PR #9, opened 2015-02-10
Applied Security patch from Magento: PATCH_SUPEE-5344_CE

manish172:


app/code/core/Mage/Admin/Model/Observer.php
```diff
@@ -69,13 +73,14 @@ public function actionPreDispatchAdmin($observer)
                     $session->login($username, $password, $request);
                     $request->setPost('login', null);
                 }
-                if (!$request->getParam('forwarded')) {
```
LeeSaferite: For the record I'd like to point out that adding this getInternallyForwarded/setInternallyForwarded is silly as the request object supports parameters that are set via code only. When you call setParam you are seting 'userland' parameters. If you want to get a parameter and be sure it was set via code and not the inbound requests you use getUserParam and NOT getParam like the original code had.

app/code/core/Mage/Oauth/controllers/Adminhtml/Oauth/AuthorizeController.php
```diff
@@ -55,7 +55,7 @@ class Mage_Oauth_Adminhtml_Oauth_AuthorizeController extends Mage_Adminhtml_Cont
      */
     public function preDispatch()
     {
-        $this->getRequest()->setParam('forwarded', true);
+        Mage::app()->getRequest()->setInternallyForwarded();
```
LeeSaferite: Why on earth are they breaking the DI by calling out to the static object???

PR #12, opened 2015-03-27
Fix catalog seo route incorrect showing disabled categories in multistore

seansan:


LeeSaferite:
We will be publishing some PR rules soon.

Until then, PRs that don't give a semi-detailed explanation and aren't totally obvious will not be merged.
If you can reply to the question and explain the logic behind this change then we will consider merging the PR.

LeeSaferite:
Closing as the PR author ( @seansan ) is not responding and providing more information.

seansan:
I thought this to be obvious

Out of the box Magento doesn't work correctly with catalog Seo map 






Sent from my iPhone.

On 17 Jun 2015, at 15:42, Lee Saferite <notifications@github.com> wrote:

Closing as the PR author ( @seansan ) is not responding and providing more information.

LeeSaferite:
@seansan Can you address the question from @drobinson and update the description of the PR?

seansan:
Unless it was updated in 1.9.1.1 - yes I see the comment, but we tested it and it does show "inactive"

Maybe it is only sitemap.phtml

Just sharing our fix as we saw this was terrible for SEO where inactive categories suddenly appeared in webmaster tools + in search results that were not meant to be there ... and if you clicked the link in Google the result was a 404. 

If somebody has more time to investgate this it would be great

I noticed that even the Magento example code online has an IsActive in the loop

http://www.magentocommerce.com/wiki/4_-_themes_and_template_customization/catalog/getting_and_using_categories_and_subcategories

LeeSaferite:
@drobinson Do you have some time to verify this PR for me?

LeeSaferite:
@seansan Thanks for responding. I'll try to make sure this gets merged.

seansan:
Probably the best way is as suggested: to alter the collection in core;oour
reasoning was to do a 'minor' update in the template file - this way we are
sure we do not break anything else. Also - if the filter is already there
then it should not matter

seansan:
Will submit new

PR #15, opened 2015-05-21
Update the hostname validation to include the latest gTLD and update …

MrGekko:
…the class to latest zend v1


Flyingmana:
Iam (currently) against this PR

1. it inserts the require statements
2. it changes the format how the gTLD get listed
3. it changes paths, which possibly is wrong

LeeSaferite:
I agree.

To fix this:
* Comment out the requires, like the original Magento version of the file
* Reformat the static list to match the existing one. This makes it trivial to see changes.
* Not sure why those file paths changed but I don't see a valid reason for it as the value is used in an 'include' call

MrGekko:
Thank you for the comments, all valid thoughts! I didnt really think this through propperly. And im fairly new to git so Ill try to be more carefull.

My new edited commit:
https://github.com/MrGekko/magento-lts/commit/e699f6c836ecc6860790095d41aae318bb5a9bd2

drobinson:
MrGekko@e699f6c looks great - can you submit a pull request for that one?

PR #16, opened 2015-05-21
Update Zend_Validate_Hostname to include the latest gTLD

MrGekko:


drobinson:
I checked - it's the exact same list as here: http://data.iana.org/TLD/tlds-alpha-by-domain.txt

+1

MrGekko:
I have found another issue:
Fix comment history with % sign.
http://www.magentocommerce.com/bug-tracking/issue/index/id/697
http://www.magentocommerce.com/bug-tracking/issue/index/id/519

But that commit got included to this pull request. How do I separate them?

Flyingmana:
use a different branch for every Pull Request, you can overwrite a branch by doing a force push.
And for future Pull Requests, always create a new branch like "patch-1", "patch-4" and so on

LeeSaferite:
Please split this into two PRs and we will accept them.

PR #17, opened 2015-06-09
Better log msg when template file doesn't exist.

tmotyl:
Before, if Magento was not able to locate a file, and error like this was logged:
2015-06-09T11:32:35+00:00 ERR (3): Warning: include(): Filename cannot be empty  in app/code/core/Mage/Core/Block/Template.php on line 241

Which is not very helpful. 
This change makes Magento log a filename and also fixes the one parameter too much in the Mage::log call.

### DIFF
--- a/app/code/core/Mage/Core/Block/Template.php
+++ b/app/code/core/Mage/Core/Block/Template.php
@@ -237,10 +237,10 @@ HTML;
 
         try {
             $includeFilePath = realpath($this->_viewDir . DS . $fileName);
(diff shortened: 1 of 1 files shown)


LeeSaferite:
+1

alexkirsch:
+1

Flyingmana:
generally ok with this, but I would like some test for this before merging, just to make sure it works as expected. Or at least try it out myself.

LeeSaferite:
I agree with @Flyingmana about the need to put a unit test requirement in place. We will have to discuss this and document it properly so people know how to contribute properly.

PR #19, opened 2015-06-17
Delete get.php.orig

davidwindell:
This shouldn't be in the repo ;)


Flyingmana:
:+1: 
sure was an accident

LeeSaferite:
As far as I can tell, this was was an artifact left over after applying the SUPEE-5994 patch from Magento. 
I will follow up with @drobinson to ensure there are no more mainline commits so things like this are visible before making it to master.

PR #20, opened 2015-06-17
Applied patch SUPEE-3762 to address XML entity escaping vulnerability

LeeSaferite:


LeeSaferite:
Magento sourced security patch

PR #22, opened 2015-06-18
Fix comment history with % sign.

MrGekko:
The issue will be solved in the upcomming 1.9.2 release of magento.
See the issues below:
http://www.magentocommerce.com/bug-tracking/issue/index/id/697
http://www.magentocommerce.com/bug-tracking/issue/index/id/519

Steps to reproduce:
Add comment to order in 'Comments History' containing '%' symbol

Expected Result:
Comment should be displayed after being saved

Actual Result:
Comment without message content is being displayed

### DIFF
--- a/app/code/core/Mage/Adminhtml/Helper/Sales.php
+++ b/app/code/core/Mage/Adminhtml/Helper/Sales.php
@@ -121,6 +121,7 @@ class Mage_Adminhtml_Helper_Sales extends Mage_Core_Helper_Abstract
     public function escapeHtmlWithLinks($data, $allowedTags = null)
     {
         if (is_string($data) && is_array($allowedTags) && in_array('a', $allowedTags)) {
+            $data = str_replace('%', '%%', $data);
             $links = array();
             $i = 1;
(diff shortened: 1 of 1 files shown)


Flyingmana:
against merging it without good review
@LeeSaferite see Skype for reason

MrGekko:
Could you tell me how I could assist in a 'good review' because this works for me, but I know that it could break other parts of the code..
So how would we go about this. 
And I am interested in the reason that you conveyed to @LeeSaferite

LeeSaferite:
@MrGekko The concern expressed by @Flyingmana over Skype was just that he didn't want to make the issue any worse and felt the issue warranted more review.

Having looked over the bug reports and the change I'm personally ok with the fix, but I think it may be prudent to wait a short amount of time to see what to official fix in 1,9,2 happens to be. I'll check the official patches to see if the issue is addressed as well.

LeeSaferite:
Also, one other suggested solution was:

Do the str_replace with the HTML equivalent &#37; 

    $data = str_replace('%', '&#37;', $data);

PR #23, opened 2015-06-19
prevent HEADERS ALREADY SENT error

Flyingmana:
$image->display() directly outputs the image together with the mimeType header,
causing the Mage_Core_Controller_Varien_Front->dispatch() to log an 
Headers already sent error on sending the actual Response object

The error occurs if magento tries to display an image in the cms wysiwyg editor mode.
The error occurs only in log, it seems to not affect the output of the image.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Cms/WysiwygController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Cms/WysiwygController.php
@@ -43,6 +43,7 @@ class Mage_Adminhtml_Cms_WysiwygController extends Mage_Adminhtml_Controller_Act
         $directive = $this->getRequest()->getParam('___directive');
         $directive = Mage::helper('core')->urlDecode($directive);
         $url = Mage::getModel('cms/adminhtml_template_filter')->filter($directive);
+        ob_start();
         try {
             $image = Varien_Image_Adapter::factory('GD2');
             $image->open($url);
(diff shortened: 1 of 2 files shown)


tmotyl:
How can I test it?

app/code/core/Mage/Adminhtml/controllers/Cms/WysiwygController.php
```diff
@@ -62,5 +63,8 @@ public function directiveAction()
             imagedestroy($image);
             */
         }
+        $this->getResponse()->setHeader('Content-type', $image->getMimeType(), true);
+        $this->getResponse()->setBody(ob_get_contents());
```
LeeSaferite: Why not use `ob_get_clean()` here and drop the next line?

LeeSaferite:
I've reviewed this PR and will merge it if you make the suggested change or explain why it's better to leave it as you originally wrote it.

drobinson:
@Flyingmana have you seen this review?

Flyingmana:
sorry for letting you wait, have made the change

LeeSaferite:
@Flyingmana I know this PR has been in the queue for a VERY long time. I finally had time to review it again this morning and I only have one request. Could you extend the fix to \Mage_Adminhtml_Cms_Wysiwyg_ImagesController::thumbnailAction as well?

Flyingmana:
next try :)
have applied it in \Mage_Adminhtml_Cms_Wysiwyg_ImagesController::thumbnailAction too

colinmollenhour:
Agh, and therein lies the problem with using one branch per version.. This PR was against 1.9.1 and apparently there are many files which conflict between 1.9.1 and 1.9.2.2 so even though it let me merge it into 1.9.1, merging 1.9.1 into 1.9.3.0 is a huge mess...

colinmollenhour:
Nevermind, looks like this PR was already in upstream 1.9.3.0 (slightly different)

PR #30, opened 2015-08-04
Update String.php

rafaelstz:


Flyingmana:
could you explain this update? What Issue does it solve?

alexkirsch:
@rafaelstz  This would break functionality as passing by reference is the only way to get that data back.

PR #31, opened 2015-08-04
Application of the correct semantics

rafaelstz:


Flyingmana:
semantic correct would be `json`, which then would need surrounding `{ }` I think

LeeSaferite:
I agree with @Flyingmana on this. I'm closing the PR, but feel free to re-open if you update.

rafaelstz:
@LeeSaferite @Flyingmana I fix it in https://github.com/OpenMage/magento-lts/pull/39

PR #33, opened 2015-08-06
Apply SUPEE-6482 patch

richdel:


richdel:
Seems it didn't apply all changes...

PR #34, opened 2015-08-07
Magento 1.9.2.1

roman204:


LeeSaferite:
Thank you for providing this PR.

I think we have solidified a process for bringing in Magento versions and @drobinson will be documenting this for future contributors.

drobinson:
Hi - here's the procedure we're following from now on: https://github.com/OpenMage/magento-lts/wiki/Procedures

In Short - we're creating a new branch for each new version release from Magento. This way we have less to deal with when considering community and Magento supplied patches (in terms of avoiding reversion of patches from one version release to another).

PR #39, opened 2015-08-11
Semantic JSON

rafaelstz:


Flyingmana:
rendering seems ok for me, so

:+1:

PR #40, opened 2015-08-18
[BUG] Fix memory leaks during image upload

masmrlar:
The image upload - especially during mass imports - leads to memory leaks due to consumed memory of the image adapter.

The fix frees the consumed image resources.

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Image.php
+++ b/app/code/core/Mage/Catalog/Helper/Image.php
@@ -639,7 +639,13 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
         }
 
         $_processor = new Varien_Image($filePath);
-        return $_processor->getMimeType() !== null;
+        $result = $_processor->getMimeType() !== null;
+
+        // free resources to prevent memory leaks
+        $_processor->close();
+        unset($_processor);
+
+        return $result;
     }
 
 }
--- a/lib/Varien/Image.php
+++ b/lib/Varien/Image.php
@@ -85,8 +85,9 @@ class Varien_Image
      * @return void
      */
     public function close() {
-        if ($this->_fileName) {
-            $this->_getAdapter()->destruct();
+        $adapter = $this->_getAdapter();
+        if ($this->_fileName && method_exists($adapter, 'destruct')) {
+            $adapter->destruct();
         }
     }
 
--- a/lib/Varien/Image.php
+++ b/lib/Varien/Image.php
@@ -54,6 +54,15 @@ class Varien_Image
     }
 
     /**
+     * Class destructor
+     * @return void
+     */
+    function __destruct()
+    {
+        $this->close();
+    }
+
+    /**
      * Opens an image and creates image handle
      *
      * @access public
@@ -68,6 +77,17 @@ class Varien_Image
         }
 
         $this->_getAdapter()->open($this->_fileName);
+    }
+
+    /**
+     * Free resources
+     *
+     * @return void
+     */
+    public function close() {
+        if ($this->_fileName) {
+            $this->_getAdapter()->destruct();
+        }
     }
 
     /**
--- a/lib/Varien/Image/Adapter/Abstract.php
+++ b/lib/Varien/Image/Adapter/Abstract.php
@@ -65,6 +65,8 @@ abstract class Varien_Image_Adapter_Abstract
 
     abstract public function open($fileName);
 
+    abstract public function destruct();
+
     abstract public function save($destination=null, $newName=null);
 
     abstract public function display();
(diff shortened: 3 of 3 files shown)


Flyingmana:
sounds reasonable to me, and I dont see a downside on this patch.
So :+1: for merging

alexkirsch:
:+1:

LeeSaferite:
The downside is that there is now an added method to the interface, potentially breaking any module that has a class implementing the changed interface.

While I greatly prefer defining things like this in an interface, in an effort to no break other modules I suggest using a method_exists call instead of changing the interface. If the interface is not modified then I see no problem with the rest.

LeeSaferite:
Thinking about it a little more, I think dropping Varien_Image::close() is better as well. The __destruct method is called when there are no more references to the object or when PHP shuts down. The code in the close() method could directly live in __destruct and the call to close() followed by the unset() could be simplified as a simple unset() which will trigger the __destruct call.

That change plus using `method_exists($this->_getAdapter(), 'destruct')` instead of the interface change would be best I think.

Flyingmana:
you are right, did not look carefully enough to see the changed interface.

the `close()` call could still be needed, if the memory leak is part of the inner image extension, which does not clean up correctly via destruct.

masmrlar:
Yes, you are right - changing interfaces is really bad habid - going to fix that.

Thanks for your hints.

LeeSaferite:
Sorry for the long delay.

I still say that adding a close() method is not needed. You can move all of that code into the __destruct call directly. Also, no need to manually call unset() prior to returning as once the variable leaves scope, due to the method return, it will automatically be destroyed.

app/code/core/Mage/Catalog/Helper/Image.php
```diff
@@ -639,7 +639,13 @@ public function validateUploadFile($filePath) {
         }
 
         $_processor = new Varien_Image($filePath);
-        return $_processor->getMimeType() !== null;
+        $result = $_processor->getMimeType() !== null;
```
LeeSaferite: None of this is needed. When the method returns the $_processor object will be destroyed automatically and trigger the __destruct() method. That method currently will trigger a call to close() you added and really should just have the close() code directly embedded.

lib/Varien/Image.php
```diff
@@ -71,6 +80,18 @@ public function open()
     }
 
     /**
+     * Free resources
+     *
+     * @return void
+     */
+    public function close() {
```
LeeSaferite: No reason to add this method. The contained code should live directly in __destruct() instead.

lib/Varien/Image.php
```diff
@@ -54,6 +54,15 @@ function __construct($fileName=null, $adapter=Varien_Image_Adapter::ADAPTER_GD2)
     }
 
     /**
+     * Class destructor
+     * @return void
+     */
+    function __destruct()
```
LeeSaferite: The whole fix for this issue should live here really.

ktomk:
unset as well as leaving scope in a method / function will call `__destruct()` in PHP. Can only confirm and underline that fix should go into there then. I will try to run a test on one install I have where there is a problem with the memory on media import.

drobinson:
Any update on this one?  Otherwise I'll close pending a new pull request...

masmrlar:
Nope, no update - but you can fix the fix if you like.

ktomk:
I created a fresh PR: #61 - It's adding the dtor and it does not touch the public interface. of the abstract class.

A little write up why we see a problem here:

The original developer of **Varien_Image_Adapter_Gd2** was just that bit too clever we all feel ashemd of if we see this in our own code years later.

Within the *ctor* (!) of it, a self-reference of *every* instance is created effectively preventing it from unloading:

    register_shutdown_function(array($this, 'destruct'));
                                     ^^^^^

Which means that memory will effectively filled until either the script normally exists or perhaps crashes due to not enough memory.

Just in a quickly mocked case of a class that only registers such a shutdown method:

	class SkyHigh
	{
		static $count = 0;

		public function __construct()
		{
			register_shutdown_function(array($this, 'destruct'));
		}

		public function destruct() {
			printf("%s - %d\n", number_format(memory_get_usage()), ++self::$count);
		}
	}

And then creating instances of it in a blatant manner:

	$count = 30000;

	echo "staring with $count \n";

	while($count--) {
		new SkyHigh();
	}

Gives me a memory use of ~23 megabyte (this will vary across PHP versions I suppose, my test was with PHP 5.6). Because of the self-reference, the instances are kept in memory without destruction. And that is a shallow class only (!), imagine the one in Magento has plenty of properties and might even keep images data in memory.

That being said, part of the memory consumption problem is still unfixed because the instances remain in memory. Only the image handles are destroyed.

LeeSaferite:
See #61

PR #57, opened 2015-11-20
Limit Collection Loads When Getting One Item

kojiromike:
I couldn't immediately think of any compelling reason not to do this.

### DIFF
--- a/lib/Varien/Data/Collection.php
+++ b/lib/Varien/Data/Collection.php
@@ -265,14 +265,19 @@ class Varien_Data_Collection implements IteratorAggregate, Countable
      */
     public function getFirstItem()
     {
+        $prevPageSize = $this->getPageSize();
+        $this->setPageSize(1);
         $this->load();
 
         if (count($this->_items)) {
             reset($this->_items);
-            return current($this->_items);
+            $result = current($this->_items);
+        } else {
+            $result = new $this->_itemObjectClass();
         }
 
-        return new $this->_itemObjectClass();
+        $this->setPageSize($prevPageSize);
+        return $result;
     }
 
     /**


kojiromike:
@drobinson honestly, I have not measured it expressly, but it is certainly documented in enough places. I always wondered why Magento didn't build this in.

As for `getLastItem`, that's definitely trickier. In `getFirstItem`, if the collection is already loaded, then it won't be re-loaded. If I do as you suggest and reverse the order I would need to force the collection to re-load.

Another approach would be to use `getLastPageNumber`, but that involves `getSize`, which is another whole (albeit cheaper) query. It would make getting the last member of very large collections faster, but it would make getting the last member of small collections slower.

```
diff --git a/lib/Varien/Data/Collection.php b/lib/Varien/Data/Collection.php
index c7ca088..28dadc4 100644
--- a/lib/Varien/Data/Collection.php
+++ b/lib/Varien/Data/Collection.php
@@ -287,13 +287,22 @@ class Varien_Data_Collection implements IteratorAggregate, Countable
      */
     public function getLastItem()
     {
-        $this->load();
+        $prevPageSize = $this->getPageSize();
+        $prevPage = $this->getCurPage();
+        $this
+            ->setPageSize(1)
+            ->setPage($this->getLastPageNumber());
+            ->load();
 
         if (count($this->_items)) {
-            return end($this->_items);
+            $result = end($this->_items);
+        } else {
+            $result = new $this->_itemObjectClass();
         }
-
-        return new $this->_itemObjectClass();
+        $this
+            ->setPageSize($prevPageSize)
+            ->setPage($prevPage);
+        return $result;
     }
 
     /**
```

Flyingmana:
there is a good reason, it changes the state of the collection in an often unintended way. You dont know how many first fetch the first item, and next iterate over the collection, or do other things with it.

It may break some code, which is not obvious for the user and may need a lot of time till noticed, and also to analyze it.

LeeSaferite:
I agree with Daniel. If someone just wants the first item then they would set the limit to one themselves. Trying to guess the usage pattern and doing something unexpected for other use cases is not a good idea.

So, while interesting and while I appreciate the PR, my vote is to decline this PR.

kojiromike:
I expected this resistance to merging it, but I think it still needed to be brought up. The way that collections hang on to the state of being "loaded" with one select without being aware if the select has changed makes for a conflict of interest between optimization and usability. So close away if you think it's the right thing to do, and have a nice day.

But I submit that there are very few unaware users that are using _this_ repo as opposed to vanilla Magento. I bet most users of this repo are at least marginally aware of this flaw with `getFirstItem`, since it comes up any time you do any kind of linting of Magento code. In @Flyingmana's example, if you intend to iterate over the entire collection anyway, you should probably not use `getFirstItem` to begin with.

Flyingmana:
I admit, that users of "this repo" know more about internals. But that should not lead to increase the difficult level when less experienced users want to switch. Also there is still a big number of extensions.

And I would say, "this repo" has a goal to be an acceptable alternative to the main source for magento, especially as the support from magento1 ends in ~3years.

PR #59, opened 2015-11-25
Fixes multistore problem where categories are not active in every storeview

seansan:
SNH 25-11-2015 Fixes multistore problem where categories are not active in every storeview

Can be applied to every version


Flyingmana:
not tested from my side, but it seems to make sense to add this filter.

drobinson:
+1

PR #60, opened 2016-01-17
Improve extensibility of get.php file

thaiphan:
Rather than directly calling the database storage model, we use the storage helper to call it instead. This makes it easier to customise how the file storage system works.

### DIFF
--- a/get.php
+++ b/get.php
@@ -143,11 +143,13 @@ if (0 !== stripos($pathInfo, $mediaDirectory . '/')) {
 }
 
 try {
-    $databaseFileSotrage = Mage::getModel('core/file_storage_database');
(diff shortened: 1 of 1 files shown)


LeeSaferite:
While I understand the desire, I'm not sure this PR does what you want. All you are doing is abstracting the Mage:getModel call into the helper. Wouldn't it be better to make use of the 'core/file_storage' helper instead?

drobinson:
I think @thaiphan is saying it would be less intrusive to rewrite the helper as opposed to the model to get what he wants done.  If I'm not mistaken, wouldn't using 'core/file_storage' also require a helper rewrite to redefine the the storage model used?

PR #62, opened 2016-01-29
PHP 7 support

drobinson:
Credit to Inchoo (https://github.com/Inchoo/Inchoo_PHP7) for the work - this is just a core implementation of the PHP 7 support module.

### DIFF
--- a/app/code/core/Mage/Core/Model/Layout.php
+++ b/app/code/core/Mage/Core/Model/Layout.php
@@ -552,7 +552,7 @@ class Mage_Core_Model_Layout extends Varien_Simplexml_Config
         $out = '';
         if (!empty($this->_output)) {
             foreach ($this->_output as $callback) {
-                $out .= $this->getBlock($callback[0])->$callback[1]();
+                $out .= $this->getBlock($callback[0])->{$callback[1]}();
             }
         }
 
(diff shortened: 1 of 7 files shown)


Flyingmana:
in generell I support this changes, they dont have a major downside.

But please explain the changes of `app/code/core/Mage/Sales/etc/config.xml`

MrGekko:
I am using this module to get PHP 7 support, so I support adding this.

The changes to `app/code/core/Mage/Sales/etc/config.xml` where orriginally made by:
> @durzel make sorting more consistent
https://github.com/danslo/Rubic_FixTotalSorting/commit/086ca9941285cdcbc2a347086148b28cd381b5a6

Hope that helps...

Flyingmana:
I see, than I would like to see this excluded from the PR, as it is not related to php7 support, so it can/needs to be discussed separate.

drobinson:
I made an assumption that PHP 7 also rewrote the sorting algorithm to make it stable (and therefore required this explicit calculation ordering) but I did not test it without this. I'll check it out.

durzel:
PHP7 does change the sort behaviour. In my experience on an install that otherwise had no bespoke sorting code moving from PHP 5.3.3 to PHP 7 resulted in a calculation issue in the cart. 

This is discussed here:
http://stackoverflow.com/questions/34281113/the-different-behavior-of-the-function-uasort-in-php-5-5-and-php-7-0

..among other places.

Flyingmana:
ok, that makes sense. In this case I give my ok. 
:+1:

seansan:
Just to confirm - this Magento LTS branch 1.9.2.4 has php 7 support?

tmotyl:
True. I'm using it on some projects already.

luchiano1:
1

PR #64, opened 2016-02-06
FIX. Enable re-sending again - http://magento.stackexchange.com/a/376…

seansan:
…90/78

FIX. Enable re-sending again - http://magento.stackexchange.com/a/37690/78

Irritating "new" feature so order emails cannot be resent ... fixed

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Sales/OrderController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Sales/OrderController.php
@@ -136,6 +136,8 @@ class Mage_Adminhtml_Sales_OrderController extends Mage_Adminhtml_Controller_Act
     {
         if ($order = $this->_initOrder()) {
             try {
+                // FIX. Enable re-sending again - http://magento.stackexchange.com/a/37690/78
+				$order->setEmailSent(false);
                 $order->sendNewOrderEmail();
                 $historyItem = Mage::getResourceModel('sales/order_status_history_collection')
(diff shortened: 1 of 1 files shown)


LeeSaferite:
I think the better solution is to switch to `queueNewOrderEmail(true)`

seansan:
As long as the problem is fixed ;P Can you create a pull request and add it?

shirtsofholl:
Can this be accepted?

diglin:
+1 for the @shirtsofholl solution

app/code/core/Mage/Adminhtml/controllers/Sales/OrderController.php
```diff
@@ -136,6 +136,8 @@ public function emailAction()
     {
         if ($order = $this->_initOrder()) {
             try {
+                // FIX. Enable re-sending again - http://magento.stackexchange.com/a/37690/78
+				$order->setEmailSent(false);
```
colinmollenhour: Tabs vs spaces... We gotta have some standards.. :)

seansan: Agree .... +1 .. just get this one in ;)

colinmollenhour: Preferred way to fix is while on your PR branch just replace the tabs with spaces, use `git commit -a --amend` to fix the existing commit and then `git push -f` to force-push over the old commit.

colinmollenhour:
Merged 1.9.2.3 into 1.9.3.0

PR #65, opened 2016-02-09
Standardize toOptionArray() method

sreichel:
There are some models that use `getOptionArray()` instead of `toOptionArray()` making it impossible to define a source model in system.xml like `<source_model>catalog/product_visibility</source_model>`


Flyingmana:
Instead of renaming real methods, I would suggest to only add an alias to them.
This should also reduce the amount of changed lines of code by a lot.

LeeSaferite:
@sreichel I agree with @Flyingmana here. Better to limit the scope of the changes by defining an alias instead, no?

LeeSaferite:
@sreichel That you for the PR, but due to the lack of feedback on the suggested changes I'm closing this. If you'd like to make the changes suggested and resubmit then I'll review and likely merge.

PR #68, opened 2016-02-29
Fix bug with wrong shipping cost

avstudnitz:
I have confirmed this to be a bug, as assumed in the original comment:
```
                    } else {
                        /**
                         * possible bug: this should be setBaseShippingAmount(),
                         * see Mage_Sales_Model_Quote_Address_Total_Shipping::collect()
                         * where this value is set again from the current specified rate price
                         * (looks like a workaround for this bug)
                         */
                        $this->setShippingAmount($rate->getPrice());
```
It caused the wrong shipping cost appear in the shopping cart - including tax instead of excluding tax as configured. The bug is fixed by using BaseShippingAmoun.


LeeSaferite:
@avstudnitz So, is the correct solution to remove the code or use the proper setter?

LeeSaferite:
@avstudnitz No feedback on my question?  Should I close the PR?

PR #70, opened 2016-03-04
suggested improvement - everybody does it why not add it

seansan:
suggested improvement - everybody does it why not add it

### DIFF
--- a/index.php
+++ b/index.php
@@ -56,7 +56,10 @@ if (!file_exists($mageFilename)) {
     exit;
 }
 
-if (file_exists($maintenanceFile)) {
+$ip = $_SERVER['REMOTE_ADDR'];
+$allowed = array('127.0.0.1'); // replace with your IP
+ 
+if (file_exists($maintenanceFile) && !in_array($ip, $allowed)) { {
     include_once dirname(__FILE__) . '/errors/503.php';
     exit;
 }


drobinson:
One reason would be that it won't work universally - many projects are behind some kind of load balancer or a firewall like Cloudflare - in both cases REMOTE_ADDR won't be the end-user.

Even if you changed it so that it first checked HTTP_X_FORWARDED_FOR it still probably isn't a great idea - the maintenance flag serves more of a purpose than just not showing the frontend when something  is happening.  You could for instance trigger install scripts before they should be triggered, causing all sorts of issues.

seansan:
Agree - but then it would just not work and not break anything. Also
wouldnt they use EE?

drobinson:
IMO "It won't break anything" is not a good enough reason to add code.

EE isn't universally adopted by merchants who this would not work for.

I'm curious - what purpose does this change serve for your projects?

seansan:
none - just a simple - possible - addition. Not necessary at all

LeeSaferite:
I have done similar in the past, but I agree with @drobinson here that it's not a change I can see the value in. The possibility that it introduces a bug in someone's setup is enough reason to at least look at it critically.

Flyingmana:
Possible Bug Case:

1. you activate maintainance
2. your local monitoring/Nagios/Icinga/anything accesses pages
3. tears :crying_cat_face:

LeeSaferite:
Based on the feedback I think it's appropriate to close this PR.  Thank you for the PR in any case.

PR #72, opened 2016-03-16
Solves configurable products dropdown sort order not being respected

seansan:
Solves configurable products dropdown sort order not being respected

http://stackoverflow.com/questions/27270918/magento-1-9-1-not-sorting-configurable-product-attributes-dropdown-by-position

Please verify

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Type/Configurable/Attribute/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Type/Configurable/Attribute/Collection.php
@@ -302,6 +302,28 @@ class Mage_Catalog_Model_Resource_Product_Type_Configurable_Attribute_Collection
                 }
             }
 
+            /**
+            * Mage 1.9+ fix for configurable attribute options not sorting to position
+            **/
+            $sortOrder = 1;
+            foreach ($this->_items as $item) {
+               $productAttribute = $item->getProductAttribute();
+               if (!($productAttribute instanceof Mage_Eav_Model_Entity_Attribute_Abstract)) {
+                   continue;
+               }
+               $options = $productAttribute->getFrontend()->getSelectOptions();
+               foreach ($options as $option) {
+                   if (!$option['value']) continue;
+                   if (isset($values[$item->getId() . ':' . $option['value']])) {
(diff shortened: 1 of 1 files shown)


seansan:
Fixes our problem and using in production

seansan:
Just had this problem again on a new installation - I think it is important to accept this (or a similar solution, but sorting really needs to work)

drobinson:
Hi @seansan - sorry you haven't heard anything on this yet.  I'm testing this out today and will hopefully have it merged shortly after.

drobinson:
I've verified this works, but to avoid looping again, I ended up with this patch (so far):

```
@@ -272,11 +272,16 @@ class Mage_Catalog_Model_Resource_Product_Type_Configurable_Attribute_Collection
                                 'store_label'                => $optionsByValue[$optionValue],
                                 'is_percent'                 => 0,
                                 'pricing_value'              => null,
-                                'use_default_value'          => true
+                                'use_default_value'          => true,
+                                'order'                      => array_search($optionValue, array_keys($optionsByValue))
                             );
                         }
                     }
                 }
+
+                usort($values, function($a, $b) {
+                   return $a['order'] - $b['order'];
+                });
             }

             foreach ($pricings[0] as $pricing) {
```

I still don't feel fuzzy, since this is happening in the _loadPrices method.  However, the only reason the ordering is incorrect is because of how this method loads the options in the first place...so I think this is probably OK.  What do you think?

seansan:
I understand the discussion. Maybe @colinmollenhour can also take a peek at this one. 

But realize the impact: you sell T-shirts and on the frontend it looks like

Small
Extra large
Medium
Large
Extra Small

colinmollenhour:
I think in a perfect world the `getUsedProducts` method would return items already properly sorted, but that looks pretty complicated to implement and @drobinson's modified version of the patch looks very efficient. Besides, sometimes sorting application-side is a lot faster if it saves an expensive join in the database. Nice work.

seansan:
Not being the local PHP , nor github expert .. would you be abel to shoot in a new PR?

drobinson:
New PR https://github.com/OpenMage/magento-lts/pull/114

Closing this one..

PR #75, opened 2016-03-23
I think it is important to mention separately that php 7.0 support wa…

seansan:
…s added

I think it is important to mention separately that php 7.0 support was added

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,8 @@ This allows you to define your version dependencies safely in composer.json:
 "openmage/magento-lts": "1.9.2.1-dev"
 ```
 
+## Important to note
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +34,8 @@ This allows you to define your version dependencies safely in composer.json:
 "openmage/magento-lts": "1.9.2.1-dev"
 ```
 
+## Important to note
+PhP 7 support was added as as from version 1.9.2.3. based on the Inchoo article and [described here](https://github.com/OpenMage/magento-lts/pull/62).
```
tmotyl: should be PHP, not PhP

LeeSaferite:
Please change the PhP to PHP as suggested by @tmotyl and I'll merge this PR

seansan:
should be changed

Although on this minor changes: 1 letter there is maybe a lot of overhead
with emails and calls en replies: I suggest one just changes it when merging

PR #77, opened 2016-04-11
Fix cache : Config Cache Corruption Fix - 100 Router Match Iterations…

seansan:
… - Object Caching and Race Conditions

Config Cache Corruption Fix - 100 Router Match Iterations - Object Caching and Race Conditions

https://github.com/convenient/magento-ce-ee-config-corruption-bug/blob/master/IMPROVEMENTS.md

**Please check my code additions**

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -176,6 +176,14 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     protected $_canUseLocalModules = null;
 
     /**
+     * Flag which is used to prevent cache corruption 
+     * Coding suggestion comes from: https://github.com/convenient/magento-ce-ee-config-corruption-bug/blob/master/IMPROVEMENTS.md
+     *
+     * @var bool
+     */
+    protected $allowSaveCache = false;
+
+    /**
      * Active modules array per namespace
      * @var array
      */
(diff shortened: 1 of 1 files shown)


seansan:
Our server ran into this issue last night: quite terrible!

LeeSaferite:
While I like the idea, I take issue with how you made the changes. You've put the proprty in the middle of the class and even worse you split the method docblock away from the method to do this.

Please clean up the PR by moving the new property into the top of the class. Also, I think having a single link to the ratiional behind the fix is sufficient. I would suggest you attach a good description and link to the docblock for the property itself.

If you can clean up the PR I'll review it and have some others do the same. Assuming it doesn't break things (doesn't initially look like it would) then I'll merge the PR.

Thanks!

seansan:
Please check again

https://github.com/OpenMage/magento-lts/pull/77/files

seansan:
@LeeSaferite would be great to merge

LeeSaferite:
@drobinson @Flyingmana Do you guys see any issues here? Can I merge this?

tmotyl:
I had this fix (https://github.com/convenient/magento-ce-ee-config-corruption-bug/blob/master/IMPROVEMENTS.md) applied in one project and it fixed the issue.

seansan:
Posting an easier and more confirmed fix

tmotyl:
Maybe the chmod was a soultion in that particular case, but the race condition issue still exists.

seansan:
Hmmm may be true. Did you find an acceptable fix?






Sent from my iPhone.

On 16 Aug 2016, at 09:57, Tymoteusz Motylewski <notifications@github.com> wrote:

Maybe the chmod was a soultion in that particular case, but the race condition issue still exists.

tmotyl:
see above - https://github.com/convenient/magento-ce-ee-config-corruption-bug/blob/master/IMPROVEMENTS.md

seansan:
should we add this as a pull request?

seems 2 changes: 1 fix, 1 improvement

PR #81, opened 2016-04-26
Fix typo in sales/order model _afterSave() method

KBortnick:
Getter in if statement has a typo which causes it to attempt to grab the wrong attribute. Since result is null, the comparison always fails, and shipping_address_id attribute is always flagged for update when Order model is saved.


Flyingmana:
:+1:
I remember this bug, always thought there was some odd reason behind this. Worth to be fixed.

PS: did not validate the code yet, but if the corrected one points to the right attribute, Iam for merging this

PR #83, opened 2016-05-03
fix jsonDecode() PHP7 compatibility

pschnell:
When saving a customer in the admin panel, Magento always throws an error regarding `jsonDecode` - `Exception message: Decoding failed: Syntax error`.

I think `jsonDecode` should be changed to the one in the Inchoo_PHP7 extension [here](https://github.com/Inchoo/Inchoo_PHP7/blob/master/app/code/local/Inchoo/PHP7/Helper/Data.php) instead, which looks like this and has that switch statement added:

    public function jsonDecode($encodedValue, $objectDecodeType = Zend_Json::TYPE_ARRAY) {
        switch (true) {
            case (null === $encodedValue)  : $encodedValue = 'null'; break;
            case (true === $encodedValue)  : $encodedValue = 'true'; break;
            case (false === $encodedValue) : $encodedValue = 'false'; break;
            case ('' === $encodedValue)    : $encodedValue = '""'; break;
            default    : // do nothing
        }
        return Zend_Json::decode($encodedValue, $objectDecodeType);
    }

### DIFF
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -656,6 +656,14 @@ XML;
      */
     public function jsonDecode($encodedValue, $objectDecodeType = Zend_Json::TYPE_ARRAY)
     {
+        switch (true) {
+            case (null === $encodedValue)  : $encodedValue = 'null'; break;
+            case (true === $encodedValue)  : $encodedValue = 'true'; break;
+            case (false === $encodedValue) : $encodedValue = 'false'; break;
+            case ('' === $encodedValue)    : $encodedValue = '""'; break;
+            default    : // do nothing
+        }
+        
         return Zend_Json::decode($encodedValue, $objectDecodeType);
     }
 


Flyingmana:
Looks ok for me and without downsides.

also the same as this stackoverflow answer suggests http://stackoverflow.com/questions/36362856/php7-json-and-zend-decoding-failed-syntax-error

have not validated it, but I think its ok to merge

:+1:

LeeSaferite:
@drobinson Can you look at this one since you did the original PHP7 support based on the Inchoo module?  There look to be several changes since PHP7 support was added in commit a07ad9c65d0b0235bcc19792f0035911f306795d

drobinson:
Looks good to me.  There are a couple more interesting changes made over there - I'll create issues for us to look at them. Most of the other commits are whitespace/documentation.  

This one is https://github.com/Inchoo/Inchoo_PHP7/commit/63976858e94793c4fdd4bd0040a1cd875fdef898

Issue #80, opened 2016-04-14
Twice tax in Grand Total


idziakjakub:
After upgrade Magento from 1.9.2.2 to 1.9.2.4 I have doubled tax amount in Quote.

I have found solution for this issue: 
http://stackoverflow.com/questions/17590364/magento-applying-tax-vat-twice/35870460#35870460

The problem occurs when you have additional totals in project.
The problem can be solved by adding correct dependencies on tax subtotal, so Magento keep the correct order.

tmotyl:
+1

tmotyl:
fixed in https://github.com/OpenMage/magento-lts/commit/b9726b9997e2bece26d37d91c92f32fbb18aebb4

hannes011:
This is a major change which is causing troubles for the calculation of shipping rate! Is "shipping" really needed as prerequisite for subtotal?
A: shipping is not included within subtotal
B: some shipping methods calculate there rate based on subtotal

This change led to strange behavior: once the totals are already calculated and stored/cached then the subtotal doesn't include tax anymore (as expected) BUT when totals will be collected first time the "shipping" total is executed before "tax_subtotal", so the subtotal still contains tax, which leads to inconsistency...
![image](https://user-images.githubusercontent.com/12843419/87032376-d21c5d80-c1e4-11ea-81f8-af7ee1fa3656.png)

FYI @tmotyl @sreichel @idziakjakub

Schrank:
I saw it and maybe this is important for this: https://blog.fabian-blechschmidt.de/mage_weee-and-why-it-is-important-for-tax-calculation/
(didn't dig into it)

Schrank:
Maybe you want to include https://github.com/ikonoshirt/TopologicalSearch into OpenMage, happy if you do! <3

hannes011:
Simple fix would be just this I guess:
```diff --git a/app/code/core/Mage/Tax/etc/config.xml b/app/code/core/Mage/Tax/etc/config.xml
--- a/app/code/core/Mage/Tax/etc/config.xml
+++ b/app/code/core/Mage/Tax/etc/config.xml
@@ -162,7 +162,7 @@
                 <totals>
                     <tax_subtotal>
                         <class>tax/sales_total_quote_subtotal</class>
-                        <after>subtotal,nominal,shipping,freeshipping</after>
+                        <after>subtotal,nominal,freeshipping</after>
                         <before>tax,discount</before>
                     </tax_subtotal>
[...]

PR #84, opened 2016-05-06
Fixed tax subtotal calculation order

idziakjakub:
https://github.com/OpenMage/magento-lts/issues/80

This change fix doubled tax amount in Quote.

### DIFF
--- a/app/code/core/Mage/Tax/etc/config.xml
+++ b/app/code/core/Mage/Tax/etc/config.xml
@@ -162,7 +162,7 @@
                 <totals>
                     <tax_subtotal>
                         <class>tax/sales_total_quote_subtotal</class>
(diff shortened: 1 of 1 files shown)


ADDISON74:
**This change was a bad idea.** It creates an issue with "Free Shipping" feature. For more details check this link #1813.

Maybe this PR really solves this issue #1104.

ADDISON74:
From what I read above in the comments in 2019 an issue was reported that orders cannot be created by the administrator in the Backend and that Free Shipping is no longer displayed in the shopping cart. After 2 years from the post no maintainer took the reporting into account. Interestingly, this commit is still in all Magento LTS and OpenMage variants. I ask myself the natural question, how many such wrong changes could exist in the source codes?

PR #85, opened 2016-05-12
is_readabe fix

drashmk:
Added @ on ``is_readable`` to suppress warning thrown if ``$src`` is file content, not a path (e.g. when image is updated through the REST API) to prevent failure when the PHP server is set to stop on warning like on development server


tmotyl:
+1

sreichel:
Not sure if supress error is best idea ...

Untested, but a `file_exists()` check before c/should cover this.

manezaf8:
Is there a way of stopping the `is_readable()` warning on magento 1 though other than suppressing the error with @ ?

And what is causing it ?

PR #86, opened 2016-05-16
Replace font LinLibertineFont with FONT_HELVETICA: reduces PDf filesi…

seansan:
…zes almost 20x fold

Replace font LinLibertineFont with FONT_HELVETICA: reduces PDf filesizes almost 20x fold

Also officialy logged in Magento 1.9x stream

Also HELVETICA is a standard font (instead of loading a lib file)

$font = Zend_Pdf_Font::fontWithName(Zend_Pdf_Font::FONT_HELVETICA);


seansan:
HUGE! file size improvement. Use case: suppopse you send along the invoice with an order. Knowing the basic PDF layout (some tables, data) how could this ever become 1-2mb????

tmotyl:
+1 by reading

Flyingmana:
also picked in 1.9.3.1

PR #95, opened 2016-07-14
Add progressive to image resize

seansan:
Add progressive to image resize

Taken from; http://magento.stackexchange.com/questions/40865/google-pagespeed-image-cache/57788#57788

### DIFF
--- a/lib/Varien/Image/Adapter/Gd2.php
+++ b/lib/Varien/Image/Adapter/Gd2.php
@@ -194,6 +194,12 @@ class Varien_Image_Adapter_Gd2 extends Varien_Image_Adapter_Abstract
             $functionParameters[] = $this->quality();
         }
 
+		// make jpegs progressive and improve product file size (when resizing without changing the library)
+		$interlaceIt = 0;
+		if ($this->_fileType == IMAGETYPE_JPEG) {
+			$interlaceIt = imageinterlace($this->_imageHandler, 1);
+		}
+
         // set quality param for PNG file type
         if (!is_null($this->quality()) && $this->_fileType == IMAGETYPE_PNG)
         {


Flyingmana:
Hi, thank you for this patch.
But I read a bit into interlacing, and it seems only be an advantage for big jpg, but not for example for png, or smaller jpg (thumbnails).
But I have this mostly from http://stackoverflow.com/questions/13449314/when-to-interlace-an-image

Maybe there is someone else with more in detail knowledge or actual practice experience?

seansan:
I think where this is most important is when using large product images > 1200px 1600px (500Kb files) of 'real' products often taken in jpeg

colinmollenhour:
How about only interlacing if the original and target size are over some threshold such as 1 megapixels?

seansan:
Just to be clear the interlacing mosty (qua name) applies to GIF and PNG ... JPG calls it progressive. The idea is the same. 

The problme with findng a threshold is: what;s your bandwidth? Is it necessary

Read;
http://yuiblog.com/blog/2008/12/05/imageopt-4/

But follow me on this
- if you have high high high speed internet
- suppose interlacing is on: who cares?

- if you have slow internet
- you would be glad

So I dont really se the point. I would progressively optimize all product images over 50Kb I think/ ot 100Kb.... Suppose you are on an iphone in the tube .....

colinmollenhour:
Seems the main drawback is that it takes a bit longer to process progressive. For a single image this is not a big deal but if you invalidate your entire image cache by changing something like the watermark and you are in a high-traffic scenario then it can really stress the server because for a while each page load could be doing dozens of image resizes and then there is stampeding, etc.. Ask me how I know.. :)

Since there are pros and cons and it is likely that not everyone will agree I would suggest making it configurable. E.g. a configuration like "Enable Progressive Images Threshold" with a comment: "Specify a non-zero size in megapixels to enable progressive image compression. Smaller images will not use progressive compression. E.g. 0.5 will enable progressive mode for images equal to or larger than half a megapixel.". I would suggest to expose this flag in a new public method and set it from the catalog/product_image model so that other image adapters could support this option as well.

The reason to express the threshold in megapixels instead of bytes is because bytes is not known for the target image beforehand, only the source image. The other option would just be to make it a Yes/No which I think would also be acceptable.

colinmollenhour:
@seansan What do you think about adding the ""Enable Progressive Images Threshold" config option to your PR? I think everyone would be satisfied then.

seansan:
Fine. But not sure how to do it. And should default to false not to break
previous versions

colinmollenhour:
if ($threshold <= imagesx($this->_imageHandler) * imagesy($this->__imageHandler))

seansan:
yeah thanks that part is clear. I do small fixes, but dont know how to add a new config option in backend, make sure it is added, retrieve it, etc etc

colinmollenhour:
Ahh, just add the field to the appropriate location in system.xml and use Mage::getStoreConfig() to retrieve the value. That's all there is.

sreichel:
@colinmollenhour what is the "appropriate location"?

colinmollenhour:
app/code/core/Mage/Catalog/etc/system.xml at config/sections/catalog/groups/product_image/fields/progressive_threshold

Another place that would be appropriate is a third-party extension such as https://github.com/magento-hackathon/Perfect_Watermarks

sreichel:
Please close.

PR #96, opened 2016-07-14
Add visibility and qty in stock to product selection grid in backend order creation

seansan:
Add new fields to collection

doesnt break anything just makes more logic to be able to filter your grid based on stock or visibility. FOr instance one could choose to show only visibility=4 so we see the same products as in the frontend

See also here: http://magento.stackexchange.com/questions/17566/magento-add-stock-qty-in-backend-new-order-creation/17898#17898

Optionally we could add the column per default to the grid: although I think this is quite logic and should be in core. Others would maybe not agree. So I will leave that up to you.

```
    if (Mage::helper('catalog')->isModuleEnabled('Mage_CatalogInventory')) {
        $this->addColumn('inv_qty',
        array(
        'header'=> Mage::helper('catalog')->__('Qty'),
        'width' => '30px',
        'type'  => 'number',
        'index' => 'inv_qty',
        ));
    }
```

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Search/Grid.php
@@ -101,6 +101,7 @@ class Mage_Adminhtml_Block_Sales_Order_Create_Search_Grid extends Mage_Adminhtml
             ->setStore($this->getStore())
             ->addAttributeToSelect($attributes)
             ->addAttributeToSelect('sku')
+            ->addAttributeToSelect('visibility')            
             ->addStoreFilter()
             ->addAttributeToFilter('type_id', array_keys(
                 Mage::getConfig()->getNode('adminhtml/sales/order/create/available_product_types')->asArray()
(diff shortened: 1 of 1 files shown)


drobinson:
In my experience, every client wants something different for these grids.  In my opinion, it probably makes sense to keep these types of changes in client-specific modules using rewrites or something.

seansan:
Inventory and Visibility are quite necessary

Inventory to see if it is there
Visibility to filter on frontend salable products only

(for instance when you have configurable with 8 underlying simples in a 10.000 products dataset: you want only the configurables: not the simples too)

colinmollenhour:
@drobinson I agree with you on the main product grid but for the Add Products grid on order create page I don't see the harm in adding the inventory column since it actually affects the order unlike something arbitrary like "color" or "size". Definitely subjective. I think it'd be great to make all of this configurable, but I'm not volunteering. :)

seansan:
I would even add visibiltiy as a column. In most cases our (backend) store via the telephone sells ofcourse the products that you see on the frontend .... not the disables, views individually and all other products that are non-existant on the frontend

tmotyl:
what about the performance impact of this change? maybe we should make a setting for disabling this feature?

seansan:
is there a big impact? one could argue that fronted is always fastest ...
and there it is in the collection

tomekjordan:
QTY availalbe is a must. visibility - I don't see any reason to add this in backorder view. Merchant knows what he sells.

seansan:
Both are important based on several use cases - esp when entering orders via telephone for customers or manual (live) entry

There is no real impact as it is limited and uses index. 

Making configurable might be a solution. 

+1

seansan:
Doesnt solve it though for configurables ... reverting changes ... for simples would be handy

PR #97, opened 2016-07-14
If add qty field is empty set a default vallue of 1

seansan:
If add qty field is empty set a default vallue of 1

Solution is maybe not the most elegant but makes sure we don't overwrite existing values. Only oly if the fields is empty set a default value. A an improvement one could get the value from the mimimal_order_qty settings in settings (this is the way you set a default order qty value in frontend instead of leaving it empty)

please imagine use entering orders every day via adminhtml and every every time this fields is empty

a starting value of 1 already solves 95% of these cases

Major improvement for us, please accept

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/composite/fieldset/qty.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/composite/fieldset/qty.phtml
@@ -34,3 +34,13 @@
         </dl>
     </div>
 </div>
+
+<script type="text/javascript">
+    //<![CDATA[
+    var input = document.getElementById('product_composite_configure_input_qty');
+    if(input.value.length == 0) { 
+        input.value = "1";
+    }
+    //]]>
+</script>
+


drobinson:
Can you clarify where you're running into the issue this solves?  I can't find a place in the admin where I'm required to enter a quantity after selecting products.

shirtsofholl:
Configurable products via admin

seansan:
Simple win - compare it to: the situation where you have to change the default min qty in the config to 1, so that Magento auto sets a product qty to 1 on the frontend. In this case when you add a config product to an order in backend the qty field is empty when it bothers you to ask for a size (read super attribute). But in this case I have already selected the products ... so yeah! at least I want 1 piece ....

Flyingmana:
tested it and works.

actually in 1.9.3.1 I could even with empty field add them and it added them as "1" with mostly default product setups. But still, it cant harm, so it gets added

also picked for 1.9.3.1

PR #99, opened 2016-08-12
Fix for 100 router match iterations + source

seansan:
Fix for 100 router match iterations + source

https://github.com/convenient/magento-ce-ee-config-corruption-bug#the-fix


alexkirsch:
:thumbsdown:

PR #100, opened 2016-08-19
Added stock update event to registerProductsSale + revertProductsSale

seansan:
Added stock update event to registerProductsSale + revertProductsSale - much needed event to solve complex stock event capture methods. Suggestion is made here bij Fabian: http://magento.stackexchange.com/questions/25207/detect-inventory-change

Maybe the event prefix which is defined elsewhere should be use, but for now just added event name as CONST

Breaks nothing, adds cool stuff

Please verify + upvote if you like it

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Stock.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Stock.php
@@ -162,7 +162,6 @@ class Mage_CatalogInventory_Model_Stock extends Mage_Core_Model_Abstract
         Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_AFTER, array(
         'stock'          => $this,
         'items'          => $items,
-        'fullsave_items' => $result,
         'operator'       => '+'
         ));
         
@@ -185,7 +184,6 @@ class Mage_CatalogInventory_Model_Stock extends Mage_Core_Model_Abstract
         Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
         'stock'          => $this,
         'items'          => $items,
-        'fullsave_items' => $result,
         'operator'       => '-'
         ));
         return $this;
--- a/app/code/core/Mage/CatalogInventory/Model/Stock.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Stock.php
@@ -50,6 +50,9 @@ class Mage_CatalogInventory_Model_Stock extends Mage_Core_Model_Abstract
     const STOCK_IN_STOCK            = 1;
 
     const DEFAULT_STOCK_ID          = 1;
+    
+    const EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE = 'cataloginventory_stock_item_correct_qty_before';
+    const EVENT_CORRECT_STOCK_ITEMS_QTY_AFTER = 'cataloginventory_stock_item_correct_qty_after';
 
     protected function _construct()
     {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -159,8 +174,18 @@ public function registerProductsSale($items)
...
     public function revertProductsSale($items)
     {
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
+        'stock'     => $this,
+        'items_obj' => (object)array('items' => &$items),
+        'operator'  => '-'
+        ));
         $qtys = $this->_prepareProductQtys($items);
         $this->_getResource()->correctItemsQty($this, $qtys, '+');
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
```
colinmollenhour: I think this is a copy/paste error. Should be `..._QTY_AFTER`

app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -50,6 +50,9 @@ class Mage_CatalogInventory_Model_Stock extends Mage_Core_Model_Abstract
     const STOCK_IN_STOCK            = 1;
 
     const DEFAULT_STOCK_ID          = 1;
+    
+    const EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE = 'cataloginventory_stock_item_correct_qty_before';
```
colinmollenhour: Being a bit pedantic here but I think a better term than "correct" would be "adjust" since correct implies that it was previously incorrect which could be confusing.

seansan:
OK

app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -159,8 +174,18 @@ public function registerProductsSale($items)
      */
     public function revertProductsSale($items)
     {
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
+        'stock'     => $this,
+        'items_obj' => (object)array('items' => &$items),
+        'operator'  => '-'
```
drobinson: This also seems to be copy/paste error, why doesn't the operator match the one passed to correctItemsQty?

app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -159,8 +174,18 @@ public function registerProductsSale($items)
...
+        ));
         $qtys = $this->_prepareProductQtys($items);
         $this->_getResource()->correctItemsQty($this, $qtys, '+');
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
+        'stock'          => $this,
+        'items'          => $items,
+        'operator'       => '-'
```
drobinson: copy/paste error?

app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -132,6 +135,11 @@ protected function _prepareProductQtys($items)
      */
     public function registerProductsSale($items)
     {
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_BEFORE, array(
+        'stock'     => $this,
+        'items_obj' => (object)array('items' => &$items),
+        'operator'  => '+'
```
drobinson: copy/paste error?

app/code/core/Mage/CatalogInventory/Model/Stock.php
```diff
@@ -150,6 +158,13 @@ public function registerProductsSale($items)
         }
         $this->_getResource()->correctItemsQty($this, $qtys, '-');
         $this->_getResource()->commit();
+        
+        Mage::dispatchEvent(self::EVENT_CORRECT_STOCK_ITEMS_QTY_AFTER, array(
+        'stock'          => $this,
+        'items'          => $items,
+        'operator'       => '+'
```
drobinson: copy/paste error?

app/code/core/Mage/CatalogInventory/Model/Stock.php
colinmollenhour: Ahh, good point.

seansan: Agree. Feel free to adjust.

tmotyl:
@seansan  will you push an update?

PR #101, opened 2016-09-16
remove url from homepage in sitemap

shirtsofholl:
tja. works for 99% is important to have right for 100%. But this solution is not sound proof. Still I would advise to accept it. As de base-url is always the homepage ... and I am not sure we would want to create another config option with "use baseurl als main homepage yes/no"

### DIFF
--- a/app/code/core/Mage/Sitemap/Model/Sitemap.php
+++ b/app/code/core/Mage/Sitemap/Model/Sitemap.php
@@ -195,13 +195,16 @@ class Mage_Sitemap_Model_Sitemap extends Mage_Core_Model_Abstract
         /**
          * Generate cms pages sitemap
          */
+        $homepage = (string)Mage::getStoreConfig('web/default/cms_home_page', $storeId);
         $changefreq = (string)Mage::getStoreConfig('sitemap/page/changefreq', $storeId);
         $priority   = (string)Mage::getStoreConfig('sitemap/page/priority', $storeId);
         $collection = Mage::getResourceModel('sitemap/cms_page')->getCollection($storeId);
         foreach ($collection as $item) {
+            $url = $item->getUrl();
+			if ( $url == $homepage) { $url = ''; $priority = 1; }
             $xml = sprintf(
                 '<url><loc>%s</loc><lastmod>%s</lastmod><changefreq>%s</changefreq><priority>%.1f</priority></url>',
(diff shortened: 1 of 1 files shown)


drobinson:
I'm not sure I understand the use case here, you don't want the homepage to be in the sitemap?

shirtsofholl:
Homepage is bare domain

Not the whole uri of the Cms page of the homepage

Site map often contains

Domain.com/home

Instead of domain.com

app/code/core/Mage/Sitemap/Model/Sitemap.php
```diff
@@ -195,13 +195,16 @@ public function generateXml()
         /**
          * Generate cms pages sitemap
          */
+        $homepage = (string)Mage::getStoreConfig('web/default/cms_home_page', $storeId);
         $changefreq = (string)Mage::getStoreConfig('sitemap/page/changefreq', $storeId);
         $priority   = (string)Mage::getStoreConfig('sitemap/page/priority', $storeId);
         $collection = Mage::getResourceModel('sitemap/cms_page')->getCollection($storeId);
         foreach ($collection as $item) {
+            $url = $item->getUrl();
+			if ( $url == $homepage) { $url = ''; $priority = 1; }
```
colinmollenhour: Tabs vs spaces...

seansan: And if someone responds about the prio autosetting it to 1 (who know a domain "could" have an main page that is not the most important) ... I understand. But I do think this is best. 

It has always bugged me that not the real domain, but rather the CMS handle was sent to Google (or Bing for that matter) .... then we block it via robots or in nginx create a 301 for the CMS handle .... simplest solution is to: just send the main URL

seansan: Also: cool that Colin has joined this whole project. Thx

seansan:
Was edited on web. Please change and accept.

colinmollenhour:
I can't edit your personal branches.

Flyingmana:
i removed the priority part, as it had a bug to also affect all cms pages afterwards. As the other PR related to an event was merged, this should also be easier to customize now so I think we can left this out.

seansan:
yes see the bug/ thx.

PR #107, opened 2016-10-07
Support models "getCommentText" method for group note generation

sreichel:
In system.xml `<comment>` has a diffent behavior for `<fields>` and `<groups>`.
This only works for `<fields>`:

```
<comment>
    <model>namespace/module<model>
</comment>
```


rafaelpatro:
Will it be merged to master (1.9.3.x)?

PR #109, opened 2016-10-09
Fix possible option id/label collisions in getOptionId()

edannenberg:
See https://gist.github.com/edannenberg/15c0eddfddad40eca32793708f67c331 for details.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Abstract.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Abstract.php
@@ -94,10 +94,18 @@ abstract class Mage_Eav_Model_Entity_Attribute_Source_Abstract
 
     public function getOptionId($value)
     {
+        $bc_warning = false;
         foreach ($this->getAllOptions() as $option) {
-            if (strcasecmp($option['label'], $value)==0 || $option['value'] == $value) {
+            if (strcasecmp($option['label'], $value)==0) {
                 return $option['value'];
             }
+            if ($option['value'] == $value) {
+                $bc_warning = true;
+            }
+        }
+        if ($bc_warning) {
+            Mage::log('Mage_Eav_Model_Entity_Attribute_Source_Abstract::getOptionId() no longer accepts option_id as param',
+                      Zend_Log::WARN);
         }
         return null;
     }


colinmollenhour:
A fix which for all intents and purposes is completely valid yet does *potentially* break backwards compatibility.. In my opinion these kinds of bugs are worth fixing if there is a very low probability of actually causing any harm, but I suspect others will want to stay by the more strict rule of no BC-breaking changes..

The only time this should break BC is if some code were to pass the option_id to getOptionId and expect back the same.. That makes no sense whatsoever since you would be using this method to get the value that you already have. So it is an absolutely terrible API.

However, it would be possible to half-fix the issue without completely breaking BC.. Like so:

```php
    public function getOptionId($value)
    {
        foreach ($this->getAllOptions() as $option) {
            if (strcasecmp($option['label'], $value)==0) {
                return $option['value'];
            }
        }
        // Keep support for the senseless behavior:
        foreach ($this->getAllOptions() as $option) {
            if ($option['value'] == $value) {
                return $option['value'];
            }
        }
        return null;
    }
```

Personally, I would accept this PR as-is though...

alexkirsch:
I agree with you Colin;  I can not see an valid reason for returning the value that was passed in..   This looks like a bug to me.

I am okay with merging this PR as is.

tmotyl:
Could we somehow detect the edge case and log the about breaking change in the log?

edannenberg:
@tmotyl Not really, I guess we could log a notice before returning `null` but then again the fix only breaks pants-on-head silly code. Shouldn't be an issue in practice.

colinmollenhour:
@Flyingmana I assume you mean if a value was matched *and* no label was matched?

Flyingmana:
actually I dont know.
Because it would return the first find in original code. I first thought it may be good to log a notice when this wrong case would happen, but its probably unlikely that wrong usage AND a collision appear together, so your interpretation is I think the better one.

edannenberg:
More importantly it minimizes false positive log spamming. Updated the PR, let me know if you want different wording.

edannenberg:
We might as well use the fix proposed by @colinmollenhour then and save people the hassle. Full BC without possible collisions.

colinmollenhour:
@LeeSaferite  and @Flyingmana are you on board with my proposed solution (first comment)? That is, two loops, second one keeps BC, no config flag added.

Flyingmana:
yes

edannenberg:
Still waiting for approval from @LeeSaferite before I update the PR again. In case my last edit went unnoticed: Using @colinmollenhour's solution could still lead to collisions when passing option ids. IMO this is a case where breaking BC is justified, it's clearly a bug and not a feature.

LeeSaferite:
I was unaware this was waiting on input from me for almost a year. Yikes. FWIW, things like this just need 2 approvals to get merged.

edannenberg:
And I was unaware this PR still existed, accidentally deleted the branch while cleaning up my fork.

PR #117, opened 2016-10-14
Add custom source for multiselect product attributes

bastienlm:
When you create your own multiselect product attribute with a custom source model, this attribute is not displayed on the layered navigation.

Thanks to this fix, you can show your attribute on the filter. In first time I browse all `attribute_id` and if the attribute have a custom source model I set all custom values on `options` array.

(Inspirated by : http://stackoverflow.com/questions/7597448/multi-select-filter-in-layered-navigation)

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Source.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Source.php
@@ -182,6 +182,21 @@ class Mage_Catalog_Model_Resource_Product_Indexer_Eav_Source extends Mage_Catalo
             $options[$row['attribute_id']][$row['option_id']] = true;
         }
 
+        foreach ($attrIds as $attId) {
+            $attribute = Mage::getResourceSingleton('catalog/product')->getAttribute($attId);
+            if ($attribute->getSourceModel() != 'eav/entity_attribute_source_table') {
+                unset($options[$attId]);
+                $sourceOptions = $attribute->getSource()->getAllOptions();
+                if ($sourceOptions) {
+                    foreach ($sourceOptions as $sourceOption) {
+                        if (isset($sourceOption['value'])) {
+                            $options[$attId][$sourceOption['value']] = true;
+                        }
+                    }
+                }
+            }
+        }
+
         // prepare get multiselect values query
         $productValueExpression = $adapter->getCheckSql('pvs.value_id > 0', 'pvs.value', 'pvd.value');
         $select = $adapter->select()


LeeSaferite:
@drobinson Can you run this code and check it is working?

@drobinson  On a side note, we need to get a working testing framework in place and mandate tests for PRs.

colinmollenhour (dismissed):
Nice work, although code style is not exactly the same as the core code (space after control keywords).

sreichel:
On hold ... broke indexer (or ive done something wrong)

bastienlm:
@sreichel Strange, I have this fix on a big project since many month and I have no problem. When you say 'broke indexer', you can't reindex or it's broke after reindex ? product attributes indexe ?

sreichel:
With this fix I wasnt able to reindex product attributes.  As said ... maybe Ive done something wong (just copy&paste to our dev server where I have some multiselect attributes with custom source model)

Just put on hold to verify :)

bastienlm:
I've update my PR for respect space, thanks @colinmollenhour for the report.

colinmollenhour:
@sreichel Did you ever figure out if there is indeed a bug with this PR or not?

sreichel:
@colinmollenhour I will check it on sunday.

sreichel:
With this code i cannot reindex product attribtues.

@bastienlm which magento version do you use for your project? This PR was for 1.9.3.0, (in 2016) maybe something else has changed.

@colinmollenhour @Flyingmana can you please test it, too?

bastienlm:
This code work on Magento ver. 1.9.3.7 (no LTS)

I try to test this on the last LTS version this week.

ADDISON74:
This PR needs to be rebased to OpenMage:1.9.4.x.

fballiano (approved):
it seems to work

ADDISON74:
What about @sreichel remarks?

tobihille (approved):

colinmollenhour (approved):

fballiano:
I've applied this patch on a customer's project (on my local copy) and I can reindex without problems:
```
fab@MBP-Fab ~/P/x (prod)> php shell/indexer.php reindexall
Category Flat Data index was rebuilt successfully in 00:00:00
Product Flat Data index was rebuilt successfully in 00:00:00
Stock Status index was rebuilt successfully in 00:00:00
Catalog product price index was rebuilt successfully in 00:00:00
Category URL Rewrites index was rebuilt successfully in 00:00:00
Product URL Rewrites index was rebuilt successfully in 00:00:00
URL Redirects index was rebuilt successfully in 00:00:00
Catalog Category/Product Index index was rebuilt successfully in 00:00:00
Search Index index was rebuilt successfully in 00:00:02
Tag Aggregation Data index was rebuilt successfully in 00:00:00
Product Attributes index was rebuilt successfully in 00:00:01
fab@MBP-Fab ~/P/x (prod)> 
```

so I don't know...

sreichel:
Has someone example code to test with? (I have just plain OM with sample data)

fballiano:
I tested on an empty store and one of my customer's data, but only the reindex process.

fballiano (approved):
I've rebased and retested the indexing without problems.

I've to say that I didn't test the custom source but it doesn't seem to create problems to existing code.

fballiano:
mmmm but if the "keys" of the custom source were "int" then it would work? because it would probably be better than nothing? maybe documenting the limit?

kiatng:
The core really does not expect custom `source_model` for attributes that use layered navigation. This PR is half-baked solution for it. Do we really want to cater for these edge cases? In any case, the following code prevents the integrity constraint violation, but it won't help with layered nav:

```php
        $select  = $adapter->select()
            ->from($this->getTable('eav/attribute'), ['attribute_id'])
            ->where('source_model != ?', 'eav/entity_attribute_source_table')
            ->where('attribute_id IN (?)', $attrIds);
        $query = $select->query();
        while ($row = $query->fetch()) {
            $attribute = Mage::getResourceSingleton('catalog/product')->getAttribute($row['attribute_id']);
            if ($source = $attribute->getSource()) {
                if ($sourceOptions = $source->getAllOptions()) {
                    unset($options[$row['attribute_id']]);
                    foreach ($sourceOptions as $sourceOption) {
                        // Values can only be integer
                        if (is_int($sourceOption['value']) || is_string($sourceOption['value']) && ctype_digit($sourceOption['value'])) {
                            $options[$row['attribute_id']][$sourceOption['value']] = true;
                        }
                    }
                }
            }
        }
```

Because there is no good solution, I think this PR should be closed.

PR #118, opened 2016-10-14
Fix shipping street and vat label

bastienlm:
Fix bad label attribute 'for' on shipping.phtml for the street (only rwd) and vat id.


FredericMartinez:
Since 2008

PR #119, opened 2016-10-14
Fix cms breadcrumb issue in 1.9.3.0

drobinson:
$breadcrumbs is not always defined and should have at least _some_ checking.  Results in fatal exception on frontend in some cases.


Flyingmana:
also picked into 1.9.3.1

PR #122, opened 2016-10-18
Fix SOAP error

drobinson:
See http://magento.stackexchange.com/questions/140761/soap-connection-problems-after-1-9-3-0-update

### DIFF
--- a/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
+++ b/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
@@ -556,11 +556,21 @@ abstract class Mage_Api_Model_Server_Handler_Abstract
      * @param array $result
      * @return mixed
      */
-    public function processingMethodResult(array $result)
+    public function processingMethodResult($result)
     {
-        foreach ($result as &$row) {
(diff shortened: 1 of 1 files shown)


infabo:
could be an array?

infabo:
could be an array?

colinmollenhour:
Yes, so:

~~~
if (is_string($row)) {
 ...
} else if (is_array($row)) {
 ...
}
~~~

infabo:
Yes, for the first if. for the second one just

```php
if(is_string($result)) {...}
```

cieslix:
the **processingRow** is using internally **preg_replace_callback** it accepts just array or string if $row is an array it calls **processingMethodResult** (recursion), so the following method should be fine.

```
    public function processingMethodResult($result)
    {
        if (is_array($result)) {
            foreach ($result as &$row) {
                $row = $this->processingMethodResult($row);
            }
        } elseif (is_string($result)) {
            $result = $this->processingRow($result);
        }
        return $result;
    }
```

@drobinson can you check if this work as intended ?

colinmollenhour:
Was fixed in 1.9.3.1

PR #123, opened 2016-10-24
Fixes Magento bug 648

seansan:
Read here
http://magento.stackexchange.com/questions/60743/undefined-variable-freepackagevalue-in-tablerate-php-on-line-130/60779#6077
and here
https://www.magentocommerce.com/bug-tracking/issue/index/id/648


infabo:
+1

PR #137, opened 2016-11-22
Update and rename custom.css to custom.css.template

seansan:
Useless during upgrades


seansan:
rename + add .template

Flyingmana:
is this file not referenced somewhere? I would guess its needed for a proper default fallback

sreichel:
It is referenced ... https://github.com/OpenMage/magento-lts/blob/1.9.3.1/app/design/adminhtml/default/default/layout/main.xml#L84

Nice idea, but vote 4 close ;)

colinmollenhour:
Agreed it should be left alone. There are other ways to override it if needed.

PR #150, opened 2017-01-04
Fix $this->setMassactionIdField('id_here') in custom/all grids

seansan:
I could not get setMassactionIdField to work in a custom grid and stumbled on I think what is an error in th code. setMassactionIdField  is never being respected is seems .... (please comfirm)

What I did
- Add use_index to massaction checkbox data array
- Remove check for getNoObjectId
- Fix reversal of merge (better other way around)

Needs logic confirmation
Also there are only a handfull of grids that use setMassactionIdField if they actually start working that may break something again

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -611,9 +611,10 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
     protected function _prepareMassactionColumn()
     {
         $columnId = 'massaction';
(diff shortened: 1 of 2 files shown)


sreichel:
@seansan can you provide some sample code to reproduce?  (I agree `setMassactionIdField` is broken/has no effect ... but i've no code to play around ...)

seansan:
Difficult. I have a custom extension but it is based of another extension. In short: we cant execute any massaction because the ID's are unknown when sent back to the server. Reverse test: I just upgraded to M 1.9.3.x where this pull is not yet in: and we see the error again. No way to update status. 

Maybe create some kind of simple custom grid?

Maybe we can code review ... dont know any other quick test

sreichel:
Duplicate of #247

PR #159, opened 2017-01-18
Added skipSaleableCheck to attribute products list

nrunde:
List all available attributes products in Product.Config json by using magentos skipSaleableCheck functionallity


tmotyl:
Hi
Thanks for your contribution. Can you explain a little bit more what this change is about? What problem does it solves? and how to test it?
Good description will help us review and merge the change.
Thanks

nrunde:
In my case I wanted to show all simple products as options of a configurable product except of their stock status. Without the skipSaleableCheck all products which are not-in-stock weren't added to the product.config json and I couldn't show them as I want to.

seansan:
Maybe similar we have seen this requirement also

Where out of stock is either
- not shown (as current)
- shown but greeted out using distinct class
- string - out of stock is added via translations

Don't know how easy or hard. Just liked to share

sreichel:
@nrunde can you please explain how this fix works. For me `$skipSaleableCheck` is always `false` and has no effect.

Btw ... show out-of-stock items as config options seems not to be a good idea. You see it as available option, but your not able to add it to cart.

sreichel:
Closed until feedback.

PR #160, opened 2017-01-19
Added missing functionality for getThumbnailUrl

sreichel:
Return thumbnail URL instead of empty string.


cieslix:
+1

tested, it works like Mage_Catalog_Helper_Product::getSmallImageUrl or Mage_Catalog_Helper_Product::getImageUrl

sreichel:
Same here ... please check and include into 1.9.3.3 ... it's just the same code that is used for other image URLs.

akira28:
@colinmollenhour shouldn't it be merged in 1.9.3.x branch?

colinmollenhour:
Ahh yes, good catch. Merged.

PR #165, opened 2017-02-08
Simple improve of Not valid template file debugging

seansan:
Simple improve of Not valid template file debugging

### DIFF
--- a/app/code/core/Mage/Core/Block/Template.php
+++ b/app/code/core/Mage/Core/Block/Template.php
@@ -241,7 +241,7 @@ HTML;
                 include $includeFilePath;
             } else {
                 $thisClass = get_class($this);
-                Mage::log('Not valid template file:'.$fileName." class:{$thisClass}", Zend_Log::CRIT, null, true);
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Core/Block/Template.php
```diff
@@ -240,7 +240,8 @@ public function fetchView($fileName)
             if ($includeFilePath != '' && (strpos($includeFilePath, realpath($this->_viewDir)) === 0 || $this->_getAllowSymlinks())) {
                 include $includeFilePath;
             } else {
-                Mage::log('Not valid template file:'.$fileName, Zend_Log::CRIT, null, true);
+                $thisClass = get_class($this);
+                Mage::log('Not valid template file:'.$fileName." class:{$thisClass}", Zend_Log::CRIT, null, true);
```
tmotyl: just a nitpick, I would prefer not to mix style of concatenating strings - single qoutes/double quotes. and sugges sth like:

 Mage::log('Not valid template file:' . $fileName . ' class: ' . $thisClass, Zend_Log::CRIT, null, true);

seansan:
You can adjust in transit and accept.

PR #169, opened 2017-02-09
Fixed the CSS merging data uri bug

royduin:
See: https://github.com/just-better/magento1-css-merge-data-uri-fix

As requested here: https://github.com/just-better/magento1-css-merge-data-uri-fix/issues/1

Currently only for 1.9.3.1, but this patch should be applied to all versions @seansan.


colinmollenhour:
Good fix. Thanks @royduin !

PR #175, opened 2017-02-15
fix: admin group_prices: disable is_percent input

ollb:
When modifying product group prices in admin, is_percent inputs were not disabled as the other inputs leading to a wrong validation error "Duplicate website group price customer group".

It may be related to : https://community.magento.com/t5/Technical-Issues/Can-t-save-product-Duplicate-website-group-price-customer-group/td-p/51039

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/edit/price/group.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/edit/price/group.phtml
@@ -124,7 +124,7 @@ var groupPriceControl = {
         <?php endif;?>
 
         if (data.readOnly == '1') {
-            ['website', 'cust_group', 'price', 'delete'].each(function(element_suffix) {
+            ['website', 'cust_group', 'price', 'delete', 'is_percent'].each(function(element_suffix) {
                 $('group_price_row_' + data.index + '_' + element_suffix).disabled = true;
             });
             $('group_price_row_' + data.index + '_delete_button').hide();


tmotyl:
Hi @ollb 
Thanks for your contribution. Can you describe the steps needed to reproduce the issue? And how the system should work instead?
Good description will help us review and merge the change.
Thanks

cieslix:
+1
I was able to test it and it works, it requires change price attribute to be website related and you have to use Magento multistore/multiwebsite. This patch bug related to "Duplicate website group price customer group."

wigman:
Please merge, this fix helped me.

sreichel:
+1

edannenberg:
FYI: Looks like an accident while merging the PR but the original contributor attribution is now lost. Almost dismissed this too because fd6351d looks like a useless merge commit at first glance.

edannenberg:
NVM, false alarm. Commit was back from february, one of the reasons I dislike merging this way. Sorry!

sreichel:
> one of the reasons I dislike merging this way.

What i your prefered way? Rebase first?

edannenberg:
Yep, a fast-forward merge is the ideal merge. Was cherry-picking the latest commits for my fork and was temporary confused because I'm bad at github. :P You can find the original commit easy enough if you look at the "useless" merge commit's parents but makes the cherry picking a bit annoying. It's also not always obvious when a specific commit actually entered a branch (date-wise) if you are digging through a history and the merge commit and actual commit have date differences such as in this case.

PR #177, opened 2017-02-15
Better HTML tagging

blopa:
I find this way better than the regular HTML tagging from the original source. It removes the uneeded use of the li tag twice. It's just a organized version of the same code.


sreichel:
IMHO i would not touch template files for better styling. (Who uses base template? It's easy to override. ...)

LeeSaferite:
I agree, this is not a change I feel we should pull in (even if I agree with the desire to fix the structure)

PR #183, opened 2017-02-25
fix unserialized data clitch

blopa:
refer: http://magento.stackexchange.com/questions/146978/default-configuration-value-for-serialized-array-settings-is-ignored/146981#146981


cieslix:
+1

PR #184, opened 2017-02-25
Add option reindexallrequired

seansan:
Add option reindexallrequired
as described here: http://magento.stackexchange.com/questions/3028/permanent-solution-for-the-common-indexing-issue

### DIFF
--- a/shell/indexer.php
+++ b/shell/indexer.php
@@ -152,7 +152,7 @@ class Mage_Shell_Compiler extends Mage_Shell_Abstract
                     echo $e . "\n";
                 }
             }
-        } else if ($this->getArg('reindex') || $this->getArg('reindexall')) {
+        } else if ($this->getArg('reindex') || $this->getArg('reindexall') || $this->getArg('reindexallrequired')) {
             if ($this->getArg('reindex')) {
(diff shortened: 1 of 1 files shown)


bob2021:
I think the condition on line 166 should also check that 

`count($process->getUnprocessedEventsCollection()) === 0`

Without this, indexers would be skipped even if they require a partial update

There are also still cases where an indexer will run even if it is not required - for example if the price indexer requires a re-index then it will also run the stock indexer as it is a dependency.  A simple/ugly workaround might be to add this loop at line 164 (untested):

```php
if ($this->getArg('reindexallrequired')) {
    foreach ($processes as $process) {
        if ($process->getStatus() == Mage_Index_Model_Process::STATUS_PENDING && count($process->getUnprocessedEventsCollection()) === 0) {
            $process->setData('runed_reindexall', true);
        }
    }
}
```

seansan:
I think the main issue is to allow for a "re-run all" and if some exotic cases are missed this is OK. 
- So missing  a required re-run is bad and should be caught in this fix
- But if an exotic case exists where a re-run occurs - I personally would be OK with this ... 
Hope it helps

seansan:
testing the extra condition

Issue #152, opened 2017-01-05
Improve post-checkout reindexing


bob2021:
Many of the indexers call `Mage_Index_Model_Resource_Abstract::disableTableKeys()`, which runs a statement like this: `ALTER TABLE ... DISABLE KEYS`.  Then, after some rows are inserted into the index table, `enableTableKeys()` is called.  

As far I can see `DISABLE/ENABLE KEYS` only works with MyISAM tables, so it isn't used by modern Magento installations.  What's worse is in MySQL 5.5+ the `ALTER TABLE` statement requires a table metadata lock.

As the statement doesn't actually do anything it could be removed and the lock would be avoided (though its possible it might cause problems for older installations that might still have MyISAM tables)

It may seem like it wouldn't have that much impact as the lock should only be for a very short time, but in production I've seen large queues of queries 'Waiting for table metadata lock' on tables like cataloginventory_stock_status and catalog_product_index_price when the ALTER TABLE statement is blocked by some other very slow transaction.

colinmollenhour:
Interesting.. Also worth noting is that `ALTER TABLE` causes in-progress transactions to be implicitly committed. I didn't look to see if these happen in the middle of a transaction or not but I'm definitely in favor of completely removing support for MyISAM wherever possible (see #105 for all reasons).

LeeSaferite:
While we are discussing this I guess I should also point out that there is
at least one instance of an index temp table being a real table and causing
race conditions. I'll have to find it later and point it out.

bob2021:
Following up on this I've been looking into the indexing that happens during checkout (specifically price & stock).  I was seeing a couple of database errors when many orders were placed simultaneously:

- A deadlock on the event_* tables
- A duplicate insert error on the price indexer *_tmp tables

There are a few requirements to see this behaviour:
- The order must bring a product's inventory level below the 'Qty for Item's Status to Become Out of Stock' or 'Notify for Quantity Below'
- Ordered products must have backorders set to 'No backorders' or 'Allow Qty Below 0 and Notify Customer'
- The config option 'Display out of stock products' must be set to 'No'

While looking for the solution I found a few issues:



[Some database transactions with missing/unsafe try catch blocks](https://github.com/bob2021/magento-lts/commit/7f53ae4b92ac6ab4a45ac0d5339351009ad2968f) (though most are probably harmless)



[...]

PR #187, opened 2017-02-27
Safer DB transactions

bob2021:

I've reviewed magento's database transactions and found some that could potentially affect a parent transaction or not close properly (though most are probably harmless)


tmotyl:
shouldn't we make the pull against 1.9.3.x branch?

bob2021:
Rebased on 1.9.3.x

PR #188, opened 2017-02-27
Disable `DISABLE KEYS` statements

bob2021:

Indexers that extend `Mage_Index_Model_Resource_Abstract` run an `ALTER TABLE ... DISABLE|ENABLE KEYS` statement before and after inserting data.  

The statement is only supported by MyISAM tables so it is useless for modern magento installations, but it still requires a table metadata lock even if MyISAM is not used.

The catalogsearch index does still use MyISAM, but it doesn't extend `Mage_Index_Model_Resource_Abstract` so it isn't affected.

This PR disables the statements by default, but leaves the functionality in place if still required.  

It also removes calls to `Mage_Index_Model_Resource_Abstract::useDisableKeys`.  In most cases it was called by mistake and had no effect (it should have been `Mage_Index_Model_Resource_Abstract::disableTableKeys`), but now they aren't required at all.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Abstract.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Abstract.php
@@ -99,9 +99,7 @@ abstract class Mage_Catalog_Model_Resource_Product_Indexer_Eav_Abstract
             $adapter->delete($this->getMainTable(), $where);
 
             // insert new index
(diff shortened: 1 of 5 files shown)


colinmollenhour:
Would you mind rebasing or resubmitting on 1.9.3.x branch? Otherwise I approve.

bob2021:
Rebased on 1.9.3.x

tmotyl:
@colinmollenhour  would you mind merging it once you approve?

PR #191, opened 2017-02-27
Don't save mass_action event to db when saving stock item model

bob2021:

When a stock item model is saved independently of a product model (eg. during checkout) and 'Display Out of Stock Products' = 'No', the following happens:

- A mass_action index event is registered
- The stock indexer runs
- The mass_action event from earlier is processed; these indexers run:
  - `catalog_product_attribute`
  - `catalog_product_price`
  - `catalogsearch_fulltext`

The problems are:
- Saving a mass_action event just to load it later on in the same request is inefficient
- The mass_action event isn't actually required unless the stock status has changed
- One mass_action record is shared between all requests:
  - If indexers are set to manual mode the event data can grow very large (megabytes)
  - When the event is loaded it could contain data from other requests; events could be processed more than once across different requests

This PR has been updated since I initially opened it, the original approach had the advantage that it indexed everything in one pass, but it was more complex and less backwards compatible.  [The original is here.](https://github.com/OpenMage/magento-lts/compare/1.9.2.4...bob2021:avoid-stock-item-mass_action-old?expand=1)

This PR keeps the mass_action event in memory instead of saving it to the database, and it only processes the mass_action if the stock status has changed.


LeeSaferite:
@bob2021 Why did you close the PR?

bob2021:
Sorry, I should have added a comment - I've made some mistakes in my understanding of how mass_action events are processed.  I think this PR is probably ok but I wanted to double check it before anyone else spent time looking at it - I'll reopen if it looks ok.

seansan:
+1 for the idea (not tested) any update on this one?

fballiano:

bob2021:

ADDISON74:

fballiano:
I've solved the conflicts (hopefully correctly since this code is extremely old) and rebased to main. I think this PR is worth checking and testing, please some admins take a look at this.

app/code/core/Mage/CatalogInventory/Model/Observer.php
```diff
@@ -1041,10 +1041,21 @@ public function refundOrderItem($observer)
      */
```
fballiano, line 1041:
```suggestion
     * @return void
     */
```

just to trigger the workflow, they do not run on a force push (for rebase)

app/code/core/Mage/CatalogInventory/Helper/Data.php
```diff
@@ -41,6 +41,11 @@ class Mage_CatalogInventory_Helper_Data extends Mage_Core_Helper_Abstract
      */
     protected static $_isQtyTypeIds;
 
+    /**
+     * Product ids to be re-indexed due to stock status change
+     */
+    protected $_productIdsForMassReindex = array();
```
fballiano, line 47:
```suggestion
     * Product ids to be re-indexed due to stock status change
     * @var array
     */
    protected $_productIdsForMassReindex = [];
```

app/code/core/Mage/CatalogInventory/Model/Observer.php
```diff
@@ -1038,13 +1038,27 @@ public function refundOrderItem($observer)
...
     {
-        Mage::getSingleton('index/indexer')->indexEvents(
+        $helper = Mage::helper('cataloginventory');
+        $ids = $helper->getProductIdsForMassReindex();
+        if (!$ids) {
+            return;
+        }
+
+        $massObject = new Varien_Object();
+        $massObject->setAttributesData(array('force_reindex_required' => 1));
```
fballiano, line 1052:
```suggestion
        $massObject->setAttributesData(['force_reindex_required' => 1]);
```

fballiano:
So, I took a better look at this and:
- it could be needed to reindex a product also if the is_outofstock didn't change, because there could be low stock triggers or something like that
- removing the "save to db" is interesting but everything has to be processed in the same run, or it won't be resumable
- the way it's coded now kinda resemble a queue like approach, it's not the best since OM doesn't really have queues but I (today) feel it's better to have it this way in case we can snap-in a queue in the future

So, for the time being, I prefer to close it

PR #193, opened 2017-02-28
admin panel customer address form fixes

bob2021:
Switching between saved customer addresses when creating an order in the admin panel can sometimes result in a '0' being set for the region field.

Also, when editing customer addresses in the admin panel, the same form instance is reused for each address, but it isn't reset properly for each iteration, so data can leak between addresses.

eg. address1 has a postcode set, but address2 has a null value for postcode.  Because `addValues()` is used instead of `setValues()`, the postcode field wont get cleared for address2 (it will still have address1's postcode)


cieslix:
+1 
Tested, works as intended. Before I was able to see '0' as my region when I was changing between addresses in Admin Panel Order Create form.

PR #194, opened 2017-02-28
fix paypal IPN postback response parsing

bob2021:
I'm not sure where this fix first came from, I've seen it posted on a few different blogs and forums.

The IPN class always expects a status code to be on the second line of the response, but curl sometimes adds extra HTTP headers, so it is better to look for the status code on the last line of the response instead.


tbaden (dismissed):
LGTM

seansan:
Whats the status of this? Has it been tested ... can it be merged?

dbachmann:
Hey there,

a customer of mine got this mail from paypal too and I've checked this issue. After changed the code and tested again everything works fine now.

Flyingmana:
that means we have 2 approves and can merge it now.

PR #195, opened 2017-02-28
fix paypal settlement report parsing

bob2021:
These are the changes I had to make to get paypal settlement reports to parse correctly

I'm not sure if its a regional issue (I'm in the UK), but these fixes shouldn't change any previous behaviour.  

The only exception might be if the date format used in reports varies between regions, but that seems unlikely as the format used is not regional (`yyyy/MM/dd HH:mm:ss Z`)

### DIFF
--- a/app/code/core/Mage/Paypal/Model/Report/Settlement.php
+++ b/app/code/core/Mage/Paypal/Model/Report/Settlement.php
@@ -269,6 +269,14 @@ class Mage_Paypal_Model_Report_Settlement extends Mage_Core_Model_Abstract
 
         $flippedSectionColumns = array_flip($sectionColumns);
         $fp = fopen($localCsv, 'r');
+
+        // skip BOM if present
+        $bom = pack("CCC", 0xef, 0xbb, 0xbf);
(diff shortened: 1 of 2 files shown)


colinmollenhour (dismissed):

fballiano:
Do we know is this PR is still needed?

ADDISON74:
From what I see it is almost 5 years ago. It was not even integrated but left open. I see that no one complained about any problem in the OM code so this PR could come and solve it. And that there are many others left without feedback without detailed checks. It is a pity that the contribution is not capitalized in its time.

fballiano:
I've no way of checkin if the problem exists, actually I didn't fully understand where (in the backend) it happens. Also, probably it was happening on an old magento version so maybe it's not relevant anymore, who knows.

fballiano:
from what I can see app/code/core/Mage/Paypal/Model/Report/Settlement.php it's still the same used in this PR

fballiano:

bob2021:

ADDISON74:

PR #197, opened 2017-03-01
Added progressive to image resize, see #95

sreichel:
Just added config option to @seansan PR.


sreichel:
Bump

sreichel:
Should we also add new translatable strings to CSV?

colinmollenhour:
Ahh, yes.

PR #198, opened 2017-03-02
Account Edit: Magento requests current password, even when you don't change it

sreichel:
Just copied from https://magento.com/tech-resources/bug-tracking/issue/index/id/1525/

> Steps to reproduce:
With the update to version 1.9.3.0 or 1.9.3.1 Magento displays the error message "Invalid current password" and don't save any changes when you try to edit your account data without setting a new password.

> the problem is introduced in 1.9.3.0 and is caused by the fact that the password is always being validated even though the users has not checked the "Change password" combobox.

### DIFF
--- a/app/code/core/Mage/Customer/controllers/AccountController.php
+++ b/app/code/core/Mage/Customer/controllers/AccountController.php
@@ -964,10 +964,6 @@ class Mage_Customer_AccountController extends Mage_Core_Controller_Front_Action
                 $customerForm->compactData($customerData);
                 $errors = array();
 
-                if (!$customer->validatePassword($this->getRequest()->getPost('current_password'))) {
-                    $errors[] = $this->__('Invalid current password');
-                }
-
                 // If email change was requested then set flag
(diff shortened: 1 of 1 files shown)


cieslix:
-1
IMHO this validation is fine like it is. If developer's intention was to check current password just for changing password this input would be hidden until marking checkbox "change password".

sreichel:
Agreed, but you got me wrong. ATM you'llget an "invalid password" message even when trying to edit some "non-wassword" account information. E.g you can't change yor name cause of hidden pw request.

cieslix:
@sreichel IMHO to edit non-passwords account data from version 1.9.3.0 you have to provide current password too.

**Magento CE 1.9.3.0 Release Notes, section Password enhancements**
- When a user changes their e-mail address, they are required to provide their password and to acknowledge the change from the previous address.

sreichel:
Close ... custom template issue where PW input is hidden as long you dont want to change PW.

Issue #196, opened 2017-02-28
patch SUPEE-7136 - secure cookie


bob2021:
I was going to open a PR with @colinmollenhour's patch to properly set & renew the secure session cookie added in 1.9.1.0

I thought it was strange that he hadn't submitted it already, then I read that there is an official patch SUPEE-7136 which does the same thing, but it isn't public.

It seems odd that neither one has been merged either here or officially, is there a reason?

Colin's patch is here:
https://gist.github.com/colinmollenhour/5066a3220881a9c0c2dd42fa1593cbff/revisions

Official patch has been posted here:
https://github.com/nexcess/magento-turpentine/issues/725#issuecomment-259718106

They are pretty much the same except:
- Cookie lifetime
  - Official: cookie lifetime from config
  - CM: 1 year
- HTTP-only flag
  - Offical: no
  - CM: yes

Assuming a PR for this would be accepted - which version is preferred?

colinmollenhour:
I'm not yet using magento-lts in production so have not yet gone though the process of porting in every patch I've made.. Hope to do that in the coming months though.

Cookie lifetime: I suppose "null" (use config) is the more correct option here.
HTTP-only: I see no reason that this cookie should be accessed by Javascript so for security reasons HTTP-only is better.

Feel free to submit a PR. Thanks!

PR #236, opened 2017-03-06
renew secure cookie lifetime (SUPEE 7136)

bob2021:
This PR fixes an issue where a session is reset when a secure page is accessed after the cookie lifetime expires.

It is the official SUPEE-7136 patch, but has been modified to set the http-only flag as suggested by @colinmollenhour in issue #196.


seansan:
@colinmollenhour 

Close? I already see this code in latest code

sreichel:
@seansan its not completly merged ... parameters are different then in SUPEE-7136.

colinmollenhour:
Merged latest so now diff is only 2 lines setting `true` for `$httponly` flag for renew/set.

PR #238, opened 2017-03-08
Added "System"-Tab to product attribute edit page

sreichel:
This PR adds a new tab to product attribute edit page where you can change :

- attribute model
- backend model
- backend table (used for??)
- frontend model
- input renderer
- note
- visibility
- source model

System tab was already in core code, it was just not visible. I've just added ...

- some fields
- ACL option
- field validation

All tests were fine so far, but please check field validation.


sreichel:
To much changes for rare use cases .... ? Adding this tab can done via event/observer if needed ...

Issue #24, opened 2015-06-20
composer.json type value


Flyingmana:
is there a specific reason why the `type` is set to `magento-source`? Other projects choose `magento-core` which is supported by at least 2 composer installers

drobinson:
Not that I'm aware of - other than "it works".  Do you know what compability we gain/lose by going with another type?

infabo:
I like this core installer: https://github.com/AydinHassan/magento-core-composer-installer
It uses `magento-core`

tmotyl:
you can use both core installers
https://github.com/AydinHassan/magento-core-composer-installer
and 
aoepeople/composer-installers

See instructions here:
http://openmage.github.io/magento-lts/install.html

tmotyl:
@LeeSaferite can you shed some light why magento-lts is using different type then other forks?

So if there is no good reason to keep different type I would opt for changing it. If there is, I would close the pull request and issues asking for the change as wont fix.

LeeSaferite:
Well, historically, 'magento-source' came before 'magento-core'.

https://github.com/AOEpeople/composer-installers/commit/163aa5194e9b8b0bc864637260920304034a05ca

https://github.com/magento-hackathon/magento-composer-installer/commit/e61b4e4088fd4751d6f8f6eb1bdadf9b296a764d

drobinson:
I think the question should be *is there a good reason to change it* since it will break backwards compatibility.  I would prefer to err on the side of not forcing anyone to update their config.

tmotyl:
ok, lets leave it as it is

PR #239, opened 2017-03-09
composer.json type value

mc388:
Change the composer.json type value from `magento-source` to `magento-core` to enable composer installers like https://github.com/AydinHassan/magento-core-composer-installer 

See issue #24

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "openmage/magento-lts",
   "license": "OSL-3.0",
-  "type": "magento-source",
+  "type": "magento-core",
   "require": {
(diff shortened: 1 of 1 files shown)


infabo:
See https://github.com/AydinHassan/magento-core-composer-installer/releases/tag/1.4.0

> Added the option magento-core-package-type to make the core package type configurable. By default it is set to magento-core. Useful if your core package has a type of magento-source or similar

You can make use of this new option.

tmotyl:
here you can find instruction how to install Magento LTS using both installers:
http://openmage.github.io/magento-lts/install.html

To keep backward compatibility we're not going to change package type.

PR #241, opened 2017-03-09
Fixed abandoned cart report

sreichel:
Fix for 1.9.3.1/1.9.3.x

```
`a:5:{i:0;s:64:"You cannot define a correlation name 'cust_mname' more than once";i:1;s:5216:"#0 /var/www/html/magento-1.9.1.0/lib/Varien/Db/Select.php(281): Zend_Db_Select->_join('inner join', Array, 'cust_mname.enti...', Array, NULL)
#1 /var/www/html/magento-1.9.1.0/app/code/core/Zend/Db/Select.php(352): Varien_Db_Select->_join('inner join', Array, 'cust_mname.enti...', Array, NULL)
#2 /var/www/html/magento-1.9.1.0/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php(224): Zend_Db_Select->joinInner(Array, 'cust_mname.enti...', Array)
#3 /var/www/html/magento-1.9.1.0/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php(90): Mage_Reports_Model_Resource_Quote_Collection->addCustomerData(NULL)
```

### DIFF
--- a/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php
+++ b/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php
@@ -206,7 +206,7 @@ class Mage_Reports_Model_Resource_Quote_Collection extends Mage_Sales_Model_Reso
                 )),
                 array('firstname' => 'cust_fname.value')
             )
-            ->joinLeft(
+            ->joinInner(
                 array('cust_mname' => $attrMiddlenameTableName),
                 implode(' AND ', array(
                     'cust_mname.entity_id = main_table.customer_id',
--- a/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php
+++ b/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php
@@ -215,14 +215,6 @@ class Mage_Reports_Model_Resource_Quote_Collection extends Mage_Sales_Model_Reso
                 array('middlename' => 'cust_mname.value')
             )
             ->joinInner(
-                array('cust_mname' => $attrMiddlenameTableName),
-                implode(' AND ', array(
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Nit-pick: You left the joinLeft and removed the joinInner but the core code has joinInner. Probably doesn't really make a difference but is generally better to stick with upstream unless there is a reason to change.

https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php#L209

sreichel:
@colinmollenhour can you pls merge to 1.9.3.1, too?

colinmollenhour:
As far as I understand it the 4-number branches are essentially deprecated and only the n.n.n.x branches will be maintained going forward (and mainly just the latest one).

Flyingmana:
But if someone needs it, you can still do a PR referencing the main PR. But please use a "Backport" label then.

sreichel:
@colinmollenhour i agree for some features that were added, but bugfixes could be backported (to at least last two versions) as in readme ...

> Each Magento Version release will get its own branch that will be independently maintained with patches and backported bug fixes.

sreichel:
@colinmollenhour 

> Nit-pick: You left the joinLeft and removed the joinInner but the core code has joinInner. Probably doesn't really make a difference but is generally better to stick with upstream unless there is a reason to change.
> https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Reports/Model/Resource/Quote/Collection.php#L209

Can you please check #41?

Think  `Fixed joinLeft / joinInner` should be reverted.

colinmollenhour:
Ahh, agreed.. Looks like comparing to core was a mistake.

PR #243, opened 2017-03-13
Made gender translatable, fixes #237

### LINKED ISSUE #237: Gender is not translatable
... cause values come directly from `eav_attribute_option_value` table.

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/design/frontend/base/default/template/customer/widget/gender.phtml#L30-L34

```
<?php $options = Mage::getResourceSingleton('customer/customer')->getAttribute('gender')->getSource()->getAllOptions();?>
<?php $value = $this->getGender();?>
<?php foreach ($options as $option):?>
    <option value="<?php echo $option['value'] ?>"<?php if ($option['value'] == $value) echo ' selected="selected"' ?>><?php echo $option['label'] ?></option>
<?php endforeach;?>
```
Is it enough to change label to `$this->__($option['label'])` or do you have a better idea?

sreichel:


blopa:
+1 for this

sreichel:
Question .... maybe this is out of scope of this project, but how about adding a "customer attribute manager"?

For those who won't change `gender.phtml` manually this could be usefull ... https://magento.stackexchange.com/a/196815/46249

Flyingmana:
Most shops I worked with use the prefix customer attribute for this, which is already configurable.
But yes, its out of scope and deserves an own ticket.

PR #247, opened 2017-03-20
Fix this >get massaction id field

seansan:
I could not get setMassactionIdField to work in a custom grid and stumbled on I think what is an error in th code. setMassactionIdField is never being respected is seems .... (please comfirm)

What I did

- Add use_index to massaction checkbox data array
- Remove check for getNoObjectId
- Fix reversal of merge (better other way around)

Needs logic confirmation
Also there are only a handfull of grids that use setMassactionIdField if they actually start working that may break something again

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -611,9 +611,10 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
     protected function _prepareMassactionColumn()
     {
         $columnId = 'massaction';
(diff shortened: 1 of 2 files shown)


seansan:
Did someone test this by any chance? 

I am seeing some strange behavior where clicking 

- "select all" in a grid (AOE scheduler) **unchecks** all checkboxes, and the counter states 94 selected

- and when I click select none it **checks** all checkboxes, and the counter states 0 selected

One would expect the other way around. Dont know of related to this pull. But might be. 

image

image

sreichel:
Use AOE scheduler to and cannot reproduce.

Please share some (custom grid) code to reproduce.

sreichel:
Would like to reproduce. Can you please share some code for?

PR #248, opened 2017-03-22
Sitemap dispatch event to add custom data

bastienlm:
Create new Dispatch event before end of generate xml function for the sitemap.

Perfect to add your custom link on the sitemap.
Use exemple :
```
/**
     * Add brand url to sitemap
     *
     * @event sitemap_generate_before
     * @param Varien_Event_Observer $observer
     */
    public function addBrandToSitemap(Varien_Event_Observer $observer)
    {
        $io               = $observer->getEvent()->getFile();
        $baseUrl    = $observer->getEvent()->getBaseUrl();
        $date         = $observer->getEvent()->getDate();
        $storeId    = $observer->getEvent()->getStoreId();

        $changefreq = (string)Mage::getStoreConfig('sitemap/brand/changefreq', $storeId);
        $priority   = (string)Mage::getStoreConfig('sitemap/brand/priority', $storeId);

        $collection = Mage::getResourceModel('brand/brand_collection')
            ->addActiveFilter()
        ;


        foreach ($collection as $brand) {
            $url = htmlspecialchars($baseUrl . Mage::helper('brand')->getBrandUrl($brand));

            if (Mage::getStoreConfig('web/seo/trailingslash'))
            {
                if (!preg_match('/\\.(rss|html|htm|xml|php?)$/', strtolower($url)) && substr($url, -1) != '/')
                {
                    $url .= '/';
                }
            }

            $xml = sprintf('<url><loc>%s</loc><lastmod>%s</lastmod><changefreq>%s</changefreq><priority>%.1f</priority></url>',
                $url,
                $date,
                $changefreq,
                $priority
            );
            $io->streamWrite($xml);
        }
        unset($collection);

    }
```

### DIFF
--- a/app/code/core/Mage/Sitemap/Model/Sitemap.php
+++ b/app/code/core/Mage/Sitemap/Model/Sitemap.php
@@ -221,6 +221,13 @@ class Mage_Sitemap_Model_Sitemap extends Mage_Core_Model_Abstract
         }
         unset($collection);
 
+        Mage::dispatchEvent('sitemap_generate_before', array(
+            'file'      => $io ,
+            'base_url'  => $baseUrl ,
+            'date'      => $date,
+            'store_id'  => $storeId
+        ));
+
         $io->streamWrite('</urlset>');
         $io->streamClose();
 


colinmollenhour:
Makes sense although I think the event name is inconsistent. E.g. the one before is "sitemap_cms_pages_generat**ing**_before" and this one is the final one before the close tag so I would propose something like "sitemap_urlset_generating_before" and move it to the top instead of the end. I actually agree that "generate" is more correct than "generating" but better to be consistent with existing convention than break it. :)

sreichel:
@colinmollenhour @Flyingmana @LeeSaferite want this one in? If so, i'd update PR to close this.

seansan:
+1 update the PR and add it ... (my opinion)

I dont think new extensions etc will take it into account, but rather for custom extensions based on LTS we have a way of "reaching" the XML

sreichel:
Mhhh ... still would change the event name?

seansan:
​go for it ;)​




Not using Gmelius <https://gmelius.com?ref=mail> yet?

bastienlm:
Arf, new PR for that ? :(

PR #88, opened 2016-05-18
Fix creditmemo tax calculation bug when using multiple discounts

seansan:
Fix creditmemo tax calculation bug when using multiple discounts

Needs code review and additional testing

But this has been tested in production and solved our errors with creditmemo calculation missing value

PR #250, opened 2017-03-26
Fix creditmemo tax calculation bug when using multiple discounts (again)

seansan:
Taken from #88 / maybe can be edited now

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Creditmemo/Total/Subtotal.php
+++ b/app/code/core/Mage/Sales/Model/Order/Creditmemo/Total/Subtotal.php
@@ -58,8 +58,8 @@ class Mage_Sales_Model_Order_Creditmemo_Total_Subtotal extends Mage_Sales_Model_
         $creditmemo->setSubtotalInclTax($subtotalInclTax );
         $creditmemo->setBaseSubtotalInclTax($baseSubtotalInclTax);
 
-        $creditmemo->setGrandTotal($creditmemo->getGrandTotal() + $subtotal);
-        $creditmemo->setBaseGrandTotal($creditmemo->getBaseGrandTotal() + $baseSubtotal);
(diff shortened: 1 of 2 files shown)


seansan:
@tmotyl Can you access now? It has been a while since I made these changes. Also need to dive in again. This is why I also asked for confirmation from other testers/people that work with credit memo

tmotyl:
yes, it's editable now

sreichel:
From #88 

> When adding several product types and utilizing several (more than 1) discount code the credit memo calculation is not correct.

Please correct me, but magento does not support multiple discount codes by default.

app/code/core/Mage/Sales/Model/Order/Creditmemo/Total/Subtotal.php
```diff
@@ -58,8 +58,8 @@ public function collect(Mage_Sales_Model_Order_Creditmemo $creditmemo)
         $creditmemo->setSubtotalInclTax($subtotalInclTax );
         $creditmemo->setBaseSubtotalInclTax($baseSubtotalInclTax);
 
-        $creditmemo->setGrandTotal($creditmemo->getGrandTotal() + $subtotal);
-        $creditmemo->setBaseGrandTotal($creditmemo->getBaseGrandTotal() + $baseSubtotal);
+        $creditmemo->setGrandTotal($creditmemo->getGrandTotal() + $subtotalInclTax);
+        $creditmemo->setBaseGrandTotal($creditmemo->getBaseGrandTotal() + $baseSubtotalInclTax);
```
sreichel: Is this tested with different tax settings?

LeeSaferite:
@sreichel You are correct, Magento does not support multiple coupon codes.

sreichel:
So it should be fixed in 3rd-party code?

PR #261, opened 2017-04-24
Fixes issue with wysiwyg editor re-creating thumbnails 

salubrium:
This causes performance / database connection limit issues when using the media manager as the existing thumbnails aren't found to exist.


LeeSaferite:
This PR includes a non-related commit (the one related to PayPal). If you fix the PR then we can evaluate the fix for inclusion.

seansan:
+1 squashing is a recommendation.. and seems to block the merge. How about we merge seeing this is from 1 year ago.

sreichel:
I can confirm that this problem still exists, but i'm not sure if this fix is complete.

If i have time for it, i'll check end maybe open new PR or approve this one.

PR #262, opened 2017-05-04
Add SUPEE-4814

edannenberg:
This fixes an infinite recursion loop when validating a "Products subselect" cart price rule.
Apparently the patch only got released for EE, but CE suffers from the same problem. :/

To reproduce the issue create a new cart price rule and add a "Products subselect" condition with a constraint on manufacturer or any other product attribute. Then try to apply the coupon in the frontend.

source: https://gist.github.com/piotrekkaminski/54529dadb0bc01a62a2d

### DIFF
--- a/app/code/core/Mage/SalesRule/Model/Rule/Condition/Product/Subselect.php
+++ b/app/code/core/Mage/SalesRule/Model/Rule/Condition/Product/Subselect.php
@@ -114,7 +114,7 @@ class Mage_SalesRule_Model_Rule_Condition_Product_Subselect
         $attr = $this->getAttribute();
         $total = 0;
         foreach ($object->getQuote()->getAllVisibleItems() as $item) {
-            if (parent::validate($item)) {
+            if (Mage_Rule_Model_Condition_Combine::validate($item)) {
                 $total += $item->getData($attr);
             }
         }


Flyingmana:
are we actually allowed to take an EE only patch? I had thought there is a non-disclosure clause somewhere

sreichel:
@Flyingmana understand your concerns, but if we arnt allowed to, it should be enough to change PR title? ;)

Flyingmana:
you are missing the point of copyright here...
its the content, which is problematic. Besides you cant change the title from inside the already merged commit.

LeeSaferite:
First, the content of a patch is debatable of it's copyrightable. Second, if the fix is to the open source code, then there should be a patch for the core code as well, no? (On my phone or I'd check)

sreichel:
How can the content be copyrightable? What if you found exactly the same solution for this problem? Not allowed to merge b/c there exits an EE-only patch?

LeeSaferite:
I am not arguing that it's copyrighted, I'm saying the opposite. This single-line change does not rise to the level of being copyrightable. Additionally, it's a change that's part of an open source codebase. My only real question was if they did not release this patch for the CE codebase at all?

sreichel:
> I am not arguing that it's copyrighted, I'm saying the opposite.

I know, was addressed to Daniel.

> My only real question was if they did not release this patch for the CE codebase at all?

As for as I know there was no patch for CE.

edannenberg:
The reason this SUPEE never made it into an official CE is not copyright or intentional, but rather a slip up by Magento. Most likely an EE customer filed a bug with Magento EE support, they resolved the issue and didn't bother to check if this should also be applied to CE. Not the first time this happened.

PR #263, opened 2017-05-05
Updated Mage_Catalog_Model_Resource_Url

googlygoo:
Updated Mage_Catalog_Model_Resource_Url, fixed is_active attribute when calling _getCategories()

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Url.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Url.php
@@ -702,7 +702,7 @@ class Mage_Catalog_Model_Resource_Url extends Mage_Core_Model_Resource_Db_Abstra
         if (!is_array($categoryIds)) {
             $categoryIds = array($categoryIds);
         }
-        $isActiveExpr = $adapter->getCheckSql('c.value_id > 0', 'c.value', 'c.value');
+        $isActiveExpr = $adapter->getCheckSql('c.value_id > 0', 'c.value', 'd.value');
         $select = $adapter->select()
             ->from(array('main_table' => $this->getTable('catalog/category')), array(
                 'main_table.entity_id',


seansan:
This may make the pull from last year fixing the frontend sitemap obsolete
(need id here)

Am on Phone

tmotyl:
Can somebody give more verbose explanation of this chnage and steps how to reproduce the error, so I can validate the fix?

LeeSaferite:
The original code is buggy because both paths out of the IF use the same exact column ```c.value```.

    $isActiveExpr = $adapter->getCheckSql('c.value_id > 0', 'c.value', 'c.value');

The tables ```c``` and ```d``` are both the same table with ```c``` being the store specific value and ```d``` being the default or global value.
    
    $table = $this->getTable(array('catalog/category', 'int'));
    $select->joinLeft(array('d' => $table),
        'd.attribute_id = :attribute_id AND d.store_id = 0 AND d.entity_id = main_table.entity_id',         
        array()
    )
    ->joinLeft(array('c' => $table),
        'c.attribute_id = :attribute_id AND c.store_id = :store_id AND c.entity_id = main_table.entity_id',
        array()
    );

The ```$table``` reference above is the ```catalog/category``` entity ```int``` value table and more specifically is being used to look up the ```is_active``` attribute.

The whole point of the buggy code is to find the store-specific ```is_active``` setting for a category and, if that doesn't exist, fall back to the default value for ```is_active``` for that category. The bug is that it was ignoring the default value and always using the store-specific value. Since the store specific value doesn't exist in many case it will end up as a NULL and then a FALSE. Meaning that unless you set the ```is_active``` flag for the category at the store level, it was going to be false.

As for testing this, I think the original PR author (@googlygoo) is better suited to give you a test case. I just looked at the code in abstract and confirmed the code was buggy and this PR fixed that bug.

googlygoo:
The description the @LeeSaferite has given is 100% on point. With regards to proving that there is a error, out of the default Magento code base I have not got any. The reason for this is that through my investigation the is_active attribute is not getting when generating the URL's for the url rewrite table, if it was this issue would of been picked up as it would of had a profound impact. 

For full disclosure, this error was only discovered through the building of a custom module and noticing that all data for is_active was coming through as null.

If you would like to see the change in action here is the generated query before and after that you should be able to run on any Magento installation as long as you still have the default category that is created on installation. In the result you will notice that in the column is_active, there will be a number of null records returned vs the working query.
``

PR #271, opened 2017-06-01
Suggesting to change default font to HELVETICA in pdf & reduce PDF fi…

seansan:
…lesize

Suggesting to change default font to HELVETICA in pdf & reduce PDF filesize

We have seen LinLibertine Pdf files ranging from 1.5 = 2 MB files .... per file whereas the same PDF in Helvetica is 40Kb ... 

This is not the end solution as I can imagine "just changing" the font has too much impact - maybe ... 

A suggestion would be to add a FONT selector in the backend? Maybe populated from Font.php? https://github.com/colinmollenhour/magento-zend/blob/master/Pdf/Font.php

This is a recording of the fix we execute every time after an upgrade

Feedback welcome on how to tackle something like this; without breaking BC

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php
+++ b/app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php
@@ -790,8 +790,9 @@ abstract class Mage_Sales_Model_Order_Pdf_Abstract extends Varien_Object
      */
     protected function _setFontRegular($object, $size = 7)
     {
-        $font = Zend_Pdf_Font::fontWithPath(Mage::getBaseDir() . '/lib/LinLibertineFont/LinLibertine_Re-4.4.1.ttf');
-        $object->setFont($font, $size);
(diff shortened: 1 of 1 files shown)


Flyingmana:
is the space the font needs similar to the old one? If yes, and if all characters are covered which are also part of the previous used one, then I would say this is not a BC breaking change.

Do you have a screenshot to show the differences side by side?

sreichel:
This should already be merged with https://github.com/OpenMage/magento-lts/pull/86/commits/2ecd07a91835a177feb2bb5439849e7c3c5aa319???

Flyingmana:
right, it should... https://github.com/OpenMage/magento-lts/pull/86
but then the diff should not show differences...
I can check later what is going wrong there

sreichel:
Got it ...

#86 was for app/code/core/Mage/Sales/Model/Order/Pdf/Items/Abstract.php
#271 is for app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php

seansan:
Aha! I agree. Now do see I was missing BOLD and ITALICs ...

I'll shoot in a pull fixing it.

https://github.com/OpenMage/magento-lts/pull/272

PR #272, opened 2017-06-01
Fix missing BOLD and ITALIC in HELVETICA FONT - Order/Pdf/Items/Abstract.php

seansan:


sreichel:
Why did you close #271?

There are two files where font has to be replaced ...

- app/code/core/Mage/Sales/Model/Order/Pdf/Items/Abstract.php (#86, #272)
- app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php (#271)

seansan:
Yes I realize that now too .... I will close all ... and resubmit 1 complete

PR #276, opened 2017-06-04
Release merge 1.9.3.3

Flyingmana:
there were no major conflicts during the merge.

does not yet contain additional fixes released by the community.

Should we merge it now, or wait for an Update by magento regarding the Patches?

- [ ] remove old var/package files


Flyingmana:
the additional commits seem to be a side effect of the last release was not merged from the mirror repo.
I could not see strange side effect from this, seems git handles this well enough as the diffs were identical.

drobinson:
Typically what I do for these is 

1. Merge the new codebase into OpenMage/magento-mirror
2. Generate a patch from there via `git diff HEAD~1 HEAD > diff.patch`
3. Apply this patch to OpenMage/magento-lts

Not sure how you created this one, but the process above seems to take care of all of our changes here and results in a clean single commit merge.  However, I was planning on waiting for V2 of the patch to avoid any messiness with 'upgrading' the patch 🙄 .  Seems the patch covers issues that require a secondary attack vector, so I wasn't too worried about getting it up before V2 of the patch.  Thoughts?

Flyingmana:
I see, I directly merge from the magento-mirror, to make continuous applying easier and have a better prove of being the original. (you dont need to review the whole changeset again with a merge, if you already did it on the mage-mirror repository)

the patch contains a fix for an attack, which can circumvent path based firewalls. For example if users protect their admin url over an additional htaccess rule, this does not offer additional protection without the patch. (thats just what I did understand from one of the security hole descriptions).

But people can still use the official patch file, so iam ok with just waiting for the next patch.
I dont expect much messiness, as we dont add the patch, but the newer magento version which contains the patch.

googlygoo:
To add my two cents, the code looks good. To test I went and created a patch between 1.9.3.2. and 1.9.3.3 to confirm that only the upgraded files have been modified and all community commits are still in place. The patch I created can be found [https://gist.github.com/googlygoo/cf5268a4752942d074d775300d949de0](https://gist.github.com/googlygoo/cf5268a4752942d074d775300d949de0)

The one thing I will mention though is some of the version-specific files in var/package like var/package/Phoenix_Moneybookers-1.9.3.2.xml have not been removed.

colinmollenhour:
I suggest fixing the "Allow symlinks" debacle before release. My preferred fix is to remove the option entirely and only check the paths for '..' and not use `realpath()` so the files may be located anywhere, just the template paths are restricted from traversing backwards.

Flyingmana:
removing the option completely will break a lot of dev setups which are using symlinks because of for example modman or composer. I would prefer to talk about this in a complete own Issue/PR.

And yeah, the old var packages are a good point, will remove them with an additional commit

colinmollenhour:
> removing the option completely will break a lot of dev setups which are using symlinks because of for example modman or composer

My proposal is to remove the config option and replace the faulty fix with one that is compatible with symlinks yet still addresses the real issue which is referencing unauthorized template files. Here is the fix that has existed for over a year publicly and I've been using for several years: https://gist.github.com/colinmollenhour/35c0d32dd651c4c8c840/revisions

The only case in which this would break existing installations would be if users were intentionally referencing templates paths with ".." in them which is so appallingly *wrong* that these users would almost deserve to have their stores broken... 😆  Even then, worst case a few templates would be broken and paths would need to be fixed which is a lot better than everyone having to completely change their deployment setups to not use symlinks.

Flyingmana:
Regarding re-enabling the Symlink setting in any way.
I want to remind everyone, that over a course of 5 years more then a hand full of Magento Community members did a more or less intensive Review of this setting and did not find any security problems regarding enabling it, despite the very clear warning on the setting itself which did tell you about possible security Issues with this setting.
Some of them are seen as security experts inside the magento community.

My Question is, do "you"(and I mean everyone, who recently did talk about this being harmless to re-enable) really think, you are a lot more capable to evaluate such a security risk, then the ones who have long experience and still did fail?



Also, there is a new patch release, I would like to have this one accepted for merge before the other one, or should I just add it to this PR?

infabo:
@Flyingmana I don't get the message.

drobinson:
@Flyingmana I would like merge this one, then apply the diff from magento-mirror for 1.9.3.4 here as well.  I think we should tackle the symlink issue in a separate discussion, even though IMO it makes this "patch level" release backwards incompatible.

One question - why does this PR contain commits relating to the 1.9.3.2 release?

Flyingmana:
because I pulled it directly from the openMage mirror, and the 1.9.3.2 release was added as a normal patch, not via pull from there

drobinson:
Got it - it's probably less confusing for the commit history here if we can continue with the normal patching process.  It would be weird to see something about 1.9.3.2 at the top of the commits at this point.  I'm also not sure the diff would be 100% correct since we have so many extra changes in this repo.  Do you want to do that or shall I?

Flyingmana:
I removed the 1.9.3.2 files in var/packages in an additional commit. 
Also I checked if the merge to the 1.9.3.4 is working, and it is https://github.com/Flyingmana/magento-lts/tree/release-merge-1.9.3.4

So after this was merged, I can directly add the PR for the next version

drobinson:
Ok - I will merge this one then, though I'm still unsure how the older commits also being merged will affect things (if at all).  Will need to test.

PR #286, opened 2017-06-13
Fix mem limit

seansan:
Described here: https://magento.com/tech-resources/bug-tracking/issue/index/id/1540/

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Image.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Image.php
@@ -183,7 +183,7 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     {
         $memoryLimit = trim(strtoupper(ini_get('memory_limit')));
 
-        if (!isSet($memoryLimit[0])){
+        if (!isSet($memoryLimit)){
             $memoryLimit = "128M";
         }
 


sreichel:
Original Post:

__Steps to reproduce:__

in `app/code/core/Mage/Catalog/Model/Product/Image.php`, on line 186,

    if (!isSet($memoryLimit[0])) {

incorrectly refers to $memoryLimit as an array with index [0]. The variable is not an array, so the if statement always evaluates to true and the memoryLimit is ALWAYS interpreted to be 128M regardless of the setting in php.ini

    if (!isSet($memoryLimit[0])) {
        $memoryLimit = "128M";
    }

Corrected code below:

    if (!isSet($memoryLimit)) {
        $memoryLimit = "128M";
    }

__Expected Result:__

Memory limit should be pulled from php.ini from ini

__Actual Result:__

always evaluates to true and uses 128M hardcoded setting from code.

---

@seansan This is not a bug ... it's absolutly corret! `$memoryLimit[0]` gives you the first char of $memoryLimit __string__. If `ini_get('memory_limit')` returns an empty string limit is set to 128M.

With this PR `!isSet($memoryLimit)` would be `false` because it is set - it's just an empty string - and return value would be `string '' (length=0)`

See: http://php.net/manual/en/function.isset.php (Example #1)

So please check code before PR :)

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -183,7 +183,7 @@ protected function _getMemoryLimit()
     {
         $memoryLimit = trim(strtoupper(ini_get('memory_limit')));
 
-        if (!isSet($memoryLimit[0])){
+        if (!isSet($memoryLimit)){
```
LeeSaferite: I don't think this is doing what you think it's doing. 
`$memoryLimit = trim(strtoupper(ini_get('memory_limit')));` will always set the variable. As far as I can tell, `!empty($memoryLimit)` is the proper check in this case.

seansan:
@sreichel right you are
So inderdaad seems to be correct (or at least checking for that first char)

app/code/core/Mage/Catalog/Model/Product/Image.php
seansan: Maybe keep it the way it is?

sreichel: For me ... I see no need to change this.

colinmollenhour: The value of `isset()` on an empty string is `true`. This should be `if (empty($memoryLimit)) {`.

```
php > $a = '';
php > var_dump(isset($a));
php shell code:1:
bool(true)
```

PR #293, opened 2017-06-16
Typo in category controller

DmitryFursNeklo:
_initCatagory -> _initCategory


Flyingmana:
thats right, but Iam afraid this will break some setups. I saw to often that people did rewrite this kind of controllers

colinmollenhour:
You could add a new method with the old name which just calls the correct one and mark that one `@deprecated` in the vardoc so that old rewrites would not be broken.

Issue #267, opened 2017-05-22
TBD: setEmailSent on Invoice


infabo:
This one should be discussed if it is a bug or an intended behaviour.

When you create a new invoice, you have this checkbox:

![grafik](https://cloud.githubusercontent.com/assets/3999104/26299839/8e75bd6c-3edc-11e7-897f-f2a2a7e4d5ef.png)

If not checked and no "invoice-copy-to-recepiants" defined (configuration / sales emails) -> all working fine and you get the expected status-message:

![grafik](https://cloud.githubusercontent.com/assets/3999104/26300027/35cdab6a-3edd-11e7-92d7-fb42c2c5c4cb.png)

Once you have entered some "invoice-copy-to-recepiants":

![grafik](https://cloud.githubusercontent.com/assets/3999104/26300095/6ca1f326-3edd-11e7-9cf9-9b1496076601.png)

In both cases no email is sent to the customer customer -> expected behaviour.

Just because an CC email is sent, the "email_sent" flag is set to true.

Received feedback from several merchants, saying this is confusing.

source:

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Sales/Model/Order/Invoice.php#L847

seansan:
I am a merchant ;) And would agree: all messages on order level implicitly
implie *the customer*

LeeSaferite:
I'd personally say this is 100% a bug. The implicit "sent to customer" in that message is something I've always taken for granted.

You can see that Invoice and Creditmemo are both exhibiting this behaviour.
[Invoice](https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Sales/Model/Order/Invoice.php#L847)
[Creditmemo](https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Sales/Model/Order/Creditmemo.php#L794)

It seems like Shipments are not just due to someone forgetting to add that code in the sendEmail method. The controllers that generate the calls to sendEmail are *also* setting the flag, but for Invoice and Creditmemo that are being overridden.

[...]

PR #297, opened 2017-07-08
[FIX] set email as sent only if customer notified

infabo:
In all other places where a "email" flag is used, email_sent is only set if the flag is true.

PS:
In my opinion, the $notifiyCustomer condition is a better solution than assigning $notifiyCustomer directly to "email_sent". One can send invoices/creditmemos multiple times. Just a second run could make it look like an email was never sent (even though it was sent in the first run).


infabo:
This is meant as a hotfix for a bug described here #267.
Still needs the [mentioned refactoring](https://github.com/OpenMage/magento-lts/issues/267#issuecomment-303122459). But out of scope IMHO -> how to handle bigger code changes when Magento still did not drop support for 1.x?

Issue #295, opened 2017-06-28
Slow query with MySQL 5.7


infabo:
After upgrading from MySQL 5.6 to 5.7 on my devbox, I have experienced a massive performance drop on Magento's homepage. After some investigation (profiling, ...) I have quickly determined a single database query to be responsible for it (at least it had the biggest impact).
Can't remember exactly, but was 45sec pageload and with the "hack" I got down to 11 secs.
Seems like there are several other fuckups due to MySQL 5.7, because on 5.6 the Magento instance is/was rocking fast. (maybe those visitor logs, etc. are not optimized too).

http://itstickers.besthw.net/en/magento-mysql-5-7-slow-queries/

So....I'd like to ask here: does or should magento-lts accept patches like these?

tomekjordan:
is Maria DB 5.7 also suffering the same problem?

colinmollenhour:
This is exactly the kind of patch this project should accept with substantial proof there are no regressions.

PR #303, opened 2017-07-17
Fix slow url_rewrite query on MySQL 5.7

tmotyl:
Fixes: https://github.com/OpenMage/magento-lts/issues/295

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Category/Url/Rewrite.php
+++ b/app/code/core/Mage/Catalog/Helper/Category/Url/Rewrite.php
@@ -75,6 +75,7 @@ class Mage_Catalog_Helper_Category_Url_Rewrite
             array('request_path'),
             "{{table}}.is_system=1 AND " .
                 "{{table}}.store_id='{$storeId}' AND " .
+                "{{table}}.category_id IS NOT NULL AND " .
                 "{{table}}.id_path LIKE 'category/%'",
             'left'
         );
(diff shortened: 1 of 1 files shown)


tmotyl:
I'm planning to test that next days

tomekjordan:
What about 3rd party modules for SEO like Mageworx SEO Ultimate, will it brake anything if they rewrite urls?

LeeSaferite:
If you read the code, it's joining the category entries to the url rewrite table based on the category_id field of the rewrite table. This added WHERE clause is just an optimization to exclude records that wouldn't be joined anyway.

tmotyl:
@tomekjordan I don't expect this to be breaking however
I'm not using this module for 100%, so can't tell. 
If you have access to this module, please test the patch and let us know.

darnux:
any news on that? in which version is this fixed?

colinmollenhour:
While I approve of the change in principle, there is no proof provided that it doesn't break anything and I don't have time to test it thoroughly myself, so perhaps if someone could test it in some capacity and report back that would help get this merged.

We really need some unit tests... 😕

darnux:
@colinmollenhour thanks for your quick reply. 
I am working for a webhosting company and we are using this fix in production for several projects (about 5-10) at the moment due to the need of MySQL 5.7. 
We did not notice any negative impact on this by now.

BassemN:
Thanks allot for that fix

wilfriedwolf:
Thanks a lot from my side too! I had a 7 s load after adding a second web-site to one of our magento installations. When I tested it locally after switching to magento_lts loading time dropped to 0.32s!
Great!

Issue #305, opened 2017-07-19
admin create order -> shipping_as_billing is 0 when set shipping method


zamoroka:
steps to reproduce: 
1. Create New Order from admin
2. Set "Shipping Address" Same As Billing.
3. Choose Shipping Method. 
> now Mage::getModel('Mage_Sales_Model_Quote_Address')->getSameAsBilling() == 1
4. Change Billing Address
5. Press "Update Items and Qty's"
6. Choose Shipping Method
> now Mage::getModel('Mage_Sales_Model_Quote_Address')->getSameAsBilling() == 0

tomekjordan:
Save addres for new customer should be also chcked as default for admin orders.

seansan:
@Tomek: as a merchant: I agree

tomekjordan:
UPDATE: save adress for new customer - the same issue comes when you as admin create order for old customer but for other website - order gets all address data, but due to not-checked save address - new customer re-created for new websites has empty address in the customer account. so definitiely check SAVE address should be set by default. thanks!

sreichel:
Fixed in #306.

PR #306, opened 2017-07-19
shipping_as_billing is 0

zamoroka:
issue: https://github.com/OpenMage/magento-lts/issues/305

'data' is sends via ajax to `Mage_Adminhtml_Sales_Order_CreateController`. In  `_processActionData` method there is a check for `$syncFlag = $this->getRequest()->getPost('shipping_as_billing');` and in this case `(int)$syncFlag == 0`. So we need to send data['shipping_as_billing'] via post request to save the right information


seansan:
+1 can this be merged?

Issue #307, opened 2017-07-26
Remove password from account_new email template


kkrieger85:
It is not 2017 to send choosen password in email:

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/locale/en_US/template/email/account_new.html#L25

Flyingmana:
only applicable if the email authentification is forced to be always on (actually I dont know how the state is for this currently, I just know people who disabled it for conversion improvements)


But yes, Iam absolutely for this change

tomekjordan:
There are many studies from UX that shows that removing plain passwords is bad for store, because customers are not able to retrieve passwords, and they are creating new accounts and whole mess. Risk for frauds is really low. Because if email is sent to EMAIL - user must login to his service - so I would not change it. Each store can remove this one line in template by themself.
As I said - if customer don't have access to its email - he cannot change anything - so risk is LOW.

tomekjordan:
Proof:

"Why is it being sent to an email when it is being reminded of a password?

In response to the questions our clients have sent us, is there an available, one-of-a-kind way to remember passwords for online shoppers at IAI-Shop.com? This is the answer why we have not decided to lose customers' passwords irretrievably by their "hashing". The basic reason why we have not decided on this solution is that the solution chosen by us simply pays for online stores, especially the bigger one. For this reason too, we have tried not to betray these reasons as long as possible.

By doing our daily work we face thousands of technological decisions. While making them, not only do we reflect on our 12-year experience, but also on the behavioral analysis of clients and their clients. One such observation is that a typically organized reminder, previously "uploaded" password is based on sending a link email, upon which users must enter a new password. As our experience shows, a few years ago, with such an organized process, it is simply more convenient to set up a new account by a customer. And with this movement, consciously or not, all efforts are undermined, for example, in rewarding customers with prizes, rebates, points or even hindering analysis of sales of goods, or advanced marketing techniques such as recommendation systems or targeted email marketing. These and other solutions are sewn at IAI-Shop.com. So if we were to put on a client account every time a new account was created, what would we build these systems for?

Therefore, hogging the passwords of customers in the online store, in our opinion, is a bad idea and using this solution effectively reduces turnover on returning customers.

[...]

PR #308, opened 2017-07-26
remove customer password from email template

kkrieger85:
refer to #307

### DIFF
--- a/app/locale/en_US/template/email/account_new.html
+++ b/app/locale/en_US/template/email/account_new.html
@@ -22,7 +22,6 @@
             <p class="highlighted-text">
                 Use the following values when prompted to log in:<br/>
                 <strong>Email</strong>: {{var customer.email}}<br/>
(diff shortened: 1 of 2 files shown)


app/locale/en_US/template/email/account_new.html
```diff
@@ -22,7 +22,6 @@
             <p class="highlighted-text">
                 Use the following values when prompted to log in:<br/>
                 <strong>Email</strong>: {{var customer.email}}<br/>
-                <strong>Password</strong>: {{htmlescape var=$customer.password}}
```
infabo: What about

`<strong>Password</strong>: Password you set when creating account`

blopa: That's what I use in my store.

kkrieger85: Let's vote both options :)

PR #317, opened 2017-07-26
Fixes system config dependencies for multiselect fields

sreichel:
Fixes an issue where `<depends>` on a multiselect config field does not work correctly if multiple values are selected.

Example config code here: https://magento.stackexchange.com/questions/179664/admin-config-show-field-depending-on-selected-multiselect-values


sreichel:
Yep. I think in realcase it's nearly useles - who uses such settings? :) However, another review would be nice  ... :)

sreichel:
@colinmollenhour "review needed" can be removed, if code style is OK ... it works as it should :)

PR #318, opened 2017-07-26
Fixed undefined variable when creating shipping labels

sreichel:
Not sure about exact setup, but wen creating shipping labels an undfined variable error is logged because of setting `$shipment` variable inside try/catch block an use it outside arround L253

Review needed.


infabo:
LGTM

PR #321, opened 2017-07-26
Add possibility to remove link at sales order and account edit page

sreichel:
... if i can add something from layout.xml, I should be able to remove it too ...


sreichel:
I just took this from [N98_LayoutHelper](https://github.com/netz98/N98_LayoutHelper), that I alway install ... if full code would be accepted I'll create another PR.

PR #323, opened 2017-07-27
Fixes wrong usage of getSelect()->order() method

sreichel:
Fixed `order()` methods where direction is used as (non-existing) second parameter.

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/lib/Zend/Db/Select.php#L577-L587

    /**
     * Adds a row order to the query.
     *
     * @param mixed $spec The column(s) and direction to order by.
     * @return Zend_Db_Select This Zend_Db_Select object.
     */
    public function order($spec)
    {
        if (!is_array($spec)) {
            $spec = array($spec);
        }

Used `$ grep -rn "order('[a-z_]*'," ./app/code/core/Mage/` ... so that should be all.


sreichel:
Please merge. :P

PR #324, opened 2017-07-30
Magento order status updated to “Processing” even when PayPal payment fails

dvdsndr:
Must remove invoices before cancel, otherwise order cancel will fail and status will go to go to processing

### DIFF
--- a/app/code/core/Mage/Paypal/Model/Ipn.php
+++ b/app/code/core/Mage/Paypal/Model/Ipn.php
@@ -532,6 +532,10 @@ class Mage_Paypal_Model_Ipn
     protected function _registerPaymentFailure()
     {
         $this->_importPaymentInformation();
+        // paypal bug, must remove invoice before order cancel 
+        foreach ($this->_order->getInvoiceCollection() as $invoice){
+            $invoice->cancel()->save();
+        }
         $this->_order
             ->registerCancellation($this->_createIpnComment(''), false)
             ->save();


dvdsndr:
Hi more detail on this at:
https://stackoverflow.com/questions/30936587/magento-order-status-updated-to-processing-even-when-paypal-payment-fails.

This is really bad for stores with ERP or warehouse integrations. Most integrations will use the status of "processing" to assume that the payment was ok and thus the goods can be shipped.

dvdsndr:
Really surprised that others are not seeing this issue. 

If payment fails, status should go Payment-Reveview > Cancelled
This bug is causing the status to go Payment-Review > Processing

Processing is the status for completed payments so consumers will get free goods!

Paypal does try to cancel the order but the cancellation fails as the order has a valid invoice, the invoice must be removed for the cancel to work

tomekjordan:
You cannot just remove invoice. because for EU vat invoicing you must have all invoices in the system, one by one, without gap in numbering. the best if paypal would not create invoice automatically before successful payment is made. which probably is a standard behavior

dvdsndr:
Thats a fix out of my depth. Will need to leave that for others. Wonder if there is a way to mark the invoice cancelled instead of having to remove it??

tbaden:
agree with @tomekjordan 
the best way is to stop the automatic generation of invoices before the successfull payment

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @dvdsndr  if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

ADDISON74:
@fballiano - I noticed that you closed some of them and then rebased to 1.9.4.x. Where you can't do this please mark them with this label "needs rewrite for new versions" to keep track of them. When you finish please leave a message in any post you want to double check too.

fballiano:
> @fballiano - I noticed that you closed some of them and then rebased to 1.9.4.x. Where you can't do this please mark them with this label "needs rewrite for new versions" to keep track of them. When you finish please leave a message in any post you want to double check too.

as you see from the comment above that's precisely what I wrote :-)
in other PRs I removed the message since @tmotyl explained me that I could rebase the PRs myself ;-)

ADDISON74:
Good advice from @tmotyl. 

Shouldn't you rebase this PR too?

Did you rebase the PR's from this here #2184?

fballiano:
can't be done since original author deleted the repo. anyway https://github.com/OpenMage/magento-lts/pull/2183 it's exactly the same thing

ADDISON74:
In this case I will mark this one and #2183 with that label and we will have to recreate them at some point if those changes still exist. It was expected that we would encounter such issues in the conditions in which some PRs had been posted for a long time.

fballiano:
Again, if we approve that other pr it is exactly the same as this one. No need for anything.

PR #326, opened 2017-07-31
Use joinLeft instead of joinInner to fix customers missing from reports.

colinmollenhour:
Refs #41
Refs #241
This commit also reverts db69295f8


infabo:
wow

PR #327, opened 2017-08-04
Fix date selection when current date is disabled

mattdavenport:
When working with `calendar.js` (utilized in admin grid filtering and other date selector functions) in a recent project, I came across a bug that this PR resolves.

Using the `disableFunc` property within the `Calendar.setup` method, if the current date is disabled selection of other dates will not work properly. In full disclosure, I discovered this fix via [this Stackoverflow answer](https://stackoverflow.com/a/28073450/1412625), but it works quite well and has no sides effects that I've noticed during testing.

### DIFF
--- a/js/calendar/calendar.js
+++ b/js/calendar/calendar.js
@@ -1246,6 +1246,8 @@ Calendar.prototype._init = function (firstDayOfWeek, date) {
                 }
                 if (weekend.indexOf(wday.toString()) != -1)
                     cell.className += cell.otherMonth ? " oweekend" : " weekend";
+            } else {
+                this.currentDateEl = cell;
             }
         }
         if (!(hasdays || this.showsOtherMonths))


sreichel:
Can you please provide some code to test?

mattdavenport:
Hey @sreichel, apologies for the delay on this one. When digging back into the code, I found I've actually extended the [TBT_EnhancedGrid](https://github.com/jayelkaake/enhancedgrid) module with the functionality. I don't think this issue is related to that module, however, as it is still utilizing Magento's `calendar.js` for the frontend js functionality. I've create a gist [here](https://gist.github.com/mattdavenport/c42d539f60c2db7d85529337ae07e2cf) and the grid column+filter code is as follows:

`Grid Column`
```php
$this->addColumn('created_at', [
            'header'                    => $this->__('Week'),
            'align'                     => 'center',
            'header_css_class'          => 'filter-dateweek',
            'index'                     => 'created_at',
            'type'                      => 'dateweek',
            'filter_condition_callback' => array($this, '_weekFilter'),
        ]);
```

`Grid Filter Callback`
```php
    /**
     * Using a single date, determine the week range
     * and filter BETWEEN the week start and end date
     */
    protected function _weekFilter($grid, $column)
    {
        if (!$value = $column->getFilter()->getValue()) {
            return $this;
        }

        if (!is_array($value) || !isset($value['week'])) {
            return $this;
        }

        /**
         * Calculate week start and end dates from
         * selected filter date
         */
        /** @var Zend_Date $weekDate */
        $weekDate = $value['week'];
        $filterTimestamp = $weekDate->getTimestamp();

        $weekDay   = date('w', $filterTimestamp);
        $startDate = date('Y-m-d 00:00:00', $filterTimestamp - (86400 * $weekDay));
        $endDate   = date('Y-m-d 23:59:59', $filterTimestamp + 86400 * (6 - $weekDay));

        $this->getCollection()->addFieldToFilter('created_at', ['gteq' => $startDate]);
        $this->getCollection()->addFieldToFilter('created_at', ['lteq' => $endDate]);

        return $this;
    }
```

sreichel:
Thanks @mattdavenport, I will test it later :)

PR #329, opened 2017-08-18
Add username to order comments

seansan:
Add username to order comments (we we know who actually did it ;) - Lets make it a pull to discuss because there are some caveats ... 

Improvements
- Can add system setting to turn on / off 

Problems
- When we check "Visible on frontend" then the customer could potentially see the username / dont know how to circumvent this
- this message can be shown in transactional emails
- Solution would be to maybe only capture backend history comments by real users (not system/default processes) 

Cleaner solution is
- https://stackoverflow.com/questions/6219620/add-username-to-order-comment-history

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -1083,6 +1083,12 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
         } else {
             $this->setStatus($status);
         }
+        
+         //get username and append
+        $session = Mage::getSingleton('admin/session');
+        $username = $session->getUser()->getUsername();
+        $comment .= ($username) ? " ({$username})" : ' (system)';       
+        
         $history = Mage::getModel('sales/order_status_history')
             ->setStatus($status)
             ->setComment($comment)


tomekjordan:
3rd party modules like IWD Order Manager did the same.....

seansan:
dont need extension so complex. Found way around issues. Breaks nothing. Will post new merge.

LeeSaferite:
This is 100% possible using events (I know as this is how I've done it since the early days).

While I like the idea of the feature being in core, I see two issues here. First, I don't know if this is something that should be added in LTS, we'd need a community discussion at least. Second, if we are going to add the feature then we should do it correctly. By correctly I mean adding a nullable FK column to the history table pointing to the user table and editing the template file to list out the user or something like "SYSTEM" for null entries .

    /** @var Mage_Core_Model_Resource_Setup $this */
    $this->startSetup();
    
    $this->getConnection()->addColumn(
        $this->getTable('sales/order_status_history'),
        'user_id',
        array(
            'type'     => Varien_Db_Ddl_Table::TYPE_INTEGER,
            'unsigned' => true,
            'nullable' => true,
            'comment'  => 'User that created this record'
        )
    );
    
    $this->getConnection()->addForeignKey(
        $this->getFkName('sales/order_status_history', 'user_id', 'admin/user', 'user_id'),
        $this->getTable('sales/order_status_history'),
        'user_id',
        $this->getTable('admin/user'),
        'user_id',
        Varien_Db_Ddl_Table::ACTION_SET_NULL,
        Varien_Db_Ddl_Table::ACTION_CASCADE
    );
    
    $this->endSetup();

    <sales_order_status_history_save_before>
        <observers>
            <sales_order_status_history_user>
                <class>sales/observer</class>
                <method>addUserToComment</method>
            </sales_order_status_history_user>
        </observers>
    </sales_order_status_history_save_before>

    public function addUserToComment(Varien_Event_Observer $observer)
    {
        /** @var Mage_Sales_Model_Order_Status_History $history */
        $history = $observer->getStatusHistory();
        if (!$history instanceof Mage_Sales_Model_Order_Status_History) {
            return;
        }

        // Don't touch existing history records
        if ($history->getId()) {
            return;
        }

        /** @var Mage_Admin_Model_Session $session */
        $session = Mage::getSingleton('admin/session');

        /** @var Mage_Admin_Model_User $user */
        $user = $session->getUser();
        if (!$user instanceof Mage_Admin_Model_User) {
            return;
        }

        // If there is a user and that user is not already set on the record
        if ($user->getId() && $user->getId() != $history->getUserId()) {
            $history->setUserId($user->getId());
        }
    }

sreichel:
> This is 100% possible using events
So there's no need to include this in core. Imho LTS should only include additions/new features that realy need changes in core code. Everything that can be added without rewrites belongs to an extension.

> While I like the idea of the feature being in core,
How about adding a new repo for such things and and set this repo as suggestion to `composer.json`?

sreichel:
> This is 100% possible using events (I know as this is how I've done it since the early days).

@LeeSaferite do you have this as extension?

kiatng:
I have had this implemented for a number of years in a project with a few hundreds admin users. There were a few amendments, and this is the current version:

```
    /**
     *  Observe event sales_order_status_history_save_before
     *  Add admin username and userrole to Mage_Sales_Model_Order_Status_History     
     */              
    public function orderStatusHistorySaveBefore($observer)  
    {
        $history = $observer->getEvent()->getStatusHistory();
        if (!$history->getId()) {
            $role = '';
            if (Mage::getSingleton('admin/session')->isLoggedIn()) {
                $user = Mage::getSingleton('admin/session')->getUser();
                $role = $user->getRole()->getRoleName();
                $name = $user->getUsername();         
            } elseif (Mage::getSingleton('customer/session')->isLoggedIn()) {
                $role = 'Frontend';
                $name = Mage::getSingleton('customer/session')->getCustomer()->getName();
            } elseif (Mage::getSingleton('api/session')->isLoggedIn()) {
                $user = Mage::getSingleton('api/session')->getUser();
                $role = 'API';
                $roleIds = $user->getRoles();
                if (!empty($roleIds)) {
                    $role .= '>'.Mage::getModel('api/role')->load($roleIds[0])->getRoleName();
                }
                $name = $user->getUsername();
            }
            if ($role) {
                $history->setData('username', $name);
                $history->setData('userrole', $role);
            }
        }
    }
```

I added 2 new columns:

```
$installer = $this;
$installer->startSetup();
$table = $installer->getTable('sales/order_status_history');
$installer->getConnection()
    ->addColumn($table, 'username', array(
        'type'      => Varien_Db_Ddl_Table::TYPE_TEXT,
        'length'    => 40,
        'nullable'  => true,
        'comment'   => 'Admin user name'
    ));
$installer->getConnection()
    ->addColumn($table, 'userrole', array(
        'type'      => Varien_Db_Ddl_Table::TYPE_TEXT,
        'length'    => 50,
        'nullable'  => true,
        'comment'   => 'Admin user role'
    ));    
$installer->endSetup();
```
As you can see, _username_ can cater for frontend and api users, not just admin users. This is how it looks in the order page in backend for the 3 different types of users. FYI, the API users are updating the status with a mobile device.

image

LeeSaferite:
Here's what I use. I extracted this from a project. It's so trivial I never made it into a module.

https://gist.github.com/LeeSaferite/e95310c53fc3c14c9fd5c11e57c204b1

seansan:
​yes something like this should have been added years ago
how often do we say: who did this or that in admin?​

LeeSaferite:
I like that the one from @kiatng supports identifying Admin, API, and FE users. I would suggest that if we want this feature in LTS core we take that code, clean it up a bit, make the needed template changes and make it happen. 

Progress is exciting.

PR #337, opened 2017-08-30
Show coupon code on order view in adminhtml

seansan:
…view)

solve longstanding order shortcoming (so actual coupon code on order view in adminhtml)
now our logistics see the description @ totals overview: but have to ask the customer: which code was it? esp. handy when using generated codes
simple fix: please improve if you think this can be done better. 
Breaks nothing, quickwin for admin users


seansan:
ideas? downsides? 

+1

tmotyl:
I had the same issue once. So i'm in favor of getting it in.

blopa:
+1

seansan:
+1 have tested in production works fine over here (and backend people very happy with it)

seansan:
Close and move to https://github.com/OpenMage/magento-lts/pull/447

PR #339, opened 2017-08-31
Fix errors found via PHP lint.

colinmollenhour:
Testing new PHP lint functionality.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,7 @@ matrix:
   allow_failures:
     - php: 7.2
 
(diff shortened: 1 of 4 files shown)


colinmollenhour:
Unfortunately the PHP lint is really slow.. I excluded lib/Zend since there is a lot there that isn't used and it isn't touched frequently, but we could possibly find a better way to optimize.

I think 5.6 and 7 are passing currently. Will see about the others tomorrow perhaps.

PR #343, opened 2017-09-15
Import Magento Release 1.9.3.6

drobinson:


colinmollenhour:
Hey they finally added a check for `..` in the template paths!  However, I noticed it is actually `../`. I wonder if this leaves Magento on Windows vulnerable, or is Magento on Windows even supported (I've never tried it)?  Also it seems kinda redundant to check the paths both in layout updates and right before actually loading the template (the horrid Allow Symlinks functionality), no? Odd..

drobinson:
Running `grep -r '\.\./' . | less` on the codebase shows a few other places where this is used, but only some of them also mention the "..\" notation.  ¯\_(ツ)_/¯

fheyer:
Your first commit contains a .rej file with changes that were rejected for app/code/core/Mage/Shipping/Model/Carrier/Tablerate.php

drobinson:
Sorry about that, removed in #344

PR #345, opened 2017-09-26
Fixed removeAccents method for german umlauts

sreichel:
If setting 2nd parameter to yes in `removeAccents($string, $german=false)` german umlauts should be translated from `ä` to 'ae' ...

    $str = 'Ae-Ä Oe-Ö Ue-Ü ae-ä oe-ö ue-ü';
    echo Mage::helper('core')->removeAccents($str, true);

    
**Expected:**
`Ae-Ae Oe-Oe Ue-Ue ae-ae oe-oe ue-ue`

**Current:**
`DnnÄaSnnÖaZnnÜalnnäaznnöassnnü`

### DIFF
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -317,9 +317,9 @@ class Mage_Core_Helper_Data extends Mage_Core_Helper_Abstract
 
             if ($german) {
                 // umlauts
-                $subst = array_merge($subst, array(
+                $subst = array(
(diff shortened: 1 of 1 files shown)


Flyingmana:
the problem was the array operation? we should make sure this behavior is consistent between the php versions (can someone check this?)

sreichel:
PHP version doesn't matter. `array_merge` is just wrong here, because array indexes are renumbered.

infabo:
LGTM.
This is a classic. No version thing like @sreichel already said.

Issue #350, opened 2017-10-09
Method "getChildren" sort ordering


EliasKotlyar:
Issue is described here:
https://github.com/magento/magento2/issues/11310

This issue happens in all Magento Versions.

PR #351, opened 2017-10-09
Sort categories returned by  "getChildren" by position

EliasKotlyar:
Change "getChildren" category method to return categories sorted by position.
Fixes issue https://github.com/OpenMage/magento-lts/issues/350

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Category.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Category.php
@@ -727,8 +727,8 @@ class Mage_Catalog_Model_Resource_Category extends Mage_Catalog_Model_Resource_A
                 'c.attribute_id = :attribute_id AND c.store_id = :store_id AND c.entity_id = m.entity_id',
                 array()
             )
-            ->where($checkSql . ' = :scope');
-        $select->order("m.position ASC");
+            ->where($checkSql . ' = :scope')
+            ->order("m.position ASC");
 
         return $adapter->fetchCol($select, $bind);
     }
--- a/app/code/core/Mage/Catalog/Model/Resource/Category.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Category.php
@@ -728,6 +728,7 @@ class Mage_Catalog_Model_Resource_Category extends Mage_Catalog_Model_Resource_A
                 array()
             )
             ->where($checkSql . ' = :scope');
+        $select->order("m.position ASC");
 
         return $adapter->fetchCol($select, $bind);
     }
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Catalog/Model/Resource/Category/Flat.php
```diff
@@ -1320,6 +1320,8 @@ public function getChildren($category, $recursive = true, $isActive = true)
         if ($isActive) {
             $select->where('is_active = ?', '1');
         }
+        $maintable = $this->getMainStoreTable($category->getStoreId());
```
LeeSaferite: Can you move this further up and use the variable in both locations?

EliasKotlyar: Thanks for your feedback. I have refactored it.

app/code/core/Mage/Catalog/Model/Resource/Category.php
```diff
@@ -728,6 +728,7 @@ public function getChildren($category, $recursive = true)
                 array()
             )
             ->where($checkSql . ' = :scope');
+        $select->order("m.position ASC");
```
LeeSaferite: Can we include this as part of the fluent call right after the `where` and drop the `$select`?

app/code/core/Mage/Catalog/Model/Resource/Category/Flat.php
```diff
@@ -1311,15 +1311,17 @@ public function isInRootCategoryList($category)
...
-            ->from($this->getMainStoreTable($category->getStoreId()), 'entity_id')
+            ->from($maintable, 'entity_id')
             ->where('path LIKE ?', "{$category->getPath()}/%");
         if (!$recursive) {
             $select->where('level <= ?', $category->getLevel() + 1);
         }
         if ($isActive) {
             $select->where('is_active = ?', '1');
         }
+        $select->order($maintable.".position ASC");
```
LeeSaferite: Can we include this as part of the fluent call right after the `where` and drop the `$select`? Moving this up above the if statements would be needed.

app/code/core/Mage/Catalog/Model/Resource/Category.php
EliasKotlyar: Thanks for your support. I have refactored it.

app/code/core/Mage/Catalog/Model/Resource/Category/Flat.php
EliasKotlyar: Thanks for your support. I have refactored it.

LeeSaferite:
This still needs a functional review from one of the other members as I do not currently have a good setup to test things (in the middle of moving).

sreichel:
LGTM.

Question. How about adding a `$sortByPositionFlag` like in M2s PR? (https://github.com/magento/magento2/pull/11342)

sreichel:
> This still needs a functional review from one of the other members ...

- it's not used in core
- the change works as intended

I agree with this PR with still would suggest to use an optional flag (#429) so we can enure not breaking anything,

PR #356, opened 2017-10-18
Change minimal passwd req from >5 to 7> (8 min chars)

seansan:
Change minimal passwd req from >5 to 7> (8 min chars)

not saying 8 is ideal but most systems default to 8

(for example: we interface with external system where users cannnot be created because their passwd is too short)

### DIFF
--- a/app/code/core/Mage/Customer/Model/Customer/Attribute/Backend/Password.php
+++ b/app/code/core/Mage/Customer/Model/Customer/Attribute/Backend/Password.php
@@ -43,8 +43,8 @@ class Mage_Customer_Model_Customer_Attribute_Backend_Password extends Mage_Eav_M
         $password = trim($object->getPassword());
         $len = Mage::helper('core/string')->strlen($password);
         if ($len) {
-             if ($len < 6) {
-                Mage::throwException(Mage::helper('customer')->__('The password must have at least 6 characters. Leading or trailing spaces will be ignored.'));
+             if ($len < 8) {
+                Mage::throwException(Mage::helper('customer')->__('The password must have at least 8 characters. Leading or trailing spaces will be ignored.'));
             }
             $object->setPasswordHash($object->hashPassword($password));
         }


tomekjordan:
As an merchant and user. I don't want any changes in it. If IT likes 2wer#$@^#$@SDF password cool, but don't force users to play that games.

julienloizelet:
In my opinion, I think that it could be in a config ( with a default value equals to 6 for backward compatibility).

seansan:
​this sounds logical (actually my 1st action was to Google if there was a
config setting I overlooked)

​

sreichel:
I believe this PR isnt complete. If you want to change min PW  length, it should be done everywhere ...

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Customer/Model/Customer.php#L71-L74

    /**
     * Minimum Password Length
     */
    const MINIMUM_PASSWORD_LENGTH = 6;

seansan:
​hmm I thought there was 1 location only (was reading some sources claiming
this).

Okay so in short.
 - make setting in config
 - default to current 6
 - change in all the right places

colinmollenhour:
Definitely should be both backward compatible and configurable.

sreichel:
Another one `js\prototype\validation.js`

    ['validate-password', 'Please enter 6 or more characters without leading or trailing spaces.', function(v) {
                var pass=v.strip(); /*strip leading and trailing spaces*/
                return (!(v.length>0 && v.length < 6) && v.length == pass.length);
            }],

$this and changes from `Customer/Model/Customer/Attribute/Backend/Password.php` would also require change translation in `app\locale\en_US\Mage_Core.csv`. **This would break all language packs.**

If its configurable `Mage_Customer_Model_Customer::validate()` has to be changed too. Whats about current pasword that are shorter then new config settings?

**tl;dr** Better close for now. This something for a `dev` or `feature` branch. May backporting some `M2` code where minimum length is configurable.

LeeSaferite:
Old passwords shouldn't matter for this change BTW. It *should* be impossible to determine the length of a stored password hash so you'd just issue a blanket password expiration or let the passwords naturally expire and be reset at that point.

sreichel:
Closed for now. Please submit complete PR.

PR #360, opened 2017-10-31
Fix rounding bug? Supposed to be solved in 1.9.3.1

seansan:
https://magento.stackexchange.com/questions/108570/1-9-2-catalogue-price-rules-rounding-when-more-then-one-rule-applied-with-stop

Might be that this bug was fixed elsewhere - or not? 

please cancel the pull if wrong (we are seeing this rounding bug in our stores now and on 1.9.3.x)


seansan:
+1 I have now tested this 1 week in production. And all "strange" cases with 1 cent difference have been solved

PR #361, opened 2017-11-01
Add tax_amount column to sales order grid

seansan:
We often add tax_amount column to sales order grid, so might be something to add

For now I have included it in the query (and using EAG enhanced grids you can choose to show it)

Or if people vote for it one can add the column by default using addColumn (now commented out)

If there is a better way to do it, or if there are maybe more columsn that we can add that make sense but do not affect performance too much - plz reply


colinmollenhour:
This should be custom functionality. Joining tables on the grid can slow it down significantly and there are many columns that *could* be added and unless it is one that nearly 100% of users want it should just be custom. However, I would be in total agreement with making it easier to add columns, such as new events events like 'adminhtml_sales_order_grid_prepare_collection_after'.

Issue #364, opened 2017-11-03
Incorrect stock versus sales 


googlygoo:
### Object
Open a discussion on the issue. What has been found and start a conversation how how best to resolve this issue while maintaining compatibility.

### Summary
During a sale where we had high concurrency and popularity of a small number of products we noticed in our final reports that we had oversold a few items. 

### Investigation
While doing some research (a few hours googling) I came across the below mentioned magento bug/forum post that had similarities to our situation. I was not happy with the solution but found it to be a great starting point to my investigation. 

The issue presents itself around the following event listeners in the Mage_CatalogInventory modue
```
<sales_model_service_quote_submit_before>
    <observers>
        <inventory>
            <class>cataloginventory/observer</class>
            <method>subtractQuoteInventory</method>
        </inventory>
    </observers>
</sales_model_service_quote_submit_before>
<sales_model_service_quote_submit_success>
    <observers>
        <inventory>
            <class>cataloginventory/observer</class>
            <method>reindexQuoteInventory</method>
        </inventory>
    </observers>
</sales_model_service_quote_submit_success>
```

`Mage_CatalogInventory_Model_Observer::subtractQuoteInventory()` function Mage calls `Mage_CatalogInventory_Model_Stock::registerProductsSale()` to reduce the stock and set the stock items to be used in `Mage_CatalogInventory_Model_Observer::reindexQuoteInventory()`.  

The simplified process is as follows
- A transaction is started, 
- The rows are locked in the cataloginventory_stock_item table for the affect items
- Stock level is correct, if so continue
- A stock item object is create (keep this in mind)
- The row is updated
- The transaction is committed releasing the lock
- Finally the array of stock item objects it returned and stored in the observer to be used later.

So far everything is good, we lock the table to ensure no other process is running while we check stock and update on sale.

Now the next step, `Mage_CatalogInventory_Model_Observer::reindexQuoteInventory(), skip past the re-indexing of the stock table as this uses the product data from the quote. We now loop through the list of previously fetched stock items, perform a save and add the items to the list of products to perform a price index for. This second save is where all the issues creep in. 

Let me try explain how errors can happen. I have tried to simplify it to better explain myself but if something is unclear please ask in the comments. In my examples below I will starting with a stock total of 5 and each order will be order 1 product. 

So in normal operation for a single product you will see db events as follows, 
**Order 1**: Select for update (row lock), Updated, Commit (Unlock) (Stock Before 5, After 4)
[...]

PR #373, opened 2017-11-14
Addresses Incorrect stock versus sales. #364

googlygoo:
Solution taken from @bob2021 #189

Looking to address this issue I have implemented two solutions. This solution and #372. Each solution has there merits. I prefer the solution #372 with one of the biggest reasons being it's simplicity and staying closer to what was originally being done in Magento. 

This solution moves the update logic to a Sql query, the benefit is that it all stock item records are updated with a single query, the down side is that that event that gets fired `cataloginventory_stock_item_save_after` may have different information to that that is in the database.


googlygoo:
Closing this pull request as through more testing I have found that there is more chance for things not working correctly by replacing the save() function with different logic. The better solution is to match the save function behaviour more closely.

PR #383, opened 2017-11-20
Use describeTable and then use prepareColumnValue to actually prepare…

pepijnblom:
… the value being used in the where section of select statements

Please see discussion on closed pull request here about casting incrementId to string I had with @Flyingmana @LeeSaferite :
https://github.com/OpenMage/magento-lts/pull/381

The discussion got me thinking and I decided to go ahead nip it in the bud instead of just solving it superficially. This will need a bit more testing but so far I have not experienced any downsides to it. I have yet to test it on a giant database though but I'll be sure to put it on our acceptance server and seeing how it fairs with 2 million records.

Anybody know how to send this in to Magento themselves if it proves to be stable?
I'm kind of surprised they didn't use the methods they already had available to them, they seem to only use prepareColumnValue before writes.


app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
```diff
@@ -400,6 +400,10 @@ public function load(Mage_Core_Model_Abstract $object, $value, $field = null)
      */
     protected function _getLoadSelect($field, $value, $object)
     {
+        $fields = $this->_getReadAdapter()->describeTable($this->getMainTable());
+        if(isset($fields[$field]) && isset($fields[$field]['DATA_TYPE']))   {
```
tmotyl: it's enough to just call `isset($fields[$field]['DATA_TYPE'])`. isset is NOT throwing notice/warning when accessing non existing keys.

pepijnblom: you are totally correct, i hadn't noticed this before

pepijnblom:
I've been running this on our acceptance server today during testing of a release. I placed about 30 orders and ran exports & imports using Xtento modules. The client was adding & modifying products during this time and not a single error was logged. So far so good, I'll be putting it live soon.

cieslix:
+1

app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
```diff
@@ -400,6 +400,10 @@ public function load(Mage_Core_Model_Abstract $object, $value, $field = null)
      */
     protected function _getLoadSelect($field, $value, $object)
     {
+        $fields = $this->_getReadAdapter()->describeTable($this->getMainTable());
+        if(isset($fields[$field]['DATA_TYPE']))   {
```
colinmollenhour: Nit-pick, but the check for `['DATA_TYPE']` is already performed within prepareColumnValue so is redundant to do here.

pepijnblom: You're very right, I hadn't actually checked prepareColumnValue too well but actually saw that this morning when I put it live. Unfortunately we're still seeing queries for increment_id without parentheses so my quest continues.

tmotyl: @pepijnblom a side question - how do you monitor such a queries? Is there a generic way to log all queries using wrong data types?

pepijnblom: @tmotyl Not that I know of. In this case I noticed it because the queries were stacking up during peak hours and they were showing up in the mysql slow log because they were not using indexes. It's one thing to see a Magento query with 10 joins showing up in there, but you know something's wrong when it's a single select * from table :)

I found the other location causing the problem. It's in:
app/code/core/Mage/Sales/Model/Resource/Quote.php

I'll make a pull request for that too.

tmotyl: @pepijnblom will you take care of the comment from @colinmollenhour  ?

pepijnblom:
Sorry about the extra commits, I had worked in the wrong branch.. Should definitely be squashed and the only extra that came with it now is the one from my other pull request which is already merged.

colinmollenhour:
Squashed and merged via command line due to conflicts. See c04788c

PR #388, opened 2017-11-25
Add quotes to numeric incrementIds - The Sequel

pepijnblom:
Here's the second location where queries were built without quotes around increment_id if it was numeric.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Resource/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Resource/Quote.php
@@ -163,9 +163,7 @@ class Mage_Sales_Model_Resource_Quote extends Mage_Sales_Model_Resource_Abstract
     public function isOrderIncrementIdUsed($orderIncrementId)
     {
         $adapter   = $this->_getReadAdapter();
-        $fields    = $adapter->describeTable($this->getMainTable());
-        $orderIncrementId = $adapter->prepareColumnValue($fields['increment_id'], $orderIncrementId);
-        $bind      = array(':increment_id' => $orderIncrementId);
+        $bind      = array(':increment_id' => (string)$orderIncrementId);
         $select    = $adapter->select();
         $select->from($this->getTable('sales/order'), 'entity_id')
             ->where('increment_id = :increment_id');
--- a/app/code/core/Mage/Sales/Model/Resource/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Resource/Quote.php
@@ -169,7 +169,8 @@ class Mage_Sales_Model_Resource_Quote extends Mage_Sales_Model_Resource_Abstract
         $select = $adapter->select()
             ->from($this->getTable('sales/order'), 'entity_id')
             ->where('increment_id = :increment_id');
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Sales/Model/Resource/Quote.php
```diff
@@ -156,17 +156,20 @@ public function getReservedOrderId($quote)
...
     public function isOrderIncrementIdUsed($orderIncrementId)
     {
-        $adapter   = $this->_getReadAdapter();
-        $bind      = array(':increment_id' => $orderIncrementId);
-        $select    = $adapter->select();
-        $select->from($this->getTable('sales/order'), 'entity_id')
+
+        $adapter = $this->_getReadAdapter();
+        $fields = $adapter->describeTable($this->getTable('sales/order'));
+        $orderIncrementId = $adapter->prepareColumnValue($fields['increment_id'], $orderIncrementId);
```
colinmollenhour: To make merging upstream changes easier commits should have minimal number of lines impacted. In this case this new line 168 could simply be added as the first or second line of the method and all of the other lines left unchanged.

pepijnblom:
@colinmollenhour Hi, sorry for getting back so late on this one. I've reverted the original code and only added the single line as per your request. It does make more sense to it that way considering possible future patches.

app/code/core/Mage/Sales/Model/Resource/Quote.php
```diff
@@ -156,12 +156,14 @@ public function getReservedOrderId($quote)
...
      *
-     * @param string $orderIncrementId
+     * @param string|integer $orderIncrementId
+     *
      * @return boolean
      */
     public function isOrderIncrementIdUsed($orderIncrementId)
     {
         $adapter   = $this->_getReadAdapter();
+        $orderIncrementId = $adapter->prepareColumnValue($fields['increment_id'], $orderIncrementId);
```
LeeSaferite: Where does `$fields` come from?

pepijnblom: Oh crap, sorry about that.. Forgot one line

app/code/core/Mage/Sales/Model/Resource/Quote.php
```diff
@@ -156,12 +156,15 @@ public function getReservedOrderId($quote)
...
-     * @param string $orderIncrementId
+     * @param string|integer $orderIncrementId
+     *
      * @return boolean
      */
     public function isOrderIncrementIdUsed($orderIncrementId)
     {
         $adapter   = $this->_getReadAdapter();
+        $fields    = $adapter->describeTable($this->getMainTable());
+        $orderIncrementId = $adapter->prepareColumnValue($fields['increment_id'], $orderIncrementId);
```
tmotyl: Do we need it at all here? Cant we just cast the value to string? We know which column w are binding (increment_id), so there is no need to detect the coumn type.

pepijnblom: There's something to say for that as well, indeed. I'll make the necessary change.

Issue #155, opened 2017-01-09
[Suggestion] Add Changlog?


sreichel:
Well ... why?

In my opinion the actual description does not point out what the advantages of Magento-LTS vs Vanilla Magento are. Uhhm ... OK ..."They accecpt bug fixes". Sounds not so bad, but it's not clear whats already in ...

There are 
- applied patches (even for older versions)
- bug fixes
- performance increasements
- even some new (small) features
- ...

Why dont tell that - to make this project more attractive?

Cause i'm lazy i wouldn't add a CHANGLOG.md ... i would add a link README.md that points to https://github.com/OpenMage/magento-lts/pulls?page=1&q=is%3Apr+is%3Aclosed+is%3Amerged&utf8=%E2%9C%93 and add some labels for all merged PR (bug, enhancement, patch ...)

colinmollenhour:
Definitely a good idea although I think CHANGELOG.md would be more user-friendly it would also be more annoying to maintain so I'm open to either proposed method.

Flyingmana:
I support this idea, as long as we dont require it for PRs.
Going from time to time trough contributions to add missing entries is not this hard.

The more problematic part is removing stuff which got fixed upstream by magento

sreichel:
Agreed, it shouln't be a requirement . I'd prefer the "easier to maintain" solution.  Add a first label for kind of this PR (bug/patch/...).

I also wouldn't remove entries that got fixed in later Magento version. It should be enough to add a second "branch-label" that match magentos versions, where the issue was fixed.

sreichel:
Don't know where to ask ...

... is there any "schedule" for merging approved PRs? I don't want to stress someone, but there are some PRs that should go live ASAP .... (i know i could use a/my fork, but i definitly want to rely on "official" LTS)

tmotyl:
+1 for making human readable changelog.
Unfortunately neither git history nor list of closed PR's is useful here.

Git history is hard to read because we use non linear history (merge commits) and commit descriptions/titles are also not too nice.

List of merged PR's is also not a solution (in the form it is right now) as you don't know which PR was applied on the current branch.

I think we should require every PR to add appropriate changelog entry.

Flyingmana:
`git log` is perfectly able to produce a linear representation even with merge commits.
Require changelog entries on PRs could(depending on format, config, activity and time to merge) lead to regular merge conflicts, especially if people want to backport them.

But I would like to see references to Issues in commit messages

[...]

PR #21, opened 2015-06-18
Zend_Validate_Hostname with latest gTLD

MrGekko:
Update Zend_Validate_Hostname to include the latest gTLD
Version 2015061701, Last Updated Thu Jun 18 07:07:01 2015 UTC

PR #387, opened 2017-11-22
Update - gTLD list for Zend Hostname Validator

spinsch:
It was already updated #21 in branch "OpenMage:1.9.1" but never migrated in 1.9.3.x branch.  
Now this is a full update from the current[ iana list](
https://data.iana.org/TLD/tlds-alpha-by-domain.txt).

```Version 2017112100, Last Updated Tue Nov 21 07:07:01 2017 UTC```

@Flyingmana .tp is now also removed from the list by iana ;)

PR #390, opened 2017-11-27
Update list of valid TLDs

colinmollenhour:
Converted list at http://data.iana.org/TLD/tlds-alpha-by-domain.txt to PHP array.

Note, I don't know anything about the UTF-8 characters domains at the end of the list. They are not on the IANA list but I left them intact.

As mentioned by @edannenberg on #155, the changes in #21 appear to have been clobbered some time ago and rather than restore the original commit I just grabbed the latest list.


spinsch:
Duplicate of #387

colinmollenhour:
Ahh, so it is...

PR #392, opened 2017-11-28
Fix Missing Mage_Api_Model_Server_V2_Handler in SOAP v2 WSDL.

SRaromicon:
Hey Guys,

this is a quick fix for the missing Handler in WSDL of SOAP API v2. Handler was always null, because the correct Handler was set after Model Instantiation.

### DIFF
--- a/app/code/core/Mage/Api/Model/Server/V2/Adapter/Soap.php
+++ b/app/code/core/Mage/Api/Model/Server/V2/Adapter/Soap.php
@@ -41,8 +41,6 @@ class Mage_Api_Model_Server_V2_Adapter_Soap extends Mage_Api_Model_Server_Adapte
     protected function _getWsdlConfig()
     {
         $wsdlConfig = Mage::getModel('api/wsdl_config');
-        $wsdlConfig->setHandler($this->getHandler())
-            ->init();
         return $wsdlConfig;
     }
 
@@ -57,6 +55,9 @@ class Mage_Api_Model_Server_V2_Adapter_Soap extends Mage_Api_Model_Server_Adapte
         $apiConfigCharset = Mage::getStoreConfig("api/config/charset");
 
         if ($this->getController()->getRequest()->getParam('wsdl') !== null) {
+            $this->wsdlConfig->setHandler($this->getHandler())
+                ->init();
+            
             $this->getController()->getResponse()
                 ->clearHeaders()
                 ->setHeader('Content-Type','text/xml; charset='.$apiConfigCharset)


tmotyl:
Hi @SRaromicon 
Thanks for the patch.
How can it be tested? What are the steps to reproduce the issue?

SRaromicon:
Sorry for my late response. I don't know, if this is really an issue, because it never worked correctly in Magento. 

If you do look at Mage_Api_Model_Wsdl_Config_Base there is a handler Variable, that should store the Handler Name for different API WSDL (Soap v1 and v2). The Variable is used in app/code/core/Mage/Api/etc/wsdl.xml, for example. The variable is always null, when WSDL is generated.

```
    <service name="{{var wsdl.name}}Service">
        <port name="{{var wsdl.handler}}Port" binding="typens:{{var wsdl.handler}}Binding">
            <soap:address location="{{var wsdl.url}}" />
        </port>
    </service>
```
You can test this very easy by getting the wsdl from https://www.yourdomain.de/api/v2_soap/index/?wsdl=1

Without Patch the WSDL will look like:

```
    <service name="MagentoService">
        <port name="Port" binding="typens:Binding">
            <soap:address location="https://www.yourdomain.de/index.php/api/v2_soap/index/" />
        </port>
    </service>
```

With Patch:

```
<service name="MagentoService">
    <port name="Mage_Api_Model_Server_V2_HandlerPort" binding="typens:Mage_Api_Model_Server_V2_HandlerBinding">
        <soap:address location="https://www.yourdomain.de/index.php/api/v2_soap/index/"/>
    </port>
</service>
```

tmotyl:
thanks @SRaromicon, can you also share some php snippet which fails without the patch and works with it? How did you find out this issue?
It would help a lot validating the solution.

edannenberg:
Usually api clients are generated against the wsdl, so I suspect this change could be breaking all existing api clients. As far as I can tell this fix is only cosmetic.

To test if this is a breaking change:

1. install SoapUI
2. create new service from wsdl without this fix being active
3. test api works in SoapUI
4. apply fix
5. test api still works in SoapUI without having to regenerate the client

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @SRaromicon  if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

fballiano:
I use SOAP v2 from WSDL since years without problems so maybe it was solved in 1.9.4.something?

ADDISON74:
If you did not replicate this issue we can take a look to the source code. Maybe the change was fixed in Magento 1.9.4.5 and it should be found in OM-19. In fact it is a small change as I see.

elidrissidev:
This issue is reproducible for me, but I don't know what difference the solution makes, if any?

fballiano:
you're right, I re-tested more carefully and the "handler" is empty, the PR works although I don't know what difference does it make.

fballiano (approved):

ADDISON74 (approved):

elidrissidev:
Did someone test @edannenberg's suggestion just in case?

elidrissidev:
I tested using SoapUI, requests are working fine after applying the fix and refreshing the cache.

PR #397, opened 2017-11-30
The productIds variable is already used as an array expression

lfluvisotto:


seansan:
good find +1

sreichel:
Someone using PhpStorm? :)

Btw ... got a free license for contributing to this project.

lfluvisotto:
PhpStorm and this one:

https://github.com/phpqa/phpqa.github.io

tmotyl:
@lfluvisotto have you investigated how this change influence the rendering, or what bug does it fix? 
From the reading the code its clear that the code is wrong, but would be very interesting what is the effect.

sreichel:
> From the reading the code its clear that the code is wrong

Seems not to be wrong.

	foreach ($arr as $arr) {
		var_dump($arr); // retruns correct array value
	}

So for my tests (PHP 5.5.9) changing the variable name has no effect.

lfluvisotto:
foreach ($productIds as $productIds) is not wrong.

There are no official recommendations that I am aware of, but I tend to use the foreach ($plural as $singular) pattern.

Just for good practice, more readable, avoiding confusion.

tmotyl:
@lfluvisotto I'm totally with you on the best practices side. I just want to understand the impact.
Have you investigated if fixing this bug has any side effects on how shop works?

lfluvisotto:
@tmotyl I do not consider this case as a bug but rather as a quality assurance improvement, just this.

Follow my test scenario:

I have 3 products ID: 1, 2, 3



Reindex Data on 'Catalog URL Rewrites' Index Process



Xdebug on branch 1.9.3.x









Xdebug on branch variable-is-already-used-as-an-array-expression









Same behavior.

LeeSaferite:
@tmotyl Looking at the code change in-context, it will break nothing and fix nothing. It *will* make the code a lot clearer though.

See:

https://3v4l.org/o13e4
`
<?php
$a = array(1,2,3,4);
foreach($a as $a) {
    var_dump($a);
}
var_dump($a);
`

This code shows that using the same variable name inside the loop is fine, but if you tried to use the source array after the loop you'd hit a bug. So, the code, as written, is a bug waiting to happen. This PR fixes that possible bug and improves readability.

LeeSaferite:
Just to be clear, I see zero reason not to merge this.

PR #398, opened 2017-11-30
Missing break statement

lfluvisotto:


LeeSaferite:
I'd like to know the same as @colinmollenhour. How can we see this bug in practice and how can we test this PR to verify operation?

lfluvisotto:
> Is this bug manifested in any way on a bare installation?

No, because none part of core code calls the outputValue method of class Mage_Eav_Model_Attribute_Data_Multiselect but this method could be called via custom module.

What I noticed was that the other classes all cases are with their due breaks.

app/code/core/Mage/Eav/Model/Attribute/Data/Multiline.php
app/code/core/Mage/Eav/Model/Attribute/Data/Select.php
app/code/core/Mage/Eav/Model/Attribute/Data/Date.php
app/code/core/Mage/Eav/Model/Attribute/Data/File.php

And according to PHP documentation:

http://php.net/manual/en/control-structures.switch.php

> PHP continues to execute the statements until the end of the switch block, or the first time it sees a break statement. If you don't write a break statement at the end of a case's statement list, PHP will go on executing the statements of the following case.

In app/code/core/Mage/Eav/Model/Attribute/Data/Multiselect.php

If $format is either Mage_Eav_Model_Attribute_Data::OUTPUT_FORMAT_JSON or Mage_Eav_Model_Attribute_Data::OUTPUT_FORMAT_ARRAY

This is also going to pass on default block.

lfluvisotto:
Hello @LeeSaferite @colinmollenhour  are you guys going to merge this PR?

Cheers!

sreichel:
> How can we see this bug in practice and how can we test this PR to verify operation?

If we dont care about breaking _bad_ written 3rd party code, this should be accepted.

I've not found any uage of this mehtod in practice ([maybe](http://www.divisionlab.com/solvingmagento/magento-eav-system/#mage-eav-model-form) cutomer cutomer attritubes?), but the current behavior is clearly unexpected.

If I set format to `OUTPUT_FORMAT_JSON` or `OUTPUT_FORMAT_ARRAY` i expect an array is returned, but currently you'll get always a string - so `$format` parameter is just ignored.

From just reading the code this is a bug to me, but I also undertand concerns to change such thing since it can break methods that rely on such (wrong) behavoir.

Flyingmana:
please add at least the PR number to the commit message when merging in this way

PR #401, opened 2017-11-30
The key and weight variable used as a key and weight foreach loop is already used in the same way by the outer foreach loop.

lfluvisotto:


sreichel:
Think this is not an "error", The outer variables are not used later (after next loop), so it shouldnt make a difference.

+1 for better readablity

lfluvisotto:
Yep, it not an error.

We know that each foreach loop has its own local variable independent of the variable having the same name inside (inner) or outside (outer).

Just for good practice, more readable, avoiding confusion.

lfluvisotto:
Hello @sreichel are you guys going to merge this PR?

Cheers!

sreichel:
@lfluvisotto Not able to.

PR #403, opened 2017-11-30
Exception has already been caught

lfluvisotto:


sreichel:
Maybe you can add this to you commit ...

There is exists no `Mage_Core_Model_Exception`. Guess we can remove first catch block?

https://github.com/OpenMage/magento-lts/blob/47cfea4525c4d0bba47601ce055dc56a78a3da63/app/code/core/Mage/Adminhtml/controllers/Catalog/ProductController.php#L941-L944

lfluvisotto:
@sreichel great!

PR #405, opened 2017-11-30
Case mismatch call (class / method / function)

lfluvisotto:


app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Theme.php
```diff
@@ -66,7 +66,7 @@ public function getOptions()
         }
         if (empty($options) || !is_array($options)) {
             $options = Mage::getModel('core/design_source_design')
-                ->setIsFullLabel(true)->getAllOptions(false);
+                ->setIsFulllabel(true)->getAllOptions(false);
```
LeeSaferite: Why is this case wrong?

LeeSaferite:
Looks like some good housekeeping work. Thanks and keep it up!

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Theme.php
tmotyl: will this still continue to work? 
if this property is magic, then changing the case will result in different property being set.

sreichel: @tmotyl is right. Seems to break functionality ...

app/code/core/Mage/Core/Model/Design/Source/Design.php

    /**
     * Setter
     * Add package name to label
     *
     * @param bool $isFullLabel
     * @return Mage_Core_Model_Design_Source_Design
     */
    public function setIsFulllabel($isFullLabel)
    {
        $this->_isFullLabel = $isFullLabel;
        return $this;
    }

lfluvisotto: Following the naming convention of getter and setter.

I think is better rename the function/method from setIsFulllabel to setIsFullLabel

LeeSaferite: @sreichel How do you think this breaks anything? It's simple correcting the case of the function name being called to the proper case. PHP doesn't enforce this so it currently works being mis-cased.

LeeSaferite:
I'm with @colinmollenhour on not changing the method name case. Yes, it's annoying they used weird casing on the method name, but it breaks that external API since the method is public. Just fix the calls to the method to match the actual method name case.

seansan:
​agree​

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Theme.php
tmotyl: After investigating this a little bit, this is NOT breaking.

tmotyl:
changing "setIsFulllabel" to "setIsFullLabel" is NOT breaking as this method is implemented as getter (in contrast to magic getters and setters, where changing the casing would be breaking).
PHP is case insensitive on method names.

So I'm in favor of fixing it.

However I saw few other changes like this in the patch. All of them should be checked whether they refer to implemented methods or magic getters. 
In later case, changing name would be breaking.

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Theme.php
sreichel: @LeeSaferite Wasn't aware the just variables are case sensitive, and methods are not ...

lfluvisotto:
If it is foggy, we can change the cases that there is no doubt such as:

MAge::

->rollback()

Pow()

StdClass

Other simple cases where there is not the ambiguity of calling the magic method or method.

tmotyl:
@lfluvisotto for me it's ok to change other occurrences too, if they are not breaking (they are not magic). We just need to be cautious.

LeeSaferite:
@tmotyl FWIW, I actually looked up and validated every case change prior to approving the initial PR.

tmotyl:
Thanks @LeeSaferite

LeeSaferite:
@lfluvisotto Are you planning on reverting the method name change?

lfluvisotto:
@LeeSaferite  Whatever you all think is the best I do here.

Issue #394, opened 2017-11-29
Session validation failure in Magento 1 EE v 1.14.3.x (and CE 1.9.3.x)


proxium:
Hi, we are seeing an increase amount of this exception:

    2017-11-22T08:30:27+00:00 ERR (3): 
    exception 'Mage_Core_Model_Session_Exception' in 
    /magento/src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php:421

I'm using Magento ver. 1.9.3.6 and Memcached also switched the session storage to Files, both session storage systems didn't avoid the exception.

There is a thread on StackOverflow that explains the exception and present a "workaround", you can check it here: https://magento.stackexchange.com/questions/167616/session-validation-failure-in-magento-1-ee-v-1-14-3-x-and-ce-1-9-3-x

Do you have an idea how to avoid rewriting the core files or how to avoid this error?

colinmollenhour:
I think this should be fixed by modifying core code and moving the validator data to it's own key and not duplicating it for every namespace which would also shrink the overall session storage considerably.

sreichel:
@colinmollenhour if you have time for, can you take a look at this? Have same issue since 1.9.3.x and no better solution as suggested in the linked SO question ...

colinmollenhour:
I just committed the better solution in b8c5f4c. Please give it some testing and review and I'll create a proper PR if it looks all good.

seansan:
HI, I am using I'm using Magento ver. 1.9.3.7 and Redis. Strangely enough we also see this error now. Was this introduced recenlty with other merges?  

I just cherrypicked this fix and testing

seansan:
Most errors are gone, new one introduced: 

Notice: Undefined index: http_x_forwarded_for  in           app/code/core/Mage/Core/Model/Session/Abstract/Varien.php on line 456

nginx does not set http_x_forwarded_for

rvelhote:
I'm getting the same PHP notice in `system.log` as @seansan which is weird because the key is indeed set in the `$sessionData`. Also using Nginx; don't know if it's related. I will continue investigating.

colinmollenhour:
This key should never not be set because it is initialized with an empty array and never unset:

```
        $parts = array(
            self::VALIDATOR_REMOTE_ADDR_KEY             => '',
            self::VALIDATOR_HTTP_VIA_KEY                => '',
            self::VALIDATOR_HTTP_X_FORVARDED_FOR_KEY    => '',
[...]

PR #406, opened 2017-12-04
Store session validator data once and validate only once. Refs #394

colinmollenhour:
The session code currently stores the session validator data within every namespace and also validates it every time the namespace is inited. This is bad because:

1. Extremely inefficient of session storage space. The validator data often comprises over 50% of the space used by a namespace and when there are many namespaces this adds up to a ton of waste. Session storage can be cut drastically with this patch and when using an in-memory storage like Redis or Memcached that matters a lot.
2. Inefficient of compute cycles since multiple namespaces means multiple validations and there is no good reason for these to differ from each other.
3. Actually creates bugs such as #394 where the validator data is updated on some requests but not others (so it can differ but shouldn't). I haven't tested but I believe this will also fix this issue.

I've been using a similar patch for over 2 years and would have submitted a PR long ago had there been an official public repo...

### DIFF
--- a/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
+++ b/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
@@ -35,6 +35,9 @@ class Mage_Core_Model_Session_Abstract_Varien extends Varien_Object
     const VALIDATOR_SESSION_EXPIRE_TIMESTAMP    = 'session_expire_timestamp';
     const SECURE_COOKIE_CHECK_KEY               = '_secure_cookie_check';
 
+    /** @var bool Flag true if session validator data has already been evaluated */
+    protected static $isValidated = FALSE;
+
     /**
      * Map of session enabled hosts
      * @example array('host.name' => true)
@@ -406,16 +409,21 @@ class Mage_Core_Model_Session_Abstract_Varien extends Varien_Object
     /**
      * Validate session
      *
-     * @param string $namespace
+     * @throws Mage_Core_Model_Session_Exception
      * @return Mage_Core_Model_Session_Abstract_Varien
      */
     public function validate()
     {
-        if (!isset($this->_data[self::VALIDATOR_KEY])) {
-            $this->_data[self::VALIDATOR_KEY] = $this->getValidatorData();
+        // Backwards compatibility with legacy sessions (validator data stored per-namespace)
+        if (isset($this->_data[self::VALIDATOR_KEY])) {
+            $_SESSION[self::VALIDATOR_KEY] = $this->_data[self::VALIDATOR_KEY];
+            unset($this->_data[self::VALIDATOR_KEY]);
+        }
+        if (!isset($_SESSION[self::VALIDATOR_KEY])) {
+            $_SESSION[self::VALIDATOR_KEY] = $this->getValidatorData();
         }
         else {
-            if (!$this->_validate()) {
+            if ( ! self::$isValidated && ! $this->_validate()) {
                 $this->getCookie()->delete(session_name());
                 // throw core session exception
                 throw new Mage_Core_Model_Session_Exception('');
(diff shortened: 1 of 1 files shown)


seansan:
​thx! and the legacy code bit: this would support legacy sessions from the
get-go of this codde, but this route would then slowly become extinct?​

proxium:
I applied in line 443

    $sessionData = $_SESSION[self::VALIDATOR_KEY]; 

But then I found that some session index are not found ex:

    # not implementing "isset($sessionData[self::VALIDATOR_HTTP_X_FORVARDED_FOR_KEY])" 
    # before assigning those value will generate this error
    # Notice: Undefined index: http_x_forwarded_for ...

    $sessionData[self::VALIDATOR_HTTP_X_FORVARDED_FOR_KEY];
    $sessionData[self::VALIDATOR_REMOTE_ADDR_KEY];

tmotyl:
this looks very promissing, as session storage size becomes an issue with bigger shops

colinmollenhour:
@proxium Can you provide some more details.. It's not clear to me how this PR can cause that to happen given that getValidatorData returns an array that will always have all of these keys set.

colinmollenhour:
I agree with @seansan's commit changes, but I don't see how it is specific to this PR.

sreichel:
Using this since some weeks and all session are are gone.

+1

BTL-UG:
> The session code currently stores the session validator data within every namespace and also validates it every time the namespace is inited. This is bad because:
> 
> 1. Extremely inefficient of session storage space. The validator data often comprises over 50% of the space used by a namespace and when there are many namespaces this adds up to a ton of waste. Session storage can be cut drastically with this patch and when using an in-memory storage like Redis or Memcached that matters a lot.
> 2. Inefficient of compute cycles since multiple namespaces means multiple validations and there is no good reason for these to differ from each other.
> 3. Actually creates bugs such as #394 where the validator data is updated on some requests but not others (so it can differ but shouldn't). I haven't tested but I believe this will also fix this issue.
> 
> I've been using a similar patch for over 2 years and would have submitted a PR long ago had there been an official public repo...

Not sure if here is a better place for my comment that I also have put in the respective commit.

Actually, " and there is no good reason for these to differ from each other." is not completely true since the validation data is differing, as described below:

This "fix" is bypassing the session validation via password timestamp. Since the Validator data is gathered only once (at core session validation and NOT at customer session validation) it will never have a value for:

```php
public function getValidatorData()
    {
        [...]

        if (isset($this->_data['visitor_data']['customer_id'])) {
            $parts[self::VALIDATOR_PASSWORD_CREATE_TIMESTAMP] =
                Mage::helper('customer')->getPasswordTimestamp($this->_data['visitor_data']['customer_id']);
        }

        return $parts;
    }
```

This means that in 

```php
protected function _validate()
    {
       [...]
        if ($this->useValidateSessionPasswordTimestamp()
            && isset($validatorData[self::VALIDATOR_PASSWORD_CREATE_TIMESTAMP])
            && isset($sessionData[self::VALIDATOR_SESSION_EXPIRE_TIMESTAMP])
            && $validatorData[self::VALIDATOR_PASSWORD_CREATE_TIMESTAMP]
            > $sessionData[self::VALIDATOR_SESSION_EXPIRE_TIMESTAMP] - $this->getCookie()->getLifetime()
        ) {
            return false;
        }
        return true;
    }
```

the condition

```php
&& isset($validatorData[self::VALIDATOR_PASSWORD_CREATE_TIMESTAMP])
```

will always resolve to false and therefore never check this

```php
&& $validatorData[self::VALIDATOR_PASSWORD_CREATE_TIMESTAMP]
            > $sessionData[self::VALIDATOR_SESSION_EXPIRE_TIMESTAMP] - $this->getCookie()->getLifetime()
```

Any thoughts on this? @colinmollenhour 

(By the way: Just coming across this issue because I traced down a bug in vanilla 1.9.4.0 where customer always get's logged out when editing account data. This is caused by the password timestamp validation, since in the editPostAction of the AccountController the password timestamp is mistakenly set to time() even if the password has not been changed. That means that if one would recover the password timestamp validation here on magento-lts it would be recommended to update the account controller to only set the password created timestamp to time() if the password has actually been changed.)

PR #407, opened 2017-12-04
misreference $array

seansan:


downloader/lib/Mage/HTTP/Client/Curl.php
```diff
@@ -514,7 +514,7 @@ protected function curlOption($name, $value)
      * Set curl options array directly
      * @param array $array
```
LeeSaferite: This needs to be fixed as well

tmotyl: or make all occurrences "$array"

LeeSaferite: Since the word `array` is a reserved word, I personally wouldn't use it as a variable name just to remove any ambiguity or possible programmer confusion. A better choice would be to change it to something like `$options` instead. That being said, I'm ok with the original change, I just wanted the docblock to match.

PR #409, opened 2017-12-06
Fix 1.9.3.x upstream merges

edannenberg:
The upstream merges for some 1.9.3.x releases introduced a few inconsistencies:

* cms_generate_breadcrumbs event is fired twice
* related to the above dd412309 is applied to wrong code block
* few locale .csv have now escaped single quote entries along with the entries where those got removed in #46
* #21 got reverted at some point
* var/package dir has 5 obsolete xml files
* slightly different configuration in Sales, SalesRule and Tax config.xml
* Mage_SalesRule_Model_Validator is outdated compared with latest from 1.9.3.7

The issues were found by extracting and applying the patch set on top of a vanilla Magento 1.9.3.7 commit, finally doing a directory compare with the current 1.9.3.x branch. The branch used for comparison can be found [here](https://github.com/edannenberg/magento-lts/commits/1.9.3.7)

#21 is not in this PR in favour of #387


seansan:
Good work!

​Maybe, we need to double check some of these reverted/re-applied changes
against PHP7 support ... esp the Sales and SalesRule XML config

https://github.com/OpenMage/magento-lts/pull/62​

As an example:
In a recent 193.x. install instance we were seeing the 1 cent rounding
error again ... that should have been long gone

edannenberg:
> Maybe, we need to double check some of these reverted/re-applied changes against PHP7 support ... esp the Sales and SalesRule XML config 

The Inchoo extension added a cleaner [fix](https://github.com/Inchoo/Inchoo_PHP7/commit/1c1c5e452eb77b71e578d2f66d2aa4ca59628fcb) for the .xml workaround, not sure about any BC issues with PHP5.6, else this would be a good addition for this repo as it fixes the issue at the root.

> In a recent 193.x. install instance we were seeing the 1 cent rounding error again ... that should have been long gone

Not sure this can ever be truly fixed unless an effort is made to finally use the intended precision for product prices in the db. All that `deltaRound` shenanigans could be avoided..

seansan:
(not on topic, but mentioned here:
https://github.com/Inchoo/Inchoo_PHP7/issues/50)

​Interesting is that I see this exact xml snippet removed in the Inchoo
update

https://github.com/Inchoo/Inchoo_PHP7/commit/1c1c5e452eb77b71e578d2f66d2aa4ca59628fcb#diff-a4345d83fc8c1f52d7fe717e321e2619L10

We recently added this again in local.xml because by *magic* we were seeing
recent 1 cent rounding differences again ... ​


Not sure how/why .... just sharing

seansan:
+1 ignore my rounding issues ... this was solved. And dont want to block other good things in this change

I will test again after merged

Issue #382, opened 2017-11-19
Configuration > Default Page - should not list disabled CMS pages


tomekjordan:
Hi. even disabled CMS pages are listed to choose for home page.
It should not list disabled CMS pages, and list current available per certain website/storeview or global scope
Screenshot: [http://prntscr.com/hceyjv](http://prntscr.com/hceyjv)

Thanks

colinmollenhour:
If you remove them from the list and it was already configured then you can unintentionally change the configuration..

1. Create CMS page.
2. Select CMS page in config.
3. Disable created CMS page.
4. Reload config page and click save.

Correct behavior: configuration is not changed since user did not intentionally make any changes.

Behavior with your proposal: some other CMS page will be selected since the previously selected page will not be in the list and configuration can thus be changed completely unintentionally.

My proposal would be to list the pages but denote them somehow as disabled. E.g. use two groups, Active and Inactive and group the pages accordingly, or add an "(inactive)" suffix to the inactive pages. Or just leave it as-is.

tmotyl:
I'm also in favor of marking disabled or unavailable pages visually.

sreichel:
How about an observer that disables deactivate default page?

sreichel:
I think this should be changed - somehow. At the moment you can also select pages that are not assigned to specfic storeview ....

Another idea ....

- prevent disable/delete CMS that is used in config

config.xml

    <cms_page_save_before>
        <observers>
            <prevent_disable_default_cms_page>
                <class>sr_debug/observer</class>
                <method>preventDiableDefaultCmsPage</method>
            </prevent_disable_default_cms_page>
        </observers>
    </cms_page_save_before>
    <cms_page_delete_before>
        <observers>
            <prevent_delete_default_cms_page>
                <class>sr_debug/observer</class>
                <method>preventDiableDefaultCmsPage</method>
            </prevent_delete_default_cms_page>
        </observers>
    </cms_page_delete_before>

Oberver.php

    public function preventDiableDefaultCmsPage(Varien_Event_Observer $observer)
    {
        /* @var Mage_Cms_Model_Page $page */
        $page = $observer->getObject();

        if ($page->isObjectNew()) {
            return;
        }

        $isDeleteAction = $observer->getEvent()->getName() === 'cms_page_delete_before';

        if (!$page->getIsActive() || $isDeleteAction) {
            /* @ref #415 */
[...]

PR #415, opened 2018-01-03
Fix "Store View" field value empty when reloading form after failed save.

colinmollenhour:
The form uses 'store_id' as the element id but 'stores[]' as the input name since the model after save uses `getStores` for the new store ids. Without this change the store_id is empty and so the Store View field values get wiped when the form is reloaded.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Cms/PageController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Cms/PageController.php
@@ -102,6 +102,7 @@ class Mage_Adminhtml_Cms_PageController extends Mage_Adminhtml_Controller_Action
         // 3. Set entered data if was error when we do save
         $data = Mage::getSingleton('adminhtml/session')->getFormData(true);
         if (! empty($data)) {
+            $data['store_id'] = $data['stores'];
             $model->setData($data);
         }
 


sreichel:
Fix works.

Are there any downsides with fixing the field name? Why is it `store_id` when deleting CMS page and `stores` when editing? Should it be consistent?

app/code/core/Mage/Adminhtml/controllers/Cms/PageController.php
```diff
@@ -102,6 +102,7 @@ public function editAction()
         // 3. Set entered data if was error when we do save
         $data = Mage::getSingleton('adminhtml/session')->getFormData(true);
         if (! empty($data)) {
+            $data['store_id'] = $data['stores'];
```
LeeSaferite: If the form is using ```stores[]``` as the element name then it's meant to be an array. Here you seem to just be grabbing the array (or possible array) and assigning it to a field that's meant to be a scalar, no?

LeeSaferite:
@sreichel They tend to use ```stores``` when the value is a multi-select for multiple store selections.

app/code/core/Mage/Adminhtml/controllers/Cms/PageController.php
sreichel: Um, `store_id` i not meant to be a scalar. Think these lines from #382 make it more clear ...

    # is delete or edit action?
    $isDeleteAction = $observer->getEvent()->getName() === 'cms_page_delete_before';
    $storeIds = $isDeleteAction ? $page->getStoreId() : $page->getStores();

LeeSaferite: I disagree.

https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Cms/Model/Resource/Page.php#L122

The underlying model resource clearly expects that ```store_id``` is a scalar value that is used as a fallback when ```stores``` is not present. The code example you listed is faulty.

LeeSaferite: And I'll follow up with this example that is contradictory to what I just said but which I consider a bug.

https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Cms/Model/Resource/Page.php#L184

```store_id``` is clearly singular and in every other context on other models is a scalar. They clearly expect it to be a scalar on save since they cast it to an array. The fact that after loading they assign the array of store IDs to ```store_id``` vs. ```stores``` is, IMHO, 100% a bug in the core code.

LeeSaferite: Further evidence that the core code is buggy regarding the treatment of ```store_id```: https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Cms/Model/Resource/Page.php#L203-L204

Here the object loading code clearly expects ```store_id``` to be a scalar, not an array.

sreichel: It's a fallback, but for magic method `$object->getStoreId()` `store_id` can be anything.

The "problem" is that input fields name is (intentionally) `stores[]`. This form data is used when editing/saving CMS pages. When delete a page `getData()` contains the DB fields, so it is `store_id`.

> The code example you listed is faulty. 
Why?

However, fix works :)

sreichel: > IMHO, 100% a bug in the core code.

Thats what i've meant. :)

LeeSaferite: The input name is ```stores``` because that is the data field that is expected when saving a page. The fact that the core afterLoad code is dumping the store IDs into ```store_id``` instead of ```stores``` is clearly a bug. There is no DB field for ```store_id``` of a page (directly). There is an additional link table that links pages to stores so a page has an array of store ids always.

sreichel: Isn't it posible to add an alias to `cms_page_store.store_id as stores`?

(But then you need a fallback to `store_id`, at least for 3rd party code.)

sreichel:
Think this fix is good as it is. Fixing `stores` vs. 'store_id' can be done later.

app/code/core/Mage/Adminhtml/controllers/Cms/PageController.php
LeeSaferite: I would say adding ```stores``` in the after load is a good first step in the overall issue.

sreichel: Adding `$object->setData('stores', $stores);` to `_afterLoad` solves only my problem (different getters for store ids), but it doesnt solve the problem @colinmollenhour describes.

To reproduce ...

- add an observer to `cms_page_save_before`-
- just throw an exeption to display an error message

Selected stores are still empty.

ADDISON74:
> Think this fix is good as it is. Fixing `stores` vs. `store_id` can be done later.

... and what happened to this post? Has that problem been solved or not?

PR #419, opened 2018-01-08
Removed deprecated method _setAttribteValue()

sreichel:
> deprecated after 1.5.1.0 - mistake in method name

... and mistake in fix :) Better remove then fix it?


LeeSaferite:
Magento really needs a documented deprecate then remove procedure.

Flyingmana:
please add at least the PR number to the commit message when merging in this way

sreichel:
OK.

PR #420, opened 2018-01-08
Fixes "wrong" array key type

sreichel:
Not really a bug, but still worth to fix?

PHP interally casts `float` values key to `int`, but I think it should be better done before ...

    $arr = array();
    $arr[1.4] = '1st';
    $arr[1.1] = '2nd';
    var_dump($arr);

    array (size=1)
      1 => string '2nd' (length=3)


colinmollenhour:
Wouldn't casting to string be more correct based on what the original author was probably intending? (although it would be more likely to change behavior)

sreichel:
Guess you are right. I used `int` to not break anything/keep current behavior.

Closed, till tested with `string`.

PR #425, opened 2018-01-11
Use FONT_HELVETICA intsead of LinLibertineFont in PDF, ref #86

sreichel:
Credits to @seansan, just merged #272 and #274 and removed commented lines from #86.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php
+++ b/app/code/core/Mage/Sales/Model/Order/Pdf/Abstract.php
@@ -790,7 +790,7 @@ abstract class Mage_Sales_Model_Order_Pdf_Abstract extends Varien_Object
      */
     protected function _setFontRegular($object, $size = 7)
     {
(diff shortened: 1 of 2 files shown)


tomekjordan:
does it support all international chars? because I saw som problems and many PDF invoice extensions use Dejavu Sans ...

sreichel:
Not sure about this. #86 is already merged, so this is just to complete/fix this PR.

sreichel:
Not sure about this. #86 is already merged, so this is just to complete/fix this PR.

Think this PR is OK since it fixes (at least) the files size problem.

According to https://magento.stackexchange.com/questions/166205/problem-no1-solved-of-invoice-but-how-to-solve-problem2/166244 magentos `Libertine ` doesnt support all chars too, so its another thing that should be fixed.

sreichel:
I think this could be merged, because it only fixes/completes #86.

If there are any problems with international chars (i guess there are), you can add a new issue.

seansan:
+1 can be merged

Font has always been a problem ;) So one could argue that if using in a very specific language and having formatting issues : that in this case one can manually (re)set the font  

For 99% HELVETICA should work fine

PR #427, opened 2018-01-12
 Added website name on order create page, fixes #358

### LINKED ISSUE #358: Create backorder - shows only store view name instead of website name
Hi. as a merchant many times I need to create backorder, but order making screen shows only store view (which is usually language name) and gives nothing to admin. Many times I needed to cancel and start again because I wasn't sure this is correct website.

Screen: [http://prntscr.com/h0fy3b](http://prntscr.com/h0fy3b)

BTW: it would be great to change website during order creation process (something similar is done in IWD Order Manager)

Thanks

sreichel:
Usefull for multistore with duplicate storeview names (e.g. languages)


LeeSaferite:
FWIW, I agree here. Build the string for the 'website name' should be left to the translation system. Pass in three params and use positional replacements like "%1$s" and "%2$s" in the translation source string.

sreichel:
For "positional replacements" ... are they required in **source** string?

This should work for custom tranlations:

    "One %s, two %s";Two %2$s, one %1$s"

LeeSaferite:
You are correct, they are not required in the source string. I'm just generally explicit. :) Feel free to not use positional markers.

sreichel:
Just to mention ... if we add/change translations this should be documentated at least!

May add a new `Mage_LTS.csv` with all added/changed strings?

colinmollenhour:
> May add a new Mage_LTS.csv with all added/changed strings?

That's a good idea except then the helpers used to load the translations would have to be updated everywhere.

sreichel:
Should be possible to do this w/o changing helpers.

We could add 2nd translation file (`Mage_Sales_LTS.csv`) to `app/code/core/Mage/Sales/etc/config.xml`

        <translate>
            <modules>
                <Mage_Sales>
                    <files>
                        <default>Mage_Sales.csv</default>
                    </files>
                </Mage_Sales>
            </modules>
        </translate>

colinmollenhour:
I think adding a separate .csv per module is a good solution.

seansan:
I have the feeling we are making this a little complex. Discussion on adding new CSV translation files and deviating from the core CSV's ... for solving which problem? 

I have reviewed the original pull request ... and dont see what could go wrong here? The improvement is fine. Adding websitename - storecode is fine, solves a "marginal" problem ... 

Is there an issue logged by someone for this? (I  mean is it a real problem for someone)

We moved to the symlink and subdirectory solution a long time ago. And use nginx variables to set the store code. Hence we can use some more alternative store view codes with more info then just EN NL or DE

image

seansan:
It is the adminhtml ... you could make showing the storename optional via config
And then it is fine too

This mainly helps admin/shipping operators that only work in the backend

tomekjordan:
Current solution is fine, works for me. I dont see any reason to mak translation in here.

sreichel:
> so there nothing to translate for me

Just think about different languages or RTL-supports.

Yes, seems unnecessary complex, but i still a aggree. If we change such things it should be in "the most correct" way. (w/o positional replacements :P)

LeeSaferite:
> w/o positional replacements :P

Well, "Example: %s - %s" and "Example: %1$s - %2$s" are functionally equivalent. If a translation desires to reorder the arguments they can use positional arguments in their translation string and it'll just work as expected.

LeeSaferite:
Oh, I also disagree with adding a new CSV file. There's no reason we can't/shouldn't just update the core translation file(s)

sreichel:
> Oh, I also disagree with adding a new CSV file. There's no reason we can't/shouldn't just update the core translation file(s)

There is a reason! Locale packs ...

I'm from germany and use https://github.com/riconeitzel/German_LocalePack_de_DE for example ...

With this change default translation is already brokenen .... if we change `Mage_Sales.cvs` I have also to adjust DE local pack to make it work with LTS. Otherwise I can keep it as it is and add my own `Mage_Sales_LTS.csv` to de_DE directory.

colinmollenhour:
Any final objections to accepting in the current state?

tomekjordan:
Too much discussion for such small task :) I use it everyday and it works as expected. Thanks

sreichel:
Please let me know how we should fix this ... i'll update PR asap. (Btw ... I still prefer own CSV, but it's on you :))

tomekjordan:
Hi. one thing need update, when you create order from Orders Page. its shows customer names list,
and where they Signed Up From. The problem is - it also shows Store View name instead of Website name in that column
Can somebody fix it? or add Website column? Store view doesn't give useful information when you like to create order - you can even create order in different website - where customer is not registered in.

Screenshot
[http://prntscr.com/l2xokx](url)

Thanks a lot

PR #429, opened 2018-01-13
Added $sortByPostion flag to getChildren(), refs #351

sreichel:
Backport of  MAGETWO-81456 (https://github.com/magento/magento2/commit/83a8ef53c246cf1e382639e6ebb1e3de59aae0eb)


seansan:
Visual code review ;) and read backport issue desccriptions and fixes. Looks OK.

app/code/core/Mage/Catalog/Model/Resource/Category/Flat.php
```diff
@@ -1307,21 +1307,23 @@ public function isInRootCategoryList($category)
...
      * @return array
      */
-    public function getChildren($category, $recursive = true, $isActive = true)
+    public function getChildren($category, $recursive = true, $isActive = true, $sortByPosition = false)
     {
-        $maintable = $this->getMainStoreTable($category->getStoreId());
         $select = $this->_getReadAdapter()->select()
-            ->from($maintable, 'entity_id')
-            ->where('path LIKE ?', "{$category->getPath()}/%")
-            ->order($maintable.".position ASC");
```
Flyingmana: I expect the `$maintable` part is there for a specific reason. Without having a more in detail look on how this behaves in production I cant review this. Would be worried of an `column name is ambiguous` error because of some joins

sreichel: Please see collins change blow ...

colinmollenhour:
> This sorted children stuff u can easier solve with an group concat function inside mysql.

I'd be interested in seeing a proposal for it. Note the limits of group concat sometimes need to be bumped to avoid truncated results:

```
SET SESSION group_concat_max_len = 4096;
```

@sreichel the new default is `$sortByPosition = false` but the original code did sort by position by default so this is a change effectively reverting #351. Why not make `$sortByPosition = true` the default?

PR #434, opened 2018-01-19
Fixes catalog pricerule with date attributes in conditions, closes #433

### LINKED ISSUE #433: Catalog price rule not working with date attributes
Can please someone test catalog price rules with date attributes?

Issue and possible fix here: https://magento.stackexchange.com/questions/210006/catalog-price-rules-with-date-as-conditions-cannot-possibly-work-due-to-strtotim

sreichel:
Fixes a bug where a string is compared to a timestamp when trying to apply prices rules that use `datetime` product attributes.

See: https://magento.stackexchange.com/questions/210006/catalog-price-rules-with-date-as-conditions-cannot-possibly-work-due-to-strtotim

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Rule/Condition/Product.php
+++ b/app/code/core/Mage/CatalogRule/Model/Rule/Condition/Product.php
@@ -115,7 +115,10 @@ class Mage_CatalogRule_Model_Rule_Condition_Product extends Mage_Rule_Model_Cond
     {
         $attribute = $object->getResource()->getAttribute($this->getAttribute());
         if ($attribute && $attribute->getBackendType() == 'datetime') {
-            $value = strtotime($value);
+            $this->setValue(strtotime($this->getValue()));
+            if (is_scalar($value)) {
+                $value = strtotime($value);
+            }
         }
         return $value;
     }


seansan:
Looks good. 
 
(Indirect) Confirmation also here: https://magento.stackexchange.com/questions/210006/catalog-price-rules-with-date-as-conditions-cannot-possibly-work-due-to-strtotim#comment292467_210037

app/code/core/Mage/CatalogRule/Model/Rule/Condition/Product.php
```diff
@@ -115,7 +115,10 @@ protected function _prepareDatetimeValue($value, $object)
     {
         $attribute = $object->getResource()->getAttribute($this->getAttribute());
         if ($attribute && $attribute->getBackendType() == 'datetime') {
-            $value = strtotime($value);
+            $this->setValue(strtotime($this->getValue()));
```
LeeSaferite: What is the purpose of this being done here exactly? You've added a side-effect that seems like a generally bad idea as it's happening in an unexpected location. Also, it seems a sanity check for a scalar value was added for the return value from this method, but this internal `get/set` does not include the same safeguard.

Reading through the related code, it seems like there are better locations for this behavior change or at the very lease a better method to expose the change.

- https://github.com/OpenMage/magento-lts/blob/1.9.3.x//app/code/core/Mage/Rule/Model/Condition/Abstract.php#L279

- https://github.com/OpenMage/magento-lts/blob/1.9.3.x//app/code/core/Mage/Rule/Model/Condition/Abstract.php#L304

The core validation code uses `getValueParsed` which seems like the semantically correct location to pick up the date/time string that's been converted to a timestamp. Additionally, looking over that abstract condition code, the value vs. parsed value code is a mess and likely deserves a little TCL.

sreichel: Tried both of them. `_prepareDatetimeValue()` is the only method where attributes backend type is already checked. You have no access to this in abstract class. 

This affects only catalog price rules for date product attributes, so this looks like the right place. :)

---

For the two lines mentioned above ... 

- added line is to fix wrong comparison. `getValue()` is the date string stored in DB (`yyyy-mm-dd`), that needs to be converted to timestamp. Since there is no other method called before where this can be converted (`getValue()` in abstract class does not work, beacause this is used to display a date in rule conditions), I've put it here.

- the check for scalar value is added because `strtotime(null);` returns `false` (for products w/o `special_date_from` set)

https://github.com/OpenMage/magento-lts/blob/7be0774fef2f7a34b0dc0852073d24d81c4dbd17/app/code/core/Mage/Rule/Model/Condition/Abstract.php#L610-L616

Without this it does not return `false` at L612, it woud return `timestamp <= false` and add pricerules for products without `special_date_from` set.

sreichel:
W'd like to see this in, but no idea to improve this PR (last changes from comments aren't in yet)

tmotyl:
hi
Insteaf of merging into this branch I propose you reset the branch to current 1.9.4.x and cherry-pick your change, as it got messy.

sreichel:
Ooops. Something went wrong. I'll reopen ... :/

PR #435, opened 2018-01-22
Fixed php error in file, usage of undefined variable

manuperezgo:
M1 team clearly did a typo with: $txnId


seansan:
How did this ever work before?

Issue #416, opened 2018-01-03
[WIP] Project cleanup?


sreichel:
I doubt there will be a M1.10 with bigger code changes, so why not to tidy up this project?

**Cleanup:**

- [ ] remove dead code
  - [x] unused libraries (some `Zend` as @colinmollenhour already suggested #374) *
    - [ ] `lib/Mage/?`
    - [x] `lib/PEAR` (#952)
    - [x] `lib/Zend/?`
  - [ ] unused modules
    - [x] `Mage_Compiler` (#534)
    - [x] `Mage_Connect` (#952)
    - [x] `Mage_GoogleBase` (deprecated after 1.5.1.0) (#1002)
    - [ ] `Mage_Usa` (_now the process of creating the label is on DHL side_)
    - [x] `Mage_XmlConnect` (disabled by default, extends `Enterprise_`) (#1040)
    - [x] `Phoenix_Moneybookers` (#1044)
  - [ ] unused methods
  - [ ] deprecated methods
  - [ ] deprecated classes (`Mysql4`)
  - [x] remove downloader (#952)
- [x] general fixes
  - [x] like session validtion (#406)
  - [x] unused/undefined method/variable (e.g. #436)
- [ ]  check if overriden classes are newer

**"Improvements _beyond bug fixes_"** (features)**:**

- [ ] improve layout system
  - [ ] Aoe_Layout (#292)
  - [x] N98_LayoutHelper (#1051, netz98/N98_LayoutHelper#12)
- [ ] improve config loading (system.xml)
  - [x] load order _child -> parent_ (#259)
  - [ ] add default/values to config
  - [ ] add field validation
- [ ] improve permission system
  - [ ] add check for all massactions (#437)
- [ ] improve modularization
  - [ ] _see comment @LeeSaferite_
- [ ] improve event system
  - [ ] add mor named events  (#249 needs update)
- [x] add translation CSV for changed/added tranlations (#358)
- [x] Stampede-Resistant Config Cache (#422)

**"full" IDE support / DOC block updates**

- [x] DOC block updates
  - [x] Admin (#778)
  - [ ] Adminhtml (no PR, big one)
  - [x] AdminNotification (#695)
  - [x] Api (#693)
  - [x] Api2 (#779)
  - [ ] Authorizenet (#1139)
  - [x] Backup (no PR)
  - [x] Bundle (#782)
  - [x] Captcha (#698)
  - [x] Catalog (#783)
  - [x] CatalogIndex (#781)
[...]

PR #436, opened 2018-01-22
Fix undefined var in Mage_Core_Model_Url_Rewrite::removeTag()

edannenberg:
`$k` is referenced but never defined.

refs #416


sreichel:
How did you found this?

PR #440, opened 2018-01-29
Fixed another bug where it was not possible to retrieve payment child transactions

manuperezgo:
As it can be seen everywhere around this class, the correct way to get the transaction type is:
->getTxnType() no ->getType()

Without this fix, its not possible to retrieve child transaction of a transaction f.e getting the child VOID transaction of an AUTH transaction


seansan:
+1 merge?

Issue #443, opened 2018-02-01
Catalog Price Rules End Date not working as expected


seansan:
Catalog Price Rules End Date not working as expected. M 1934, LTS version

Problem: When setting a Catalog Price Rule to start 1-12 and end 31-12. For this to work we need to set start date _instead_ as 30-11 and end date to 30-12. Backend adminhtml US language. European Time. Server time (now) = Thu Feb  1 13:38:11 CET 2018 and corresponds with the actual time.

Problem also confirmed here:
- https://magento.stackexchange.com/questions/170961/magento-catalog-price-rule-date-not-working-properly#
- https://magento.stackexchange.com/questions/171093/magento-issue-with-dates-for-catalog-price-rule
- (possible solution) https://magento.stackexchange.com/questions/61045/catalog-price-rule-from-today-does-not-apply

![image](https://user-images.githubusercontent.com/652395/35678805-fe2722c2-0754-11e8-8933-f6a0ec580ce7.png)

In file 'Mage/CatalogRule/Model/Resource/Rule.php' I dont know why they are adding 1 day?

![image](https://user-images.githubusercontent.com/652395/35679011-caabcf64-0755-11e8-887e-d027cba68598.png)

sreichel:
Can you please try patch from #434. Guess its the same issue.

seansan:
​not the same;

I am testing ​a solution now will update later




Powered by Gmelius <https://gmelius.com?ref=mail>

sreichel:
(@seansan removed my comment before your answer ... so also deleted it to keep it clean)

Just a reminder to me ... check timezone issues and customers last visit date - there was something wrong too.

tomekjordan:
I dont remeber, but I thought it was fixed in 1931. Im using it, and there are no catalog price rules porblems anymore. Im in UTC+2 timezone

tomekjordan:
UPDATE: I checked and I managed it by fixing 
catalogrule/observer::dailyCatalogUpdate

with AOE Scheduler to fit my timezone: 
from: Original: 0 1 * * *
to: 1 0 * * *
and it works for me (GMT+1)

seansan:
A patch was shared in pull. Should fix it for all time zones without making
other changes.
Please test that and confirm.

tomekjordan:
Tested. 100% works. Start date,end date for catalog rules know works properly (tested on GMT+1). Thanks

tomekjordan:
[...]

PR #444, opened 2018-02-01
Fix for #443, catalog pricerules start and end date not respected (co…

seansan:
…rrect)

Fix for #443, catalog pricerules start and end date not respected (correct)

Please validate my logic. 

1. Fromtime and totime are now measured in GMT
2. And compared to GMT time (now) in $timestamp

This way (if and) when the CRON runs the rules at that specific time are checked and activated (or disabled) - as this cron job runs per default at 2 AM as I understand. This would mean that all rules enabled and disable are in effect turned ON and OFF at 2AM during the night the rule is activated (1st day) and de-activated after the end date 2 hours after midnight. 

Reference 1: https://magento.stackexchange.com/questions/61045/catalog-price-rule-from-today-does-not-apply
Reference 2: (under catalog rules) http://jokiruiz.com/magento/magento-timezones-timestamps/

Test data:
- for a rule with start date 10 dec 2017 and end date 30 jan 2018
- we see the following comparison on Thu 1 feb 2018 @ 14:28 GMT

TIMESTAMP: Thu, 01 Feb 2018 14:28:03 GMT
START; Sat, 09 Dec 2017 23:00:00 GMT
END: Tue, 30 Jan 2018 22:59:59 GMT

So rule in this case is turned OFF

Please confirm

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Resource/Rule.php
+++ b/app/code/core/Mage/CatalogRule/Model/Resource/Rule.php
@@ -213,7 +213,7 @@ class Mage_CatalogRule_Model_Resource_Rule extends Mage_Rule_Model_Resource_Abst
         $toTime = (int) Mage::getModel('core/date')->gmtTimestamp(strtotime($rule->getToDate()));
         $toTime = $toTime ? ($toTime + self::SECONDS_IN_DAY - 1) : 0;
 
-        $timestamp = Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP);
+        $timestamp = time();
         if ($fromTime > $timestamp
             || ($toTime && $toTime < $timestamp)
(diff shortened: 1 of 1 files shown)


seansan:
+1 can someone review our logic? and/or test? Appreciate it ...

tomekjordan:
Tested. 100% works. Start date,end date for catalog rules know works properly (tested on GMT+1). Thanks

seansan:
Second chunk does give the same output for both​

$timestamp = Mage::app()->getLocale()->date(null, null, null,
true)->get(Zend_Date::TIMESTAMP);
time();​

(no rationale other than gut) going through Mage: app sounds better ....
anyways: it works! (finally because this has been a real culprit over the
years)







Powered by Gmelius <https://gmelius.com?ref=mail>

Issue #422, opened 2018-01-09
Stampede-Resistant Config Cache


seansan:
@colinmollenhour 
nice work
is this something to add to LTS? (Or was it added before?)

https://gist.github.com/colinmollenhour/7a91c4a92ccfd2adaeb6

LeeSaferite:
Perhaps we should add this to https://github.com/OpenMage/magento-lts/projects/3

colinmollenhour:
Yes, I think it should be added as it definitely fixes some significant stability issues, although others have since forked my gist and maintained the patch more carefully so those may be a better starting point than the original. E.g.: https://github.com/magehost/Cm_StampedeResistantConfig

seansan:
​good to hear; maybe give them a nudge? maybe they are willing to
contribute ... ​

sreichel:
> Perhaps we should add this to Project Cleanup

Yep. Same for #267 or #374

seansan:
@colinmollenhour Is this still something to add? Or have you/we changed our views on this?

PR #446, opened 2018-02-08
Trying diff on magehost/Cm_StampedeResistantConfig

seansan:
Trying diff on magehost/Cm_StampedeResistantConfig fixes https://github.com/OpenMage/magento-lts/issues/422

At least we can test it like this 

I see it also changes some license data and extra spaces etc ... maybe someone who knows best can suggest an additional pull to set it to the correct values?

I think in general we should think about: 
1. is the described problem real? and thus something we should fic
2. is this a good solution? and should we thus implement it


seansan:
​Would it be smarter if you created the pull? ... I ust copied and pasted
to show a diff

​




Powered by Gmelius <https://gmelius.com?ref=mail>

colinmollenhour:
Thanks @seansan - Created PR #545

PR #447, opened 2018-02-08
Couponcode adminhtml updated pull

seansan:
Couponcode adminhtml updated pull
- added string
- added %s
- added to translation files
no squash merge here otherwise would have bundled in 1


pocallaghan:
Whilst I haven't investigated, how the values pass through the system, both `$_order->getCouponCode()` and  `$_order->getDiscountDescription()` should likely be escaped before passing them to translate. Off the top of my head, I don't think a frontend user can insert an arbitrary value, so it's probable that you'd require admin access to exploit, but it's always good practice to escape such data. It's possible that with access to a limited administrative account, privileges could be escalated through use of persistent XSS in these fields.

seansan:
If it doesn’t break anything why not add the change? Please edit the change

Also
In some advanced cases more than 1 code can be applied. So maybe check for
multiple values too for these exotic cases

colinmollenhour:
Good observation, @pocallaghan. PR updated. Thanks!

tbaden:
@Flyingmana

PR #450, opened 2018-02-09
PHP 7.2 fixes (method signature)

jtnw:
PHP 7.2 requires method signatures to match when a class overrides a method from the parent. This PR fixes the getAllOptions method in classes that extend Mage_Eav_Model_Entity_Attribute_Source_Table.

In each instance, the child method used no parameters, thus including them will not break anything. There may be other places where this is a problem, and I will fix as I come across them.

### DIFF
--- a/app/code/core/Mage/Customer/Model/Customer/Attribute/Source/Group.php
+++ b/app/code/core/Mage/Customer/Model/Customer/Attribute/Source/Group.php
@@ -33,7 +33,7 @@
  */
 class Mage_Customer_Model_Customer_Attribute_Source_Group extends Mage_Eav_Model_Entity_Attribute_Source_Table
 {
-    public function getAllOptions()
+    public function getAllOptions($withEmpty = true, $defaultValues = false)
     {
         if (!$this->_options) {
             $this->_options = Mage::getResourceModel('customer/group_collection')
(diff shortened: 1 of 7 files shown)


colinmollenhour:
Since the new arguments have no effect perhaps they should be renamed or documented to make it clear that using them will have no effect?

jtnw:
Added phpdoc info. I think it's probably cleaner than naming the arguments $doNotUse1 and $doNotUse2 unless you had a better idea?

tmotyl:
I don't care that much about argument names, as modern IDEs has visual indication that argument is not used. So having doc comment is good enough for me.

Flyingmana:
I prefer to keep the argument names the same, in case other extended classes want to re-add functionality there

jtnw:
Before this is merged, should we change the `$withEmpty` default option to `false`? This is what the code looks like in `Mage_Eav_Model_Entity_Attribute_Source_Table`. Whether we implement this code in the children methods is another question, as it isn't likely to be used anywhere.

`
        if ($withEmpty) {
            array_unshift($options, array('label' => '', 'value' => ''));
        }
`

jtnw:
So for example, the method in file `app/code/core/Mage/Customer/Model/Customer/Attribute/Source/Group.php` would look like this:

```
public function getAllOptions($withEmpty = true, $defaultValues = false)
{
    if (!$this->_options) {
        $this->_options = Mage::getResourceModel('customer/group_collection')
                        ->setRealGroupsFilter()
                        ->load()
                        ->toOptionArray()
                        ;
    }
    if ($withEmpty) {
        array_unshift($options, array('label' => '', 'value' => ''));
    }
    return $this->_options;
}
```

jtnw:
So, after looking at all this, I'd recommend keeping the PR as it is, and if the community decides to implement the functionality of `$withEmpty` and `$defaultValues` for whatever reason, then we will need to dig into `app/code/core/Mage/Adminhtml/Block/Widget/Form.php` and maybe other files as well. But I don't really see a need to implement the extra functionality as it isn't used anywhere.

PR #451, opened 2018-02-12
[PHP-7] fix memory leak

altdovydas:


colinmollenhour:
So PHP doesn't already destroy the Varien_Image instance that is assigned to the local variable, or does this just force the garbage collection to happen sooner?

altdovydas:
@colinmollenhour This is tricky, because if you test memory usage with:
`memory_get_usage(true);`
You will not see any difference, because not PHP are using RAM.
If you destroy Varien_Image it destroy -> Varien_Image_Adapter_Gd2 which destroy:
`@imagedestroy($this->_imageHandler);`
Which free RAM instant (destroy image handler resource).
I notice this, after changed PHP from 5.6 to 7 because import is always killed (because exceeded of RAM). Garbage collector are initiating after process reach memory_limit, but in this case PHP don't count image handler as PHP process, but Linux will kill process if it reach limit.
Example:
Importing 10 000 products with images (each has ~3 img)
After importing ~500 products memory_get_usage(true);  ~270MB RAM
Linux top (php process) ~ 2GB RAM

seansan:
Add comment line and merge?

Flyingmana:
reminds me of https://github.com/OpenMage/magento-lts/pull/61
is this somehow related?

Flyingmana:
also, if PHP is not executing the destructor correctly without the unset, this sounds like a bug in PHP itself which is worth to get reported. (or its just a wrong setting for garbage collection)

colinmollenhour:
No idea how to go about addressing this via PHP source code, but I'd consider it one of those edge cases where even with a scripting language you may occasionally need to help the memory manager out a bit. The workaround is completely safe from being broken in the future so I see no reason not to accept the PR. I added a comment in the source for better context/readability.

simbus82:
Is it related to this problem from 1.9.2.0?
https://4rtis.com/fix-magento-running-out-of-memory-while-resizing-images

I need to add this hack to every Magento "vanilla" (1.9.4.0 too) otherwise my memory consuption become very high!

colinmollenhour:
Looks like it might be related, but it does not fix the issue referred to in the blog post with the shutdown function being registered. We'd be happy to accept a PR for that fix! Thanks!

Issue #455, opened 2018-02-24
Add To Cart Button in Category has faulty Redirect Url


SRaromicon:
Hey Guys,

since SUPEE-8788 or Magento CE Version > 1.9.2.4, the redirect urls at Cart Buttons are faulty.
& is replaced with \&amp; and query part of the url is broken. 

Steps to reproduce:

1. Set Option "After Adding a Product Redirect to Shopping Car" (System->Checkout->Shopping Cart) to "No"
2. Go to Category with Layered Navigation enabled and select at least 2 Filter
3. Hit "Add To Cart"-Button directly in Category
4. Look at your Url in Browser. You will be redirected to Category URL with broken Query Part with replaced & -> \&amp;
e.g. https://www.example.com/testcategory.html?color=rot&amp;manufacturer=nike

This issue can be reproduces in many active Magento Stores.

Caused by:

This ist caused by Mage_Core_Helper_Url:

```php    
public function getCurrentUrl()
    {
        $request = Mage::app()->getRequest();
        $port = $request->getServer('SERVER_PORT');
        if ($port) {
            $defaultPorts = array(
                Mage_Core_Controller_Request_Http::DEFAULT_HTTP_PORT,
                Mage_Core_Controller_Request_Http::DEFAULT_HTTPS_PORT
            );
            $port = (in_array($port, $defaultPorts)) ? '' : ':' . $port;
        }
        $url = $request->getScheme() . '://' . $request->getHttpHost() . $port . $request->getServer('REQUEST_URI');
        return $this->escapeUrl($url);
//        return $this->_getUrl('*/*/*', array('_current' => true, '_use_rewrite' => true));
    }
```

Magento has replaced 

`return $url`
by
`return $this->escapeUrl($url);`

Which uses a htmlspecialchars. 

I think the currentUrl should only be escaped in Templates, so you won't break the query Part.

seansan:
My concern is: why did they replace it? Was it to fix something else?

SRaromicon:
Sorry don't know, what they were trying to fix, but it was the wrong place and breaking the redirect urls is a big issue.

In my opinion escapeUrl should be used, if your are writing the url into a template or html output.

colinmollenhour:
Probably a case of overzealousness.. Agree that escaping should happen outside of the get method.

SRaromicon:
[...]

PR #456, opened 2018-02-24
Fix wrong escaped currentUrl.

SRaromicon:
Pull Request for Issue #455.

### DIFF
--- a/app/code/core/Mage/Core/Helper/Abstract.php
+++ b/app/code/core/Mage/Core/Helper/Abstract.php
@@ -277,7 +277,7 @@ abstract class Mage_Core_Helper_Abstract
      */
     public function escapeUrl($data)
     {
-        return htmlspecialchars($data);
+        return str_replace(array('<','>','"','\''), array('&lt;','&gt;','&quot;','&#039;'), $data);
     }
 
     /**
--- a/app/code/core/Mage/Core/Helper/Url.php
+++ b/app/code/core/Mage/Core/Helper/Url.php
@@ -51,7 +51,7 @@ class Mage_Core_Helper_Url extends Mage_Core_Helper_Abstract
             $port = (in_array($port, $defaultPorts)) ? '' : ':' . $port;
         }
         $url = $request->getScheme() . '://' . $request->getHttpHost() . $port . $request->getServer('REQUEST_URI');
-        return $url;
+        return $this->escapeUrl($url);
 //        return $this->_getUrl('*/*/*', array('_current' => true, '_use_rewrite' => true));
     }
 
--- a/app/code/core/Mage/Core/Helper/Url.php
+++ b/app/code/core/Mage/Core/Helper/Url.php
@@ -51,7 +51,7 @@ class Mage_Core_Helper_Url extends Mage_Core_Helper_Abstract
             $port = (in_array($port, $defaultPorts)) ? '' : ':' . $port;
         }
         $url = $request->getScheme() . '://' . $request->getHttpHost() . $port . $request->getServer('REQUEST_URI');
-        return $this->escapeUrl($url);
+        return $url;
 //        return $this->_getUrl('*/*/*', array('_current' => true, '_use_rewrite' => true));
     }
 


colinmollenhour:
I noticed that this is used in templates such as `app/design/frontend/default/iphone/template/page/switch/stores.phtml`:

```php
<a href="<?php echo $_group->getCurrentUrl() ?>"><?php echo $this->escapeHtml($_group->getName()) ?></a>
```

This should then be escaped like:

```php
<a href="<?php echo $this->jsQuoteEscape($this->$_group->getCurrentUrl()) ?>"><?php echo $this->escapeHtml($_group->getName()) ?></a>
```

Otherwise ostensibly an XSS could be a carried out like:

[https://example.com/index.php/admin/catalog_product/index/foo/" onclick="alert('booyah')](https://google.com)

SRaromicon:
Let me check this again. In my opinion they are using a different function add those templates, not that one from Mage_Core_Helper_Url. getCurrentUrl from Mage_Core_Helper_Url is always used as base64encoded string as uenc parameter in Magento Core.

colinmollenhour:
Ahh ok I didn't check to see if they were one and the same. I think in general it would not hurt to add `jsQuoteEscape` to PHP templates for href attribute values. Perhaps getCurrentUrl should do some filtering for quotes and angle brackets in the parameter keys and values (if it doesn't already)?

tmotyl:
Please note that this change (adding escaping) was introduced in Magento security release.

pocallaghan:
Having a quick look at this, the `Mage_Checkout_MultishippingController::registerAction` method seems to assign this return value to the `customer_form_register` block. An example template that I believe may be used by that block is `app/design/frontend/rwd/default/template/persistent/customer/form/register.phtml`, which renders `<input type="hidden" name="error_url" value="<?php echo $this->getErrorUrl() ?>" />`, this PR would therefore be unsafe. 

NB: If you're looking to amend the pull request to fix the value on output, rather than generation (which is arguably valid, though hard to not leave 3rd party modules vulnerable with that approach), I didn't look any further after seeing this, there may be other occurances.

SRaromicon:
You are right, but leaving the catalog redirect urls broken, isn't a really good opportunity. There only 20 occurrence of getCurrentUrl(). Most of them are not from Mage_Core_Helper_Url. 

Any suggestions to fix this issue?

pocallaghan:
I'm not entirely familiar with the issue, I just had some free time so looked through the PR's and commented where I saw potenial issues. The "correct" solution as far as core Magento is concerned is probably the change you made, but it would also require additional updates for anywhere calling the function, in order to ensure it's escaped in a contextually aware manner.

If the decision is to try and make the function secure for 3rd party extensions that might call it, things get a lot trickier because you don't know in what context they're using the URL return from the method (output in body text, as a js variable, as an attribute on a dom node, for a redirect etc). The best middle ground that springs to mind is to consider implementing something more akin to `Mage_Core_Model_Url:escape`, which escapes`<>'"`, essentially not escaping the `&` characters that have meaning in your context. This character in itself is also the least likely to cause security issues in any of the context I'm aware of.

SRaromicon:
So maybe its better to replace 

```php
return $this->escapeUrl($url)` 
```
by
```php
return str_replace(array('<','>','"','\''), array('&lt;','&gt;','&quot;','&#039;'), $url);
```
Or we can alter the Mage_Core_Helper_Abstract::escapeUrl function from

```php
public function escapeUrl($data)
{
     return htmlspecialchars($data);
}
```

to 

```php
public function escapeUrl($data)
{
   return str_replace(array('<','>','"','\''), array('&lt;','&gt;','&quot;','&#039;'), $data);
}
```

I've found a 3rd Party Extension, that is using the escapeUrl function for a redirect. That url is broken   too, if there are parameters like ?utm_source=domain.com&utm_campaign=price-list.

colinmollenhour:
I think that's a very reasonable solution. It solves possible accidental and double-encoding of '&' and gets rid of the dangerous characters that can affect the HTML parser which aren't likely to be used legitimately in urls anyway.

colinmollenhour (dismissed):

tbaden (dismissed):
thanks for the work 👍 
the code itself looks good.

maybe some optimization using strtr:
https://micro-optimization.com/strtr-vs-str_replace.html

tbaden:
@Flyingmana

Flyingmana:
Iam voting against a merge for now, this needs a more in detail security check.

It is for example used in \Mage_Widget_Model_Widget::getWidgetDeclaration
where it is indirectly part of processing a **POST** parameter.

Additionally this method has changed since then to also escape Script Identifierts. So it is likely an entry point, which was already used in past attacks.

Therefore I would even go so far to say, we should not touch it at all, especially not to remove safeguards.

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @SRaromicon  if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

colinmollenhour:
This code has since been updated with what appears to be a fairly robust javascript filtration so I think this can just be closed.

That said, it seems silly to go through such great lengths if you know you are escaping a url, just reject (return empty string) anything that doesn't match `/^https?:/` in addition to escaping quotes and brackets, right?

PR #461, opened 2018-02-28
Remove empty JS file

bastienlm:
Method use for add swatches JS on product page
(call: PRODUCT_TYPE_configurable Handle, rwd/default/layout/configurableswatches.xml)
But if the condition is false, this method return an empty string AND include empty js.
http://storage3.static.itmages.com/i/18/0228/h_1519839970_6082466_fdd17a4d7d.png

### DIFF
--- a/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
+++ b/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
@@ -118,7 +118,7 @@ class Mage_ConfigurableSwatches_Helper_Data extends Mage_Core_Helper_Abstract
     /**
      * Get swatches product javascript
      *
-     * @return string
+     * @return string | null
      */
     public function getSwatchesProductJs()
     {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -135,6 +135,5 @@ public function getSwatchesProductJs()
                 }
             }
         }
-        return '';
```
jacquesbh: A correct way is to `return null;`. But this works too.

Also this change is correct, returning `null` (by avoiding the `return` statement) will avoid to deal with an empty javascript filename. Which is better.

The impact is in `\Mage_Page_Block_Html_Head::getCssJsHtml`;

:+1:

bastienlm: Really better of course ! I've updated my PR. Thanks !

PR #465, opened 2018-03-09
Fix backend creationdate bug

seansan:
We dont want system time in backend. 

https://magento.stackexchange.com/questions/38143/customer-created-logged-in-timestamps-in-admin

Discovered whilst testing for GDPR: when customer asks "when" did I subscribe our data is always 1 hour off

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
@@ -78,8 +78,12 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_View
      */
     public function getCreateDate()
     {
-        return $this->_getCoreHelper()->formatDate($this->getCustomer()->getCreatedAt(),
-            Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true);
+        $date = Mage::app()->getLocale()->storeDate(
+        $this->getCustomer()->getStoreId(),
+        $this->getCustomer()->getCreatedAtTimestamp(),
+        true
+    );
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
```diff
@@ -78,8 +78,12 @@ public function getCustomerLog()
      */
     public function getCreateDate()
     {
-        return $this->_getCoreHelper()->formatDate($this->getCustomer()->getCreatedAt(),
-            Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true);
+        $date = Mage::app()->getLocale()->storeDate(
```
LeeSaferite: Indentation level is off on this whole block.

LeeSaferite: Generally speaking I disagree with this PR since it's not fixing a bug but instead changing behavior to what you expect. This method is meant to return date as it's stored in the DB (which should be UTC I would hope). If you notice, the very next method is `getStoreCreateDate()` and it does exactly what you want. If you think admin should be displaying the store date then instead why not edit the template to use the correct method? Even better would be a new bool config path and then check that in the template to have both behaviors. You could leave the default as the old behavior and use the flag to change to using store times instead.

seansan:
Agree. Config flag little too much. But yes.

seansan:
Looks like the template already accounts for difference for timezone vs time

But in our case the 2 are the same which is strange because it should not be (there is 1 hour offset)

image

seansan:
See https://github.com/OpenMage/magento-lts/issues/466

PR #469, opened 2018-03-13
Apply the Page Layout handle set in config.xml under global/page/layouts

jacquesbh:
In case the Page Layout is changed for a product, the handle is not applied as it is for a CMS Page.

Since this was not working before this PR, merging is a no risk operation.

Example: in `config.xml`
```xml
<?xml version="1.0" encoding="utf-8"?>
<config>
    <modules>
        <Example_Tmp>
            <version>0.1.0</version>
        </Example_Tmp>
    </modules>
    <global>
        <page>
            <layouts>
                <my_new_page_layout module="example_tmp">
                    <label>My New Page Layout</label>
                    <template>page/my_new_page_layout.phtml</template>
                    <layout_handle>my_new_page_layout</layout_handle>
                </my_new_page_layout>
            </layouts>
        </page>
    </global>
    <frontend>
        <layout>
            <updates>
                <example_tmp module="Example_Tmp">
                    <file>example_tmp.xml</file>
                </example_tmp>
            </updates>
        </layout>
    </frontend>
</config>
```

Then in your `example_tmp.xml` in layout directory:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<layout version="0.1.0">
    <my_new_page_layout>
        <label>My New Page Layout</label>
        <reference name="root">
            <action method="setTemplate"><template>page/my_new_page_layout.phtml</template></action>
            <action method="setIsHandle"><applied>1</applied></action>
        </reference>
    </my_new_page_layout>
</layout>
```

You can now add in your own layouts something like that as example:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<layout version="0.1.0">
    <my_new_page_layout>
        <remove name="content"/>
    </my_new_page_layout>
</layout>
```

This update in case the Layout is used for Products is not working in the current Magento version.
But It's already working for CMS Pages.

I fix this bug in local for many years now and I think it's time to share it.

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Product/View.php
+++ b/app/code/core/Mage/Catalog/Helper/Product/View.php
@@ -58,6 +58,11 @@ class Mage_Catalog_Helper_Product_View extends Mage_Core_Helper_Abstract
 
         $update->addHandle('PRODUCT_TYPE_' . $product->getTypeId());
         $update->addHandle('PRODUCT_' . $product->getId());
+
+        if ($settings->getPageLayout()) {
+            $update->addHandle($settings->getPageLayout());
+        }
+
         $controller->loadLayoutUpdates();
 
         // Apply custom layout update once layout is loaded


app/code/core/Mage/Catalog/Helper/Product/View.php
```diff
@@ -58,6 +58,11 @@ public function initProductLayout($product, $controller)
 
         $update->addHandle('PRODUCT_TYPE_' . $product->getTypeId());
         $update->addHandle('PRODUCT_' . $product->getId());
+
+        if ($settings->getPageLayout()) {
+            $update->addHandle($settings->getPageLayout());
```
LeeSaferite: I think you should actually be using this, no?
   
``$controller->getLayout()->helper('page/layout')->applyHandle($settings->getPageLayout());

LeeSaferite:
Geez. Looking at the core code around this I'm surprised *any* of the custom theme and custom page layout stuff actually works. Looking over the CMS code it looks like the customer theme from/to info is unused. Looking at categories it looks like the page layout handle is never applied. Instead it just triggers the template change for the page layout. It sounds like a minor difference, but it's meaningful. Same with the product code. It's not applying the page layout handle *and* it's not running the ```applyTemplate``` which is why you noticed the bug. Calling ```applyTemplate``` masks the bug of not calling ```applyHandle```.

As far as I can tell, Products and Catalogs use the `custom_from_date` and `custom_to_date` to control application of custom design, page layout, and custom update XML. This means a custom page layout is tied to a custom design and custom XML. You either apply all or none. That means you cannot have a permanent page layout selection for a product and a timed custom theme or custom XML. Also, the product review page doesn't properly use `catalog/design` to apply the from/to dates for the layout changes.

app/code/core/Mage/Catalog/Helper/Product/View.php
LeeSaferite: Actually, looking at the core config XML around `global/page/layouts` I see that the layout code and the layout handle *don't* match for the core page layouts. They use a `page_` prefix on all of the page layout handles. This is even more of a reason to apply the page layout handle via the helper.

LeeSaferite:
An just in case it was unclear, I highly welcome this PR. Layout related code is near to my heart. :)

app/code/core/Mage/Catalog/Helper/Product/View.php
jacquesbh: 👍 I'll change that.

LeeSaferite: Also, the `if` surrounding it isn't strictly needed. The `page/layout::applyHandle` code will be a no-op if the page layout code is invalid (null and '' are invalid).

LeeSaferite:
@jacquesbh I would suggest maybe also changing the example in the description so that the layout handle doesn't match the page layout code so there is no ambiguity.

Issue #472, opened 2018-03-22
Cancel payment return to cart: uer has to re-enter address


seansan:
Problem: Cancel payment return to cart: user has to re-enter address

I am being told there is an old error here: 

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Checkout/Block/Onepage/Billing.php#L117

Mage_Checkout_Block_Onepage_Billing

If the customer is logged in then the current line is fine (supposedly)
If the customer is NOT logged in then it should be

`$this->getQuote()->getBillingAddress()` or make it `$this->_address = $this->getQuote()->getBillingAddress();`

The same exists here: Mage_Checkout_Block_Onepage_Shipping

```
public function getAddress()
{
    if (is_null($this->_address)) {
        $this->_address = $this->getQuote()->getShippingAddress();
    }

    return $this->_address;
}
```

ADDISON74:
If this is still an issue we should create a PR for it. The author has a very limited activity on GitHub since 2021

PR #478, opened 2018-03-30
Fix issue #472 /// empty billing and ship details after cancel payment @ PSP

seansan:
Fixes issue: https://github.com/OpenMage/magento-lts/issues/472

From what I hear this is a long standing bug/unexpected behavior

If a user cancels his order at the PSP, they are redirected to the cart, if they then proceed back to the checkout: guess what? all billing information is empty and they have to re-enter all

so whatever the final fix: we should consider the customer experience. If you cancel your payment at the PSP and return to cart and continue: you expect your data to still be there

### DIFF
--- a/app/code/core/Mage/Checkout/Block/Onepage/Billing.php
+++ b/app/code/core/Mage/Checkout/Block/Onepage/Billing.php
@@ -102,20 +102,7 @@ class Mage_Checkout_Block_Onepage_Billing extends Mage_Checkout_Block_Onepage_Ab
     public function getAddress()
     {
         if (is_null($this->_address)) {
-            if ($this->isCustomerLoggedIn()) {
-                $this->_address = $this->getQuote()->getBillingAddress();
-                if (!$this->_address->getFirstname()) {
-                    $this->_address->setFirstname($this->getQuote()->getCustomer()->getFirstname());
-                }
-                if (!$this->_address->getMiddlename()) {
-                    $this->_address->setMiddlename($this->getQuote()->getCustomer()->getMiddlename());
-                }
-                if (!$this->_address->getLastname()) {
-                    $this->_address->setLastname($this->getQuote()->getCustomer()->getLastname());
-                }
-            } else {
-                $this->_address = Mage::getModel('sales/quote_address');
-            }
+            $this->_address = $this->getQuote()->getBillingAddress();
         }
 
         return $this->_address;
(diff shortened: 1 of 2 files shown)


seansan:
+1 review needed ofcourse

app/code/core/Mage/Checkout/Block/Onepage/Billing.php
```diff
@@ -102,20 +102,7 @@ public function getMethod()
     public function getAddress()
     {
         if (is_null($this->_address)) {
-            if ($this->isCustomerLoggedIn()) {
-                $this->_address = $this->getQuote()->getBillingAddress();
-                if (!$this->_address->getFirstname()) {
```
markvds: I would not remove these lines. They've been added somewhere in the past years and serve their purpose: it prefills the name for an address. There's nothing wrong with that.

seansan: So the suggestion is to 
1. move `$this->_address = $this->getQuote()->getBillingAddress();` above `if ($this->isCustomerLoggedIn()) {`
2. and remove the else?

colinmollenhour: Yes, I think that would make this PR complete.

seansan:
Feel free to update code.

colinmollenhour:
Thanks, but in general I prefer the original authors to update their own PRs as most project maintainers do because otherwise the expectation that a maintainer will fix things up results in poor quality PRs.

ADDISON74:
... and the original author did not update his own PR after 3 years. Good job.

A lot of PR's are waiting for feedbacks from their authors too.

ADDISON74:
What do we do with this PR? As far as I understand it would be OK if a change is made in the source code that the author did not want to operate for almost 4 years.

fballiano:
if you check the https://github.com/OpenMage/magento-lts/blob/a640d7b0af2adc2ec01119d7abe1055ac8079797/app/code/core/Mage/Checkout/Block/Onepage/Shipping.php#L69 you can see it's already been modified as suggested in this PR.

For the first part of the PR (regarding Billing.php) I don't understand the patch and the suggestions...

ADDISON74:
The author had activity on GitHub in 2020, since then his presence has decreased considerably. If we don't understand what this PR solves I don't see why it would remain open after so long.

fballiano:
let's close and and it could always be reopened if new info comes up :-)

PR #483, opened 2018-04-19
Fix issue with CRON where storeid is not known/not set

seansan:
Fix issue with CRON where storeid is not known/not set

https://github.com/AOEpeople/Aoe_Scheduler/issues/297#issuecomment-382660307

Not sure where this can have other unforeseen effect

### DIFF
--- a/app/code/core/Mage/Core/Model/App.php
+++ b/app/code/core/Mage/Core/Model/App.php
@@ -839,7 +839,7 @@ class Mage_Core_Model_App
         }
 
         if (!isset($id) || ''===$id || $id === true) {
-            $id = $this->_currentStore;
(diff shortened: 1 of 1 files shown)


colinmollenhour:
I don't think the default store should be assumed in the core code. If that behavior is desired then it can be set in the code that bootstraps Magento.

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @seansan if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

fballiano:
Matching my opinion and Colin's I think it's better to close this PR, assuming a store code in the core is a bit dangerous, at least with the information we have at this time.

PR #495, opened 2018-07-17
Remove diff trash from 40720cab87d614a57b1396800936fa86c6dc098b

edannenberg:
Leftover file from some patch attempt.

There should also have been a merge conflict when merging 1.9.3.9 as #192 is fixed in core now, although slightly less optimal. Dropped the fix in my fork to match the 1.9.3.x branch.


Flyingmana:
for reference, this file was added as part of https://github.com/OpenMage/magento-lts/pull/343

PR #496, opened 2018-07-26
Fixed ability to save config by a user with limited access

vovayatsyuk:
Related to configs that use `config_path` parameter - Paypal config section for example.

### Steps to reproduce

 1. Create new Admin Role at _System > Permission > Roles_ with access to "Payment Methods Section" and "PayPal Section" configurations.
 2. Create a new Admin User with this Role at _System > Permission > Users_
 3. Login by this user and try to save config.
 4. "Access denied" message will be shown

p.s. Solution is taken from https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Adminhtml/controllers/System/ConfigController.php#L280-L288


tbaden:
@Flyingmana

PR #504, opened 2018-08-23
Fix to allow for always reorder

seansan:
Fix to allow for always reorder

Works on backend. Tested on frontend. Maybe not all cases. 

Rationale
- if you want to reorder Magento should at least try
- if there is no stock: then we see this in admin order entry; can remove product or change: why block even trying?
- customerid check seems to work even with anonymous checkout

In short. Reorder is something you "want" so if you click it you are performing an action where the user "wants" to reorder
Currently there are all these safeguards in the code protecting against what? Magento should at least "try" to reorder. And the reorder screen or cart is a perfect place to do corrections (out of stock etc) if necessary: in short => there is a flow to highlight and correct potentials low stock or missing info anyways. Dont understand why this is so heavily blocked. 
Another option would be to offer a new config setting under sales->reorder but I think the current setting is already clear enough: we want the ability to reorder!

image

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -775,7 +775,7 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
      */
     public function canReorder()
     {
-        return $this->_canReorder(false);
+        return $this->_canReorder(true);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


sreichel:
However ... this should be merged into `1.9.4.x` ...

fballiano (approved):

PR #509, opened 2018-08-24
Bots bypass logs

rafaelpatro:
Improved bot list that could be ignored by logs saving two sql queries per bot access.
[Read more...](https://magento.stackexchange.com/questions/102326/how-to-prevent-bots-from-appearing-in-online-customers-in-backend)
https://github.com/OpenMage/magento-lts/blob/afc58e89714d5ec3c7a24b4af571bc3d4d66f0e3/app/code/core/Mage/Log/Model/Visitor.php#L198

_p.s. sorry about prototype.js... fixed!_

### DIFF
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -43,6 +43,605 @@
             <google1>Googlebot/1.0 (googlebot@googlebot.com http://googlebot.com/)</google1>
             <google2>Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)</google2>
             <google3>Googlebot/2.1 (+http://www.googlebot.com/bot.html)</google3>
+            <google4>Googlebot/2.1 (+http://www.google.com/bot.html)</google4>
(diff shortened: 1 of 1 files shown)


js/prototype/prototype.js
```diff
@@ -609,7 +609,7 @@ Object.extend(String.prototype, (function() {
   }
 
   function evalScripts() {
-    return this.extractScripts().map(function(script) { return eval(script) });
+	return this.extractScripts().map(function(script) { with(this){ return eval(script);} });
```
tmotyl: this change doesn't seem related to bot thing

rafaelpatro:
I didn't make this test but _500 strings in_array_ should be fast than _saving logs in database_ I guess.
But it would be awesome a regexp.

edannenberg:
Relevant code is in `Mage_Log_Model_Visitor::_construct()`:

```
        $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
        if ($ignoreAgents) {
            $ignoreAgents = $ignoreAgents->asArray();
            if (in_array($userAgent, $ignoreAgents)) {
                $this->_skipRequestLogging = true;
            }
        }
```

Integrating [Crawler-Detect](https://github.com/JayBizzle/Crawler-Detect) could also be an option, probably better to create an extension for that though.

Issue #508, opened 2018-08-23
Unable To Remove Coupon From Cart When Shipping/Tax Is Estimated


SafrazSSears:
In the Shopping Cart, when you use the estimate shipping/tax feature, apply the estimations to the order, add a coupon code to the order, and then remove the coupon, the Shopping Cart returns the "Coupon code was canceled." messages, however, the coupon remains attached to the quote/sale.
If you do not use the estimate shipping/tax feature, the coupon can be added and then removed without issue.

**LAMP Stack**
Linux: Ubuntu 16.04.10 (AWS Ubuntu 16.04 AMI)
Apache: Apache/2.4.18 (Ubuntu)
MySQL: Ver 15.1 Distrib 10.1.35-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
PHP: PHP 7.0.30-0ubuntu0.16.04.1 (cli)

**Magento CE Version:** 1.9.3.9

**Modules Active In Local Pool:**
1. Inchoo_PHP7

**Modules Active In Community Pool:**
1. Cm_RedisSession

**Active Theme:** base/default

Is anyone else experiencing this issue? I am experiencing this after upgrading from Magento CE 1.9.2.4 to Magento CE 1.9.3.9.
The only variable under that consideration is that I am not running a clean Magento CE 1.9.3.9 database.

rafaelpatro:
Yes, I can reproduce the issue in my store too.

Similar environment...
 - Linux: Ubuntu 16.04.3 (Hetzner dedicated)
 - Apache: Apache/2.4.27 (Ubuntu)
 - MySQL: Ver 14.14 Distrib 5.7.20, for Linux (x86_64) using  EditLine wrapper
 - PHP: PHP 7.0.28-0ubuntu0.16.04.1 (fpm-fcgi)

I'm using the same modules...
 - Inchoo_PHP7
 - Cm_RedisSession

Theme: rwd/default

SafrazSSears:
Has anyone tried with a similar environment, but using PHP 5 instead? This could be a PHP 7 compatibility issue.

SafrazSSears:
This seems to be an old issue that has existed across a few minor versions, and I dug up a solution in Magento's Issue Tracking.

The issue lies in the app/code/core/Mage/Checkout/controllers/CartController.php file, within the couponPostAction() function.
Follow these steps to fix the issue -
1. Find this line:
[...]

PR #512, opened 2018-08-30
Fix Coupon Removal Failure When Shipping/Tax Is Applied To Quote

SafrazSSears:
This pull request is based on Issue #508 

In the Shopping Cart, when you use the estimate shipping/tax feature, apply the estimations to the order, add a coupon code to the order, and then remove the coupon, the Shopping Cart returns the "Coupon code was canceled." messages, however, the coupon remains attached to the quote/sale.
If you do not use the estimate shipping/tax feature, the coupon can be added and then removed without issue.

LAMP Stack
Linux: Ubuntu 16.04.10 (AWS Ubuntu 16.04 AMI)
Apache: Apache/2.4.18 (Ubuntu)
MySQL: Ver 15.1 Distrib 10.1.35-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
PHP: PHP 7.0.30-0ubuntu0.16.04.1 (cli)

Magento CE Version: 1.9.3.9

Modules Active In Local Pool:

Inchoo_PHP7
Modules Active In Community Pool:

Cm_RedisSession
Active Theme: base/default

### DIFF
--- a/app/code/core/Mage/Checkout/controllers/CartController.php
+++ b/app/code/core/Mage/Checkout/controllers/CartController.php
@@ -620,6 +620,7 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
                 }
             } else {
                 $this->_getSession()->addSuccess($this->__('Coupon code was canceled.'));
+                $this->_getSession()->setCartCouponCode($couponCode);
             }
 
         } catch (Mage_Core_Exception $e) {


app/code/core/Mage/Checkout/controllers/CartController.php
```diff
@@ -620,6 +620,7 @@ public function couponPostAction()
                 }
             } else {
                 $this->_getSession()->addSuccess($this->__('Coupon code was canceled.'));
+                $this->_getSession()->setCartCouponCode($couponCode);
```
Flyingmana: I would suggest to do it above the message, to if something goes wrong with setting, it happens before we send out the message.

besides that I think it would make more sende, to directly set en empty string, or even Null(does Null work there?)
The empty string, because if the else is triggered, this means $codeLength did resolve to false

sreichel:
Closed with #656

PR #514, opened 2018-09-05
Template bugs

kiatng:
Bug 1. In PHP7, `$this->getFieldValue('qty')*1` will produce a warning _A non-numeric value encountered_ when qty is 0. 

Bug 2. The first key of the street array is 0 and not 1, resulting in missing street[0] and erroneous street address in the register form. Note that this only affects re-rendering of the register form when there was an exception in an earlier form submit.

### DIFF
--- a/app/design/frontend/base/default/template/persistent/customer/form/register.phtml
+++ b/app/design/frontend/base/default/template/persistent/customer/form/register.phtml
@@ -99,14 +99,14 @@
                 <li class="wide">
                     <label for="street_1" class="required"><em>*</em><?php echo $this->__('Street Address') ?></label>
                     <div class="input-box">
-                        <input type="text" name="street[]" value="<?php echo $this->escapeHtml($this->getFormData()->getStreet(1)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__('Street Address')) ?>" id="street_1" class="input-text <?php echo $_streetValidationClass ?>" />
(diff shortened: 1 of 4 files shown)


app/design/frontend/base/default/template/persistent/customer/form/register.phtml
```diff
@@ -99,14 +99,14 @@
...
                     </div>
                 </li>
             <?php $_streetValidationClass = trim(str_replace('required-entry', '', $_streetValidationClass)); ?>
             <?php for ($_i = 2, $_n = $this->helper('customer/address')->getStreetLines(); $_i <= $_n; $_i++): ?>
                 <li class="wide">
                     <div class="input-box">
-                        <input type="text" name="street[]" value="<?php echo $this->escapeHtml($this->getFormData()->getStreet($_i)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__('Street Address %s', $_i)) ?>" id="street_<?php echo $_i ?>" class="input-text <?php echo $_streetValidationClass ?>" />
+                        <input type="text" name="street[]" value="<?php echo $this->escapeHtml($this->getFormData()->getStreet($_i-1)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__('Street Address %s', $_i)) ?>" id="street_<?php echo $_i ?>" class="input-text <?php echo $_streetValidationClass ?>" />
```
Flyingmana: you missed to also change the title part

Flyingmana:
I would appreciate if you could split this up in two separate PullRequests.

Bug 1 is pretty obvious for me

But for Bug2 I would need to look a bit more into the code, so this will take a bit longer to review. And it would be sad, if we need to delay the Fix for Bug1 because of this.

app/design/frontend/base/default/template/persistent/customer/form/register.phtml
kiatng: Hmm...yes I did.

Issue #149, opened 2017-01-04
PHP7 .htaccess


ghost:
the .htaccess file should also include directives for mod_php7.c

    <IfModule mod_php7.c>

colinmollenhour:
I would prefer to just remove `<IfModule mod_php5.c>` instead of duplicate. If PHP isn't present then you've got major problems and probably don't want your webserver starting anyway and AFAIK all mod_php5 directives are compatible with all mod_php7 directives.

jtnw:
Removing the `<IfModule mod_php5.c>` may cause issues with apache+fastcgi. I don't use fastcgi myself so I can't say for certain, but from what I've read using php_value in .htaccess without mod_php causes a 500 error. See http://stackoverflow.com/questions/4117572/using-htaccess-with-fastcgi#4118492

colinmollenhour:
Ehh, what a mess. Here is a possible hack:

```
<IfModule mod_php5.c>
// Set doPhp = 1
</IfModule>
<IfModule mod_php7.c>
// Set doPhp = 1
</IfModule>
<If doPhp>
// php flags here
</If>
```

I don't know the exact syntax to use offhand though, perhaps SetEnv.

jtnw:
Yup. I came across this earlier, although it's only supported with apache 2.4. Can we drop support for older versions of apache? 
    <IfModule mod_php5.c>
        Define PHP_IS_ENABLED
    </IfModule>
    
    <IfModule mod_php7.c>
        Define PHP_IS_ENABLED
    </IfModule>
    
    <IfDefine PHP_IS_ENABLED>
    </IfDefine>
Otherwise I'm just inclined to add a comment to the htaccess noting about PHP7, but I did have a few hours of wondering why my PHP's max memory wasn't being read before I realized the php5 bit. Fixing this would definitely help out people down the line.

seansan:
@jtnw Any thoughts on this? Or has this been solved?

jtnw:
@seansan for me, I just changed php5 to php7 in htaccess, haven’t touched it in the past year.

Apache 2.4 came out in 2012, so the Define directive should be widely supported. I would recommend we use the structure in my previous comment.

seansan:
Can you submit a simple pull request? This would then close the issue

(great! how time and waiiting does sometimes fix common problems)

jtnw:
Sure thing, let me just do a quick test to make sure that the directive works.

Flyingmana:
[...]

PR #517, opened 2018-09-05
added php7 htaccess settings

jtnw:
Since there seems to be no way to have shared settings for both mod_php5 and mod_php7, the PR simply duplicates the directives in .htaccess for both. While not ideal, it will save headaches from people who run on php7 and haven't manually edited .htaccess.


jtnw:
Solves #149

PR #519, opened 2018-09-06
Revert changes in getOrderItem in Mage_Bundle

### LINKED ISSUE #513: getOrderItem don't always return OrderItem
https://github.com/OpenMage/magento-lts/blob/34afee089b4f4c80bf0340fb0c6c9b761b04d48b/app/code/core/Mage/Bundle/Block/Adminhtml/Sales/Order/Items/Renderer.php#L180

If you check this code, I think the logic is not correct.
In the `getValueHtml()` we have the line : 
`$result .= " " . $this->getOrderItem()->getOrder()->formatPrice($attributes['price']);`

And with this call, I have this error : 
```
PHP Fatal error:  Call to a member function formatPrice() on null in .../app/code/core/Mage/Bundle/Block/Adminhtml/Sales/Order/Items/Renderer.php on line 203
```

It seems logic because the `getOrderItem` does not always return an `OrderItem`

maximehuran:
fixes #513

### DIFF
--- a/app/code/core/Mage/Bundle/Block/Adminhtml/Sales/Order/Items/Renderer.php
+++ b/app/code/core/Mage/Bundle/Block/Adminhtml/Sales/Order/Items/Renderer.php
@@ -177,17 +177,14 @@ class Mage_Bundle_Block_Adminhtml_Sales_Order_Items_Renderer extends Mage_Adminh
         return $result;
     }
 
+    /**
+     * Retrieve Order Item
+     *
+     * @return Mage_Sales_Model_Order_Item
+     */
     public function getOrderItem()
     {
-        if ($this->getItem() instanceof Mage_Sales_Model_Order_Invoice_Item) {
-            return $this->getItem();
-        } else if ($this->getItem() instanceof Mage_Sales_Model_Order_Shipment_Item) {
-            return $this->getItem();
-        } else if ($this->getItem() instanceof Mage_Sales_Model_Order_Creditmemo_Item) {
-            return $this->getItem();
-        } else {
-            return $this->getItem()->getOrderItem();
-        }
+        return $this->getItem()->getOrderItem();
     }
 
     public function getValueHtml($item)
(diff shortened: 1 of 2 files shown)


sreichel:
This was changed because `Mage_Sales_Order_Item` isnt a valid class. I've tested this with different bundled products from our shop and got no errors.

I've no time to teset until the weekend ... instead of revert it, can you please try to change

```
return $this->getItem();
```

to

```
return $this->getItem()->getOrderItem();
```

Thanks.

maximehuran:
Did you try in forcing `isChildCalculated` to false ? 
Because in my case, this is the value of that.

I don't understand the change you want, can you provide a more information ?
You want the change on my code or on the initial ?

maximehuran:
I have the same problem in another project. I will check another time what we can do with this issue

maximehuran:
@sreichel what do you think about my new pull request ?

sreichel:
@maximehuran thanks for this. LGTM. Only annotation seems to be wrong. (should be `Mage_Sales_Model_Order_Item`?)

maximehuran:
@sreichel updated 🎉

sreichel:
Sorry for introducing this with https://github.com/OpenMage/magento-lts/commit/9312e9f6693b9d17fe2927222a7079a1f38c5606 ... none of my product matchedthe criteria. :(

sreichel:
@Flyingmana 

I tired to fix non-existing classs - `if ($this->getItem() instanceof Mage_Sales_Order_Item)`, but did not change the next lines from

    return $this->getItem();

to

    return $this->getItem()->getOrderItem();

It's correct to remove the orignal wrong check against a non existing class.

rafaelpatro:
Hi guys! Will it be reverted?
After applying commit 9312e9f6693b9d17fe2927222a7079a1f38c5606 I'm facing the same issue in my bundle products.
```log
PHP Fatal error:  Uncaught Error: Call to a member function formatPrice() on null in /var/www/app/code/core/Mage/Bundle/Block/Adminhtml/Sales/Order/Items/Renderer.php:203

PR #520, opened 2018-09-06
Fix wrong street index in customer register form

kiatng:
Fix wrong street index in customer register form, which uses the` Varien_Object` form_data in `getStreet($line)`, $line is zero-based. See `Mage_Customer_Block_Form_Register`:

    /**
     * Retrieve form data
     *
     * @return Varien_Object
     */
    public function getFormData()
    {
        $data = $this->getData('form_data');
        if (is_null($data)) {
            $formData = Mage::getSingleton('customer/session')->getCustomerFormData(true);
            $data = new Varien_Object();
            if ($formData) {
                $data->addData($formData);
                $data->setCustomerData(1);
            }
            if (isset($data['region_id'])) {
                $data['region_id'] = (int)$data['region_id'];
            }
            $this->setData('form_data', $data);
        }
        return $data;
    }

In other address edit forms, `getStreet($line)` is of `Mage_Customer_Model_Address_Abstract` which correctly returns `$arr[$line-1]`:

    /**
     * get address street
     *
     * @param   int $line address line index
     * @return  string
     */
    public function getStreet($line=0)
    {
        $street = parent::getData('street');
        if (-1 === $line) {
            return $street;
        } else {
            $arr = is_array($street) ? $street : explode("\n", $street);
            if (0 === $line || $line === null) {
                return $arr;
            } elseif (isset($arr[$line-1])) {
                return $arr[$line-1];
            } else {
                return '';
            }
        }
    }

But in the register forms, the street index need to be adjusted. Note that the object form_data is only populated when the controller throws an exception, see `Mage_Customer_AccountController::createPostAction()`

       } catch (Mage_Core_Exception $e) {
            $session->setCustomerFormData($this->getRequest()->getPost());
            if ($e->getCode() === Mage_Customer_Model_Customer::EXCEPTION_EMAIL_EXISTS) {
                $url = $this->_getUrl('customer/account/forgotpassword');
                $message = $this->__('There is already an account with this email address. If you are sure that it is your email address, <a href="%s">click here</a> to get your password and access your account.', $url);
            } else {
                $message = $this->_escapeHtml($e->getMessage());
            }
            $session->addError($message);
        } catch (Exception $e) {
            $session->setCustomerFormData($this->getRequest()->getPost());
            $session->addException($e, $this->__('Cannot save the customer.'));
        }

### DIFF
--- a/app/design/frontend/base/default/template/customer/form/register.phtml
+++ b/app/design/frontend/base/default/template/customer/form/register.phtml
@@ -102,14 +102,14 @@
                 <li class="wide">
                     <label for="street_1" class="required"><em>*</em><?php echo $this->__('Street Address') ?></label>
                     <div class="input-box">
-                        <input type="text" name="street[]" value="<?php echo $this->escapeHtml($this->getFormData()->getStreet(1)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__('Street Address')) ?>" id="street_1" class="input-text <?php echo $_streetValidationClass ?>" />
+                        <input type="text" name="street[]" value="<?php echo $this->escapeHtml($this->getFormData()->getStreet(0)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__('Street Address')) ?>" id="street_1" class="input-text <?php echo $_streetValidationClass ?>" />
                     </div>
                 </li>
             <?php $_streetValidationClass = trim(str_replace('required-entry', '', $_streetValidationClass)); ?>
             <?php for ($_i = 2, $_n = $this->helper('customer/address')->getStreetLines(); $_i <= $_n; $_i++): ?>
                 <li class="wide">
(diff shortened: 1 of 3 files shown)


kiatng:
My bad, I don't know how I replaced every line in the rwd register.phtml even though I just changed 2 lines.  It's embarrassing but what did I do wrong?

Flyingmana:
Oh, did not see this for some time. Usually this is caused by using a different line ending, happens most times on windows.
Can you go through the answers in https://stackoverflow.com/questions/3920650/git-diff-show-me-line-ending-changes/11509629 and check if it is the lineendings and fix the last commit?

In case you are not experienced with fixing pushed commits, I can search for a good tutorial

kiatng:
OK, I revert the commit , fix the end of line thingy with `git config --global core.autocrlf true`, hope it's OK now.

PR #523, opened 2018-09-13
Reduce unnecessary History updates on order saving.

spinsch:
if store-id stored into model data array, the whole history collection of an order will get updated.

to reduce unnecessary mysql updates, i've used the store id directly from the order object.

fyi: all changes into a model data array will set a flag (save me again).

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Status/History.php
+++ b/app/code/core/Mage/Sales/Model/Order/Status/History.php
@@ -75,7 +75,7 @@ class Mage_Sales_Model_Order_Status_History extends Mage_Sales_Model_Abstract
     }
 
     /**
-     * Set order object and grab some metadata from it
+     * Set order object
      *
      * @param   Mage_Sales_Model_Order $order
      * @return  Mage_Sales_Model_Order_Status_History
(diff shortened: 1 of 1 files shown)


Flyingmana:
how does this affect order comments done from the Customer via Frontend?

colinmollenhour:
Good observation, I've seen this in many places where the setter causes the `hasDataChanges` flag to be true resulting in needless saves. The pattern I prefer to use is to check the value before setting it. In this case something like:

```php
    public function setOrder(Mage_Sales_Model_Order $order)
    {
        $this->_order = $order;
        if ($this->getStoreId() != $order->getStoreId()) {
            $this->setStoreId($order->getStoreId());
        }
        return $this;
    }
```

tmotyl:
Can we agree how to proceed with this one?

kiatng:
I tested this PR in backend order page by adding test comments, it doesn't prevent saving of the history object as intended. This is probably due to `load()` in the collection that calls `setData()`:

https://github.com/OpenMage/magento-lts/blob/8e9ac45bb2383aabdbb746497369dbb53884fdf7/lib/Varien/Data/Collection/Db.php#L588-L595

I will create another PR that will fix this issue.

PR #529, opened 2018-09-20
Fix in ACL checking for newsletter subscription

markvds:
1. The checked ACL identifier (`customer/newsletter`) does not exist, so any restricted admin user cannot manage subscriptions.
2. It should check the same identifier that the customer edit page uses to detemine if the tab with checkbox should be visible (see https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php#L92).

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/CustomerController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/CustomerController.php
@@ -317,7 +317,7 @@ class Mage_Adminhtml_CustomerController extends Mage_Adminhtml_Controller_Action
                 }
             }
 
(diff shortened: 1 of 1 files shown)


markvds:
Well this is embarrassing! Submitting a pull request with a typo :confused: Sorry!

Although I'm not 100% familiar with the PR system of Github, I think I have made the changes and the PR is ready to merge.

colinmollenhour:
It's often the one-liners where mistakes slip through. :)

Just an FYI, personally I would "ammend" the commit and then force-push it to replace the incorrect commit instead of add to it, but we can also squash in the merge.

PR #347, opened 2017-10-04
EAV - Query improvements

dng-dev:
This pull requests converts EAV - based queries for $model->load() and $collection->load() queries. Affected Entities are catalog_product catalog_categories customers. The idea behind is to convert the SQL - Queries to simpler one which also make proper use of the query cache. Instead of having one HUGE query with UNION ALL for the eav tables each table is queried separatly. Another improfement is that the if expression is gone and the order by store_id is happening in PHP which reduces the overall load on the database. Another improfement is the static sorted attribute cache to avoid double sort of attributes when 2 products whinin an listing have the same attribute set.

PR #531, opened 2018-09-21
splitted union queries for eav entities to simpler queries

dng-dev:
This is a follow up to the past PR #347.

This pull requests converts EAV - based queries for $model->load() and $collection->load() queries. Affected Entities are catalog_product catalog_categories customers. The idea behind is to convert the SQL - Queries to simpler one which reduces the load on the database and don't create tmp-tables and also depending on the configuration tmp-tables on disc. Instead of having one HUGE query with UNION ALL for the eav tables each table is queried separatly. Another improfement is that the if expression is gone and the order by store_id is happening in PHP which reduces the overall load on the database. Another improfement is the static sorted attribute cache to avoid double sort of attributes when 2 products whinin an listing have the same attribute set.

@seansan @colinmollenhour @tmotyl @partounian the other PR is closed

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Abstract.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Abstract.php
@@ -982,27 +982,78 @@ abstract class Mage_Eav_Model_Entity_Abstract extends Mage_Core_Model_Resource_A
 
         Varien_Profiler::start('__EAV_LOAD_MODEL_ATTRIBUTES__');
 
-        $selects = array();
-        foreach (array_keys($this->getAttributesByTable()) as $table) {
-            $attribute = current($this->_attributesByTable[$table]);
-            $eavType = $attribute->getBackendType();
-            $select = $this->_getLoadAttributesSelect($object, $table);
-            $selects[$eavType][] = $this->_addLoadAttributesSelectFields($select, $table, $eavType);
-        }
-        $selectGroups = Mage::getResourceHelper('eav')->getLoadAttributesSelectGroups($selects);
-        foreach ($selectGroups as $selects) {
-            if (!empty($selects)) {
-                $select = $this->_prepareLoadSelect($selects);
-                $values = $this->_getReadAdapter()->fetchAll($select);
-                foreach ($values as $valueRow) {
-                    $this->_setAttributeValue($object, $valueRow);
+        $tableAttributes = array();
+        $attributeTypes  = array();
+
+        foreach ($this->getAttributesById() as $attributeId => $attribute){
+            /** @var Mage_Catalog_Model_Resource_Eav_Attribute $attribute */
+            if ($attribute instanceof  Mage_Catalog_Model_Resource_Eav_Attribute && !$attribute->isStatic()) {
+                $tableAttributes[$attribute->getBackendTable()][] = $attributeId;
+                if (!isset($attributeTypes[$attribute->getBackendTable()])) {
+                    $attributeTypes[$attribute->getBackendTable()] = $attribute->getBackendType();
                 }
             }
+        }
+
+        // get eav select skeleton
+        $select = $this->_getLoadAttributesSelectSkeleton($object);
+        foreach ($tableAttributes as $table=>$attributes) {
+            try {
+                /** @var Varien_Db_Select $select */
+                $select->reset(Varien_Db_Select::FROM);
+                $select->from(array('eav_value' => $table), null);
+                $values = $this->getReadConnection()->fetchAll($select);
+            } catch (Exception $e) {
+                throw $e;
+            }
+
+            $storeValues = [];
+            foreach ($values as $valueRow) {
+                if ($valueRow['store_id'] == 0) {
+                    $this->_setAttributeValue($object, $valueRow);
+                } else {
+                    $storeValues[] = $valueRow;
+                }
+            }
+            foreach ($storeValues as $valueRow) {
+                $this->_setAttributeValue($object, $valueRow);
+            }
+            unset($values);
+            unset($storeValues);
         }
 
         Varien_Profiler::stop('__EAV_LOAD_MODEL_ATTRIBUTES__');
 
         return $this;
+    }
+
+    /**
+     * load attribute select skeleton query
+     *
+     * @param   Mage_Eav_Model_Entity_Attribute
+     * @return  Varien_Db_Select
+     */
+    protected function _getLoadAttributesSelectSkeleton($object)
+    {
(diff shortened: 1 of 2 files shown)


colinmollenhour:
Nice work, Daniel, reducing on-disk temp tables would be a great improvement!

I can see for loading collections of products that the result size would be potentially very large, but what about for loading a single product? I'm wondering if in this case the UNION method could be faster due to network latency? If so perhaps the UNION method could be used for single loads and the new one for collection loads?

Would be great to include some examples of the before/after you have experienced from this update. Thanks!

dng-dev:
Hey @colinmollenhour will try to make some benchmarks.

IvanChepurnyi:
Maybe it makes sense instead of separating queries use UNION ALL, as UNION ALL does not do any table on disk materialization as it does not need to sort results for merged result.

IvanChepurnyi:
So this code can be improved in couple of ways:
1. We can remove `store_id` left join in original Magento Catalog EAV collection [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Model/Resource/Collection/Abstract.php#L103) and replace with `IN(default_store_id, current_store_id)`
2. Then rewrite this method to avoid using `fetchAll` as it basically wastes CPU time on copying data into array that gets disposed just right away [here](https://github.com/OpenMage/magento-lts/blob/6282a9a6b8a4452543b82f52ea785cfff3b6ea79/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php#L1097). It needs to use `query` method on adapter that returns back iterator of PDO statement that can be directly passed into foreach.
3. Override `_setItemAttributeValue` method in `Mage_Catalog_Model_Resource_Collection_Abstract` and check if record has if `store_id` is not zero set value to product, but if `store_id` equal to 0 and value is already present in product - don't do anything. 

In such a way you even get better performance then eliminating UNION ALL as it does reduce latency, the issue in the select query itself.

colinmollenhour:
Great suggestions! So happy to have contributions from you guys! I think before we can merge any improvements we should have some real-world numbers, though.

P.S. My team uses Ecomdev_PHPUnit daily and are so grateful for it, @IvanChepurnyi!

dng-dev:
Will work on it in the next time

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -1092,37 +1092,64 @@ public function _loadAttributes($printQuery = false, $logQuery = false)
...
+            ->where("$entityIdField IN (?)", array_map('intval', array_keys($this->_itemsById)))
+            ->where('attribute_id IN (?)', array_map('intval', $this->_selectAttributes))
+            ->order(new Zend_Db_Expr('NULL'));
+        if (false === method_exists($this, 'getStoreId')) {
+            return $select;
+        }
+        $storeId = $this->getStoreId();
+        if (true === method_exists($this, 'getDefaultStoreId') && (int)$storeId > 0) {
+            $select->where('store_id IN(?)', array((int)$storeId, (int)$this->getDefaultStoreId()));
+        } else if (false === method_exists($this, 'getDefaultStoreId') || $storeId == 0) {
```
joshua-bn, line 1146: This can never get to here because line 1140 does a check for `false === method_exists($this, 'getStoreId')`

app/code/core/Mage/Eav/Model/Entity/Abstract.php
```diff
@@ -982,29 +982,80 @@ protected function _loadModelAttributes($object)
...
+            ->where('attribute_id IN (?)', array_map('intval', array_keys($this->getAttributesById())))
+            ->order(new Zend_Db_Expr('NULL'));
+        if (false === method_exists($object, 'getStoreId')) {
+            return $select;
+        }
+
+        $storeId = $object->getStoreId();
+        if ((int)$storeId > 0 && true === method_exists($this, 'getDefaultStoreId')) {
+            $select->where('store_id IN(?)', array((int)$storeId, (int)Mage_Catalog_Model_Abstract::DEFAULT_STORE_ID));
+        } else if (false === method_exists($this, 'getDefaultStoreId') || $storeId == 0) {
```
joshua-bn, line 1052: It can never get to here because line 1045

joshua-bn (changes requested):

joshua-bn:
> So this code can be improved in couple of ways:
> 
> 1. We can remove `store_id` left join in original Magento Catalog EAV collection [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Model/Resource/Collection/Abstract.php#L103) and replace with `IN(default_store_id, current_store_id)`
> 2. Then rewrite this method to avoid using `fetchAll` as it basically wastes CPU time on copying data into array that gets disposed just right away [here](https://github.com/OpenMage/magento-lts/blob/6282a9a6b8a4452543b82f52ea785cfff3b6ea79/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php#L1097). It needs to use `query` method on adapter that returns back iterator of PDO statement that can be directly passed into foreach.
> 3. Override `_setItemAttributeValue` method in `Mage_Catalog_Model_Resource_Collection_Abstract` and check if record has  `store_id` that is not zero, we set value to product, but if `store_id` equal to 0 and value is already present in product - don't do anything.
> 
> In such a way you even get better performance then eliminating UNION ALL as it does reduce latency, the issue in the select query itself.

Maybe a PR?

joshua-bn:
> Then rewrite this method to avoid using fetchAll as it basically wastes CPU time on copying data into array that gets disposed just right away here. It needs to use query method on adapter that returns back iterator of PDO statement that can be directly passed into foreach.

That might be a micro-optimization that doesn't really do anything for this. I see a benefit in it collecting the values first. That way you can throw/catch any issues before calling _setItemAttributeValue() (which sets values and may throw its own exceptions)

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @dng-dev if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

fballiano:
it's sad to see such great ideas posted, commented and abandoned :-(

fballiano:
This PR has too many unanswered questions, it was a great idea but it totally abandoned since more than 2 years and it needs to be rebased to "main" and has conflicts. sorry guys, I've labeled "don't forget this PR" but I'll close it.

I'd love to see it fixed and reopened but I've zero hope for that.

PR #533, opened 2018-09-21
added more Profiler statements

dng-dev:
Added more profiler statements for templates

### DIFF
--- a/app/code/core/Mage/Core/Block/Abstract.php
+++ b/app/code/core/Mage/Core/Block/Abstract.php
@@ -567,6 +567,7 @@ abstract class Mage_Core_Block_Abstract extends Varien_Object
      */
     public function getChildHtml($name = '', $useCache = true, $sorted = false)
     {
+        Varien_Profiler::start('getChildHtml: '.$name);
         if ($name === '') {
             if ($sorted) {
                 $children = array();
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Core/Block/Abstract.php
```diff
@@ -582,8 +583,10 @@ public function getChildHtml($name = '', $useCache = true, $sorted = false)
             }
             return $out;
```
colinmollenhour: Needs a `stop` before the return to keep the stack aligned.

app/code/core/Mage/Core/Block/Text/List.php
```diff
@@ -36,11 +36,13 @@ protected function _toHtml()
     {
         $this->setText('');
         foreach ($this->getSortedChildren() as $name) {
+            Varien_Profiler::start('getSortedChildren: ' . $name);
             $block = $this->getLayout()->getBlock($name);
             if (!$block) {
                 Mage::throwException(Mage::helper('core')->__('Invalid block: %s', $name));
```
colinmollenhour: Needs a `stop` before the exception is thrown to keep the stack aligned.

app/code/core/Mage/Core/Block/Abstract.php
Flyingmana: or removing the return $out in this place, as it is returned in the end

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @dng-dev  if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

fballiano:
No feedback to the reviews from original author. I'm closing since I don't even think this PR is really that important.

Issue #525, opened 2018-09-14
Remove "compiler" related code


Flyingmana:
For Years the Magento1 Compiler code is mostly unused and recommended against to use it.
To reduce Code/Logic (and because someone suggested it) I would like to see the Compiler related code be removed (maybe leave empty methods, in case its needed for compatibility reasons?)

What are your takes on this?

Context what it is: 
https://magento.stackexchange.com/questions/179093/what-is-compilation-in-magento-and-its-use
and prove I have prejudice here: 
https://magento.stackexchange.com/questions/12996/when-should-i-use-the-magento-compiler/15204#15204

rafaelpatro:
[Referring Alan Storm article 2011:](https://alanstorm.com/magento_compiler_path/)
_"In computer science the term compiler usually refers to transforming code from one language (like C) into another language (like binary machine code). The Magento compiler transforms PHP code into – more PHP code!"_

I think since 2011 PHP performance was so increased that searching for M1 files in separate directories is no more a bottleneck. I don't remember where to find the reference but some time ago I've read a benchmark showing that M1 compiler has no effect in PHP7.

Flyingmana:
a big part of this improvement is because of the OpCache, which additionally caches file access so effectively, that the number of files became not relevant for PHP anymore

FredericMartinez:
+1

Ig0r-M:
:+1:

PR #534, opened 2018-09-21
removed Mage_Compiler

dng-dev:
Removed Mage_Compiler #525


LeeSaferite:
Does it make sense to fix the autoloader at the same time?

It should be returning a false if the class isn't found. That would allow people to easily use composer autoloader in conjunction.

dng-dev:
Will look over this later and implement it.

Am Freitag, 21. September 2018 schrieb Lee Saferite :

dng-dev:
@seansan as mentioned i added the missing return statement for the autoloader.

Flyingmana:
the __autoload function can get deleted completely. Since the Varien Autoloader uses the SPL autoloader, this one is never in use.

PR #535, opened 2018-09-25
Add Customer Email unsubscribe option to credit memo

seansan:
Add Customer Email unsubscribe option to credit memo

**The rationale**

- email subscriber management is becoming increasingly important (esp email sender credability)
- it is a given that customers sometimes cancel orders, make returns, buy products twice with both you and the competitor and return 1, are (structurally) unhappy .. no problem, daily business ;)
- however we have seen that this niche customer group is a very large contributor to email unsubscribes (even marking it as spam, when they did subscribe themselves)
- for years we have focussed on OPTIN and list growth, now the time has come for email OPT OUT

So our backoffice team has made the request to add an option during credit memo process to more easily unsubscribe customers in the above situation ... (the workaround is via account etc etc)

**Coding**

I have added a simple update that does not break anything (so it seems). I did not manage to add it into the test scripts where I can see it also exists. But see this as logic to be improved by a more senior programmer. I am no PHP developer. 

I think it would be nice/good to have this ... again it does not break anything either

Hope it helps

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Sales/Order/CreditmemoController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Sales/Order/CreditmemoController.php
@@ -295,6 +295,16 @@ class Mage_Adminhtml_Sales_Order_CreditmemoController extends Mage_Adminhtml_Con
                     }
                 }
 
+                $succesubscribemsg = '';
+                if (isset($data['newsletter_customer_unsubscribe'])) {
+                    $order = $creditmemo->getOrder();
+                    $email = (string)$order->getBillingAddress()->getEmail();
+                    if (!$email) $email = $order->getCustomerEmail();
+                    if (Mage::getModel('newsletter/subscriber')->loadByEmail($email)->unsubscribe()) {
+                        $succesubscribemsg = $this->__(' Newsletter unsubscribe success (%s).', $email);
+                    }
+                }                
+                
                 if (isset($data['do_refund'])) {
                     $creditmemo->setRefundRequested(true);
                 }
@@ -310,7 +320,7 @@ class Mage_Adminhtml_Sales_Order_CreditmemoController extends Mage_Adminhtml_Con
                 $creditmemo->getOrder()->setCustomerNoteNotify(!empty($data['send_email']));
                 $this->_saveCreditmemo($creditmemo);
                 $creditmemo->sendEmail(!empty($data['send_email']), $comment);
-                $this->_getSession()->addSuccess($this->__('The credit memo has been created.'));
+                $this->_getSession()->addSuccess($this->__('The credit memo has been created.') . $succesubscribemsg);
                 Mage::getSingleton('adminhtml/session')->getCommentText(true);
                 $this->_redirect('*/sales_order/view', array('order_id' => $creditmemo->getOrderId()));
                 return;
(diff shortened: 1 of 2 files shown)


app/design/adminhtml/default/default/template/sales/order/creditmemo/create/items.phtml
```diff
@@ -104,6 +104,10 @@
               <label class="normal" for="notify_customer"><?php echo Mage::helper('sales')->__('Append Comments') ?></label>
               <input id="notify_customer" name="creditmemo[comment_customer_notify]" value="1" type="checkbox" />
           </p>
+          <p>
```
colinmollenhour: Why not first check that customer is subscribed and only show this option if so?

app/code/core/Mage/Adminhtml/controllers/Sales/Order/CreditmemoController.php
```diff
@@ -310,7 +320,7 @@ public function saveAction()
                 $creditmemo->getOrder()->setCustomerNoteNotify(!empty($data['send_email']));
                 $this->_saveCreditmemo($creditmemo);
                 $creditmemo->sendEmail(!empty($data['send_email']), $comment);
-                $this->_getSession()->addSuccess($this->__('The credit memo has been created.'));
+                $this->_getSession()->addSuccess($this->__('The credit memo has been created.') . $succesubscribemsg);
```
colinmollenhour: I've not seen this pattern anywhere else in Magento. There is no harm in calling `addSuccess` multiple times.

app/code/core/Mage/Adminhtml/controllers/Sales/Order/CreditmemoController.php
```diff
@@ -295,6 +295,16 @@ public function saveAction()
                     }
                 }
 
+                $succesubscribemsg = '';
+                if (isset($data['newsletter_customer_unsubscribe'])) {
+                    $order = $creditmemo->getOrder();
+                    $email = (string)$order->getBillingAddress()->getEmail();
+                    if (!$email) $email = $order->getCustomerEmail();
+                    if (Mage::getModel('newsletter/subscriber')->loadByEmail($email)->unsubscribe()) {
```
colinmollenhour: I think it would Be good to check if the subscriber was actually loaded before calling `unsubscribe` to avoid an exception thrown.

seansan:
All true. This was a quick fix.

If we think it can be added we can update The code to standards and create
proper pull

seansan:
Ho do I approve to these changes and continue from there? I cannot select
the 2nd option approve

[image: image.png]

seansan:
Yes can do local but the web GUI apparently also has some functionality

Only I cannot use  it

Thx

app/code/core/Mage/Adminhtml/controllers/Sales/Order/CreditmemoController.php
```diff
@@ -295,6 +295,19 @@ public function saveAction()
                     }
                 }
 
+                $succesubscribemsg = '';
+                if (isset($data['newsletter_customer_unsubscribe'])) {
+                    $order = $creditmemo->getOrder();
+                    $email = (string)$order->getBillingAddress()->getEmail();
```
colinmollenhour: Should probably not assume that `getBillingAddress` returns an object.

fballiano:
This PR is based on 1.9.3.x which is a non-maintained branch at this point. Since we can't work on this anymore I'm closing it. @seansan if you feel like rewriting the PR on 1.9.4.x feel absolutely free to reopen it, it will be much appreciated.

fballiano:
> Credit memo is not a newsletter, I don't really understand why you want to allow unsubscribe here.

agree

PR #536, opened 2018-09-27
Update Mysqli.php

henrykbrzoska:
No comment needed IMO


Flyingmana:
huh, obvious this should get fixed. Does anyone know if and in which context this is still used?
As far as I know, Magento is using PDO. Maybe we can just delete this all together?

henrykbrzoska:
Delete 'anything' in mage1 isn't best idea.

PR #538, opened 2018-09-28
Unable to search customer by first and last name on sales_order_grid

kestraly:
Unable to search customer by first and last name on sales_order_grid
This is due to the middle name requiring a space.

More info on this can be seen at, which still exists in current 1.9.3.10 version
https://magento.stackexchange.com/questions/80196/magento-1-9-2-0-table-sales-flat-order-grid-contains-extra-space-in-customer

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Observer.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Observer.php
@@ -737,6 +737,21 @@ class Mage_CatalogInventory_Model_Observer
             }
         }
 
+        /**
+         * ref: https://github.com/OpenMage/magento-lts/issues/152
+         *
+         * get product stock setup.
+         * We only want to re-index products that actually have stock management setup.
+         * This limits the number of stock re-indexing that takes place,
+         * especially in stores where stock is not managed
+         **/
+        $productIds = array_map('intval', $productIds);
+        $stockCollection = Mage::getModel('cataloginventory/stock_item')->getCollection()
+            ->addFieldToFilter('product_id', array('in' => $productIds))
+            ->addFieldToFilter('manage_stock', array('eq' => 1));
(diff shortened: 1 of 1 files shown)


seansan:
This would explain countless occasions of surprise and frustration

If this is the fix then I suggest a quick push

kestraly:
Have it working on production env.
Just noticed same issue is on Invoice, Shipments and Creditmemo Grids

sreichel:
> If this is the fix then I suggest a quick push

I can reprouce it, but the fix seems not to work. (?)

kestraly:
Fix only works for new orders coming in. To back port you have to run 

`Mage::getModel('sales/order')->getResource()->updateGridRecords(Mage::getResourceModel('sales/order_collection')->getAllIds());`

sreichel:
@kestraly I should better read the hole answer :) Thx.

colinmollenhour:
Also there is `getCheckSql` method which would be more inline with the Magento way of building queries even though it really doesn't matter since MySQL is the only database actually implemented.

kestraly:
I'll adjust this then to use getCheckSql

kestraly:
In changing to CheckSql - It goes back to requiring a double space to search. Scratching my head as to why that would be. @colinmollenhour - Any thoughts?

`$ifnullMiddle  = $adapter->getCheckSql('{{table}}.middlename IS NULL OR {{table}}.middlename="" OR {{table}}.middlename=" "', '""', '" "');`

colinmollenhour:
I don't know other than I always use single quotes in SQL for strings. Does this work?

```
$ifnullMiddle = $adapter->getCheckSql("{{table}}.middlename IS NULL OR {{table}}.middlename=''", "''", "' '");
```

kestraly:
@colinmollenhour - Unfortunately gives the same problem. Hmmm

kestraly:
Updated all grids

PR #539, opened 2018-09-29
Upstream 1.9.3.10

Flyingmana:
please do merge with doing a merge commit, to keep the merges for the next versions as easy as just doing a PR from the OpenMage Mirror


Flyingmana:
the commit was already checked on the OpenMage mirror, and there is no conflict shown.
I would say its safe to merge and if there is no explicite veto, will do this next week

seansan:
Can’t wait;)thanks!

PR #555, opened 2018-10-17
Fix calculation of base total in magento report with discount

kpitn:
Actually magento store discount with negative value :  - 10

If you have an order with subtotal = 90 and base_discount_invoiced = -10

When Magento do base_subtotal_invoiced  - base_discount_invoiced => 90 -- 10 => 90 +10 => 100


kpitn:
1. Make an order with some coupon with discount amount.
2. In admin goes to > Report > Sales > Coupon
3. Display your order : with the correct date / order status

You will see the last column name "Total amount" with incorect values.  
discount

colinmollenhour:
Would `* -1` be more correct?

tbaden:
@LeeSaferite @Flyingmana @drobinson 
seems ready to merge

PR #556, opened 2018-10-18
Add URI info to log files.

kiatng:
Add URI info to the log files, which looks like this 

in system.log:
> 2018-10-13T03:32:43+00:00 DEBUG (7) **[/assets.html]**: Security problem: customs/asset_list has not been whitelisted.

in exception.log:
> 2018-10-18T05:55:08+00:00 ERR (3) **[/index.php/emgs/adminhtml_sales_order_itemOption/save/key/7f...dc7/]**: 
> exception 'Exception' with message 'Warning: explode() expects parameter 2 to be string, array given  in /home/web/public_html/magento/emgsihe/app/code/core/Mage/Catalog/Model/Product/Option/Type/Select.php on line 128' in /home/web/public_html/magento/emgsihe/app/code/core/Mage/Core/functions.php:245
> Stack trace:
> #0 [internal function]: mageCoreErrorHandler(2, 'explode() expec...', '/home/web/publi...', 128, Array)
> ...

I find the URI info useful in debugging during development.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -838,7 +838,7 @@ final class Mage
                 }
                 $writer->setFormatter($formatter);
                 $loggers[$file] = new Zend_Log($writer);
-                $loggers[$file]->setEventItem('uri', $_SERVER["REQUEST_URI"]);
+                $loggers[$file]->setEventItem('uri', isset($_SERVER["REQUEST_URI"]) ? $_SERVER["REQUEST_URI"] : 'No URI');
             }
 
             if (is_array($message) || is_object($message)) {
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -827,7 +827,7 @@ final class Mage
                     chmod($logFile, 0640);
                 }
 
-                $format = '%timestamp% %priorityName% (%priority%): %message%' . PHP_EOL;
+                $format = '%timestamp% %priorityName% (%priority%) [%uri%]: %message%' . PHP_EOL;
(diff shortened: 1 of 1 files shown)


seansan:
for debugging errors like "HEADERS ALREADY SENT" this is very very handy

Flyingmana:
what is the expected output for cli errors?

kiatng:
If cli, we will have _PHP Notice:  Undefined index: REQUEST_URI_. Thanks for the catch. Let me fix that.

colinmollenhour:
Not opposed, but have you seen https://github.com/firegento/firegento-logger? This works amazingly well with Sentry (PR not yet merged for lack of active reviewers).

Why not use a separate format for CLI and then use something like `if (PHP_SAPI == 'cli') { ...`?

kiatng:
No, I haven't seen _firegento_logger_, which seems to be rather complex. Thanks for the info. 

I don't have experience with CLI in Magento. I have never seen any log messages from CLI. What would be useful information to include in the event of exception or log from CLI? I thought of including `$_SERVER['SCRIPT_FILENAME']`, which is the absolute pathname of the currently executing script. However, the file info would be in the trace anyway, so including 'SCRIPT_FILENAME' would not be useful. Any other idea?

dng-dev:
I would recommend to test for session id in url and remove it

kiatng:
Agreed. Firegento_Logger extension can fulfill this and do much more. Thank you all for the time and thoughts.

Issue #505, opened 2018-08-23
Does M LTS support Php 7.2.x?


seansan:
Does M LTS support Php 7.2.x?

colinmollenhour:
I think the main blocker is that PHP 7.2 removed mcrypt so see #129 and #449.

Aside from that I don't know of any major issues but wouldn't be surprised if there are minor issues from things like PCRE2.

AFAIK, PHP 7.1 is the latest fully supported version. Pull requests welcome! :)

See: https://github.com/php/php-src/blob/PHP-7.3/UPGRADING

seansan:
https://github.com/Inchoo/Inchoo_PHP7/wiki/PHP-7.1-and-7.2

Flyingmana:
> getimagesize() and related functions now report the mime type of BMP images
    as image/bmp instead of image/x-ms-bmp, since the former has been registered
with the IANA (see RFC 7903).

may need a change

most other 7.3 related BC breaks dont affect Magento I think, but who knows which odd constructs are hidden internally somewhere.

Flyingmana:
can someone look trough the changes of https://magento.com/tech-resources/download#download2240 ?

colinmollenhour:
The bulk of it is updating the libraries like phpseclib and adding mcrypt_compat. As it is the official solution I see no reason to not go ahead and merge it even though if you're not using downloader probably a lot of that code is not needed. There are some other changes in there like some explicit type casting and adding arguments to overridden methods so it's better that we've done to date.

seansan:
https://inchoo.net/magento/magento-1-official-php-7-2-patches/

Quazz:
So the Inchoo solution needs to be removed and the patch applied.

seansan:
https://github.com/hans2103/Magento?files=1

corbinu:
After https://github.com/OpenMage/magento-lts/pull/557 I think all that needs to be done for this that the composer requirement should include 7.2 as a valid version.

seansan:
For means of closure ;)

Has anyone tested 7.2 PHP support on their server? (with the merges above)

Running on production?

(if yes: then maybe I dare to take a leap on our accept server .... and we can close the issue)

@corbinu @Flyingmana @colinmollenhour

seansan:
Update. Tested today on DEV. No problems so far. On moving to production nginx env. 
So seems to work fine. 1.9.4.x branch.

corbinu:
[...]

PR #557, opened 2018-10-19
apply php 7.2 official patch

cieslix:
I applied official php 7.2 patch 

I saw there is conversation related https://github.com/OpenMage/magento-lts/issues/505 

I did some testing on PHP 7.1 with mcrypt / PHP 7.1 without mcrypt / PHP 7.2 without mcrypt on vanilla Magento 1.9.3.10 and there was no errors.


app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -651,17 +651,28 @@ public function jsonEncode($valueToEncode, $cycleCheck = false, $options = array
...
+                $encodedValue = 'true';
+                break;
+            case (false === $encodedValue):
+                $encodedValue = 'false';
+                break;
+            case ('' === $encodedValue):
+                $encodedValue = '""';
+                break;
+            default:
+                // do nothing
```
colinmollenhour: This is just whitespace change which replaces the original update from https://github.com/OpenMage/magento-lts/pull/83. I think whitespace-only changes should be avoided unless there is good reason, particularly when the author is not the same.

lib/Varien/Autoload.php
```diff
@@ -91,7 +91,7 @@ public function autoload($class)
         }
         $classFile.= '.php';
         //echo $classFile;die();
-        return include $classFile;
```
Flyingmana: we dont have a `if file_exists()` check here yet, thats something we should keep in mind as a future change

app/code/core/Mage/Core/functions.php
```diff
@@ -178,6 +160,11 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline){
         define('E_DEPRECATED', 8192);
     }
 
+    // Suppress deprecation warnings on PHP 7.x
+    if ($errno == E_DEPRECATED && version_compare(PHP_VERSION, '7.0.0', '>=')) {
```
Flyingmana: we should think about removing this again at a later point, to not make future updates harder

app/code/core/Mage/Core/Helper/Data.php
Flyingmana: if this reflexs what is in magento now, I think its better to keep the coding style they do use in case of future changes.

lib/Varien/Autoload.php
colinmollenhour: I think suppressing the `include` warning is functionally equivalent to using a conditional on `file_exists`, right?

app/code/core/Mage/Core/Helper/Data.php
cieslix: it was included into the patch, i didn't add anything from myself. nevertheless new code use zend code style guide https://framework.zend.com/manual/1.11/en/coding-standard.coding-style.html#coding-standards.coding-style.control-statements.switch

lib/Varien/Autoload.php
Flyingmana: except you set scream.enabled=true, which makes all the @ ignored

colinmollenhour: Then it seems like a good solution, the user that wants to see all warnings can see them and everyone else will not. Also I just remembered that `file_exists` does not search the include path like `include` so they are not functionally equivalent anyway. You could use `stream_resolve_include_path` but I still think `@include` is fine.

Flyingmana: interesting, I never faced a problem because of the include path not respected in this check.
The problem with this is, thats not a warning which just lands in the log, but it breaks any application which somewhere relies on a `class_exist()` which should return `false`, because if you set `scream.enabled=true`, the application will not work anymore.

And this added error supressor now makes sense, some of the new added code relies on class_exists, so it would break if the error is triggered.

anyway, we should move this to a separate discussion and decide later. The PR is good so far.

tomlankhorst:
#566

lib/Varien/Autoload.php
bramstroker: Hi, we are facing a lot of issues in our shops with this error suppressing introduced. 
I don't think it is a good idea to suppress all kind of errors which could be triggered when including a file. Fatal errors like ` Declaration of xx must be compatible with yy` and `Cannot redeclare` will also be silently suppressed now and the application will stop excecuting resulting in a blank page (or partial HTML). The fatal error is nowhere to be found because it is suppressed. This makes debugging very time consuming and we will not get noticed if a customer experiences this kind of error. There are legitimate cases to use the @ sign, but autoloading is definitely not a place where you want to use it imo. 
`scream.enabled=true` on production environments is definitely a no go, so there is no way for us to log this errors on production.
I am not really sure which issue you are trying to resolve with the error suppressing. Is the `include` behaviour different in PHP7.2?
Please give me your thoughts about this. There must be another solution as all other autoloaders (Composer, Symfony etc.) will also not suppress include errors, and they have no problems with `class_exists()` calls.

edwinljacobs: Why introduce error suppressing for one of the most critical components in an application. If something is wrong with code being included or some file is missing it should be abundantly clear as in almost all cases it will result in the application not being able to function correctly.

colinmollenhour: These are great points. @cieslix do you know the reason this change was introduced specifically?

PR #562, opened 2018-11-03
Limit post-checkout stock re-index to only products that manage stock…

ProxiBlue:
ref #152

The core post checkout stock reindex routine will stock re-indx all products in order placed.
This is not needed, when an item is not stock managed.

This change has been used in a live environment on two sites. Both sites only manage about 10% of their stock items in magento.
This change removed all instanced of the dreaded Lock Table during checkout.

The change will most likely have zero effect on stores with full stock management.

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Observer.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Observer.php
@@ -745,9 +745,11 @@ class Mage_CatalogInventory_Model_Observer
          * This limits the number of stock re-indexing that takes place,
          * especially in stores where stock is not managed
          **/
+        $productIds = array_map('intval', $productIds);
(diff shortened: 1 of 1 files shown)


ProxiBlue:
@colinmollenhour 

Changed as requested.
Normal collection usage of ->addFieldToSelect() still left other columns, so I want with the code as suggested.

The resulting SQL is:

```SELECT `main_table`.`product_id` FROM `cataloginventory_stock_item` AS `main_table` INNER JOIN `catalog_product_entity` AS `cp_table` ON main_table.product_id = cp_table.entity_id WHERE (`product_id` IN(11)) AND (`manage_stock` = 1)```

Also note that I cast all values of the $product_id array to INT, prior to SQL usage.
As far as I am aware, using INTS in SQL IN statements are more performant than using strings.

PR #564, opened 2018-11-08
fixed wrong creditmemo base tax calculation

sprankhub:
See https://github.com/PAYONE-GmbH/magento-1/issues/267#issuecomment-436907099


kestraly:
+1 For Passing

PR #566, opened 2018-11-13
composer.json require php < 7.3

tomlankhorst:
Allow `composer upgrade` on PHP 7.2.x

#557


Flyingmana:
we did not have someone test it yet, but I think we can change this to <7.4
at least on the syntax level, there are no issues anymore for PHP 7.3
Not sure if we should wait for someone testing it in production or just expect it to work an fix the bugs as they come in.

PR #570, opened 2018-11-22
Url reindexer update

Ig0r-M:
**This work was created more for discussion point of view**

For internal projects we are using Fast Asynchronous Re-indexing for Magento 1 which will be present is some of screenshots.

**Problem**
While working at an internal task following was discovered through debugging (hard debugging) i.e. Function `\Mage_Index_Model_Event::_mergeNewDataRecursive` produces duplicates
image
Based on the image above it's clear that there will be duplicated values in `$current` array
image1
Unfortunately did not dig too much into understand a fix for this approach as i have no information if that logic is indented thus for now you can view my approach at fixing it https://github.com/OpenMage/magento-lts/commit/9119c011461c36772633d1627bb83b833bf0b535

On side not this was impacting category url rewrite logic
image2

Which in turn is used only in https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Catalog/Model/Indexer/Url.php#L240-L244 so for that just making values unique https://github.com/OpenMage/magento-lts/commit/51a94e3ed7ed7e4397cd274a931cae874e864973 was fastest way to update it.

P.S. i have raised this PR mainly to try and understand more if that is a bug or a feature :smile:

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Indexer/Url.php
+++ b/app/code/core/Mage/Catalog/Model/Indexer/Url.php
@@ -237,9 +237,12 @@ class Mage_Catalog_Model_Indexer_Url extends Mage_Index_Model_Indexer_Abstract
                  $urlModel->refreshProductRewrite($productId);
             }
         }
-        if (isset($data['rewrite_category_ids'])) {
+
+        $rewriteCategoryIds = $data['rewrite_category_ids'] ?? null;
+        if ($rewriteCategoryIds) {
+            $rewriteCategoryIds = array_unique($rewriteCategoryIds);
             $urlModel->clearStoreInvalidRewrites(); // Maybe some categories were moved
-            foreach ($data['rewrite_category_ids'] as $categoryId) {
+            foreach ($rewriteCategoryIds as $categoryId) {
                 $urlModel->refreshCategoryRewrite($categoryId);
             }
         }
--- a/app/code/core/Mage/Index/Model/Event.php
+++ b/app/code/core/Mage/Index/Model/Event.php
@@ -171,6 +171,8 @@ class Mage_Index_Model_Event extends Mage_Core_Model_Abstract
                 }
             } elseif (!array_key_exists($key, $current) || is_null($current[$key])) {
                 $current[$key] = $previous[$key];
+            } elseif (array_key_exists($key, $current) && array_key_exists($key, $previous) && ($previous[$key] == $current[$key])) {
+                return $current;
             } elseif (!is_array($previous[$key]) && !is_string($key)) {
                 $current[] = $previous[$key];
             }


app/code/core/Mage/Catalog/Model/Indexer/Url.php
```diff
@@ -237,9 +237,12 @@ protected function _processEvent(Mage_Index_Model_Event $event)
                  $urlModel->refreshProductRewrite($productId);
             }
         }
-        if (isset($data['rewrite_category_ids'])) {
+
+        $rewriteCategoryIds = $data['rewrite_category_ids'] ?? null;
```
colinmollenhour: I love the null coalescing operator as much as the next PHP guy, but unfortunately we haven't abandoned PHP 5.x yet..

app/code/core/Mage/Index/Model/Event.php
```diff
@@ -171,6 +171,8 @@ protected function _mergeNewDataRecursive($previous, $current)
                 }
             } elseif (!array_key_exists($key, $current) || is_null($current[$key])) {
                 $current[$key] = $previous[$key];
+            } elseif (array_key_exists($key, $current) && array_key_exists($key, $previous) && ($previous[$key] == $current[$key])) {
+                return $current;
```
colinmollenhour: Since this is inside a foreach loop it should not `return`  in this case. Take for example:

```
$previous = ["a", "b"];
$current = ["a", "c"];
```

My understanding is the result should ideally be `["a", "b", "c"]` but this patch would yield `["a", "c"]` (with the current code probably yielding `["a", "b", "a", "c"]`).

It may be better to allow a duplicated value and then address that using `array_unique` like your other commit does since it is hard to say what making changes to this method could impact.

app/code/core/Mage/Catalog/Model/Indexer/Url.php
Ig0r-M: this is for our internal project.
Do you want for me to update to be 5.x+ compatible ?

app/code/core/Mage/Index/Model/Event.php
Ig0r-M: that's why i did not dare to edit properly this function as it has a broader area of impact, do you want for me to revert these changes ?

app/code/core/Mage/Catalog/Model/Indexer/Url.php
colinmollenhour: Yes, that would be great! Thanks

app/code/core/Mage/Index/Model/Event.php
colinmollenhour: Unless you have a lot of time to spend improving the method then that would probably be the best way forward. While the old behavior is not optimal it can still yield a correct outcome given your other patch using array_unique, from what I understand.

I'd love to know more about what issue this PR fixes though.

Ig0r-M: current spot is in regard with reindexing category url which added duplication thus making extra cycle for already parsed categories.

As for other, will try to take a look and see if i can improve core function, which i am scared to add a fix for :fearful:

Ig0r-M:
https://github.com/OpenMage/magento-lts/pull/582 Raised to address url reindexing for product and category.

app/code/core/Mage/Catalog/Model/Indexer/Url.php
Ig0r-M: Raised https://github.com/OpenMage/magento-lts/pull/582 for that

Ig0r-M:
Will close now this PR.

PR #572, opened 2018-11-27
getAttributeRawValue return false if value is '0'

bastienlm:
Something I think crazy
When you call 
`Mage::getResourceModel('catalog/product')->getAttributeRawValue($productId, 'attribute', $storeId);`
if the value is 0 in database (the real value), this function return FALSE (boolean) :interrobang: 

Yes, for this function empty value and 0 return the same thing 'FALSE'.
I think It's more logic to return 0 if value is 0 (string or numeric). I know it's not a little update and can be affected many thing, but... 0 != false... No ?! :)


Flyingmana:
huh, thats a risky field to act on.

```
php > var_dump(false == '0');
bool(true)
```

which then still leads to return false, and its returning false for null and `""`

In the end it still seems to lead to the expected result for any cases I can imagine. But I wonder if we should handle also the empty string case to return an empty string?
What does the database abstraction layer return in general?

This may also have a minimal impact on performance for some usecases.

colinmollenhour:
I think the problem is with the fact that one return statement deals with two cases: single attribute and multiple attributes. It is better to handle each separately.

```php
        if (sizeof($attributesData) == 1) {
            $_data = each($attributesData);
            return $_data[1];
        }
        return $attributesData ? $attributesData : false;
```

bastienlm:
One simple example for understand the problems :
If you have one product with 'attribute_1' = 0;
`$product->getAttribute1()` => 0
`Mage::getResourceModel('catalog/product')->getAttributeRawValue($productId, 'attribute_1', $storeId);` => false

with getAttributeRawValue you can't know if the real value is empty or 0. I think it's not a problem for the Core of Magento, but for custom devlopment (in my case ^^, for example, you need different process if value is empty or if value is 0) this 'issue' can cause trouble.

bastienlm:
> I think the problem is with the fact that one return statement deals with two cases: single attribute and multiple attributes. It is better to handle each separately.
> 
> ```
>         if (sizeof($attributesData) == 1) {
>             $_data = each($attributesData);
>             return $_data[1];
>         }
>         return $attributesData ? $attributesData : false;
> ```

More lite :)  better !

sreichel:
>  I think it's not a problem for the Core of Magento, but for custom devlopment (in my case ^^, for example, you need different process if value is empty or if value is 0) this 'issue' can cause trouble.

Could reproduce with one of my modules ... I do nut use 0 value so it doesnt appeared.

---

Bit oftopic ... 

Should we care about impact on other people code? If someone expects that `0` is treated as `false` it could break his code ... 

I agree with a fix, but we should document it. (and maybe add unit tests for it).

luigifab (approved):

colinmollenhour (approved):

sreichel:
With 1.9.4.4 (after this PR) code changed from ...

```
        if (sizeof($attributesData) == 1) {
            $_data = each($attributesData);
            $attributesData = $_data[1];
        }
```

to

```
        if (sizeof($attributesData) == 1 && !$returnArray) {
            $_data = reset($attributesData);
            $attributesData = $_data;
        }
```

@colinmollenhour did you verify that this change is still needed?

kiatng (approved):
Tested. It now 

> Retrieve attribute's raw value from DB.

as originally intended. But note that int 0 becomes "0". It return `null` if the value in DB is `null`.

tmotyl:
I consider it a bug and had to work around it in my projects several times.

luigifab (approved):

fballiano (approved):
tested, I don't know how we lived without this until now

fballiano:
I'm merging it and opening a new PR to document it in the README

fballiano:
cherry-picked to v20 with no conflicts

PR #574, opened 2018-11-28
Mark indexProcess as STATUS_REQUIRE_REINDEX; it is cleared after

seansan:
Mark indexProcess as STATUS_REQUIRE_REINDEX

it is cleared directly after if the indexer can run

$indexProcess->changeStatus(Mage_Index_Model_Process::STATUS_REQUIRE_REINDEX);

if it cannot then at least the indexer is marked as dirty

(sometimes the indexer is already running and the process cannot start; in this case there is no way to see this for the admin or some watching process)

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Rule.php
+++ b/app/code/core/Mage/CatalogRule/Model/Rule.php
@@ -318,6 +318,7 @@ class Mage_CatalogRule_Model_Rule extends Mage_Rule_Model_Abstract
         $this->_getResource()->applyAllRules();
         $this->_invalidateCache();
         $indexProcess = Mage::getSingleton('index/indexer')->getProcessByCode('catalog_product_price');
+        $indexProcess->changeStatus(Mage_Index_Model_Process::STATUS_REQUIRE_REINDEX);
         if ($indexProcess) {
             $indexProcess->reindexAll();
         }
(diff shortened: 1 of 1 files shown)


colinmollenhour (dismissed):
Looks like a great suggestion to me!

Also, shouldn't the cache be invalidated *after* the reindex is done rather than before?

seansan:
To my understanding we are setting reindex status - rather than cache
status? Or am I misreading your comment

colinmollenhour:
Sorry, two lines up it calls `$this->_invalidateCache();`. It is a separate issue but it seems to me invalidating the cache before the indexes are updated is useless since the cache will potentially be re-warmed with the old data before the indexing is complete.

kkrieger85 (dismissed):

app/code/core/Mage/CatalogRule/Model/Rule.php
```diff
@@ -318,6 +318,7 @@ public function applyAll()
         $this->_getResource()->applyAllRules();
         $this->_invalidateCache();
         $indexProcess = Mage::getSingleton('index/indexer')->getProcessByCode('catalog_product_price');
+        $indexProcess->changeStatus(Mage_Index_Model_Process::STATUS_REQUIRE_REINDEX);
```
tmotyl: getProcessByCode('catalog_product_price') can return `false` so in order to avoid fatal error we need to call this method inside if statement

tmotyl:
@seansan can you point me to the place where the flag is removed?

Flyingmana (approved):

PR #578, opened 2018-11-29
Sales_Order_Grids > Search By Customer First and Last Name Fixes

kestraly:
Orignal pull request ticket - Unable to search customer by first and last name on sales_order_grid #538

Apologies, had to reset my fork after I'd made the commits to the wrong branch. So just cleared and started again.

As said in previous PR, this fixes searching both customer first and last name.
Other notes can be found https://github.com/OpenMage/magento-lts/pull/538

### DIFF
--- a/app/code/core/Mage/Sales/Model/Resource/Order.php
+++ b/app/code/core/Mage/Sales/Model/Resource/Order.php
@@ -89,15 +89,14 @@ class Mage_Sales_Model_Resource_Order extends Mage_Sales_Model_Resource_Order_Ab
         $adapter       = $this->getReadConnection();
         $ifnullFirst   = $adapter->getIfNullSql('{{table}}.firstname', $adapter->quote(''));
         $ifnullMiddle  = $adapter->getIfNullSql('{{table}}.middlename', $adapter->quote(''));
-        $middelnameEnd = $adapter->getCheckSql('{{table}}.middlename IS NULL OR {{table}}.middlename="" OR {{table}}.middlename=" "', '""', '" "');
         $ifnullLast    = $adapter->getIfNullSql('{{table}}.lastname', $adapter->quote(''));
-        $concatAddress = $adapter->getConcatSql(array(
+        $concatAddress = trim(replace($adapter->getConcatSql(array(
             $ifnullFirst,
             $adapter->quote(' '),
             $ifnullMiddle,
-            $middelnameEnd,
+            $adapter->quote(' '),
             $ifnullLast
-        ));
+        )),'  ', ' '));
         $this->addVirtualGridColumn(
                 'billing_name',
                 'sales/order_address',
--- a/app/code/core/Mage/Sales/Model/Resource/Order.php
+++ b/app/code/core/Mage/Sales/Model/Resource/Order.php
@@ -89,12 +89,13 @@ class Mage_Sales_Model_Resource_Order extends Mage_Sales_Model_Resource_Order_Ab
         $adapter       = $this->getReadConnection();
         $ifnullFirst   = $adapter->getIfNullSql('{{table}}.firstname', $adapter->quote(''));
         $ifnullMiddle  = $adapter->getIfNullSql('{{table}}.middlename', $adapter->quote(''));
+        $middelnameEnd = $adapter->getCheckSql('{{table}}.middlename IS NULL OR {{table}}.middlename="" OR {{table}}.middlename=" "', '""', '" "');
         $ifnullLast    = $adapter->getIfNullSql('{{table}}.lastname', $adapter->quote(''));
         $concatAddress = $adapter->getConcatSql(array(
             $ifnullFirst,
             $adapter->quote(' '),
             $ifnullMiddle,
-            $adapter->quote(' '),
+            $middelnameEnd,
             $ifnullLast
         ));
         $this->addVirtualGridColumn(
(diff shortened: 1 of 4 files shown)


colinmollenhour:
What about just wrapping the concat in trim and replace two spaces with one?

```
trim(replace(concat('first', ' ', 'middle', ' ', 'last'),'  ', ' '))
```

That way if any of the fields are empty or a space there is no extra space and all of the 'if's can't be removed.

colinmollenhour:
This is what I had in mind. (not tested but should work)

```php
$concatAddress = $adapter->getConcatSql(array(
            $ifnullFirst,
            $adapter->quote(' '),
            $ifnullMiddle,
            $adapter->quote(' '),
            $ifnullLast
        ));
$concatAddress = new Zend_Db_Expr("TRIM(REPLACE($concatAddress,'  ', ' '))");
```

colinmollenhour:
I merged this commit into 1.9.4.x as well.

ADDISON74:
In OpenMage v20.0.6 this fix created more issues. Here is my report #1414.

For the moment I will revert the code to Magento vanilla. I prefer using two spaces but getting a complete filtered list of names than getting only a part of it.

Hope someone will find a solution once and for all to this bug.

sreichel:
@ADDISON74 why is this related to #1414?

ADDISON74:
@sreichel: after removing these changes I was able to filter both columns correctly even I put two spaces between words.

sreichel:
@ADDISON74 understood, but #1414 is something about broken wishlist.

ADDISON74:
You are right. I am sorry I mentioned a wrong issue number. I updated my initial post.

ADDISON74:
I am investigating this pull request. There are 4 files involved:

/app/code/core/Mage/Sales/Model/Resource/Order.php
/app/code/core/Mage/Sales/Model/Resource/Order/Creditmemo.php
/app/code/core/Mage/Sales/Model/Resource/Order/Invoice.php
/app/code/core/Mage/Sales/Model/Resource/Order/Shipment.php

In all these files there is a protected function called **_initVirtualGridColumns()**. I commented this function then I reloaded the pages in Backend (Orders, Invoices, Shipments, Credit Memo) . I did not see any changes in pages content, filtering the columns for Bill to Name / Ship to Name returns the same results.

ADDISON74:
This pull request was merged without testing different scenarios to prevent all issues. In Magento a complete name could have 3 parts: First, Middle and Last names. 

1. What happens if a store is not using Middle name in forms when a visitor registers as a customer or places orders? 
2. What happens if a part has more than one word?

I tested the following situation:

First Name: Joseph Robinette
Middle Name: 
Last Name:  Biden

If I filter the "Billing to Name" column for ```Joseph Robinette``` I get all results. Now if I filter for complete name ```Joseph Robinette Biden``` I get no results. By inserting one more space between Robinette and Biden I get again all results.

As a personal opinion we should pay attention to that TRIM function and how it deals with spaces.

kiatng:
Names are complicated because different cultures have different ways of naming their children. I don't have a solution but based on my actual encounters:

1. Some Indonesians have only one-word name, e.g. Jessica. It's true, I didn't believe at first and thought it was a mistake until I saw their passports.
2. On the other end of the spectrum, some names have many words, e.g. WASANA BUDDHI MADUBASHINI BALASOORIYA BALASOORIYA MUDIYANSELAGE
3. Most people in the world do not have middle name.
4. A large percentage of names begin with surname (family name) follows by given name. So _first name_ has different meaning, it is not given name but surname, and _last name_ is the given name and not surname.

Because the project deals with nationalities and cultures from around the world,  I needed to add a field called `fullname` in different tables and forms to get around the problems. I also needed to parse the fullname into `firstname` and `lastname` in `Mage_Customer` to avoid Mage throwing errors, but they aren't used.

ADDISON74:
When this change is implemented by an OM upgrade all records in the **bill_name** and **shipping_name** columns will remain in their original form with a double space between the names. My opinion that a sql upgrade should be created to modify these records too not only the php files. 

I did it quickly directly in the database with the help of the SQL script bellow and I didn't limit myself to just two spaces between the names, there can be even more. **In this way not only the recordings after the upgrade are found correctly, but all of them!**

```
UPDATE `sales_flat_creditmemo_grid` 
SET `billing_name`= REGEXP_REPLACE(billing_name, '[[:space:]]+', ' ');

UPDATE `sales_flat_invoice_grid` 
SET `billing_name`= REGEXP_REPLACE(billing_name, '[[:space:]]+', ' ');

UPDATE `sales_flat_order_grid` 
SET `shipping_name`= REGEXP_REPLACE(shipping_name, '[[:space:]]+', ' ');

UPDATE `sales_flat_order_grid` 
SET `billing_name`= REGEXP_REPLACE(billing_name, '[[:space:]]+', ' ');

UPDATE `sales_flat_shipment_grid` 
SET `shipping_name`= REGEXP_REPLACE(shipping_name, '[[:space:]]+', ' ');
```

colinmollenhour:
@ADDISON74 `REGEXP_REPLACE` sure is handy but unfortunately it requires MySQL 8.0..

Flyingmana:
I dont think the framework should manipulate existing user or even order data.
We can note this query in our documentation, if people want to cleanup their data. But it should not happen as part of an OM update. As we see, it happens that we sometimes miss details. We can revert the code, but if we also mess up the database, we cant fix it anymore with another update.
Besides, we dont know what side effects this could have with 3th party systems.

ADDISON74:
This PR has solved an important issue when the middle name is missing and which is inherited from Magento. But once merged in OpenMage it should be explained that the filtering of the columns in the grid will be done correctly only for the data collected after its integration in production. For the previous ones, either an extra space is used or an SQL command should be applied.

@colinmollenhour - indeed it works with MySQL 8 but the command can be adapted for previous versions. Unfortunately some forms in Frontend do not trim the data and then you can have more space. It can be used REPLACE command and search for 2 or more spaces that are replaced with a single space.

colinmollenhour:
Perhaps it should go in a shell script like `shell/maintenance.php`?

ADDISON74:
@colinmollenhour - It is a a good approach to run a script from that location which solves this issue in database. I guess many have had to deal with this problem in Magento but have not wondered why no results appear even though the names were entered correctly in the filter input box.

PR #581, opened 2018-12-07
Add Duplicate Catalog Rule Button and Action

seansan:
Add Duplicate Catalog Rule Button and Action

Please excuse my coding. But it works fine. And was very needed over here.

Edit it directly to improve to coding standards.


seansan:
Maybe someone can do the same for pricerules ;)

app/code/core/Mage/Adminhtml/controllers/Promo/CatalogController.php
```diff
@@ -308,6 +308,38 @@ public function applyRulesAction()
...
 
+     /**
+     * Duplicate catalog rule and edit
+     */
+    public function duplicateAction()
+    {
+        $errorMessage = Mage::helper('catalogrule')->__('Cannot copy rule ');
+        $params = $this->getRequest()->getParams();
+        try {
+            if($model = Mage::getModel('catalogrule/rule')->load($params['id'])) {
```
colinmollenhour: The return value of `load()` is an object so always truthy. The assignment should go before the `if` clause and then `getId` should be used to see if the load was successful.

app/code/core/Mage/Adminhtml/controllers/Promo/CatalogController.php
```diff
@@ -308,6 +308,38 @@ public function applyRulesAction()
...
+                $_currentrule = $model->getData();
+                $_newrule = Mage::getModel('catalogrule/rule');
+                $_newrule->setData($_currentrule);
+                unset($_newrule['rule_id']);
+                $_newrule['is_active'] = 0;
+                $_newrule['name'] = $_currentrule['name'] . " (Copy)";
+                $_newrule['identifier'] = $_currentrule['identifier'] . "-copy";
+                unset($_currentrule);
+                $_newrule->save();
```
colinmollenhour: The `$_currentrule` and `$_newrule` variable names are not consistent with Magento conventions of using camelCase and not adding underscore prefixes to variable names (these serve no purpose inside class methods).

While `Varien_Object` does support ArrayAccess it's use is very limited by convention so I would prefer to see the getters/setters used instead.

app/code/core/Mage/Adminhtml/controllers/Promo/CatalogController.php
```diff
@@ -308,6 +308,38 @@ public function applyRulesAction()
...
+                $this->_getSession()->addSuccess(Mage::helper('catalogrule')->__('The rule has been duplicated. Create a new Rule Name and Save to apply changes.'));
+                $this->_redirect('*/*/edit', array('id' => $_newrule->getId(), '_current' => true));
```
colinmollenhour: I think you're going about this the wrong way.. Rather than duplicate and edit, I think it is best to just preload the form data from an existing rule so that the rule is not created until it is actually saved. In the case of rules especially they can cause quite a bit of work by indexers and such when they are saved, cache invalidations, etc. To do this you would probably copy or modify the `editAction`.

seansan:
Thanks. Though I can make changes using this method ... I see another one
is suggested.

I will leave it for now .. maybe to someone else to create the code the
right way?

sreichel:
Adding butons with custom controllers can be don w/o change or rewrites ....

PR #582, opened 2018-12-07
Remove duplication during url indexer event process

Ig0r-M:
Following https://github.com/OpenMage/magento-lts/pull/570 which explains that function `\Mage_Index_Model_Event::_mergeNewDataRecursive` created duplicates, thus creating unnecessary cycles for url event reindex


Ig0r-M:
Changed base from `OpenMage:1.9.3.x` to `OpenMage:1.9.4.x`

Ig0r-M:
Now worries, my concern is that base method. Yep your updates makes sense

PR #583, opened 2018-12-17
fixed return value of saveCache method

sprankhub:
The original method did not return a boolean value, but the return value of `Mage_Core_Model_App::saveCache`. See https://github.com/OpenMage/magento-mirror/blob/fe6a94e281cb3e8e166f4eece31c4df1faa8cec6/app/code/core/Mage/Core/Model/Layout/Update.php#L198.


tmotyl:
@colinmollenhour the other change was yours, so please review it. For me it looks ok.

colinmollenhour:
This method is only ever called once and the return value is ignored in that case. Also the `MCMApp#saveCache` method just returns `$this` so it will always be the same value and has no special significance other than it is not "false" which is returned when the cache is disabled. So I don't really see a reason to change it..

sprankhub:
There may be others depending on that return value. Changing this just does not make any sense IMHO. It is just not proper OOP to completely change a return value (even the type!).

PR #589, opened 2019-01-15
Fix HTTP/2 errors with cURL post

wpdevteam:
Latest cURL update defaults to HTTP/2, but Magento's HTTP CURL Adapter is incompatible. This patch is necessary for newer server environments.

### DIFF
--- a/lib/Varien/Http/Adapter/Curl.php
+++ b/lib/Varien/Http/Adapter/Curl.php
@@ -180,7 +180,8 @@ class Varien_Http_Adapter_Curl implements Zend_Http_Client_Adapter_Interface
         $options = array(
             CURLOPT_URL                     => $url,
             CURLOPT_RETURNTRANSFER          => true,
-            CURLOPT_HEADER                  => $header
+            CURLOPT_HEADER                  => $header,
+            CURLOPT_HTTP_VERSION            => CURL_HTTP_VERSION_1_1
         );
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Can you describe how it is incompatible? Perhaps it could be made compatible rather than disabling it?

wpdevteam:
For cURL requests, there's no need to use HTTP/2, which would cost a lot more for a single request.
And this function call does not use input param $http_ver, which is a bug.
To make it better, perhaps can base on input param to set proper protocol.

LeeSaferite:
@wpdevteam How is Magento's HTTP curl adapter incompatible with curl selecting HTTP/2?

wpdevteam:
@LeeSaferite I think the main issue is that input param '$http_ver' in 'public function write' of 'lib/Varien/Http/Adapter/Curl.php' does nothing and 'CURL_HTTP_VERSION_1_1' should be the default header. It has better performance and compatibility if defaulted this way.

colinmollenhour:
Oops, merged into 1.9.3.x.. Just cherry picked into 1.9.4.x.

Thanks for the PR @wpdevteam!

simbus82:
Nice to see that i have solved this same problem in the same manner in a plugin that connect Magento to Joomla :-)
https://github.com/simbus82/MageBridgeCore/commit/f95b3fb794563523e6064ff27f00b390ed13e18d

PR #590, opened 2019-01-15
Fix issues with json_encode() and floats after upgrading to PHP 7.1

dariuszmaciejewski:
Issue description:
https://bugs.php.net/bug.php?id=72567

More info:
https://wiki.php.net/rfc/precise_float_value

Workaround in Magento 2:
https://github.com/magento/magento2/blob/2.3-develop/app/bootstrap.php#L74

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -32,3 +32,8 @@
 if (function_exists('libxml_disable_entity_loader')) {
     libxml_disable_entity_loader(false);
 }
+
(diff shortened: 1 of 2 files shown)


colinmollenhour:
What issue does this fix? One comment on the RFC states:

> Instead of changing serialize_precision to 14, it is recommended to leave it at the new default value of -1.

It seems setting it to 14 circumvents the purpose for the RFC which was to allow serialized floats to maintain their full precision rather than them being unnecessarily truncated.

PR #594, opened 2019-01-25
Remove whitespace in addBodyClass($className)

AndresInSpace:
Remove whitespace that is generated in body class string. 

if no result from $this->getBodyClass() output just classname instead of null+whitespace+classname
E.g. 
`<body class=" classnames">`

Fixed Result:
`<body class="classnames">`


kiatng:
Files affected by this commit:

in [app/design/frontend/rwd/default/template/page](https://github.com/OpenMage/magento-lts/tree/1.9.4.x/app/design/frontend/rwd/default/template/page)
print.phtml
popup.phtml

in [app/design/frontend/base/default/template/page](https://github.com/OpenMage/magento-lts/tree/1.9.4.x/app/design/frontend/base/default/template/page)
print.phtml
popup.phtml

PR #599, opened 2019-02-04
Allow 3rd-party applications to operate in get.php scope

ericseanturner:
Sometimes it might be useful to allow third-party modules to operate in the get.php scope. To accommodate this, a configuration point called 'loaded_modules' is added, into which third-party modules may register, thereby allowing their classes to be loaded during requests to get.php.


LeeSaferite:
I'm not sure I get what this is meant to allow. Could you explain it to me?

colinmollenhour:
get.php avoids loading the full Magento environment by caching a few key pieces of information in `resource_config.json`. When this file is present it calls `Mage::init` with the third and fourth arguments so only `Mage_Core` is loaded and not the gazillion other extensions. However, some third-party extensions might need to be loaded in order to override core functionality. An example is `Thai_S3` which we've been working on integrating. When the media file is missing from the local media directory and it must fetch the file off of S3 it needs to do so using the Thai_S3 rewritten classes but since Magento is initialized with only Mage_Core the `Thai/S3/etc/config.xml` is never loaded so the rewrites are not in effect. This PR makes the list of extensions which are necessary for get.php to work to be configurable. The list of modules to load is persisted in `resource_config.json` and then used to inform the partial-init.

colinmollenhour:
Also cherry-picked onto 1.9.4.x

PR #605, opened 2019-02-15
surpress array to string conversion

seansan:
surpress array to string conversion

not sure about all the use cases here but (string) would only affect FALSE to be empty string ""

all other variables woould be something else than an empty string

feedback welcome

### DIFF
--- a/app/code/core/Mage/Catalog/Block/Product/View/Attributes.php
+++ b/app/code/core/Mage/Catalog/Block/Product/View/Attributes.php
@@ -63,7 +63,7 @@ class Mage_Catalog_Block_Product_View_Attributes extends Mage_Core_Block_Templat
 
                 if (!$product->hasData($attribute->getAttributeCode())) {
                     $value = Mage::helper('catalog')->__('N/A');
-                } elseif ((string)$value == '') {
+                } elseif ($value == false || $value === '') {
                     $value = Mage::helper('catalog')->__('No');
                 } elseif ($attribute->getFrontendInput() == 'price' && is_string($value)) {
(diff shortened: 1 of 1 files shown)


colinmollenhour:
So an array value is appearing as "No"? What about `null`? Perhaps just adding ` && ! is_array($value)` would be safer?

seansan:
Hi there yes null should be included

So in the end the "No" clause should be for Null, False and Empty string

  echo "'".(string)null ."' <br/>";
  echo  "'".(string) array(1,2) ."' <br/>";
  echo  "'".(string)0 ."' <br/>";
  echo  "'".(string)1 ."' <br/>";
  echo  "'".(string)false ."' <br/>";
  echo  "'".(string)true ."' <br/>";

colinmollenhour:
Now it seems both array and 0 is affected. Should 0 not be "No"?

seansan:
dont know

Casting string on zero (0) would end up in "0"

Which can be correct, in like: number of buttons = 0

(and then you dont want No)

colinmollenhour:
Int zero is loosely equal to the empty string, but I suppose that int 0 should never really happen since PDO is set to return everything as strings. LGTM

LeeSaferite:
I tested this with the following:

```
<?php
$values = [null, true, false, 0, 1, "0", "1", "", []];
foreach($values as $value) {
    echo "Checking with: " . json_encode($value) . " ";
    if((@(string)$value == '') === (is_null($value) || $value === false || $value === '')) {
        echo "Pass\n";
    } else { 
        echo "Fail!\n";
    }
}
```

colinmollenhour:
I merged this commit into 1.9.4.x as well.

PR #607, opened 2019-02-18
Get cookie restriction block for current store

julienloizelet:
Issue : Magento EU Cookie law restriction block is not loaded per store id.

Fix found here : https://magento.stackexchange.com/questions/70931/magento-eu-cookie-law-restriction-block-not-loaded-per-store-id-cookie-restrict


julienloizelet:
Thanks,
Is it possible to merge it on 1.9.4.x too ?

PR #614, opened 2019-03-01
Fix #609 Add Captcha translate

### LINKED ISSUE #609: Untranslatable Mage_Captcha module
Sorry for not having time for a PR .... just found this on MSE ...

https://magento.stackexchange.com/questions/236088/magento-translation-not-working/263334#263334

Checked ... seems captcha module never was translatable.

seansan:
https://github.com/OpenMage/magento-lts/issues/609

### DIFF
--- a/app/code/core/Mage/Captcha/etc/config.xml
+++ b/app/code/core/Mage/Captcha/etc/config.xml
@@ -157,6 +157,15 @@
                 </captcha>
             </updates>
         </layout>
+        <translate>
+            <modules>
+                <captcha>
+                    <files>
+                        <default>Mage_Captcha.csv</default>
+                    </files>
+                </captcha>
+            </modules>
+        </translate>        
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Captcha/etc/config.xml
```diff
@@ -157,6 +157,15 @@
                 </captcha>
             </updates>
         </layout>
+        <translate>
+            <modules>
+                <captcha>
```
tmotyl: shouldn't this be `<Mage_Captcha>` ?

kkrieger85: Yes, you're right. 

Did not noticed this

sreichel:
@seansan Do do want to update?

Flyingmana:
after merge of #733 this should not be needed anymore

PR #625, opened 2019-03-22
Remove memory_limit in .htaccess

VincentMarmiesse:
Hello,
Is there any reason to let these `memory_limit` instruction in `.htaccess`?

It's very low and can't be overriden by web server configuration. The only solution is to edit this file but it get erased on every Magento upgrade.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -32,10 +32,8 @@
 <IfModule mod_php5.c>
 
 ############################################
-## adjust memory limit
+## adjust max execution time
 
-#    php_value memory_limit 64M
-    php_value memory_limit 256M
     php_value max_execution_time 18000
 
 ############################################
@@ -71,8 +69,6 @@
 ## see above for php_flag descriptions
 
 <IfModule mod_php7.c>
-    #php_value memory_limit 64M
-    php_value memory_limit 256M
     php_value max_execution_time 18000
     php_flag magic_quotes_gpc off
     php_flag session.auto_start off
--- a/.htaccess.sample
+++ b/.htaccess.sample
@@ -29,9 +29,8 @@
 <IfModule mod_php5.c>
 
 ############################################
-## adjust memory limit
+## adjust max execution time
 
-    php_value memory_limit 512M
     php_value max_execution_time 18000
 
 ############################################


tmotyl:
Hi
From my experience, 256MB memody limit is quite high. Maybe the issue is that some requests are using too much memory in the first place? Which requests do fail for you with 256MB memory limit?

VincentMarmiesse:
Hi @tmotyl,
We have a lot of cron tasks which perform operations on a quite big number of products (80k+).
Even the standard Magento export fails with this number of products.

Furthermore, [Magento recommands 512 MB](https://docs.magento.com/m1/ce/user_guide/magento/system-requirements.html). Maybe we should set it to 512 MB and not remove it?

edannenberg:
Maybe just comment it out so people that are on a managed hoster can easily enable this again. We use nginx but [this](https://github.com/OpenMage/magento-lts/blob/1.9.3.x/shell/abstract.php#L114) has bitten me more than once in the past, so yea :+1:  for removal.

LeeSaferite:
Well, your cron tasks shouldn't be using this value since it's only applicable to web requests.

That being said, I have no issue with raising the limit or removing it totally. You should be settling the limit in you INI anyway.

VincentMarmiesse:
> Maybe just comment it out so people that are on a managed hoster can easily enable this again.

It might be the best solution indeed, should I update my PR?

> That being said, I have no issue with raising the limit or removing it totally. You should be settling the limit in you INI anyway.

I agree, that's why this limit should not be here IMO.

edannenberg:
@LeeSaferite said:
> Well, your cron tasks shouldn't be using this value since it's only applicable to web requests.

Exactly, unfortunately Magento tries to be cute and if `.htaccess` exists *any* `php_value` in the file will be used for all scripts that extend `Mage_Shell_Abstract`.

Flyingmana:
Iam also ok with removing it. I remember especially for the Admin area it happens quite often to need more.
And CLI scripts need often more anyway

akira28:
I would also change Mage_Shell_Abstract to avoid using .htaccess to define php configs :P

colinmollenhour:
For frontend page requests 256M is a lot and if a page uses more than that there is some bug like loading every row in a large table without a where or limit clause. I'd rather have that page fail than consume all of the server's memory and cause things to start crashing (esp for servers running MySQL on the same machine).

I do agree however that for admin and CLI the 256M limit is sometimes too small. However in those cases the scripts should probably use `ini_set()` to set the appropriate limit. Having no limit at all is not a good idea in my opinion. I do that that for shell scripts the searching of .htaccess files for the memory limit should be removed.

So, I don't object to merging this but it is not the best solution IMO. Personally I don't recommend or use Apache anyway making .htaccess files moot. This is why I think it would be nice to somehow include a "reference" server setup using nginx and php-fpm based on Docker, but that is another subject entirely... :)

colinmollenhour:
This is missing the update to `.htaccess.sample`.

VincentMarmiesse:
> This is missing the update to `.htaccess.sample`.

Done.

PR #628, opened 2019-03-27
Load countries directory select by store

fplantinet:
Fix countries allowed in checkout 

Source: https://blog.fabian-blechschmidt.de/all-countries-in-checkout-although-allowed_countries-set-correctly/

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
@@ -46,6 +46,9 @@ class Mage_Catalog_Model_Product_Attribute_Source_Countryofmanufacture
             $options = unserialize($cache);
         } else {
             $collection = Mage::getModel('directory/country')->getResourceCollection();
+            if (!Mage::app()->getStore()->isAdmin()) {
+                $collection->loadByStore();
+            }
             $options = $collection->toOptionArray();
             if (Mage::app()->useCache('config')) {
                 Mage::app()->saveCache(serialize($options), $cacheKey, array('config'));


app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
```diff
@@ -45,10 +45,15 @@ public function getAllOptions()
...
+            if (Mage::app()->getStore()->isAdmin()) {
+                $collection = Mage::getModel('directory/country')->getResourceCollection();
+            } else {
+                $collection = Mage::getModel('directory/country')->getResourceCollection()->loadByStore();
+            }
             $options = $collection->toOptionArray();
+
             if (Mage::app()->useCache('config')) {
-                Mage::app()->saveCache(serialize($options), $cacheKey, array('config'));
+                Mage::app()->saveCache(serialize($options), $cacheKey, ['config']);
```
edannenberg: If this is a purely cosmetic change I would recommend removing it. Same for the introduced new line. Less changes = better.

app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
```diff
@@ -45,10 +45,15 @@ public function getAllOptions()
         if (Mage::app()->useCache('config') && $cache = Mage::app()->loadCache($cacheKey)) {
             $options = unserialize($cache);
         } else {
-            $collection = Mage::getModel('directory/country')->getResourceCollection();
+            if (Mage::app()->getStore()->isAdmin()) {
```
edannenberg: Refactor suggestion:

            $collection = Mage::getModel('directory/country')->getResourceCollection();
            if (!Mage::app()->getStore()->isAdmin())
                  $collection->loadByStore();

This way the patch becomes additive only.

app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
fplantinet: ok thanx!

app/code/core/Mage/Catalog/Model/Product/Attribute/Source/Countryofmanufacture.php
fplantinet: ok, i update commit ! thanx

PR #630, opened 2019-03-27
Fix types in doc comments

tmotyl:
Here is a bunch of fixes to types in doc comments (as well as few typos fixed).
This is a pat of previous PR, I've extracted here the ones which should be easy to review.


sreichel:
Can this be merged?

PR #633, opened 2019-03-28
Fix wrong types in doc comments part 2

tmotyl:


colinmollenhour:
Thanks for tackling this!

PR #639, opened 2019-03-29
Upstream merge 1.9.4.1

Flyingmana:
as there were issues with the last merge from the mirror, I do provide it now as a PR for an extra review


Flyingmana:
the bootstrap mismatch did already happen on the 1.9.4.0 import(I probably chose the wrong side for the conflict. So I would push this into a separate PullRequest afterwards

I merged #602 

I did forward port the current changes in 1.9.3.x to 1.9.4.x via #645

I remember looking into the wishlist observer thing and discussing it in a slack, sadly the history there is already lost, and it was not yet discussed here in the repository. The change was introduced by the PHP 7.2 patch, but was not included into official releases, so Iam not sure if it can be called a regression. Maybe, as it is about multiple wishlists, it was not meant to be for CE and just accidentally made it into the patch back then. (just writing it down here, so Iam not so confused about this topic the next time)
Am I right, that we dont need to do anything related to the wishlist?

Anything else I missed?

edannenberg:
> the bootstrap mismatch did already happen on the 1.9.4.0 import(I probably chose the wrong side for the conflict. So I would push this into a separate PullRequest afterwards

Yea looks like I missed that one when creating #602.

> I remember looking into the wishlist observer thing and discussing it in a slack, sadly the history there is already lost, and it was not yet discussed here in the repository. The change was introduced by the PHP 7.2 patch, but was not included into official releases, so Iam not sure if it can be called a regression.

I took another look at it, pretty sure this is a regression. The relevant PHP7.2 patch change:

```diff
-if (count($wishlistIds) && $request->getParam('wishlist_next')){
+if (!empty($wishlistIds) && $request->getParam('wishlist_next')){
```

Then in 1.4.0.0 this changed to:

```diff
-if (!empty($wishlistIds) && $request->getParam('wishlist_next')){
+if (empty($wishlistIds) && $request->getParam('wishlist_next')){
```

> Am I right, that we dont need to do anything related to the wishlist?

The regression never made it into this repo when 1.9.4.0 was merged, so right now no further actions are needed.

> Anything else I missed?

Looked good otherwise from my end. You can rebase the PR branch and do another directory compare with my fork, if both match everything is in order.

app/Mage.php
```diff
@@ -806,16 +806,22 @@ public static function log($message, $level = null, $file = '', $forceLog = fals
...
 
         // Validate file extension before save. Allowed file extensions: log, txt, html, csv
-        if (!self::helper('log')->isLogFileExtensionValid($file)) {
+        $_allowedFileExtensions = explode(
+            ',',
+            (string) self::getConfig()->getNode('dev/log/allowedFileExtensions', Mage_Core_Model_Store::DEFAULT_CODE)
+        );
+        $logValidator = new Zend_Validate_File_Extension($_allowedFileExtensions);
+        $logDir = self::getBaseDir('var') . DS . 'log';
+        if (!$logValidator->isValid($logDir . DS . $file)) {
```
sreichel: Not sure about to change this ... see #646

edannenberg: Another patch another regression, le sigh. Should get addressed in a separate PR though.

edannenberg: After digging a bit into the issue the code is actually correct, for some reason Zend_Validate_File_Extension::isValid() also does a file check, which really is none of it's business.

sreichel: Not said its wrong. Finally it does nearly the same as the old check - if the file check is removed :)

PR #640, opened 2019-03-29
Suggestion to add some logging if in error

seansan:
Suggestion to add some logging if in error

https://community.magento.com/t5/Payments-PayPal/PayPal-response-hasn-t-required-fields-AMT-field-is-deprecated/td-p/51852

### DIFF
--- a/app/code/core/Mage/Paypal/Model/Api/Nvp.php
+++ b/app/code/core/Mage/Paypal/Model/Api/Nvp.php
@@ -1164,6 +1164,7 @@ class Mage_Paypal_Model_Api_Nvp extends Mage_Paypal_Model_Api_Abstract
         if (isset($this->_requiredResponseParams[$method])) {
             foreach ($this->_requiredResponseParams[$method] as $param) {
                 if (!isset($response[$param])) {
+                    Mage::log("Expected PayPal field not found in NVP Response: $param");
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Paypal/Model/Api/Nvp.php
```diff
@@ -1164,6 +1164,7 @@ protected function _validateResponse($method, $response)
         if (isset($this->_requiredResponseParams[$method])) {
             foreach ($this->_requiredResponseParams[$method] as $param) {
                 if (!isset($response[$param])) {
+                    Mage::log("Expected PayPal field not found in NVP Response: $param");
```
tmotyl: Please log more context, right now the log is not really useful. You see that sth is wrong but has no idea which order was it and which paypal method caused the issue.

seansan:
According to OP this works fine. Also I am glad "at least" there is some
log.
Please add to the pull request if there is more info you need that you are
missing

app/code/core/Mage/Paypal/Model/Api/Nvp.php
Flyingmana: do we have any order identifying data available in this context?

Flyingmana (approved):

Issue #646, opened 2019-04-02
Mage::log() not working on new Magento update (1.9.4.1)


sreichel:
Log seems broken, when log files does not exist .... https://magento.stackexchange.com/a/268396/46249

Hope for next official update?

Flyingmana:
Iam aware of this Issue, I guess we will accept a community provided fix for this Issue after the merge

themew:
Fixed here > https://gist.github.com/piotrekkaminski/0596cae2d25bf467edbd3d3f03ab9f8f
but nothing from Magento about a V2 patch.

seansan:
Hi, was this fixed maybe and pushed to 4.x? (having this probem too now)

themew:
> Hi, was this fixed maybe and pushed to 4.x? (having this probem too now)

The patch works or you can manually edit the app/Mage.php file

colinmollenhour:
Can someone else approve one or more of #648, #652 or #670? All three are correct in my opinion, but we need at least one of them to fix this issue.

colinmollenhour:
Merged #670.

aadmathijssen:
This issue has been fixed in CE 1.9.4.2 with exactly the same changes as in #670.

PR #648, opened 2019-04-02
Zend_Validate_File_Extension::isValid() fails if filename doesn't exist

edannenberg:
Fixes regression introduced in 1.9.4.1. See #646. The root cause has been in the Magento core since probably forever, the class is just barely used so it wasn't an issue until now.

The few places that actually use the class have their own file checks in place already, like [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Core/Model/Email/Template/Abstract.php#L242).

### DIFF
--- a/lib/Zend/Validate/File/ExcludeExtension.php
+++ b/lib/Zend/Validate/File/ExcludeExtension.php
@@ -60,12 +60,6 @@ class Zend_Validate_File_ExcludeExtension extends Zend_Validate_File_Extension
      */
     public function isValid($value, $file = null)
     {
-        // Is file readable ?
-        #require_once 'Zend/Loader.php';
-        if (!Zend_Loader::isReadable($value)) {
-            return $this->_throw($file, self::NOT_FOUND);
-        }
-
         if ($file !== null) {
             $info['extension'] = substr($file['name'], strrpos($file['name'], '.') + 1);
         } else {
--- a/lib/Zend/Validate/File/Extension.php
+++ b/lib/Zend/Validate/File/Extension.php
@@ -185,12 +185,6 @@ class Zend_Validate_File_Extension extends Zend_Validate_Abstract
      */
     public function isValid($value, $file = null)
     {
-        // Is file readable ?
-        #require_once 'Zend/Loader.php';
-        if (!Zend_Loader::isReadable($value)) {
-            return $this->_throw($file, self::NOT_FOUND);
-        }
-
         if ($file !== null) {
             $info['extension'] = substr($file['name'], strrpos($file['name'], '.') + 1);
         } else {


Flyingmana:
I vote against this change, as it changes the API of this function in a way, which creates an incompatibility against ZF2 which also checks if the file is readable

https://github.com/zendframework/zend-validator/blob/master/src/File/Extension.php

edannenberg:
@sreichel 
> It's used in all Zend File validator classes ...

Not really, it's only used by `Zend_Validate_File_ExcludeExtension`. All other references are wrong doc blocks: https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Zend/Validate/File/Exists.php#L103

@Flyingmana 
> I vote against this change, as it changes the API of this function in a way, which creates an incompatibility against ZF2 which also checks if the file is readable

Why do we care about ZF2? The API as it is now is broken. It makes no sense that you have to create the file first before you can check if it has a valid extension. If you look at the implementation in ZF2 it's pretty clear that it got blindly refactored without much thought. For example [File/Exists](https://github.com/zendframework/zend-validator/blob/64c33668e5fa2d39c6289a878f927ea2b0850c30/src/File/Exists.php#L103) still sports the same wrong doc-blocks from ZF1 even though it got refactored.

Flyingmana:
its in this case not only backwards, but also forward compatibility

Flyingmana:
To explain it a bit more. I care, because 3th party modules may use this, and it will change behaviour for them.
Also we increase incompatibility with Magento2, which is still using the ZF Validator to validate file extensions, so instead of making it easier to do a future switch, we make it harder, as someone cant make use of this for module parts which should work for both versions in the same way.

edannenberg:
> I care, because 3th party modules may use this, and it will change behaviour for them.

This would only change behaviour if someone relied on `Zend_Validate_File_Extension` to validate if a file exists, which is highly unlikely IMO.

> Also we increase incompatibility with Magento2, which is still using the ZF Validator to validate file extensions, so instead of making it easier to do a future switch, we make it harder, as someone cant make use of this for module parts which should work for both versions in the same way.

I would recommend to open a ticket/PR with ZF2 or Magento 2 then so this can get fixed. It's clearly a bug.

sreichel:
@edannenberg if I watch the same file again and again ... :) You're right.

Instead of removing this check ... how about check if `$file` param is set (`@param  array   $file  File data from Zend_File_Transfer`)?

        if ($file !== null && !Zend_Loader::isReadable($value)) {
            return $this->_throw($file, self::NOT_FOUND);
        }

chelevich:
In my project 
- I created `MyOwn_Validate_File_Extension`, 
- overridden `isValid` method
- called parent and if error was `NOT_FOUND` I created the file,
- then I re-validated the file with the second parent call
And I used it in `Mage::log()`
Here we can do the same in `lib/Varien/Validate/File/Extension.php`

cslogic:
At Mage.php on line 817, `$logValidator` is there to check if the file has one of `$_allowedFileExtensions`, which is ok. My bet is that there is nothing wrong with `Zend_Validate_File_Extension`, I just think it is not the right class to check the extension of an file that doesn't even exist. We gotta find another way to do that.

edannenberg:
@sreichel 
> Instead of removing this check ... how about check if $file param is set (@param array $file File data from Zend_File_Transfer)?

Good point but I would be surprised if `Zend_File_Transfer` doesn't handle this already so it would be kinda pointless to check again. Checking if a file exists is simply out of scope for this class as it only has a single job, check the file extension for a given **filename**, straight from the ZF1 api docs:

```
isValid (line 189)

Defined by Zend_Validate_Interface

Returns true if and only if the fileextension of $value is included in the set extension list.
```

colinmollenhour:
Given the context that this is an extremely simple class/method where the documentation clearly states what it was intended to do and that the code doesn't do what all other indicators say it *should* do, and this incorrect behavior is actually causing a serious bug that needs to be fixed, I think this BC break is acceptable. ZF is full of terrible code so this does not surprise me and seems pretty clearly to be a case where the author mistakenly thought that the file must exist for the `pathinfo` function to work. On top of that it is using `Zend_Loader` which is meant to be used for auto-loading classes and is dangerously close to methods like `Zend_Loader::loadFile` that would be used for RCE attacks.

What is the purpose of tracking ZF1 code closely with ZF2 code? ZF2 is hardly any better than ZF1, basically ZF1 with namespaces and it is basically a dead project itself.

sreichel:
@edannenberg I agree with "it is out of scope" in that file - so I approved first.

Maybe I'm wrong - just quickly searched - but it seeems to be used in `Mage_Catalog_Model_Product_Option_Type_File` ... but also there is a `is_readable($fileFullPath)` check.

However ... if we decide to remove it, it should be removed from `Zend_Validate_File_ExcludeExtension` too.

theroch:
@sreichel I confirm with you, removing is not the correct approach.
How it is with this solution?

```
if (file_exists($value) && !Zend_Loader::isReadable($value)) {
            return $this->_throw($file, self::NOT_FOUND);
        }
```
It solves the problem for me.

edannenberg:
@sreichel 
> Maybe I'm wrong - just quickly searched - but it seeems to be used in Mage_Catalog_Model_Product_Option_Type_File ... but also there is a is_readable($fileFullPath) check.

The 2 places in Magento core where `Zend_Validate_File_Extension` is used don't rely on the faulty behaviour to ensure that the file exists. As it should be.

> However ... if we decide to remove it, it should be removed from Zend_Validate_File_ExcludeExtension too.

Sure, we might as well. I amended the commit.

@theroch 
> @sreichel I confirm with you, removing is not the correct approach.
How it is with this solution?

It's a bit late for april's fools.. ;) Take a look at your patch, the if condition will never trigger, it's functionally equivalent to removing.

theroch:
@edannenberg That was not a joke. file_exists return true if the user has no read access to the file and is_readable return only true if the user has readd access or the file is not locked.

edannenberg:
> The file_exists funtion returns true if the file exists but the user has no read access to the file. is_readable returns only true if the user has read access or the file is not locked by another process.

Ok so it will virtually never trigger. What is accomplished in the rare case this actually does trigger? How is this relevant to a validator class who's sole purpose is to check a given string for valid file extensions?

sreichel:
@Flyingmana I understand your concerns ... but IMHO we need a quick fix here. Would it be OK to approve this and later change it, if there is a better solution?

theroch:
> Ok so it will virtually never trigger. What is accomplished in the rare case this actually does trigger? How is this relevant to a validator class who's sole purpose is to check a given string for valid file extensions?

It has nothing to do with checking the permitted file extensions. However, the function was probably implemented, since it should usually be possible to read a file.
The function is also used in Mage_Core_Model_Email_Template_Abstract, but there the is_readable function is checked explicitly.
So the part could also be removed from the isValid method.

colinmollenhour:
Alternatively we could just change the Mage code to check the file extensions using a non-Zend method since the Zend code seems to be the hang-up.. E.g.:

```php
if ( ! ($extension = pathinfo($file, PATHINFO_EXTENSION)) || ! in_array($extension, $_allowedFileExtensions)) {
    return;
}
```

I never was a fan of using all of these silly wrappers for such basic functionality.. Example: `Zend_Loader::isReadable()` vs `is_readable()`.

cslogic:
> Alternatively we could just change the Mage code to check the file extensions using a non-Zend method since the Zend code seems to be the hang-up..

That's exactly what I have meant earlier, and I still think that this is the best solution as it keep compatibility and fixes the problem...

sreichel:
> 
> 
> Alternatively we could just change the Mage code to check the file extensions using a non-Zend method since the Zend code seems to be the hang-up.. E.g.:
> 
> ```
> if ( ! ($extension = pathinfo($file, PATHINFO_EXTENSION)) || ! in_array($extension, $_allowedFileExtensions)) {
>     return;
> }
> ```
> 
> I never was a fan of using all of these silly wrappers for such basic functionality.. Example: `Zend_Loader::isReadable()` vs `is_readable()`.

This is excatly what tho old method has done ...

```
    /**
     * Checking if file extensions is allowed. If passed then return true.
     *
     * @param $file
     * @return bool
     */
    public function isLogFileExtensionValid($file)
    {
        $result = false;
        $validatedFileExtension = pathinfo($file, PATHINFO_EXTENSION);
        if ($validatedFileExtension && in_array($validatedFileExtension, $this->_allowedFileExtensions)) {
            $result = true;
        }

        return $result;
    }
```

May "fix" it with changing this method? Just have to get allowed extensions from `'dev/log/allowedFileExtensions'` config, instead of  a private var?

edannenberg:
I really don't get the BC worries for this one. BC is important but really shouldn't cater to support shoddy code. This PR fixes the root cause while also restoring the intended functionality of the class, preventing any future issues. Do we really want to declare `Zend_Validate_File_Extension` to be broken with `#wontfix` because there is a tiny probability that some code relies on undocumented buggy behavior? Even though all usage we have seen so far suggests that users of the class did not mistake this for a file exists validator?

Flyingmana:
just for clarification. Iam against it, but I will not block it. If you have the required 2 approvers, I will merge it.

But, BC is not a topic which depends on how much sense a behavior makes, or if we like it or not. 
Its about trust, trust that we do not break other peoples usecases, regardless how odd they may seem.
Its why there is semantic versioning.
There are reasons why some "bugs" in behavior are only fixed with minor or even major version increase. ZF2 may be already dead, but it is still used. Also remember the effort they had with replacing ZF1 with ZF2 in M2. Think how long it will take to get rid of ZF2
In my opinion this should either get fixed on Magento side, or by forking this class into an own namespace.
But as I said, I am not here as a blocker, just want to make sure you are aware of this and keep the longterm effect of this in mind.

sreichel:
>  just for clarification. Iam against it, but I will not block it. If you have the required 2 approvers, I will merge it. [...] But as I said, I am not here as a blocker, just want to make sure you are aware of this and keep the longterm effect of this in mind.

@Flyingmana Thanks. This is why i've asked for review .... personally I don't want to overrule any of the initators/longest-time members. So i'm not going to approve.

When a raise an issue, i normally do not ask if we should wait for Magento fixing it - it does not happen :) In the case im "pretty" sure it will be fixed in "some" way "soon" ... maybe with fixing ZF.

If you dont want to change ZF, do the most simple for now and revert this part of the patch or "adjust" `isLogFileExtensionValid()`?

edannenberg:
> But, BC is not a topic which depends on how much sense a behavior makes, or if we like it or not.
Its about trust, trust that we do not break other peoples usecases, regardless how odd they may seem.

I'm not sure how we can fix anything in this repo then, for example #525 should have never get merged then because there might be a poor soul that still used it.. I understand the importance of BC and have to deal with it on a daily basis, blindly following that rule isn't the way to go about it. Instead we should look at each BC issue indiviually and assess the actual impact. Theoretical or buggy 3rd party code BC issues should never prevent valid use cases / bug fixes unless the expected impact outweights the benefits.

piotrekkaminski:
This is the current official patch that will be ported to earlier versions (this should work on latest) https://gist.github.com/piotrekkaminski/0596cae2d25bf467edbd3d3f03ab9f8f

royduin:
Did anyone contact Magento about this? What did they say? Do they confirm this is a bug and are they going to fix this?

FredericMartinez:
@piotrekkaminski It works on production with 1.9.4.1 LTS, thanks!

BassemN:
@piotrekkaminski Thanks a lot :+1:

BassemN:
> This is the current official patch that will be ported to earlier versions (this should work on latest) https://gist.github.com/piotrekkaminski/0596cae2d25bf467edbd3d3f03ab9f8f

@piotrekkaminski It works on production with 1.9.4.1 LTS, thanks!

sreichel:
Closed with #670.

edannenberg:
This is not fixed with #670. The bug in Zend_Validate_File_Extension still remains.

kkrieger85:
@edannenberg On which version? 1.9.?.?

edannenberg:
@kkrieger85 I'm not saying that #670 isn't fixed, it worked around the issue by not using `Zend_Validate_File_Extension` at all.

FYI: I asked on ZF's Slack a while ago and it was acknowledged as a bug, however the PR hasn't been processed yet. It does indeed look like a dead project.

kkrieger85:
@edannenberg Could you reference this PR in this issue/PR?

aadmathijssen:
Yesterday, Magento released CE 1.9.4.2, which contains the fix that has been implemented in #670.

fballiano:
is this still needed guys?

ADDISON74:
**OFFTOPIC**
@fballiano - is there a procedure for closing PR's and Issues in the event that a requested response is not received within a deadline? In other projects I found that if there are no answers then it is closed with the mention that it can be opened at any time at the request of the participants. In OpenMage have a lot of open posts and some are 5 years old. Even if we can't evaluate each one I think we can implement a rule to close them. They can be found, developed at any time, they do not disappear. A spring cleaning would be necessary after all this time. However, 300 issues are enough to show the inefficiency of a project. Personally I use those where the number of Issues/PR's are controllable and the posts are not older than 6 months, otherwise for me it is a dead project.

fballiano:
@ADDISON74 I agree with you. I'd not like to just start closing PR/issues, that's why I'm starting to ping the ones that don't have comments since a while and then close them if no feedback is provided in some reasonable time (like a couple of weeks?)

They can be reopened anytime if needed so I too think it's not that big of a deal :-)

But you're also right when you say we should have a proper protocol for that ad AFAIK there's no official rule.

ADDISON74:
Two weeks should be fine for waiting a requested reply. 

I would recommend two things: a public rule and a standard message from maintainers to be pasted in Issues/PR's.

fballiano:
I'm too against this patch, not for compatibility but because we need to stop changing Zend Framework's files. I take responsability for closing this since anyway it's dead since 3 years.

Issue #647, opened 2019-04-02
Wrong price displayed in cart when qty in cart is higher than stock.


Hexmage:
Since 1.9.3.0 a bug was introduced which may cause the cart to display the wrong price for a product. 

How to reproduce:
- Create a product that has managed stock and isn't backorderable
- Add product to cart.
- Add the product to your cart until its qty exceeds its stock.

Expected result:
- Product price is still the original price and subtotal is the total for all products in the cart. You can't go to the checkout.

Actual result:
- Product price is significantly lower than the original price (depends on how far you went over the stock). You can't go to the checkout.

The cause:
Mage_Sales_Model_Quote_Address_Total_Subtotal line 114
`} else if (!$quoteItem->getParentItem() && !$item->getHasError()) {`
Because `&& !$item->getHasError()` got added the product price doesn't update anymore after it has gained an error, like in this case not having enough stock. But it is using the not updated value to incorrectly calculate the new subtotal.

sreichel:
Closed with #650.

PR #650, opened 2019-04-03
painful protection in Subtotal.php

henrykbrzoska:
Linking the tax/prices calculation with that if quote item is 'wrong'(errored) item is pointless. nonsense
Protection through which I lost 4 hours of life.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote/Address/Total/Subtotal.php
+++ b/app/code/core/Mage/Sales/Model/Quote/Address/Total/Subtotal.php
@@ -112,7 +112,7 @@ class Mage_Sales_Model_Quote_Address_Total_Subtotal extends Mage_Sales_Model_Quo
             $item->setPrice($finalPrice)
                 ->setBaseOriginalPrice($finalPrice);
             $item->calcRowTotal();
-        } else if (!$quoteItem->getParentItem() && !$item->getHasError()) {
+        } else if (!$quoteItem->getParentItem()) {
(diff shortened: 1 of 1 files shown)


tmotyl:
Hi @henrykbrzoska 
Can you elaborate a little bit more about the issue youre fixing? 
What is the current state? What is the result after applying the pr?

henrykbrzoska:
Yes, ofc.
1.Quote item with 'error' isn't calculated. Why? I can't imagine reason.
2. But…Cart Rules still works.
3. Cart and taxes are ruined.

http://i.imgur.com/E6IZ2Kb.png
item for testing
Add 4+ pieces http://openmage.bssolutions.it/shea-enfused-hydrating-body-lotion.html

edannenberg:
Looks like this should fix #647.

Hexmage:
This will indeed fix #647, because it reverts the change that breaks in the first place.

henrykbrzoska:
yes.
If someone has the possibility of contact, it is probably Magento Team should find out about it.

Hexmage:
There is an issue about this on the Magento forums themselves. Which is about 2 years old by now.

henrykbrzoska:
If this error is explained and intended for improvement, then please do the next :)
>> https://github.com/OpenMage/magento-lts/pull/662
^^please delete if you consider spam

henrykbrzoska:
next problem with this sh*t
- when reorder an order with product which 'has error'(i.e. insufficient stock qty) => price = 0.

henrykbrzoska:
bump!

PR #651, opened 2019-04-04
Fix broken logging in 1.9.4.1, also see #648

sreichel:
This reverts that part of last patch that uses Zend_Validate_File_Extension. The new feature, that gets allowed extensions from system.xml has moved to `Mage_Log_Helper_Data::isLogFileExtensionValid().`

Ref #646 #648

### DIFF
--- a/api.php
+++ b/api.php
@@ -24,8 +24,8 @@
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-if (version_compare(phpversion(), '5.2.0', '<')) {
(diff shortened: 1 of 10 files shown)


cslogic:
Removing the allowed extensions verification wouldn't be a security problem?

sreichel:
It is not removed - just moved to `isLogFileExtensionValid()`.

Before patch allow extensions was a private property. Now it takes allowed extensions from config node as introduced with 1941. 

Benefit (?)

`Mage_Adminhtml_Model_System_Config_Backend_Filename::_beforeSave()` uses same mehtod and would now use extensions from confignodes a `log()` method does.

sreichel:
If this is okay, a make a new clean PR.

PR #653, opened 2019-04-06
Removed Magento compiler leftovers

cslogic:
I have found no use for the things I removed, so for me there is no reason to maintain them.


sreichel:
Looks like dead code?

Issue #654, opened 2019-04-09
PHP 7.3 - mcrypt_module_open - FUNCTION REMOVED


Sekiphp:
```
[Tue Apr 09 10:44:18.546069 2019] [php7:error] [pid 6418] [client 127.0.0.1:43814] PHP Fatal error:  
Uncaught Error: Call to undefined function mcrypt_module_open() in /var/www/developer/web/lib/Varien/Crypt/Mcrypt.php:63
```
*Link to PHP doc:* https://www.php.net/manual/en/function.mcrypt-module-open.php (removed in version 7.2.0)

*Possible solution:* https://stackoverflow.com/questions/41740600/php7-1-mcrypt-alternative

Sekiphp:
phpseclib

Issue #129, opened 2016-10-25
If You're Typing the Word MCRYPT Into Your PHP Code, You're Doing It Wrong


colinmollenhour:
https://paragonie.com/blog/2015/05/if-you-re-typing-word-mcrypt-into-your-code-you-re-doing-it-wrong

Mage_Core_Model_Encryption uses Varien_Crypt which implements mcrypt by default.
I propose the following:
- Add Varien_Crypt_Openssl to implement encryption via Openssl using whatever the best practices are for secure encryption.
- Read a config key global/crypt/method to get a specific method if specified. Default to mcrypt if not specified for backwards compatibility.
- Add global/crypt/method -> openssl to the local.xml.template so that for new installations the default is openssl. I don't really care to support platforms without openssl so I don't even know if one exists or if they are common but I think it is perfectly reasonable to require a supported encryption library for an ecommerce platform.. It can easily be added to list of modules required during install.
- Write a script to migrate encrypted data from old method/key to new method/key. (find all encrypted config values, decrypt, re-encrypt with new key/method and save over old local.xml)
  - Note, Magento CE had a major bug up until 1.8/1.9 whereby a VERY poor encryption key was generated at installation so without a migration script if you are using one of these keys you basically might as well not even be using any encryption it is so bad. I tried to raise this with Magento on a magento2 issue but my warnings were completely ignored since they already have such a migration script for EE apparently.

EDIT:
Oh yeah, mcrypt will be deprecated in PHP 7.1 and removed in the version after. More info: https://wiki.php.net/rfc/mcrypt-viking-funeral

Flyingmana:
👍 
And yes, there are platforms without openSSL out there, but we can sure add alternates then.

sreichel:
Reference ... https://github.com/Inchoo/Inchoo_PHP7/issues/98

inluxc:
This is a nice read about this problem.
https://www.zimuel.it/slides/phpce2017#/

colinmollenhour:
Just pushed PR for this issue. See #506

kkrieger85:
Close due to https://github.com/OpenMage/magento-lts/pull/506

seansan:
Should this also be changed? https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Zend/Filter/Encrypt.php#L85

colinmollenhour:
[...]

PR #655, opened 2019-04-09
Varien_Crypt_Mcrypt added PHP 7.3 support

Sekiphp:
Added support for PHP 7.3. Removed functions are deprecated as PHP 7.1 and removed in PHP 7.2.

More info in issue #654

### DIFF
--- a/lib/Varien/Crypt/Mcrypt.php
+++ b/lib/Varien/Crypt/Mcrypt.php
@@ -55,6 +55,23 @@ class Varien_Crypt_Mcrypt extends Varien_Crypt_Abstract
         }
     }
 
+	/**
+	 * Transition table
+	 */
+    protected function _getCipher()
+    {
+        $transition = [
+            (defined('MCRYPT_BLOWFISH') ? MCRYPT_BLOWFISH : 'blowfish') => 'bf-ecb',
+            // more constants need to be implement here
+        ];
+
+        if (!array_key_exists($this->getCipher(), $transition)) {
+            return NULL;
+        }
+
+        return $transition[$this->getCipher()];
+    }
+
     /**
      * Initialize mcrypt module
      *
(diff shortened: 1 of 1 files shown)


lib/Varien/Crypt/Mcrypt.php
```diff
@@ -55,6 +55,23 @@ public function destruct()
         }
     }
 
+	/**
+	 * Transition table
+	 */
+    protected function _getCipher()
+    {
+        $transition = [
+            (defined('MCRYPT_BLOWFISH') ? MCRYPT_BLOWFISH : 'blowfish') => 'bf-ecb',
```
Sekiphp: `ecb`: https://www.php.net/manual/en/mcrypt.constants.php

seansan:
@Sekiphp Can you explain how this works and why you closed this issue/pull?

We are having problems with CRYPT in M19.4.x LTS ... we tried the solution by @colinmollenhour https://github.com/OpenMage/magento-lts/issues/129 witht the title "If You're Typing the Word MCRYPT Into Your PHP Code, You're Doing It Wrong"

But the reality is ... people DO type the word mcrypt in their code and wont change it

So we've tried

- Nothing, upgraded to PhP 7.2 + LTS and got errors mainly in (popular!) extensions
- Implemented the fix in issue #129 but then the code obfuscation stops working in vendor code + paypal express stopped working (header encodign wrong)
- maybe try this solution?


Maybe I should type this into the issue itself for a better discussion there ... https://github.com/OpenMage/magento-lts/issues/129

But the long and the short is: we need to get this fixed. And Im not an expert in any coding/decoding algoritms ... ;(

What to do

colinmollenhour:
It is hard to pinpoint a specific example but I think this openssl method sometimes works and sometimes doesn't depending on the input and it has to do with the padding at the end of the string if I remember correctly.. It has been a while, though..

Issue #658, opened 2019-04-11
getAttributeRawValue() - when call with array of attrs unexpectable behavior


Sekiphp:
**Class:** Mage_Catalog_Model_Resource_Abstract

**Example call:**
```
$productData = Mage::getResourceModel('catalog/product')->getAttributeRawValue(
                    (int) $suborder->getProductId(),
                    ['package_height', 'package_width', 'package_length', 'weight'],
                    Mage_Core_Model_App::ADMIN_STORE_ID
                );
```

**Expected result:** indexed array
```
Array
(
    [weight] => 53.5000
    [package_height] => 0
    [package_width] => 0
    [package_length] => 0
)
```

**Given result:** `53.5000` (Because only one of given attributes exists for this product in admin store.)

colinmollenhour:
I thought something similar came up before? It seems the correct approach is to look at the original number of arguments to determine whether or not to return a scalar or an array, not the result for reasons such as this.

midlan:
@colinmollenhour I've created fix (pull request #664), where I look if the list of attributes is array. If yes, I disabled conversion to scalar.

Sekiphp:
Solved in pull request #664

PR #664, opened 2019-04-17
Fixed unexpectable behavior of getAttributeRawValue(), issue #658

midlan:


app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -584,8 +584,12 @@ public function getAttributeRawValue($entityId, $attribute, $store)
     {
         if (!$entityId || empty($attribute)) {
             return false;
-        }
+	}
```
colinmollenhour: Looks like you replaced spaces with a tab.

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -584,8 +584,12 @@ public function getAttributeRawValue($entityId, $attribute, $store)
...
+
+        $wasArray = true;
```
colinmollenhour: To better maintain BC I think `$wasArray` should also be false if the size of the `$attribute` array was 1. In this case perhaps it should be renamed to something like `$returnArray`.

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
midlan: my fault, removed

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
midlan: You are right, I didn't saw any code like that in Magento, but some 3rd party module maybe relying on it. Fixed that also.

PR #665, opened 2019-04-18
Make Adminhtml field comment translatable

rodrigoriome:
I'm on a mission of translating a lot of modules and I just noticed that the field comments (those messages that shows when you hover the "?" next to the field") ware not being translated. I think Magento tries to translate it somewhere but it's not working.


sreichel:
To enable translation for system config fields you have to edit modules `system.xml`.

Just add `translate="label comment tooltip"` to the config field you want translate. E.g.

```
<flat_catalog_product translate="label comment">
```

- label: field label
- comment: small text below input
- tooltip: questionmark

rodrigoriome:
My mistake, I confused my Magento instances and ended up fixing a third-party-extension problem. Won't happen again, sorry.

PR #671, opened 2019-05-17
Fix double load at customer edit page in administration

Sekiphp:
In customer edit are duplicated all requests in Google Chrome and Mozilla Firefox. In Internet Explorer this works correctly. This changes fix this issue.

For example - try behavior in different browsers:
```
<iframe src="">
<p>Your browser does not support iframes.</p>
</iframe>

<iframe src="#">
<p>Your browser does not support iframes.</p>
</iframe>
```


sreichel:
Can you please add steps to reproduce?

kkrieger85:
Bildschirmfoto 2019-06-01 um 14 16 48


In FF it looks like inception. :D

kkrieger85:
@Flyingmana your turn

PR #675, opened 2019-05-23
Mage_Admin - IDE improvements

sreichel:
Blueprint. Just tried to get rid of most IDE errors - especially "undfined classes" (IDE forwarding) -  ... and make it work with phpstan at the end ...

- added/fixed missing DOC blocks
- added @method annotations
- changed @return to $this for methods
- changed @return to $this for method annotations
- use @inheritDoc for methods that return parent call
- <s>added class check (instanceof) for those methods (instead adding @var SomeClass)</s>
- fixed get var before try/catch
- <s>changed typehint to concrete class instead of Mage_Core_Model_Abstract</s>
- <s>changed some helper calls to add @var annotation</s>
- flipped @var annotations 

 <sup>(1)</sup> temp?

`Mage_Admin` is just an example ... Any thoughts? :)


colinmollenhour:
Very nice!

Under what circumstance is `$installer->getConnection();` in the scope of the scripts that were updated not an instance of `Varien_Db_Adapter_Pdo_Mysql`? If none, perhaps `Varien_Db_Adapter_Pdo_Mysql` should be added to the `@return` vardoc for `getConnection` even though it doesn't necessarily return that but in Magento core code it always does in practice.

sreichel:
> Under what circumstance is `$installer->getConnection();` in the scope of the scripts that were updated not an instance of `Varien_Db_Adapter_Pdo_Mysql`?

Guess none. It just because `getConnection()` returns `Varien_Db_Adapter_Interface` what seems to be correct. Methods like `modify()` or `addKey()` are part of `Varien_Db_Adapter_Pdo_Mysql`.

PR #676, opened 2019-05-23
Mage_AdminNotification - IDE improvements

sreichel:
- added/fixed missing DOC blocks
- fixed exception translation
- flipped @var annotations


sreichel:
Current result from phpstan ... 

```
 ------ -------------------------------------------------------------------------------------------------------- 
  Line   Model/Feed.php                                                                                          
 ------ -------------------------------------------------------------------------------------------------------- 
  85     Right side of && is always true.                                                                        
  85     Right side of && is always true.                                                                        
  165    Strict comparison using === between string and false will always evaluate to false.                     
  189    Parameter #1 $data of class SimpleXMLElement constructor expects string, SimpleXMLElement|false given.  
 ------ -------------------------------------------------------------------------------------------------------- 
```

PR #677, opened 2019-05-23
Mage_Api - IDE improvements

sreichel:
- added/fixed missing DOCs
- changed @return to $this for methods
- use @inheritdoc for methods that return parent call
- <s>added class check (instanceof) for those methods (instead adding @var SomeClass)</s>
- <s>changed typehint to concrete class instead of Mage_Core_Model_Abstract</s>
- flipped @var annotations


app/code/core/Mage/Api/Model/Resource/Role.php
```diff
@@ -44,27 +44,21 @@ protected function _construct()
...
-     * @param Mage_Core_Model_Abstract $object
-     * @return $this
+     * @inheritDoc
      */
-    protected function _beforeSave(Mage_Core_Model_Abstract $object)
+    protected function _beforeSave(Mage_Core_Model_Abstract $role)
     {
-        if (!$object->getId()) {
-            $object->setCreated(now());
+        if ($role instanceof Mage_Admin_Model_Role) {
```
Flyingmana: That looks like more then just an IDE improvement, did I miss the description in the PR for this?

sreichel: No, wasnt mentioned in this PR. Updated <sup>(1)</sup>.

This only applies to methods that inherit from abstract `Mage_Core_Model_Resource_Db_Abstract` class like `_beforeSave` etc ... 

Renaming wasnt neccessry, just for better readablity ... Is it possible that `$object` is something else then `Mage_Admin_Model_Role` in this case - if this was your question?

app/code/core/Mage/Api/Model/Resource/Role.php
```diff
@@ -44,27 +44,21 @@ protected function _construct()
...
+            if (!$role->getId()) {
+                $role->setCreated(now());
+            }
+            $role->setModified(now());
         }
-        $object->setModified(now());
-        return $this;
+        return parent::_beforeSave($role);
```
Flyingmana: the call of the parent method is also new if I see this correctly.

Flyingmana:
sorry for the extra work this causes, but I think the changes inside the code (like in the method where I did comment in) should be better done in a separate PR.
The non-code changes look fine, but reviewing the code part is currently to much extra work for me

app/code/core/Mage/Api/Model/Resource/Role.php
sreichel: Also just for inherited methods from `Mage_Core_Model_Resource_Db_Abstract` ... just for code consistency. (parent just returns `$this`)

sreichel:
No problem. Will split this (same for other PRs). Thanks for review. :)

sreichel:
@Flyingmana no problem to split this.

Are you okay with revert `instanceof` and `parent::method()` changes, but keep `set/getData()` instead of magic methods on `Varien_Object`?

app/code/core/Mage/Api/Model/Acl/Assert/Time.php
```diff
@@ -49,6 +49,9 @@ public function assert(Mage_Api_Model_Acl $acl, Mage_Api_Model_Acl_Role $role =
         return $this->_isCleanTime(time());
     }
 
+    /**
+     * @param $time
```
Flyingmana: no benefit without type for IDE

app/code/core/Mage/Api/Model/Config.php
```diff
@@ -38,9 +38,10 @@ class Mage_Api_Model_Config extends Varien_Simplexml_Config
     /**
      * Constructor
      *
+     * @param null $sourceData
```
Flyingmana: incorrect datatype

app/code/core/Mage/Api/Model/Resource/Role/Collection.php
```diff
@@ -46,7 +46,7 @@ protected function _construct()
     /**
      * Aet user filter
      *
-     * @param int $userId
+     * @param string $userId
```
Flyingmana: user ID is not a string, is it?

app/code/core/Mage/Api/Model/Resource/Roles.php
```diff
@@ -77,7 +77,7 @@ protected function _beforeSave(Mage_Core_Model_Abstract $role)
         }
 
         if ($role->getPid() > 0) {
-            $row = $this->load($role->getPid());
+            $row = $this->load($role->getPid(), null);
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Resource/Roles.php
```diff
@@ -94,8 +94,10 @@ protected function _beforeSave(Mage_Core_Model_Abstract $role)
      */
     protected function _afterSave(Mage_Core_Model_Abstract $role)
     {
-        $this->_updateRoleUsersAcl($role);
-        Mage::app()->getCache()->clean(Zend_Cache::CLEANING_MODE_MATCHING_TAG);
+        if ($role instanceof Mage_Api_Model_Roles) {
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Resource/Rules/Collection.php
```diff
@@ -46,12 +46,12 @@ protected function _construct()
...
      * Retrieve rules by role
      *
-     * @param int $id
+     * @param string $id
      * @return $this
      */
     public function getByRoles($id)
     {
-        $this->getSelect()->where("role_id = ?", (int)$id);
+        $this->getSelect()->where("role_id = ?", $id);
```
Flyingmana: logic/code change

app/code/core/Mage/Api/Model/Resource/User.php
```diff
@@ -219,16 +219,16 @@ public function hasAssigned2Role($user)
     /**
      * Action before save
      *
-     * @param Mage_Core_Model_Abstract $user
-     * @return $this
+     * @param Mage_Api_Model_User $user
```
Flyingmana: change does not reflect code

app/code/core/Mage/Api/Model/Resource/User.php
```diff
@@ -219,16 +219,16 @@ public function hasAssigned2Role($user)
...
+     * @inheritDoc
      */
     protected function _beforeSave(Mage_Core_Model_Abstract $user)
     {
         if (!$user->getId()) {
             $user->setCreated(now());
         }
         $user->setModified(now());
-        return $this;
+        return parent::_beforeSave($user);
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Resource/User.php
```diff
@@ -335,14 +335,14 @@ public function _getRoles(Mage_Core_Model_Abstract $user)
...
-     * @param Mage_Core_Model_Abstract $user
+     * @param Mage_Core_Model_Abstract|Mage_Api_Model_User $user
      * @return $this
      */
     public function add(Mage_Core_Model_Abstract $user)
     {
         $adapter = $this->_getWriteAdapter();
         $aRoles  = $this->hasAssigned2Role($user);
-        if (sizeof($aRoles) > 0) {
+        if (is_array($aRoles) && count($aRoles) > 0) {
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Resource/User.php
```diff
@@ -354,7 +354,7 @@ public function add(Mage_Core_Model_Abstract $user)
         if ($user->getId() > 0) {
             $role = Mage::getModel('api/role')->load($user->getRoleId());
         } else {
-            $role = new Varien_Object(array('tree_level' => 0));
+            $role = new Mage_Api_Helper_Object_Role(array('tree_level' => 0));
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Wsdl/Config.php
```diff
@@ -68,26 +71,47 @@ public static function getNamespacesPrefix()
...
     {
         return Mage::app()->loadCache($id);
     }
 
-    protected function _saveCache($data, $id, $tags=array(), $lifetime=false)
+    /**
+     * @param string $data
+     * @param string $id
+     * @param array $tags
+     * @param bool $lifetime
```
Flyingmana: wrong datatype, its int|bool

app/code/core/Mage/Api/Model/Wsdl/Config/Base.php
```diff
@@ -51,7 +54,7 @@ public function __construct($sourceData=null)
         unset($queryParams['wsdl']);
 
         // set up default WSDL template variables
-        $this->_wsdlVariables = new Varien_Object(
+        $this->_wsdlVariables = new Mage_Api_Helper_Object_Wsdl(
```
Flyingmana: code change

app/code/core/Mage/Api/Model/Resource/User.php
sreichel: Missed to revert. Thx.

app/code/core/Mage/Api/Model/Resource/User.php
sreichel: Cause by wrong DOC. It is an array in any case ...

app/code/core/Mage/Api/Model/Resource/User.php
sreichel: Pls see #679

app/code/core/Mage/Api/Model/Wsdl/Config.php
sreichel: Yep.

app/code/core/Mage/Api/Model/Resource/Role.php
sreichel: Reopend below ...

PR #679, opened 2019-05-28
Mage_Cms - IDE improvements

sreichel:
- added/fixed missing DOCs
- added @method annotations
- changed @return to $this for methods and annotation
- use @inheritdoc for methods that return parent call
- flipped @var annotations


app/code/core/Mage/Cms/Model/Resource/Page.php
```diff
@@ -390,7 +375,7 @@ public function lookupStoreIds($pageId)
 
         $select  = $adapter->select()
             ->from($this->getTable('cms/page_store'), 'store_id')
-            ->where('page_id = ?',(int)$pageId);
+            ->where('page_id = ?', $pageId);
```
kkrieger85: 1) Code change

2) Why do you remove cast to `int` if id field in DB ist integer?

mysql> describe cms_page_store;
+----------+----------------------+------+-----+---------+-------+
| Field    | Type                 | Null | Key | Default | Extra |
+----------+----------------------+------+-----+---------+-------+
| page_id  | smallint(6)          | NO   | PRI | NULL    |       |
| store_id | smallint(5) unsigned | NO   | PRI | NULL    |       |
+----------+----------------------+------+-----+---------+-------+

sreichel: Well ... DOC block change is correct. `$cmsPage->load(1)->getId()` returns a string.

Is cast to int required in select statement? The `where()` method accepts string (should be string|array) ... the only difference is in `Zend_Db_Adapter_Abstract::_quote()`

    /**
     * Quote a raw string.
     *
     * @param string $value     Raw string
     * @return string           Quoted string
     */
    protected function _quote($value)
    {
        if (is_int($value)) {
            return $value;
        } elseif (is_float($value)) {
            return sprintf('%F', $value);
        }
        return "'" . addcslashes($value, "\000\n\r\\'\"\032") . "'";
    }

Will revert and update according blocks.

Thanks for review!

PR #681, opened 2019-05-28
Mage_Customer - IDE improvements

sreichel:
- added/fixed missing DOCs
- changed @return to $this for methods and annotations
- flipped @var annotations

### DIFF
--- a/app/code/core/Mage/Customer/Block/Account/Dashboard/Newsletter.php
+++ b/app/code/core/Mage/Customer/Block/Account/Dashboard/Newsletter.php
@@ -34,9 +34,17 @@
 
 class Mage_Customer_Block_Account_Dashboard_Newsletter extends Mage_Core_Block_Template
 {
+    /**
+     * @var Mage_Newsletter_Model_Subscriber
+     */
+    public $_subscription;
+
+    /**
+     * @return Mage_Newsletter_Model_Subscriber
+     */
     public function getSubscriptionObject()
     {
-        if(is_null($this->_subscription)) {
+        if (is_null($this->_subscription)) {
             $this->_subscription = Mage::getModel('newsletter/subscriber')
                 ->loadByCustomer(Mage::getSingleton('customer/session')->getCustomer());
         }
--- a/app/code/core/Mage/Customer/Model/Convert/Parser/Customer.php
+++ b/app/code/core/Mage/Customer/Model/Convert/Parser/Customer.php
@@ -63,6 +63,9 @@ class Mage_Customer_Model_Convert_Parser_Customer
      */
     protected $_customerGroups = null;
 
+    /**
+     * @return Mage_Core_Model_Config_Element[]|SimpleXMLElement|null
+     */
     public function getFields()
     {
         if (!$this->_fields) {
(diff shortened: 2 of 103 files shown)


app/code/core/Mage/Customer/Block/Account/Dashboard/Newsletter.php
```diff
@@ -34,9 +34,17 @@
 
 class Mage_Customer_Block_Account_Dashboard_Newsletter extends Mage_Core_Block_Template
 {
+    /**
+     * @var Mage_Newsletter_Model_Subscriber
+     */
+    public $_subscription;
```
kkrieger85: Variables with underscore are hint for private attributes.

See [PEAR Manual](https://pear.php.net/manual/en/standards.naming.php)

app/code/core/Mage/Customer/Model/Convert/Parser/Customer.php
```diff
@@ -696,10 +606,9 @@ public function parse()
                         if ($customer->getDefaultBilling()) {
                             $model->setDefaultBilling($customer->getDefaultBilling());
                         } else {
-                            $shippingAddress->save();
+                            $billingAddress->save();
```
kkrieger85: This is a fundemental code change. 

Even if this is right, it should not be in this commit.

app/code/core/Mage/Customer/Block/Account/Dashboard/Newsletter.php
sreichel: This var wasnt set and public before. Make it protected/private could cause problem with code relying on this?

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -652,7 +653,7 @@ public function confirmAction()
     public function confirmationAction()
     {
         $customer = $this->_getModel('customer/customer');
-        if ($this->_getSession()->isLoggedIn()) {
+        if (!$customer instanceof Mage_Customer_Model_Customer || $this->_getSession()->isLoggedIn()) {
```
sreichel: Revert .... add @var annotation?

app/code/core/Mage/Customer/Model/Convert/Parser/Customer.php
sreichel: Will be a own PR.

colinmollenhour:
Just an FYI on the off chance you aren't already aware of this feature, to separate these changes into separate commits one good way is to use `git add --interactive` and once you get the hang of it, it is pretty fast. It basically presents each "chunk" one at a time so you can stage parts of a file without staging other parts.

app/code/core/Mage/Customer/Model/Convert/Parser/Customer.php
sreichel: Resolved in #701

app/code/core/Mage/Customer/Block/Account/Dashboard/Newsletter.php
sreichel: Fixed in #701

PR #682, opened 2019-05-29
Mage_Core - IDE improvements

sreichel:
- added/fixed missing DOCs
- add @method annotations
- changed @return to $this for methods and annotations
- flipped @var annotations
- ...

### DIFF
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -121,8 +121,7 @@ class Mage_Core_Helper_Data extends Mage_Core_Helper_Abstract
             }
 
             $value = $store->convertPrice($value, $format, $includeContainer);
-        }
-        catch (Exception $e){
+        } catch (Exception $e) {
             $value = $e->getMessage();
         }
 
@@ -156,7 +155,7 @@ class Mage_Core_Helper_Data extends Mage_Core_Helper_Abstract
     /**
      * Format date using current locale options and time zone.
      *
-     * @param   date|Zend_Date|null $date
+     * @param   string|Zend_Date|null $date
      * @param   string              $format   See Mage_Core_Model_Locale::FORMAT_TYPE_* constants
      * @param   bool                $showTime Whether to include time
      * @return  string
@@ -171,7 +170,7 @@ class Mage_Core_Helper_Data extends Mage_Core_Helper_Abstract
         }
         if (is_null($date)) {
             $date = Mage::app()->getLocale()->date(Mage::getSingleton('core/date')->gmtTimestamp(), null, null);
-        } else if (!$date instanceof Zend_Date) {
+        } elseif (!$date instanceof Zend_Date) {
             $date = Mage::app()->getLocale()->date(strtotime($date), null, null);
         }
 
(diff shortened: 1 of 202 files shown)


app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -200,7 +199,7 @@ public function formatTime($time = null, $format = Mage_Core_Model_Locale::FORMA
 
         if (is_null($time)) {
             $date = Mage::app()->getLocale()->date(time());
-        } else if ($time instanceof Zend_Date) {
+        } elseif ($time instanceof Zend_Date) {
```
kkrieger85: Difference `else if` and `elseif`:

> Note: Note that elseif and else if will only be considered exactly the same when using curly brackets as in the above example. When using a colon to define your if/elseif conditions, you must not separate else if into two words, or PHP will fail with a parse error.

#ref https://www.php.net/manual/en/control-structures.elseif.php

app/code/core/Mage/Core/Model/App/Emulation.php
```diff
@@ -88,9 +90,9 @@ public function startEnvironmentEmulation($storeId, $area = Mage_Core_Model_App_
         $initialLocaleCode = $this->_emulateLocale($storeId, $area);
 
         $initialEnvironmentInfo = new Varien_Object();
-        $initialEnvironmentInfo->setInitialTranslateInline($initialTranslateInline)
-            ->setInitialDesign($initialDesign)
-            ->setInitialLocaleCode($initialLocaleCode);
+        $initialEnvironmentInfo->setData('initial_translate_in_line', $initialTranslateInline)
```
kkrieger85: As far as I read this line it should be

`$initialEnvironmentInfo->setData('initial_translate_inline', $initialTranslateInline)`

app/code/core/Mage/Core/Model/App/Emulation.php
```diff
@@ -106,12 +108,12 @@ public function startEnvironmentEmulation($storeId, $area = Mage_Core_Model_App_
      */
     public function stopEnvironmentEmulation(Varien_Object $initialEnvironmentInfo)
     {
-        $this->_restoreInitialInlineTranslation($initialEnvironmentInfo->getInitialTranslateInline());
-        $initialDesign = $initialEnvironmentInfo->getInitialDesign();
+        $this->_restoreInitialInlineTranslation($initialEnvironmentInfo->getData('initial_translate_in_line'));
```
kkrieger85: As far as I read this line it should be

`$this->_restoreInitialInlineTranslation($initialEnvironmentInfo->getData('initial_translate_inline'));`

app/code/core/Mage/Core/Model/App/Emulation.php
sreichel: Will be reverted.

app/code/core/Mage/Core/Model/App/Emulation.php
sreichel: Will be reverted.

sreichel:
For some reason it looked like "InLine" ... it works since getter/setter have same key, but it will be reverted.

Q: idea was to make phpstan/phpstorm work with Varian_Objects. You cannot use annotations or something else, so the idea was to replace magic methods .... I've already reverted this in another PR and tried something else ... https://github.com/OpenMage/magento-lts/pull/677/commits/088d028850a95af3baabe3e6c91f374d4e24cbfe Maybe this is better?

app/code/core/Mage/Core/Helper/Data.php
sreichel: Since we have curly brackets it doesnt matter?

PR #694, opened 2019-06-02
Mage_Cms - DOC block update

sreichel:
Cleaned #679 ... only DOC changes left.


app/code/core/Mage/Cms/Block/Block.php
```diff
@@ -24,17 +24,19 @@
...
 /**
  * Cms block content block
  *
+ * @method int getBlockId()
+ *
  * @category   Mage
  * @package    Mage_Cms
  * @author     Magento Core Team <core@magentocommerce.com>
+ *
+ * @method $this setBlockId(int $int)
```
tmotyl, line 36: lets group `@method` together

sreichel: Good catch. :)

app/code/core/Mage/Cms/Model/Resource/Block.php
```diff
@@ -63,8 +59,9 @@ protected function _beforeDelete(Mage_Core_Model_Abstract $object)
     /**
      * Perform operations before object save
      *
-     * @param Mage_Cms_Model_Block $object
+     * @param Mage_Core_Model_Abstract $object
```
tmotyl, line 62: Mage_Cms_Model_Block is correct

sreichel: Yes and no.

`Mage_Cms_Model_Block`  is expected, but signature says ... `protected function _beforeSave(Mage_Core_Model_Abstract $object)`

I'll update to `Mage_Core_Model_AbstractMage_Cms_Model_Block |`

app/code/core/Mage/Cms/Helper/Data.php
```diff
@@ -52,7 +52,7 @@ public function getPageTemplateProcessor()
     /**
      * Retrieve Template processor for Block Content
      *
-     * @return Varien_Filter_Template
+     * @return Mage_Core_Model_Abstract
      */
```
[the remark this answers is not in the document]
sreichel: Problem for static analysis is this: `return Mage::getModel($model);`

How about `Mage_Core_Model_Abstract|Varien_Filter_Template`?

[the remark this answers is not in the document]
tmotyl: ok

app/code/core/Mage/Cms/Model/Resource/Block.php
tmotyl: ok, this will work.
I saw that you did add different type to doc comment then typehint in other places. even in this patch. see:

``` 
    * @param Mage_Cms_Model_Page $object
     * @inheritDoc
     */
    protected function _beforeSave(Mage_Core_Model_Abstract $object)

```

PR #697, opened 2019-06-02
Mage_Cron - DOC block update

sreichel:


app/code/core/Mage/Cron/Model/Schedule.php
```diff
@@ -61,10 +65,15 @@ public function _construct()
...
+    /**
+     * @param string $expr
+     * @return $this
+     * @throws Mage_Core_Exception
+     */
     public function setCronExpr($expr)
     {
         $e = preg_split('#\s+#', $expr, null, PREG_SPLIT_NO_EMPTY);
-        if (sizeof($e)<5 || sizeof($e)>6) {
+        if (is_array($e) && (sizeof($e) < 5 || sizeof($e) > 6)) {
```
kkrieger85: Code change

sreichel: `preg_split` returns `array|false` ... so no code change, just an improvement that belongs to next phpstan related PR.

I see no harm, but can exclude this here, if wanted?!?

kkrieger85: I like this small improvement. But I think it should be separated into a new PR

sreichel: There will be a few of such changes ... if accepted :)

PR #700, opened 2019-06-02
Possible improvement for register, should break nothing

seansan:
Maybe two changes in 1 ... 
1. graceful is still gracefull, but does! tell the admin about via log
2. Exception logs the value also. Not sure about this one ... normally register registers simple values, but if whole objects are stored then maybe this is not smart

just an idea

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -219,9 +219,10 @@ final class Mage
     {
         if (isset(self::$_registry[$key])) {
             if ($graceful) {
+                self::log('Mage registry key "'.$key.'" already exists');
                 return;
             }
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Thanks for suggesting these changes, but I disagree with them..

- If logging is desired then the `$graceful` flag should not be used and exception can easily be caught and logged.
- Using `print_r` is extremely prone to consuming a ton of memory and with any circular references (common in Magento code) can easily cause out of memory errors resulting in the original error not being logged in place of a fatal error.

seansan:
Yes I agree and have seen the same. Will close.

PR #703, opened 2019-06-02
Varien_ - DOC block update (1)

sreichel:
- Varien_Data_
- Varien_Event_
- Varien_Object

Big part for PHPStorm/phpstan :)


lib/Varien/Data/Form/Element/Fieldset.php
```diff
@@ -41,6 +43,11 @@ class Varien_Data_Form_Element_Fieldset extends Varien_Data_Form_Element_Abstrac
      */
     protected $_sortChildrenByKey = '';
 
+    /**
+     * @var string
+     */
+    protected $_sortDirection = '';
```
Flyingmana: This is a logic change, which prevents the current behaviour of this value beeing part of _data over __set and __get, which also means you change the visibility from public to protected.

sreichel: Is it intentional logic or just a missed field? There arn't getters/setters for this. I'll revert.

Flyingmana: It may not be intentional, but its there for a long time, so there may be a lot of people who did make use of it in the way it is currently over the getData()

lib/Varien/Data/Collection/Db.php
```diff
@@ -775,7 +786,7 @@ protected function _saveCache($data, $select)
     /**
      * Check if cache can be used for collection data
      *
-     * @return bool
+     * @return Zend_Cache_Core
```
tmotyl, line 789: I would say this should stay bool, as everywhere in the core it's used as bool.
In a separate patch we can change the return type to bool always.

lib/Varien/Data/Tree/Db.php
```diff
@@ -114,11 +114,17 @@ public function __construct($connection, $table, $fields)
         $this->_select->from($this->_table, array_values($fields));
     }
 
+    /**
+     * @return Zend_Db_Select
```
tmotyl, line 118: Varien_Db_Select ? and in the property too?

lib/Varien/Data/Collection/Db.php
sreichel: > I would say this should stay bool, as everywhere in the core it's used as bool.
> In a separate patch we can change the return type to bool always.

Then everywhere else it is wrong too. Again, the "Doc-Updates" depend on each other ... i'm pretty sure it has been corrected in all classes (also Varien_). Espeacially "core methods" like this have been checked manually via xdebug.

This sh*** took probably the most time ... :(

lib/Varien/Data/Tree/Db.php
sreichel: I'm pretty sure if xdebuged this .... i'll check again.

sreichel: It is ok. Please check `__construct()` ...

`$this->_select  = $this->_conn->select();` returns ....

```
    /**
     * Creates and returns a new Zend_Db_Select object for this adapter.
     *
     * @return Zend_Db_Select
     */
    public function select()
    {
        return new Zend_Db_Select($this);
    }
```

lib/Varien/Data/Collection/Db.php
sreichel: @tmotyl it is `Zend_Cache_Core|false`.

tmotyl:
please resolve the conflict

PR #707, opened 2019-06-04
Mage_Poll - DOC block update

sreichel:
Docs only.


app/code/core/Mage/Poll/Model/Resource/Poll/Answer/Collection.php
```diff
@@ -64,7 +66,7 @@ public function addPollFilter($pollId)
     public function countPercent($pollObject)
     {
         if (!$pollObject) {
-            return;
+            return $this;
```
Flyingmana: change acknowledged

sreichel:
Duplicatates #758 (later one is more up to date)

PR #708, opened 2019-06-04
Mage_CurrencySymbol - DOC block update

sreichel:
- add missing properties

### DIFF
--- a/app/code/core/Mage/CurrencySymbol/Block/Adminhtml/System/Currencysymbol.php
+++ b/app/code/core/Mage/CurrencySymbol/Block/Adminhtml/System/Currencysymbol.php
@@ -30,14 +30,11 @@
  * @category   Mage
  * @package    Mage_CurrencySymbol
  * @author     Magento Core Team <core@magentocommerce.com>
+ *
+ * @property string $_blockGroup
  */
 class Mage_CurrencySymbol_Block_Adminhtml_System_Currencysymbol extends Mage_Adminhtml_Block_Widget_Form
 {
-    /**
-     * @var string
-     */
-    private $_blockGroup;
-
     /**
      * @var string
      */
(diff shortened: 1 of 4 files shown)


app/code/core/Mage/CurrencySymbol/Block/Adminhtml/System/Currencysymbol.php
```diff
@@ -33,6 +33,16 @@
  */
 class Mage_Currencysymbol_Block_Adminhtml_System_Currencysymbol extends Mage_Adminhtml_Block_Widget_Form
 {
+    /**
+     * @var string
+     */
+    private $_blockGroup;
```
Sekiphp: I think that _blockGroup, _controller should be protected or deleted. As I see in project this attributes is not used. From this class is not extended any class, so I think, that these variables are pointless.

sreichel: Fixed.

app/code/core/Mage/CurrencySymbol/Model/System/Currencysymbol.php
```diff
@@ -30,6 +30,8 @@
  * @category    Mage
  * @package     Mage_CurrencySymbol
  * @author      Magento Core Team <core@magentocommerce.com>
+ *
+ * @method $this resetValues()
```
tmotyl, line 34: what is it used for? what does it mean?

sreichel: Seems to be wrong commited .... can't find this in my patches. I'll check ....

sreichel: It is "used" in `Mage_CurrencySymbol_Adminhtml_System_CurrencysymbolController::resetAction()`, but this method does not exist ... not sure why, but i guess it was just to supress warnings.

tmotyl: ok, we have bigger issue here. adding this doc comment doesn't hurt

PR #711, opened 2019-06-08
Add missing method _beforeLoadCollection()

Sekiphp:
Magento grids are extended from `Mage_Adminhtml_Block_Widget_Grid` class. In this class is method `_afterLoadCollection()` but not the before method.

**Example of usage:**
If you want to extend, for example, customers grid you shold create rewrite and extend this Grid. And if you want to add some column, you have to add some joins to `_prepareCollection` method, but if you don't want to copy-paste code from parent, currently there is only one way how to do this - use `_preparePage` method and inside get collection, modify and return parent. But I think, that this is little bug. Mofify collection in `_prepareCollection` is not possible because collection is loaded, when I call parent.

_I think that this is non-breaking feature._

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -547,6 +547,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
             }
 
             if (!$this->_isExport) {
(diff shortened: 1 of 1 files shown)


seansan:
Would this work with the - used by many - eag grid manager?

Sekiphp:
@seansan Do you talking about BL custom grids?

seansan:
Yes that one

Issue #713, opened 2019-06-08
[Bug] Mage_Directory_Helper_Data::getRegionCollection()


sreichel:
> Fatal error: Uncaught Error: Call to undefined method Mage_Directory_Helper_Data::getAddress()

```
Mage::helper('directory')->getRegionCollection();
```

Sekiphp:
This error is generated here:
`    /**
     * Retrieve region collection
     *
     * @return Mage_Directory_Model_Resource_Region_Collection
     */
    public function getRegionCollection()
    {
        if (!$this->_regionCollection) {
            $this->_regionCollection = Mage::getModel('directory/region')->getResourceCollection()
                ->addCountryFilter($this->getAddress()->getCountryId())
                ->load();
        }
        return $this->_regionCollection;
    }`

Similar method is implemented in `Mage_Directory_Block_Data` but there is defined country filter using this function: 
`    public function getCountryId()
    {
        $countryId = $this->getData('country_id');
        if (is_null($countryId)) {
            $countryId = Mage::helper('core')->getDefaultCountry();
        }
        return $countryId;
    }`

It's hard to say, what implementation is correct. It seems, that this function never worked before. I think, that right implementation is create optional parameter for filtering.

sreichel:
#718

real34:
For the record, this issue is still occurring since 19.4.4.
The fix was reverted as part of https://github.com/OpenMage/magento-lts/pull/773#pullrequestreview-409273882 as it seemed to create another problem.

Should it be reopened?

Flyingmana:
should be fixed with next release, please reopen if its not the case

PR #718, opened 2019-06-11
Fix fatal error when getRegionCollection() is called

Sekiphp:
**Issue:** #713 

Tonight I spent some time with research - this bug is in Magento since 1.1.1, and I think that this method never worked before.

**Problematic line:** `->addCountryFilter($this->getAddress()->getCountryId())`

**Implementation of `addCountryFilter`:**
```
    public function addCountryFilter($countryId)
    {
        if (!empty($countryId)) {
            if (is_array($countryId)) {
                $this->addFieldToFilter('main_table.country_id', array('in' => $countryId));
            } else {
                $this->addFieldToFilter('main_table.country_id', $countryId);
            }
        }
        return $this;
    }
```

**Explanation of my code:** `addCountryFilter` accept NULL as param and when getRegionCollection is called, we wants to get full collection without filter (expected behavior according to method name).

### DIFF
--- a/app/code/core/Mage/Directory/Helper/Data.php
+++ b/app/code/core/Mage/Directory/Helper/Data.php
@@ -107,14 +107,14 @@ class Mage_Directory_Helper_Data extends Mage_Core_Helper_Abstract
 
     /**
      * Retrieve region collection
-     *
+     * @param string|array|null $countryFilter
      * @return Mage_Directory_Model_Resource_Region_Collection
      */
-    public function getRegionCollection()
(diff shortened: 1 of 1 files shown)


sreichel:
Thanks. :+1: 

I had just deleted it, because there a several posibilities ...

```
public function getAddress()
{
    $address = Mage::registry('current_address'); # does not exist
    if ($address instanceof Varien_Object) {
        return $address;
    }
    return new Varien_Object();
}
```

PR #720, opened 2019-06-13
Fatal error: Call to protected function in Varien_Date

Sekiphp:
**Error message:** PHP Fatal error:  Uncaught Error: Call to protected method Zend_Date_DateObject::getUnixTimestamp() from context 'Varien_Date' in lib/Varien/Date.php:127

Method `getUnixTimestamp` is protected. This method is located in `Zend_Date_DateObject` and its protected. But In `Zend_Date_DateObject::getUnixTimestamp()` is this line: `if ($date instanceof Zend_Date) {` and if i go to `Zend_Date` there is this public method:

```
    public function getTimestamp()
    {
        return $this->getUnixTimestamp();
    }
```

### DIFF
--- a/lib/Varien/Date.php
+++ b/lib/Varien/Date.php
@@ -124,7 +124,7 @@ class Varien_Date
     public static function toTimestamp($date)
     {
         if ($date instanceof Zend_Date) {
-            return $date->getUnixTimestamp();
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Thanks!

Issue #719, opened 2019-06-13
Mage_Downloadable_Adminhtml_Downloadable_Product_EditController::linkAction()


sreichel:
> Mage_Downloadable_Adminhtml_Downloadable_Product_EditController::linkAction()

Method `_getCustomerSession` does not exist ...

```
            try {
                $this->_processDownload($resource, $resourceType);
            } catch (Mage_Core_Exception $e) {
                $this->_getCustomerSession()->addError(Mage::helper('downloadable')->__('An error occurred while getting the requested content.'));
            }
```

mattdavenport:
This appears to be fixed. On the `1.9.4.x` branch it is now using `$this->_getSession()`.

colinmollenhour:
Almost exactly one year to the day! :)

PR #723, opened 2019-06-13
Fix missing method in Mage_Downloadable_Adminhtml_Downloadable_Product_EditController::linkAction()

Sekiphp:
### Issue: #719 

Currently untested, but should work :-)

### DIFF
--- a/app/code/core/Mage/Downloadable/controllers/Adminhtml/Downloadable/Product/EditController.php
+++ b/app/code/core/Mage/Downloadable/controllers/Adminhtml/Downloadable/Product/EditController.php
@@ -127,14 +127,4 @@ class Mage_Downloadable_Adminhtml_Downloadable_Product_EditController extends Ma
         exit(0);
     }
 
-    /**
-     * Return admin session object
-     *
-     * @return Mage_Admin_Model_Session
-     */
-    protected function _getSession()
-    {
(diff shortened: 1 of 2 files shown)


Sekiphp:
> Either it's `customer/session` for customer session, or `_getSession()` for admin session.

I don't think so.... This class is in adminhtml scope - so I think, that `admin/session` is correct way. Now I see, that this is more complicated.....but I have an idea how to fix this.

sreichel:
> > Either it's `customer/session` for customer session, or `_getSession()` for admin session.
> 
> I don't think so.... This class is in adminhtml scope - so I think, that `admin/session` is correct way.

You got me wrong. Sorry. Next time i'll be more clear :)

If we follow **wrong** naming it "**should**" be `customer/session` . Your first commit was nearly right, but instead of 

    Mage::getSingleton('admin/session')

use helper method (from Mage_Adminhtml_ControllerAction)

    $this->_getSession()

Sekiphp:
@sreichel I think, that is correct solution now. You have right with existing `_getSession()` method, but we need to conserve overriding this method in child (because this is not used in admin scope)

Sekiphp:
Someone to re-check @sreichel ?

Sekiphp:
Issue was closed by colinmollenhour on 15 Jun 2020

Issue #280, opened 2017-06-11
Bug: Fatal error in salesOrderBeforeSave event for deleted payment


sreichel:
See; https://magento.stackexchange.com/questions/104468/getting-php-fatal-error-call-to-a-member-function-getmethodinstance-on-boolea

Possible fix:
https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Payment/Model/Observer.php#L46

Change:

    if ($order->getPayment()->getMethodInstance()->getCode() != 'free') {

To:

    if ($order->getPayment() && $order->getPayment()->getMethodInstance()->getCode() != 'free') {

Flyingmana:
in which cases is a deleted payment to be expected?

sreichel:
Not tested yet, but one (for me) reproducable case could be .... disable M2E Pro Extension and modify an Order, where payment method is 'm2e_payment'.

Flyingmana:
in this case you dont delete the payment methods with this code, you configure a pseudo method with the same type to avoid all the errors.(I know there are some errors in the admin area, if the payment method to the code is not there anymore)

I think it is a bad pattern to delete payments. Once there may be a lot of modules, which expect a payment to exist for an order. And you need to reserve it for quite some time, to being able to handle returns correctly.

So I continue asking, is there another case where a payment is expected to be deleted?

colinmollenhour:
I think @sreichel's suggested case is valid since stores can be installed for many years and naturally things change.. New payment methods are adopted and old ones deprecated. Perhaps the original author quit supporting it or it was replaced with a different method due to PCI compliance changes, etc..

In my opinion if @getPayment()@ method can return something other than a MSMO_Payment then every time it is used the return value should be checked before assuming that it is an object (unless for some reason that is guaranteed to be a safe assumption such as it was just set in the same request). This applies as a general rule for all methods that can return both objects and non-objects and in this case I think the suggested fix is appropriate.

rafaelpatro:
Hi guys
I was facing this issue frequently.
I have applied a Try Catch and it works too.
Thanks

damien-biasotto:
Should it sill be open since PR #730 was merged?

sreichel:
@damien-biasotto #730 is not related.

ADDISON74:
This report is almost 5 years old. If there is still a problem in OpenMage and having a solution a PR should be created.

sreichel:
> This report is almost 5 years old.

[...]

PR #730, opened 2019-06-16
Prevent fatal error in Mage_Catalog_Model_Resource_Category_Collectio…

sreichel:
…n::setVisibility()

Pass a non-product eav colletion ...

```
$collection = Mage::getModel('catalog/category')->getCollection();
Mage::getSingleton('catalog/product_visibility')->addVisibleInSearchFilterToCollection($collection);
```

>Uncaught Error: Call to undefined method Mage_Catalog_Model_Resource_Category_Collection::setVisibility()

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Visibility.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Visibility.php
@@ -60,41 +60,36 @@ class Mage_Catalog_Model_Product_Visibility extends Varien_Object
      * Add visible in catalog filter to collection
      *
      * @deprecated
-     * @param Mage_Eav_Model_Entity_Collection_Abstract $collection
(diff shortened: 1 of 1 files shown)


sreichel:
@colinmollenhour you're right. It just results in a more clear error message.

sreichel:
It's phpstan "playground" ... an easy avoidable error in static code analyse ... my goal is to reduce these errors as far as possibly, I just want to see "real"  problems at the end ...  e.g. #280

tmotyl:
@sreichel change looks good codewise, but please change the commit message to
"Set correct collection types in Mage_Catalog_Model_Product_Visibility (#730)"
then I will merge it.

sreichel:
Changed message.

PR #731, opened 2019-06-16
 Fix call to undefined method (Mage_Api) 

sreichel:
... + fix pass variable to array_pop

Reproduce:

Pass valid website and store ID to ...

```
$class = new Mage_Catalog_Model_Category_Api();
$class->level(1, 1);
```

> Call to a member function getRootCategoryId() on integer


sreichel:
Bump.

app/code/core/Mage/Catalog/Model/Category/Api.php
```diff
@@ -65,7 +65,7 @@ public function level($website = null, $store = null, $categoryId = null)
                         $ids = $categoryId;
                     }
                 } elseif (in_array($store, $website->getStoreIds())) {
-                    $storeId = Mage::app()->getStore($store)->getId();
+                    $store = Mage::app()->getStore($store);
```
Flyingmana: From the code I would expect this to cause propblems, as `$storeId` is initially set with DEFAULT_STORE_ID, and then used in line 100 after the if,elseif,else in 
```
 $collection = Mage::getModel('catalog/category')->getCollection()
            ->setStoreId($storeId)	            ->setStoreId($storeId)
            ->addAttributeToSelect('name')	            ->addAttributeToSelect('name')
            ->addAttributeToSelect('is_active');	            ->addAttributeToSelect('is_active');
```

sreichel: @Flyingmana thanks for review. I've tested it, but I'm not sure I've covered eventualities. I'll update later.

colinmollenhour: Suggestion:

```suggestion
                    $store = Mage::app()->getStore($store);
                    $storeId = $store->getId();
```

PR #735, opened 2019-06-16
Cast types instead of function

sreichel:


colinmollenhour:
What is the motivation/benefit of making this change?

sreichel:
These functions are "forbidden" in EQP standard, so they should not be used in core too.

Sekiphp:
@sreichel You talking about this repository https://github.com/magento/marketplace-eqp?

I think, that these changes are pointless. There is no bug fixes, no performance improvments....but too many changes.

sreichel:
@midlan the functions will not deprecate, because of additional features (at least intval() & strval()).

@Sekiphp to many changes for what? I do not believe thery will be big code changes anyymore, that could cause merge conflicts.

I dont think its pointless. What's wrong with consistent code? Magento 2 replaced all PHP function with typecasting. Btw ... typecasting is much more faster (i know benchmarks vs real usage)

Flyingmana:
a few things.

* Having a common standard with Magento may be beneficial.
* every change has the possibility to introduce bugs, even if it may seem impossible
* the performance difference is becoming less with every php version.


benchmark case for PHP 7.1: https://stackoverflow.com/questions/1912599/is-there-any-particular-difference-between-intval-and-casting-to-int-int-x
result: performance gain of ~0,00000014ms per call, thats an improvement of <1%

app/code/core/Mage/XmlConnect/Block/Checkout/Payment/Method/List.php
```diff
@@ -229,8 +229,8 @@ public function addCustomerBalanceToXmlObj(Mage_XmlConnect_Model_Simplexml_Eleme
                     'post_name' => 'payment[use_customer_balance]',
                     'code'      => 1,
                     'label'     => $this->__('Use Store Credit (%s available)', $balance),
-                    'is_cover_a_quote' => intval($customerBalanceBlock->isFullyPaidAfterApplication()),
-                    'selected'  => intval($customerBalanceBlock->isCustomerBalanceUsed())
+                    'is_cover_a_quote' => (int)$customerBalanceBlock->isFullyPaidAfterApplication(),
+                    'selected'  => (int)($customerBalanceBlock->isCustomerBalanceUsed()
```
midlan: @Flyingmana can't understand why you approved the pull request since there is obvious error on this line! Travis CI pass failed for all PHP versions

Flyingmana: because I missed it, and I do check travisCI not always on a review, only always on a merge

midlan: ok, consider checking travisCI every time :) It looked scary to me that fatal syntax error is approved.

colinmollenhour:
I scanned through and didn't see any errors, but what method did you use to make this change, was it manual or automated? Did you account for cases like this?

```php
intval($a * $b) // original
(int)$a * $b // incorrect
(int)($a * $b) // correct
```

sreichel:
Thanks @colinmollenhour . Good point. I manually checked all edit, but also missed that opening parenthesis)

Update tomorrow.

tmotyl:
@sreichel can you bring the patch back to life?

sreichel:
> 
> 
> @sreichel can you bring the patch back to life?

Thanks for bump. During next days ...

sreichel:
IMHO it looks good. I'd prefer the other PRs get merge and let this open (for now).

tmotyl:
ok

sreichel:
Something went wrong .... i'll create a new one.

sreichel:
> @sreichel can you bring the patch back to life?

:sunglasses:

Issue #691, opened 2019-06-02
composer.json PHP minimum Requirement => 7.x


kkrieger85:
Replaces #412 

Update PHP minimum requirements to 7.1. for at least 1.9.4.x branch

Question: Also use php min req to 7.x for 1.9.3.x? We could add https://github.com/Inchoo/Inchoo_PHP7 to composer.json requirements

reference: https://php.net/supported-versions.php

Flyingmana:
there was a comment, that there are current LTS OperatingSystems shipping with a 5.x release, are they out of support now already?

colinmollenhour:
Ubuntu 14.04 LTS is now EOL (as of April 2019). Ubuntu 16.04 LTS uses PHP 7.0.

Do we care about RHEL/CentOS? I don't think we should because do we really want to support PHP 5.4 until June 2024?

kkrieger85:
From my point of view, there is no need to support PHP 5.x. 

There's the [Inchoo/PHP7](https://github.com/Inchoo/Inchoo_PHP7) package which brings PHP7 [Support to Magento Core from 1.9.2.2. upon](https://github.com/Inchoo/Inchoo_PHP7#compatibility)

With Magento Core 1.9.4.x this package was included in Magento Core


Besides that: We can strongly recommend using PHP7 at least for performance reasons

colinmollenhour:
My vote is for 7.0 to be the minimum since it isn't that old and is still supported by Ubuntu 16.04 LTS for two more years.

Debian 8 still uses 5.6 and will not be EOL for 11 more months, but there is an upgrade path to Debian 9 which uses 7.0.

Of course, for people using Docker (which I advocate should be everyone) which distro you run doesn't matter. The official docker images actually stopped supporting 7.0 already with 7.1 being the oldest getting updates.

Flyingmana:
then Iam fine with it, too.

sreichel:
>Question: Also use php min req to 7.x for 1.9.3.x?

I'd leave 1.9.3x as it is ... just add official patches (?), all other changes should be 1.9.4.x only.

Flyingmana:
done via #736

PR #736, opened 2019-06-17
Drop support for EOL PHP5

sreichel:
Nails ... ref  #691

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: php
 
 php:
-  - 5.6
   - 7.0
   - 7.1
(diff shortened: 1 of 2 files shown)


sreichel:
> jaroschek pushed a commit to sandstein/magento-lts that referenced this pull request on 27 Aug 2019
Stefan Jaroschek
Revert "Drop support for EOL PHP5 (OpenMage#736)" 

Gerade zufällig drüber gestolpert .... omg sorry, aber ich muss gerade herzlich lachen. :)

---

We should clearly announce dropped PHP5 support!

PR #740, opened 2019-06-19
Fixed typos in method names

sreichel:


joshua-bn:
> OMG... how could this be ignored/missed for years.

At least for `addStockStatusFilterToSelect` it is not used anywhere

PR #748, opened 2019-06-20
Mage_Shipping - DOC block update

sreichel:
- doc blocks added/fixed


app/code/core/Mage/Shipping/Model/Rate/Result/Error.php
```diff
@@ -24,10 +24,18 @@
...
+ *
+ * @method $this setCarrier(string $value)
+ * @method $this setCarrierTitle(string $value)
+ * @method $this setErrorMessage(string $value)
+ */
 class Mage_Shipping_Model_Rate_Result_Error extends Mage_Shipping_Model_Rate_Result_Abstract
 {
-
+    /**
+     * @return mixed
```
tmotyl, line 37: string

sreichel: Had checked all `mixed` types later ... ;)

PR #754, opened 2019-06-22
Mage_Review - DOC block update

sreichel:
- doc blocks added/fixed
- fixed undefined property used instead of magic getter (Block/Form.php::L62)


app/code/core/Mage/Review/Block/Customer/List.php
```diff
@@ -130,7 +130,7 @@ public function getProductLink()
     /**
      * Format date in short format
      *
-     * @param $date
+     * @param string $date
```
tmotyl, line 133: string|Zend_Date|null

app/code/core/Mage/Review/Model/Review/Summary.php
```diff
@@ -24,14 +24,15 @@
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-/**
+/**$this
```
tmotyl, line 27: copy paster error?

app/code/core/Mage/Review/Block/Customer/List.php
sreichel: I'll change and added comment to #784 to recheck after all is merged (Mage_Core eg.)

PR #758, opened 2019-06-22
Mage_Poll - DOC block update

sreichel:
- doc blocks added/fixed


app/code/core/Mage/Poll/Model/Poll.php
```diff
@@ -196,15 +209,14 @@ public function addVote(Mage_Poll_Model_Poll_Vote $vote)
      * Check answer existing for poll
      *
      * @param   mixed $answer
-     * @return  boll
+     * @return  false|string
```
Flyingmana: while checkAnswerId() indeed can return the ID (as a string), its clearly not intended, and definitely something which should change, so it returns only a bool

sreichel: I was told to split doc changes from all others ... so I'd like to this this for now and fix it in a later PR. (it's just casting to boolean?)

Leave a note here #784 or raise another issue for such fixes?

tmotyl: I suggest - fix the boll to bool in this patch and make second patch fixing returned value.

sreichel: > 
> 
> I suggest - fix the boll to bool in this patch and make second patch fixing returned value.

No. These PRs only reflects current behavior. No logic changes.

- merge this
- open ticket to fix it + docs 

????

tmotyl: I would not like to make a change which will suggest that the correct type is "false|string", which we will change soon again.
So I maybe skip changing this line (keep the old broken boll), to keep this patch a "nobrainer", and then open a new PR which fixes the return type and doc comment.
Is that ok for you?

sreichel: > I would not like to make a change which will suggest that the correct type is "false|string", which we will change soon again

It is not about to suggest something. DOC should reflect the current behavoir, NOT what seems to be intenional.

Please leave a note at #784 and i'll fix it in time after the other PRs are merged ;)

sreichel: I better agree and fix id later ... thanks @tmotyl @Flyingmana

PR #760, opened 2019-06-22
Mage_Payment - DOC block update

sreichel:
- doc blocks added/fixed

### DIFF
--- a/app/code/core/Mage/Payment/Block/Form/Container.php
+++ b/app/code/core/Mage/Payment/Block/Form/Container.php
@@ -45,7 +45,7 @@ class Mage_Payment_Block_Form_Container extends Mage_Core_Block_Template
          */
         foreach ($this->getMethods() as $method) {
             $this->setChild(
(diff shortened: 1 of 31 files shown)


app/code/core/Mage/Payment/Block/Form/Container.php
```diff
@@ -45,8 +45,8 @@ protected function _prepareLayout()
          */
         foreach ($this->getMethods() as $method) {
             $this->setChild(
-               'payment.method.'.$method->getCode(),
-               $this->helper('payment')->getMethodFormBlock($method)
+                'payment.method.'.$method->getCode(),
```
damien-biasotto, line 48: Ideally, there should a space before and after the concatenation operator i.e : `'payment.method.' . $method->getCode(),`

colinmollenhour:
```suggestion
                'payment.method.' . $method->getCode(),
```

PR #762, opened 2019-06-22
Mage_Page - DOC block update

sreichel:
- doc blocks added/fixed


sreichel:
Bump. ;)

app/code/core/Mage/Page/Block/Html/Footer.php
```diff
@@ -62,12 +62,19 @@ public function getCacheKeyInfo()
...
+     * @return $this
+     */
     public function setCopyright($copyright)
     {
         $this->_copyright = $copyright;
         return $this;
     }
 
+    /**
+     * @return mixed
```
tmotyl, line 76: Return string? Also the propert misses a type

app/code/core/Mage/Page/Block/Html/Toplinks.php
```diff
@@ -60,16 +60,16 @@ function __construct()
      * @param string|array $liParams
      * @param string|array $aParams
      * @param string $innerText
-     * @param int $position
+     * @param string $position
```
tmotyl, line 63: Int/string - see usage inside function

PR #766, opened 2019-06-23
Mage_Media - mini DOC block update

sreichel:
- doc blocks added/fixed


sreichel:
Small one :)

app/code/core/Mage/Media/Model/File/Image.php
```diff
@@ -111,7 +126,7 @@ public function getImage(Mage_Media_Model_Image $object)
      * Create tmp image resource for operations
      *
      * @param Mage_Media_Model_Image $object
-     * @return $this
+     * @return false|resource
```
kiatng, line 129: Shouldn't it be `@return $this`?

app/code/core/Mage/Media/Model/Image.php
```diff
@@ -72,7 +74,7 @@ protected function _construct()
      * Set media image config instance
      *
      * @param Mage_Media_Model_Image_Config_Interface $config
-     * @return unknown
+     * @return Mage_Media_Model_Image
```
kiatng, line 77: Shouldn't it be `@return $this`?

app/code/core/Mage/Media/Model/File/Image.php
sreichel: Ne, it returns `        $resource = imagecreatetruecolor($object->getDestanationDimensions()->getWidth(), $object->getDestanationDimensions()->getHeight());
        return $resource;`

app/code/core/Mage/Media/Model/Image.php
sreichel: Thsi was right, but collin merged some seconds ago ... will be fixed in "Round 2".

app/code/core/Mage/Media/Model/File/Image.php
kiatng: Yes, indeed. Thanks!

app/code/core/Mage/Media/Model/Image.php
kiatng: Thanks for your effort.

PR #767, opened 2019-06-23
Mage_Log - DOC block update

sreichel:
- doc blocks added/fixed

### DIFF
--- a/app/code/core/Mage/Log/Model/Log.php
+++ b/app/code/core/Mage/Log/Model/Log.php
@@ -58,6 +58,9 @@ class Mage_Log_Model_Log extends Mage_Core_Model_Abstract
         $this->_init('log/log');
     }
 
+    /**
+     * @return float|int
+     */
     public function getLogCleanTime()
     {
         return Mage::getStoreConfig(self::XML_LOG_CLEAN_DAYS) * 60 * 60 * 24;
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -24,19 +24,44 @@
  * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-
 /**
  * Enter description here ...
  *
  * @method Mage_Log_Model_Resource_Visitor getResource()
- * @method string getSessionId()
- * @method Mage_Log_Model_Visitor setSessionId(string $value)
- * @method Mage_Log_Model_Visitor setFirstVisitAt(string $value)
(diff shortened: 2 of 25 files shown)


app/code/core/Mage/Log/Model/Log.php
```diff
@@ -58,6 +58,9 @@ protected function _construct()
         $this->_init('log/log');
     }
 
+    /**
+     * @return float|int
```
tmotyl, line 62: why float? the config description is "Save Expired Items Days" and default value is "180"

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -158,6 +187,9 @@ public function getUrl()
         return $url;
     }
 
+    /**
+     * @return mixed
```
tmotyl, line 191: why mixed? I wold expect rather int

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -166,6 +198,9 @@ public function getFirstVisitAt()
         return $this->getData('first_visit_at');
     }
 
+    /**
+     * @return mixed
```
tmotyl, line 202: same here

app/code/core/Mage/Log/Model/Log.php
sreichel: 'Mage::getStoreConfig(self::XML_LOG_CLEAN_DAYS)' returns a 'string' ... multiplied its float.

tmotyl: ok, so we need a casting to int here.
I propose we either skip this docbloc change, and open a new PR fixing the docblock and adding cast.

app/code/core/Mage/Log/Model/Visitor.php
sreichel: I've used mixed only in a very few cases, where multiple return types are possble. Mixed/object are on todo list as mentioned in #784.

app/code/core/Mage/Log/Model/Visitor.php
sreichel: See above.

app/code/core/Mage/Log/Model/Log.php
tmotyl: created an issue for this change https://github.com/OpenMage/magento-lts/issues/949

tmotyl:
thanks @sreichel ! :tada:

sreichel:
> 
> 
> thanks @sreichel ! tada

I am happily surprised :)

Didnt check ... if this is discussed in discord plmk.

PR #769, opened 2019-06-23
Mage_ImportExport - DOC block update

sreichel:
- doc blocks added/fixed


app/code/core/Mage/ImportExport/Model/Import/Uploader.php
```diff
@@ -43,6 +43,10 @@ class Mage_ImportExport_Model_Import_Uploader extends Mage_Core_Model_File_Uploa
     );
     const DEFAULT_FILE_TYPE = 'application/octet-stream';
 
+    /**
+     * Mage_ImportExport_Model_Import_Uploader constructor.
+     * @param null $filePath
```
kkrieger85, line 48: This should be "string"?

kkrieger85: Or string|null

tmotyl, line 48:
```suggestion
     * @param string|null $filePath
```

tmotyl: right

PR #776, opened 2019-06-23
Mage_CatalogSearch - DOC block update

sreichel:
- doc blocks added/fixed

### DIFF
--- a/app/code/core/Mage/CatalogSearch/Block/Term.php
+++ b/app/code/core/Mage/CatalogSearch/Block/Term.php
@@ -34,7 +34,7 @@
 class Mage_CatalogSearch_Block_Term extends Mage_Core_Block_Template
 {
     /**
-     * @var Mage_CatalogSearch_Model_Resource_Query_Collection
+     * @var array
      */
     protected $_terms;
     protected $_minPopularity;
@@ -88,7 +88,7 @@ class Mage_CatalogSearch_Block_Term extends Mage_Core_Block_Template
     }
 
     /**
-     * @return Mage_CatalogSearch_Model_Resource_Query_Collection
+     * @return array
      */
     public function getTerms()
     {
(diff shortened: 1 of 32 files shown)


app/code/core/Mage/CatalogSearch/Block/Term.php
```diff
@@ -83,12 +87,19 @@ protected function _loadTerms()
         return $this;
     }
 
+    /**
+     * @return Mage_CatalogSearch_Model_Resource_Query_Collection
```
tmotyl, line 91: isn't it array?
```
 foreach ($termKeys as $termKey) {
                $this->_terms[$termKey] = $temp[$termKey];
            }
```

app/code/core/Mage/CatalogSearch/Model/Layer/Filter/Attribute.php
```diff
@@ -33,12 +33,11 @@ class Mage_CatalogSearch_Model_Layer_Filter_Attribute extends Mage_Catalog_Model
     /**
      * Check whether specified attribute can be used in LN
      *
-     * @param Mage_Catalog_Model_Resource_Eav_Attribute  $attribute
-     * @return bool
+     * @param Mage_Catalog_Model_Resource_Eav_Attribute $attribute
+     * @return Mage_Catalog_Model_Resource_Eav_Attribute
```
tmotyl, line 37: bool was correct

sreichel: This is wrong (proberbly copied param to return), but ...

- in DB its tinynt/bool
- parrents `@method` say `int`
- in code ìnt` is used

I'll revert, but tis has to be chekced again. (W'd love to have strict types :P)

PR #778, opened 2019-06-23
Mage_Admin - DOC block update

sreichel:
- doc blocks added/fixed

### DIFF
--- a/app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
+++ b/app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
@@ -53,7 +53,7 @@ class Mage_Admin_Model_Acl_Assert_Ip implements Zend_Acl_Assert_Interface
     }
 
     /**
-     * @param bool $ip
+     * @param string|false $ip
      */
     protected function _isCleanIP($ip)
     {
--- a/app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
+++ b/app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
@@ -37,18 +37,24 @@ class Mage_Admin_Model_Acl_Assert_Ip implements Zend_Acl_Assert_Interface
     /**
      * Check whether ip is allowed
      *
-     * @param Zend_Acl $acl
-     * @param Zend_Acl_Role_Interface $role
-     * @param Zend_Acl_Resource_Interface $resource
+     * @param Mage_Admin_Model_Acl $acl
+     * @param Mage_Admin_Model_Acl_Role $role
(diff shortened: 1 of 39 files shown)


app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
```diff
@@ -37,18 +37,24 @@ class Mage_Admin_Model_Acl_Assert_Ip implements Zend_Acl_Assert_Interface
...
+        Mage_Admin_Model_Acl $acl,
+        Mage_Admin_Model_Acl_Role $role = null,
+        Mage_Admin_Model_Acl_Resource $resource = null,
+        $privilege = null
+    ) {
         return $this->_isCleanIP(Mage::helper('core/http')->getRemoteAddr());
     }
 
+    /**
+     * @param bool $ip
```
tmotyl, line 56: this should be string as Mage::helper('core/http')->getRemoteAddr() returns string

tmotyl: maybe return should be bool

app/code/core/Mage/Admin/Model/Acl/Assert/Time.php
```diff
@@ -27,28 +27,34 @@
...
+        Mage_Admin_Model_Acl $acl,
+        Mage_Admin_Model_Acl_Role $role = null,
+        Mage_Admin_Model_Acl_Resource $resource = null,
+        $privilege = null
+    ) {
         return $this->_isCleanTime(time());
     }
 
+    /**
+     * @param bool $time
```
tmotyl, line 56: $time is not bool, but return type yes

app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
sreichel: Mhhh ... this methode is empty ....  How abaout ...

- `@param`: `Mage::helper('core/http')->getRemoteAddr()` returns `string` or `false` ... 
- `@return void`

```
    protected function _isCleanIP($ip)
    {
        // ...
    }
```

app/code/core/Mage/Admin/Model/Acl/Assert/Time.php
sreichel: Similar to above ....

app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
tmotyl: ok, seems the classes are not used in the core anyway

PR #781, opened 2019-06-23
Mage_CatalogIndex - DOC block update

sreichel:
- doc blocks added/fixed

### DIFF
--- a/app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
+++ b/app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
@@ -49,7 +49,7 @@ class Mage_CatalogIndex_Model_Data_Abstract extends Mage_Core_Model_Abstract
     /**
      * Defines when product type has children
      *
-     * @var array<int, bool>
+     * @var int[]|bool[]
      */
     protected $_haveChildren = array(
         Mage_CatalogIndex_Model_Retreiver::CHILDREN_FOR_TIERS => true,
@@ -150,7 +150,7 @@ class Mage_CatalogIndex_Model_Data_Abstract extends Mage_Core_Model_Abstract
     /**
      * Fetch final price for product
      *
-     * @param array $product
+     * @param int $product
      * @param Mage_Core_Model_Store $store
      * @param Mage_Customer_Model_Group $group
      * @return float
--- a/app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
+++ b/app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
@@ -31,6 +31,11 @@
  * @category   Mage
  * @package    Mage_CatalogIndex
  * @author     Magento Core Team <core@magentocommerce.com>
+ *
+ * @method Mage_CatalogIndex_Model_Resource_Data_Abstract getResource()
+ *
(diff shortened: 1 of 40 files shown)


app/code/core/Mage/CatalogIndex/Model/Data/Configurable.php
```diff
@@ -34,13 +34,13 @@ class Mage_CatalogIndex_Model_Data_Configurable extends Mage_CatalogIndex_Model_
     /**
      * Defines when product type has children
      *
-     * @var boolean
+     * @var array<int, bool>
```
colinmollenhour, line 37: Not a blocker, but I don't think PHPStorm recognizes this syntax. However, it would recognize `int[]|bool[]`.

sreichel:
@colinmollenhour Thanks for review .... please note all minor issues at #784 ... i'll update later.

app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
```diff
@@ -118,6 +123,7 @@ public function getParentProductIds($store, $childIds)
      * @param array $settings
      * @param int $type
      * @param int|array $suppliedId
+     * @return mixed
```
tmotyl, line 126: it should be array as fetchLinkInformation returns array (it has a wrong doc comment, but it always returns $this->_getWriteAdapter()->fetchCol($this->_getLinkSelect()); and fetch call returns an array

sreichel: @tmotyl ... `getParentProductIds()` return either `false` or `fetchLinkInformation()`, that returns `array` indeed.

So it should be `array|false`? Can you please approve and add a note to #784? `mixed` will be checked later!

tmotyl: I think you're looking at wrong method.
This method `fetchLinkInformation` has only one return `        return $this->getResource()->fetchLinkInformation($store->getId(), $settings['table'], $idField, $whereField, $suppliedId, $additional);`
Which always returns an array.
Please fix it in this patch.

The other function `getChildProductIds` (which also has "mixed" return type, should have, as you said "array|false"
for this issue I've added a coment in #784 as can be fixed later.

tmotyl:
```suggestion
     * @return array
```

app/code/core/Mage/CatalogIndex/Model/Data/Configurable.php
sreichel: I thought it works too ... however , changed it ...

app/code/core/Mage/CatalogIndex/Model/Data/Abstract.php
```diff
@@ -144,7 +150,7 @@ protected function fetchLinkInformation($store, $settings, $type, $suppliedId)
     /**
      * Fetch final price for product
      *
-     * @param int $product
+     * @param array $product
```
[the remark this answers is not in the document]
sreichel: Just reverted to `int` ... if `array` is possible, it will pop up again later :)

PR #783, opened 2019-06-23
Mage_Catalog - DOC block update

sreichel:
- doc blocks added/fixed


app/code/core/Mage/Catalog/Block/Layer/View.php
```diff
@@ -105,7 +105,7 @@ protected function _getAttributeFilterBlockName()
     /**
      * Prepare child blocks
      *
-     * @return $this
+     * @inheritDoc
```
tmotyl: what is the benefit of this change?

sreichel: `@return $this` means an instance of `Mage_Catalog_Block_Layer_View `, but the mehtod returns `Mage_Core_Block_Template` (see `return parent::_prepareLayout();`)

colinmollenhour: But `parent::_prepareLayout()` just returns `$this` so `@return $this` is  still correct - the return value is an instance of `Mage_Catalog_Block_Layer_View`.

So both ways are correct but it seems the `@inheritDoc` is used very infrequently. For consistency and simplicity sake I think we should not adopt the use of `@inheritDoc`. Also while `@inheritDoc` may be functionally equivalent to IDEs and doc generators, I think it is not preferred for humans because it just requires another navigation in the IDE to find the full docs.

sreichel: @colinmollenhour You're "wrong" in this case. `@return $this` in a DOC block expects the **exact** current class to be returned. So `$this` for `Mage_Catalog_Block_Layer_View` does not equals  `$this` in `Mage_Core_Block_Template`.

For every method that returns `parent::something()` `@inheritDoc` should be used.

(seems to be a new feature, but also PhpStorm gives a warning for wrong return type ... )

sreichel: Btw ... thanks for review @tmotyl :)

colinmollenhour: @sreichel I believe it is you who is wrong in this case. 😃  Here is an example showing that `@return $this` correctly resolves the auto-completion based on the class instantiated in PhpStorm 2019.2 (although it has worked this way for as long as I can remember):

image

However, I do think there is a bug in PhpStorm because it does hint there is an error when `return parent::...` is used even though this is perfectly valid and the return value is the same (it doesn't matter what method used `return $this`, the `$this` refers to an instance of `B` in this case so the IDE should treat it as such.

image

Flyingmana: Interesting case. 
hint: phpstorm is not always reliable for special cases, as they also analyze the sourcecode itself for knowing the return type, not sure if they only warn on a conflict, or also ignore the docblock.

Instead of changing this place, we should change the docblock of the base definition of this method to "$this"

We can then keep the current $this here, or inherite, its not making a difference.

for future reference when to use $this, self, static https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#keyword
PSR-5 never was accepted, but its the best I found for defining this.

@sreichel as you can see, $this is refering to the current class in the current context, not the current class where it was defined. You seem to mix this up with static or maybe self, according to PSR-5

Or do you have another documentation which is more up-to-date?

sreichel: > Here is an example showing that `@return $this` correctly resolves the auto-completion based on the class instantiated in PhpStorm 2019.2

PhpStorms auto-completion works completly independed of annotations. 

> However, I do think there is a bug in PhpStorm because it does hint there is an error when `return parent::...` is used even though this is perfectly valid and the return value is the same (it doesn't matter what method used `return $this`, the `$this` refers to an instance of `B` in this case so the IDE should treat it as such.

From "code view" it it perfectly valid of course. From code analysis like phpstan it is not. For me it is pretty obvious ... :)

For your example .... if a CA tool reads your class `B::foo()` it expects an instance of class B returned. In other word, the CA tool does not know from annotion, that the return value comes from parent class. The doc block dosn't tell it.

If you use `@inheritDoc` it uses parents doc block information ... and that matches `parent::...`.

Not sure on this, but PhpStorm 2019.1 (or at least 2018.x) didn't hint this as error. Does it disappear when you use `inheritDoc`?

> Instead of changing this place, we should change the docblock of the base definition of this method to "$this"

Changing base class doc block to `$this` is mandatory (and should be already done ... other PR?). If parent class uses a FQCN in doc, all inherited methods returning `$this` will throw errors. (missmatch of child-parent return type definition)

> You seem to mix this up with static or maybe self, according to PSR-5
> 
> Or do you have another documentation which is more up-to-date?

`static` would be wrong and `self` would give the same (wrong) result.

Inheritance is mentioned here: https://docs.phpdoc.org/guides/inheritance.html As far as i know, phpdoc follows latest PSR-5 draft.

However ... the idea behind all these PRs was to make LTS work with static code analyzation tools like phpstan. You'll find this changes in a lot of PRs.

colinmollenhour: When the parent class has `@return $this` and the super class has `@return $this`, how is this different than `@inheritDoc`? If PHPStan doesn't work with `@return $this` then it seems that is a bug or serious limitation in PHPStan? In my example, `B#foo()` returns an instance of `B`, not of `A` so `@return $this` is not incorrect.

Can you provide an example of how PHPStan gives an incorrect result without `@inheritDoc` in the super class?

sreichel: Does Magentos parent or super classes have/had return type `$this`?

There is nothing wrong with phpstan (other tools work the same way). It just compares the promised return type from annotation with the real one ... 

> In my example, `B#foo()` returns an instance of `B`, not of `A` so `@return $this` is not incorrect.

**No.**  `B#foo()` returns  `parent::foo()` that has declared return type `$this`. This `$this` referrs to class `A`. The `$this` annotation of `B#foo()` expects a class `B` object.

Even PhpStorm says "Return value is expected to be 'B', 'A' returned"

Just a add strict type declaration to your example ...

```
declare(strict_types=1);

class A
{
    /**
     * @return $this
     */
    public function foo(): A
    {
        return $this;
    }
}

class B extends A
{
    /**
     * @return $this
     */
    public function foo(): B
    {
        return parent::foo();
    }
}
```

> Declaration of B::foo(): B must be compatible with A::foo(): A

sreichel: I dont know how to explain ... `$this` in a doc block isn't the same as is code :(

Flyingmana: but the PSR standard says `$this` in a doc block is the same as in code

Flyingmana: > No. B#foo() returns parent::foo() that has declared return type $this. This $this referrs to class A.

what you refer to here is not the docblock `$this`, but the docblock `self`

sreichel: `self` also referres to class `B` ... `$this` or `self` would be correct if the method uses `return $this`.

Example:

```
class A
{
    /**
     * @return A
     */
    public function foo()
    {
        return $this;
    }
}

class B extends A
{
    /**
     * @inheritDoc
     */
    public function foo()
    {
        return parent::foo();
    }
}

class C extends B
{
    /**
     * @return $this
     */
    public function foo()
    {
        return parent::foo();
    }
}

```
A: correct
B: correct
C: incorrect (expected C, got A)

If you change `@return A` to `@return $this` 

A: correct
B: correct
C: incorrect (expected C, got B)

If you `@return $this`  for all

A: correct
B: incorrect (expected B, got A)
C: incorrect (expected C, got B)

Use `@inheritDoc` fo B & C and its all correct.

colinmollenhour: An instance of a class can only have one type, so `return $this` is always returning the same instance regardless of what class the function returning it is declared in.

In your last example, the thing is, an instance of C is still an instance of C even if a method was not overridden calls `return $this`. Your static type example that you deleted was incorrect, it should have been:

```php
<?php
declare(strict_types=1);

class A
{
    /**
     * @return $this
     */
    public function foo(): A
    {
        return $this;
    }
}

class B extends A
{
    /**
     * @return $this
     */
    public function foo(): A
    {
        return parent::foo();
    }

    public function bar()
    {
        return 1;
    }
}

$b = new B;
$b->foo()->bar();
```

The above is valid. In PhpStorm if you change `B#foo`'s vardoc to `@inheritDoc` then PhpStorm does not recognize the call to `->bar()` on the last line and warns "Method 'bar' not found in A".

tmotyl: Hi little update:
following snippet is all OK, in PHPStorm 
```
<?php
class A
{
    /**
     * @return $this
     */
    public function foo()
    {
        return $this;
    }
}

class B extends A
{
    /**
     * @return $this
     */
    public function foo()
    {
        return parent::foo();
    }
}

class C extends B
{
    /**
     * @return $this
     */
    public function foo()
    {
        return parent::foo();
    }
}
```
So can we agree on something and continue with the topic?

colinmollenhour:
I think PHPStorm 2020.1 has gotten a little smarter, it now recognizes the auto-completion better in @sreichel's example. It recognizes `(new B)->foo()` as returning `A|B`. However, using `@return $this` everywhere it says `(new B)->foo()` only returns `B` which I believe is more accurate (although not significantly so).

However, using @sreichel's example, PHPStorm 2020.1 flags a warning on `C`:

image

Using only `@return $this` there are no errors.

In my opinion we should always use `@return $this` when a method effectively does `return $this;` and not use `@inheritDoc` since the IDE's just are not as smart with handling it.

tmotyl:
I agree with @colinmollenhour, especially as `return $this` is understandable for a human, whithout the need to dig deeper, while inheridoc is not.

sreichel:
> However, using @sreichel's example, PHPStorm 2020.1 flags a warning on `C`:

Yes, it does. Just the tooltip is smarter.

> In my opinion we should always use `@return $this` when a method effectively does `return $this;` and not use `@inheritDoc` since the IDE's just are not as smart with handling it.

Absolutely agree, but in this case its not `return $this`, it is `parent::some()` ....

> But parent::_prepareLayout() just returns $this so @return $this is still correct - the return value is an instance of Mage_Catalog_Block_Layer_View.

----

I have used `@inheritDoc` only  if return value calls `parrent::x()`, for `return $this;` i've used  `@return $this`!

PLEASE leave it as it is for now.

tmotyl:
I've resolved the conflict

tmotyl:
conflict resolved once again

tmotyl:
please revote

app/code/core/Mage/Catalog/Model/Product.php
```diff
@@ -607,7 +843,7 @@ protected function _afterLoad()
     /**
      * Retrieve resource instance wrapper
      *
-     * @return Mage_Catalog_Model_Resource_Eav_Mysql4_Product
+     * @@inheritDoc
```
colinmollenhour, line 846: Double @
```suggestion
     * @inheritDoc
```

app/code/core/Mage/Catalog/Model/Product/Action.php
```diff
@@ -46,7 +46,7 @@ protected function _construct()
     /**
      * Retrieve resource instance wrapper
      *
-     * @return Mage_Catalog_Model_Resource_Eav_Mysql4_Product_Action
+     * @inheritDoc
```
colinmollenhour, line 49: This is a step backward for type inference. `@inheritDoc` will mean that all methods added by `Mage_Catalog_Model_Resource_Eav_Mysql4_Product_Action` will no longer be detected by the IDE because the parent method's docs say the return is `Mage_Core_Model_Mysql4_Abstract`.

tmotyl: right. So what do you suggest? 
do 
```suggest
     * @return Mage_Catalog_Model_Resource_Product_Action
```
?
Probably other places need to be rechecked

sreichel: I'd leave it as-is ...

- `Mage_Catalog_Model_Resource_Eav_Mysql4_Product_Action` is an empty deprecated method
- `return parent::_getResource();` uses return type from parent DOCs ... so it seems correct
- `@return Mage_Core_Model_Mysql4_Abstract` should be changed to resource (later)

Problem is `Mage_Core_Model_Abstract::_getResource()` ... 
```
return Mage::getResourceSingleton($this->_resourceName);
```
... neither IDE nor phpstan can detect correct child-model of `Mage_Core_Model_Mysql4_Abstract`.

PR #596, opened 2019-01-26
Fix <br/> in gallery header

nintenic:
#553 Keeps the escapehtml() by adding a str_replace() to convert back just the `<br/>` tag.

Issue #786, opened 2019-06-26
Patch SUPEE-11155/Magento 1.9.4.2


cundd:
Magento published version 1.9.4.2 including SUPEE-11155

cundd:
See pull request on magento-mirror: https://github.com/OpenMage/magento-mirror/pull/75

Flyingmana:
as usuall, we will merge from the Mirror as soon as it did go through the merge process there and was approved.

cundd:
@Flyingmana are my issues not helpful? I thought that it may help to have it noted here

Flyingmana:
it is usefull :)
I just wanted to add a status, before someone puts work on creating a separate PullRequest, which then may be less helpful for the existing process

wytzejan:
Just want to notify that OpenMage/magento-mirror#75 is merged and done! :-)

cundd:
Can this issue be closed? I think I've overseen the update.

Flyingmana:
update was done with https://github.com/OpenMage/magento-lts/pull/791
and no complains yet, so I think its done, so will close this Issue here

PR #788, opened 2019-06-29
Revert change that was for 1.9.3.10 and is not applicable to 1.9.4.x

ProxiBlue:
ref: https://github.com/OpenMage/magento-lts/pull/596


ProxiBlue:
Will be resolved with https://github.com/OpenMage/magento-lts/issues/786

PR #790, opened 2019-07-05
Fixed currency convert exception

rafdol:
Parameter `$toCurrency` can be `null` or `string` or `Mage_Directory_Model_Currency` object. If `$toCurrency` is `string` and rate doen't exists then we have fatal error (`Call to a member function ... on string` against exception. This is fix for described problem.

### DIFF
--- a/app/code/core/Mage/Directory/Model/Currency.php
+++ b/app/code/core/Mage/Directory/Model/Currency.php
@@ -183,15 +183,22 @@ class Mage_Directory_Model_Currency extends Mage_Core_Model_Abstract
     {
         if (is_null($toCurrency)) {
             return $price;
-        } else {
-            $rate = $this->getRate($toCurrency);
-            if ($rate) {
-                return $price * $rate;
-            }
         }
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Seems this could have been fixed with impact to only one line of code using something like:

`$toCurrency instanceof Mage_Directory_Model_Currency ? $toCurrency->getCode() : $toCurrency`

sreichel:
Nice, this was on my to-do list, too.

dariuszmaciejewski:
@colinmollenhour class names are pretty long in Magento 1, so I am not sure your suggestion is easier to read. Especially if a method call (getter) occurs.

colinmollenhour:
> @colinmollenhour class names are pretty long in Magento 1, so I am not sure your suggestion is easier to read. Especially if a method call (getter) occurs.

Sure, but the main motivation is +1,-1  lines changed instead of +13,-6. Fewer lines changed, less chance for regression or future conflict.

PR #792, opened 2019-07-12
Fix region API list response - Update Api.php --> bug

kyansi:
Region list request returns null if there is no name value. This way we return default_name instead if it exists.

### DIFF
--- a/app/code/core/Mage/Directory/Model/Region/Api.php
+++ b/app/code/core/Mage/Directory/Model/Region/Api.php
@@ -53,8 +53,8 @@ class Mage_Directory_Model_Region_Api extends Mage_Api_Model_Resource_Abstract
 
         $result = array();
         foreach ($country->getRegions() as $region) {
-            $region->getName();
-            $result[] = $region->toArray(array('region_id', 'code', 'name'));
+            $name = $region->getName();
+            $result[] = $region->toArray(array('region_id', 'code', $name));
         }
 
         return $result;
--- a/app/code/core/Mage/Directory/Model/Region/Api.php
+++ b/app/code/core/Mage/Directory/Model/Region/Api.php
@@ -53,8 +53,11 @@ class Mage_Directory_Model_Region_Api extends Mage_Api_Model_Resource_Abstract
 
         $result = array();
         foreach ($country->getRegions() as $region) {
-            $region->setName($region->getName());
-            $result[] = $region->toArray(array('region_id', 'code', 'name'));
+            $result[] = array(
+                'region_id' => $region->getRegionId(),
+                'code' => $region->getCode(),
+                'name' => $region->getName(), //use the logic of default name
+            );
         }
 
         return $result;
--- a/app/code/core/Mage/Directory/Model/Region/Api.php
+++ b/app/code/core/Mage/Directory/Model/Region/Api.php
@@ -53,8 +53,8 @@ class Mage_Directory_Model_Region_Api extends Mage_Api_Model_Resource_Abstract
 
         $result = array();
         foreach ($country->getRegions() as $region) {
-            $name = $region->getName();
-            $result[] = $region->toArray(array('region_id', 'code', $name));
+            $region->setName($region->getName());
+            $result[] = $region->toArray(array('region_id', 'code', 'name'));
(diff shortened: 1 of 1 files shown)


colinmollenhour:
`toArray` takes an array of keys from which the values should be returned, so changing `'name'` to `$name` is not going to have the effect I believe you are looking for.

kyansi:
In my observations $this->getName() triggers a method in magento-lts/app/code/core/Mage/Directory/Model/Region.php.
The method checks whether region has data "name" and gets "default_name" in case if "name" is null empty.
We save this value to variable and use it to pass "name" or "default_name" value from region to toArray function.
This solution is implemented and working.

kyansi:
Please refer to updated commit.
We set "name" value from $this->getName() which triggers a method in magento-lts/app/code/core/Mage/Directory/Model/Region.php.
The method checks whether region has data "name" and gets "default_name" in case if "name" is empty.
We save this value to "name" and use it when calling toArray.
This solution is implemented and working.

midlan (dismissed):

Sekiphp:
Please add short comment to class, because it's not self documenting code :-)

kkrieger85:
How can I test this PR?

spinsch:
> How can I test this PR?

```php
<?php

/* 
 * you need to create an user and group
 * 
 * System > Webservices > SOAP/XML-RPC - User
 * System > Webservices > SOAP/XML-RPC - Roles
 */

$apiV1 = "http://openmage.localhost/api/soap/?wsdl=1";

$username = 'spinsch';
$password = 'spinsch';

$cli = new SoapClient($apiV1);

$session = $cli->login($username, $password);

// the call name can be found under app/code/core/Mage/Directory/etc/api.xml
$result = $cli->call($session, 'directory_region.list', array('CH'));

var_dump($result);
```

reference: https://devdocs.magento.com/guides/m1x/api/soap/introduction.html

app/code/core/Mage/Directory/Model/Region/Api.php
```diff
@@ -53,8 +53,8 @@ public function items($country)
 
         $result = array();
         foreach ($country->getRegions() as $region) {
-            $name = $region->getName();
-            $result[] = $region->toArray(array('region_id', 'code', $name));
+            $region->setName($region->getName());
```
uni-t: wtf is this.

Using setter to store dta from own getter function?

colinmollenhour: The getter is overridden to return either the 'name' or the 'default_name' if the name is null. Probably at one point in time the getter would update the name itself which is why it was called without using the value, which I think is probably a worse infraction. The equivalent code for the fix without calling a setter or relying on mysterious behavior would be something like this:

```php
$regionData = $region->toArray(['region_id', 'code', 'name']);
if (empty($regionData[2])) {
    $regionData[2] = $region->getDefaultName();
}
$result[] = $regionData;
```

Or this would probably also work:

```php
$result[] = array_filter($region->toArray(['region_id', 'code', 'name', 'default_name']));
```

midlan: @colinmollenhour good point, but to respect the original implementation of getName, the code it should look like:

```php
$regionData = $region->toArray(['region_id', 'code', 'name']);
if (is_null($regionData['name'])) {
    $regionData['name'] = $region->getData('default_name');
}
$result[] = $regionData;
```
**But it's not ideal too, the logic of the getName is cloned to other place**. At the moment, the best solution to me would be:

```php
foreach ($country->getRegions() as $region) {
    $result[] = array(
        'region_id' => $region->getRegionId(),
        'code' => $region->getCode(),
        'name' => $region->getName(), //use the logic of default name
    );
}
```

ADDISON74:
If the author is not changing the PR @midlan or @colinmollenhour could you please edit the file?

We have so many PR's where no changes are made by the authors because they no longer have activity on GitHub. Either we close and create new ones or we just edit them.

midlan:
@ADDISON74 file edited

kiatng (approved):

ADDISON74 (approved):

PR #796, opened 2019-08-05
Replacing Admin Notifications feed URL

netalico:
This will allow us to control the Feed so that only relevant/important notifications are displayed to merchants.

### DIFF
--- a/app/code/core/Mage/AdminNotification/etc/config.xml
+++ b/app/code/core/Mage/AdminNotification/etc/config.xml
@@ -91,7 +91,7 @@
     <default>
         <system>
             <adminnotification>
-                <feed_url>notifications.magentocommerce.com/community/notifications.rss</feed_url>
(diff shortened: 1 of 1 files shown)


kkrieger85:
@colinmollenhour @Flyingmana Could you please review?

Flyingmana:
could please someone check how it works and post a screenshot here?

spinsch:
> could please someone check how it works and post a screenshot here?

it works when the flag "use_https" is set.
http request returns 301 to redirect https and this will broken the feed update.

Configuration___System___Magento_Admin

Flyingmana:
ahh yes, I enforce https on the website (for reasons)
is Magento using http by default?

the ReadMore is a link to the post, or showing post content inside Magento?

but in general, seems to work, thats good :)

spinsch:
yes magento use "http" by default.

"Read detail" goto => https://openmage.github.io/Web_Notifications//2019/06/24/OpenMage.html

PR #799, opened 2019-08-13
Force HTTPS for Admin Notifications

netalico:


Sekiphp:
@netalico What is the main reason for forcing? Please write some explanation :-)

netalico:
Security. :) Also, the new feed is only HTTPS so we either need to force it or change the default

netalico:
Any other reviewers? Comments/Issues?

app/code/core/Mage/AdminNotification/Helper/Data.php
```diff
@@ -100,9 +100,7 @@ public function getUnreadNoticeCount($severity)
     public function getPopupObjectUrl($withExt = false)
     {
         if (is_null($this->_popupUrl)) {
-            $sheme = Mage::app()->getFrontController()->getRequest()->isSecure()
-                ? 'https://'
-                : 'http://';
+            $sheme = 'https://';
```
spinsch: can you remove the $sheme var and add https directly ?
```php
$this->_popupUrl = 'https://' . Mage::getStoreConfig(self::XML_PATH_POPUP_URL);
```

netalico:
Done

shineability:
I'm all for security, but this could break (maybe not that well written) third party code...

E.g. We are using Amasty's [Admin Actions Log](https://amasty.com/admin-actions-log.html) module which rewrites `Mage_AdminNotification_Model_Feed` and overrides the `getFeedUrl()` method like this...

``` 
public function getFeedUrl()
{
    if (is_null($this->_feedUrl)) {
        $this->_feedUrl = (Mage::getStoreConfigFlag(self::XML_USE_HTTPS_PATH) ? 'https://' : 'http://') . self::URL_NEWS;
    }
    return $this->_feedUrl;
}
```
The now missing constant `XML_USE_HTTPS_PATH` triggers an error.

mark-netalico:
I added it back in https://github.com/OpenMage/magento-lts/pull/800 because I had the same issue.

PR #802, opened 2019-08-22
Fix widgets layout handle on edit

VincentMarmiesse:
Hi,
There is an issue when you edit a widget: it does not select the right layout handle.

Example, if you create a `core/text_list` block on handle `cms_index_index`:
```
<?xml version="1.0" encoding="UTF-8" ?>
<layout version="0.1.0">
    <cms_index_index>
        <reference name="content">
            <block type="core/text_list" name="home.area" translate="label">
                <label>Homepage area</label>
            </block>
        </reference>
    </cms_index_index>
</layout>
```

The area will appear when you create a widget but not when you edit one. That's because the select is not correct in template file.


julienloizelet (dismissed):
Test ok.

To reproduce bug : 
- create local.xml as suggested
- create a widget and select `Display on specific pages` > `Cms home Page` : you will be able to choose the Homepage area added in local.xml
- But as soon as you save the widget (i.e you are now in edition mode), the option disappears from the select.
- On the PR branch, it works as expected.

colinmollenhour (dismissed):

PR #804, opened 2019-09-06
Add created_at and updated_at to all relevant REST API resources

mattdavenport:
When developing an integration with the Magento 1.9 REST API, I came across an issue in accessing fields `created_at` and `updated_at` for certain resources. Orders by default have `created_at` available, but not `updated_at` which is useful for incremental syncing. In addition, other resources such as `/products` and `/stockitems` have neither `created_at` or `updated_at` fields available to the consumer.

These dates are valuable for applications consuming the API, and should be available as attributes which can be enabled for applicable roles (Admin, Customer). In addition, these values should never be writable, and should only be included in the configuration for resources which have these dates in the database.

This PR adds both `created_at` and `updated_at` fields to all applicable resources, while excluding them from writes via configuration files.

### DIFF
--- a/app/code/core/Mage/Catalog/etc/api2.xml
+++ b/app/code/core/Mage/Catalog/etc/api2.xml
@@ -60,7 +60,7 @@
                         <retrieve>1</retrieve>
                     </guest>
                 </privileges>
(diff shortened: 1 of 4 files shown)


mattdavenport:
Could we get this merged in? We'd like to start tracking magento-lts `master` soon and having this PR merged would be of great help. It appears it has enough reviews, so if there is any other issue please let me know and I'll push a fix. Thanks!

kkrieger85:
Ping @Flyingmana

Issue #807, opened 2019-09-11
Theme data not updated when switching package


azambon:
When one changes the currently active package by invoking `Mage_Core_Model_Design_Package::setPackageName()`, the theme data is not updated or reset.

Suppose that one has a non-default admin theme, `default/customername` and suppose that there is a cronjob that tries to send emails from some specific, non admin store view, let's say store view 1. Also assume that the package and theme of such store view are something like `package1` and `default` respectively.
When the `Mage_Core_Model_Design_Package` is first loaded, it correctly loads the admin configuration with package `default` and theme `customername`. Then, when the code tries to render the email, Magento initiates an app emulation and switches the package to the one configured for store view 1 by calling `Mage_Core_Model_Design_Package::setPackageName('package1')`. The problem is that no one changes the theme data, and so Magento looks for the data in `package1/customername`, which does not exist.
It is true that the app emulation code calls `Mage_Core_Model_Design_Package::setTheme('')`, but such a call does not reset the default theme. And this can cause issues.

fballiano:
I'm closing this issue since we have its PR https://github.com/OpenMage/magento-lts/pull/808

PR #808, opened 2019-09-11
Clears the currently active theme when changing package

azambon:
Fix for issue #807

### DIFF
--- a/app/code/core/Mage/Core/Model/Design/Package.php
+++ b/app/code/core/Mage/Core/Model/Design/Package.php
@@ -173,6 +173,7 @@ class Mage_Core_Model_Design_Package
      */
     public function setPackageName($name = '')
     {
+        $origName = $this->_name;
         if (empty($name)) {
             // see, if exceptions for user-agents defined in config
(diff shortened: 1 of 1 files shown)


azambon:
> However, there is a small risk that someone sets the theme first and then the package in which case this could clobber their theme setting

Good point, technically speaking this PR is a BC break, I hadn't thought about it.
Without this PR, one can set the theme first, then set the package and everything works (if the package and theme form a sensible combination). After merging this PR, one would need to set the package first and then set the theme. Otherwise setting the package would clear the theme.

Offhand I can't see a way to ensure a consistent package-theme pair while also allowing freedom in the order that these values are set.
If anyone has a better solution I'm all ears :)

fballiano:
I'd prefer this PR not to pass, when a developer is switching package/theme is always manually checking everything so I may want to just switch the package and keep the theme as it is. Nobody would expect the system to clear data from another section of the config, it's a behaviour we don't have anywhere else and I doesn't feel natural in my opinion.

colinmollenhour:
@azambon please reopen with justification if you feel it is prudent but given the time passage I'm assuming this is ok to close.

PR #811, opened 2019-09-12
fopen() and fclose()

### LINKED ISSUE #810: fopen() and fclose() missing as public methods for  Varien_Io_File class
Hi, team.
Congrats for the library Varien_Io_File. It has been useful.
On https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Varien/Io/File.php
many functions helps Magento developers to avoid coding standard notifications about using directly **mkdir(), file_exists()**...So I have used this library in order to "clean" our project code. And I arrived to the point where I didn't find any function which allow developers replace **fopen()** and **fclose()** PHP functions.
And maybe it will be necessary a new method which allows to access the resource returned by the new **fopen() function** equivalent.
I think this is a need, there is not an equivalent way to access to the returned value for **fopen()** function. And sometimes, some third party functions/methods use this **resource** as a param.
Thanks.

roberto-ebizmarts:
Hello. This is just an idea for solving #810  closes #810

### DIFF
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -845,4 +845,16 @@ class Varien_Io_File extends Varien_Io_Abstract
     {
         return $this->getCleanPath(dirname($file));
     }
+    
+    public function fopen($fileName, $mode)
+    {
+        $this->_streamHandler = @fopen($fileName, $mode);
+        return $this->_streamHandler;
+    }
+    
+    public function fclose($fileName, $mode)
+    {
+        return $this->streamClose();
+    }
+    
 }
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -846,15 +846,8 @@ class Varien_Io_File extends Varien_Io_Abstract
         return $this->getCleanPath(dirname($file));
     }
     
-    public function fopen($fileName, $mode)
+    public function getStreamHandler()
     {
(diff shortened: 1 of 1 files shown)


roberto-ebizmarts:
Hello, @colinmollenhour 
Thanks for answering.
Please, notice that I don't pretend to re-implement methods that already exist.
I know that streamOpen() and streamClose() exist, but they return boolean values.

I just want access to the result of executing **fopen()** function without using the function itself, directly. Also **fclose()**.
But...why?
    -Well, I use Magento 1 for developing an extension, but **Marketplace** demands to respect **MEQP1 standard** (coding standard for Magento 1) which doesn't allow to use directly some functions like: **curl_exec(), file_exists(), mkdir(), unlink()**. These functions I could replace them thanks to this library (Varien_IO_File) and Varien_Http_Adapter_Curl. I explained this better on #810 
The unique two replacements I have not made has been this only two functions: **fopen()** and **fclose()**. I am trying to get homologue functions which let me replace them.

colinmollenhour:
So why not just add a getter?

```php
    public function getStreamHandler() {
        return $this->_streamHandler;
    }
```

As for `streamClose` it should probably be modified to check for a `false` return value from `fclose` and `return false` if so.

roberto-ebizmarts:
Nice, @colinmollenhour 
I think it's good and simple solution.
Thanks.

PR #813, opened 2019-09-20
Update Emulation.php

Schrank:
https://magento.stackexchange.com/a/25617/217
Emulation doesn't load frontend translation, so we do it.

### DIFF
--- a/app/code/core/Mage/Core/Model/App/Emulation.php
+++ b/app/code/core/Mage/Core/Model/App/Emulation.php
@@ -92,6 +92,8 @@ class Mage_Core_Model_App_Emulation extends Varien_Object
             ->setInitialDesign($initialDesign)
             ->setInitialLocaleCode($initialLocaleCode);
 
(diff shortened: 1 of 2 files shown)


colinmollenhour:
Should this line then be removed so init is not called twice?

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/ProductAlert/Model/Email.php#L258

Schrank:
Ah, thanks! You already have a fix - the problem is, that most likely it doesn't cover all emails, so I decided to put it directly in the emulation.
I can remove all (if there are more than the one) calls or close the PR - as you wish :-)

stijnhau:
seems like a good fix

colinmollenhour:
Can you remove the line https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/ProductAlert/Model/Email.php#L258 and then I think this could be merged.

Issue #741, opened 2019-06-19
Array to string conversion when adding reviews from backend


sreichel:
Repoduce:

- set a filter on entity column in Catalog -> Mange Products grid
- go to Catalog -> Customer Reviews
- add New Review

Expected:

- product list apears

Actually:

- empty product list or array to string conversion when dev mode is on.
- need to reset filter in product grid to add reviews

colinmollenhour:
I would guess that the product grid for adding a review should use a unique grid id to avoid the conflict with the main product grid?

PR #821, opened 2019-10-01
Fixed grid filter reset

### LINKED ISSUE #820: [Bug] Reset filter on adminhml grid does not reset page status
### Preconditions

1. Valet+ 1.0.27, MacOS
2. Nginx
3. PHP 7.0.*
4. OpenMage 1.9.3.10 (Also in 1.9.4.x)

### Steps to reproduce

1. Open a grid with more than one page: Catalog > Products
2. Go to page 2
3. Add filter by ID 1 to 1
4. Trigger the "Search" Button

(5.) Also can trigger the "Reset filter" Button

### Expected result

Go back to the first page

### Actual result

Stays on the current page

spinsch:
Fixes #820 

**Edited:**

The second commit also reset sorting- and direction parameter. Was never reset if flag "_setSaveParametersInSession" is set.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -1289,12 +1289,18 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
     {
         $session = Mage::getSingleton('adminhtml/session');
         $sessionParamName = $this->getId().$paramName;
-        if ($this->getRequest()->has($paramName)) {
-            $param = $this->getRequest()->getParam($paramName);
-            if ($this->_saveParametersInSession) {
-                $session->setData($sessionParamName, $param);
+        $param = $this->getRequest()->getParam($paramName);
+
+        if ($param !== null) {
+            if (trim($param) !== '') {
+                if ($this->_saveParametersInSession) {
+                    $session->setData($sessionParamName, $param);
+                }
+                return $param;
+            } else {
+                $session->unsetData($sessionParamName);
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
```diff
@@ -1289,12 +1289,18 @@ public function getParam($paramName, $default=null)
...
         $session = Mage::getSingleton('adminhtml/session');
         $sessionParamName = $this->getId().$paramName;
-        if ($this->getRequest()->has($paramName)) {
-            $param = $this->getRequest()->getParam($paramName);
-            if ($this->_saveParametersInSession) {
-                $session->setData($sessionParamName, $param);
+        $param = $this->getRequest()->getParam($paramName);
+
+        if ($param !== null) {
+            if (trim($param) !== '') {
```
colinmollenhour: Can param also be an array in this context?

spinsch: no it can't be an array - multiple filter will serialize to a string. sorry for my lazy response :)

colinmollenhour:
I'm pretty sure this wouldn't fix #741 as this affects sorting and paging but that issue is about filters.

fballiano (approved):

ADDISON74:
This PR created trouble here #2177. I suggest to revert this PR and leave it a while for further investigations.

All who upgraded already to versions 1.9.4.16/20.0.14 must replace this file /app/code/core/Mage/Adminhtml/Block/Widget/Grid.php with those from 1.9.4.15/20.0.13.

PR #825, opened 2019-10-08
Fix 404 page - getUrl for unexists store

Sekiphp:
**Problematic usage:**
`echo Mage::getUrl("module/index/policy", array('_store'=>15);`

This calling provides this output (in case when store with id 15 not exists):
Screenshot_1

but I only try to output route to page - not redirecting.....

Problematic passage:
```
    public function setStore($data)
    {
        $this->setData('store', Mage::app()->getStore($data));
        return $this;
    }
```
here is thrown `Mage_Core_Model_Store_Exception` exception without message, which is catched in `Mage` class:

```
        } catch (Mage_Core_Model_Store_Exception $e) {
            require_once(self::getBaseDir() . DS . 'errors' . DS . '404.php');
            die();
        }
```

**Error after this changes:**

Screenshot_2

### DIFF
--- a/app/code/core/Mage/Core/Model/Url.php
+++ b/app/code/core/Mage/Core/Model/Url.php
@@ -361,8 +361,6 @@ class Mage_Core_Model_Url extends Varien_Object
         }
         catch (Mage_Core_Model_Store_Exception $e) {
             Mage::logException($e);
-
-            throw new Mage_Core_Exception("Store with id {$data} was not found!");
         }
 
         return $this;
--- a/app/code/core/Mage/Core/Model/Url.php
+++ b/app/code/core/Mage/Core/Model/Url.php
@@ -362,6 +362,7 @@ class Mage_Core_Model_Url extends Varien_Object
         catch (Mage_Core_Model_Store_Exception $e) {
             // store not found (e.g. on developer server)
             Mage::logException($e);
+            throw new Mage_Core_Exception("Store with id {$data} was not found!");
         }
 
         return $this;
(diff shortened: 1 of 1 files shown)


midlan:
@Flyingmana
It changes behavior from empty 404 this result:
https://user-images.githubusercontent.com/9967016/66381538-d4cdf600-e9b9-11e9-85df-6ddd97bbb90e.png

to this:
https://user-images.githubusercontent.com/9967016/66385324-98ea5f00-e9c0-11e9-8085-6335bc19f243.png

I think the second is much more better for debugging. I agree it can break someone's code expecting the Mage_Core_Model_Store_Exception type.

Maybe it would be better to change this code:
```
        } catch (Mage_Core_Model_Store_Exception $e) {
            require_once(self::getBaseDir() . DS . 'errors' . DS . '404.php');
            die();
        }
```
(it's displaying the nothing-telling 404 error)

colinmollenhour:
I think the code in the description could easily be wrapped with it's own try/catch so this is left up to the user as to how to handle the error. The code code should continue to throw an exception as it does.

midlan:
@colinmollenhour 
The point is that the 404 error is not descriptive at all, so it is hard to debug it. Who would expect that this call:

`echo Mage::getUrl("module/index/policy", array('_store'=>15);`

would end up with the 404 error with any description?

colinmollenhour:
The author of the code in question using a method that must load a store by an arbitrary id that is not guaranteed to exist, should take extra precaution to handle the resulting error in case the store does not actually exist. Just my .02.

Flyingmana:
having the exception logged should already help enough for now.
And there are maybe more people affected by changing the exception, then there are currently affected by the current behavior.

If there is still a strongly need for changing the exception, I would like to split this up, so we can discuss this in a bigger round, when we plan for a release which allows BC breaks.

LukasCesal (approved):

fballiano:
agree with @Flyingmana and @colinmollenhour thus i'm closing this, sorry.

PR #827, opened 2019-10-10
Fix email/admin_new_user_notification template

mickys:
Someone missed a } in the template @vars header line 6 which then results in a "Fix Decoding failed: Syntax error" when loading said template in the "System -> Transactional Emails" area.

Issue introduced in 1.9.3.10.

https://magento.com/tech-resources/bug-tracking/issue/index/id/1994/

This should allow users to load this template in the "System -> Transactional Emails" area.

### DIFF
--- a/app/code/core/Mage/Core/etc/system.xml
+++ b/app/code/core/Mage/Core/etc/system.xml
@@ -1094,6 +1094,15 @@
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
                         </password_reset_link_expiration_period>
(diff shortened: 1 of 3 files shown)


mickys:
Completed functionality by adding the required system field, so admins can now choose a transactional email if they choose to.

Config option already existed.

Config field exposed at 
System -> Configuration -> Advanced -> Admin -> Admin User Emails as implemented by config path admin/emails/admin_notification_email_template 


Admin can now choose a transactional email when enabling
System -> Configuration -> Advanced -> Admin -> Security -> New Admin User Create Notification

PR #834, opened 2019-10-14
Suggested fix to make $quote->getItemById more robust

seansan:
Several new quote and order flows do not always trigger and set the quoteid ... the quote is then created with items but has no id yet ...  and getItemById () does not work in these case because there is a different key. This happens for instance when orders are created via

            /** @var Mage_Sales_Model_Service_Quote $service */
            $service = Mage::getModel('sales/service_quote', $quote);
            $service->submitAll();

Please review and UPDATE my code suggestion directly. Needs a proper review as it is quite in the core of quotes of M1

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -831,7 +831,18 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
      */
     public function getItemById($itemId)
     {
-        return $this->getItemsCollection()->getItemById($itemId);
+        $quoteItem = null;
+        if ($quoteItem = $this->getItemsCollection()->getItemById($itemId)) {
+            return $quoteItem
+        } else {
+            foreach (this->getItemsCollection() as $item) {
+                if ($item->getId() == $itemId) {
+                    $quoteItem = $itemId;
+                    return $quoteItem;
+                } 
+            }
+        }
+        return $quoteItem;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


seansan:
I agree with you. But it does fix a lot of problems or exotic situations
when you cannot grasp why it does not work.
Please correct my if else code if necessary.
Thx sean

joshua-bn:
Scary that this got committed missing those two giant issues that got fixed in the other two commits

henrykbrzoska:
if someone have exception like 
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`magento`.`sales_flat_quote_item_option`, CONSTRAINT `FK_5F20E478CA64B6891EA8A9D6C2735739` FOREIGN KEY (`item_id`) REFERENCES `sales_flat_quote_item` (`item_id`) ON DELETE CASCADE O) 
it's because of this change.
in my example it occurs in IWD order manager when trying to add disabled product to cart

digitalpianism:
@henrykbrzoska same issue here with the Amasty_Promo module with an "Auto add product to cart" rule. the $itemId being null, the comparator isn't correct I'm going to make a PR that I will link here.

Flyingmana:
@henrykbrzoska  @digitalpianism 
Could one of you create an Issue for this and reference it here?

digitalpianism:
@Flyingmana I'm not entirely sure this is an issue.

Basically the module's code look like this:

`$cart->removeItem($item->getId());`

On Magento, as the $item does not have an id, it does nothing.

On Openmage, this piece of code here:

```
foreach ($this->getItemsCollection() as $item) {
                if ($item->getId() == $itemId) {
                    $quoteItem = $item;
                    return $quoteItem;
                } 
            }
```

the condition is true as there's an item without id and the $itemId is null.

I've just added a condition in the module's code as I assumed it was the module badly written rather than this openmage change.

digitalpianism:
@Flyingmana I'm not entirely sure why this change was made to start with even though I read @seansan comment.

If the condition I mentionned above, maybe we should add an "isset" or "!is_null" condition. I dont know which one of the two is supposed to be set.

PR #837, opened 2019-10-15
Add source file support to SFTP wrapper

spinsch:
When writing only a string could be specified as source but no local file.

### DIFF
--- a/lib/Varien/Io/Sftp.php
+++ b/lib/Varien/Io/Sftp.php
@@ -233,4 +233,15 @@ class Varien_Io_Sftp extends Varien_Io_Abstract implements Varien_Io_Interface
         return $list;
     }
 
+    /**
+     * Write a file
+     * @param  string $filename remote filename
+     * @param  string $src local filename
+     * @return boolean
+     */
+    public function writeFile($filename, $src)
+    {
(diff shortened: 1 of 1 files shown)


colinmollenhour:
A separate method like `writeFile()` would be much safer. For example, if you are writing user input then the user could provide a file with the contents "app/etc/local.xml" and then the `is_file` test would return true and now you are uploading the local.xml file to an SFTP server.

spinsch:
Yeah, that's a good reason to reconsider my change :scream:. The parameter $mode (file chmod) is only used with Varien_Io_File::write. Now I changed the interface and made a better alternative to my first commit.

colinmollenhour:
I don't think we want to change the Varien_Io_Interface as that would be a BC breaking change. What is wrong with my suggestion of simply having a separate method called `writeFile($remotePath, $localPath)`?

spinsch:
Ich empfinde eine zusätzliche Methode für eine Änderung wie diese als unnötig. Aber bin nun dem Wunsch gefolgt. :)

spinsch:
I find an additional method for a change like this unnecessary. But I have now followed the wish. :)

PR #840, opened 2019-10-18
Fix performance issue with bruteforce search in review observer

IvanChepurnyi:
The current observer does a brute-force search to find a current summary related to a product in the collection. It becomes visible when you show 1000 products on a page and increases exponentially with the increased size of product collection. 
Fix replaces brute force search (O(n^m)) with a hash map lookup (O(n))


tmotyl:
@IvanChepurnyi can you share some code snippet you used to test the issue? It would ease up the review. 
Thanks

IvanChepurnyi:
Here is a blackfire profile of a customer where this issue is visible.

PR #842, opened 2019-10-23
Add missing EU country

spinsch:
Croatia joined the EU in 2013.

ref. https://europa.eu/european-union/about-eu/countries_en#tab-0-1

### DIFF
--- a/app/code/core/Mage/Core/etc/config.xml
+++ b/app/code/core/Mage/Core/etc/config.xml
@@ -440,7 +440,7 @@
         </admin>
         <general>
             <country>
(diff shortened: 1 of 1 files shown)


Flyingmana:
not sure about the update script, there may still come an official patch adding an update script for this version.
also, isnt there a chance for complications if we just delete the config entry?

spinsch:
The EU list can be adjusted in the backend. It may happen that someone has complications.
I saw no other way to add a new country to existing installations. Should I skip the upgrade script?

Flyingmana:
For now its probably better to skip the update script.
We can add it at a later point (after the EOL of M1), and then also wait for a bit of feedback from other people.
But in general, as its already configurable, there is no need to update it for existing installations I think

hhirsch:
@spinsch If I understand correctly this can be pulled if you remove the upgrade script from the PR. Can you upgrade your PR?

spinsch:
@hhirsch yes it was understood. I'm a little busy at the moment.

Issue #592, opened 2019-01-23
Discovered another performance issue, this time in Mage_Adminhtml_Block_Sales_Order_Create_Customer_Grid


pepijnblom:
A client was complaining about slow performance when re-ordering a previous order for their customers. I traced it to Mage_Adminhtml_Block_Sales_Order_Create_Customer_Grid which is added to the /admin/sales_order_create page even if you have already chosen a previous order to re-order. Normally you'd go to Sales/Orders and click Create New Order after which you must select a customer from the grid. 

Considering the order & customer have already been chosen when re-ordering, this grid never needs to be added. It's currently running a query which is taking 29 seconds to run (11 or 12 left joins. Their customer_entity table holds 700,000+ customers and the customer_entity_varchar table holds 2.9 million records. 

**EDIT:** I just did a bit more searching and you should see:
`app/design/adminhtml/default/default/template/sales/order/create/form.phtml`
and then
`\Mage_Adminhtml_Block_Sales_Order_Create_Form::getCustomerSelectorDisplay`

Seriously??

I wonder what you thoughts are on fixing this. I'm not sure what the ideal solution is here.
I've managed locally by retrieving the adminhtml/session_quote and then the customer from that in the __construct of Mage_Adminhtml_Block_Sales_Order_Create_Customer_Grid. If a customer Id can be gotten then it just returns. Works just fine but I'm not sure it's the best solution to this problem.

colinmollenhour:
I'd encourage you to go ahead and submit a PR with your working fix so we can see how it works.

pepijnblom:
I decided to implement it as a module in the shop I mentioned.

You can check it out here and implement the change if you see fit.
It shaved off 30s of the load time for creating new orders from an existing order.

https://github.com/graciousagency/magento1-gridoptim

colinmollenhour:
Hmm, what about also using wrapping the rendering of the block in a conditional in the `form.phtml`:

```
<?php if ($this->getCustomerId()): ?>
    <?php echo $this->getChildHtml('customer') ?>
<?php endif ?>
```

pepijnblom:
Hmm that might work actually, but you'd have to flip the logic of the if. I'm going to have to test this some time and I'll make an MR if it works.

```
<?php if (!$this->getCustomerId()): ?>
    <?php echo $this->getChildHtml('customer') ?>
<?php endif ?>
```

colinmollenhour:
Doh, thanks for the correction.. :)

PR #844, opened 2019-10-24
load customers block only if a customer was not set

pepijnblom:
As discussed here:   #592 

Sorry this took so long, I kind of forgot about it since my personal solution worked as well but confirmed it worked on another project now.

### DIFF
--- a/app/design/adminhtml/default/default/template/sales/order/create/form.phtml
+++ b/app/design/adminhtml/default/default/template/sales/order/create/form.phtml
@@ -36,7 +36,9 @@
         <?php echo $this->getChildHtml('message') ?>
     </div>
     <div id="order-customer-selector" style="display:<?php echo $this->getCustomerSelectorDisplay() ?>">
-        <?php echo $this->getChildHtml('customer') ?>
+        <?php if (!$this->getCustomerId()): ?>
+            <?php echo $this->getChildHtml('customer') ?>
+        <?php endif ?>
     </div>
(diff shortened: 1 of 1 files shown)


app/design/adminhtml/default/default/template/sales/order/create/form.phtml
```diff
@@ -36,7 +36,9 @@
         <?php echo $this->getChildHtml('message') ?>
     </div>
     <div id="order-customer-selector" style="display:<?php echo $this->getCustomerSelectorDisplay() ?>">
-        <?php echo $this->getChildHtml('customer') ?>
+        <?php if (!$this->getCustomerId()): ?>
```
tmotyl: This condition is an oposite of the commit title.
"don't load customers block if a customer is already set "
Either the commit should read "don't load customers block if a customer is set" or the condition should be chnaged?

tmotyl:
Hi @pepijnblom you don't have to close PR when you want to change code. Just push to the branch and the PR will be automatically updated.

pepijnblom:
> Hi @pepijnblom you don't have to close PR when you want to change code. Just push to the branch and the PR will be automatically updated.

Yeah I get that but I prefer to have as little commits on my merges as possible /ocd 😅

tmotyl:
then rebase the branch, squashing commits together, and push --force. 
It's making reviewing much harder when you open new PR's and the dicscussion history is lost.

PR #849, opened 2019-11-01
Fix loading of theme based translation

spinsch:
If the frontend is emulated in the backend with the same language as currently in the backend, the emulator does not load the language file from the frontend theme.

I mean this with theme based translation:
` /app/design/frontend/<THEME>/<DESIGN>/locale/<ISO-LOCALE>/translate.csv`


spinsch:
i found a early pull request #813 with the same stuff.

PR #851, opened 2019-11-08
Fix in dataflow profile file upload on windows

pointia:
If you try to upload a csv file from windows in the dataflow profile you get an error.
Since M1.9.3.4 mime type checking of the upload was introduced but windows changes the mime type of the file on upload.

If MS Excel is installed the mime type is application/vnd.ms-excel otherwise it is application/octet-stream

### DIFF
--- a/app/code/core/Mage/Dataflow/Model/Profile.php
+++ b/app/code/core/Mage/Dataflow/Model/Profile.php
@@ -72,6 +72,15 @@ class Mage_Dataflow_Model_Profile extends Mage_Core_Model_Abstract
      * @var array
      */
     protected $_customerTablePermanentAttributes = array('email', 'website');
+    
+    /**
+     * Allowed mime types for csv files
+     * Linux / Mac text/csv
+     * Windows application/octet-stream
+     * Windows with excel installed application/vnd.ms-excel
+     * @var array
+     */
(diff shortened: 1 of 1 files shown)


colinmollenhour:
But what mime type is sent for XML Excel files? I haven't tested but if it was also application/vnd.ms-excel or application/octet-stream then this would break those uploads. Perhaps it would be better to just examine the first few bytes to see if it looks like an xml file?

Flyingmana:
related: https://stackoverflow.com/a/28233618/716029

Iam also for doing an additional check, but only if it is detected as `application/vnd.ms-excel`
Alternatively (or additionally) we use the mime type detection on System side.

pointia:
@Flyingmana I knew that link, this is where i got my information from.
@colinmollenhour You might be right, but I tried to stick to the validation process that Magento provided even though it is OS related. I think Magento has chosen that way of validating the mimtype provided by the OS, to validate the mimetype without opening the file in PHP. 

Maybe the best process would be to check the mimetype provided by the OS and after that check it with mime_content_type to get a clue if we can use it or not.

ADDISON74:
LGTM. Any other opinions?

fballiano (approved):

fballiano:
Almost 3 years with no feedback, I'm closing this.
Feel free to reopen in case you can overtake it.

PR #852, opened 2019-11-12
Stop calling getLastPageNumber() when not necessary

midlan:
Community, what do you think?

When the `$this->_curPage` is `1` the `$this->getLastPageNumber()` is called. In eav collection it is slow operation, because `$this->getSize()` is called inside.

### DIFF
--- a/lib/Varien/Data/Collection.php
+++ b/lib/Varien/Data/Collection.php
@@ -205,7 +205,7 @@ class Varien_Data_Collection implements IteratorAggregate, Countable
      */
     public function getCurPage($displacement = 0)
     {
-        if ($this->_curPage + $displacement < 1) {
+        if ($this->_curPage + $displacement <= 1) {
             return 1;
         }
         elseif ($this->_curPage + $displacement > $this->getLastPageNumber()) {


colinmollenhour:
Very nice optimization!

midlan:
I've got feedback that it may break some category listing. At this moment I don't see the reason, I will try to research.

Flyingmana:
@midlan did you find anything out?

midlan:
@Flyingmana I didn't, my Magento instance is pretty modified and I am not what broken the category listing. Can someone try it on yours Magento?

sreichel:
Just played bit with ... no real usage, but if `$displacement` is something below `-1` (yes, negative ints here) it could break custom code ....

colinmollenhour:
I've been running this since April 23 on a project with tons of custom code and had no issues. 

MySQL doesn't support negative limits: "LIMIT takes one or two numeric arguments, which must both be nonnegative integer constants". So, I don't see any reason anyone would ever use a negative `$displacement` argument.

sreichel:
FYI `$displacement` with `-1` is used in core. @midlan mentioned problems w/ category listings, that i cant reproduce .... maybe related to custom code?

midlan:
@sreichel 
Yes I think it is related to some custom/community code. Did someone tested that in combination with Lesti_Fpc?

tmotyl:
The only place in the core where negative displacement is used is:
https://github.com/OpenMage/magento-lts/blob/performance_tuning/app/code/core/Mage/Page/Block/Html/Pager.php#L284-L284


btw, in Magento2 the `getLastPageNumber` was removed from the function

```
    public function getCurPage($displacement = 0)
    {
        if ($this->_curPage + $displacement < 1) {
            return 1;
        }

        return $this->_curPage + $displacement;
    }
```

sreichel:
> This change could break 3rd party code, if the code depends on the side effect

This it is. I also tested with vanilla install and it did work, but it could break 3rd-party code ...

Issue #826, opened 2019-10-08
Travis not working


midlan:
It seems, that travis stopped working few days ago. New pull requests are not tested with travis:

https://github.com/OpenMage/magento-lts/pull/825

Flyingmana:
they are still running: https://travis-ci.org/OpenMage/magento-lts/pull_requests
But they dont give any feedback to Github.

I also notice, it still has the old default branch(1.9.3.x) configured, not the current one.
And there are entries missing in the Build History.
Also it does not list 1.9.4.x as active branch(or branch at all)
I did change the config on Travis Side to also "Build pushed branches" and not only "Build pushed pull requests", this seemed to trigger the PullRequest runs again.

I cant check the configured Service on the Repository, I think only @LeeSaferite can.

colinmollenhour:
I just manually triggered a build on 1.9.4.x as it said it had never been triggered. I think this will fix it but we'll see.

midlan:
on my PR #852 travis didn't triggered

Flyingmana:
it was at least started https://travis-ci.org/OpenMage/magento-lts/builds/610750039

either travis has problems, or they only send a state after the build ended

midlan:
it finished but the check did not appear at the PR

Flyingmana:
I think everything which we had run via travis, now also runs via github actions

midlan:
seems like everything is now moved to travis. Let's delete the `.travis.yml`?

colinmollenhour:
Yes, I think we can abandon Travis now

midlan:
Travis removed in #1250, closing.

PR #853, opened 2019-11-12
add php syntax check as github action

Flyingmana:
to slowly move away from TravisCI ( partially to solve/workaround #826)

### DIFF
--- a/.github/workflows/static-code-analyses.yml
+++ b/.github/workflows/static-code-analyses.yml
@@ -1,0 +1,61 @@
+name: "Static Code Analyses"
+on:
+  push:
+  pull_request:
+
+
+jobs:
+  php_syntax:
+    name: PHP Syntax ${{ matrix.php-versions }}
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      max-parallel: 5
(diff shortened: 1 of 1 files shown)


Flyingmana:
one example for it can be seen at https://github.com/Flyingmana/magento-lts/runs/299937656

Flyingmana:
with this command
```
      - name: Check .php files
        run: '! find . -not \( -path ./.phpstorm.meta.php -prune \) -not \( -path ./lib/PEAR -prune \) -not \( -path ./lib/phpseclib -prune \) -not \( -path ./lib/Zend -prune \) -type f -name "*.php" -exec php -d error_reporting=32767 -l {} \; 2>&1 1> /dev/null | grep "^"'
      - name: Check .phtml files
        run: '! find app/design -type f -name "*.phtml" -exec php -d error_reporting=32767 -l {} \; 2>&1 1> /dev/null | grep "^"'
```
and a forced error in `/app/design/frontend/base/default/template/bundle/email/order/items/invoice/default.phtml`, we will get

image

Flyingmana:
currently still misses handling of expected failures

PR #856, opened 2019-11-23
remove outdated Undo MagicQuotes function

Flyingmana:
php 7.4 related change, as the used function get_magic_quotes_gpc() is now deprecated

### DIFF
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -25,32 +25,6 @@
  */
 
 /**
- * Disable magic quotes in runtime if needed
- *
- * @link http://us3.php.net/manual/en/security.magicquotes.disabling.php
- */
-if (get_magic_quotes_gpc()) {
-    function mageUndoMagicQuotes($array, $topLevel=true) {
-        $newArray = array();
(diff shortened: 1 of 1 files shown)


kkrieger85:
```
/**
 * Gets the current configuration setting of magic quotes gpc
 * @link https://php.net/manual/en/function.get-magic-quotes-gpc.php
 * @return int 0 if magic quotes gpc are off, 1 otherwise.
 * @since 4.0
 * @since 5.0
 * @deprecated 7.4
 */
function get_magic_quotes_gpc () {}
```

@Flyingmana Do we need to test for backward compatibility?

colinmollenhour:
What about changing the first line to:

```php
if (function_exists('get_magic_quotes_gpc') && get_magic_quotes_gpc()) {
```

Flyingmana:
@colinmollenhour do you have a suggestion, when we can retire this code block completely then?

colinmollenhour:
Ahh, my bad, I just noticed this:

> 5.4.0 | Always returns FALSE because the magic quotes feature was removed from PHP.

So it can be removed now since 5.3 is already EOL and not supported and there is no effect on any supported version.

PR #857, opened 2019-11-23
correct argument order of implode calls

Flyingmana:
reversed order is deprecated from php 7.4 on

https://wiki.php.net/rfc/deprecations_php_7_4
https://www.php.net/manual/en/function.implode.php


kkrieger85:
As I understand rfc, in PHP 7.4 following happens to implode() calls:

implode( $pieces, $glue) => deprecation warning
implode( $glue, $pieces) => OK
implode ( $pieces ) => OK

PR #829, opened 2019-10-11
add php 7.4 to travis.yml

Flyingmana:

PR #859, opened 2019-11-24
fix php7.4 syntax deprecated warnings

Flyingmana:
PHP Deprecated:  Array and string offset access syntax with curly braces
is deprecated

see https://github.com/OpenMage/magento-lts/pull/829


kkrieger85:
https://wiki.php.net/rfc/deprecate_curly_braces_array_access

PR #865, opened 2019-12-02
Grid container without header

Sekiphp:
New little feature. When you have grid only for showing some "static" data, you sometimes no need button and header text. When there are no buttons and header text is empty is hidden header block. Now if there are no buttons and header text is empty, there is gray border.

### DIFF
--- a/app/design/adminhtml/default/default/template/widget/grid/container.phtml
+++ b/app/design/adminhtml/default/default/template/widget/grid/container.phtml
@@ -24,6 +24,7 @@
  * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
+<?php if (!empty($this->getHeaderText()) || !empty($this->getButtonsHtml())): ?>
 <div class="content-header">
     <table cellspacing="0">
         <tr>
@@ -32,6 +33,7 @@
         </tr>
     </table>
 </div>
+<?php endif; ?>
 <div>
     <?php echo $this->getGridHtml() ?>
 </div>


app/design/adminhtml/default/default/template/widget/grid/container.phtml
```diff
@@ -24,6 +24,7 @@
  * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
+<?php if (!empty($this->getHeaderText()) || !empty($this->getButtonsHtml())): ?>
```
midlan: 1. there should be getHeaderHtml instead of getHeaderHtml?
2. save the value of getButtonsHtml into variable, echo it later (instead of calling the function twice, it contains non-trivial code to create the HTML)

Sekiphp:
```    /**
     * Get header HTML
     *
     * @return string
     */
    public function getHeaderHtml()
    {
        return '<h3 class="' . $this->getHeaderCssClass() . '">' . $this->getHeaderText() . '</h3>';
    }```

Sekiphp: Example of alternative solution:

`        $gridContainer = $this->getLayout()
            ->createBlock('namespace_module/adminhtml_parcelHistory')
            ->setBlockId('adminhtml_parcelHistory');

        $grid = $gridContainer->getChild('grid');
        $gridContainer->setData('entity_id', (int) $this->getRequest()->getParam('entity_id'));

        $tabContent .= $gridContainer->toHtml();`

you can replace to:

`        $grid = $this->getLayout()
            ->createBlock('namespace_module/adminhtml_parcelHistory_grid')
            ->setBlockId('adminhtml_parcelHistory');


        $grid->setData('entity_id', (int) $this->getRequest()->getParam('entity_id'));

        $tabContent .= $grid->toHtml();`

and you get grid without header. In other case you can create method `isHeaderHidden` in `Mage_Adminhtml_Block_Widget_Container`

PR #876, opened 2020-01-06
Update .htaccess

miguelbalparda:
Added mod_expires headers for common file types. This shouldn't cause any issues if mod_expires is not present in the server since it's inside a conditional.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -216,9 +216,18 @@
 ############################################
 ## Add default Expires header
 ## http://developer.yahoo.com/performance/rules.html#expires
-
+    ExpiresActive On
+    ExpiresByType image/jpg "access plus 1 year"
+    ExpiresByType image/jpeg "access plus 1 year"
+    ExpiresByType image/gif "access plus 1 year"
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -216,9 +216,18 @@
...
+    ExpiresByType image/jpeg "access plus 1 year"
+    ExpiresByType image/gif "access plus 1 year"
+    ExpiresByType image/png "access plus 1 year"
+    ExpiresByType text/css "access plus 1 month"
+    ExpiresByType application/pdf "access plus 1 month"
+    ExpiresByType text/x-javascript "access plus 1 month"
+    ExpiresByType application/x-shockwave-flash "access plus 1 month"
+    ExpiresByType image/x-icon "access plus 1 year"
+    ExpiresDefault "access plus 2 days"
     ExpiresDefault "access plus 1 year"
```
colinmollenhour: There should not be two "ExpiresDefault" lines, right?

miguelbalparda: Fixed, sorry about that.

PR #877, opened 2020-01-07
Issue with payment method instance on Mage_Payment_Model_Info

Qronicle:
The method `Mage_Payment_Model_Info::getMethodInstance()` does not return the correct instance model after changing the method.

For example, say we are working with a quote, the method instance is "cached" the first time it is gotten, but if you then update the quote's method, and call getMethodInstance again, it will still return the old instance. (Which has caused issues for us were clients payed by invoice and weren't redirected to our payment provider)

I'm not sure the way I solved it is the best, but I don't feel it's safe to check `$instance->getCode() == $this->getMethod()`, so I made a data property on the instance to keep track of the method name that it was made with.


ADDISON74:
Do you get any errors in logs?

fballiano:
So, I was checking this one but
- I don't fully understand how to reproduce the problem (and we've no sample in the PR description)
- I don't 100% like the implementation and, instead of adding a new property, I think it should be proabably solved elsewhere
- there's no feedback since many months and I think it will never come

So I'll take responsability and I'll close it, if we can discuss it further with the author feel free to reopen it.

PR #878, opened 2020-01-08
Create SECURITY.md

Flyingmana:
To make the way reporting more clear

still a draft / the Github default

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
+Use this section to tell people about which versions of your project are
+currently being supported with security updates.
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
+
+Tell them where to go, how often they can expect to get an update on a
+reported vulnerability, what to expect if the vulnerability is accepted or
+declined, etc.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,21 +1,24 @@
 # Security Policy
 
+OpenMage LTS is a fork of Magento CE 1.9.4.4 which provides a place for the Magento community to continue to contribute to the Magento 1 code base. We appreciate you disclosing important security vulnerabilities responsibly and privately by following the easy process defined below.
+
+We will keep the details of your security advisory private and only share it with verified members of our organization or our partner organizations and only on an as-needed basis.
+
 ## Supported Versions
 
-Use this section to tell people about which versions of your project are
-currently being supported with security updates.
-
-| Version | Supported          |
-| ------- | ------------------ |
-| 5.1.x   | :white_check_mark: |
-| 5.0.x   | :x:                |
-| 4.0.x   | :white_check_mark: |
-| < 4.0   | :x:                |
+| Version   | Branch           | Supported          |
+| --------- | ---------------- | ------------------ |
+| 1.9.4.4   | 1.9.4.x          | :white_check_mark: |
+| < 1.9.4.4 | 1.9.3.x, etc.    | :x:                |
 
 ## Reporting a Vulnerability
 
-Use this section to tell people how to report a vulnerability.
+To report a vulnerability, please *DO NOT* open a public Issue or Pull Request.
 
-Tell them where to go, how often they can expect to get an update on a
-reported vulnerability, what to expect if the vulnerability is accepted or
-declined, etc.
+Please log in to Github (sign up is free) and report the vulnerability by opening a draft security advisory [here](https://github.com/OpenMage/magento-lts/security/advisories/new).
+
+https://github.com/OpenMage/magento-lts/security/advisories/new
+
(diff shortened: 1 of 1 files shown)


kkrieger85:
@Flyingmana Maybe we could recommend https://github.com/kkrieger85/magento-module-security-txt and add additional information about how to report issues to OpenMage

colinmollenhour:
Is there a way you can see to include a default static security.txt (referring to OpenMage) that the user can still override easily (referring to themselves)? I suppose they could always use Apache's `Alias` or nginx's `location {}` to override it if they wanted to. So I'd probably lean towards just adding a static `.well-known/security.txt` file to the repo since probably less than 0.1% of Magento sites have one of these files already.

kkrieger85:
If we would extend getVersionInfo() https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/Mage.php#L169 by adding "security"-hint, it would be possible to automatically add OpenMage reference to security.txt

Proposal:

```
        return array(
            'major'       => '1',
            'minor'       => '9',
            'revision'    => '4',
            'patch'       => '3',
            'stability'   => '',
            'number'      => '',
            'name'        => 'OpenMage',
            'source'      => 'https://github.com/OpenMage/magento-lts',
            'security'    => 'https://github.com/OpenMage/magento-lts/issues',
        );
```

Name should be Unique - but that would be quite hard to convince other Magento1 Forks to adopt this.
Source should be a link to original source code
Security should be a link to page where someone could report security issues

colinmollenhour:
What are the other active M1 forks besides OpenMage and mage-one.com?

Shouldn't the security url be https://github.com/OpenMage/magento-lts/security/advisories/new ?

Flyingmana:
So the primary usecase of this file is, so people wanting to report a security issue in OpenMage know where and how.
The module of @kkrieger85 is for offering websites a way to contact them.

For OpenMage we do want to explain an alternative way to report security issues in a more private way instead of creating a public github issue.
The Security Advisory may be the best for this. Will test if this can be used by external people to report one. (Did let it test, outsiders just see a 404 page)
So I would fall back to my initial suggestion of having voluntairs provide their email for contact, and they then create internally the advisory

Regarding Other Forks, I started a list recently https://openmage.github.io/magento-lts/alternatives.html

colinmollenhour:
Bummer on the 404.. Is there a way to make the submission available to anyone with a github account? Looks like a nice feature but not as useful if only maintainers can use it...

Hey, @mark-netalico, are you interested in joining OpenMage to help maintain M1? If not, what are your reservations?

mark-netalico:
@colinmollenhour 100%. I started Mage1SE because I think there needs to be some sort of funding/sponsorship of bug bounties and I'm working to set that up. But I already talked to @Flyingmana about having some official collaboration (especially in the case of something like a zero-day vulnerability).

kkrieger85:
@mark-netalico AFAIK: Bug bounties and zero day is the idea behind https://mage-one.com/ 
I like the idea of an open source bug bounty program :D

kkrieger85:
Alternatives: https://github.com/bragento/magento-core
and there is a fork by some french, but I can't remember the name

mark-netalico:
@kkrieger85 Indeed. I have a lot of concerns with the viability/legality/ethic of Mage One... Unfortunately, I don't think it's a viable long term solution as I don't they they'll be able to truly keep the security updates closed source because doing so would be an OSL 3.0 license violation as it requires Derivative Work disclosure. They'll try to keep them private, but their customers will eventually release them publicly. And I think that the project will shut down fairly quickly after the EOL due to lack of funding.

colinmollenhour:
I updated the default template, any objection to merging it so I can test the security advisories page after it is published? I'm wondering if the link only works for non-project members after the SECURITY.md is published.

Flyingmana:
see https://help.github.com/en/github/managing-security-vulnerabilities/creating-a-security-advisory

> Anyone with admin permissions to a repository can create a security advisory.

colinmollenhour:
I don't know what "anyone" means in this context? Any github user?

colinmollenhour:
Ahh, thanks. It is odd to me that they don't make this feature public... I updated the language to just start by email for now. We can update as things change. Does it look ok for merge now?

SECURITY.md
```diff
@@ -0,0 +1,20 @@
+# Security Policy
+
+OpenMage LTS is a fork of Magento CE 1.9.4.4 which provides a place for the Magento community to continue to contribute to the Magento 1 code base. We appreciate you disclosing important security vulnerabilities responsibly and privately by following the easy process defined below.
```
tmotyl, line 3:
```suggestion
OpenMage LTS is a fork of Magento CE 1.9 which provides a place for the Magento community to continue to contribute to the Magento 1 code base. We appreciate you disclosing important security vulnerabilities responsibly and privately by following the easy process defined below.
```

colinmollenhour:
Confirmed that this page is now accessible without Github authentication:
https://github.com/OpenMage/magento-lts/security/policy

colinmollenhour:
Note, the Security policy references the README but the list of maintainers in the README doesn't actually have email addresses, just links to Github profiles, but Github profile don't expose email addresses either.. I think perhaps we should instead link to the HackerOne page. Thoughts?

tmotyl:
I definitelly prefer one contact poinf for security, otherwise it doesnt look professional, so either openmage emaik or hacker one pge

Issue #882, opened 2020-01-16
Upgrade to 1.9.4.3 breaks grids using `filter_condition_callback` SUPEE-11219 [confirmed+fix]


seansan:
Upgrade to 1.9.4.3 breaks grids using `filter_condition_callback` SUPEE-11219 [confirmed]

Due to code changes here https://github.com/OpenMage/magento-lts/commit/cce46fced84e2e61d20fa7227b80f1f652df538a#r36881190 belonging to SUPEE-11219

New problems occur in various extensions a code utilizing `filter_condition_callback` as described here and as ... well we have always done

To me this seems as a breaking change ... and as such breaking more than necessary .. although I must say it is not clear why the extra if check was added in the first place

The problem is described here: https://magento.stackexchange.com/a/292307/78 and here in more (user) detail https://magento.stackexchange.com/questions/294824/magento-1-9-4-3-custom-column-filter-not-working

Expected
Custom filters to work in grids

Actual
Custom filters do no longer work

Solution
Remove the second if statement again

![image](https://user-images.githubusercontent.com/652395/72530712-3bdc3000-3870-11ea-9b49-1fbebef52673.png)

The error is 

```
invalid attribute name: warehouse_ids

#0 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(1295): Mage::exception('Mage_Eav', 'Invalid attribu...')
#1 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(1388): Mage_Eav_Model_Entity_Collection_Abstract->_addAttributeJoin('warehouse_ids', 'inner')
#2 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(322): Mage_Eav_Model_Entity_Collection_Abstract->_getAttributeConditionSql('warehouse_ids', Array, 'inner')
#3 app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php(1446): Mage_Eav_Model_Entity_Collection_Abstract->addAttributeToFilter('warehouse_ids', Array, 'inner')
#4 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(342): Mage_Catalog_Model_Resource_Product_Collection->addAttributeToFilter('warehouse_ids', Array)
#5 app/code/core/Mage/Adminhtml/Block/Widget/Grid.php(472): Mage_Eav_Model_Entity_Collection_Abstract->addFieldToFilter('warehouse_ids', Array)
[...]

PR #887, opened 2020-01-20
Fixing Magento Invalid attribute name issue after SUPEE-11219

seansan:
Removed 2nd if check after M version upgrade broke this for many grids
Described here 
https://www.garybell.co.uk/fix-magento-invalid-attribute-error-supee-11219/
and here https://magento.stackexchange.com/a/292307/78

Fixes https://github.com/OpenMage/magento-lts/issues/882


sreichel:
I'd like to close ....

https://magento.stackexchange.com/questions/292092/supee-11219-potential-issues/292307#292307

> Edit: Fixed the issue by moving callbacks to another class which extends proper parent class in this case Mage_Adminhtml_Block_Widget_Grid. In the past custom callback functions could basically be in any class and we had these in plain Helper\Data.php which just extended Mage_Core_Helper_Abstract and that didn't work anymore.

Just put callback into right class.

PR #890, opened 2020-01-23
Remove this->_debug Paypal

seansan:
Not supposed to be there?

Please check if any functionality depends on it


seansan:
Feel free to make the change. If a new product: then I would remove, seeing M1 is somehwhat legacy I tend to keep the history. You can decide and merge

PR #892, opened 2020-01-31
Merge pull request #1 from OpenMage/1.9.4.x

roman204:
fix Nesting level too deep - recursive dependency?


Flyingmana:
I think there did something go wrong? I dont see any changes listed

roman204:
Nahh did it from the phone which was horrible

PR #895, opened 2020-02-04
Removed ES6 JS introduced in 1.9.4.4 for IE compability

samsteele:


samsteele:
Added missing bracket

PR #896, opened 2020-02-06
Fix for critical checkout breaking bug for IE11 users

fitzjohnit:
Magento 1.9.4.4 added a new function to js/varien/js.js called buttonDisable (only used by the Compilation page in admin). Unfortunately it uses arrow functions which IE11 doesn't support and so breaks the ability of IE11 users to purchase anything.

I've rewritten and tested the function to make it compatible with IE11.


fitzjohnit:
Apologies missed that!

PR #897, opened 2020-02-06
When category default_sort_by not available, use the system config

midlan:
When you remove sort method from Magento and some categories is sorted by it, theese categories are sorted by the first availible method. I fixed the method, to first try use the sort from system config.


app/code/core/Mage/Catalog/Model/Category.php
```diff
@@ -918,10 +918,16 @@ public function getAvailableSortByOptions() {
      * @return string
      */
     public function getDefaultSortBy() {
-        if (!$sortBy = $this->getData('default_sort_by')) {
+
+        $sortBy = $this->getData('default_sort_by');
+
+        //when not set or not available use default from system config
+        if (!$sortBy || !isset($available[$sortBy])) {
```
Sekiphp: Variable `$available` is not defined yet

seansan: agree code needs logic update by OP

colinmollenhour: Dunno if this violates PSR or not but I often use the assignment in the if statement but it has to be wrapped in parentheses:

```
if ( ! ($sortBy = $this->getData('default_sort_by'))) {
```

midlan: fixed now, the $available variable is initialized on the right place

app/code/core/Mage/Catalog/Model/Category.php
```diff
@@ -918,11 +918,17 @@ public function getAvailableSortByOptions() {
...
             $sortBy = Mage::getSingleton('catalog/config')
                 ->getProductListDefaultSortBy($this->getStoreId());
         }
-        $available = $this->getAvailableSortByOptions();
+
+        //even the sort from system config not available, use the first of available
         if (!isset($available[$sortBy])) {
```
colinmollenhour: Doesn't this `if` block basically clobber the effects of the first `if` block when `!isset($available[$sortBy])` is true?

midlan: no, the current logic is:

1. try sort from category object; cames from: `$this->getData('default_sort_by')`
2. sort from category object is empty or not available, try use sort from system config; cames from: `Mage::getSingleton('catalog/config')->getProductListDefaultSortBy($this->getStoreId());`
3. sort from system config is also not available, use the first available; cames from `$sortBy[0]`


In the original logic was the sort from system config used only when category sort was empty.

colinmollenhour: The current logic can be simplified like so:

```php
if ($a || $b) {
   $c = 1;
}
if ($b) {
  $c = 2;
}
```

This is equivalent to:

```php
if ($a) {
  $c = 1;
}
if ($b) {
  $c = 2;
}
```

midlan: I didn't catch your hint. Show me example on actually the method without changing the logic:

```php
    public function getDefaultSortBy() {

        $sortBy = $this->getData('default_sort_by');
        $available = $this->getAvailableSortByOptions();

        //when not set or not available use default from system config
        if (!$sortBy || !isset($available[$sortBy])) {
            $sortBy = Mage::getSingleton('catalog/config')
                ->getProductListDefaultSortBy($this->getStoreId());
        }

        //even the sort from system config not available, use the first of available
        if (!isset($available[$sortBy])) {
            $sortBy = array_keys($available);
            $sortBy = $sortBy[0];
        }

        return $sortBy;
    }
```

colinmollenhour: My bad, I misunderstood the code.. So hard to read with comments inline.. :)

I think all I'd change is to add `! isset($available[$sortBy])` to the last `if` in case the system config value is not an available option.

```php
    public function getDefaultSortBy() {
        $sortBy = $this->getData('default_sort_by');
        $available = $this->getAvailableSortByOptions();
        if ( ! $sortBy || ! isset($available[$sortBy])) {
            $sortBy = Mage::getSingleton('catalog/config')
                ->getProductListDefaultSortBy($this->getStoreId());
        }
        if ( ! $sortBy || ! isset($available[$sortBy])) {
            $sortBy = array_shift($available);
        }
        return $sortBy;
    }
```

midlan: fixed now with your hint

PR #899, opened 2020-02-15
Set default sort order for some grids

luigifab:
Set default sort order for customer group and promo catalog and promo quote grids.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Promo/Quote/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Promo/Quote/Grid.php
@@ -41,8 +41,8 @@ class Mage_Adminhtml_Block_Promo_Quote_Grid extends Mage_Adminhtml_Block_Widget_
     {
         parent::__construct();
         $this->setId('promo_quote_grid');
-        $this->setDefaultSort('sort_order');
-        $this->setDefaultDir('ASC');
+        $this->setDefaultSort('rule_id');
+        $this->setDefaultDir('desc');
         $this->setSaveParametersInSession(true);
     }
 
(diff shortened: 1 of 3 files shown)


app/code/core/Mage/Adminhtml/Block/Promo/Quote/Grid.php
```diff
@@ -41,8 +41,8 @@ public function __construct()
     {
         parent::__construct();
         $this->setId('promo_quote_grid');
-        $this->setDefaultSort('sort_order');
-        $this->setDefaultDir('ASC');
+        $this->setDefaultSort('rule_id');
+        $this->setDefaultDir('desc');
```
sreichel, line 45: I disagree with this one.

luigifab: I prefer to see the latest created rule in first :)

colinmollenhour: If we were to apply a convention to all grids it seems like most recently created (or primary key desc) would generally be the best rule. While you could make a case that in some cases it makes sense to deviate from that, you could also make a case that you should still not deviate just so the user always knows what to expect. I don't particularly care in this case but would probably lean towards establishing a generalized rule and sticking with that.

tmotyl: Please also check if there is an index on the column you are going to sort with. If not this change would be a performance issue if there are many rules

sreichel: Its not about performance .... espacially for promo rules i'd leave "sort order" to show priority.

luigifab: Do I need to change something?

colinmollenhour: Note, you can set the grid column to use as the sort order (e.g. created_at) but still use a different column for the actual sorting (e.g. entity_id) so that the proper indexes are used:

```php
$this->setDefaultSort('created_at');

...

$this->addColumn('created_at',[
  'index' => 'created_at',
  ...
  'sort_index' => 'entity_id',
]);
```

luigifab:
I can not push again, so I close it.

PR #901, opened 2020-02-15
fix PHP Warning: "continue" targeting switch is equivalent to "break"…

luigifab:
A fix for PHP 7.3/7.4 Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"?


colinmollenhour:
Replaced by #1013

PR #902, opened 2020-02-15
Fix some PHP mistakes

luigifab:
Initialize $newValue for 2 sql upgrade.
Add default parameter to idn_to_ascii().
Fix some PHP mistakes (undefined index, @new, sizeof on null).

### DIFF
--- a/app/code/core/Mage/AdminNotification/Model/Feed.php
+++ b/app/code/core/Mage/AdminNotification/Model/Feed.php
@@ -169,7 +169,7 @@ class Mage_AdminNotification_Model_Feed extends Mage_Core_Model_Abstract
         $curl->close();
 
         try {
-            $xml  = new SimpleXMLElement($data);
+            $xml = @new SimpleXMLElement($data);
         }
         catch (Exception $e) {
             return false;
--- a/app/code/core/Mage/Core/Helper/Url.php
+++ b/app/code/core/Mage/Core/Helper/Url.php
@@ -180,7 +180,7 @@ class Mage_Core_Helper_Url extends Mage_Core_Helper_Abstract
         $parsedUrl = parse_url($url);
         if (!$this->_isPunycode($parsedUrl['host'])) {
             if (function_exists('idn_to_ascii')) {
-                $host = idn_to_ascii($parsedUrl['host']);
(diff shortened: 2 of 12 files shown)


app/code/core/Mage/Core/Helper/Url.php
```diff
@@ -180,7 +180,7 @@ public function encodePunycode($url)
         $parsedUrl = parse_url($url);
         if (!$this->_isPunycode($parsedUrl['host'])) {
             if (function_exists('idn_to_ascii')) {
-                $host = idn_to_ascii($parsedUrl['host']);
+                $host = idn_to_ascii($parsedUrl['host'], IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46);
```
spinsch: the parameters are already set as default. why must this be specified explicitly now?

app/code/core/Mage/AdminNotification/Model/Feed.php
```diff
@@ -169,7 +169,7 @@ public function getFeedData()
         $curl->close();
 
         try {
-            $xml  = new SimpleXMLElement($data);
+            $xml = @new SimpleXMLElement($data);
```
spinsch: what happens now if an exception is thrown?

app/code/core/Mage/Core/Helper/Url.php
luigifab: I have added these default parameters some months ago, at this time, there was a notice or a warning with php 7.2? (I don't remember exactly). I tested again with php 7.2 & 7.4, and I'm agree these default parameters are useless. Sorry.

app/code/core/Mage/AdminNotification/Model/Feed.php
luigifab: Wow! I'm stupid, ignore that.

lib/Zend/Service/WindowsAzure/Storage/Batch.php
```diff
@@ -105,7 +105,7 @@ protected function _clean()
         unset($this->_operations);
         $this->_storageClient->setCurrentBatch(null);
         $this->_storageClient = null;
-        unset($this);
+        //unset($this);
```
colinmollenhour, line 108: Code should just be deleted rather than commented out.

luigifab:
I will resend a pull request, I don't understand how to edit.

PR #907, opened 2020-02-24
Warning: count(): Parameter must be an array.

michalszdeligo:
I added fix for sometimes it appears "Warning: count(): Parameter must be an array.".

### DIFF
--- a/lib/Varien/File/Uploader.php
+++ b/lib/Varien/File/Uploader.php
@@ -504,9 +504,9 @@ class Varien_File_Uploader
             $this->_uploadType = self::MULTIPLE_STYLE;
             $this->_file = $fileId;
         } else {
-            preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
+            $result = preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
 
-            if (count($file) > 0 && (count($file[0]) > 0) && (count($file[1]) > 0)) {
+            if ($result && count($file) === 3) {
                 array_shift($file);
                 $this->_uploadType = self::MULTIPLE_STYLE;
 


lib/Varien/File/Uploader.php
```diff
@@ -504,9 +504,9 @@ private function _setUploadFileId($fileId)
             $this->_uploadType = self::MULTIPLE_STYLE;
             $this->_file = $fileId;
         } else {
-            preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
+            $result = preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
 
-            if (count($file) > 0 && (count($file[0]) > 0) && (count($file[1]) > 0)) {
+            if ($result && count($file) === 3) {
```
tmotyl: Why ==3 ? Previous code checked only whether first two items are

colinmollenhour: index 0 is the whole string matched, 1 is the first capture and 2 is the second capture, so three elements. The full matched string is dropped on the next line by array_shift. I kinda think the regex should be changed to be `.+?` rather than `.*?` so that all captures must be non-empty.

colinmollenhour:
```suggestion
            if (preg_match("/^(.+?)\[(.+?)\]$/", $fileId, $file)) {
```

lib/Varien/File/Uploader.php
```diff
@@ -504,9 +504,9 @@ private function _setUploadFileId($fileId)
             $this->_uploadType = self::MULTIPLE_STYLE;
             $this->_file = $fileId;
         } else {
-            preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
+            $result = preg_match("/^(.*?)\[(.*?)\]$/", $fileId, $file);
```
colinmollenhour, line 507:
```suggestion
```

lib/Varien/File/Uploader.php
```diff
@@ -504,9 +504,9 @@ private function _setUploadFileId($fileId)
...
```
colinmollenhour, line 508:
```suggestion
```

ADDISON74:
... and what is the final version of this PR after suggestions?

fballiano:
I'm closing this, the description is not really easy to understand, we lack feedback from @michalszdeligo since forever

PR #911, opened 2020-03-16
Optimize EAV collections

Sekiphp:
### Some DB background

Screenshot_7

### Overview
In cases when Magento collections are extended from `Mage_Eav_Model_Entity_Collection_Abstract` there are filtering by `entity_type_id` but for example for customer entity_type_id is always 1 (viz. screenshot)
Screenshot_8

But this conditions slow down multiple queries and I think, that at this time is abandoned using entity_type_id for most of queries.


app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -167,9 +167,7 @@ protected function _prepareStaticFields()
     protected function _initSelect()
     {
         $this->getSelect()->from(array('e' => $this->getEntity()->getEntityTable()));
-        if ($this->getEntity()->getTypeId()) {
-            $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
-        }
+
```
colinmollenhour, line 170: I believe this will accomplish the same but still keep compatibility with the default entity table (which is not used in core but is still present).

```suggestion
        if ($this->getEntity()->getEntityTable() != Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
            $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
        }
```

colinmollenhour:
@Sekiphp Do you agree with the changes? Please update PR if so. Thanks!

joshua-bn:
Bump @Sekiphp. Looks like a good suggestion from @colinmollenhour and pretty simple. Looks like a good change to implement.

Sekiphp:
Your suggestion implemented @colinmollenhour ;-)

colinmollenhour (approved):
Looks good, thanks!

app/code/core/Mage/Directory/Model/Resource/Currency.php
```diff
@@ -183,16 +183,22 @@ public function saveRates($rates)
...
-        $bind    = array(':config_path' => $path);
-        $select  = $adapter->select()
-                ->from($this->getTable('core/config_data'))
-                ->where('path = :config_path');
-        $result  = array();
-        $rowSet  = $adapter->fetchAll($select, $bind);
-        foreach ($rowSet as $row) {
-            $result = array_merge($result, explode(',', $row['value']));
+        $result  = [];
+        $config = Mage::app()->getConfig();
```
tmotyl, line 187: this looks like a change from https://github.com/OpenMage/magento-lts/pull/1252/files, it shouldn't be here, right?

Sekiphp: Thank you! I mischeckouted to another branch :-)

Sekiphp:
Finally fixed @colinmollenhour @joshua-bn @tmotyl

joshua-bn:
It's been a while since I looked at this so I might've lost some knowledge here.

For the entity tables like customer, catalog, etc, the entity_type_id will always be the same. For the eav_entity table (I don't think anyone ever uses that table) this does matter. The condition seems backwards then. If it _is_ the default table, we should add the entity_type_id filter. If it _is not_ then there's no need, right?

colinmollenhour (approved):

colinmollenhour (dismissed):
@joshua-bn is right, I had the logic backwards! Doh!

Sekiphp:
> It's been a while since I looked at this so I might've lost some knowledge here.
> 
> For the entity tables like customer, catalog, etc, the entity_type_id will always be the same. For the eav_entity table (I don't think anyone ever uses that table) this does matter. The condition seems backwards then. If it _is_ the default table, we should add the entity_type_id filter. If it _is not_ then there's no need, right?

So the right solution is?: 
```
if ($this->getEntity()->getEntityType()->getEntityModel() == Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) { 
    $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
}
```

@colinmollenhour I did some changes, because: `$this->getEntity()->getEntityTable()` returns table name and `Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE` contains model path.

joshua-bn:
@Sekiphp it should be the table, not the model.

I think this is what we're looking for:
```
    protected function _initSelect()
    {
        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);

        if ($this->getEntity()->getEntityTable() === Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
            $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
        }

        return $this;
    }
```

joshua-bn:
btw, here's what M2 has for this https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Eav/Model/Entity/Collection/AbstractCollection.php#L217

luigifab:
I'm not sure, what is the status of this PR?

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -167,9 +167,11 @@ protected function _prepareStaticFields()
     protected function _initSelect()
     {
         $this->getSelect()->from(array('e' => $this->getEntity()->getEntityTable()));
-        if ($this->getEntity()->getTypeId()) {
+
+        if ($this->getEntity()->getEntityTable() != Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
             $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
         }
```
kiatng, line 173: I tested both the PR and @joshua-bn's suggestion, and found that @joshua-bn produces the correct SQL statement.

```suggestion
        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);

        if ($this->getEntity()->getEntityTable() === Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
            $this->addAttributeToFilter('entity_type_id', $this->getEntity()->getTypeId());
        }
```

Test code:
```php
$collection = Mage::getResourceModel('customer/customer_collection');
$collection->getSelect()->__toString();
// @joshua-bn Correct Result: SELECT `e`.* FROM `customer_entity` AS `e`
// PR Incorrect Result: SELECT `e`.* FROM `customer_entity` AS `e` WHERE (`e`.`entity_type_id` = '1')
```

colinmollenhour: Oops, had my boolean logic backwards with the original suggestion.. Agreed that your suggestion is correct.

ADDISON74:
With a little changes in code as requested we can merge this PR .

luigifab:
The SQL queries with our MariaDB 10.7:
```
MariaDB []> SELECT SQL_NO_CACHE COUNT(*) FROM customer_entity AS `e` WHERE (`e`.entity_type_id = '1');
+----------+
| COUNT(*) |
+----------+
|  1214517 |
+----------+
1 row in set (0.283 sec)

MariaDB []> SELECT SQL_NO_CACHE COUNT(*) FROM customer_entity AS `e` WHERE 1;
+----------+
| COUNT(*) |
+----------+
|  1214517 |
+----------+
1 row in set (0.245 sec)
```

I don't see problems at localhost with this change.

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -167,9 +167,11 @@ protected function _prepareStaticFields()
     protected function _initSelect()
     {
         $this->getSelect()->from(array('e' => $this->getEntity()->getEntityTable()));
-        if ($this->getEntity()->getTypeId()) {
+
+        if ($this->getEntity()->getEntityTable() != Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
```
colinmollenhour, line 171: I believe the logic was supposed to be flipped so that the entity_type_id filter is only applied for the default `eav_entity` table (which nobody uses)?

```suggestion

        if ($this->getEntity()->getEntityTable() == Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
```

colinmollenhour (changes requested):

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -159,10 +159,12 @@ protected function _prepareStaticFields()
      */
     protected function _initSelect()
     {
-        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);
-        if ($this->getEntity()->getTypeId()) {
+        $this->getSelect()->from(array('e' => $this->getEntity()->getEntityTable()));
+
+        if ($this->getEntity()->getEntityTable() === Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
```
kiatng, line 164: Resolved conflict with 1.9.4.x, short array:

```suggestion
        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);
 
        if ($this->getEntity()->getEntityTable() === Mage_Eav_Model_Entity::DEFAULT_ENTITY_TABLE && $this->getEntity()->getTypeId()) {
```

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -159,10 +159,12 @@ protected function _prepareStaticFields()
      */
     protected function _initSelect()
     {
-        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);
-        if ($this->getEntity()->getTypeId()) {
+        $this->getSelect()->from(array('e' => $this->getEntity()->getEntityTable()));
```
kiatng, line 162: Resolved conflict with 1.9.4.x, short array:

```suggestion
        $this->getSelect()->from(['e' => $this->getEntity()->getEntityTable()]);
```

kiatng (approved):
Tested.

luigifab:
So I tested the wrong version at localhost, but I didn't see any bad changes...

colinmollenhour (approved):

fballiano:
this one is approved for merging, is everybody comfortable with this?

kiatng:
I have this in production a month now. No issue so far.

PR #912, opened 2020-03-21
Do not emit warning on null byte in $src in io_file

tmotyl:


Flyingmana:
is there not a way to check for a null byte, without suppressing errors?
I would like to have the code in a way, that does not cause issues if someone activates the ini setting `scream.enabled`

kkrieger85:
Proposal: $src = str_replace(chr(0), '', $src);

Issue #915, opened 2020-04-04
Warning in product sold report - magento tries to sum product names (order_items_name)


tmotyl:
Warning: A non-numeric value encountered  in app/code/core/Mage/Reports/Model/Totals.php on line 61
The line is 
```
    $columns[$field]['value'] = $columns[$field]['value'] + (isset($data[$field]) ? $data[$field] : 0);
```
$field is 'order_items_name' in this case 
$data[$field] - is a product name

Grid configuration (Block/Report/Product/Sold/Grid.php)
```
        $this->addColumn('name', array(
            'header'    =>Mage::helper('reports')->__('Product Name'),
            'index'     =>'order_items_name'
        ));
```

PR #916, opened 2020-04-04
Do not sum columns with undefined total function

tmotyl:
This function only supports non-empty total functions (avg, sum, or total with slash)
so we can safely ignore fields which have 'total' undefined (null).
Fixes: https://github.com/OpenMage/magento-lts/issues/915


tmotyl:
here you go

tmotyl:
Please review again, i had to fix whitespace issue

Issue #917, opened 2020-04-04
Add runtime cache to zend locale data


tmotyl:
This modified version of Zend_Locale_Data (based on version from Magento CE 1.6.2.0) uses a local cache in addition to the shared cache to prevent repeated cache requests for the same data. This can save literally hundreds of cache requests per page depending on how heavily the page uses Zend_Locale (dates and currency).

https://gist.github.com/colinmollenhour/2715828

PR #918, opened 2020-04-04
Add runtime cache to Zend_Locale_Data

tmotyl:
Fixes: https://github.com/OpenMage/magento-lts/issues/917
This imporves performance,as Zend is not calling cache backend multiple times for locale information.

### DIFF
--- a/lib/Zend/Locale/Data.php
+++ b/lib/Zend/Locale/Data.php
@@ -75,6 +75,13 @@ class Zend_Locale_Data
     private static $_cacheDisabled = false;
 
     /**
+     * Internal cache, prevent repeated cache requests
+     *
+     * @var array
+     */
+    private static $_localCache = array();
+
+    /**
      * Read the content from locale
      *
      * Can be called like:
@@ -335,8 +342,15 @@ class Zend_Locale_Data
 
         $val = urlencode($val);
         $id  = self::_filterCacheId('Zend_LocaleL_' . $locale . '_' . $path . '_' . $val);
+
+        // add runtime cache to avoid callng cache backend multiple times during one request
+        if ( isset(self::$_localCache[$id])) {
+            return self::$_localCache[$id];
+        }
         if (!self::$_cacheDisabled && ($result = self::$_cache->load($id))) {
-            return unserialize($result);
+            $result = unserialize($result);
+            self::$_localCache[$id] = $result;
+            return $result;
         }
 
         $temp = array();
@@ -946,11 +960,13 @@ class Zend_Locale_Data
         }
 
         if (isset(self::$_cache)) {
+            $data = serialize($temp);
             if (self::$_cacheTags) {
-                self::$_cache->save( serialize($temp), $id, array('Zend_Locale'));
+                self::$_cache->save( $data, $id, array('Zend_Locale'));
             } else {
-                self::$_cache->save( serialize($temp), $id);
+                self::$_cache->save( $data, $id);
             }
+            static::$_localCache['id'] = $data;
         }
 
         return $temp;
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Minor improvement: might as well save the result (`$temp`) in the local cache as well at the end of the `getList` method to avoid reading from cache or when a cache backend is disabled.

colinmollenhour:
Haha, I just realized that this PR is based on my patch from almost 8 years ago! Thanks for digging it up!

tmotyl:
yep, oldi but goldi ;) 
I've stu,mbled upon similar issues when working with Magento 2. https://github.com/magento/magento2/issues/20973

lib/Zend/Locale/Data.php
```diff
@@ -946,11 +960,13 @@ public static function getList($locale, $path, $value = false)
...
         if (isset(self::$_cache)) {
+            $data = serialize($temp);
             if (self::$_cacheTags) {
-                self::$_cache->save( serialize($temp), $id, array('Zend_Locale'));
+                self::$_cache->save( $data, $id, array('Zend_Locale'));
             } else {
-                self::$_cache->save( serialize($temp), $id);
+                self::$_cache->save( $data, $id);
             }
+            static::$_localCache['id'] = $data;
```
ilnytskyi, line 969: @tmotyl 
why this key used `'id'` ?
should it be the `$id` variable ?

tmotyl: damn, it should. Can you make a PR?

tmotyl: and in line 1526 too

colinmollenhour:
Fixed in 8b89754e884d5d9b8907311ac121f26bc7809ef0. Thanks @ilnytskyi!

ilnytskyi:
We tried this patch with M2 and we got an error :)

```
{"message":"Warning: Illegal string offset 'decimal' in /var/www/html/vendor/magento/zendframework1/library/Zend/Locale/Format.php on line 427","context":[],"level":500,"level_name":"CRITICAL","channel":"main"}
```

lib/Zend/Locale/Data.php
Flyingmana: one reason, why its always good to wait for a second Review 👀

lib/Zend/Locale/Data.php
```diff
@@ -985,7 +1001,9 @@ public static function getContent($locale, $path, $value = false)
         $val = urlencode($val);
         $id  = self::_filterCacheId('Zend_LocaleC_' . $locale . '_' . $path . '_' . $val);
         if (!self::$_cacheDisabled && ($result = self::$_cache->load($id))) {
```
ilnytskyi, line 1003: btw if I understand correctly lack of 
```
        // add runtime cache to avoid callng cache backend multiple times during one request
        if (isset(self::$_localCache[$id])) {
            return self::$_localCache[$id];
        }
``` 
before the `if` will trigger `$_cache->load($id)` and data is not taken from runtime

Noticed that after the patch we anyway have many `Zend_LocaleC_` coming from cache

tmotyl: can you propose a patch/pull request? thanks!

tmotyl: https://github.com/OpenMage/magento-lts/pull/1018
I've pushed the change, and tested it.

kiatng:
We ran into seemingly random warnings in both frontend and backend:

> Warning: Invalid argument supplied for foreach() in lib/Zend/Locale/Format.php on line 853

Code adjacent to line 853:

```php
// erase day string
    $daylist = Zend_Locale_Data::getList($options['locale'], 'day');
foreach($daylist as $key => $name) {
    if (iconv_strpos($number, $name) !== false) {
        $number = str_replace($name, "EEEE", $number);
        break;
    }
```

`$daylist` is a serialized string:

```
a:7:{s:3:"sun";s:6:"Sunday";s:3:"mon";s:6:"Monday";s:3:"tue";s:7:"Tuesday";s:3:"wed";s:9:"Wednesday";s:3:"thu";s:8:"Thursday";s:3:"fri";s:6:"Friday";s:3:"sat";s:8:"Saturday";}
```

Not sure if #1018 would fix it.

PR #919, opened 2020-04-04
Improve performance of the products sold report

tmotyl:
Disable inventory status loading event on after_load
And prevent collection from fetching all attributes (at later point name and entity_id is added).
I did a blackfire profile with monthly report 
And had 30% speed improvement on both CPU, Memory and DB queries.
image

Alternative implementation woud be to add new flag to product collection and read it in inventory observer similar to the "require_stock_items" flag which is already in place there see:

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/CatalogInventory/Model/Observer.php#L112


colinmollenhour:
I think I'd prefer using the action flags vs modifying the config in memory. Not a huge risk but if somehow the config was saved later in the request it would potentially cause this value to be saved. Also there could be future developments like storing config in shared memory which would not play well with this kind of practice, although that too seems highly unlikely. :)

Issue #920, opened 2020-04-06
Performance issued due too old_fields_map functionality


tmotyl:
Magento 1.6 has introduced a `_initOldFieldsMap` functionality in VarienObject.
It was used to keep backward compatibility with code written pre 1.6.
As in 1.6 some db fields names has been changed.
In some models the _initOldFieldsMap returns static array (like in stock), in others like product it gets configuration from xml.

There are 3 solutions:
1) either remove the functionality completely which would be breaking fro pre 1.6 code
2) instead of checking configuration just return hard coded mapping in _initOldFieldsMap (this will be non breaking for pre v1.6 code, but would not allow people to add/change mapping through xml) 
3) create a static runtime cache in every odel, so configuration is read only once

I would prefer solution 1)

See: https://github.com/lizards-and-pumpkins/magento-connector/issues/143

I've spotted the issue when profiling product sales report.
And saw that on every prodct creation Magento is asking configuration for some node, creating tons of objects which are discarded right after and this triggers PHP garbage collection.

![image](https://user-images.githubusercontent.com/515397/78546284-6f8a7e80-77fd-11ea-933c-02e218fb33c2.png)

kiatng:
I agree with option 1, remove it completely.

Flyingmana:
removing it completely is a serious BC break. (we also do not know, if people did misuse it in some way)

We will probably not include it in Version 19.

The process for serious BC breaks should in best case go over our RFC process, to give all participating parties enough time to comment on it.
https://github.com/OpenMage/rfcs

solution 3. seams to be a reasonable short term solution  which can be accepted in our current version 19.

I absolute support Solution 1, but it has to go over the RFC process.
And should probably suggest a deprecated message, so people notice they are using it.
Therefore we should in parallel take Solution 3.

tmotyl:
@Flyingmana 
I think that instead of trying to be non breaking at all cause, I would say we should do breaking changes but document them.
I think we should say "we don't support pre 1.6 extensions code". Magento 1.6 was released Nov 2009, so it was plenty of time for people to adopt the new API.
If people will complain, we can always add a compatibility layer as an external module.
We all have limited resources and I think we should spend it on making improvements rather than maintaining (for free) old code which maybe nobody uses.

Feel free to convert this issue and attached PR as RFC. My objective in next weeks is rather solving issues and perfromance optimizations so I will keep pushing new PR's to the LTS repo.

[...]

PR #921, opened 2020-04-06
Make oldFields mapping hardcoded to speed up system

tmotyl:
The mapping came with Magneto 1.6 to provide a backward compatibility
for old (pre 1.6) db field names.
Now the mapping is hardcoded in the class instead of saved
in the configuration.
This improves performance as Magento will not create tons of objects for
every new product/order/... entities.

Related: https://github.com/OpenMage/magento-lts/issues/920

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Do not use 20.x.x if you need IE support.
 - add redis as a valid option for `global/session_save` #1513
 - possibility to disable global search in backend #1532
 - reduce needless saves by avoiding setting `_hasDataChanges` flag #2066
+- removed support for `global/sales/old_fields_map` defined in XML #921
 
 For full list of changes, you can [compare tags](https://github.com/OpenMage/magento-lts/compare/1.9.4.x...20.0).
 
--- a/app/code/core/Mage/Catalog/Model/Product.php
+++ b/app/code/core/Mage/Catalog/Model/Product.php
@@ -362,7 +362,6 @@ class Mage_Catalog_Model_Product extends Mage_Catalog_Model_Abstract
      */
     protected function _initOldFieldsMap()
     {
-        $this->_oldFieldsMap = Mage::helper('catalog')->getOldFieldMap();
         return $this;
     }
 
--- a/app/code/core/Mage/CatalogInventory/Model/Stock/Item.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Stock/Item.php
@@ -177,6 +177,7 @@ class Mage_CatalogInventory_Model_Stock_Item extends Mage_Core_Model_Abstract
      */
     protected function _initOldFieldsMap()
     {
+        // pre 1.6 fields names, old => new
         $this->_oldFieldsMap = array(
             'stock_status_changed_automatically' => 'stock_status_changed_auto',
             'use_config_enable_qty_increments'   => 'use_config_enable_qty_inc'
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -526,7 +526,12 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
      */
     protected function _initOldFieldsMap()
     {
-        $this->_oldFieldsMap = Mage::helper('sales')->getOldFieldMap('order');
+        // pre 1.6 fields names, old => new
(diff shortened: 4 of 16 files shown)


colinmollenhour (dismissed):

tmotyl:
@colinmollenhour can you take a look at the issue #920
I've described there 3 options how to solve it.
To be honest I would love to start removing deprecated stuff from pre 1.6 era

Flyingmana (dismissed):
as mentioned in the related Issue, this is a serious BC break and should follow a certain process therefore.

tmotyl:
I agree it's breaking (no configuration option any more) but I would not call it serious BC. I have not seen a project or module which would use this configuration options.
But anyway feel free to make RFC out of it, you will have my vote ;)
I will not have time in the next weeks to proceed with formal RFC. But I will continue to push any new fixed I will find of course.

tmotyl:
FYI, there are few places more with getOldFieldMap:
order address
order creditmemo
order invoice
quote
quote address
quote item

sreichel:
> as mentioned in the related Issue, this is a serious BC break and should follow a certain process therefore.

If its about 11Y old extensions ... just drop it. (IMHO)

I guess with raising composer requirments to PHP7 we broke far more (users/agencies, that still rely on 5.x). If removal is a no-go I'd agree to hardcoded values ..

tmotyl:
@Flyingmana Can we agree on following:
for branch 1.9 I will make the values hardcoded inside _initOldFieldsMap functions, and for 20.x I will remove the method ?

ADDISON74:
This PR needs to be rebased to OpenMage:20.0

tmotyl:
@Flyingmana @fballiano This patch is now less breaking:
- no methods are removed
- mapping which was present in magento is still working

However adding new through configuration will not work ($this->_oldFieldsMap = Mage::helper('sales')->getOldFieldMap).
This mechanism of fetching config on every object creation was the biggest performance bottleneck. 

I'm ok with merging this patch to 1.9 branch as it's breaking potential is very limited, but If you prefer to have it only in v20, then lets' be it :)

luigifab:
I vote for: drop them all, everywhere.

fballiano:
pre 1.6? let's kill them already!

colinmollenhour (approved):

fballiano:
I think @luigifab and me wanted to say that if it's such an old functionality maybe we could remove it?

fballiano:
I don't know, since it was discussed to have it in v20 I've no idea what to do then :-D

tmotyl:
Will I have your vote if we keep it for 20 only? 
If yes then please vote and let's merge the thing :) lets not overthink it ;)
We can then promote v20 as much faster in some areas ;)

ADDISON74:
As far as I understand it would not be wanted in OM-19 because it would create (possible) trouble with a source code before Magento version 1.6, especially extensions. That means a very long time ago. Who else uses extensions before M1-1.6 that weren't updated later? Hard to say.

I personally think that if the PR is a real improvement to go both versions and if it will create issues we will find out then. It can be quickly reverted for any version and is only a commit. I am also for OM-19 (even I don't use this version in production, OM-20 is concrete).

colinmollenhour:
I propose we leave it alone for v19 (the PR is already based on v20 so we're good) and add a note to the README.md section of change between v19 and v20 along the lines of:

> Removed support for `global/sales/old_fields_map` defined in XML. #921

Then this can be merged and this PR will be listed in the release notes so users of v20 should do a quick check of their code for `old_fields_map` to see if they need to make any changes (highly unlikely afaik).

ADDISON74:
I agree with the opinion expressed by @colinmollenhour.

fballiano:
@tmotyl I can't add the note to the README since you locked the PR from modifications :-)

tmotyl:
@fballiano thats strange... consulting github documentation ... seems you can only enable "Allow edits from maintainers" on personal forks and not on organizational forks... see https://github.com/orgs/github-community/discussions/5634 
So please let me know what you want to add or lets add the note in the followup PR

fballiano:
I just wanted to add this line:

Removed support for global/sales/old_fields_map defined in XML. https://github.com/OpenMage/magento-lts/pull/921

to the README in the "differences between v19 and v20" as suggested by Colin ;-)

fballiano (approved):

ADDISON74 (approved):

PR #923, opened 2020-04-07
Fix 'No date part in '' found. Upgrade from 1.9.0.x to 1.9.2.2

seansan:
https://magento.stackexchange.com/questions/93202/no-date-part-in-found-when-upgrading-from-magento-1-9-0-x-to-1-9-2-2/99914

### DIFF
--- a/lib/Zend/Locale/Format.php
+++ b/lib/Zend/Locale/Format.php
@@ -905,9 +905,8 @@ class Zend_Locale_Format
 
         // split number parts
         $split = false;
-        preg_match_all('/\d+/u', $number, $splitted);
 
-        if (!empty($number) && count($splitted[0]) == 0) {
+        if ( ! preg_match_all('/\d+/u', $number, $splitted)) {
             self::_setEncoding($oenc);
             #require_once 'Zend/Locale/Exception.php';
             throw new Zend_Locale_Exception("No date part in '$date' found.");
--- a/lib/Zend/Locale/Format.php
+++ b/lib/Zend/Locale/Format.php
@@ -907,7 +907,7 @@ class Zend_Locale_Format
         $split = false;
         preg_match_all('/\d+/u', $number, $splitted);
 
-        if (count($splitted[0]) == 0) {
+        if (!empty($number) && count($splitted[0]) == 0) {
             self::_setEncoding($oenc);
             #require_once 'Zend/Locale/Exception.php';
             throw new Zend_Locale_Exception("No date part in '$date' found.");


sreichel:
I'd **not** merge this.

The error `No date part in ` is correct, because `0000-00-00 00:00:00` is not a valid date.

Even ...

```
sql> UPDATE `prefix_customer_entity` SET created_at =  "0000-00-00 00:00:00"
```

Results in ....
```
Incorrect datetime value: '0000-00-00 00:00:00' for column 'created_at' at row 1
```

colinmollenhour:
That may be an invalid date, but this patch will avoid there being a PHP warning. That is, the execution result for the case where the value being parsed is an empty string is the same, just without the PHP warning.

sreichel:
If `created_at` is empty or invalid semething went wrong before. (imho)

colinmollenhour:
If I recall correctly I've seen this error triggered many times from things like automated scanners. It is annoying that it throws PHP warnings because those get logged in a more urgent way for me so it is like a false alarm whereas fixing the PHP warning lets the error just bubble up to the user and I can happily ignore it. :)

sreichel:
I'll not approve this. Sorry.

colinmollenhour:
@sreichel Is this even related to `0000-00-00 00:00:00`? If it was, wouldn't the error be "No date part found in '0000-00-00 00:00:00'"?

sreichel:
You're right, but even an empty value is not intended .... the field isn't nullable.

Instead of "fixing" the output, i'd prefer input validation. (dont know where these wrong values come from)

seansan:
Why I logged this is because it breaks a quite important element in our
shop. Feed generation that halted altogether.

It certainly is something related to php versioning somewhere. But we don’t
know where.

So maybe both views are true?

sreichel:
> So maybe both views are true?

Of course! From merchants view i'm with you, but imho it is still wrong. Howerever, i'll neither approve nor block this.

colinmollenhour:
Ahh, I just pushed a commit and then realized that it actually doesn't fix anything because the `$splitted[0]` will still be an array even if there were no matches.. I was thinking that `$splitted[0]` would be an undefined index. Sorry for the confusion... I think maybe a better solution is just to track down individual issues where an empty string is parsed and fix them there. I'll push PR for the one in the stackexchange thread shortly.

colinmollenhour:
@seansan please see #980, does that address the issue in your view?

seansan:
thanks

the problem arises not in view, but when exporting data (so an extension or
otherwise is accessing the data and cannot handle it ... because it is
different than expected)

this is why we proposed solution 2 also (in addition to 1)

colinmollenhour:
@seansan I think any code accessing `created_at` where it is not guaranteed to have a valid value should check first if the value is not null before trying to parse it. So I think there is a strong case that it is a bug in the extension code. I don't think we should change the behavior of a core library that attempting to parse an empty string throws an exception.

PR #927, opened 2020-04-14
Fix file delete of customer attribute

kiatng:
[edit] This bug affects **file deletion** of cusotmer attribute in both frontend and backend. It was introduced in [ver 1.9.4.1](https://github.com/OpenMage/magento-lts/commit/62fae32153d34cfae1377cbc18937def7ef8c501):

```php
        if (!$attribute->isAttributeValidationPassed()) {
            return $this;
        }
```

in the class `Mage_Eav_Model_Attribute_Data_File`.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Attribute/Data/File.php
+++ b/app/code/core/Mage/Eav/Model/Attribute/Data/File.php
@@ -172,6 +172,9 @@ class Mage_Eav_Model_Attribute_Data_File extends Mage_Eav_Model_Attribute_Data_A
         }
 
         if (!$attribute->getIsRequired() && !$toUpload) {
+            if ($toDelete) {
+                $attribute->setAttributeValidationAsPassed();
+            }
             return true;
         }
 
--- a/.htaccess
+++ b/.htaccess
@@ -212,9 +212,17 @@
 ############################################
 ## Add default Expires header
 ## http://developer.yahoo.com/performance/rules.html#expires
-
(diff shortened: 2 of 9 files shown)


tmotyl:
Hi
Please provide a detailed description of the issue, how to reproduce it and what would you expect to happen.

kiatng:
To reproduce:

1. add a customer attribute of  `input_type` _file_, add the `attribute_id` to table `customer_form_attribute`, `form_code` _customer_account_edit_
2. render the attribute in frontend _My Account_ page for login customer
3. upload a file to the attribute and _Save_
4. select the _Delete_ checkbox and _Save_; expect file to be deleted, but it isn't

The reason it doesn't delete is because `$attribute->isAttributeValidationPassed()` is `false`.

app/code/core/Mage/Eav/Model/Attribute/Data/File.php
```diff
@@ -172,6 +172,9 @@ public function validateValue($value)
         }
 
         if (!$attribute->getIsRequired() && !$toUpload) {
+            if ($toDelete) {
+                $attribute->setAttributeValidationAsPassed();
+            }
             return true;
         }
```
kiatng, line 180: To trace the logic, please expand to code below to show the function `compactValue`, where the delete operation is carried out.

fballiano (approved):

ADDISON74 (approved):

PR #929, opened 2020-04-20
fix apply discount amount to FPT percent discount

Gaelle:
When apply percent discount, FPT were only added to first cart's item


colinmollenhour:
Hi @Gaelle can you provide some more info on exactly how this bug is manifested (how to reproduce) and why this fixes it?

Gaelle:
Hi,
You must : 
 - Apply FPT on products on back-office 
 - Create discount code with percent action
 - On front-office, you add minimum 2 products to your cart
- You apply your discount code on cart

Result : the discount amount is wrong because FPT is only added on first item of your cart.

In Magento code, break is IN the foreach items.

PR #930, opened 2020-04-21
Send order and agreement variables to the view

jacquesbh:
Especially the `$order` which is really really good to have there.  
It allows to custom the success page without loading the order again.


tmotyl:
Shoul the phtml file be also updated with this PR (to demonstrate the usefullness)?

jacquesbh:
> Shoul the phtml file be also updated with this PR (to demonstrate the usefullness)?

I see no need here since the phtml contains everything which is needed right now.
The issue is pure OOP for me, why load an object and not sending it to the view since it may be very useful. It leads to more load over the database, which makes no sens.

It's still possible to add a comment saying these two variables are available in the templates, but I don't see the need here.

tmotyl:
> "why load an object " 

where is the object loaded again?
If you're saying this patch can improve performance by passing existing variable to the view, then plase change also the view to use the performance benefit.

tmotyl:
So do I get it right, that default templates are not using the order objects (they are not fetching it), but it's a common customization to show some order properties, so its good to have order object available?

Flyingmana:
yes

PR #936, opened 2020-04-24
Fix SQL query quoting/casting when type is passed to where function

tmotyl:
The `$type` variable can be both string or int, so before comparing it to
`TYPE_CONDITION` string it has to be casted to avoid comparing integer zero
with string `(0 == 'TYPE_CONDITION')` which will wrongly return true,
and remove the information about type.

Pass type provided to where function down the chain to allow automatic
casting of arrays of values e.g. to int.

This fixes following cases:
```
->where('attr_table.store_id IN (?)', $storeIds, Zend_Db::INT_TYPE);
```
and
```
->where('attr_table.store_id = ?', $storeId, Zend_Db::INT_TYPE);
```
In both cases now passed value is correctly casted to int
(either single value, or each value from array)

Passing `Zend_Db::INT_TYPE` to where condition will prevent mysql performance
issues which might occur when mixed types are passed in "`in()`" condition.
related: https://github.com/OpenMage/magento-lts/pull/381

Also fixes type hints along the way.

Related PR on Magento2 side https://github.com/magento/magento2/pull/27980

### DIFF
--- a/lib/Magento/Db/Adapter/Pdo/Mysql.php
+++ b/lib/Magento/Db/Adapter/Pdo/Mysql.php
@@ -124,9 +124,9 @@ class Magento_Db_Adapter_Pdo_Mysql extends Varien_Db_Adapter_Pdo_Mysql
      * If an array is passed as the value, the array values are quote
      * and then returned as a comma-separated string.
      *
-     * @param mixed $value The value to quote.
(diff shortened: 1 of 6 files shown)


tmotyl:
FYI, it successfully went through the review in Magento 2.

PR #938, opened 2020-04-29
Fix _getSortedCollectorCodes to allow multiple totals calculation (and multiple PSPs)

seansan:
See explanation here: https://stackoverflow.com/questions/9194281/sort-algorithm-magento-checkout-totals-sorted-wrongly-causing-wrong-shipping-ta
and maybe too detailed here: https://github.com/mollie/Magento/issues/130

Problem is with totals calculation whit multiple PSPs

### DIFF
--- a/app/code/core/Mage/Sales/Model/Config/Ordered.php
+++ b/app/code/core/Mage/Sales/Model/Config/Ordered.php
@@ -126,8 +126,7 @@ abstract class Mage_Sales_Model_Config_Ordered extends Mage_Core_Model_Config_Ba
      *
      * @return array
      */
-    protected function _getSortedCollectorCodes()
-    {
+    protected function _getSortedCollectorCodes() {
         if (Mage::app()->useCache('config')) {
             $cachedData = Mage::app()->loadCache($this->_collectorsCacheKey);
             if ($cachedData) {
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Yes I think shell_exec may be ok for someone that needs a fast hack, but I don't think it belongs in an actual patch. The PHP project you linked to looks like a good solution.

seansan:
Hi! I see the same ... hence : the bug was now! and seeing EOL M1 a lot of PSP changes are being made. It is almost a shame M1 could not extend support jusssst a little longer into 2021 ... 

Adding a new library is too difficult to me ... so please advise. However the bug is real

seansan:
Possible other solution .. actually does not belong in the pull ... for discussion, maybe tax_subtotal extra xml lines can be added somewhere in core also solving the issue in general

                    
                    <tax_subtotal>
                        <after>nominal,subtotal,msrp,freeshipping</after>
                        <before>tax_subtotal,weee,shipping,tax_shipping,discount,tax,grand_total</before>
                    </tax_subtotal>

colinmollenhour:
Probably using composer is the correct way to include the library, but that will make composer a requirement to install OpenMage.. What are your thoughts on how to include composer modules @Flyingmana?

seansan:
see the comment in the pull on adjust tax order xml

colinmollenhour:
@seansan Please submit a PR for the XML change and explain what effects it has exactly.

fballiano:
This PR doesn't have updates since 3 years, should be rebased, has conflicts, uses file_put_contents which is really different from anything semilar done in OM. Sorry but for the time being I think it's better to close it.

I'll label it as "don't forget this PR" because it would be a very interesting topic but likely it will never be revived again.

Issue #934, opened 2020-04-24
(Possible) Layered Navigation Performance Improvement


rvelhote:
When building the layered navigation tree Magento will [fetch the full attribute option list](https://github.com/OpenMage/magento-lts/blob/2056b4bc8d7f3e033b7996acfc7c9eff24cfc30e/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Table.php#L54) of all attributes that have _Use In Layered Navigation_ set to _Filterable_.

It might be possible to cache these results using the built-in `initCache` method for collections. I defined two tags that will clear this cache but in order to have the best exprience we should also clear the tag cache for the attribute when saving the attribute  in `Mage_Adminhtml_Catalog_Product_AttributeController`.

```
        if (!isset($this->_options[$storeId])) {
            $key = 'ATTRIBUTE_OPTIONS_ID_' . $this->getAttribute()->getId();
            $tag = array(
                'eav', Mage_Eav_Model_Entity_Attribute::CACHE_TAG,
            );
            $collection = Mage::getResourceModel('eav/entity_attribute_option_collection')
                ->setPositionOrder('asc')
                ->setAttributeFilter($this->getAttribute()->getId())
                ->setStoreFilter($this->getAttribute()->getStoreId())
                ->initCache(Mage::app()->getCache(), $key, $tag)
                ->load();
            $this->_options[$storeId]        = $collection->toOptionArray();
            $this->_optionsDefault[$storeId] = $collection->toOptionArray('default_value');
        }
```
With this change I was able to reduce the number of executed SQL queries by 14 - it will depend on the amount of attributes you have defined of course.

I am happy with this change and I am currently testing in development but I was wondering is anyone sees any drawback/side-effects to this.

Thank you,
Ricardo

colinmollenhour:
Looks like a nice improvement! I don't see any issues with it. Would you submit a PR?

rvelhote:
Thank you for the feedback Colin. I will work on a full PR for further analysis as soon as possible!

joshua-bn:
Should this have been closed with #942 ?

sreichel:
@joshua-bn Thanks.

Fixed with #942.

PR #942, opened 2020-05-01
Cache all attribute options for display in layered navigation #943

rvelhote:
Related to issue #934. 

This will improve performance by reducing the amount of SQL Queries when fetching all possible attributes to display in the layered navigation. The amount of queries you will save is related to the amount of attributes you each store has configured to be available for layered navigation.

When saving attributes Magento will [clear the translate cache tag](https://github.com/OpenMage/magento-lts/blob/4c72dcd46a20ec1032b638c79280b453fc84e357/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/AttributeController.php#L315) so that cache tag as well as the EAV cache tag are included with this cache object.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Table.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Table.php
@@ -51,17 +51,16 @@ class Mage_Eav_Model_Entity_Attribute_Source_Table extends Mage_Eav_Model_Entity
             $this->_optionsDefault = array();
         }
         if (!isset($this->_options[$storeId])) {
-            $prefixId = 'ATTRIBUTE_OPTIONS_ID_' . $this->getAttribute()->getId();
-            $tags = array(
-                'eav',
-                Mage_Core_Model_Translate::CACHE_TAG,
-                Mage_Eav_Model_Entity_Attribute::CACHE_TAG,
+            $idPrefix = 'ATTRIBUTE_OPTIONS_ID_' . $this->getAttribute()->getId();
+            $tags = array_merge(
+                array('eav', Mage_Core_Model_Translate::CACHE_TAG),
+                $this->getAttribute()->getCacheTags()
             );
             $collection = Mage::getResourceModel('eav/entity_attribute_option_collection')
                 ->setPositionOrder('asc')
                 ->setAttributeFilter($this->getAttribute()->getId())
                 ->setStoreFilter($this->getAttribute()->getStoreId())
-                ->initCache(Mage::app()->getCache(), $prefixId, $tags)
+                ->initCache(Mage::app()->getCache(), $idPrefix, $tags)
                 ->load();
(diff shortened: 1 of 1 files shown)


colinmollenhour:
I may be missing something, but you do not need to add the translate cache tag because updating translations will not have a material affect on the attribute options.

You should add the individual id tag as well to be safe. So the final `$tags =` line can be:

```php
            $tags = array_merge(['eav'], $this->getAttribute()->getCacheTags());
```

rvelhote:
I included the translate cache tag because Magento also clears it after saving an attribute (with the comment `Clear translation cache because attribute labels are stored in translation`). If the user updates the attribute with a new translation then no one would see it until the translation cache was cleared.

The suggestion you made to use `$this->getAttribute()->getCacheTags()` is much better and it also includes `Mage_Eav_Model_Entity_Attribute::CACHE_TAG` plus an attribute specific tag. I will update the PR.

rvelhote:
I updated the PR to match @colinmollenhour suggestions.

I've been using this in production for a couple of months and works well in reducing the number of SQL queries. Of course it will hit the cache backend instead so if it's the database it's not much of a difference but if it's Redis or Memcached it will be better.

tmotyl:
Hi
I've took the patch for a little ride. Here are my findings:
1) The SQL query we will avoid looks like:
‌```
SELECT `main_table`.*, `tdv`.`value` AS `default_value`, `tsv`.`value` AS `store_default_value`, IF(tsv.value_id > 0, tsv.value, tdv.value) AS `value` FROM `eav_attribute_option` AS `main_table`
 INNER JOIN `eav_attribute_option_value` AS `tdv` ON tdv.option_id = main_table.option_id
 LEFT JOIN `eav_attribute_option_value` AS `tsv` ON tsv.option_id = main_table.option_id AND tsv.store_id = '7' WHERE (`attribute_id` = '184') AND (tdv.store_id = 0) ORDER BY main_table.sort_order ASC, value ASC
```
2) In my case on category page the `$storeId = $this->getAttribute()->getStoreId();` was always null, which resulted in storing optionArray under empty string in `$this->_options[$storeId]`.
I recommend changing the line to:

```
$storeId = $this->getAttribute()->getStoreId();
if (is_null($storeId)) {
      $storeId = Mage::app()->getStore()->getId();
}
```
The same way it's done in `setStoreFilter`. This way it will be clear what is stored in $this->_options.
Anyway, to be honest I don't see how would one be able to change attribute store to have multiple store ids inside `_options`, but its another story.

3) In my case I saw multiple calls to `getAllOptions` with the same attribute, so we might want to consider using a runtime cache too. I would be interested in getting feedback from you whether its also a case for you, or it depends on project customizations.

colinmollenhour:
@tmotyl A runtime cache usually makes sense because this will not be GCd in the lifetime of the request anyway and memory is cheap.

rvelhote:
@tmotyl I will add a local runtime cache. I only see one call for each attribute on my store so it might be something specific to your store.

PR #944, opened 2020-05-04
Upstream merge 1.9.4.5

Flyingmana:


joniklas:
I don't know if this is the right place to ask: 

But is there any background information why the password upgrade process is now checking the hash version against Mage_Core_Model_Encryption::HASH_VERSION_SHA256 and not against Mage_Core_Model_Encryption::HASH_VERSION_LATEST anymore? 

Isn't this reducing security? Thanks for pointing to some insides.

seansan:
Is this The V2 version of the patch? There seems to be a critical hack in v1 version https://twitter.com/colinodell/status/1260683841452363776

colinmollenhour:
@seansan Version 2 was applied, the `$authResult` check already exists. Thanks for checking, though!

joniklas:
@colinmollenhour And why is it checkin hash version against Mage_Core_Model_Encryption::HASH_VERSION_SHA256 and NOT Mage_Core_Model_Encryption::HASH_VERSION_LATEST ? Don't we want to use bcrypt implementation from PHP?

colinmollenhour:
There are two methods, `getHash` and `getHashPassword`. This update upgrades the former from MD5 to SHA256 by default and the latter remains using bcrypt as long as the password_hash function exists.

joniklas:
@colinmollenhour Also this upgrade changes some occurrences of getHashPassword to getHash. (in Mage_Admin_Model_User::_getEncodedPassword, Mage_Api_Model_User::_getEncodedApiKey and Mage_Customer_Model_Customer::hashPassword)
Aren't these the places where we want bcrypt!?

colinmollenhour:
You are probably right, but I don't know the reasoning behind why the Magento team made those changes.. If you want to submit a working and tested PR that changes it we would be happy to review and consider changing it.

PR #955, opened 2020-05-08
added some missing top level domain names

paulrodriguez:
Some top-level domain names are missing.
the ones I found that kept customers from creating an account were:
blog
design
group
page
pet
phone


paulrodriguez:
never mind.
didnt see that the file was fixed under the core directory

Flyingmana:
still thank you. thats also a hint for something we should clean up at some point.

Probably by moving to one of the ZF1 forks, which still provide maintenance so we have forces joined.

PR #963, opened 2020-05-11
Dropped useless semicolons

sreichel:


Flyingmana:
For this specific change, Iam against.
But maybe I do not understand the benefit.
Which Coding Style/Tool does recommend removing them, and for what reason?

sreichel:
> For this specific change, Iam against.
> But maybe I do not understand the benefit.
> Which Coding Style/Tool does recommend removing them, and for what reason?

Oh, thought this is a no-brainer ... These semicolons are superfluous and i guess nobody would write code like this. Also phpStorm shows a weak warning for.

IMHO ... M1 is EoL very soon, so why not fix also small code style issues?

tmotyl:
@Flyingmana to valiate you can run php lint on these classes.
I also think this is a nobrainer

kiatng:
The benefit is coding style consistency.

PR #965, opened 2020-05-11
Fixed calls to static methods

sreichel:


sreichel:
How to re-run checks?

colinmollenhour:
It passed after merging. :)

boesbo:
Questo ha generato un bug: https://github.com/OpenMage/magento-lts/issues/1135

tmotyl:
FYI the error @boesbo mentions is related to usage of "self::" instead of "static::".
I think this patch should be revisited to see if in other places we should also use.

sreichel:
@boesbo yes i did. This changes were made manually and i've checked for this, but missed that one. Sorry.

@tmotyl already reviewed this PR and found nothing else.

boesbo:
> @boesbo yes i did. This changes were made manually and i've checked for this, but missed that one. Sorry.
> 
> 
> 
> @tmotyl already reviewed this PR and found nothing else.
> 
> 
> 
> 

!thank you so much!

Issue #959, opened 2020-05-10
Solve #912 without the use of the error suppressor


Flyingmana:
Find a way to solve #912 without the use of the error suppressor `@`, so we do not cause Issues if `scream.enabled` is activated

sreichel:
Fixed in #966

ADDISON74:
Was this report resolved by #966? It was closed in June 2020 and reopened on the same day.

fballiano:
I think all of the PRs related to this issue are merged, closing

PR #966, opened 2020-05-11
Check for null byte at the time of writing a file.

damien-biasotto:
Good afternoon everyone,

I decided to tackle the #959 issue, by adding a byte null check in the protected method `_IsValidSource`. 

However, I introduced a new behaviour, I arbitrarily decided to flag any strings that contain a null byte character as invalid (for safety reasons).

Then removed any null byte related error suppressors.

I'm not sure what are the guidelines for Pull Requests. Shall I provide a unit test to highlight this change?

### DIFF
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -486,7 +486,8 @@ class Varien_Io_File extends Varien_Io_Abstract
      */
     protected function _IsValidSource($src)
     {
-        if (is_string($src) || is_resource($src)) {
+        //Treat string that contains a null byte as invalid
+        if ((is_string($src) && strpos($src, char(0)) === false) || is_resource($src)) {
             return true;
         }
 
@@ -505,7 +506,7 @@ class Varien_Io_File extends Varien_Io_Abstract
     {
         $error = false;
         @chdir($this->_cwd);
-         if (file_exists($filename)) {
+        if (file_exists($filename)) {
             if (!is_writeable($filename)) {
                 $error = "File '{$this->getFilteredPath($filename)}' isn't writeable";
             }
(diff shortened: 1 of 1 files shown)


tmotyl:
what else can cause a warning except for the null byte?

damien-biasotto:
> what else can cause a warning except for the null byte?

Well, the documentation is not really exhaustive : 

> Upon failure, an E_WARNING is emitted.

I'd assume if you're not passing a string nor a resource?

tmotyl:
hmm seems this change introduced an issue: https://github.com/OpenMage/magento-lts/pull/1084
How can it be possible that the problem was not caught by linter?

sreichel:
@tmotyl this is no problem with linter, it's a problem with reviews ... :(

I've checked https://www.php.net/manual/en/security.filesystem.nullbytes.php, but did not see the typo in that PR AND it was not really tested. `Varien_Io_File::write()` is only used in a few places so this does not popup during normal frontend/backend tests.

A clear description, **steps to reproduce** and maybe a code snippet (if easier) should be mandatory.

damien-biasotto:
Oh my, I'm really sorry guys. I don't know how i didn't catch that.

Flyingmana:
@tmotyl I added an Issue for this, I also think we should have a static analyser, which can find such a case.
https://github.com/OpenMage/magento-lts/issues/1087

The current one did not catch it, as we just check for syntax error, and technically this is not a syntax error.

digitalpianism:
I made a PR to address the bug that was introduced by that change: https://github.com/OpenMage/magento-lts/pull/1302

Tomasz-Silpion:
After hours of debug why our image import has stopped to work what was spotted more than 2 weeks from OpenMage update I got into the same Varien_Io_File _IsValidSource method _$src, chr(0)_ check that fails on base64 content. Replacing the class with @digitalpianism patch has solved the problem. @damien-biasotto @tmotyl can you review the change and merge it for the next release?

PR #969, opened 2020-05-12
Fix checkout address for guest order

spinsch:
If the guest wants to change the shopping cart again during checkout or the payment process has failed, the guest does not have to fill in his address again.


colinmollenhour:
The behavior change makes sense to me, it is extremely frustrating to lose all progress if you leave checkout.. Do some people consider this a possible security risk, though?

What about just making the `else` case always load the model from the quote? E.g., remove the added lines and change the last line:

```php
            } else {
                $this->_address = $this->getQuote()->getBillingAddress();
            }
```

kiatng:
I agree with @colinmollenhour, the code can be simplified to

```php
    public function getAddress()
    {
        if (is_null($this->_address)) {
+           $this->_address = $this->getQuote()->getBillingAddress();
            if ($this->isCustomerLoggedIn()) {
-               $this->_address = $this->getQuote()->getBillingAddress();
```

spinsch:
I'm probably always a little too careful with changes like this.

PR #989, opened 2020-05-19
Apply some Brand Changes, to have Magento appearing less

Flyingmana:
Apply some Brand Changes, to have Magento appearing less

Also change some URLs and the display of the OpenMage Version


Flyingmana:
I also wanted to change the Logo, but had some Issues related to Quality. (Magento does use .gif in most places, and I was not able to render good looking versions for the size used to display it.)

I can provide the original PSD file if anyone wants to take this over.

PR #992, opened 2020-05-20
update install instructions to reflect the further documentation link

fitzjohnit:
the current composer install instruction below doesn't work:

`
"openmage/magento-lts": "1.9.4.x"
`

so have updated to reflect the instructions in http://www.openmage.org/magento-lts/install.html that do

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,18 +7,18 @@ level of backwards compatibility to the official releases.
 **Pull requests with unofficial bug fixes and security patches from the community are encouraged and welcome!**
 
 Though Magento does not follow [Semantic Versioning](http://semver.org/) we aim to provide a workable system for
-dependency definition. Each Magento `1.<minor>.<revision>` release will get its own branch (named `1.<minor>.<revision>.x`)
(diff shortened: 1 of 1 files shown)


Flyingmana:
There is a bit of confusion.
The part for the Magento Releases regarding branches is still correct.

the branches for them are still `1.<minor>.<revision>.x`

We do have additionally our own Major Versions with own branches http://www.openmage.org/supported-versions.html

README.md
```diff
@@ -7,18 +7,18 @@ level of backwards compatibility to the official releases.
 **Pull requests with unofficial bug fixes and security patches from the community are encouraged and welcome!**
 
 Though Magento does not follow [Semantic Versioning](http://semver.org/) we aim to provide a workable system for
-dependency definition. Each Magento `1.<minor>.<revision>` release will get its own branch (named `1.<minor>.<revision>.x`)
+dependency definition. Each Magento `1.<minor>.<revision>` release will get its own branch (named `1<minor>.<revision>.x`)
```
Flyingmana, line 10: This line needs to be changed to <major>.<minor>.x

PR #996, opened 2020-05-20
Fixed bug on Exception message: Notice: Trying to access array offset on value of type bool  in /app/code/core/Mage/Catalog/Model/Resource/Url.php on line 388

kiatng:
This happened when I tried to reindex _Catalog URL Rewrites_ in backend.


kiatng:
I did something wrong, will try again.

Issue #987, opened 2020-05-19
1.9.4.5 (SUPEE-11314) downgrades password hashes to SHA256 (from "latest" Bcrypt)


joniklas:
The patched password hash upgrade functionality changes password hashes if the hash version is NOT SHA256. This means is also changes password hashes if the "latest" hashing type Bcrypt is used! So it is a kind of downgrade. 

This affects `customers`, `admin` users and `api` users.

**How to reproduce:**
* Check in database the password_hash for an excisting customer/admin is bcrypt (starts with $2y$...)
* Login with a customer/admin
* Check in database the password_hash is now SHA256

**Code References:**
Customer:
https://github.com/OpenMage/magento-lts/commit/16dc8f76f1f6bbbaf18005e88ae08df218d60aac#r39249967
```
if (Mage_Core_Model_Encryption::HASH_VERSION_SHA256 !== $currentVersionHash) {
$model->changePassword($password, false);
}
```
https://github.com/OpenMage/magento-lts/commit/16dc8f76f1f6bbbaf18005e88ae08df218d60aac#r39249998

**Enterprise Editon:**
In EE the end result of the password hashes is okay and stays with bcrypt. But in EE the password hash gets changes on each and every login since it also compares against `HASH_VERSION_SHA256` instead of `HASH_VERSION_LATEST`.

I expect OpenMage Community Edition also wants to stay with bcrypt for password hashing?

colinmollenhour:
Thanks for your clear bug report, @joniklas. Can you submit a PR with a proposed fix? I think we all will agree that bcrypt is the preferred algorithm where available but I can't say when any of us will have time to create a patch ourselves.

jfksk:
That really looks like Adobe did this for M1 -> M2 migrations.
M2 uses Argon2id via Sodium, if available, and SHA256 as fallback
[Here is the M2 Encryptor](https://github.com/magento/magento2/blob/2d09485b8e00d149e936412dc0f7baa78a928931/lib/internal/Magento/Framework/Encryption/Encryptor.php#L163)

As, to my understanding, Argon2id is superior to bcrypt for storing passwords, I'd suggest to do the same. If compatibility is not a goal, bcrypt, being superior to SHA256, could be used as Fallback.

[...]

PR #998, opened 2020-05-21
Password hash upgrade to strong algorithms

jfksk:
As described in #987, Magento downgrades hashes to SHA256. Besides that, hashes are never upgraded to stronger algos. These changes upgrade all password hashes (admin, api, customer) to the latest and greatest that's offered by password_hash. 
There is no hash-version anymore, there is only the latest version offered by php. ~~The PASSWORD_DEFAULT version is bumped to  PASSWORD_ARGON2ID - if available. That's hardcoded, not sure if that is needed and/or desirable.~~ 
The intend of this PR is first and foremost to have something to discus about.

- Optimize customer password hash-upgrade and removed downgrades to SHA256
- Optimize admin password hash-upgrades and allow for upgrades past SHA256
- Password hashes are upgraded with password_hash: bcrypt or Argon2
- Add API user apy_key hash upgrades
- Add hash-upgrade logic for password_hash
- Remove password hash salts altogether: password_hash deals with it
- ~~Bump hash algo to Argon2, if available~~
- Use the encryptors getHashPassword, not getHash for passwords
- getHash now always returns SHA256 hashes
- Ensure that the hash validation is also executed when the user does not exist or a simple hash (MD5/SHA256/SHA512) validation failed


app/code/core/Mage/Core/Model/Encryption.php
```diff
@@ -81,25 +95,36 @@ public function getHash($password, $salt = false)
...
      */
     public function getHashPassword($password, $salt = null)
     {
-        if (is_integer($salt)) {
-            $salt = $this->_helper->getRandomString($salt);
+        return $this->hash($password, self::HASH_VERSION_LATEST);
+    }
+
+    /**
+     * Generate hash for customer password
```
joniklas, line 115: Comment outdated? Does not reflect return type. Remove?

jfksk: aye thx, there were more wrong... I fixed them...

colinmollenhour:
Excellent work!

jfksk:
I removed Argon2 mainly because of my inability to figure out what sane defaults are. The PHP ones seem to be a bit excessive.  The sodium ones seem to be better... We probably should just wait until the PHP core team sets Argon2 as default.
Anyhow, it shed light on another problem. It was possible to tell if a user exist by checking the response times. I found that bcrypt hashing was already slow enough to make a difference. I tried to mitigate that by always executing an hash-validation.

Any concerns about this?

956e3c575fc61b7774187d94ab47209bbe53a818
76ad021aaf689c7860cc035b0ca7ca02ae203375

jfksk:
One more important thing I'd like to highlight:
Mage_Core_Model_Encryption::getHash vs Mage_Core_Model_Encryption::getHashPassword was a BC break introduced by Magento in 19.4.3
99b772f94de693228a787d69254911e060934cf1
(I consider that a breaking change, because the encryption model is replaceable via configuration [1]) 

By the looks of it it already broke modules like Ikonoshirt_Pbkdf2.

Just using Mage_Core_Model_Encryption::getHash again IMHO wont help much, as bcrypt hashes where already introduced and custom modules still have to check whether validate against the core logic or use the custom logic. On top of that, there is now also the upgrade logic introduced in 1.9.4.4. At least this PR allows you to change the behavior by overwriting the new method Mage_Core_Model_Encryption::passwordHashNeedsUpgrade

While this PR is not introducing totally new problems, its certainly not helping in that case...

[1] https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Core/etc/config.xml

joniklas:
@jfksk Concerning configurable encryption modeL: 

Encryption model is configured in EE as: `<encryption_model>Enterprise_Pci_Model_Encryption</encryption_model>`.

Also `Enterprise_Pci_Model_Encryption` is just a small rewrite of `Mage_Core_Model_Encryption` with changed defaults (SHA256 instead of MD5, and still Bcrypt used).

jfksk:
@joniklas That's interesting. As I do not have a current M1 EE under my hands, do you know by chance, if they also changed from the methods getHash to getHashPassword and alogs back and forth?  Just out of curiosity... I wonder if they did there anything different... 

I pointed that out, because I could not come up with a better solution. To me its fine to do it that way. You'd need to support m1.9.4.3 and m1.9.4.4 in a custom module to keep it compatible and if you only support OpenMage, it can be a little bit easier...

joniklas:
@jfksk In EE they also changed from getHash to getHashPassword back and forth. Mostly just by inheriting from Core (Mage_Core_Model_Encryption).

1.) But the default value of `hash()` in EE is different:
`hash($data, $version = self::HASH_VERSION_SHA256)`
2.) and hash() also includes a "latest" case with `password_hash()` since last year. And this is unchanged.

joniklas:
So we have a statement, that the regeneration of the password is intended. To keep the compatibility of password in the migration process from M1 to M2.

So the question is still there, to keep an upgrade path wich is not an downgrade to SHA256. Maybe keeping Argon as upgrade path with a default configuration disabled could be an option to keep an potential upgrade path to M2 for OpenMage.

jfksk:
Oh dear, I totally overlooked that they are specifically not using hash_hmac for passwords -  so sha256 hashes are compatible... :facepalm:

@joniklas What I do not like about Argon2 in M2 is that they are using sodium_crypto_pwhash instead of sodium_crypto_pwhash_str or password_hash.
That means that M2 cannot take advantage of fully auto upgrading hashes and the hashes are not verifiable with sodium_crypto_pwhash_str_verify or password_verify.
If we do that, we'd therefore pass on automagic password hashing.

I'm not a big fan of diverging just yet, but if we don't, we need some sort of a compromise. 
Which kept me thinking, if it might be acceptable to offer a encryption model (maybe in a separate module) for merchants that want to migrate?
I further noticed, that the Data Migration Tool now comes with migration logic for bcrypt hashes. Maybe that's already enough? [They just added that](https://github.com/magento/data-migration-tool/commit/3c7f83b8eaab690734113f574a26595e61415b62)

app/code/core/Mage/Admin/Model/User.php
```diff
@@ -395,7 +397,7 @@ public function authenticate($username, $password)
             $this->loadByUsername($username);
             $sensitive = ($config) ? $username == $this->getUsername() : true;
 
-            if ($sensitive && $this->getId() && Mage::helper('core')->validateHash($password, $this->getPassword())) {
+            if (Mage::helper('core')->validateHash($password, $this->getPassword()) && $sensitive && $this->getId()) {
```
andthink, line 400: Testing sensitive first would be faster than testing validateHash

jfksk: That's why I swapped it. It can be possible to check if the username exists if its the other way around. AFAIK one couldn't tell with simple hashes (sha256/512) but bcrypt is quiet a bit slower.

Flyingmana:
@jfksk if it is about diverging, this sounds like something for one of the non-LTS releases (20.x or 21.x for example)

jfksk:
@Flyingmana I'm pretty torn on that subject because I am not 100% about the consequences as Magento switched back and forth on that topic. 
If that change cannot be identified as uncritical for users switching to OpenMage it's defiantly better to deal with it the next major version - or at least after some time - depending on the schedule... In that case I'll might publish a small module containing these changes - for now.

jfksk:
Is that something that needs to be handled? 
https://marc.info/?l=php-internals&m=161905610413116&w=2
IDK why one would fill his/her password w/ null-bytes but maybe there is some side-vector?

ADDISON74:
We should discuss this PR a little and decide what to do with it.

jfksk:
I cannot recommend this PR, as it is  anymore, because I already changed quiet a few more things in my fork, for a more robust solution , so I am closing it.

This is not bad - but its not complete - and there is no interest at the moment.

If there is some interest in the future, write me on discord or a email, if you like. It's IMHO fine at it's core, but there are a few more things to consider, in the grand scheme.

pquerner:
Seems like you since then deleted your module. Too bad :(

PR #1000, opened 2020-05-22
Update issue templates

sreichel:
Addes templates bases on M2 (copied) .... kept severity level, but maybe needs changes ...


colinmollenhour:
Nice! Do we need to add a .htaccess file to be safe? There is no security risk, just generally should block access to files that are not intended to be public.

sreichel:
@Flyingmana approve again?

PR #1002, opened 2020-05-23
Removed Mage_GoogleBase

sreichel:
Ref #416

`GoogleBase` is deprecated since 1.5.1 .... can we remove it?


Flyingmana:
The Question is, does it still work?
And if yes, then this has to go through an RFC

sreichel:
1.6.0.0-alpha1 changelog .... https://github.com/OpenMage/magento-lts/commit/a562c7f972d3809ef933a69dc58283b2dd85cb25#diff-ccb38245ef0f8b85d3770014ac70334b

It does not work anymore. All code is commented ...

PR #1007, opened 2020-05-27
Fix CatalogRule start date insert logic

mmenozzi:
For the same reason the rule is inserted until 24 hours after the rule end date, we should also start inserting it 24 hours before the start date. Am I wrong?


kiatng:
I think the code is correct. I made a test:

```php
// Original code.
$fromTime = (int) strtotime('2020-01-01');
$toTime = (int) strtotime('2020-01-02');
$toTime = $toTime ? ($toTime + 86400 - 1) : 0;
$result = [date('r', $fromTime), date('r', $toTime)];
```
The result is
```
array(2) {
  [0] => string(31) "Wed, 01 Jan 2020 00:00:00 +0000"
  [1] => string(31) "Thu, 02 Jan 2020 23:59:59 +0000"
}
```

Now with your code:
```php
// Changed code.
$fromTime = (int) strtotime('2020-01-01');
$fromTime = $fromTime ? ($fromTime - 86400 + 1) : 0;
$toTime = (int) strtotime('2020-01-02');
$toTime = $toTime ? ($toTime + 86400 - 1) : 0;
$result = [date('r', $fromTime), date('r', $toTime)];
``` 
The result is wrong:

```
array(2) {
  [0] => string(31) "Tue, 31 Dec 2019 00:00:01 +0000"
  [1] => string(31) "Thu, 02 Jan 2020 23:59:59 +0000"
}
```

mmenozzi:
When we faced this issue it was a timezone issue. I remember that the day before the Black Friday we set a rule to star on the next day. We are in the Europe/Rome timezone but we have an Australian store view with Sidney timezone, so we expected to see the rule already applied in the AU store view because it was already the Black Friday day there.

I remember we debugged locally and the problem was that the rule wasn’t added at the moment we created it but it should because of the Sidney timezone. So we fixed with this change.

kiatng:
Could it be that the Australian store locale was not configured correctly in the backend?

mmenozzi:
No it was configured correctly.

seansan:
There have been some changes on this in the past also ... same problem ... I remember it was fixed (Amsterdam timezone) .... it has to do with time of running, core time of the (unix) server and the timezone.

ADDISON74:
We should move forward this PR. @kiatng - what is your opinion?

fballiano:
I coulnd't really understand why this solutions fixes the problem but I think we should never use +-86400 seconds since it doesn't work the day of the daylight saving +-1 hour change.

kiatng:
> We should move forward this PR. @kiatng - what is your opinion?

I think we need to replicate the issue first.

ADDISON74:
In this case we need to replicate this issue first.

ADDISON74:
I would leave it open. I will make time to replicate the issue. If it confirms I will go further to implement the PR and check what is happening.

fballiano:
@ADDISON74 I would close this, @kiatng tested the coded and seemed not work https://github.com/OpenMage/magento-lts/pull/1007#issuecomment-635028713 (that result is wrong for sure) but using SECONDS_IN_DAY is also not a super good practise since it doesn't work the day of daylightsaving change.

fballiano:
maybe it could be converted to an issue but we need a full reproducible case

mmenozzi:
Hi guys, I'll try to provide a way to reproduce the issue in the next days.

mmenozzi:
Hi guys, the bug still exist. At least on version 19.4.21 of OpenMage.
Here the steps to reproduce it:

1. First of all make sure that the current time of the machine you'll use to do the test is after the 3 PM.
2. Install OpenMage 19.4.21 with sample data, paying attention to configure Europe/Rome as default timezone
3. Login to admin panel and create a website, called "Australian", with a store and a store view in it
4. Go in System -> Configuration -> General -> General, select the "Australian" website in the scope menu and configure the "AUS Eastern Standard Time (Australia/Sydney)" as timezone for the "Australian" website.
5. Go to Catalog -> Manage products, select all products, select the "Update attributes" mass action and add all products to the "Australian" website.
6. Now go in Promotions -> Catalog Price Rules and create a new active rule starting from tomorrow (for example if today is 2023-01-16, set the start date to 2023-01-17), which discounts all the products by 20%. Apply the rule, reindex everything and clear the cache.
7. Now go in the frontend, in the Australian store view. Search for the product "AVIATOR SUNGLASSES" and open its product page.

If the current time of is 3 PM of the 2023-01-16 in Europe/Rome, it means that in the Australia/Sydney timezone the current time is 2 AM (or 1 AM depending on the daylight saving time) of the 2023-01-17. This means that on the "Australian" store view we should see the discounted price of 236 for the "AVIATOR SUNGLASSES" but what happens is that we still see the full price of 295.

If you apply the patch provided in this PR and you save and apply the catalog price rule again you'll get the discounted price.

/cc @fballiano @ADDISON74 @kiatng @seansan

kiatng:
Note changing `fromTime` to one day earlier may cause the rule to be applied too early, although it works in the test case. Here, `$fromTime` is inserted in the table.

https://github.com/OpenMage/magento-lts/blob/7c16d61c6512e8b64fe9dff4df799f0c36ba9ae4/app/code/core/Mage/CatalogRule/Model/Resource/Rule.php#L233-L238


Instead, we can modify `$timestamp` so that it's one day after:

```php
$timestamp = time() + self::SECONDS_IN_DAY;
```

Now, the rule can be added ahead of time:
https://github.com/OpenMage/magento-lts/blob/7c16d61c6512e8b64fe9dff4df799f0c36ba9ae4/app/code/core/Mage/CatalogRule/Model/Resource/Rule.php#L208-L213

@mmenozzi Can you test line 208 `$timestamp = time() + self::SECONDS_IN_DAY;` without changing `$fromTime` and see if it works for you?

sreichel:
> Note changing fromTime to one day earlier may cause the rule to be applied too early

Also seen this .... the topic is slowly growing a beard ...

luigifab:
I tested the test case with and without the PR, and this is not working, like @kiatng says the rule is applied to early.

**australian website: ok for sydney:**
image

**default website: ko for paris:**
image

**my rule 20% for tomorrow:**
image

kiatng:
Line 209:
```php
        if ($fromTime > $timestamp
            || ($toTime && $toTime < $timestamp)
        ) {
```
The above conditions restrict the number of rules to be added to the table. If the condition(s) is(are) true, it will skip the rule. The problem with this PR is that it changed the `$fromTime`, which is then added to the table resulting in the rule being applied too early.

Instead, we can change it to
```php
        if ($fromTime > $timestamp + self::SECONDS_IN_DAY
            || ($toTime && $toTime < $timestamp)
        ) {
```
Note that the comparison for `$toTime` is correct as it is because the time is passed due date, the rule is skipped. However, the `$fromTime` is compared to 24 hours later; if `$fromTime` is not yet tomorrow, add the rule. Why this `$timestamp + self::SECONDS_IN_DAY`, has 24 hours  added? It is because the cron runs only once a day. If the cron runs every 12 hours, then we do `$timestamp + self::SECONDS_IN_DAY/2`

@luigifab Can you test if this works:

luigifab:
It seems that the @kiatng solution works. @mmenozzi can you try?

I need to run `catalogrule_apply_all` (same as apply rules I think, because the cron job is started at night when my computer is off). Perhaps it is necessary to run this cron job more than once per day.

mmenozzi:
Hi @luigifab,
@kiatng was right: with my PR the rule were applied too early. I tested it on my machine.
I updated the PR with his suggestion.

Thank you guys!

kiatng (dismissed):

luigifab (dismissed):
I have tested with a local env.
I don't use catalogrule on real store.

app/code/core/Mage/CatalogRule/Model/Resource/Rule.php
```diff
@@ -206,7 +206,7 @@ public function insertRuleData(Mage_CatalogRule_Model_Rule $rule, array $website
         $toTime = $toTime ? ($toTime + self::SECONDS_IN_DAY - 1) : 0;
 
         $timestamp = time();
-        if ($fromTime > $timestamp
+        if ($fromTime > $timestamp + self::SECONDS_IN_DAY
             || ($toTime && $toTime < $timestamp)
         ) {
```
fballiano, line 211:
```suggestion
        if ($fromTime > strtotime('+1day', $timestamp) || ($toTime && $toTime < $timestamp)) {
```
his shouldn't create problems with daylight saving time change etc

fballiano:
I know this PR is approved but I think it's not correct to sum the number of seconds in a day, since it will be broken 2 times a year (eg https://github.com/OpenMage/magento-lts/issues/3115 although it's another topic).

I wrote a couple of comments with another syntax which shouldn't suffer from that problem.

fballiano:
so sorry guys, I've rebased the PR and it screwed everything... I've rebuilt it (adding my strtotime() suggestions), do you think it misses something?

fballiano:
The 86400 second cannot be used, it does not work 2 days a year, that i am more than sure

ADDISON74:
This issue reminds me of another one that I discussed in another forum https://magento.stackexchange.com/questions/133405/issue-when-specialpricefromdate-is-equal-with-specialpricetodate/

What happens when price_from is perfectly equal to price_to? For example both having the value 2023-07-16 00:00. Naturally it should run for a day.

luigifab:
Is it possible to deploy the PR on demo.openmage.org to test it again?

sreichel:
> so sorry guys, I've rebased the PR and it screwed everything... I've rebuilt it (adding my strtotime() suggestions), do you think it misses something?

@kiatng is the original PR still available?

theroch:
I've updated this PR with the latest main branch and kept the carbon changes from @sreichel and restored the original patch from @kiatng.
Can someone test it again, maybe @luigifab?

PR #1009, opened 2020-05-29
Allow "email" to be used in address templates

woutersamaey:
Only fields from the address can be used in address templates, but a client of ours needed the "email" on the Invoice PDF, so I thought it made sense to have access to the "email" var in the address fields.

There are other options, like building the $data array in a new protected method, that could then be overridden, or throw a new event, but I wanted to start simple. This kind of thing probably won't be extended much.

### DIFF
--- a/app/code/core/Mage/Customer/Block/Address/Renderer/Default.php
+++ b/app/code/core/Mage/Customer/Block/Address/Renderer/Default.php
@@ -135,6 +135,11 @@ class Mage_Customer_Block_Address_Renderer_Default extends Mage_Core_Block_Abstr
             }
         }
 
+        if($address instanceof \Mage_Sales_Model_Order_Address){
+            $order = $address->getOrder();
+            $data['email'] = $order->getCustomerEmail();
+        }
+
         if ($this->getType()->getHtmlEscape()) {
             foreach ($data as $key => $value) {
                 $data[$key] = $this->escapeHtml($value);


sreichel:
I like the idea, but why not dispatch a new event to change `$data`? So this would not be limited to customers email.

colinmollenhour:
> I like the idea, but why not dispatch a new event to change $data? So this would not be limited to customers email.

As email is pretty broadly useful I think it makes sense to still add it so it can be used with no additional event observers, so I think a case could be made for adding both methods.

woutersamaey:
@sreichel what name would you suggest for the event?

sreichel:
Something like `<module>_<type>_<model>_<method>` .... `customer_block_address_render` or `customer_address_render` ...

sreichel:
> 
> 
> > I like the idea, but why not dispatch a new event to change $data? So this would not be limited to customers email.
> 
> As email is pretty broadly useful I think it makes sense to still add it so it can be used with no additional event observers, so I think a case could be made for adding both methods.

IMHO .... I'dont want to have that hard-coded `$data['email']` in ... even if its unlikly, if you have a custom `email` attribute for addresses, it would be overwritten .... i prefer a new event :)

ADDISON74:
... and after two years this PR tays in phrase to be or not to be an event.

fballiano:
mmm I wouldn't add the email to an address, the emailaddress can be easily accessed from the $order->getCustomerEmail()

fballiano:
I'd say the idea of making it even is probably the best but we can't modify this PR since it's blocked.

@ADDISON74 I'd close it, if somebody needs this feature they'll create a new PR, at the moment I wouldn't work on this personally.

ADDISON74:
I would set it to draft. After a while we can evaluate it again and create a new PR.

ADDISON74:
@sreichel you made some interesting observations. We keep this the idea of this PR, if you have time please implement your version.

sreichel (approved):
> mmm I wouldn't add the email to an address, the emailaddress can be easily accessed from the $order->getCustomerEmail()

:+1: 

As already said .... _ even if its unlikly_ ...

sreichel:
@woutersamaey can you please change your repo settings to allow edits?

ADDISON74:

Issue #1019, opened 2020-06-03
Emails should not contain passwords in plain text


cdbessig:
Emails, for one example, app\locale\en_US\template\email\account_new.html contain the user's password in plain text.  This should be corrected throughout all locals to not send users passwords.

This is now considered best practice.  This has always annoyed me that magento 1.9.4 never took care of this.

Now when I moved to openmage this was overwritten and I got the default functionality back.  Since openmage is meant to be Long term support AND feature enhancement of M1, I would consider this as a valid feature request, or even a security issue (possible leak).

What I propose is in each locale change this line:

```
<strong>Password</strong>: {{htmlescape var=$customer.password}}
```

To something like this:

```
<strong>Password</strong>: -- What you entered when you signed up --
```

colinmollenhour:
Thanks for opening this issue, Brian. We discussed this at length at #307 and I think came to a conclusion on what to do but it hasn't been executed on yet. If you can submit a PR that would be awesome!

kkrieger85:
Shame on me. I assigned #307 and forgot about it. 

PR coming today

sreichel:
Fixed in #1021

PR #1022, opened 2020-06-04
1.9.3.x => E-Mail templates only: replace plaintext password with hint of choosen password

### LINKED ISSUE #307: Remove password from account_new email template

kkrieger85:
See  #307

_To replace plaintext password with <strong>Password</strong>: (the password you set when creating your account) and also remove reference to plaintext password in email templates._

Fixes #307


kkrieger85:
Fixes #1019 too

Issue #725, opened 2019-06-15
[WIP] Cleanup: Round 1 - DOC block updates


sreichel:
Just an overview for already opened or ready PRs ... I'll add further PRs as soon I think I#m done with round 1 :)

Round 1:

- just doc blocks
- PSR2 automatic fixes
- all other changes (added missing properties/...) will be removed
- there are some missing typehints for parameters ... pls don't worry, will be fixed in round 2 :)

---

Pull requests ...

- [x]  Admin
- [ ]  Adminhtml - classname designpack
- [x]  AdminNotification
- [x]  Api
- [x]  Api2
- [x]  Bundle
- [x]  Captcha
- [x]  Catalog
- [x]  CatalogIndex
- [x]  CatalogInventory
- [x]  CatalogRule
- [x]  CatalogSearch
- [x]  Checkout
- [x]  Cms
- [x]  ConfigurableSwatches
- [x]  Contacts
- [x]  Core
- [x]  Cron
- [x]  CurrencySymbol
- [x]  Customer
- [x]  Directory
- [x]  Downloadable
- [x]  Eav
- [x]  GiftMessage
- [x]  ImportExport
- [x]  Index
- [x]  Log
- [x]  Media
- [x]  Newsletter
- [x]  Oauth
- [x]  Page
- [x]  PageCache
- [x]  Payment
- [x]  Persistent
- [x]  Poll
- [x]  ProductAlert
- [x]  Rating
- [x]  Reports
- [x]  Review
- [ ]  Rss
- [x]  Rule
- [x]  Sales
- [x]  SalesRule
- [x]  Sendfriend
- [x]  Shipping
- [x]  Sitemap
- [x]  Tag
- [x]  Tax
- [ ]  Uploader (0)
- [x]  Weee
- [x]  Widget
- [x]  Wishlist

Changes in lib

- [x] Varien
- [ ] Zend

Exluded (at least in "round 1"):

- Authorizenet
- Backup
- Centinel
- Connect
- Dataflow
- GoogleAnalytics
- <s>GoogleBase</s> (removed, see #1002)
- GoogleCheckout
- Install
- Paygate
- Paypal
- PaypalUk
- Usa
- XmlConnect

colinmollenhour:
Nice work, @sreichel and thanks for tackling this big project! I'm fully on-board with all of the doc block-only updates I've seen so far. Thanks for splitting them up from the other updates so they can be reviewed separately.

[...]

PR #1029, opened 2020-06-08
Correct name for Request class in Payment Module

tmotyl:
There is already a class with name
Mage_Paygate_Model_Authorizenet_Request
in Paygate module.

Alternatively we should just remove the class as it was never used (autoload couldnt reach it).
What's your opinion on that?


sreichel:
I did not touch it in #725 for this reason. I'd prefer to remove it ....

tmotyl:
I've changed th PR to remove the class.

PR #1031, opened 2020-06-08
Ensure coupon code times_used decrements on cancel

### LINKED ISSUE #167: Canceled order - reuse coupon by the same customer
Hi.
Magento problem for ages. 
User is making order, using coupon - one time. Order is canceled due to any problems with payment. Very often.
User tries to reorder - and try to use the coupon again. No luck

Can somebody fix this issue (feature) in Magento that is really annoying. Takes time for custome support, they are asking stuff to help, stuff must regenerate coupons and send them manually.
The goal to allow reuse coupon that was used for canceled order. Nothing more. Key is canceled order state.
Any options to fix this?
Related Stack Exchange topic
http://magento.stackexchange.com/questions/125849/coupon-code-should-apply-once-we-cancel-the-order
Solution to check:
http://magebase.com/magento-tutorials/quick-fix-coupon-codes-used-up-on-incomplete-transactions OR [removed bad link]

Thanks
Tomek

mattdavenport:
To implement this cleanly in the observer (without instantiating new connection objects, fetching records, etc.) I added a parameter `$decrement` to `updateCustomerCouponTimesUsed` which should be a non-breaking change.

fixes #167

### DIFF
--- a/app/code/core/Mage/SalesRule/Model/Observer.php
+++ b/app/code/core/Mage/SalesRule/Model/Observer.php
@@ -133,6 +133,40 @@ class Mage_SalesRule_Model_Observer
     }
 
     /**
+     * Registered callback: called after an order payment is canceled
+     *
+     * @param $observer
+     */
+    public function sales_order_paymentCancel($observer)
+    {
+        /** @var Varien_Event $event */
+        $event = $observer->getEvent();
+        /** @var Mage_Sales_Model_Order $order */
+        $order = $event->getPayment()->getOrder();
+
+        if ($order->canCancel()) {
+            if ($code = $order->getCouponCode()) {
+                // Decrement coupon times_used
+                $coupon = Mage::getModel('salesrule/coupon')->loadByCode($code);
+                $coupon->setTimesUsed($coupon->getTimesUsed() - 1);
+                $coupon->save();
+
+
+                if ($customerId = $order->getCustomerId()) {
+                    // Decrement coupon_usage times_used
+                    Mage::getResourceModel('salesrule/coupon_usage')->updateCustomerCouponTimesUsed($customerId, $coupon->getId(), TRUE);
+
+                    // Decrement rule times_used
+                    if ($customerCoupon = Mage::getModel('salesrule/rule_customer')->loadByCustomerRule($customerId, $coupon->getId())) {
+                        $customerCoupon->setTimesUsed($customerCoupon->getTimesUsed() - 1);
+                        $customerCoupon->save();
+                    }
+                }
+            }
+        }
+    }
+
+    /**
      * Refresh sales coupons report statistics for last day
      *
      * @param Mage_Cron_Model_Schedule $schedule
--- a/app/code/core/Mage/SalesRule/Model/Observer.php
+++ b/app/code/core/Mage/SalesRule/Model/Observer.php
@@ -154,7 +154,7 @@ class Mage_SalesRule_Model_Observer
 
                 if ($customerId = $order->getCustomerId()) {
                     // Decrement coupon_usage times_used
-                    Mage::getResourceModel('salesrule/coupon_usage')->updateCustomerCouponTimesUsed($customerId, $coupon->getId(), TRUE);
+                    Mage::getResourceModel('salesrule/coupon_usage')->updateCustomerCouponTimesUsed($customerId, $coupon->getId(), true);
 
                     // Decrement rule times_used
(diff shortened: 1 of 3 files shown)


sreichel:
LGTM ... but can you please add some steps to reproduce?

mattdavenport:
Sure thing!

1. In `Shopping Cart Price Rules`, add a new rule with a "Specific Coupon" code with > 0 uses per coupon.
2. In `Actions` tab, set a discount amount.
3. Add a product to cart. Apply promo code in cart. Checkout.
4. Upon checkout, it can be observed that `times_used` has been incremented in tables:
    - `salesrule_coupon`
    - `salesrule_coupon_usage`
    - `salesrule_customer`
5. In admin panel, cancel the order.
6. It can be observed that `times_used` has decremented to allow the coupon to be used again.

colinmollenhour:
@tomekjordan @seansan may be interested in reviewing

app/code/core/Mage/SalesRule/Model/Observer.php
```diff
@@ -132,6 +132,40 @@ public function sales_order_afterPlace($observer)
...
+            if ($code = $order->getCouponCode()) {
+                // Decrement coupon times_used
+                $coupon = Mage::getModel('salesrule/coupon')->loadByCode($code);
+                $coupon->setTimesUsed($coupon->getTimesUsed() - 1);
+                $coupon->save();
+
+
+                if ($customerId = $order->getCustomerId()) {
+                    // Decrement coupon_usage times_used
+                    Mage::getResourceModel('salesrule/coupon_usage')->updateCustomerCouponTimesUsed($customerId, $coupon->getId(), TRUE);
```
colinmollenhour, line 157:
```suggestion
                    Mage::getResourceModel('salesrule/coupon_usage')->updateCustomerCouponTimesUsed($customerId, $coupon->getId(), true);
```

app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
```diff
@@ -46,11 +46,11 @@ protected function _construct()
     /**
      * Increment times_used counter
      *
-     *
      * @param unknown_type $customerId
      * @param unknown_type $couponId
+     * @param bool $decrement   Decrement instead of increment times_used
      */
-    public function updateCustomerCouponTimesUsed($customerId, $couponId)
+    public function updateCustomerCouponTimesUsed($customerId, $couponId, $decrement = FALSE)
```
colinmollenhour, line 53:
```suggestion
    public function updateCustomerCouponTimesUsed($customerId, $couponId, $decrement = false)
```

app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
```diff
@@ -60,11 +60,11 @@ public function updateCustomerCouponTimesUsed($customerId, $couponId)
 
         $timesUsed = $read->fetchOne($select, array(':coupon_id' => $couponId, ':customer_id' => $customerId));
 
-        if ($timesUsed > 0) {
+        if ($timesUsed !== FALSE && $timesUsed > 0) {
```
colinmollenhour, line 63:
```suggestion
        if ($timesUsed !== false && $timesUsed > 0) {
```

henrykbrzoska:
Hi, this logic is not obvious. Please reanalyse if used coupon should be reused.
IMO - shouldn't. Regardless of whether the order has been placed.

sreichel:
> Hi, this logic is not obvious. Please reanalyse if used coupon should be reused.
> IMO - shouldn't. Regardless of whether the order has been placed.

Can you please explain your concerns?

If I use a voucher but cancel the order, the voucher should still be available to me.

In what cases should the voucher no longer be valid?

henrykbrzoska:
> If I use a voucher but cancel the order, the voucher should still be available to me.

are you sure?

sreichel:
> are you sure?
> I think only about ecommerce logic. If i give someone 'ONE time' coupon it should be 'ONE time' usable.

Yeah, but maybe I'm missing something...

For example: A customer orders an article not in stock with a one-time-voucher. Since it takes too long, he cancels the order.

Should the voucher expire?

henrykbrzoska:
IMO yes. he should get another one if seller permit.

sreichel:
Well ... IMO it should not :)

Just did a quick search ...

- mediamarkt/saturn voucher expires
- adidas does not expire, even after full refund

I think its easy cover both needs ... add a config section and a check to observer. Time for a PR?

henrykbrzoska:
ok, approve. :)

I wanted to be careful

PR #1032, opened 2020-06-09
Allow access to current order

woutersamaey:
It's a pain in Magento to get the order currently being created. This would be the cleanest solution, even though you could build on the `checkout_type_onepage_save_order` event, but I don't like that. You'd also need to make sure the event is always first and it created clutter.

Magento uses the registry and `current_` prefix for everything, so why not here?

I understand that for Magento 2 this would be a big no-no, but for Magento 1 I think this is perfectly acceptable and would make a lot of people happy.


tmotyl:
Will it work if there are few orders created in a batch?

woutersamaey:
Can we get this approved?

tmotyl:
Funny, today I have a case In a project where this change ould be helpful - to avoid having to fetch order again from db.

PR #1034, opened 2020-06-09
Fixed store get attribute raw value

midlan:
My collegue found another bug in method getAttributeRawValue. It retured values for other attributes when passed array of attributes to second parameter `$attribute` and store value does not exist for some of the attributes.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
@@ -660,7 +660,7 @@ abstract class Mage_Catalog_Model_Resource_Abstract extends Mage_Eav_Model_Entit
                         'store_value.value'
                     );
                     $joinCondition = array(
-                        $adapter->quoteInto('store_value.attribute_id = default_value.attribute_id', array_keys($_attributes)),
+                        'store_value.attribute_id = default_value.attribute_id',
                         'store_value.entity_type_id = :entity_type_id',
                         'store_value.entity_id = :entity_id',
                         'store_value.store_id = :store_id',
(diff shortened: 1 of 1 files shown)


tmotyl:
Can you describe steps how can I test it? (even if it requires an xdebug session)

kiatng:
I tested with

```php
$result = Mage::getResourceSingleton('catalog/product')->getAttributeRawValue(
    1,
    [74,99,98],
    0
);
```

The result is the same for the existing code and the modified code:

```
 array(3) {
  ["description"] => string(38) "Product Description"
  ["url_key"] => string(31) "product-description"
  ["status"] => string(1) "2"
}
```

tmotyl:
Thanks @kiatng. I would also like to tests the case which was broken before and now is fixed.

kiatng:
I managed to replicate the bug. As per description:

>  ... and store value does not exist for some of the attributes.

Test code:

```php
$result =$result = Mage::getResourceSingleton('catalog/product')->getAttributeRawValue(
    1,
    [74,99,98,'custom_layout_update'],
    2
);
```
Test result with code in the PR:

```
 array(4) {
  ["description"] => string(19) "Product Description"
  ["custom_layout_update"] => string(22) "<!-- custom design -->"
  ["url_key"] => string(19) "product-description"
  ["status"] => string(1) "2"
}
```

Test result with core code:
```
 array(4) {
  ["description"] => string(22) "<!-- custom design -->"
  ["custom_layout_update"] => string(22) "<!-- custom design -->"
  ["url_key"] => string(19) "product-description"
  ["status"] => string(1) "2"
}
```

The PR is correct.

midlan:
@kiatng yes thats the usage which is buggy, with different store than default 0, I am talking about

PR #1043, opened 2020-06-15
Fix Warning: filemtime(): stat failed for /js/prototype/windows/themes/magento.css

chowaniec:


colinmollenhour:
I think this will be removed in #1040. Please comment there if you know of any reasons not to remove the XmlConnect package.

chowaniec:
@colinmollenhour thanks, I didn't notice it.

PR #1044, opened 2020-06-16
Removed Phoenix_Moneybookers

sreichel:
Ref #988

Should we keep it? If required ... https://github.com/simkea/Phoenix_Moneybookers ...

TODO: setup script to delete config entries (seperate PR)


Flyingmana:
I think Iam missing the reason why this module should be removed.
isnt it part of the core for a very long time now and probably used by many?

Besides, the linked Repository has an outdated version, the marketplace already had newer versions.
https://github.com/OpenMageModuleFostering/phoenix_moneybookers/commits/master

colinmollenhour:
It can always be added back as a separate extension if anyone is actually using it but I wouldn't assume that because it exists it is being used, necessarily.. I think a more important consideration is do we want to support it? If it isn't compatible with SAQ A I think we should remove it. OpenMage is "use at your own risk" anyway, but it is still good to strongly discourage use of payment methods that are not SAQ A by not including them in core, IMHO.

sreichel:
Imho it is just shipped with core, but a community extension that has not been updated for years. Is this still maintained - after M1 EOL? There is no public repo to track or report issues, so we have to support it on our own. Do we want this?

- https://github.com/magento/magento2/issues/103
- https://www.quora.com/Why-nobody-uses-Moneybookers-now-called-Skrill-anymore-Whats-wrong-with-this-payment-system

composer.json
```diff
@@ -2,26 +2,48 @@
...
+    "squizlabs/php_codesniffer": "*"
+  },
+  "extra": {
+    "magento-root-dir": "."
+  },
+  "config": {
+    "sort-packages": true
+  },
+  "scripts": {
+    "install-hooks": ["sh ./vendor/smgladkovskiy/phpcs-git-pre-commit/src/setup.sh"],
```
Flyingmana, line 39: Does not belong to the PullRequest

Flyingmana:
I see, as it belongs to a service which is not existing anymore, removing is ok.

sreichel:
> 
> 
> +1 for removing

Ther is a aprrove-button .... ;)

viable-hartman:
I know a little to this conversation, but wouldn't we also want to delete: **skin/frontend/base/default/images/moneybookers** and **skin/adminhtml/default/default/images/moneybookers** when getting rid of this extension?

sreichel:
> I know a little to this conversation, but wouldn't we also want to delete: **skin/frontend/base/default/images/moneybookers** and **skin/adminhtml/default/default/images/moneybookers** when getting rid of this extension?

Yep. There is already a PR ... https://github.com/OpenMage/magento-lts/pull/1161

PR #1047, opened 2020-06-18
Array and string offset access syntax with curly braces

sreichel:
Fixed: "PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated" for php7.4

Ref #829, #859 (@kkrieger85)

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Increment/Alphanum.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Increment/Alphanum.php
@@ -77,7 +77,7 @@ class Mage_Eav_Model_Entity_Increment_Alphanum extends Mage_Eav_Model_Entity_Inc
                 $p = 0;
                 $bumpNextChar = true;
             }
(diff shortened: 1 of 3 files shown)


rvelhote:
There are also a bunch of equivalent deprecation warnings for ` Array and string offset access syntax with curly braces is deprecated` under `lib/Zend` however the static code analysis for 7.4 ignores this folder. Is this being done on purpose or it's an oversight?

```
find . -not \( -path ./.phpstorm.meta.php -prune \) -not \( -path ./lib/PEAR -prune \) -not \( -path ./lib/phpseclib -prune \) -not \( -path ./lib/Zend -prune \) -type f -name "*.php" -exec php -d error_reporting=32767 -l {} \; 2>&1 1> /dev/null | grep "^"
```

sreichel:
@rvelhote good catch. Libs should be included too.

sreichel:
Please merge #1058 instead. @rvelhote  please add your comment there.

PR #1048, opened 2020-06-18
Fix broken file upload for downloadables caused by PATCH SUPEE-11314

theroch:
Removed session params from upload URL to allow file uploads for downloadable products.

Signed-off-by: Frank Rochlitzer <f.rochlitzer@b3-it.de>


theroch:
Caused by app/code/core/Mage/Admin/Model/Session.php:109->127
```
/**
 * Logout user if was logged not from admin
 */
protected function logoutIndirect()
{
    $user = $this->getUser();
    if ($user) {
        $extraData = $user->getExtra();
        if (
            !is_null(Mage::app()->getRequest()->getParam('SID'))
            && !$this->allowAdminSid()
            || isset($extraData['indirect_login'])
            && $this->getIndirectLogin()
        ) {
            $this->unsetData('user');
            $this->setIndirectLogin(false);
        }
    }
}
```
`!$this->allowAdminSid()` is always true.

The Links.php and Samples.php added the SID param to all upload URLs. So it was no more possible to upload any files for downloads.

sprankhub:
I can confirm that this fixes the issue.

PR #336, opened 2017-08-29
Added functionality for $after paramenter in addCrumbs() method

sreichel:
In `Mage_Page_Block_Html_Breadcrumbs::addCrumb()` third parameter `$after` is ignored/has no affect ...

This should work correctly now ...

     $block->addCrumb('my_crumb', array(
         'label'    => label,
         'title'    => title,
     ), 'home');

PR #1051, opened 2020-06-20
Add features from N98 layout helper

sreichel:
See #321, #336 (netz98/N98_LayoutHelper#10)
Ref #416

N98_LayoutHelper is one of the modules that i have used everywhere ... most important function are already in LTS core code (remove links, improved breadcrumbs), only one usefull change at Page_Head is left ... 

This PR adds two args to `Mage_Page_Block_Html_Head::addItem()` mehtod ...

```
     * @param string $referenceName name of the item to insert the element before. If name is not found, insert at the end, * has special meaning (before all / before all)
     * @param bool $before If true insert before the $referenceName instead of after
```

(todo ... change composer.json to prevent install of N98 module, add documentation)


rjocoleman:
@sreichel do you want a hand with this?

I can add:

* [`replace`](https://getcomposer.org/doc/04-schema.md#replace) in composer.json:
```json
{
  "replace": {
    "n98/n98_layouthelper": "*"
  }
}
```

* Some docs, as this feature is super useful but quite hard to understand, so docs are important - where should they go?

sreichel:
> 
> 
> @sreichel do you want a hand with this?
> 
> I can add:
> 
>     * [`replace`](https://getcomposer.org/doc/04-schema.md#replace) in composer.json:
> 
> 
> ```json
> {
>   "replace": {
>     "n98/n98_layouthelper": "*"
>   }
> }
> ```
> 
>     * Some docs, as this feature is super useful but quite hard to understand, so docs are important - where should they go?

Thanks. @Flyingmana do you have any objections regarding "replace"? I only marked it as a draft because there are many undocumented but important changes that you might want to know about if you switch to OpenMage.

kkrieger85 (approved):

colinmollenhour (approved):

PR #1061, opened 2020-06-24
Fix bug on filter issues in Role Users.

eetzen:
Unable to filter Role Users in Permission->Roles->Role Users or go to next Page or change the number of rows displayed per page.

This bug is introduced in pull request https://github.com/OpenMage/magento-lts/commit/d644768c998cf37c0f633482e8765e4e630457cc

Mage_Adminhtml_Block_Permissions_Role_Grid_User->_getUsers() is returning an int instead of an array on line 160.


AlterWeb:
Looks good! I don't see any problem with the solution. I tested it and it works.

theroch:
I can confirm this bug.
But your description is a little bit confusing.
The filter in the grid works for me, but the first column information (in_role_users) is lost while you set any filter.

The type casting was implemented with PATCH SUPEE-11155 CE 1.9.4.0 v3-2019-06-18-08-34-01.
So this change is security related!

Maybe you make a type cast to array?

Issue #925, opened 2020-04-13
poor performance when building simplexml object for configuration


onlinebizsoft:
I have a Magento site which has quite many store (around 100 website/store). I'm trying to achieve a performance issue with Magento and I found that PHP (7.x)  has a very poor performance when building the simplexml object from cached XML string
![image](https://user-images.githubusercontent.com/22501570/79102962-f8715f00-7d95-11ea-82bb-7f3fc8dfc790.png)

I was trying to find a way to cache this object directly however simplexml object has no serialize interface, even I tried to use this library but it doesn't support as well
https://github.com/igbinary/igbinary

Has anyone gotten any idea?
@colinmollenhour @Vinai

tmotyl:
Hi
@onlinebizsoft please give more context.
Why do you think this method is a bottleneck in your case? What did you profile and how?
What cache storage do you use?

How many times this method is called?
From where?

I know indeed that simplexml is not very performant but in order to do any optimization the case you have need to be described much better.

colinmollenhour:
I've thought about improvements for SimpleXML many times but I think the only way to improve it would be to replace it (e.g. with native PHP arrays) but that would be a lot of effort and break a lot of code so I don't really see it being feasible.

However, unless I've forgotten something once the config is cached, increasing the number of stores should not really hurt performance since the cache is broken down into different "sections" for each website/store and only the ones used in a request should be loaded, so it should only be loading one store per request.

rvelhote:
I was looking into something recently that's slightly related.

Magento does not do `simple_xml_load_string` unless you try to load a configuration for a specific store view but it does [fetch all the configurations from cache](https://github.com/OpenMage/magento-lts/blob/c6f200cb6bc3c4ace0d9b54d0d0df0ad0432ba8f/app/code/core/Mage/Core/Model/App.php#L659). This results, in our case, in 98 calls to Redis.

It's not a huge performance cost for Redis but I can imagine that people that have the cache in the database and lots of storeviews might suffer a bit.

colinmollenhour:
Ahh, nice find! I had never noticed that code before and you are absolutely right, it results in a cache load/save for every store on every request! It only has a handful of keys so they are small payloads but still that is pretty significant for large numbers of stores. It seems that it would be much better to load/save all stores as one cache record since all of them are always loaded anyway.

PR #1066, opened 2020-06-25
Create a global config cache rather than a config cache per store

rvelhote:
Related to: https://github.com/OpenMage/magento-lts/issues/925

Magento stores a small configuration cache with a limited number of parameters where each cache key is scoped to a store. For shops with a lot of store view this will generate n-queries (where n is the number of stores) to the cache backend.

This patch will add a initConfigCache method to the collection level rather than applying it per store. It will loop through all the stores and assign the configuration cache to each store therefore making just one query to the cache backend.

This is an draft PR to get feedback. It works well in my test environment reducing the number of requests to Redis from 99 to 1.

### DIFF
--- a/app/code/core/Mage/Core/Model/App.php
+++ b/app/code/core/Mage/Core/Model/App.php
@@ -655,9 +655,11 @@ class Mage_Core_Model_App
         $websiteGroups = array();
         $groupStores   = array();
 
+        $storeCollection->initConfigCache();
+
         foreach ($storeCollection as $store) {
             /** @var Mage_Core_Model_Store $store */
-            $store->initConfigCache();
+            // $store->initConfigCache();
             $store->setWebsite($websiteCollection->getItemById($store->getWebsiteId()));
             $store->setGroup($groupCollection->getItemById($store->getGroupId()));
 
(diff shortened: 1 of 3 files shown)


colinmollenhour:
Awesome! What is the real difference this makes in a production environment?

rvelhote:
The current time spent loading these entries from Redis is about 10ms for 99 store views so I expect to shave at least that amount from the current page load. Perhaps even more in the future as we add more store views and even move to Elasticache (higher latency).

I think this patch/change will not be very useful if you don't have that many store views however it's here for everyone's consideration.

app/code/core/Mage/Core/Model/App.php
```diff
@@ -655,9 +655,11 @@ protected function _initStores()
         $websiteGroups = array();
         $groupStores   = array();
 
+        $storeCollection->initConfigCache();
+
         foreach ($storeCollection as $store) {
             /** @var Mage_Core_Model_Store $store */
-            $store->initConfigCache();
+            // $store->initConfigCache();
```
colinmollenhour, line 662: Delete commented out code
```suggestion

```

luigifab (commented):
We are using the PR since some weeks with our 20 stores views.
Not sure if the PR change something or not.

luigifab (approved):

tmotyl:
rebased, to see if it's still green

PR #1073, opened 2020-06-29
(BC) Removed IE compatibility

luigifab:
### Description

I suggest to kill **IE 5 / 6 / 7 / 8 / 9** compatibility.  
I suggest also to remove the `if` tag for `addItem` of `Mage_Page_Block_Html_Head`.

I don't suggest to maintain compatibility only with 3 latest recent versions of all major browsers.

10,840 deletions, I like that, and I hate IE.

### Contribution checklist


colinmollenhour:
Oops, I clicked "Ready for review" accidentally... Should this still be in draft status for some reason?

sreichel:
> I suggest also to kill the if tag for addItem (not yet done).

?

luigifab:
It's ready (I hope).

kkrieger85:
@luigifab could you please rebase/fix conflicting file?

luigifab:
So sorry for my bad commits some days ago (with xmlconnect changes).
Here is a new proposal with more deletions.

I removed `decorateGeneric` `decorateTable` `decorateList` `decorateDataList` JS functions.
This will probably break some CSS based on `.odd .even .first .last` classes, not yet investigated/fixed.

colinmollenhour:
> I removed `decorateGeneric` `decorateTable` `decorateList` `decorateDataList` JS functions.
> This will probably break some CSS based on `.odd .even .first .last` classes, not yet investigated/fixed.

Let's not do this, it will break BC and honestly I don't think there is much performance impact from using them even though it would not be considered good practice these days.

Flyingmana:
The decorate table function has a notable impact, as it can force a redraw of many page parts, after this new classes are applied.

But Iam absolutely against removing it. Its not only IE what is affected here, but all browsers, as the themes uses this classes and likely also many css and js people wrote over the years.
This function is not something like a polyfill, it adds new elements which are used, and removing this will for sure break many things.
But for new themes, js and css we should discourage making use of this party.

luigifab:
Should i change the base branch to 20.0?

colinmollenhour:
Nice work, @luigifab!

> Should I change the base branch to 20.0?

I think it could go either way but I'll defer to @Flyingmana.

kkrieger85 (approved):

Flyingmana:
yeah, definitive 20.0, I sadly know projects on M1 who are still relying on IE internally. The sad world of big corp B2B

luigifab:
So let's go for 20.0.

colinmollenhour (approved):
Excellent work @luigifab !

fballiano (approved):

fballiano:
I'd love to see this merged

kkrieger85 (approved):

Flyingmana:
small warning in advance: this PR will get at least partially reverted soon, as it breaks some usecases involving the `<action method="addItem">` where we do not have a replacement for. (Issue follows soon)

ADDISON74:
We may only have reports if it is used as an Internet Explorer browser. At this time it is official that IE 11 will be removed by June 15, 2022 from all Windows distributions. If I look at the statistics, those who still use IE are very few v11 mainly and a few v9. The market is already divided between Chrome, Edge, Firefox, Brave, Vivaldi, Opera, Safari.

But I agree with @Flyingmana if it is merged then it needs to be tested more so as not to create any issues.

Flyingmana:
the broken usecase is unrelated to Internet Explorer

ADDISON74:
Then there really is a problem with this PR and it needs to be reviewed.

ADDISON74:
On first reading of the code almost all <action method = "addItem"> removed refer to IE related files. @flyingmana could you please indicate where you identified that problems could occur?

js/varien/js.js
```diff
@@ -649,18 +649,6 @@ Element.addMethods({
...
-
-    window.console = {};
-    for (var i = 0; i < names.length; ++i)
-        window.console[names[i]] = function() {}
-}
-*/
-
 /**
  * Executes event handler on the element. Works with event handlers attached by Prototype,
  * in a browser-agnostic fashion.
```
ADDISON74, line 654: May I ask why this if statement was removed from code? What was the purpose?

joshua-bn: Are you asking why it was removed in this PR? It was commented out by Magento.

If you're asking why they commented it out... 🤷

ADDISON74: In magento there are certain parts commented in the code but not because they would not be functional but because they served some purposes. I asked why this part of the code was removed by the PR author since I see no connection to IE.

joshua-bn: `!("console" in window)` was for IE 6. It didn't have console.log (crazy, but true... that's how old it is)

joshua-bn: https://caniuse.com/console-basic

ADDISON74: Thank you for your clarification. In this case personally I no longer have any objection related to PR.

Danieleeffe1:
Hello

I read a little about the reasons for these changes.

With reference to the bug reported # 1796 I would like to specify that with no browser it loads the css file of the customized theme

the custom theme is a direct descendant of RWD

apart from the css file and a few modifications of some files the classes have not been touched in any way.

the result is that themes other than rwd are not loaded.

it's a huge problem

ps: the problem recurs also with the 20.0.13 version of openmage

fballiano:
I think that if you need that level of compatibility you should stay on v19 instead of v20.

Danieleeffe1:
> Penso che se hai bisogno di quel livello di compatibilità dovresti rimanere su v19 anziché v20.

agree
evolve a project.

but don't you risk losing that initial idea of ​​backwards compatibility to avoid losing years of work and energy poured into Magento 1?

What real (performance) advantage have you obtained by making these changes that are causing problems for web designers?

fballiano:
backward compatibility is always assured by the v19 branch, which is still maintained as promised ;-)

the benefits is that we can have a maintained code, not thousands of files that none of the contributors know/use anymore that just lay there forever, regarding super old technologies that tie us back to the past.

to switch to the v19 it depends how you installed it in the first place, if you copied the files like it was usually done, then simply overwrite everything with the new/old version, it's not gonna be super clean but still..

Danieleeffe1:
Thanks for the reply

my situation is this:

Ubuntu server 20.04
php 7.4
mysql 8

OpenMage 20.0.8 installed via composer and currently updated via composer to 20.0.12.

so: how should I do?

consider me as a child to accompany by the hand.

Thank you

fballiano:
if you're doing it with composer simply rebuild the composer.json file, delete the composer.lock and the vendor folder and the "composer install"

PR #1074, opened 2020-06-29
Added configuration for captcha failed_attempts_ip in backend

### LINKED ISSUE #1065: Admin Captcha (After number of attempts to login) has a default value of 1000 tries per IP and no way to change this
### Description (*)

The Captcha for the admin login can be set to a mode "After number of attempts to login" you then have the options to specify after how many attempts this will be required. Standard this is 3 attempts, so I thought this would mean that after I tried 3 wrong credentials I would get the Captcha. But after some testing I found out this is only applied to the attempts to login to the same account. You can hover try multiple accounts without triggering this Captcha. Not a big security risk because you already should know the password and only guessing the account name but still I think it would be nice to block a bot from guessing more than the 3 combinations of users and passwords.

There seems to be a mechanism however that does that based on IP address. This value is set to 1000 attempts however, which I find quit a lot. There is no admin setting to override this value so I can only change this by overriding this in the core config files, directly in the database or by creating a custom module for that.

### Expected behavior (*)

I see several ways to make this feature more useful.

1. Make a admin config setting for this so every shop owner can change this for them self.
2. Change this behavior to make use of the same value already set in the backend (now used for the number of attempts to login on a single account).
3. Lower this number (1000) to a more appropriate number like for instance 10. If someone is providing more then 10 different combinations I would say there is something wrong there. 

### Benefits

Improve security (slightly) and make the feature behave more like expected (or at least what I was expecting it to behave like).

### Additional information

If more people find this useful I am happy to create a pull request to add this feature. Please also let me know then which option you prefer and why or if you see a problem with one of the options.

damien-biasotto:
Made an admin config so every shop owner can set their own limit.

It will only enforce the use of the captcha, and will not lock out an account.

It should fix #1065

### DIFF
--- a/app/code/core/Mage/Captcha/etc/system.xml
+++ b/app/code/core/Mage/Captcha/etc/system.xml
@@ -77,9 +77,22 @@
                             </depends>
                             <frontend_class>required-entry validate-digits</frontend_class>
                         </failed_attempts_login>
+                        <failed_attempts_ip translate="label">
+                            <label>Number of Unsuccessful Attempts to Login per IP Address</label>
+                            <frontend_type>text</frontend_type>
+                            <sort_order>6</sort_order>
+                            <show_in_default>1</show_in_default>
+                            <show_in_website>0</show_in_website>
+                            <show_in_store>0</show_in_store>
+                            <depends>
+                                <mode>after_fail</mode>
+                                <enable>1</enable>
+                            </depends>
+                            <frontend_class>required-entry validate-digits</frontend_class>
+                        </failed_attempts_ip>
                         <timeout translate="label">
                             <label>CAPTCHA Timeout (minutes)</label>
-                            <sort_order>6</sort_order>
+                            <sort_order>7</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
@@ -89,7 +102,7 @@
                         <length translate="label comment">
                             <label>Number of Symbols</label>
                             <comment>Please specify 8 symbols at the most. Range allowed (e.g. 3-5)</comment>
-                            <sort_order>7</sort_order>
+                            <sort_order>8</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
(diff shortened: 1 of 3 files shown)


kiatng:
What's the effect if the default value in `config.xml` is changed instead of the upgrade script?

```xml
    <admin>
        <captcha>
            ...
            <failed_attempts_ip>10</failed_attempts_ip>
```

damien-biasotto:
> What's the effect if the default value in `config.xml` is changed instead of the upgrade script?
> 
> ```
> <default>
>     <admin>
>         <captcha>
>             ...
>             <failed_attempts_ip>10</failed_attempts_ip>
> ```
That's a good question, the update script makes this setting customisable in the admin area which was not the case before. I probably shouldn't have to change the value in the `config.xml` but I like consistency :)

kiatng:
I tested by changing the `<default>` in `config.xml` without the upgrqade script, and can see that the default is set to 10, also, I am able to set the value to 8:

image

I think changing the value in `config.xml` is a better way, it's unnecessary to bump the ver, which may have unintended consequences. As a bad example, Adobe decides to prolong EOL, the ver may conflict with the official ver.

damien-biasotto:
@kiatng Yup, I guess you're right :) I just wanted to create the config path in `core_config_data` table, but it will be done as soon as the store owner is saving the settings in admin. I will update the PR.

sreichel:
Translatable string should be added to csv?

damien-biasotto:
> Translatable string should be added to csv?

Yup thank you for the feedback. I added it.

AlterWeb:
Looks good to me. I would however change the label for the setting "admin/captcha/failed_attempts_login" to "Number of Unsuccessful Attempts to Login per Account" to make it more clear what the difference is.

sreichel (dismissed):
> Please also change the default in `app/code/core/Mage/Captcha/etc/config.xml`.

I'd keep default value and add documantion later. There are a few PRs with added config option, but none of them is mentioned anywhere.

app/code/core/Mage/Captcha/etc/system.xml
```diff
@@ -90,10 +90,23 @@
                             </depends>
                             <frontend_class>required-entry validate-digits</frontend_class>
                         </failed_attempts_login>
+                        <failed_attempts_ip translate="label comment">
```
luigifab, line 93: There is no comment, can you remove it?

ADDISON74:
@damien-biasotto - Could you please check the reviews from @colinmollenhour and @luigifab to get ready this PR for approving?

app/code/core/Mage/Captcha/etc/system.xml
fballiano: done

sreichel:
> but most of all I can't see the PHP logic

Code is already there, but there was no config section for it.

fballiano (approved):
I've rebased and rechecked this PR, why did we leave it dormant? I think it works and should be merged

ADDISON74 (approved):

sreichel:
Memo ... should be added to docs.

luigifab:
This is not important, but `<frontend_type>text</frontend_type>` was added, and it's useless, and removed by another PR.

fballiano:
@luigifab I made a super exception and removed that line directly, since you were totally right

PR #1075, opened 2020-06-30
memory leak when resizing multiple product images

ioweb-gr:
I've stumbled into this issue a while back and noted this from somewhere on the internet. Unfortunately I don't have the link anymore. 

The issue is that when you loop through products and resize their main image you will at some point run out of memory. It seems the GD adapter is not freeing up memory correctly with it's custom shutdown function. 

In my sites I used to override the adapter locally and it worked OK so I'm sharing the code I found. I'm not sure if you want to merge it though as it may interfere in other areas. 

It will be good to have it noted down here anyways as a record that this problem exists and has a solution. Credits go to unknown person that wrote this down


colinmollenhour (dismissed):

colinmollenhour:
Thanks for the PR!

app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
```diff
@@ -118,7 +118,7 @@ public function execute()
 
         /** @var Mage_Core_Model_Date $coreDate */
         $coreDate  = $this->_factory->getModel('core/date');
-        $timestamp = $coreDate->gmtTimestamp('Today');
+        $timestamp = Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP);
```
tmotyl, line 121: should this change be included in the PR?

ioweb-gr: Oh sorry this was an accident on my side because I was doing something else. Didn't know this wasn't merged in so long ⭕

Let me reset the branch for 1.9.4.x

ADDISON74:
Any other opinions to move this forward?

fballiano (commented):
I like it

fballiano (approved):

kiatng:
In this [stackoverflow](https://magento.stackexchange.com/questions/79672/importexport-problem-with-new-destructor-of-varien-image-adapter-gd2-in-1-9-2-0), there is a discussion about this issue. Someone commented about security concern with the `__destruct()` method.

ADDISON74:
Interesting comments on Stackoverflow. The concern about security is coming from Magento Senior Piotr Kaminski. I would prefer the last solution offered in that post by logging the error too. It is implemented here:

https://github.com/sitewards/import_image_memory_leak_fix

colinmollenhour:
Yeah I'm failing to see how this code could pose a security risk.. especially with "allowed classes".

fballiano:
fixed conflicts but we have to decide what to do with this, either we merge it or we close it.

ADDISON74:
It seems to be an issue in OM code that needs to be fixed, but I don't know how to reproduce it. 

I prefer David Manners's solution. You may find it here, the last post on the page https://magento.stackexchange.com/questions/79672/importexport-problem-with-new-destructor-of-varien-image-adapter-gd2-in-1-9-2-0.

There is another solution here https://github.com/borasocom-team/magento-gd2-memoryleak (I did not check it).

I would take in consideration what Piotr Kaminski said related to magic methods like __destruct().

ADDISON74:
On the other hand, I think that we should also deal with the errors that could appear when using the destructor. For this reason I considered that the proposal made by one of the participants in the discussion on Stack Exchange seems more appropriate to me.

colinmollenhour (approved):
Unless `imagedestroy` has some sort of explotable low level memory management issue I don't see how this is potentially harmful and think it should be merged. This has nothing to do with object serialzation/deserialization except that `__destruct` is a magic method and so are `__sleep`  and `__wakeup`.

colinmollenhour:
Looks like as of PHP 8.0 this function is now a no-op anyway so we're just wasting time discussing.. Suggest to close.

> This function has no effect. Prior to PHP 8.0.0, this function was used to close the resource.

fballiano:
true, since 8.0 imagedestroy is NOP, so it does nothing. we should close this PR.

PR #1076, opened 2020-06-30
Wexo M1 Fixes

wexo-team:
In this pull request there are two fixes which improves the stability and security on the Magento 1 shops we manage.

The "preventBadControllerLogin" fix prevents posting of login requests to controllers that is not an instance of "Mage_Adminhtml_IndexController" which is the default login controller.
This helps to stop brute force attacks on controllers that has been configured wrong in the config.xml.
 
The fix in "app/code/core/Mage/Cms/Model/Wysiwyg/Images/Storage.php" is to prevent the the GD2 library from throwing an exception when trying to resize to 0 width or height.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Observer.php
+++ b/app/code/core/Mage/Adminhtml/Model/Observer.php
@@ -76,4 +76,41 @@ class Mage_Adminhtml_Model_Observer
         Mage::app()->removeCache(Mage_Adminhtml_Block_Notification_Security::VERIFICATION_RESULT_CACHE_KEY);
         return $this;
     }
+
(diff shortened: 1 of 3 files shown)


tmotyl:
Hi @wexo-team 
Thanks a lot for your contribution! Highly appreciated.
Can you please split the 2 chages into 2 PR's? This will allow us to review and test them separately.
Thnaks.

wexo-team:
@tmotyl i have now split the Pull Request into https://github.com/OpenMage/magento-lts/pull/1077 and https://github.com/OpenMage/magento-lts/pull/1078

PR #1077, opened 2020-06-30
Prevent bad controller login.

wexo-team:
Splitting previous Pull Request into 2 as asked by @tmotyl in https://github.com/OpenMage/magento-lts/pull/1076

The observer added makes admin controllers that is configured wrong in external modules, return 403 if it is accessed directly without being logged in.
This it to prevent posting directly to these controllers for admin login.

### DIFF
--- a/app/code/core/Mage/Admin/Model/Observer.php
+++ b/app/code/core/Mage/Admin/Model/Observer.php
@@ -100,11 +100,19 @@ class Mage_Admin_Model_Observer
                             ->setActionName('deniedJson')
                             ->setDispatched(false);
                     } else {
-                        $request->setParam('forwarded', true)
-                            ->setRouteName('adminhtml')
-                            ->setControllerName('index')
-                            ->setActionName('login')
-                            ->setDispatched(false);
+                        if ($request->getRouteName() == 'adminhtml') {
+                            $request->setParam('forwarded', true)
+                                ->setRouteName('adminhtml')
+                                ->setControllerName('index')
+                                ->setActionName('login')
+                                ->setDispatched(false);
+                        } else {
+                            $request->setParam('forwarded', true)
+                                ->setRouteName('adminhtml')
+                                ->setControllerName('index')
+                                ->setActionName('forbidden')
+                                ->setDispatched(false);
+                        }
                     }
                     return false;
                 }
--- a/app/code/core/Mage/Adminhtml/Controller/Action.php
+++ b/app/code/core/Mage/Adminhtml/Controller/Action.php
@@ -280,6 +280,13 @@ class Mage_Adminhtml_Controller_Action extends Mage_Core_Controller_Varien_Actio
         $this->renderLayout();
     }
 
+    public function forbiddenAction()
+    {
+        $this->getResponse()->setHeader('HTTP/1.1','403 Forbidden');
+        $this->loadLayout(array('adminhtml_denied'));
+        $this->renderLayout();
+    }
+
     public function loadLayout($ids=null, $generateBlocks=true, $generateXml=true)
     {
         parent::loadLayout($ids, $generateBlocks, $generateXml);
--- a/app/code/core/Mage/Adminhtml/Model/Observer.php
+++ b/app/code/core/Mage/Adminhtml/Model/Observer.php
@@ -76,4 +76,41 @@ class Mage_Adminhtml_Model_Observer
         Mage::app()->removeCache(Mage_Adminhtml_Block_Notification_Security::VERIFICATION_RESULT_CACHE_KEY);
         return $this;
     }
+
+    /**
+     * Prevent login from controllers that is not Mage_Adminhtml_IndexController
+     *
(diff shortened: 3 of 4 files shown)


tmotyl:
Hi
Can you provide an example of a bad controller this change is securing?
I'm wondering whether it will be better to change Mage_Adminhtml_Controller_Action class instead of adding an observer in global scope (which would slow down every request to Magento).

If the root cause is wrong configuration, then maybe we can implement a configuration validation (or even on-the-fly migration of the wrong configuration)?

wexo-team:
HI @tmotyl,

We have found that admin controllers that is registered in the config.xml like shown below, would just show a login screen if accessed without being logged in. The controller would still need to extend `Mage_Adminhtml_Controller_Action`, so it's entirely possible to just add a version of the code to the `preDispatch` method of the `Mage_Adminhtml_Controller_Action` class, if this is prefered, then i can make changes to this Pull Request and test it to make sure it has similar effect. I don't believe the configuration is wrong, but it is just the way it defaults back to the login screen on an URL that does not fall under the backend URL, is what i would classify unwanted behavior.
```xml
<?xml version="1.0"?>
<config>
    <modules>
        <Vendor_Module>
            <version>1.0.0</version>
        </Vendor_Module>
    </modules>
    <admin>
        <routers>
            <route_name>
                <use>admin</use>
                <args>
                    <module>Vendor_Module</module>
                    <frontName>route_name</frontName>
                </args>
            </route_name>
        </routers>
    </admin>
</config>
```

tmotyl:
Hi
Yes, I would prefer a fix which doesn't add an observer but fixes directly in a proper place.

What is the problem here? Is it exposing login form outside configured custom admin url?
Or is there more?
Why is magento displaying a login form (so forward) instead of redirecting? Which code is responsible for that?


Problems like that have to be carefully fixed not to break stuff, thats why it's so important to have a good understanding of the problem and solutions.

wexo-team:
HI @tmotyl ,

After a lengthy debugging and testing session, we have now made a fix that fixes the issue at the exact point where the Internal Forward is happening.

We have also made sure that this does not cause an issue for users where the session times out on controllers which does not have this problem.
Users will only get the 403 if their session times out on a controller which has this issue, or if they access the controller directly without being logged in.

The problem is as you say that it exposes an login form from an controller which is not under the backend URL.
The problem is because of the internal forward that is right here: [app/code/core/Mage/Admin/Model/Observer.php:103](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Admin/Model/Observer.php#L103)
That internal forward should only apply as the new fix specifies, if the route used is the "adminhtml" route, so that if your session times out on the order grid, you will get the login page.

colinmollenhour:
Please let me try to summarize and tell me if I'm correct or wrong:

If accessing admin controller that requires auth from a frontend base url: user will now get 403 error.
Before: user got an admin login page?

seansan:
> Please let me try to summarize and tell me if I'm correct or wrong:
> 
> If accessing admin controller that requires auth from a frontend base url: user will now get 403 error.
> Before: user got an admin login page?

This is what I understand too from reading. 

I think the win here is : it does not expose the admin uri  (and hence a major win)

Is this assumption correct?

sicet7:
@seansan Yes your assumption is correct, the fix is simply to prevent exposing a login page on a url which is not under the configured admin url :-)

app/code/core/Mage/Admin/Model/Observer.php
```diff
@@ -100,11 +100,19 @@ public function actionPreDispatchAdmin($observer)
                             ->setActionName('deniedJson')
                             ->setDispatched(false);
                     } else {
-                        $request->setParam('forwarded', true)
-                            ->setRouteName('adminhtml')
-                            ->setControllerName('index')
-                            ->setActionName('login')
-                            ->setDispatched(false);
+                        if ($request->getRouteName() == 'adminhtml') {
```
colinmollenhour, line 103: What about alternatively using `Mage::app()->getStore()->isAdmin()` ? This way older routes that are not in the adminhtml namespace would still work.

fballiano:
this PR is a very interesting find, since the observer was already there I wouldn't care about finding another way to fix the problem. But yes, let's make the change proposed by @colinmollenhour and merge it.

app/code/core/Mage/Admin/Model/Observer.php
```diff
@@ -100,11 +100,19 @@ public function actionPreDispatchAdmin($observer)
                             ->setActionName('deniedJson')
                             ->setDispatched(false);
                     } else {
```
luigifab, line 102: also, can you merge the `else { if () {} else {} }` with `else if () {} else {}`?

ADDISON74:
With a little help this PR can be merged.

fballiano:
PR https://github.com/OpenMage/magento-lts/pull/2164 implements the modifications requested by other maintainers, since I couldn't add commits to this one.

PR #1078, opened 2020-06-30
Prevent resize to 0 width or height.

wexo-team:
Splitting previous Pull Request into 2 as asked by @tmotyl in https://github.com/OpenMage/magento-lts/pull/1076

This prevents a exception thrown in the GD2 class when resizing to 0 width or height.


seansan:
+1

PR #1081, opened 2020-06-30
Fixed Magento not detecting HTTPS behind a proxy.

wexo-team:
This fix adds a check for the "HTTP_X_FORWARDED_PROTO" header to determine if the scheme is HTTPS or HTTP.

### DIFF
--- a/app/code/core/Mage/Api2/Model/Request.php
+++ b/app/code/core/Mage/Api2/Model/Request.php
@@ -323,15 +323,4 @@ class Mage_Api2_Model_Request extends Zend_Controller_Request_Http
         }
         return false;
     }
-
-    /**
-     * @inheritDoc
-     */
-    public function getScheme()
-    {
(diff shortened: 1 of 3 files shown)


kkrieger85:
As far as i can see, both classes inherit from Zend_Controller_Request_Http. So it might be better to do these changes in this class to avoid duplicate code.

wexo-team:
Thank @kkrieger85,

I have now moved the fix into the class that you suggestet :-)

sreichel:
@kkrieger85 is it possible to use "squash merge" option for unified commit log?

kkrieger85:
Of course, it is

Am Do., 23. Juli 2020 um 15:53 Uhr schrieb sv3n <notifications@github.com>:

Flyingmana:
can confirm of one case, where this did work and solve related Issues, thank you for providing the patch 😀

PR #1085, opened 2020-07-01
Installer rebranding

dbachmann:
* Added openmage favicon
* Added openmage logo in installer
* Renamend magento to openmage in template
* Removed old magento logo gif


sreichel:
Shoult new translations not be added the own csv?

sreichel:
Should translation move to custom CSV to not break locale packs?

colinmollenhour:
@sreichel Yes, but let's handle that separately. Either way these strings will no longer be translated with existing locale packs.

PR #1091, opened 2020-07-06
Added dash(-) as an allowed character for store code

julienloizelet:
Allow dash in store code

### Description (*)

Currently, it is not possible to create a store with a store code containing dash (-). When using the feature "add store code in url", this can lead to conflicts with SEO purposes.

There is a lot of literature (for example https://stackoverflow.com/a/2318376/7497291) on it and I am not a SEO specialist but what I know is that I have often been asked to add the dash in the store code and I had to rewrite this resource model. That's why I submit this PR.

Furthermore, the reason why dash is not allowed in Magento is not clear for me. The only answer I found is here :
https://magento.stackexchange.com/a/237309/50208 

1. Create a store with code "en-uk". This is not possible without this PR and this becomes possible with it.

### Questions or comments

I am not sure it was necessary to create a Mage_Core_LTS.csv file : The text I changed is not used anymore in the source code,  so I hesitated to modify directly the Mage_Core.csv file because it could be considered as dead code. But, if for some reason, somebody has specific code using this old key and need this old translated key, this will not break it. That's why I created a new csv file as suggested in the README.

### DIFF
--- a/app/code/core/Mage/Core/Model/Resource/Store.php
+++ b/app/code/core/Mage/Core/Model/Resource/Store.php
@@ -64,9 +64,9 @@ class Mage_Core_Model_Resource_Store extends Mage_Core_Model_Resource_Db_Abstrac
      */
     protected function _beforeSave(Mage_Core_Model_Abstract $model)
     {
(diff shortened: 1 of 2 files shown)


colinmollenhour:
Thanks, @julienloizelet!

PR #1092, opened 2020-07-06
Add missing country calling code 599 for Netherlands Antilles (AN).

kiatng:
### Description (*)

Added missing country calling code 599 for Netherlands Antilles (AN) in Zend Locale.

`$callingcodes = Mage::app()->getLocale()->getTranslationList('phonetoterritory');` Netherlands Antilles (AN) is added in the return array.

### Questions or comments

I needed to implement a country calling code drop down. From the coutry list in `Mage_Directory`, there are 3 missing calling codes in `Mage::app()->getLocale()->getTranslationList('phonetoterritory')`: AN - Netherlands Antilles, BV - Bouvet Island, and HM - Heard Island and McDonald Islands, all of which are islands. From Google Maps, only AN is an inhabited island with hotels.

### DIFF
--- a/app/code/core/Mage/Core/Block/Abstract.php
+++ b/app/code/core/Mage/Core/Block/Abstract.php
@@ -755,6 +755,7 @@ abstract class Mage_Core_Block_Abstract extends Varien_Object
      */
     public function sortChildren($force = false)
     {
+        $this->_sortedChildren = array_values($this->_sortedChildren); // reset indexes which might have gaps after unsetting blocks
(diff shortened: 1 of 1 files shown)


sreichel (commented):
If BV (55) and HM (672) is also missing, why not add them too?

kiatng:
I didn't include BV and HM because both are uninhabited islands and has no sign of any commercial activity; whereas AN has hotels with phone numbers listed in google maps.

When I googled _IDD code 55_, it returns Brazil, whereas HM (672) seems correct. But how can we be more definite?

Issue #1093, opened 2020-07-06
Disable currencyconverterapi service by default


kkrieger85:
This option is activated by default: 
currency/currencyconverterapi/active

1) For shops which run on one currency only, you don't need this service
2) To use this service, you need to register(?) and also need an API key (even there is a free version)

Ref: https://www.currencyconverterapi.com/pricing

PR #1097, opened 2020-07-07
Cleanup: disable currencyconverterapi in config

kkrieger85:
### Description (*)
Update config file to disable currencyconverterapi service by default

### Related Pull Requests
----

1. Fixes OpenMage/magento-lts#1093

1. Login to Backend -> System -> Configuration -> Currency setup
2. Open CurrencyConverterAPI Tab
3. Enabled should be "no"


fballiano:
I'd be ok with the modification but.. what happens to the stores that use it and that would get it disabled? Is it an acceptable behaviour? maybe it should be done on the v20 branch instead of the v19?

sreichel:
> what happens to the stores that use it and that would get it disabled?

IMHO nothing. If you use fixer.io you have to enter your API and save changes to DB.

kkrieger85:
Same for currencyconverterapi  

If you use one of these services, you need to add API keys => They are stored in DB

Issue #684, opened 2019-05-30
Remove exceeding shipping quotes in cart


rafaelpatro:
Hi all! I would like to suggest removing this piece of code:
https://github.com/OpenMage/magento-lts/blob/9f9a99ba397e00c7471b2234ef1e61cc48d600f7/app/code/core/Mage/Checkout/Model/Cart.php#L462

Today cart index action are requesting shipping quotes every time. Decreasing page load and user experience.
I think this flag should respect the user action. This happens only in _estimatePostAction_.
https://github.com/OpenMage/magento-lts/blob/9f9a99ba397e00c7471b2234ef1e61cc48d600f7/app/code/core/Mage/Checkout/controllers/CartController.php#L552

So user could remove and update the cart without collecting totals in every action.

colinmollenhour:
> Today cart index action are requesting shipping quotes every time.

Are you sure this is happening on a clean Magento installation? This code has been there since 1.1.1 and I'm pretty sure it hasn't always requested shipping quotes every time you visit the cart.

discountscott:
We have had this issue for many years. We worked around it by adding a line in /app/code/core/Mage/Checkout/controllers/CartController.php

added right bbfore line 233 `$cart->addProduct($product, $params);`
`$cart->getQuote()->removeAllAddresses()->getShippingAddress()->setPostcode('')->setCollectShippingRates(false)->removeAllShippingRates();`

Effectively removes the customers shipping postalcode from the checkout and clears the collectShippingRates flag when adding additional items to cart.

tmotyl:
Can somebody elaborate me the details of the problem? And why would it be safe to remove the line?

discountscott:
We have not tested the previous proposed fix. We resolved by adding a line to the cart controller action.

The issue is as follows:

- Add item to cart
- Get shipping quote
- Leave cart and add another item
- 'Add to cart' appears to take a long time depending on shipping methods
   - Shipping quote gets run again during the add to cart action
   - We have customers that as a result click the add to cart again, causing even more delays (and incorrect qty in cart)

The resolution is to clear the shipping quote on 'add to cart'.

rafaelpatro:
Perfectly described! Thanks
A long time problem in Magento core. Since its construction.

tmotyl:
> * Get shipping quote

what do you mean by that?

rafaelpatro:
Estimate shipping with external APIs, webservices, that requires some time (5-10 seconds).
[...]

PR #1100, opened 2020-07-07
Clear shipping quote upon cart add/update/delete.

discountscott:
Clear shipping quote upon cart add/update/delete.
Issue #684

### Description (*)
This clears the shipping quote upon cart cart item add/update/delete. Resolves slow cart response due to external shipping API's.

1. Fixes OpenMage/magento-lts#684

1. Add item to cart
2. Get shipping quote
3. Leave cart and add another item
4. Item will be added to cart without the shipping quote (which causes excessive delays due to external shipping APIs)


discountscott:
Only removing the postcode in the shipping field.

If you've started the checkout process and are shipping to the billing address, then go back to cart and add another item. Everything will still populate OK if proceeding through checkout again. If different ship-to, only the postcode field will be required to enter again.

When we've tried to eliminate this issue, while leaving the zipcode, magento sees that as the shipping address changing and auto setting the collectRates flag.

This fix allows you get shipping rates while in the cart without problem. And if you add or change cart, clears shipping quote (to ensure it's not stale based on old cart qtys) and also clears the zipcode field so that it's not auto populated again.

rafaelpatro:
> If you've started the checkout process and are shipping to the billing address, then go back to cart and add another item. Everything will still populate OK if proceeding through checkout again. If different ship-to, only the postcode field will be required to enter again.

I guess this does not proceed as you described. Maybe you have a checkout extension or browser plugin that is filling the checkout form automatically. I've up in my mind that when you are not logged and leave the checkout page, the address and customer fields are lost.

rafaelpatro:
> When we've tried to eliminate this issue, while leaving the zipcode, magento sees that as the shipping address changing and auto setting the collectRates flag.

Well if Magento is auto setting a flag in a wrong way. A better approach should be fixing that piece of code too. Rather than circumvent a bad Magento code.

discountscott:
> > If you've started the checkout process and are shipping to the billing address, then go back to cart and add another item. Everything will still populate OK if proceeding through checkout again. If different ship-to, only the postcode field will be required to enter again.
> 
> I guess this does not proceed as you described. Maybe you have a checkout extension or browser plugin that is filling the checkout form automatically. I've up in my mind that when you are not logged and leave the checkout page, the address and customer fields are lost.

Are you sure about that? I just double checked. As soon as you hit continue on the billing address (with ship-to same checked), the 'sales_flat_quote_address' table is updated with the shipping info. Referencing the 'sales_flat_quote' (cart).

discountscott:
> > When we've tried to eliminate this issue, while leaving the zipcode, magento sees that as the shipping address changing and auto setting the collectRates flag.
> 
> Well if Magento is auto setting a flag in a wrong way. A better approach should be fixing that piece of code too. Rather than circumvent a bad Magento code.

Agreed 👍 

I had just seen this issue for something that we "resolved" on our own years ago. It was always amazing to me that this behavior existed in the first place. The FedEx rates are too slow for this...

Happy to look a bit deeper and see if it can be resolved in a better way.

rafaelpatro:
Suggestion...
Analysing `$cart->save()` in add action...
```php
    public function save()
    {
        Mage::dispatchEvent('checkout_cart_save_before', array('cart'=>$this));

        $this->getQuote()->getBillingAddress();
        $this->getQuote()->getShippingAddress()->setCollectShippingRates(true);
        $this->getQuote()->collectTotals();
        $this->getQuote()->save();
        $this->getCheckoutSession()->setQuoteId($this->getQuote()->getId());
        /**
         * Cart save usually called after changes with cart items.
         */
        Mage::dispatchEvent('checkout_cart_save_after', array('cart'=>$this));
        return $this;
    }
```

Maybe instead of the `$cart->save` we can use `$cart->getQuote()->save();` for add/delete actions.
Keeping your fixes before. But now we can save address fields. Makes sense?

discountscott:
That is the problem function 'save' is called from all the above actions. Add to cart, Update qty, Delete item, and Estimate Shipping.

if you set setCollectShippingRates(false) in the save function. The shipping quote is forever kept even if adding new items. Clicking the 'estimate shipping' button does nothing...

```
$cart->getQuote()->removeAllAddresses()->getShippingAddress()->setPostcode('')->setCollectShippingRates(false)->removeAllShippingRates();

$cart->addProduct($product, $params);

if (!empty($related)) {
  $cart->addProductsByIds(explode(',', $related));
}

$cart->save();
```

When save is called (even if you set collectShippingRates(false) before the addProduct()), it will get set back again...

rafaelpatro:
Yep agree with you. It was a catastrophic suggestion.

Issue #1102, opened 2020-07-09
Deleting reviews leaves orphan data in a database table


ADDISON74:
This issue is related to **review_entity_summary** table in Magento database.

Just delete a review in Backend then check all tables related to review inside the database. All data are deleted except those from review_entity_summary table.

fballiano:
You mean when deleting the last review of a product (or other entity)?

Because the review_entity_summary should just get updated with the count and average or the product ratings.

ADDISON74:
You are correct in my case deleting a review won't update the product ratings. It is a Magento 1 not OpenLTS. I will check more in detail. If it is a bug in Magento 1 I will close the ticket.

ADDISON74:
Please note if you delete ALL reviews **review_entity_summary** table won't be updated. A better implementation would be by deleting all reviews Magento should empty the table review_entity_summary data like starting from scratch.

fballiano:
I've tried to implement a fix.

ADDISON74:
I will close this issue. There is a fix tested by me here #1103.

PR #1103, opened 2020-07-09
Clean review_entity_summary when reviews are deleted

fballiano:
This PR should fix bug https://github.com/OpenMage/magento-lts/issues/1102

When deleting all reviews of an entity the review_entity_summary isn't cleaned correctly. I tried many different ways to approach the problem but the implemented one seems the easiest and most effective.

Also, the default implementation some of the methods of the Mage_Review modules seem incomplete, they lack the "entity_type" parameters which make the implementations prone to errors when different entities have the same PkValue.

### DIFF
--- a/app/code/core/Mage/Review/Model/Resource/Review.php
+++ b/app/code/core/Mage/Review/Model/Resource/Review.php
@@ -249,6 +249,25 @@ class Mage_Review_Model_Resource_Review extends Mage_Core_Model_Resource_Db_Abst
      */
     public function afterDeleteCommit(Mage_Core_Model_Abstract $object)
     {
+        $read_adapter = $this->_getReadAdapter();
+        $select = $read_adapter->select()
+            ->from(
+                $this->_reviewTable,
+                array(
+                    'review_count' => new Zend_Db_Expr('COUNT(*)')
+                )
+            )
+            ->where("entity_id = ?", $object->getEntityId())
+            ->where("entity_pk_value = ?", $object->getEntityPkValue());
+        $total_reviews = $read_adapter->fetchOne($select);
+        if ($total_reviews == 0) {
+            $this->_getWriteAdapter()->delete($this->_aggregateTable, array(
+                'entity_type = ?'   => $object->getEntityId(),
+                'entity_pk_value = ?' => $object->getEntityPkValue()
+            ));
+            return $this;
+        }
+
         $this->aggregate($object);
 
         // reaggregate ratings, that depended on this review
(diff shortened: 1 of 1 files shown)


fballiano:
sorry to come again on this after so long:
- tested reviews done in backend and the patch removes the review, the ratings and recalculates the aggregation, so I think it work for this case
- tested reviews done in frontend (although the review submit form wasn't showing, I had to change app/design/frontend/rwd/default/template/review/product/view/list.phtml to make it appear) and also here it seems that the patch works fine, removing the reviews, the ratings and recalculating the aggregations

Am I missing something?

ADDISON74:
I created 3 reviews in Frontend and approved them in Backend. In the database table **review_entity_summary** columns [reviews_count] and [rating_summary] have changed accordingly. In the Backend I changed the rating and in the table the percentage was recalculated correctly. Then I deleted the reviews one by one and also the changes in the table were correct. After I deleted the last review of the product in the table **review_entity_summary** there was no record for it, which is correct and the issue I reported initially is solved.

Then I created 3 reviews in Backend and I repeated the previous steps, I also didn't find anything wrong, and I didn't find any records in the logs.

If there are no objections to the code, my opinion this commit must be merged.

ADDISON74:
@fballiano: In table **review_entity_summary** for a product there are two rows. They differ only in the values on the store_id column, 0 (admin) and 1 (default). Do you know the reason why there is a row for admin?

fballiano:
> @fballiano: In table **review_entity_summary** for a product there are two rows. They differ only in the values on the store_id column, 0 (admin) and 1 (default). Do you know the reason why there is a row for admin?

no idea 😅

fballiano:
@admins what can we do to move this forward?

Flyingmana:
Getting some approving Reviews usually helps.
Shared the PR via twitter to increase its visibility.

fballiano:
@ADDISON74 I think you tested the fix, could you "approve" the PR with green button on https://github.com/OpenMage/magento-lts/pull/1103/files (and then select "approve")?

Thanks!

ADDISON74:
@fballiano - I tested it and I didn't find anything wrong, no errors, Your changes solve that initial issue. Unfortunately I don't have that button to approve your work because only a small group has this privilege and unfortunately these people don't have time to approve. There are so many untested PR's and it's a shame the work and contribution of those who are still interested in this project. I wonder what it will be like in 1 year ...

ADDISON74 (approved):

luigifab (approved):
`array(` Seems good. `)`

To clean up data:  `DELETE FROM review_entity_summary WHERE entity_pk_value NOT IN (SELECT entity_pk_value FROM review)`

PR #1104, opened 2020-07-09
Execute the total model 'shipping' after 'tax_subtotal'

hannes011:
This pull request solves the problem I've mentioned earlier in #80 :

### Description (*)
This is a major change which is causing troubles for the calculation of shipping rate! Is "shipping" really needed as prerequisite for subtotal?
A: shipping is not included within subtotal
B: some shipping methods calculate there rate based on subtotal

This change led to strange behavior: once the totals are already calculated and stored/cached then the subtotal doesn't include tax anymore (as expected) BUT when totals will be collected first time the "shipping" total is executed before "tax_subtotal", so the subtotal still contains tax, which leads to inconsistency...

1. Fixes OpenMage/magento-lts#80

This issue applies only to external shipping modules like matrixrates from webshopapps (I haven't tested it but it might apply to tablerates from Magento core, too)
1. configure two shipping rates for a certain threshold (e.g. rate A for subtotal < 10 EUR and rate B for subtotal > 10 EUR)
2. add a product to cart (with tax 11 EUR, and without tax 9.24 EUR)
3. in step "shipping method" I can choose rate A since subtotal is expected to be without tax
4. in final step I click "place order" and get an error message "Please select a shipping method"
5. I go back to step shipping method and now I get rate B

This different behavior depends on whether totals are already calculated and cached (then subtotal does not contain tax anymore because subtotal_tax is already collected) or not (then the total model "shipping" is executed before "subtotal_tax")!

### DIFF
--- a/app/code/core/Mage/Tax/etc/config.xml
+++ b/app/code/core/Mage/Tax/etc/config.xml
@@ -162,7 +162,7 @@
                 <totals>
                     <tax_subtotal>
                         <class>tax/sales_total_quote_subtotal</class>
-                        <after>subtotal,nominal,shipping,freeshipping</after>
+                        <after>subtotal,nominal,freeshipping</after>
                         <before>tax,discount</before>
                     </tax_subtotal>
                     <tax_shipping>


ADDISON74:
In the latest version of vanilla Magento 1.9.4.5 this line modified by PR is the following

``
<after>freeshipping</after>
``

In Magento LTS/OpenMage LTS  the line bellow cancels Free Shipping feature when is set in store configuration.

``
<after>subtotal,nominal,shipping,freeshipping</after>
``

For more information check the reported issue here #1813. 

I will test your proposal an come back with a feedback.

ADDISON74:
By using any of the lines bellow in /app/code/core/Mage/Tax/etc/config.xml at line 165 solves the issue reported here #1813.

<after>freeshipping</after> => Vanilla Magento 1.9.4.5

or

<after>subtotal,nominal,freeshipping</after> => PR #1104

The variant below is now in all Magento LTS/OpenMage versions but the "Free Shipping" feature no longer works. Before merging this PR I would be interested to know what are the differences and real advantages in using the original Magento version or the proposed one.

<after>subtotal,nominal,shipping,freeshipping</after>

ADDISON74 (approved):

fballiano:
This is a very delicate change, it would require extensive tests that are impossible to me to conduct.

fballiano:
This PR doesn't have updates in 3 years, it's an extremely delicate topic, the original issue is closed as fixed... sorry but I'll close this PR too.

ithuis:
can we  reopen this ... 

This PR fixes this situation for me. 

When i configure a minimum cart total before "freeShipping" is available, in file
```/app/code/core/Mage/Shipping/Model/Carrier/Freeshipping.php```

the "$request->getBaseSubtotalInclTax()" returns 0. 

after this pull request, it returns the correct cart amount.

fballiano:
github doesn't let me reopen it for some reason so it will have to be re-created

ADDISON74:
I am still using this change because of this issue #1813.

PR #1107, opened 2020-07-13
Only resize images when their file size has improved

lemundo-team:
This change is fixing issues with increasing file sizes after image upload.

### Description (*)
After uploading already web optimized images it appears that some of them have increased image size. This fix keeps resized images only when their size is decreased in comparison to the original uploaded one.

1. Upload different sized images in  WYSIWYG editor
2. Check their sizes in /media/

### DIFF
--- a/app/code/core/Mage/Core/Model/File/Validator/Image.php
+++ b/app/code/core/Mage/Core/Model/File/Validator/Image.php
@@ -98,6 +98,7 @@ class Mage_Core_Model_File_Validator_Image
                 //replace tmp image with re-sampled copy to exclude images with malicious data
                 $image = imagecreatefromstring(file_get_contents($filePath));
                 if ($image !== false) {
+                    $tmpFilePath = $filePath . '_';
                     $img = imagecreatetruecolor($imageWidth, $imageHeight);
                     imagealphablending($img, false);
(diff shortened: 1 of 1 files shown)


tmotyl:
Hi
Thanks for the PR!
Do you have an example of the file which is bigger after scaling so I can test the PR?
Have you checked if magento keeps the original file format/extension always? If not then keeping the original might not be what is expected?

colinmollenhour:
I love the concept, but what about the security implications? I don't know a whole lot about how these malicious files are formatted but I don't think we want to sacrifice security. I think I recall a previous discussion around this issue where the user wanted to accomplish the same and my suggestion was to make the auto-recompress step configurable for users that were more concerned about image compression than admin users uploading malicious files.

kkrieger85 (approved):
LGTM

fballiano:
The PR is very interesting but I have to say that I think I agree with @colinmollenhour, I'd also prefer the safe "resize anyway" approach

luigifab:
You can't do that, because you need to reprocess images to avoid malicious content (like in #475).

fballiano:
ok so it seems we kinda all agree that it would be a little bit of a risk so, sorry for that, but I'm closing this PR.

PR #1109, opened 2020-07-13
Ensure there are no emails without recipient in email queue

lemundo-team:
Additional check of email recipients in order to ensure there are no invalid emails without recipient in queue

### Description (*)
It appeared several times that there are emails without recipients in the emails queue - even if the email address is mandatory for checkout. This pull request is not fixing the origin issue but it ensures that this one wrongly added email is not blocking all queued emails which is leading to unsatisfied customers.
Consider it more as a security improvement to keep the shop running.

There is no known test scenario but it is solving issues that happens once every 2000 orders.

### DIFF
--- a/app/code/core/Mage/Core/Model/Email/Queue.php
+++ b/app/code/core/Mage/Core/Model/Email/Queue.php
@@ -95,7 +95,7 @@ class Mage_Core_Model_Email_Queue extends Mage_Core_Model_Abstract
      */
     protected function _beforeSave()
     {
-        if (empty($this->_recipients) || !is_array($this->_recipients)) {
+        if (empty($this->_recipients) || !is_array($this->_recipients) || empty($this->_recipients[0])) { // additional check of recipients information (email address)
             Mage::throwException(Mage::helper('core')->__('Message recipients data must be set.'));
         }
         return parent::_beforeSave();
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Core/Model/Email/Queue.php
```diff
@@ -238,13 +238,13 @@ public function send()
 
                 try {
                     $mailer->send();
+                    unset($mailer);
```
colinmollenhour, line 241: Why move the unset?

kkrieger85: I think the `unset ` comes from line 245 https://github.com/OpenMage/magento-lts/pull/1109/files#diff-e408845a99067b7da8df6e1c971d74dd81e8d02f15b4b6111ce1d4389ae45e49L245

kkrieger85 (approved):

ADDISON74 (approved):

luigifab:
The move of the unset is wrong no?

https://github.com/OpenMage/magento-lts/blob/3005c60fa78edc88ce34c016442b185aba0ea43e/app/code/core/Mage/Core/Model/Email/Queue.php#L239-L247

colinmollenhour:
Agreed with the change overall. Not sure if the location of the unset matters much, the PHP garbage collector should be able to clean it up either way I think.

leissbua:
i think the whole solution does not work nor is well tested - remove an recipient and the queue processing is brought to an end:

``` 
---EXCEPTION---
Mage_Core_Exception: Message recipients data must be set. in /home/cloudpanel/htdocs/www.domain.com/app/Mage.php:656
Stack trace:
#0 /home/cloudpanel/htdocs/www.domain.com/app/code/core/Mage/Core/Model/Email/Queue.php(99): Mage::throwException('Message recipie...')
#1 /home/cloudpanel/htdocs/www.domain.com/app/code/core/Mage/Core/Model/Abstract.php(329): Mage_Core_Model_Email_Queue->_beforeSave()

kiatng:
@leissbua Sorry for the problem you face. How do you remove a recipient? Was it the first recipient that was removed? Does it throw the exception when the second or third recipient is removed?

PR #1111, opened 2020-07-14
fix cancel order bug

drwilliams:
### Description (*)

When trying to cancel an order I was getting an error related to changes in pull request #1031. When trying to decrement times_used in salesrule_customer, you need to use rule_id and not the id of the coupon.

### Related Pull Requests
#1031


boesbo:
Every when are release tags updated? this bugfix is very important @colinmollenhour @kkrieger85 @drwilliams

Issue #1116, opened 2020-07-16
Unable to place order after canceled order


drwilliams:
There is an error when placing an order when using a promo code that has been used on a canceled order. This is because when you cancel an order the times_used in salesrule_coupon_usage becomes zero and it tries to insert a new entry into the database. 

### Preconditions (*)

1. Using OpenMage 19.4.5

### Steps to reproduce (*)

1. Log in to a customer account
2. Place an order and use a promo code
3. Cancel the order
4. Try to place another order with the same promo code and you will get an error when submitting the order

### Expected result (*)

1. Order placed successfully

### Actual result (*)

1. Integrity constraint violation: 1062 Duplicate entry '10961-661' for key 'PRIMARY', query was: INSERT INTO `salesrule_coupon_usage` (`coupon_id`, `customer_id`, `times_used`) VALUES (\?, \?, \?) in /cache/lib/Zend/Db/Statement/Pdo.php:235

VincentMarmiesse:
+1

sreichel:
@all-contributors add @drwilliams bug

PR #1117, opened 2020-07-16
Prevent duplicate entry when updating salesrule_coupon_usage

drwilliams:
### Description (*)

If you cancel a customer order with a promo code and try to place an order with the same promo code it fails because of an Integrity constraint violation: 1062 Duplicate entry on salesrule_coupon_usage since times_used goes to 0 when the order is canceled. I removed the check for $timesUsed > 0 so that it will run as an update instead of an insert.

### Related Pull Requests

#1031 

1. Fixes OpenMage/magento-lts#1116

1. Log in to customer account
2. Place an order with a promo code
3. Cancel the order
4. While still logged in, place another order with the same promo code (This failed before making this change.)

### DIFF
--- a/app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
+++ b/app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
@@ -62,16 +62,19 @@ class Mage_SalesRule_Model_Resource_Coupon_Usage extends Mage_Core_Model_Resourc
         $timesUsed = $read->fetchOne($select, array(':coupon_id' => $couponId, ':customer_id' => $customerId));
 
         if ($timesUsed !== false) {
-            $this->_getWriteAdapter()->update(
-                $this->getMainTable(),
-                array(
-                    'times_used' => $timesUsed + ($decrement ? -1 : 1)
-                ),
-                array(
-                    'coupon_id = ?' => $couponId,
-                    'customer_id = ?' => $customerId,
-                )
-            );
+            $timesUsed += ($decrement ? -1 : 1);
+            if($timesUsed >= 0) {
+                $this->_getWriteAdapter()->update(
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
```diff
@@ -61,7 +61,7 @@ public function updateCustomerCouponTimesUsed($customerId, $couponId, $decrement
 
         $timesUsed = $read->fetchOne($select, array(':coupon_id' => $couponId, ':customer_id' => $customerId));
 
-        if ($timesUsed !== false && $timesUsed > 0) {
```
kkrieger85, line 64: I guess there would be a problem if (somehow) the couponTimesUsed will be decremented several times until timeUsed less than 0. 
Maybe you should add $decrement to the if statement. Ensure to not descrement $timesUsed to less than 0

colinmollenhour:
I haven't had a chance to look at this closely but is it related to #1031 which has already been merged? /cc @mattdavenport

VincentMarmiesse:
> I haven't had a chance to look at this closely but is it related to #1031 which has already been merged? /cc @mattdavenport

Yes it is, but #1031 caused some issues, especially #1116.
This is critical as the user is not able to place an order in the following scenario:
* User uses a coupon
* User cancels his order
* User re-uses the same coupon => he's not able to place his order

sreichel (approved):
Tested with latest 19.4.6

app/code/core/Mage/SalesRule/Model/Resource/Coupon/Usage.php
colinmollenhour: The new code has a second if clause that would prevent this.

Issue #573, opened 2018-11-28
Mage_CatalogRule_Model_Rule: detect changes and only then run the indexer. Advise?


seansan:
In this function, should we / can we - check if there are products modified? (if there is a need to reindex)

Suppose no rule is applicable and no changes are made: then why run the reindexer?

```
    /**
     * Apply all price rules, invalidate related cache and refresh price index
     *
     * @return Mage_CatalogRule_Model_Rule
     */
    public function applyAll()
    {
        $this->getResourceCollection()->walk(array($this->_getResource(), 'updateRuleProductData'));
        $this->_getResource()->applyAllRules();
        $this->_invalidateCache();
        $indexProcess = Mage::getSingleton('index/indexer')->getProcessByCode('catalog_product_price');
        if ($indexProcess) {
            $indexProcess->reindexAll();
        }
    }
```

seansan:
Suppose we add a new method? It gives us a global hint if there are rules for any store

here app/code/core/Mage/CatalogRule/Model/Resource/Rule.php

`/**
     * Get active rules based on date now and if active
     *
     * @return array
     */
    public function getActiveRulesCount()
    {
        $date     = Mage::app()->getLocale()->date()->getTimestamp();
        $adapter = $this->_getReadAdapter();
        $select = $adapter->select()
            ->from($this->getTable('catalogrule'))
            ->where('is_active = ?', 1)
            ->where('from_time = 0 or from_time < ?', $date)
            ->where('to_time = 0 or to_time > ?', $date)
            ->order('sort_order');
        return $adapter->fetchAll($select);
    }`

tmotyl:
I've just stumbled upon similar issue.
I have 0 catalog rules in db, yet Magento runs reindex (including creation of temporary tables on each product save).
![image](https://user-images.githubusercontent.com/515397/88109855-82a63c00-cbab-11ea-8d3f-6a304ee64e0f.png)

I think that we could add a return inside applyAllRulesToProduct

```php
public function applyAllRulesToProduct($product)
    {
        if (is_numeric($product)) {
            /** @var Mage_Catalog_Model_Product $product */
            $product = Mage::getModel('catalog/product')->load($product);
        }

        $productWebsiteIds = $product->getWebsiteIds();

        /** @var Mage_CatalogRule_Model_Resource_Rule_Collection $rules */
        $rules = Mage::getModel('catalogrule/rule')->getCollection()
            ->addFieldToFilter('is_active', 1);

       //FIX
[...]

PR #1119, opened 2020-07-21
Do not calculate catalogrules when there are none

tmotyl:
### Description (*)
Add early return to applyAllRulesToProduct
to speed up product save when there are no catalog rules.
This prevents Magento from doing heavy db operations including
creating temporary tables just to find out they are empty.
Like: CREATE TEMPORARY TABLE `catalogrule_product_price_tmp`

1. Relates: OpenMage/magento-lts#573

1. Given you have magento with no catalog rules and with rules
2. Check whether saving product from admin panel works

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Rule.php
+++ b/app/code/core/Mage/CatalogRule/Model/Rule.php
@@ -342,6 +342,9 @@ class Mage_CatalogRule_Model_Rule extends Mage_Rule_Model_Abstract
         /** @var Mage_CatalogRule_Model_Resource_Rule_Collection $rules */
         $rules = Mage::getModel('catalogrule/rule')->getCollection()
             ->addFieldToFilter('is_active', 1);
+        if ($rules->count() === 0) {
+            return $this;
+        }
         foreach ($rules as $rule) {
             $websiteIds = array_intersect($productWebsiteIds, $rule->getWebsiteIds());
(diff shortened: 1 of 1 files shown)


sreichel (dismissed):
dupe ... see above

colinmollenhour:
Have you also tested removing the last rule?

tmotyl:
> I think there is no need to load rules ... how about this?
> 
> ```
> 
>         if (!$rules->getSize()) {
>             return $this;
>         }
> ```

if there are no rules, then count will not hurt, and when there are rules, they will be loaded anyway few lines later

tmotyl:
> dupe ... see above

what do you mean?

sreichel:
> 
> 
> > I think there is no need to load rules ... how about this?
> > ```
> > 
> >         if (!$rules->getSize()) {
> >             return $this;
> >         }
> > ```
> 
> if there are no rules, then count will not hurt, and when there are rules, they will be loaded anyway few lines later



> 
> 
> > dupe ... see above
> 
> what do you mean?

Accidently requested changes two times.

Can you please change `count()` to `getSize()`. I guess tools like MCG will complain about it.

tmotyl:
> Can you please change `count()` to `getSize()`. I guess tools like MCG will complain about it.

It will add one query more. 
One to get size and one to load (if there are rules).

fballiano (approved):

ADDISON74 (approved):

PR #1123, opened 2020-07-23
Use latest media editor plugin (989c4b5 from 2015030) for tinymce v3.5

theroch:
### Description (*)
 * The option to embed videos in the editor was broken in magento tinymce plugin version.
 * Added support for video track-tag to allow subtitles for accessibility (WCAG, BITV 2.0) (only src) and keeping manual entered data
 * Disabled flash fallback for mp4 videos in tinymce (magento disabled this before too)
 * Fixed problem with placeholder in editor while using percentage dimensions for videos

### Related Pull Requests

 * None known

 * None known

1. Create CMS page
2. Upload some mp4 and vtt (Web Video Text Tracks) files to the tinemcy media storage. It is not possible to do that with magento at the moment.
3. Use tinemce editor to create a html5 video tag and add a vtt track file.

### Questions or comments

* Maybe we will provide a other pull request to add support for uploading other files than images in magento.


Flyingmana:
The new code triggers some warnings in our code quality tool, do you have a link from where you downloaded it, so we can compare it with this?
Do you by chance know, if the new code also supports IE11?

theroch:
> The new code triggers some warnings in our code quality tool, do you have a link from where you downloaded it, so we can compare it with this?

Sure, the main code is from the offical tinymce repo from github for v3.x:
https://github.com/tinymce/tinymce/blob/3.x/jscripts/tiny_mce/plugins/media/editor_plugin_src.js [989c4b5 from 20150630)
All the warnings are from this code, I think.

> Do you by chance know, if the new code also supports IE11?

The video tag is supported by ie from v9.
The track tag is supported by ie from v10.
The rest of the code is from 2015, so it should support IE11.
But we didn't test the support for IE11 because we are thinking there is with Edge a better alternative for Windows 8.1 and Windows 10.

ADDISON74:
Microsoft already announced all IE versions will be removed from Windows 10.  I would not worry about these versions, the number of those who use them is decreasing and in addition they are unnecessarily exposed to security breaches in the conditions in which there are free alternatives such as Firefox, Edge, Chrome, Vivaldi and many more.

If this update does not create errors and brings new features it is a good addition.

ADDISON74:
I would recommend testing this version 4.5.x with OpenMage: https://github.com/tinymce/tinymce/tree/4.5.x/js/tinymce.

If you want to remain to version 3.x then there are some changes in all files that should be revised. This editor is an important feature in Backend and we have to avoid any issues with this upgrade in case o merging. I am concerning especially about two files editor_plugin. and editor_plugin_src.js. Magento team in many JS files adapted them to their own needs.

fballiano:
Hi @theroch, would you be able to update the PR that now has conflicts? Thank you! :-)

fballiano (approved):

PR #1128, opened 2020-07-29
New event "order_status_changed_before_save"

woutersamaey:
### Description (*)
New event "sales_order_state_change" allows custom modules to detect an order state/status change and trigger appropriate actions. I've seen modules rewrite the order model because this was missing.

Very handy improvement.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -2355,6 +2355,11 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
         }
 
         $this->setData('protect_code', substr(md5(uniqid(mt_rand(), true) . ':' . microtime(true)), 5, 6));
+
+        if ($this->getStatus() !== $this->getOrigData('status')) {
+            Mage::dispatchEvent('order_status_changed_before_save', ['order' => $this]);
+        }
+
         return $this;
     }
 


kiatng:
I actually need this event in some of my projects. However, I fire it from `_beforeSave`:

```php
    {
        // ...
        // Just before existing the function.
        if ($this->getStatus() != $this->getOrigData('status')) {
            Mage::dispatchEvent('order_status_changed_before_save', ['order' => $this]);
        }
        return $this;
    }
```

It doesn't check for state changes as condition of firing, which is less granular; a state may not have changed while status changes. However, you can easily check state changes in the observer. 

I like firing in `_beforeSave` because I know at this point the states of the `order` model have settled, ie, the states of its attributes, the order items, order history, etc., have settled and are ready to be saved.

woutersamaey:
Updated this PR to reflect @kiatng example. This is indeed better.

luigifab (commented):
Could you add the new event into [EVENTS.md](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/EVENTS.md).
Personally I would have used `!=`, but never mind.

theroch (approved):

fballiano (approved):

fballiano:
I think this one could be merged as it is (since it's not modifiable by other admins) and then I'll modify EVENTS.md in another PR.

fballiano:
oops, it's not mergeable because of conflicts and only @woutersamaey can fix them

fballiano:
completed via https://github.com/OpenMage/magento-lts/pull/2927

Issue #1135, opened 2020-08-04
Error calculating the special price for Bundle products


boesbo:
### Preconditions (*)
1. 1.9.4.x
2. Bundle product with special price

### Steps to reproduce (*)
1. Add a percentage value in the special price field of a bundled product.

### Expected result (*)
1. The price must be calculated as a percentage of the "price" field

### Actual result (*)
1. Only on the product detail page and in the cart is the price calculated by interpreting the special_price field, which in the Bundle products is a percentage, as if it were a fixed price.
2. No problem in the logic of price indexing

**Backend value**
<img width="555" alt="Screenshot 2020-08-04 at 12 42 53" src="https://user-images.githubusercontent.com/12744378/89285100-12d68d80-d650-11ea-8578-edfc5ffcbec0.png">

**Wrong frontend result**
<img width="288" alt="Screenshot 2020-08-04 at 12 43 11" src="https://user-images.githubusercontent.com/12744378/89285110-1538e780-d650-11ea-8489-c42488252853.png">

boesbo:
https://github.com/OpenMage/magento-lts/pull/1136

PR #1136, opened 2020-08-04
Fix special price for bundle products

boesbo:
### Description (*)
The bug causes: only on the product detail page and in the cart is the price calculated by interpreting the special_price field, which in the Bundle products is a percentage, as if it were a fixed price.

1. Fixes OpenMage/magento-lts#1135


tmotyl:
@boesbo would usage of  "static::" instead of "self::" solve the problem?
The method is still a static method, so it should not be called with "->"

boesbo:
> @boesbo would usage of "static::" instead of "self::" solve the problem?
> The method is still a static method, so it should not be called with "->"

! I have tested it and it is fine as "static". I have updated the PR.

PR #1137, opened 2020-08-06
Make Curl Client HTTP/2 compatible

Schrank:
- Refactoring Mage_HTTP_Client_Curl::parseHeaders

Before this fails with `Exception: Invalid response line returned from server: HTTP/2 301` after the change it works.
```
$c = new Mage_HTTP_Client_Curl();
$c->get('https://facebook.com');
```
### Questions or comments
- I'm not sure wether HTTP/2 allows Status Codes (line 473)
- this is PHP 7.1 code (not sure what openmage is expecting), list shortcut and return void

### DIFF
--- a/lib/Mage/HTTP/Client/Curl.php
+++ b/lib/Mage/HTTP/Client/Curl.php
@@ -440,8 +440,8 @@ implements Mage_HTTP_IClient
                 list($name, $value) = $out;
             }
 
-            if ('' !== $name) {
-                if ('Set-Cookie' === $name) {
+            if ($name !== '') {
+                if ($name === 'Set-Cookie') {
                     if (!isset($this->_responseHeaders[$name])) {
                         $this->_responseHeaders[$name] = [];
                     }
--- a/lib/Mage/HTTP/Client/Curl.php
+++ b/lib/Mage/HTTP/Client/Curl.php
@@ -464,7 +464,7 @@ implements Mage_HTTP_IClient
      */
     protected function validateHttpVersion(array $line)
     {
-        if ($line[0] === 'HTTP/1.1') {
+        if ($line[0] === 'HTTP/1.0' || $line[0] === 'HTTP/1.1') {
             if (count($line) !== 3) {
                 $this->doError('Invalid response line returned from server: ' . implode(' ', $line));
             }
--- a/lib/Mage/HTTP/Client/Curl.php
+++ b/lib/Mage/HTTP/Client/Curl.php
@@ -421,31 +421,29 @@ implements Mage_HTTP_IClient
      * Parse headers - CURL callback functin
      *
      * @param resource $ch curl handle, not needed
-     * @param string $data
+     * @param string   $data
+     *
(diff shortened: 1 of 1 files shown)


colinmollenhour (dismissed):

theroch:
Thanks for your PR.
Should Openmage support PHP 7.0 anymore? PHP 7.0 is EOL since over a year and most linux distributions supports newer php versions. PHP 7.1 is availabe since december 2016.

theroch (dismissed):
Code is now compatible to PHP 7.0 and newer

sreichel:
> ### Questions or comments
> * this is PHP 7.1 code (not sure what openmage is expecting), list shortcut and return void

I'd like to make use of it, but not in the main release line. Guess dropping php-7.0 support would break a lot installations.

@colinmollenhour @Flyingmana not sure about planned release strategy, but setting min requirements to php-7.3 for develop/future branch could be a start.

lib/Mage/HTTP/Client/Curl.php
```diff
@@ -456,10 +454,34 @@ protected function parseHeaders($ch, $data)
...
     }
 
+    /**
+     * @param array $line
+     *
+     * @throws Exception
+     */
+    protected function validateHttpVersion(array $line)
+    {
+        if ($line[0] === 'HTTP/1.1') {
```
tmotyl, line 467: What about http 1.0? Should it be also handled here?

tmotyl (changes requested):
add support for http 1.0

lib/Mage/HTTP/Client/Curl.php
sreichel:
```suggestion
        if ($line[0] === 'HTTP/1.0' || $line[0] === 'HTTP/1.1') {
```

tmotyl (approved):

tmotyl (changes requested):
code should be made 7.0 compatible

tmotyl:
hmm, I see it's alfready 7.0 compatible.
I've tested it and it works.
Alternative solution would be to force http 1.0 adding
```
        $this->curlOption(CURLOPT_HTTP_VERSION, CURL_HTTP_VERSION_1_1);

```
here:
https://github.com/openmage/magento-lts/blob/1.9.4.x/lib/Mage/HTTP/Client/Curl.php#L354

tmotyl (approved):

tmotyl:
some related links from community:
https://community.magento.com/t5/Magento-2-x-Version-Upgrades/Invalid-response-line-returned-from-server-HTTP-2-200/td-p/113833/page/2
https://magento.stackexchange.com/questions/249737/magento-2-invalid-response-line-returned-from-server-http-2-200

fballiano:
Sorry guys I rebased the PR and made a small change to 2 "if" statements (since the notation that was used is almost never used in our code)

colinmollenhour (approved):

fballiano (approved):

PR #1138, opened 2020-08-07
Backport Magento security Update 2.3.5-p2 from 20200728 APSB20-47

theroch:
### Description (*)
* Update _validateSecretKey in Mage_Adminhtml_Controller_Action to avoid time attacks related to CVE-2020-9690
 * It is important to provide the user-supplied string as the second parameter, rather than the first.

### Related Pull Requests

 - none

 - none

1.  Do several action in backend

### Questions or comments

 - I've reviewed the p2 patch, but I think the other fixed CVEs are realted to M2 only.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Controller/Action.php
+++ b/app/code/core/Mage/Adminhtml/Controller/Action.php
@@ -388,8 +388,8 @@ class Mage_Adminhtml_Controller_Action extends Mage_Core_Controller_Varien_Actio
             return true;
         }
 
-        if (!($secretKey = $this->getRequest()->getParam(Mage_Adminhtml_Model_Url::SECRET_KEY_PARAM_NAME, null))
-            || $secretKey != Mage::getSingleton('adminhtml/url')->getSecretKey()) {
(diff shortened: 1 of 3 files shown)


Flyingmana:
Thank you a lot for checking this and contributing.
I added you to our internal advisories regarding this.

When you contact us over the recommended way (noted in https://github.com/OpenMage/magento-lts/blob/1.9.4.x/SECURITY.md ) we can handle this in private and better coordinate the release of security relevant information and releases.


Regarding the Patch, we use hash_equals() already on other places direct, so we do not need the fallback.
Also its a risk for itself to copy this and not using it via a composer dependency, as there may be related security issues in the future and we would not get an info about it this way.
 (using such dependencies via composer is on our list)

theroch:
@Flyingmana 
Thx for your explanations and sorry for breaking the reporting process. It was the first time providing security related information for me. I will stick to it next time.

Should the PR to be closed now?

Flyingmana:
No Problem, its also for us the first time to go through this process as a whole :)

PR #1143, opened 2020-08-11
Show the pending payment status

boesbo:
### Description (*)
Make the "pending payment" status visible in the source model. It is not clear why the line has been commented in the Magento core, it is only used in the settings form there are no automatisms related to this array.

1. If I create a custom payment method I would like to be able to create a configuration form to set the default status and select "pending payment".

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Order/Status.php
+++ b/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Order/Status.php
@@ -32,7 +32,7 @@ class Mage_Adminhtml_Model_System_Config_Source_Order_Status
     // set null to enable all possible
     protected $_stateStatuses = array(
         Mage_Sales_Model_Order::STATE_NEW,
-//        Mage_Sales_Model_Order::STATE_PENDING_PAYMENT,
+        Mage_Sales_Model_Order::STATE_PENDING_PAYMENT,
         Mage_Sales_Model_Order::STATE_PROCESSING,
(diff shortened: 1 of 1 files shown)


boesbo:
News about?

boesbo:
> > there are no automatisms related to this array.
> 
> 👍

This source config (adminhtml/system_config_source_order_status) is used only in the configuration forms (sistem.xml).

I don't see automatic influences in Mage_Sales_Model_Order_Config::getStateStatuses, am I missing something?

kiatng:
For custom payment methods, we can define our own `source_model` in `system.xml` that includes any order _state-status_, including custom ones.  So, my humble opinion: this PR is unnecessary, but on the other hand, I cannot see any harm with it.  @boesbo, can you elaborate on how you use `pending_payment` in your payment flow, or how it can be useful for other payment methods?

boesbo:
> For custom payment methods, we can define our own `source_model` in `system.xml` that includes any order _state-status_, including custom ones. So, my humble opinion: this PR is unnecessary, but on the other hand, I cannot see any harm with it. @boesbo, can you elaborate on how you use `pending_payment` in your payment flow, or how it can be useful for other payment methods?

What you say is right. The reason that made me want to do the PR and that I suspect was a comment on the core code that went into release for no specific reason. I try to understand in which version it happened and if there are changelog.

boesbo:
The line has been commented since Magento version 1.7 so I don't think this is a strategic change.

PR #1149, opened 2020-08-17
Suggestion to add some debugging to lockfile (to identify fileownersh…

seansan:
…ip issues)

Suggestion to add some debugging to lockfile (to identify fileownership issues)

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Index/Model/Lock.php
+++ b/app/code/core/Mage/Index/Model/Lock.php
@@ -143,12 +143,14 @@ class Mage_Index_Model_Lock
                 $result = flock($this->_getLockFile($lockName), LOCK_EX);
             } catch(Exception $e) {
                 Mage::logException($e);
+                throw $e;
             }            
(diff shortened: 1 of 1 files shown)


jfksk:
Shouldn't that catch  ["Error"](https://www.php.net/manual/en/class.error.php) instead of ["Exception"](https://www.php.net/manual/en/class.exception.php) ?
"Error" and "Exception" do implement "[Throwable](https://www.php.net/manual/en/class.throwable.php)" but "Error" does not extend "Exception".

OT (as this is a bit of an hassle to log):
Maybe it would be nice to change 
Mage::logException(Exception $e) 
to 
Mage::logException(Throwable $e)

Flyingmana:
Reopened after talking with @seansan

Will also do a small modification later, so the exceptions get thrown again after the logging, so we do not cause a BC break.

Flyingmana (approved):

PR #1165, opened 2020-08-23
Fix Coupon related cancel order bug if the uses are equal to 0

boesbo:
### Description (*)
If the use of the coupon is equal to 0, the record must be deleted in order not to create SQL errors at the next use of the coupon.

1. Create an order using a coupon (1 use). Cancel the order and try again to place an order with the same COUPON. You will get a SQL error in the "salesrule_coupon_usage" table.


boesbo:
Useless. I saw that it has already been fixed in another way and already merged.

sreichel:
This should be fixed in https://github.com/OpenMage/magento-lts/commit/1b97c6fd5772717484586bc6bd031d850e7e1cb1 (next release!)

The db entry is not deleted, but correctly set to 0.

Issue #1167, opened 2020-08-24
Don't connect to write connection


joshua-bn:
### Summary (*)
In https://github.com/OpenMage/magento-lts/blob/20.0/app/code/core/Mage/Core/Model/Resource/Db/Abstract.php#L332 we are checking if the write connection has any transactions open. To do this, it connects to it. We can just check if there is a connection open first.

### Examples (*)
Place a break point in `Mage_Core_Model_Resource::_newConnection()`

### Proposed solution
Replace that with 
```php
    public function hasConnection(string $connectionName): bool
    {
        return isset($this->_connections[$connectionName]);
    }

    /**
     * Retrieve connection for read data
     *
     * @return Varien_Db_Adapter_Interface
     */
    protected function _getReadAdapter()
    {
        if ($this->hasConnection('write')) {
            $writeAdapter = $this->_getWriteAdapter();
            if ($writeAdapter && $writeAdapter->getTransactionLevel() > 0) {
                // if transaction is started we should use write connection for reading
                return $writeAdapter;
            }
        }

        return $this->_getConnection('read');
    }
```

I use PHP 7.4 locally and type hint so just fix it for your liking. Maybe change to protected to match the other ones, but I think this makes sense as public.

colinmollenhour:
Excellent suggestion! I am not in favor of using type hints at this point for consistency with the rest of the code, though. Would you mind creating a PR?

joshua-bn:
Possibly. I don't have OpenMage forked and don't actually run OpenMage. I just have been copying the relevant changes from here to our codebase (this is my work Github account - BloomNation.com).

So, if I anyone wants to take these changes and create a PR from them before I get around to it, be my guest. 

I haven't actually tested this. I just found this today.

joshua-bn:
btw, tested it locally and created the PR. I haven't tested what effect this has in production - which is really the only place you're going to see a difference in scalability and performance.

colinmollenhour:
Thanks for creating the PR!

I wouldn't expect there to be any noticeable impact on performance even if you've disabled the ill-conceived Mage_Log module and are using a replica for read connections, but it still makes sense not to open connections unless/until they are going to actually be used. I think for the out of the box installation the read/write connections are still shared anyway so for those installations there shouldn't be any difference.

joshua-bn:
[...]

PR #1169, opened 2020-08-26
#1167 - do not connect to write adapter when getting the read adapter

joshua-bn:
### Description (*)
Do not connect to the write adapter when getting the read adapter. We get the write adapter to check if there are any transactions. This causes it to have to connect to the write connection to do so. Most requests don't actually need that write connection.

### Related Pull Requests
None

1. Fixes OpenMage/magento-lts#1167

1. Test if it works - Place a break point in Mage_Core_Model_Resource::_newConnection()
2. Test if the read adapter is still fetched by going to really any page that is not cached
3. Test that transactions work by going to a page that does any writes


fballiano (approved):

fballiano:
this is a great contribution and should be merged ASAP

PR #1183, opened 2020-09-03
Fixes cloning of a collection by also cloning the internal select object

theroch:
### Description (*)
Without this fix, the original select object is used and every modification would also affect the original object. A second effect occurs if you clone a collection and use the load function. All the triggers in _beforeLoad function of the collection which add additional filters etc. will be modify the original select object and the select object will be scale up with identical filters and orders.

Here is an SQL example if you use the CSV export in the sales order grid without this patch:

```
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC LIMIT 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table` 
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC, created_at DESC LIMIT 1000 OFFSET 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table`
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC, created_at DESC, created_at DESC LIMIT 1000 OFFSET 2000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table` 
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC, created_at DESC, created_at DESC, created_at DESC LIMIT 1000 OFFSET 3000
```
For example, the problem is caused by Mage_Adminhtml_Block_Widget_Grid::_exportIterateCollection starting on line 977 in combination with the while loop.

### Related Pull Requests
none

none

1. Open admin panel and go to the sales order grid
2. Use the CSV export function with more than 1000 rows.

#### Result without patch
```
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC LIMIT 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table` 
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC, created_at DESC LIMIT 1000 OFFSET 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table`
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC, created_at DESC, created_at DESC LIMIT 1000 OFFSET 2000
```
The issue affects also the where clause and joins.

#### Result with patch
```
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC LIMIT 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table` 
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC LIMIT 1000 OFFSET 1000
SELECT COUNT(*) FROM `sales_flat_order_grid` AS `main_table`
SELECT `main_table`.* FROM `sales_flat_order_grid` AS `main_table` ORDER BY created_at DESC LIMIT 1000 OFFSET 2000
```
The select object doesn't scale up and the "order by", where and join conditions are not multiplicated.

### Questions or comments

none

### DIFF
--- a/lib/Varien/Data/Collection/Db.php
+++ b/lib/Varien/Data/Collection/Db.php
@@ -854,4 +854,16 @@ class Varien_Data_Collection_Db extends Varien_Data_Collection
 
         return $this;
     }
+
+    /**
+     * Magic clone function
+     *
+     * Clone also Zend_Db_Select
+     *
+     * @return void
+     */
+    public function __clone()
+    {
+        $this->_select = clone $this->_select;
+    }
 }


joshua-bn (approved):

tmotyl:
Hi @theroch 
I generally agree with your change, but please add more info about manual testing - what is happening before the patch and what is happening after the patch. Thanks

fballiano:
This is a very interesting PR, sometimes in the past I've been suggested to use unserialize/serialize to deep clone an object... but that's also not a very elegant way to do it.

I think this PR should be moved forward.

fballiano (approved):

Flyingmana (approved):

midlan:
Looking on the change, seems to me that the same issue applies for `_fetchStmt` property. Calling `Varien_Data_Collection_Db::fetchItem` would move the pointer for both - original and clonecd collection. Or I am wrong?

PR #1181, opened 2020-09-03
Updated lib Net/IDNA2 to latest version

sreichel:
### Description (*)

Updated to latest version to fix last PHP7.4 deprecated warnings ... https://github.com/pear/Net_IDNA2/commits/master

### Related Pull Requests

1. OpenMage/magento-lts#859

### Manual testing scenarios (*)

1. see https://github.com/OpenMage/magento-lts/pull/1173/checks?check_run_id=1048372650

### Questions or comments

- Mage_Eav will be fixed in own PR

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1184, opened 2020-09-03
Add php-74 to static tests

sreichel:
### Description (*)

`composer.json` allows PHP7.0-7.4, so we should also cover 7.4 with static tests.

- should be merged after #1181, #1182
- made 7.4 non-experimental
- added 7.0-7.3 experimental tests, with included lib-directories (PEAR maybe will be removed, not sure about phpseclib, but lib/Zend should be covered)

### DIFF
--- a/.github/workflows/static-code-analyses.yml
+++ b/.github/workflows/static-code-analyses.yml
@@ -12,7 +12,7 @@ jobs:
       max-parallel: 5
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.0', '7.1', '7.2', '7.3']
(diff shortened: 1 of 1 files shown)


kkrieger85 (approved):

Flyingmana (approved):

Flyingmana:
do we really need the checks for all versions of php7?
Maybe only the oldest and newest? Because, each of them takes 5min, and I dont think its likely to have a failure in between them

kkrieger85:
If we'd remove checks, than I'd ensure the latest supported versions are running. See https://www.php.net/supported-versions.php

PR #1198, opened 2020-09-08
Fix integers being cast to decimal in some particular cases

### LINKED ISSUE #1192: Fix integers being cast to decimal in some particular cases
### Preconditions (*)

1. 20.0.2
2. Disable Flat Catalog.

### Steps to reproduce (*)

1. Run the following script
```
<?php

require_once 'app/Mage.php';

Mage::app();

$collection = Mage::getModel('catalog/product')->getCollection()
    ->addAttributeToSelect('tax_class_id')
    ->addAttributeToSort('name');

// toggle this to see the weird result
$collection->addAttributeToSelect('price');

// to make the script faster and lighter
$collection->setPageSize(1);

if ($collection->getSize() > 0) {
    var_dump($collection->getFirstItem()->getData('tax_class_id'));
}

// does not help, the SQL is exactly the same
// echo (string) $collection->getSelect();
```
2.

### Expected result (*)

1. It should give you the tax class id as an integer, as declared for this attribute type.
2.

### Actual result (*)

1. It will give you the tax class id as a float.
2.

### Extra information

I spent a lot of time debugging this issue, you can find a detailed thread about it here: https://magento.stackexchange.com/questions/105244/magento-1-admin-product-grid-column-value-disappear-on-name-search/105421

So I already have a fix for it, will do a PR soon

digitalpianism:
### Description (*)

This provides a fix for the issue #1192 that I have created.
Details about the fix can be found here: https://magento.stackexchange.com/questions/105244/magento-1-admin-product-grid-column-value-disappear-on-name-search/105421

1. Fixes OpenMage/magento-lts#1192

1. Run the following script with catalog product flat disabled
```<?php

require_once 'app/Mage.php';

Mage::app();

$collection = Mage::getModel('catalog/product')->getCollection()
    ->addAttributeToSelect('tax_class_id')
    ->addAttributeToSort('name');

// toggle this to see the weird result
$collection->addAttributeToSelect('price');

// to make the script faster and lighter
$collection->setPageSize(1);

if ($collection->getSize() > 0) {
    var_dump($collection->getFirstItem()->getData('tax_class_id'));
}
```
2. Notice the bug.
3. Apply the fix.
4. Run the script again.


sreichel:
This fix does not work for me.

I use 1.9.4.x-dev with sample data:

- with flat tables ON your test-script returns `int(2)`
- with flat tables OFF before fix `string(6) "0.0000"`
- with flat tables ON after fix `int(0)`

sreichel:
This seems not fix the problem ... all tests with flat product tables OFF

```
<?php

require_once 'app/Mage.php';

Mage::app();

$collection = Mage::getModel('catalog/product')->getCollection();
$collection->addAttributeToSelect('tax_class_id');
$collection->addAttributeToSort('name');

// toggle this to see the weird result
$collection->addAttributeToSelect('price');

// to make the script faster and lighter
$collection->setPageSize(1);

if ($collection->getSize() > 0) {
    var_dump($collection->getFirstItem()->getData('tax_class_id'));
}
```

Correct output should be 2

- output (as it is): `string(6) "0.0000"`
- comment `addAttributeToSort('name')` turns it to `string(6) "2.0000"`
- comment `addAttributeToSelect('price')` returns correctly `int(2)`

We should also check `addAttributeToSort()` ...

digitalpianism:
@sreichel I've just tried again on my install and it works fine. Remember I'm running 20.0.2 so maybe something changed along the way if you're using a newer version. This bug is definitely still present in the last official Magento 1 release.

Only one product created, flat tables off.

First run without the fix gives me "2.000", then after applying the fix it gives me "2".

sreichel:
@digitalpianism tested again with 19.4.x ... 

(Can you please check if output is okay with `$test->testAddAttributeToSort();` under 20.2?)

- before fix
  - php7.4
  ```
  string(1) "2"
  string(1) "0"
  string(6) "2.0000"
  ```
  - php8.1
  ```
  int(2)
  int(0)
  string(6) "2.0000"
  ```

- after fix
  - php7.4
  ```
  string(1) "2"
  string(1) "0"
  string(1) "2"
  ```
  - php8.1
  ```
  int(2)
  int(0)
  int(2)
  ```

sreichel (approved):

sreichel:
Tested you example from stackexchange and it works.

... but something is still wrong with this testscript. And we should check why type changes from php7 to 8 ...

app/code/core/Mage/Eav/Model/Resource/Helper/Mysql4.php
```diff
@@ -108,7 +108,7 @@ public function getLoadAttributesSelectGroups($selects)
         foreach ($selects as $eavType => $selectGroup) {
             $mainGroup = array_merge($mainGroup, $selectGroup);
         }
-        return [$mainGroup];
+        return $mainGroup;
```
Flyingmana, line 111: this seems like a strong behavior change

digitalpianism:
It is a strong change but the code using this method has been changed accordingly

Raphael Petrini

sreichel:
Changed return type seem to come from here ...

https://www.php.net/manual/en/migration81.incompatible.php#migration81.incompatible.pdo.mysql

> MySQL Driver
Integers and floats in result sets will now be returned using native PHP types instead of strings when using emulated prepared statements. This matches the behavior of native prepared statements. The previous behaviour can be restored by enabling the PDO::ATTR_STRINGIFY_FETCHES option.

kiatng (approved):
Tested.

luigifab:
We have a bug with theses changes. I don't understand why.

We have an attribute, text type, id 1276.

On product edit page from OpenMage backend for store 1, I write "abc" and I save, the database is updated (see bellow), but in backend, the default value (123) is displayed (the checkbox _use default value_ is not checked).

image

--

When I revert this PR, if I refresh backend page, the good value (abc) is displayed.

--

Now, if I didn't revert the PR, I can remove the line id 888745 in `catalog_product_entity_text`, then, when I save "def" from backend, the database is updated, and the good value (def) is displayed.

Now, I update the new line id with the old line id (888745), I refresh backend page, the default value (123) is displayed.

Now, I revert this PR, I refresh backend page, the good value (def) is displayed.

sreichel:
Is it a MySQL8 DB?

PR #1199, opened 2020-09-08
Properly set timestamp for catalogrule on non GMT install Fix #1191

### LINKED ISSUE #1191: Fix timestamp not set properly for catalogrule on non GMT install.
### Preconditions (*)

1. 20.0.2
2. GMT+5 install

### Steps to reproduce (*)

1. Create a catalog rule that applies 15% off all products
2.

### Expected result (*)

1. Works fine.
2.

### Actual result (*)

1. Stops working after midnight.
2.

### Extra information

When Magento displays products on frontend, it checks if there are catalog rules for this date. And the date used for this check is our local, so in our case GMT+5. However, when catalog rules are being applied, it uses GMT date. So that means that we aren't able to apply rules until 5 AM.

Source / fix: https://stackoverflow.com/questions/25280095/magento-catalog-price-rule-disappears-at-night

Fix is ready, will do a PR soon.

digitalpianism:
### Description (*)

Fix the issue #1191 

1. Fixes OpenMage/magento-lts#1191

1. Setup the testing procedure from the issue
2. Apply the fix
3. Enjoy the rule still working after midnight.

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
+++ b/app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
@@ -116,9 +116,7 @@ class Mage_CatalogRule_Model_Action_Index_Refresh
     {
         $this->_app->dispatchEvent('catalogrule_before_apply', array('resource' => $this->_resource));
 
-        /** @var Mage_Core_Model_Date $coreDate */
-        $coreDate  = $this->_factory->getModel('core/date');
-        $timestamp = Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP);
+        $timestamp = Mage::app()->getLocale()->storeTimeStamp();
 
         foreach ($this->_app->getWebsites(false) as $website) {
             /** @var Mage_Core_Model_Website $website */
--- a/app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
+++ b/app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
@@ -118,7 +118,7 @@ class Mage_CatalogRule_Model_Action_Index_Refresh
 
         /** @var Mage_Core_Model_Date $coreDate */
         $coreDate  = $this->_factory->getModel('core/date');
-        $timestamp = $coreDate->gmtTimestamp('Today');
+        $timestamp = Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP);
 
         foreach ($this->_app->getWebsites(false) as $website) {
             /** @var Mage_Core_Model_Website $website */


app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
```diff
@@ -118,7 +118,7 @@ public function execute()
 
         /** @var Mage_Core_Model_Date $coreDate */
         $coreDate  = $this->_factory->getModel('core/date');
-        $timestamp = $coreDate->gmtTimestamp('Today');
+        $timestamp = Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP);
```
kiatng, line 121: Code wise, `$timestamp = Mage::app()->getLocale()-> storeTimeStamp();` is may be clearer.  Also lines 119 and 120 should be removed.

ADDISON74:
This should be merged.

colinmollenhour (approved):
@digitalpianism Please confirm the new version as suggest by @kiatng still works as excpected.

app/code/core/Mage/CatalogRule/Model/Action/Index/Refresh.php
```diff
@@ -116,9 +116,7 @@ public function execute()
     {
         $this->_app->dispatchEvent('catalogrule_before_apply', array('resource' => $this->_resource));
 
-        /** @var Mage_Core_Model_Date $coreDate */
```
kiatng, line 119: My bad, it's incorrect to use `storeTimeStamp()`, `$timestamp` must be GMT (same timezone as in the DB)  and not based on local timezone. So it should be:

```suggestion
        /** @var Mage_Core_Model_Date $coreDate */
        $coreDate  = $this->_factory->getModel('core/date');
        $timestamp = $coreDate->gmtTimestamp();
```

The above `$timestamp` has the same value as the original PR `Mage::app()->getLocale()->date(null, null, null, true)->get(Zend_Date::TIMESTAMP)`.

The original `$timestamp = $coreDate->gmtTimestamp('Today');` returns today's date at 0 hours and not the current time. See https://github.com/OpenMage/magento-lts/blob/cc3eff3ebec46efbcab7ae4c26dd8a754476da18/app/code/core/Mage/Core/Model/Date.php#L142-L150

colinmollenhour: I love how Magento has 99 ways to get a timestamp.. Isn't the default timezone already UTC? If so is that not the same as `time()`?

If it just uses the current timestamp is the result of the index refresh dependent on the time of day that it is run? Sorry, I haven't dug into this issue deeply enough to actually understand what the bug is and what the fix should be..

ADDISON74: Related to this commit I discovered issues with Magento installations, but I did not check in OM yet. Even I configured correctly the time zones in PHP and Magento, always the time is not correct in frontend, logs. For example, the special price period should end at 00:00 but it ends at 2 am. In logs when there are errors/warnings the time is not correct. There is a 2 - 3 hours difference related to normal time. When I check webserver logs I need to make a conversion in my mind.

fballiano:
Hi @digitalpianism, could you check @kiatng's comments so that we may go on try to merge it? Thank you!

fballiano:
this was already fixed lately

PR #1201, opened 2020-09-08
Fix wrong payment info template paths

mdlonline:
# Description (*)
 In some Payment block classes the 'toPdf' function sets a wrong template path. These templates are present in another folder.

1. Create order/invoice
2. Make sure the email is being sent with pdf attached
3. Check if payment info is added to attached pdf, and no logs about invalid template files are created.

### DIFF
--- a/app/code/core/Mage/Payment/Block/Info.php
+++ b/app/code/core/Mage/Payment/Block/Info.php
@@ -74,7 +74,7 @@ class Mage_Payment_Block_Info extends Mage_Core_Block_Template
      */
     public function toPdf()
     {
-        $this->setTemplate('payment/info/pdf/default.phtml');
+        $this->setTemplate('payment/info/default.phtml');
         return $this->toHtml();
     }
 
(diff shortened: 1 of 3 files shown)


kkrieger85:
We could discuss if this was intended to have the possibility to use different layout styles. 

But defining a default existing template is better than having no template for pdf

kkrieger85 (approved):
LGTM

sreichel (approved):
Tested.

rfeese:
I think this change caused a regression in rendering payment information as PDF. This change causes the toPdf methods to use the HTML templates instead of the PDF templates. This causes invalid PDF rendering for the payment methods. 

The PDF template files *do* exist:
https://github.com/OpenMage/magento-lts/tree/1.9.4.x/app/design/adminhtml/default/default/template/payment/info/pdf

As far as I know, there is no option for including a PDF attachment on invoice emails. Was the submitter using an add-on module for email attachments? Perhaps there is a bug in the add-on?

PR #1207, opened 2020-09-13
bugfix: don't cast min_sale_qty to int as it can be below 1

mageconsult:
### Description (*)
With Magento 1.9.4.2 there was a small change in 
app/design/adminhtml/default/default/template/catalog/product/tab/inventory.phtml.

When selling products with is_qty_decimal set to true, it is possible to sell units lower than one.

Here an example from an actual product in frontend
image

Backend when creating product
image

When creating a product in backend, this will first be save correctly. 
image

With next edit it will be casted to integer and the decimal part will be cutted of. A min_sale_qty with 0.25 for example will be casted to 0.
image

With this fix the old version will be introduced again.

1. enable is_qty_decimal in product
2. enter qty of 1.5
3. set min_sale_qty to 0.25
4. save and continue edit
5. min_sale_qty now is still 0.25 (instead of 0)

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/tab/inventory.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/tab/inventory.phtml
@@ -77,7 +77,7 @@
 
         <tr>
             <td class="label"><label for="inventory_min_sale_qty"><?php echo Mage::helper('catalog')->__('Minimum Qty Allowed in Shopping Cart') ?></label></td>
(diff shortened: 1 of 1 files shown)


colinmollenhour (approved):

kkrieger85 (approved):

joshua-bn:
Probably would've been better to cast it to a float instead of nothing. Now it could be a non-scalar value or a non-numeric string.

Issue #571, opened 2018-11-26
introduce a /pub/ directory analog to Magento2


Flyingmana:
Webapps nowdays have for mostly security reasons most of the files not as part of the webroot anymore. They use different names like `pub`, `www`,`public`, `htdocs`. 

We should make it easy to do the same.

colinmollenhour:
Totally agreed, although this would definitely break BC for anyone trying to do a `git pull`.

Flyingmana:
Not necessarily. Magento 2 did for quite a while support both on parallel?(not sure if they still do)

avstudnitz:
They do.

edannenberg:
I'm probably missing something here, but isn't that kinda pointless for Magento 1?

speedupmate:
its not pointless it hides 99% of your filesystem mishaps and lessens attack surface a lot cause you only make public what needs to be public 

take a following structure : 

> magento_base/
> htdocs/
> htdocs/website_01/
> htdocs/website_01/index.php
> htdocs/website_01/.htaccess <-- symlink or real file
> htdocs/website_01/js/ <--symlink
> htdocs/website_01/skin/ <--symlink
> htdocs/website_01/media/ <--symlink
> htdocs/website_01/errors/ <--symlink
> htdocs/website_02/
> repeat for any other site  you might have --"-- 
> 

or a  following one

> magento_base/
> magento_base/pub/
> magento_base/pub/website_01/
> magento_base/pub/website_01/index.php
> magento_base/pub/website_01/.htaccess <-- symlink or real file
> magento_base/pub/website_01/js/ <--symlink
> magento_base/pub/website_01/skin/ <--symlink
> magento_base/pub/website_01/media/ <--symlink
> magento_base/pub/website_01/errors/ <--symlink
> magento_base/pub/website_02/
> repeat for any other site  you might have --"-- 
>

benefits are: 
* lessen attack surface on filesystem configuration errors (wrong permissions , misconfigured .htacccess etc)
* lessen the attack surface by controlling file access trough webserver configuration (you can't access what is not public) 
* site configurations can be easily versioned in place as those are now real files in filesystem
* probably more as code is separated from presentation

colinmollenhour:
[...]

PR #1210, opened 2020-09-16
Introduced pub/ structure for webserver home folder mapping

speedupmate:
introducing pub/{n}_website folder structure to serve public assets while hiding away magento_root from public areas.

### Description (*)

* lessen attack surface by only serving what is public 
* allows version controlled management of any sites (including test and staging sites) configuration (if different)
* separates public folder from core code so core can be updated without needing to merge .htaccess , index.php changes etc as those would be separated

1. Fixes OpenMage/magento-lts#571

1. set your webserver config to point to a new home folder under pub/{x} 
2. alternatively symlink your webserver  home folder to desired location under pub/{x} 


colinmollenhour:
Ahh, I've always wished the repo root was not the web root. That's a clever way to avoid renaming every file in the repo. I like it! 👍

fballiano:
Questions:
- Is this compatible with the installation via composer?
- are those directories symlinks?
- I think we need to update the .htaccess in this PR cause the main one was changed a lot

speedupmate:
> Questions:
> 
> * Is this compatible with the installation via composer?
> * are those directories symlinks?
> * I think we need to update the .htaccess in this PR cause the main one was changed a lot

* I have no idea , it probably is compatible since composer only deals with filesystem
* yes those are symlinks , can be apache nginx directives/aliases also but thats more complicated 
* actually .htaccess can also be a symlink by default and whoever needs to edit this can swap it out for real one

fballiano:
this PR is tricky nowadays, htaccess and index.php are very different and we've to rebase it to main or next branch :-\

pub/default/.htaccess
```diff
@@ -0,0 +1,234 @@
+############################################
```
fballiano, line 1: I'd prefer to symlink the .htaccess if possibile, it allows for less duplication on the repo

pub/default/index.php
```diff
@@ -0,0 +1,78 @@
...
+ * http://opensource.org/licenses/osl-3.0.php
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@magento.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade Magento to newer
+ * versions in the future. If you wish to customize Magento for your
+ * needs please refer to http://www.magento.com for more information.
```
fballiano, line 19:
```suggestion
 * OpenMage
 *
 * This source file is subject to the Open Software License (OSL 3.0)
 * that is bundled with this package in the file LICENSE.txt.
 * It is also available at https://opensource.org/license/osl-3-0-php
```

fballiano:
I've a similar structure for some customers of mine, and my index.php looks like this:

```php
<?php
chdir("..");
require "index.php";
```

wouldn't it work for you? I'd try to avoid duplicate most of the code that's in this version of the index.php file

ADDISON74:
I think this PR can be closed. Moving some directories does not bring any improvement, on the contrary, it complicates a structure that has been working for more than 15 years. In addition, there are files that will create a conflict with what already exists in the repository.

colinmollenhour:
@ADDISON74 
> ... it complicates a structure that has been working for more than 15 years. ...

I think this PR has it's merits. Every other web project I've ever seen has a separate pub directory so that every project file is not in the root. M2, Laravel, Rails, etc.. I've never seen another one that relies on `.htaccess` files sprinkled around to "secure" everything.

You know what happens if you are migrating files from one directory to another and you forget to copy or delete a .htaccess file? Boom, your files are wide open! Ask me how I know.. :(

Also every time someone adds a new file to the repo it is a risk that it isn't protected.. Example:

https://demo.openmage.org/LICENSE.html

I would actually be in favor of *moving* index.php into pub for the next major version and making this mandatory. It's worth the effort to reorganize the code to expose only what you want, but probably very few do. In my opinion the project should set an example for a good deployment, not encourage bad practice. Yes, I know, kinda late to be tackling this now... :D

colinmollenhour (approved):

dev/openmage/apache/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM ghcr.io/colinmollenhour/docker-openmage-dev:8.2-apache
+
+ENV APACHE_DOCUMENT_ROOT /var/www/html/pub/default
+
+RUN sed -ri -e 's!/var/www/html!${APACHE_DOCUMENT_ROOT}!g' /etc/apache2/sites-available/*.conf
+RUN sed -ri -e 's!/var/www/!${APACHE_DOCUMENT_ROOT}!g' /etc/apache2/apache2.conf /etc/apache2/conf-available/*.conf
```
fballiano, line 6: are the changes to dev/openmage intended for this PR?

ADDISON74:
@colinmollenhour - The argument you offered is enough to find PR utility, but I have never encountered the mentioned problems and I hope this never happens. PR still needs polish and documentation. It was proposed in main branch.

dev/openmage/apache/Dockerfile
colinmollenhour: Yes, this updates dev/openmage to use the new root directory by default.

fballiano (approved):

colinmollenhour:
@ADDISON74

> PR still needs polish and documentation.
Please provide specifics regarding polish. Where do you propose this should be documented?

> It was proposed in main branch.
As implemented currently this actually doesn't break BC in any way so I think this could be added to "main". If there is support for breaking BC on purpose then it should definitely be in "next" but I think that will be too controversial..

I see @fballiano just merged while I was typing. 😄 I'm happy to add docs for this, just let me know where. It can be improved upon if any issues are found but I currently do not know of any issues with it.

fballiano:
when my question was answered I was ok with that, we can add docs in any moment ;-)

fballiano:
and i was also sure it's not BC

fballiano:
sadly this PR doesn't work when the project is installed via composer if deploy strategy is "copy"

Screenshot 2023-06-15 alle 11 31 27

can we do something about that?

speedupmate:
@fballiano you can pro provide a post script that inspects the copy strategy and fixes symlinks if those fail to be copied and need to be recreated

ADDISON74:
Shouldn't we include steps how to use this feature in OM?

Also, is there any solution when OM is used together with composer as @fballiano reported above?

fballiano:
Since the "pub" folder is totally incompatible with composer installation, which is for use the best way to install/upgrade/manage OpenMage, I'm not even sure this folder should be in our repo, because it's simply broken now (with composer)

ADDISON74:
Shall we revert this PR?

fballiano:
This PR has some other things not really related to the "pub/" structure so I'd not revert but I'd evaluate the removal of just the pub/ folder, converting it to a documentation on how to create it...

colinmollenhour:
Perhaps the composer modules which deploy the Magento repo could be fixed to support copying the symlinks? Which one did you use, @fballiano ?

fballiano:
@colinmollenhour I use "aydin-hassan/magento-core-composer-installer", both for php7 and php8 (depending on the project)

fballiano:
I think we could just introduce an exclude for the pub directoy

```
         "magento-core-deploy" : {
            "excludes": [ 
                "pub"
            ]
         }
```

the feature is still good and it's good to have it in the vendor, but deploying it is not very usefull cause it will have to be customized anyway (per website) so if we don't deploy it at all we've no problem.

ma4nn:
perhaps the `magento-core-deploy` config should be mentioned somewhere in the [installation notes](https://github.com/OpenMage/magento-lts#composer), right?

kiatng:
Ref https://github.com/OpenMage/magento-lts/blob/611bbe7ba8b1087452d18e423a93d0eb0ae52f02/errors/processor.php#L113-L116

Line 113 ` $this->_prepareConfig(); ` above load the skin defined in `errors/local.xml`, which will be overwritten in line 115 `$this->_setSkin($_SERVER['MAGE_ERRORS_SKIN'] ?? $_GET['skin']);`. 

For anyone who is wondering why the error page is rendering the default skin and not the expected skin, remove the following lines:
https://github.com/OpenMage/magento-lts/blob/611bbe7ba8b1087452d18e423a93d0eb0ae52f02/pub/default/index.php#L20-L23

in your `pub/{store_or_website_name}/index.php`.

sreichel:
@kiatng is this still valid https://github.com/OpenMage/magento-lts/blob/main/docs/ERROR_PAGE.md ?

kiatng:
`ERROR_PAGE.md` is valid if `$_SERVER['MAGE_ERRORS_SKIN']` is not set. I'm not sure why this PR introduced the new param `MAGE_ERRORS_SKIN`, it seems unnecessary to me.

PR #1213, opened 2020-09-17
Fix strpos with non-string needle

luigifab:
### Description

This fix: _Deprecated functionality: strpos(): Non-string needles will be interpreted as strings in the future. Use an explicit chr() call to preserve the current behavior in .../app/code/core/Mage/Eav/Model/Entity/Increment/Numeric.php on line 48_

OpenMage 20.0.3 / PHP 7.4.9

### Manual testing scenarios

1. Create a new website with a new store view in backend.
2. Go to frontend on this new store view.
3. Add a product to cart, go to checkout, and try to make the order.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Increment/Numeric.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Increment/Numeric.php
@@ -43,7 +43,9 @@ class Mage_Eav_Model_Entity_Increment_Numeric extends Mage_Eav_Model_Entity_Incr
     {
         $last = $this->getLastId();
 
-        if (strpos($last, $this->getPrefix()) === 0) {
+        if (empty($last)) {
+            $last = 0;
+        } else if (strpos($last, $this->getPrefix()) === 0) {
             $last = (int)substr($last, strlen($this->getPrefix()));
         } else {
             $last = (int)$last;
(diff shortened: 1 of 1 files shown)


kiatng:
Question: The needle in this case is `$this->getPrefix()`,  in the [official doc](https://www.php.net/manual/en/function.strpos.php):

> If needle is not a string, it is converted to an integer and applied as the ordinal value of a character. This behavior is deprecated as of PHP 7.3.0, and relying on it is highly discouraged. Depending on the intended behavior, the needle should either be explicitly cast to string, or an explicit call to chr() should be performed.

So, shouldn't the fix be `strpos($last, (string) $this->getPrefix()`?

luigifab:
@kiatng hum, when you are creating the first order on a new store view, `$this->getPrefix()` is empty (null?).

kiatng:
@luigifab I was thinking in a more general case: `if (strpos($last, (string) $this->getPrefix()) === 0) {` will work for not just when `$last` is `null` but also when `$this->getPrefix()` returns a non-string.

luigifab:
oh, yes, I can change with `(string)`

Flyingmana:
I added a testcase for this, and validated this PR will fix the Issue.

Maybe someone can extend this with a few more tests with real world variables, especially for the case of a not set prefix.

https://github.com/OpenMage/Testfield/blob/main/tests/Unit/Regression/Lts/Ticket1213Test.php

Issue #1219, opened 2020-09-21
Background color configuration tabs new OpenMage admin theme


AlterWeb:
### Preconditions (*)

1. OpenMage 19.4.7

### Steps to reproduce (*)

1. Install PostNL module
2. Go to configuration and look at the PostNL menu item

### Expected result (*)

1. Menu item readable:
![Screenshot from 2020-09-21 11-32-12](https://user-images.githubusercontent.com/12827587/93752100-7156cb80-fbfe-11ea-81c6-a4a5d78441d1.png)

### Actual result (*)

1. Menu item is not readable because background color of `ul.tabs dl`:
![Screenshot from 2020-09-21 11-35-01](https://user-images.githubusercontent.com/12827587/93752152-84699b80-fbfe-11ea-9e8c-b57f3b6a2438.png)

### Solution
I think the fix is as simple as setting the background color of `ul.tabs dl` to white, I don't know for sure if there is a reason it is set to #68666f but I can't think of one.

aterjung:
my last pull request solves this issue. I haven seen any side-effects until now.

sreichel:
Fixed in #1224

sreichel:
@all-contributors add @AlterWeb bug

PR #1223, opened 2020-09-22
Fix config menu css from Isse #1219

aterjung:
### Description
fix openmage admin theme css to fix error #1219 with at least PostNL Module.

BG Color of menu was dark and only overlaying a elements has white background. Custom-Menu Elements without link elements were shown with wrong background color then.

1. Fixes OpenMage/magento-lts#1219


aterjung:
sorry, messed up the pull request...

Issue #1225, opened 2020-09-23
Custom Adminhtml theme doesn't work after setup of Magento LTS


Arhimenrius:
### Summary (*)
After installing Magento LTS (replacing current Magento-Mirror version), I have noticed that our custom adminhtml theme started to return blank page.

After further investigation, I have found out that the reason is in enforced "openmage" theme. And looks like it is enforced for every single component of layout separately (default, template, skin, locale, layout).

It's partially "fine", as it can be overwritten by enforcing own theme for every single component, like: 
` <admin>
            <design>
                <package>
                    <name>mycustompackage</name>
                </package>
                <theme>
                    <default>default</default>
                    <template>default</template>
                    <skin>default</skin>
                    <locale>default</locale>
                    <layout>default</layout>
                </theme>
            </design>
        </admin>
`

However, I have investigated it deeper and found out that, there is setting "admin/design/use_legacy_theme", which should be enough to just set to 1, and it should fall back to default way. Unfortunately, looks like somebody hardcoded it to "0" inside Mage/Adminhtml config.xml file, causing that it cannot be set in local.xml.

### Proposed solution
3 proposals: 
1. Modify setting to be able to be read from local.xml, not only from custom module
2. Set "use_legacy_theme" to 1 by default
3. Don't overwrite default approach and try to promote "openmage" theme in another way.

aterjung:
could you show me some parts of your custom theme, especially naming, folders, config.xml? I can't imagine how this can leed to a white page. The openmage theme mainly modifies css - that should not break everything. The "legacy"  switch can be found in system config and saves its value to core_config_data table. local.xml is part of a themes layout definiton and can not set config values.

It would help me to get some details about your custom theme. We are evaluating some possible changes to the openmage skin structure and switcher. Probably you can test these changes with your installation?

kiatng:
As noted in issue #1164, there is an issue with custom template rewrite. I still think we need proper implementation of infinite theme fallback, see[ Alan Storm's article](https://alanstorm.com/magento_infinite_fallback_theme_xml/). That article applies to frontend, but I think it's the same for backend.  My suggestion on doing it is outlined in issue #1216, which breaks something. I will investigate more when I have the time.

rvelhote:
Setting this `use_legacy_theme` to `1` fixed an issue we had with the calendar losing its styles (in this case we did it via `System » Configuration » Admin » Enable legacy Admin Theme`).

[...]

Issue #1216, opened 2020-09-17
Admin theme changed by default config value


AlterWeb:
### Preconditions (*)

1. OpenMage 19.4.7

### Steps to reproduce (*)

1. Change config value of design/theme/template and design/theme/skin to a random theme under the default config

### Expected result (*)

1. All stores use this value for there skin unless specified different
2. The admin theme should be openmage

![Screenshot from 2020-09-17 18-17-14](https://user-images.githubusercontent.com/12827587/93498421-0eb9b300-f912-11ea-8139-90da3b55d7e0.png)

### Actual result (*)

1. All stores use this value for there skin unless specified different
2. The default Magento admin theme is used because it can not find the random theme (unless this is "openmage" of course)

![Screenshot from 2020-09-17 18-16-15](https://user-images.githubusercontent.com/12827587/93498315-ee89f400-f911-11ea-9ebd-032da7367832.png)

aterjung:
Thank you for writing this bug report! There is something wrong in the theme switcher code. I will fix this as soon as possible. As long as this is fixed and included in the next release, you could just move the skin overwrite to the website level.

AlterWeb:
@aterjung Thank you for the quick response. I think this issue is not new but we never noticed it because the (old) default theme it is falling back to was the expected behavior. If I remember correctly from debugging it, the code responsable for this behavior was not edited. You even may say that it's a feature, but if it is than it is a confusing one ;-). Thank you for noticing that workaround, we already did do exactly that for now. So no big problem for now but it makes the configuration a bit more confusing.

aterjung:
@AlterWeb where is the code you identified as source for this error? I am a little bit confused at the moment, my dev-environment where the theme switcher was developed does not show this behavior! But all of my customers live systems do have this error.

AlterWeb:
I don't have access to my workstation at the moment (typing this on my phone) but if I remember correctly it was in the class Mage_Adminhtml_Controller_Action. Around line 163 there is a foreach loop setting those values. I can check this tomorrow.

AlterWeb:
@aterjung I just checked it and this is indeed where those values are set. So this code:

```
foreach (array('layout', 'template', 'skin', 'locale') as $type) {
    if ($value = (string)Mage::getConfig()->getNode("stores/admin/design/theme/{$type}")) {
        Mage::getDesign()->setTheme($type, $value);
    }
}
```

kiatng:
[...]

Issue #1164, opened 2020-08-23
New admin theme vs custom themes?


sreichel:
### Summary (*)

I've tested the admin theme (#1008) only with vanilla LTS and it worked like a charm.  After updating some shops to latest LTS i had to realize that **all** shops made changes to admin templates. Instead of changing default templates, they added a new theme that is based on default ... (seems correct to me)

```
    <stores>
        <admin>
            <design>
                <package>
                    <name>default</name>
                </package>
                <theme>
                    <default>default</default>
                    <template>some_name</template>
                </theme>
            </design>
        </admin>
    </stores>
```

So "openmage admin theme" is never loaded ... 

- it shows default login page (default templates loaded)
- only loads default css 

### Examples (*)

See above.

### Proposed solution

- add "openmage" to fallback logic? NO! (???)

or

- replace default admin theme with  #1008
- remove admin theme switcher

It is just CSS and should break nothing.

aterjung:
ok, this behavior is intended to be shure custom themes are not affected. The shops i am supporting all use the default theme so i did not see this as a problem. Did you check if the themes would break if you change the base theme to openmage?

sreichel:
> Did you check if the themes would break if you change the base theme to openmage?

Yes, i did. It shows openmage theme, but the storeswitcher does not work anymore.

sreichel:
Do we still need the `default` theme, or can we (@aterjung) replace it? Actually it does not work with common (?) template rewrites.

Flyingmana:
We can not replace it for backwards compatibility reasons, we could just copy it, which would increase the needed maintenance

aterjung:
I am not shure if I understand the Problem! If there is a custom theme, isn’t it right to ignore the openmage theme?
Is the Problem that the theme switch dose not switch to openmage theme then? Is this really a problem?




Diplom Wirtschaftsinformatiker (FH)

kiatng:
@aterjung An example to the problem: I have a custom template in `app\design\adminhtml\default\ihe\template\sales\order\view\history.phtml`, where I set my custom theme called `ihe`, where I added the admin user name to show which comment is made by who. Now, if I want to use the `openmage` theme, how do I fallback to it?

As suggested by @sreichel, one way is to replace the `default` theme with `openmage`. 

Another way is make a new package for `openmage`, I actually prefer this, for infinite theme fallback.

PR #1229, opened 2020-09-24
Openmage Designpackage and new admin theme overwrites

aterjung:
### Description (*)
This is a draft PR for the conversion of the openmage theme to an own package and and all related changes to the default theme to make it "infinite theme fallback" compatible. Additionally i added an own admin theme overwrite section to the backend config in Advanced/Admin/Admin Theme.
Please read issue #1216 for more details.  

### Fixed Issues
1. Fixes OpenMage/magento-lts#1216
2. Fixes OpenMage/magento-lts#1225
3. Fixes OpenMage/magento-lts#1164

Install branch and test it in diffenrent conditions.

### Questions or comments
1. Is this the right way to go?
2. Should the admin theme overwrites section be more failsafe?
3. Are Overwrite-Options for layout, templates,skin and locales realy needed for backend themes? There are no different storeviews with slightly different themes, you always have one backend theme.


app/code/core/Mage/Adminhtml/etc/config.xml
```diff
@@ -236,23 +236,10 @@
         </help>
         <admin>
             <design>
```
kiatng, line 238: `<design>` should be `<theme>` . See [system,xml](https://github.com/OpenMage/magento-lts/blob/916adc7a46c704628c61ce793ed1cde24b649be8/app/code/core/Mage/Core/etc/system.xml#L1061).

kiatng:
The following in [Action.php](https://github.com/OpenMage/magento-lts/blob/916adc7a46c704628c61ce793ed1cde24b649be8/app/code/core/Mage/Adminhtml/Controller/Action.php#L155) breaks BC

```php
    public function preDispatch()
    {
        // override admin store design settings via admin config section
        Mage::getDesign()
            ->setArea($this->_currentArea)
            ->setPackageName((string)Mage::getConfig()->getNode('stores/admin/admin/theme/package'))
            ->setTheme((string)Mage::getConfig()->getNode('stores/admin/admin/theme/default'))
        ;
        foreach (array('layout', 'template', 'skin', 'locale') as $type) {
            if ($value = (string)Mage::getConfig()->getNode("stores/admin/admin/theme/{$type}")) {
                Mage::getDesign()->setTheme($type, $value);
            }
        }
```

Traditionally, theme overwrite is setting the package/theme in config.xml like this:

```xml
<config>
    <stores>
        <admin>
            <design>
                <package>
                    <name>default</name>
                </package>
                <theme>
                    <default>emgsori</default>
                </theme>
            </design>
        </admin>
    </stores>
    ...
</config>
```

Which will be ignored totally. Is there a way to get around this?

aterjung:
yes, the theme overwrite is a BC in the way it is done now. This is what i think have to discussed - the way it was implementen in magento since 1.4 is a little bit strange. I work with M1 sind 2009 and never noticed that you can change the backend theme by setting the "default" value in the Design tab in Backend. I think everybody i know use this filed as default for the frontend theme.
This is the change in M1 1.4 that contains a comment showing that this was a intended decision: 
image

They decided to simply overwrite the package via config.xml. For me it seems this was forgotten right after that or it was only used for Magento Go probably - but it is not mentioned anywhere, and its not documented in the M1 Documentation.

I think the backend theme thing was not done with much love in these days - we could repair this now and add a OpenMage Theme package - then we need a way to make these BC changes work for everyone - or we keep the old code and replace the default theme.

A possible solution could be to make a switch to let the admin chose if he would like the package read from stores/admin/design/package and therefore from config.xml or from stores/admin/admin/theme/package?
On top of that - is it really needed to have the layout, theme, locale, skin inputs for the admin theme? I would prefer two dropdowns like this: 
admin_themeswitch_mockup

and if you select "legacy" everything is handled like before.

kiatng:
For BC, we just need one dropdown for selecting legacy or openmage theme. I tested the following in [Action.php](https://github.com/OpenMage/magento-lts/blob/916adc7a46c704628c61ce793ed1cde24b649be8/app/code/core/Mage/Adminhtml/Controller/Action.php#L155):

```php
    public function preDispatch()
    {
        // User configurable setting in backend > System > Configuration > Admin > Admin Theme
        // To override openmage, create a custom package and fallback to openmage by theme.xml as in frontend.
        $switch = Mage::getStoreConfig('admin/theme/package') == 'openmage'
            ? 'admin' // Use openmage package.
            : 'design'; // Use the legacy base package for BC.

        // override admin store design settings via admin config section
        Mage::getDesign()
            ->setArea($this->_currentArea)
            ->setPackageName((string)Mage::getConfig()->getNode("stores/admin/$switch/theme/package"))
            ->setTheme((string)Mage::getConfig()->getNode("stores/admin/$switch/theme/default"))
        ;
        foreach (array('layout', 'template', 'skin', 'locale') as $type) {
            if ($value = (string)Mage::getConfig()->getNode("stores/admin/admin/theme/{$type}")) {
                Mage::getDesign()->setTheme($type, $value);
            }
        }
```

We can keep the updates on layout, template, skin, locale as is. 

If we use a `source_model` like:

```php
    {
        return array(
            ['value' => 'admin', 'label'=>Mage::helper('adminhtml')->__('OpenMage theme')],
            ['value' => 'design', 'label'=>Mage::helper('adminhtml')->__('Legacy theme for backward compatibilty')],
        );
    }
```
we can simplify `$switch = Mage::getStoreConfig('admin/theme/package');`. To reduce surprises, the default should be set to legacy theme:

```xml
        <admin>
            <theme>
                <package>design</package>
            </theme>
        </admin>
```

I tested BC for `default/custom_theme`, with the override files localed in `app/design/adminhtml/default/custom_theme`, and it works.

In this way, stores with custom theme will fallback to the `base` (legacy) package. For developers, they can make custom theme to fallback to `openmage` with full flexibility as in frontend. This is the best I can think of.

kiatng:
For existing stores that override the admin theme, and if they would like to use openmage theme and fallback to it, there is a simple way:

1. Set the `$switch` to legacy theme.
2. Add `theme.xml` to `app/design/adminhtml/default/{custom_theme}` and set the parent to `openmage/default`.

```xml
<?xml version="1.0"?>
<!--
/**
 *
 * @category    design
 * @package     default_{custom_theme}
 */
-->
<theme>
    <parent>openmage/default</parent>
</theme>
```

This works for me.

henrykbrzoska:
Why changing to app/design/adminhtml/base ?
I'm not sure if this change 'default' to 'base' is more bad than useless.

aterjung:
@kiatng thank you for your support on this topic! I have developed this a little further, cause i disliked that you have to switch to a "legacy mode" mode to make your own theme. The way i implemented it now is more open to the future. Building a custom theme dose not need to chose "legacy", instead you could build a custom package and it will in the dropdown to be selected.

aterjung:
@henrykbrzoska it is a requirement to make the "infinite theme fallback" possible like it is implemented in the frontend since 1.4.
I share your doubts, but after the first experiments it seems to work flawless. But this needs more testing.

aterjung:
@kiatng i am not so confident about the default case. 19.4.7 introduced the openmage theme as default - it would look like a regression if we change back to the default theme.

kiatng:
@aterjung I like your approach on the switcher. It's so much better. I have tested it in following cases:

1. Test for BC as an existing store with admin theme override in package `default` theme `custom`, without `theme.xml`. It works right off the box. 
image

2. Set the config:
image

3. Added `theme.xml` in above case to fallback to openmage.
image
The fallback works as well.
image

4. As new custom package with `theme.xml` fallabck, which works perfectly as expected.




In any case, I can get BC with your latest changes. Great job!

kiatng:
@aterjung On your concern on regressing back to the legacy by default, I feel that BC is important. As more stores upgrade/switch to om, it won't break the custom theme override. For stores without it, may be there is a way to greet them with the openmage theme?

aterjung:
@kiatng i don't understand what you are doing there. 

```
                 $packages['default'] ?? $package = 'openmage';
```
is that intended? Or did you mean
 
```
                 $package  = $packages['default'] ?? 'openmage';
```

If not, the way you wrote it sets openmage theme if there is no default folder and the default/base package if there is a default folder present. Depending on the update strategy it is possible that a default folder is left, so this will disable the openmage theme.

I guess your code should handle the case that a shop was updated right now and there is no stores/admin/admin/theme/package entry?

Wouldn't it be better to have a setup script that runs after the update that will set this value to the best guess that is possible? 
I can imagin these cases that should be handled:

- Update from Vanilla M1
   - without custom theme -> OpenMage Theme
   - with customised theme -> Legacy Mode with Info on first Login
- Update from OpenMage 19.4.7 with OpenMage theme V1
   - with activated legacy mode -> Legacy Mode with Info on first Login
   - with activated OpenMage theme -> OpenMage Theme

Depending on the upgrade strategy we have to expect these folder structures in app/design/adminhtml/
- empty default folder
- default folder with modifications left
- default folder with complete old content from old M1 versions
- custom theme inside default folder

I think you should only set openmage theme active if we can safely identify a clean state. If not, we should activate legacy mode and show a notice about it. We should build a info page on the website where we describe what has to be done to make the openmage theme work and link it in the notice.

What do you think?

kiatng:
@aterjung Sorry for the improper usage of null coalescing operator (??). I agree with your analysis. I think you have covered all the cases. Using setup script to set the theme is like _why-didn't-I-think-of-that_, so +1. 
+1 also on documentation and notices. With these in place, it looks perfect to me.

aterjung:
i have now added an update script that should decide if openmage theme can be activated without danger. Only roughly tested until now...

ADDISON74:
You did a great job please continue your work.

fballiano:
This PR has a lot of conflicts now @aterjung are you still interested in continuing it or should we redo it from scratch?

ADDISON74:
It can continue as a project but at least we avoid BC in OM. The list of conflicts seems to be longer and it won't be fixed soon. I express my gratitude to the efforts the author did over the years but when I see those visual issues at the top of the page I don't spend tine to fix them and switch immediately to Magento legacy. 

I propose a vote to separate this Backend theme from OM. Like what we plan with Zend Framework and CM_Redis extension.

fballiano:
uff this PR was extremely interesting but it doesn't receive feedback since 3 years and at this point in time it's really screwed, many conflicts, the committed files would be very outdated, I think we've to close it.

I'll mark it as "don't forget this PR" and it is still linked in the issues (which will remain open). It would be easier to restart it from scratch at this point, I think.

davirs:
I installed the latest version of 20.10.2, when I activate a theme for the frontend, the admin panel automatically switches to the legacy version. In the configuration panel, legacy is disabled.

I tested some things I saw here but without success.

image



image

ADDISON74:
@davirs - Does your reporting refer to this closed PR or is an issue in OpenMage? If it is the last option, please post it in the Issues section, along with all the necessary steps to be reproduced and confirmed.

davirs:
@ADDISON74 This report is related to this PR. I found other issues about exactly this problem, all closed and pointing to this PR.
Although closed, I believe that there is still some missing part of the solution.

ADDISON74:
Please read this comment https://github.com/OpenMage/magento-lts/pull/1229#issuecomment-1546684292.  

As long as the author cannot be contacted for more than 4 years, someone else can take care of this PR if he wants to have a finality.

sreichel:
@davirs can please open a new issue?

This PR would solve the problem, but I don't like to move that files around.

aterjung:
Hi, sorry, i`ve completely lost this topic out of my mind and did not recoginse the messages from github. (had quite some private problems in the last years...) Is its worth getting into this topic again?

ADDISON74:
@aterjung - This PR has been closed for more than 1 year. If you still have the strength to take it to the end, it would be better to create a new one, trying to gather all the feedback received first.

sreichel:
> Is its worth getting into this topic again?

It is :)

sreichel:
@aterjung please re-open, if possible.

PR #1230, opened 2020-09-25
Align label with value

AlterWeb:
### Description (*)
Table information was not aligned properly. There was a 5px top padding for the label that was not there for the value. I found some more small issues with the new admin layout and will fix those when I'm sure this is the correct way to do so.

Screenshot from 2020-09-25 12-26-44

1. Check the alignment in of the table content on the order detail page.

### Questions or comments
This is my first code contribution to OpenMage. If I have to do something different please let me know. I noticed that some lines are marked as changed because of different order of CSS selectors. And single quotes that are replaced by double quotes. I used Ruby Sass to compile the scss file to css with the following command "sass --style=expanded skin/adminhtml/default/openmage/scss/form.scss skin/adminhtml/default/openmage/form.css".


aterjung (approved):

kkrieger85:
Why did you change the order of some css classes ?

AlterWeb:
@kkrieger85 I did not do that manually but the Sass processor did. I only changed one line in the form.scss file and ran the Sass processor like I told in the comment:

> This is my first code contribution to OpenMage. If I have to do something different please let me know. I noticed that some lines are marked as changed because of different order of CSS selectors. And single quotes that are replaced by double quotes. I used Ruby Sass to compile the scss file to css with the following command "sass --style=expanded skin/adminhtml/default/openmage/scss/form.scss skin/adminhtml/default/openmage/form.css".

Without the --style=expanded a lot more lines changed.

aterjung:
Should I recompile your changes with my original sass settings?

AlterWeb:
@aterjung You can do so, but I think maybe it would be a good idea if everyone can do this without changing all those lines. So that we have a method of processing the scss files that everyone can use. What processor / compiler do you use and with what settings?

aterjung:
I am using the binary version of sass from homebrew (Dart-Sass V1.26.9) with no special options applied.

I am not sure what is the right way to resolve this. If the CSS Compiler is working right, It should not change anything if the order is changed a bit, but its confusing to have changes where nothing is changed. But on different platforms there are many different versions of sass available and these might change in the future.

AlterWeb:
@aterjung I don't know either.... I would say it's indeed a bit confusing that there are changes when there is nothing really changing but as long as the code is working like it should, than maybe this is also a non issue... If you're right and all the different versions of sass compilers generate different output than maybe there is not much we can do about it. I never worked with other people on the same scss files before so I don't know if there is a good solution for this.

sreichel (approved):

sreichel:
> but as long as the code is working like it should

Yep, but we should document how to use sass/css, without haveing changes that maybe come from differnt compilers.

fballiano (approved):

Issue #1237, opened 2020-09-29
"Place Order" Button disabled when trying to order virtual products with PayPal-Express


wilfriedwolf:
### Preconditions (*)
1. Use 1.9.4.x

### Steps to reproduce (*)
1. Use virtual product
2. Try to buy with PayPal Express-Checkout

### Expected result (*)
1. I am able to press the "Place Order" button

### Actual result (*)
1.The button is disabled

### References
Thanks to @sprankhub :
See 
1. https://magento.stackexchange.com/questions/62398/place-order-button-disabled-with-virtual-products-paypal-express-and-terms-and/69815#69815
2. https://github.com/customgento/CustomGento_CoreFixes-m1/pull/7/files
3. https://github.com/customgento/CustomGento_CoreFixes-m1/pull/8

wilfriedwolf:
From the - now closed PR:

So now I understand whats going on:

Basically the implementation of the PayPal-Express-Review form is just crap!

Case A: There is a shipping address. Then in 

https://github.com/OpenMage/magento-lts/blob/18739f107de21144a578c8143557f8d2ac6969b3/skin/frontend/base/default/js/checkout/review.js#L63

the Treatment of the submit form is just skipped. That's why it works.

Case B: There is no shipping address. Then in Line 

https://github.com/OpenMage/magento-lts/blob/18739f107de21144a578c8143557f8d2ac6969b3/skin/frontend/base/default/js/checkout/review.js#L77

the event listener is bound to the agreements checkbox disabling the submit button when changing an agreement checkbox.

I guess, what the author of these lines ment to do is to disabling the submit button unless all agreements are checked.

What actually shoud be done is to rewrite this crap in order to check if agreements are checked, preventing submission of the order otherwise just like in the normal checkout.

So as I see it, there is no need to merge the fix from the links, but rather to rewrite the review.js

colinmollenhour:
Thanks for your work on this @wilfriedwolf and sorry to hear your rabbit-hole has gotten deeper.. I'm sure the community would appreciate a PR if you find a solid fix for it! Of course minimal BC breakage would be best but we do have two branches now, one for full-BC (19.x) and one for future (20.x).

fballiano:
Payment method bugs are very hard to reproduce and work, feel free to reopen it or provide a PR for the fix, but since we've no updates in more that a year, I'm closing this one.

PR #1238, opened 2020-09-29
Enable virtual products paypal express

wilfriedwolf:
### Description (*)
The pull requests alters review.js in order not to disable checkout of virtual products with paypal express

### Related Pull Requests
none

1. Fixes OpenMage/magento-lts#1237

See Issue

### Questions or comments
As described here https://magento.stackexchange.com/questions/62398/place-order-button-disabled-with-virtual-products-paypal-express-and-terms-and/69815#69815 but in the core rather than in a theme.


colinmollenhour (changes requested):
I don't know the details of this fix and have never used PayPal Express, but it seems odd to be registering an event listener inside the observer of the same event listener.. Can you explain how/why this fixes the issue?

Also `input` is not defined in this context.. Did you perhaps mean for these lines to be added to the function above?

wilfriedwolf:
Hi Colin,

You are perfectly right, my mistake. In my project used a theme overload of the function and wanted it to become core stuff...
I'll push the corrected branch right away...

Best regards and thanks for the review,

Willi

colinmollenhour:
Ahh, the diff makes much more sense now. However, can you explain why the 'change' event should not be observed in these specific cases? Is there risk that this will affect other checkout methods?

wilfriedwolf:
Ok, thanks for letting me look it over again, actually I just took the references / the fix from the internet without too much understandig what was going on. 

So now I understand whats going on:

Basically the implementation of the PayPal-Express-Review form is just crap!

Case A: There is a shipping address. Then in 

https://github.com/OpenMage/magento-lts/blob/18739f107de21144a578c8143557f8d2ac6969b3/skin/frontend/base/default/js/checkout/review.js#L63

the Treatment of the submit form is just skipped. That's why it works.

Case B: There is no shipping address. Then in Line 

https://github.com/OpenMage/magento-lts/blob/18739f107de21144a578c8143557f8d2ac6969b3/skin/frontend/base/default/js/checkout/review.js#L77

the event listener is bound to the agreements checkbox disabling the submit button when changing an agreement checkbox.

I guess, what the author of these lines ment to do is to disabling the submit button unless all agreements are checked.

So my PR just disables this implementation in a rather weird way :/ 

What actually shoud be done is to rewrite this crap in order to check if agreements are checked, preventing submission of the order otherwise just like in the normal checkout.

Don't know when or if I'll find time to write this PR.

For now, just forget about this...

PR #1247, opened 2020-10-03
Fix call_user_func_array arguments for PHP 8

luigifab:
### Description
Changes second argument of call_user_func_array() from associative to indexed array. Before PHP 8 both types of arrays work the same but on PHP 8 associative arrays are interpreted as "named parameters". In our case this causes code to break.

OpenMage 20.0.3 / PHP 8.0.0-beta3
From https://github.com/joomla/joomla-cms/pull/30608.

### Manual testing scenarios
1. Install PHP 8.0.0
2. Open any page of OpenMage in your browser

### Contribution checklist


colinmollenhour (approved):
Thanks for the PR!

Flyingmana (approved):

PR #1248, opened 2020-10-06
mark trigger_recollect before collectTotals

empiricompany:
### Description (*)

Fix issue that causing the loop if any 3rd party payment/shipping method attempt to reloading the quote before collectTotals is ended
### Related Pull Requests

https://github.com/OpenMage/magento-lts/pull/1244

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -2037,7 +2037,7 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
     {
         // collect totals and save me, if required
         if (1 == $this->getData('trigger_recollect')) {
-            $this->setData('trigger_recollect', 0)->save();
+            $this->setTriggerRecollect(0)->getResource()->save($this);
             $this->collectTotals()->save();
         }
         return parent::_afterLoad();
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -2037,6 +2037,7 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
     {
         // collect totals and save me, if required
         if (1 == $this->getData('trigger_recollect')) {
+            $this->setData('trigger_recollect', 0)->save();
             $this->collectTotals()->save();
         }
         return parent::_afterLoad();


app/code/core/Mage/Sales/Model/Quote.php
```diff
@@ -2037,6 +2037,7 @@ protected function _afterLoad()
     {
         // collect totals and save me, if required
         if (1 == $this->getData('trigger_recollect')) {
+            $this->setData('trigger_recollect', 0)->save();
```
colinmollenhour, line 2040: Is it necessary to call `save()` or is it sufficient to just set the value in memory? The `save()` is not without overhead.

empiricompany: we must to save the record otherwise a subsequent load (new instance) of the quote model into collectTotals() from 3rd party payment/method/observer 
```
class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
.. ..
    public function collectTotals()
    {
.. ..
        foreach ($this->getAllAddresses() as $address) {
.. ..
            $address->collectTotals();
```
```
class Mage_Sales_Model_Quote_Address extends Mage_Customer_Model_Address_Abstract
..
    public function collectTotals()
    {
        Mage::dispatchEvent($this->_eventPrefix . '_collect_totals_before', array($this->_eventObject => $this));
        foreach ($this->getTotalCollector()->getCollectors() as $model) {
            $model->collect($this);
        }
        Mage::dispatchEvent($this->_eventPrefix . '_collect_totals_after', array($this->_eventObject => $this));
        return $this;
    }
```
will fall again into the if of the _afterLoad () method causing the loop

it is an event that occurs in a few cases, I think when the customer has in the cart some products affected by catalog rules applied that require price recalculation
for this the eventual overhead is negligible

i know it may seem incorrect to reset the flag before collectTotals () but i think it is a design problem that in production we only solved this way

colinmollenhour: Ok, what about bypassing the save model observers though? There are probably a lot of observers on `sale_quote_save_after` (or whatever it is) and we probably don't want them to run before `collectTotals` as that could change behavior in more unexpected ways. 

```suggestion
            $this->setTriggerRecollect(0)->getResource()->save($this);
```

Or even better add a new method to the quote resource model to do a very lightweight update to avoid the entire quote model being serialized again.

empiricompany: yes it would be better as it just needs to set the flag on the table without further cascading actions and possible overheads.
updating the quote model would be cleaner but I don't have time to do it and test it I'm sorry, for us this fix is ​​more than enough

empiricompany:
sorry for the delay, we tested the latest suggested patch from @colinmollenhour in a high traffic / high concurrency store (about 5 orders per second), with about 10 catalog rules and with msp_cashondelivery problematic extension.
we fixed our problems, the rules are being applied in the carts, and there are no more infinite loops
so I can definitely ask for it to be merged in the future release

colinmollenhour (approved):
Thanks for the report @empiricompany!

Flyingmana (approved):
Did not test this myself, but the logic of this change makes sense for me.
And in case an exception/error is happening, it still only leads to a new totals collect on a different request, which normally should be safe.

PR #1255, opened 2020-10-09
Change lowest PHP version to 7

spinsch:
### Description
- Removes all php version specific fallback fixes
- Bump min. php version to 7.0.0
- Replace old magento links

### Contribution checklist


Flyingmana:
PHP 5 is end of life since Dec 2018
https://www.php.net/eol.php

so, Iam not fully opposed to it

jouriy:
Please note, that dropping PHP 5.6 support may result in complications for direct upgrades from versions below M1.9.2 running on PHP5.6. So intermediate step with upgrade to M1.9.3+ will be needed first to upgrade PHP before OpenMage upgrade.

Flyingmana:
There would be the alternative, to only remove this parts from version 20.x on, to make the transition easier.

spinsch:
Another alternative would be to throw the check completely out of the files and display a hint in the backend instead.

colinmollenhour:
Just a comment on the keeping 5.6 idea for 19.x and dropping for 20.x.. This would complicate accepting patches. For example we would have to be careful not to backport 5.x incompatible code from 20.x branch or make sure that patches to 19.x are "upgraded" to PHP 7 for 20.x. Otherwise I like the idea but I'm not sure if it is worth it.

User's clinging to PHP 5.x are sinking lower and lower into that pit.. I'm not really keen on complicating our development cycle for PHP 5 users unless there are really a lot of them and the two-step upgrade path seems like a very reasonable workaround.

Flyingmana (approved):

Flyingmana:
for context: current php usage stats for composer. https://blog.packagist.com/php-versions-stats-2021-1-edition/

Flyingmana (approved):

luigifab (commented):
Not tested, but I agree, it's time to say good bye to PHP 5.
Also, can I suggest to remove `if version_compare` in index.php files? :)

Flyingmana:
Removing of the check I would see as separate topic

jouriy:
> Not tested, but I agree, it's time to say good bye to PHP 5.
> Also, can I suggest to remove `if version_compare` in index.php files? :)

de facto, OpenMage cannot run on PHP 5 since v19.4.9 due to 5dee1c09, cf25b8a2 and 26b9eee, so this check is required now. Migration from older Magento versions require intermediate step with upgrade 1.9.4.5 first followed by PHP switch to 7.3.

kiatng (approved):

kkrieger85 (approved):

Issue #950, opened 2020-05-05
No throw when Mage_Admin_Model_User::delete fails


midlan:
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Admin/Model/Resource/User.php#L219

inside this catch is not thrown any error, so user is not deleted and you get green message:

> The user has been deleted.

But the user still remains in database and also can login. It return `false` but the return value is ignored in controller: \Mage_Adminhtml_Permissions_UserController::deleteAction

**It is bug or it is there for some purpose?**

Sekiphp:
@midlan In my opinion this is bug, because is ignored returned value from `delete` method on `admin/user` model.

I suggest that there should be (at least):
`app/code/core/Mage/Adminhtml/controllers/Permissions/UserController.php:187`

`
            try {
                $model = Mage::getModel('admin/user');
                $model->setId($id);
                if ($model->delete()) {
                Mage::getSingleton('adminhtml/session')->addSuccess($this->__('The user has been deleted.'));
}
else {
                Mage::getSingleton('adminhtml/session')->addSuccess($this->__('Deleting of user was terminated!'));
}
                $this->_redirect('*/*/');
                return;
            }
            catch (Exception $e) {
                Mage::getSingleton('adminhtml/session')->addError($e->getMessage());
                $this->_redirect('*/*/edit', array('user_id' => $this->getRequest()->getParam('user_id')));
                return;
            }
`

colinmollenhour:
I think instead of returning false on that second to last line of User.php the exception should simply be rethrown (`throw $e;`). This is the common pattern throughout Magento. The error message "Deleting of user was terminated!" doesn't help the user at all or even make sense so should just be removed and the return value not checked at all.

midlan:
@colinmollenhour 
I agree. But what about backcompatibility? Someone may rely on the return value and only catching exceptions of type `Mage_Core_Exception `.

colinmollenhour:
It should be noted then as a BC change. We are still working out exactly how to deal with BC changes, though...

elidrissidev:
Closing as fixed.

sreichel:
@all-contributors add @midlan bug

PR #1257, opened 2020-10-11
Make overrides of Mage_Core_Model_Resource_Db_Abstract::delete respect parent api

midlan:
### Description (*)

I fixed overrides of Mage_Core_Model_Resource_Db_Abstract::delete to respect parent api, which means:
- throw on error
- return `$this`

1. Fixes OpenMage/magento-lts#950
#### Admin panel
1. make some constraint for `admin_user` table that not allow you delete record (e.g. foreign key referencing the table with `RESTRICT` rule `ON DELETE`)
2. go to admin panel and try delete some user, instead of green message that user was deleted, we should get red error message
#### API
There is changes also in API resource model, so we should test it also.


kkrieger85 (approved):

midlan:
@Flyingmana @sreichel review pls?

midlan:
It is BC break, labeled it and changed base branch to 20.0

Sekiphp (dismissed):
seems ok

for future compatibility, you should consider to replace catch of `Exception` to `Throwable`

fballiano:
> seems ok
> 
> for future compatibility, you should consider to replace catch of `Exception` to `Throwable`

I switched them

fballiano (dismissed):

fballiano:
But we should still add a note to the README in the "differences between v19 and v20"

elidrissidev:
Updated the branch since the README's version was ancient and added the note to the README.

elidrissidev (approved):

PR #1258, opened 2020-10-14
Optimize the indexing catalog query and remove the always true condition(1=1)

amiriun:
### Description (*)
There is an always true condition in of the joins of product indexer, so I removed it.
Also I realized that if we check whether d.value is not null in the where clause of **FROM** , the query execution time could be very faster, and you can see the difference by explaining the query and compare them after and before this change.
In my benchmark, it is **4600% faster** than before.
I believe this condition (AND d.`value` IS NOT NULL) has not any effect in the result of the query because as you can see we have been already filtering this query by d.value in the line of 147 

1. remove extra condition, which is always true (1=1)
2. add a condition to one of the where clauses to increase the query

1. Reindex the produducts
2. check the results and counts
3. checkout to my branch
4. reindex the products again
5. check the results and counts
6. both of them should be the same and the index execution should be faster

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Source.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Eav/Source.php
@@ -116,10 +116,10 @@ class Mage_Catalog_Model_Resource_Product_Indexer_Eav_Source extends Mage_Catalo
             )
             ->joinLeft(
                 array('d' => $this->getValueTable('catalog/product', 'int')),
-                '1 = 1 AND d.store_id = 0',
+                'd.store_id = 0',
                 array('entity_id', 'attribute_id', 'value')
             )
-            ->where('s.store_id != 0');
+            ->where('s.store_id != 0 AND d.`value` IS NOT NULL');
 
         $statusCond = $adapter->quoteInto(' = ?', Mage_Catalog_Model_Product_Status::STATUS_ENABLED);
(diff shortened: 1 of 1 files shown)


tmotyl:
Hi
Can you paste the query before and after the change?

kkrieger85:
Tested it

Before: 
``` 
SELECT `s`.`store_id`, `s`.`website_id`, `d`.`entity_id`, `d`.`attribute_id`, `d`.`value` FROM `core_store` AS `s`
 LEFT JOIN `catalog_product_entity_int` AS `d` ON 1 = 1 AND d.store_id = 0 WHERE (s.store_id != 0)
```

After:
``` 
SELECT `s`.`store_id`, `s`.`website_id`, `d`.`entity_id`, `d`.`attribute_id`, `d`.`value` FROM `core_store` AS `s`
 LEFT JOIN `catalog_product_entity_int` AS `d` ON d.store_id = 0 WHERE (s.store_id != 0 AND d.`value` IS NOT NULL) 
```

joshua-bn:
The explain for those queries for me are identical.

tmotyl:
@joshua-bn @kkrieger85 which db versions were you using? my experience is that often the minor version of db changes the preformance test results.

joshuaadickerson:
@tmotyl 5.6 something on RDS and my local Docker env. Doesn't matter though. It changes functionality.

fballiano:
the explain is exactly the same, the original author disappeared since the first post, so I'm closing this

PR #1120, opened 2020-07-22
Add --dev requirement captainhook & initial config

kkrieger85:
Proposal for automated code style linting

- Add [CaptianHook](https://captainhookphp.github.io/captainhook/hooks.html) to `--dev` requirements
- Add [PHPCS Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) to `--dev` requirements
- Add initial config file `captainhook.json` with `pre-commit` action 

- With commiting this changes, the git hook became active and fixes code in `app` folder with PSR2-Rules

Related:

- https://github.com/OpenMage/magento-lts/issues/947
- https://github.com/OpenMage/magento-lts/issues/982

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

Issue #982, opened 2020-05-15
Add phpCBF/captainhook to worklow?


sreichel:
Is it possible to run phpCBF (PSR2) before files are merged?

The doc PRs have automated fixes, but they dont cover all files ...  with pre-merge hook we could improve code with every PR.

tmotyl:
There are magento official rules for php code sniffer we can use. For just cgl changes (psr-2 and psr-12) we can also use the php-cs-fixer

Flyingmana:
Do you mean as a check, or as a fixer?
Check is easy, as something which changes the Files is a lot harder.

Adding https://github.com/CaptainHookPhp/captainhook could be usefull, as long as we do not require it for all contributors to use.
Also Iam not a fan of mixing code and codestyle changes together, its often enough work to check one of it :/

tmotyl:
My idea was that php-cs-fixer should be run as check (--dry-run) in gitlab actions/travis and with "fix mod" locally on git commit hook.
Then if commit fails, you wil see whitespace changes made by php-cs-fixer as unstaged changes.
So you add them to the commit (stage) and commit again.

Alternatively we can run dry-run in both cases, but provide a convenience script like "build/fixMyCommit.sh" which you can run to fix files from the last commit, and then stage the and commit --amend.

See how such a script works in TYPO3 world https://github.com/TYPO3/TYPO3.CMS/blob/master/Build/Scripts/cglFixMyCommit.sh
https://docs.typo3.org/m/typo3/guide-contributionworkflow/master/en-us/Appendix/CglFixMyCommit.html

sreichel:
> Do you mean as a check, or as a fixer?
> Check is easy, as something which changes the Files is a lot harder.

Question was for fixer. Automated fixes (psr2?) for spaces, linebreaks, ....

E.g. I dont want to ask for fix spaces here https://github.com/OpenMage/magento-lts/pull/980/commits/dca888af02ebc4c4857237be27e51e963bf6af9f ... 

Maybe we can add a CONTRIBUTING file and suggest using pre-commit hook like https://gist.github.com/fdemiramon/0423b4308218d417fbf3?

tmotyl:
Travis/github actions should check for code style automatically and go red if commit doesnt comply. So the author should run fixer and push the patch again. We should not merge any new patches with broken code style

Flyingmana:
could we formulate this a bit more positive?

[...]

Issue #947, opened 2020-05-04
Discussion: Consider using PHP CS Fixer and/or PHPStan


colinmollenhour:
- Initially run CS Fixer to update all code to conform to a standard
- Add Travis pipeline or Github Action to throw error when CS Fixer output is different from pushed code (showing diff)

https://github.com/OpenMage/magento-lts/pull/693#discussion_r416685506
#365

colinmollenhour:
@sreichel What are the pros/cons of PHP CS Fixer and PHPStan and are they completely redundant?

tmotyl:
Php fixer just take care of CGL and can fix the formatting. While phpstan does static code analysis but cant fix code. 

So we need both :)

sreichel:
> @sreichel What are the pros/cons of PHP CS Fixer and PHPStan and are they completely redundant?

As @tmotyl said. No, they aren't redundant.

IMHO cs-fixer is "just" a nice to have to have code matching PSR2 standard or whatever. phpstan/psalm is to prevent logical problems .... can't find it actually, but there was a nice article about predicted or expected variables ....

kkrieger85:
+1 for phpcsfixer for PSR2
`php php-cs-fixer.phar fix /path/to/project --rules=@PSR2`

This could also improve to review PR like https://github.com/OpenMage/magento-lts/pull/1035 and all other PR consider code style

PR #1264, opened 2020-10-18
Chore: Add Captainhook & PHPCS to composer require-dev and sample captainhook config

kkrieger85:
Add captainhook and PHPCS to composer require-dev section
Add sample captainhook config 

closes #1120 (duplication)
refs #947 #982 

Please consider, that commit hooks are disabled in captainhook.json

You could test it by installing packages with composer and using the PHPCS Rule with using the command in captainhook.json


tmotyl:
I would love to have php-cs fixer in place.
This change however is missing few pieces:
1) php-cs-fixer configuration file (lets start with config which is not making too much changes in the codebase - and add new rules later on
2) github action to validate codestyle on PR
3) use friendsofphp/php-cs-fixer v3 instead of v2

kkrieger85:
Feel free to update my PR

sreichel:
Seems related to #2400

PR #1267, opened 2020-10-18
1095 remove notification survey block

kkrieger85:
Remove block in adminhtml template:
<type="adminhtml/notification_survey" name="notification_survey" template="notification/survey.phtml"/>
Remove survey related code
Remove survey related translation

After merging, in adminhtml should be no more survey block

no_survey

closes #1095

PR #1268, opened 2020-10-18
20.0 1095 remove notification survey block

kkrieger85:
Remove block in adminhtml template:
<type="adminhtml/notification_survey" name="notification_survey" template="notification/survey.phtml"/>
Remove survey related code
Remove survey related translation

See also #1267


Flyingmana (dismissed):

Flyingmana:
I think we should keep the phpstorm metadata thing separated in the future, maybe putting this into a github action which updates this once a month

sreichel:
> 
> 
> I think we should keep the phpstorm metadata thing separated in the future, maybe putting this into a github action which updates this once a month

Is it possible w/o n98 a composer requirement and pre-commit-hook?

Flyingmana:
Yes

fballiano:
I think this will create a conflict when https://github.com/OpenMage/magento-lts/pull/1267 will be ported to the v20 branch. I think it should be closed.

luigifab:
If I remember correctly, it's the same changes.

luigifab:
So, this is the closing day?

fballiano:
exactly, same changes plus some phpstorm metadata, but I'll close this anyway

PR #1292, opened 2020-10-30
Support the logging of Throwables

woutersamaey:
With PHP Strict Types becoming more common, the "Error" class comes in the picture and this does not extend from Exception.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -905,9 +905,9 @@ final class Mage
     /**
      * Write exception to log
      *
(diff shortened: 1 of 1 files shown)


Flyingmana (approved):
Approving, with the note this increases the minimum php version to 7.0
(so if there are to many arguments against it because of this, we can move it to 20.x and then its fine)

colinmollenhour (approved):

PR #1299, opened 2020-11-05
Set logo width explicitly

loekvangool:
### Description (*)
`mod_pagespeed` and other tools may fill in missing dimensions on `image` tags for better performance, leading to distorted images.

1. Open the admin page


colinmollenhour (dismissed):

luigifab:
This change works for legacy admin theme, but break openmage admin theme.

fballiano:
I'm closing since it breaks the openmage theme (tested today). we should have a class in the body which identifies the theme and then have specific css for the themes, but that's outside the scope of this PR.

PR #1301, opened 2020-11-10
Prefix added to CM Redis cache backend to enable multiple instances

greben99:
### Description (*)
Problem is you currently cannot use the same redis cache backend for multiple instances of Magento. The cache would clash because the prefix is hard coded. To fix this, added a <site> param in the options, which will be prepended to the cache prefix, allowing multiple magento instances to connect to the same redis DB. (You can use the DB param, but that is limited)

        <cache>
            <backend>Mage_Cache_Backend_Redis</backend>
            <backend_options>
                 <server>###</server>
                 <site>97</site>
                 <port>###</port>
                 ....
            </backend_options>
        </cache>

1. Connect 2 magento instances to the same redis backend DB.
2. Listen to the redis DB and ensure every item that is saved has the specified prefix.


colinmollenhour:
Please see discussion here: https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/issues/46

I think main takeaway is this:

> the proper place to set the prefix is the frontend not the backend and the frontend already supports cache_id_prefix.

fballiano:
I think @colinmollenhour comment is conclusive

Issue #1178, opened 2020-09-02
Error in the "_IsValidSource" method in Varien/Io/File.php


felipegomesdev:
Hello.

After this change is part of the code (https://github.com/OpenMage/magento-lts/pull/966)

The SOAP API Feature of creating a product image has stopped working. Look [documentation](https://r-martins.github.io/m1docs/guides/m1x/api/soap/catalog/catalogProductAttributeMedia/catalog_product_attribute_media.create.html)

Well, the whole image has a null value. For example:
```php
$src = file_get_contents("https://pbs.twimg.com/media/EelBMRwXsAEcrM9.png");

if (strpos($src, chr(0)) === false){
    echo "success...";
}else {
    echo "error...";
}
```
Therefore, it is not possible to pass the _IsValidSource method in Varien/Io/File.php

Same problem at https://github.com/OpenMage/magento-lts/issues/1125

sreichel:
@all-contributors add @felipegomesdev bug

Issue #1125, opened 2020-07-26
Null bytes found in saving the PDF order


boesbo:
Null bytes are detected if you try to save the pdf Order. The "Varien_Io_File::_IsValidSource($src)" function blocks the upload because it finds null bytes (strpos($src, chr(0)) === false).

### Preconditions (*)
v1.9.4.x-dev

### Steps to reproduce (*)
```
// Get pdf.
$pdf = Mage::getModel('emailattachments/order_pdf_order')->getPdf([$order]);
$pdfName = 'order.pdf';
$baseDir = Mage::getBaseDir('tmp');
$filePath = $baseDir . DS . $pdfName;

$file = new Varien_Io_File();
$file->write($filePath, $pdf->render());
```

maximehuran:
I have the same problem when I use Image API to create new product image sending the content in base64.

sreichel:
@all-contributors add @boesbo bug

Issue #666, opened 2019-04-18
Warning: is_readable() expects parameter 1 to be a valid path


sreichel:
From https://magento.stackexchange.com/questions/270687/warning-is-readable-expects-parameter-1-to-be-a-valid-path

Not sure if supress error is best idea ... see #85 

Untested, but a file_exists() check before c/should cover this.

manezaf8:
The thing is we are not suppose to update the [  lib ] Folder since we working with big data and most of the functions are related, so how to go about it to stop this warning from happening ?

PR #1302, opened 2020-11-10
Fix a bug where product media upload via API was not possible anymore.

digitalpianism:
### Description (*)

The following commit (https://github.com/edannenberg/magento-lts/commit/cd0c30fe8bbd6d966e267081bca8b223f62e0ff2) has introduced a few bugs:
- https://github.com/OpenMage/magento-lts/issues/1178
- https://github.com/OpenMage/magento-lts/issues/1125

The purpose of this PR is to fix those bugs by trying to keep the original changes made in the commit I mentionned above.

1. Fixes OpenMage/magento-lts#1178
2. Fixes OpenMage/magento-lts#1125
3. Fixes the EVIL issue OpenMage/magento-lts#666

1. Run the following php file from the root of your Magento website.
```
<?php
//some settings
error_reporting(E_ALL | E_STRICT);
define('MAGENTO_ROOT', getcwd());
$mageFilename = MAGENTO_ROOT . '/app/Mage.php';
require_once $mageFilename;
Mage::setIsDeveloperMode(true);
ini_set('display_errors', 1);
umask(0);
//instantiate the app model
Mage::app();
$tmpDirectory = Mage::getBaseDir('var') . DS . 'api' . DS . "test";
$ioAdapter = new Varien_Io_File();
try {
    // TEST 1: simple text file
    $fileContent = "test";
    $fileName  = 'test1.txt';
    // Create temporary directory for api
    $ioAdapter->checkAndCreateFolder($tmpDirectory);
    $ioAdapter->open(array('path' => $tmpDirectory));
    $ioAdapter->write($fileName, $fileContent, 0666);
    unset($fileContent);
} catch (Exception $e) {
    echo $e->getMessage() . "\n";
}

$ioAdapter = new Varien_Io_File();
try {
    // TEST 2: image file from URL
    $fileContent = file_get_contents("http://www.google.com/images/branding/googlelogo/2x/googlelogo_color_92x30dp.png");
    $fileName  = 'test2.png';
    // Create temporary directory for api
    $ioAdapter->checkAndCreateFolder($tmpDirectory);
    $ioAdapter->open(array('path' => $tmpDirectory));
    // Write image file
    $ioAdapter->write($fileName, $fileContent, 0666);
    unset($fileContent);
} catch (Exception $e) {
    echo $e->getMessage() . "\n";
}

$ioAdapter = new Varien_Io_File();
try {
    // TEST 3: image file from path
    $fileContent = $tmpDirectory . DS . "test2.png";
    $fileName  = 'test3.png';
    // Create temporary directory for api
    $ioAdapter->checkAndCreateFolder($tmpDirectory);
    $ioAdapter->open(array('path' => $tmpDirectory));
    // Write image file
    $ioAdapter->write($fileName, $fileContent, 0666);
    unset($fileContent);
} catch (Exception $e) {
    echo $e->getMessage() . "\n";
}

$ioAdapter = new Varien_Io_File();
try {
    // TEST 4: image file from URL with addition of base64 encode/decode
    // To reproduce the behavior of Mage_Catalog_Model_Product_Attribute_Media_Api::create()
    $imageFile = "http://www.google.com/images/branding/googlelogo/2x/googlelogo_color_92x30dp.png";
    $imData = file_get_contents($imageFile);
    $fileContentSent = base64_encode($imData);
    $fileContent = @base64_decode($fileContentSent, true);
    $fileName  = 'test4.png';
    // Create temporary directory for api
    $ioAdapter->checkAndCreateFolder($tmpDirectory);
    $ioAdapter->open(array('path' => $tmpDirectory));
    // Write image file
    $ioAdapter->write($fileName, $fileContent, 0666);
    unset($fileContent);
} catch (Exception $e) {
    echo $e->getMessage() . "\n";
}

$ioAdapter = new Varien_Io_File();
try {
    // TEST 5: image file from path with addition of base64 encode/decode
    // To reproduce the behavior of Mage_Catalog_Model_Product_Attribute_Media_Api::create()
    $imageFile = $tmpDirectory . DS . "test3.png";
    $imData = file_get_contents($imageFile);
    $fileContentSent = base64_encode($imData);
    $fileContent = @base64_decode($fileContentSent, true);
    $fileName  = 'test5.png';
    // Create temporary directory for api
    $ioAdapter->checkAndCreateFolder($tmpDirectory);
    $ioAdapter->open(array('path' => $tmpDirectory));
    // Write image file
    $ioAdapter->write($fileName, $fileContent, 0666);
    unset($fileContent);
} catch (Exception $e) {
    echo $e->getMessage() . "\n";
}

```
2. You will get only **3 files out of 5** under `var/api/test` folder. **Both images will be empty images**
3. Apply the PR 
4. Run the file again.
5. You will get **all the files** from the 5 different tests. **All images will be correct.**

### Questions or comments

The original purpose of the buggy commit was a security fix: prevent null byte inside filename from being processed when writing a file. The **problem introduced** was that the source ([as explained in the PHPDoc)](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Varien/Io/File.php#L452) can be **either a string, a file or a stream**. Thus, what I've done is to **check if the string provided is a file, and in that case only, check for null byte**. 

Indeed, **when you use file_get_content as a source** for the Varien_Io_File::write method, **the original null byte check will return true and thus the write will fail silently.**

Note that I have also re added the @ operator to the is_readable call as it throw a warning in case you provide content directly instead of providing a file path.


colinmollenhour:
Looks good but please revert the unrelated changes to change log and Mage.php

digitalpianism:
@colinmollenhour done sorry that was a mistake I didn't want to changes those two files ^^

colinmollenhour (dismissed):

lib/Varien/Io/File.php
```diff
@@ -486,8 +486,17 @@ public function write($filename, $src, $mode=null)
...
+        // In case of a string
+        if (is_string($src)) {
+            // If its a file we check for null byte
+            // If it's not a valid path, file_exists() will return a falsey value, and the @ will keep it from complaining about the bad string.
+            if (@file_exists($src)
+                && strpos($src, chr(0)) !== false) {
+                return false;
+            } else {
+                return true;
+            }
```
sreichel, line 498:
```suggestion
            return @file_exists($src) && strpos($src, chr(0)) !== false ? false : true;
```

sreichel (approved):

maximehuran:
We applied this fix on a store which using API to create images and it works

rjocoleman:
I encountered this bug today too! Thanks for the fix @digitalpianism

Along with the fix this PR brings a few things with it, they're all cancelled out but it will introduce 9 irrelevant commits that are then undone by another commit.

The fix itself is only this: https://gist.github.com/rjocoleman/4422ca5cd56fe5ca84a316170f7196c0

@digitalpianism to fix up the commit history and not introduce the extra cruft could you please rebase? 

(One way to do this by rebasing the branch interactively against the base branch and dropping all the irrelevant commits, then "squashing" the relevant ones together https://asciinema.org/a/JzWvAeKaFcY7h51mvcFXD1X8h - if you do this you'd need to then force push the `fix-1178` branch to overwrite the old history)

Aside from keeping the commit history clean a major advantage is being about to use things like [composer patches](https://github.com/cweagans/composer-patches) to apply the fix from this PR before it's merged (as it stands the patch is really large and might not cleanly apply because of the irrelevant commits)

thanks!

wytzejan:
When can this one be merged? @digitalpianism 
I have also applied this fix on a store which using API to create images and it works. The only thing is that everytime I update Magento I manually have to make this code change in the core.

rvelhote:
We were doing some data transfer using the API and were affected by this issue. This patch to `lib/Varien/Io/File.php` fixed the issue for us. No side-effects were detected.

colinmollenhour (approved):

PR #1304, opened 2020-11-12
Fix config bug in Cm_RedisSession

henrykbrzoska:
so, we make working proper config values break_after_frontend & break_after_adminhtml from
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/etc/local.xml.additional


joshua-bn:
Can you explain what this is for? What the issue is that you're fixing?

Unless you changed the name of the sessions, they should be frontend and adminhtml (afaik). Your method is fragile.

colinmollenhour:
Is `session_name()` in this context not holding the proper session name?

henrykbrzoska:
@colinmollenhour maybe i've something wrong in php ini settings, but on load session i have try to load 'break_after_PHPSESSID' by default
@joshua-bn yes, there should be check if admin uri is set in backoffice.

henrykbrzoska:
nvm

Issue #1306, opened 2020-11-13
My concerns about _hasDataChanges and the huge performance impact


woutersamaey:
In Magento, any time you use a setter, the property _hasDataChanges is set to true, even when you're not actually changing anything.

I.e. this code will set _hasDataChanges to true, even though we're not changing anything.
```
$stockItem->setQty($stockItem->getQty());
```

This is disturbing me, because it can make Magento very slow, causing unneeded save and MySQL stuff to run.

I wanted to open this issue as a way to discuss this.

On occasion, I have in fact overridden the setData() method for specific models so _hasDataChanges remains false. In my experience it can greatly improve stock qty import scripts for example. It's mind blowing how much speed can be gained.

My implementation looks like this:
```
public function setData($key, $value = null) {
        // In our version, we first check if we are actually changing something. Not setting the _hasDataChanges flag regardless.
        if (!is_array($key) && !is_object($value)) {
            $oldValue = $this->_data[$key] ?? null;
            if ($oldValue === $value) {
                // No need to change anything
                return $this;

            } elseif ($oldValue === (string) $value) {
                // Still no need to change anything
                return $this;

            } elseif (is_numeric($value) && is_numeric($oldValue) && '' . round($oldValue, 8) === '' . round($value, 8)) {
                // The number is the same, with up to 8 decimals accuracy
                return $this;
            }
        }

        return parent::setData($key, $value);
    }
```

Why not make this standard for all models? Yes, it is a bit tricky, but isn't it worth it given the huge speed gain that is possible?

One part of the "tricky" side is that events are also triggered. IMHO I'd say "fuck it". The speed gain is very important and you're not actually saving anything, so why should an event be triggered?

### Preconditions (*)
1. Any installation since the dawn of Magento

### Steps to reproduce (*)
1. Use any setter on any DB-backed model and set to the same value already there.
2. Run save() on the model. It will save, even though there is no need for it.

### Expected result (*)
1. Don't save at all, because there is no need, greatly improving speed!

### Actual result (*)
1. It saves the existing data back to the DB like we're used to, also triggering events...

woutersamaey:
The label "bug" is wrong, but I don't know how to change it.

tmotyl:
Hi
@woutersamaey Ive also wondered the same some time ago.
Do you also observe performance improvement in some build in functionality ? Can you share some numbers?

I think that the first condition (values being equal) is safe. Im not sure how safe it would be testing the string cast.

[...]

PR #1307, opened 2020-11-13
Unset the "has data changed" flag after loading stock items linked to a product.

woutersamaey:
The flag is set incorrectly, because some product data was added to the stock item, making Magento believe there was an actual change, when in fact there wasn't. This is the best way to reset this.

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Observer.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Observer.php
@@ -121,6 +121,14 @@ class Mage_CatalogInventory_Model_Observer
         } else {
             Mage::getModel('cataloginventory/stock_status')->addStockStatusToProducts($productCollection);
         }
+
+        // Stock items don't have any real changes (yet), unset the flag!
+        foreach($productCollection as $product){
+            $stockItem = $product->getStockItem();
+            /* @var $stockItem Mage_CatalogInventory_Model_Stock_Item */
+            $stockItem->setDataChanges(false);
+        }
+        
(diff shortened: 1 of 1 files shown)


woutersamaey:
Keeping the "_hasDataChanges" clean leads to greater performance because no save will happen if there aren't any real changes. Not exactly the same, but there's also some discussion in #1306

kiatng:
May be a better fix on skipping model save is to add this method:

```php
// In app\code\core\Mage\CatalogInventory\Model\Stock\Item.php
    /**
     * Checks that stock item model has data changes.
     * Call save stock item if model isn't need to save in DB
     *
     * @return boolean
     */
    protected function _hasModelChanged()
    {
        if (!$this->hasDataChanges()) {
            return false;
        }

        return $this->_getResource()->hasDataChanged($this);
    }
```

colinmollenhour:
@woutersamaey Sorry, I should have looked at this long ago.. I approve of this change but while I was applying some other changes for #1306 I went ahead and implemented an alternative solution that is perhaps a bit more "kosher" in #2066. (adds a proper protected property to reference the stockItem instead of using _data).

colinmollenhour (approved):
Approve, but if #2066 is merged this does not need to also be merged.

fballiano:
I think https://github.com/OpenMage/magento-lts/pull/2066 would be a more comprensive solution :-)

fballiano:
Since https://github.com/OpenMage/magento-lts/pull/2066 is already merged we don't need this one.

PR #1311, opened 2020-11-18
Larger panels in attribute set editor

loekvangool:
### Description (*)
Increasing the breathing space a bit. Screens have more pixels now.

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
@@ -71,7 +71,7 @@
                     </tbody>
                 </table>
             </div>
-            <div id="tree-div2" style="height:400px; margin-top:5px;overflow:auto"></div>
+            <div id="tree-div2" style="height:700px; margin-top:5px;overflow:auto"></div>
             <script type="text/javascript">
             //<![CDATA[
             var allowDragAndDrop = <?php echo ($this->getIsReadOnly() ? 'false' : 'true'); ?>;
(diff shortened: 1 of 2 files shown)


ADDISON74:
My personal opinion related to this proposal. In Edit Attribute Set page there are two div's with different heights identified by the following id's: **tree-div1** and **tree-div2**.

To achieve a little visual improvement in window space I would take in consideration two conditions: a maximized window for a 1080p screen and equal heights for 2 and 3 columns. By playing in browser inspecting the page code in order to get the same column heights and no scroll bar based on a maximized window I got the following values:

For id=tree-div1 height should be 490 px
For id= tree-div2 height should be 541px

If you set the value for height to 700px you will get the scroll bar. Also you should change the column height in the middle to take the benefits of the space.

My suggestion is to keep the same value for column in the middle as Magento Team set up and increase it for right column (Unassigned Attribute) to 451px. Otherwise the two values above are the ones I would recommend if the window is maximized.

loekvangool:
I'm happy with any solution that allows the panels to grow a bit vertically. TA better solution would be something that grows with the screen, and I PR'ed this as a quick fix. I wonder what assumptions we can make about the resolutions used by our users (users that edit attribute sets in OpenMage). 1080 seems on the low side imo.

fballiano:
sorry guys, I'll close this in favor of https://github.com/OpenMage/magento-lts/pull/3247, also because this needs to be rebased anyway.

@loekvangool I'll mark you as author of the commit of the new onw.

loekvangool:
Fine ofc thanks for the effort

Issue #1289, opened 2020-10-28
Wrong attributes order in comparison window


ADDISON74:
This is a known issue in Magento. The attributes order listed in product page is not the same in comparison window.  In the past I reported this issue on Stack Exchange and Magento Forum. Here are the links: 

https://magento.stackexchange.com/questions/64983/attributes-order-in-comparison-window-is-wrong

https://community.magento.com/t5/Can-Magento-do/Attributes-order-in-comparison-window-is-wrong/m-p/4137

I vote for adding this missing feature to OpenMage.

seansan:
Wow! Thats a very old issue 2015 ... good one

btw is the "old" bugs database/reporting system from M1 still exist somewhere?

ADDISON74:
Between 2014 - 2018 I reported a lot of issues in Magento 1 Bugtracker. Some of them were took in consideration. Unfortunately I don't think there is a backup to it. 

- I found all Magento CE packages are here + SB Samples: https://pubfiles.nexcess.net/magento/ce-packages/

- The last Magento Connect extensions can be found here: https://pluginarchive.com/magento. At a moment there were around 7000 extensions and in the end only half remain available.

ADDISON74:
I will close this issue. Please check #1568 for a fix.

PR #1314, opened 2020-11-18
Some more room for the category menu

loekvangool:
And replacing the background image with a border.

### DIFF
--- a/skin/adminhtml/default/default/boxes.css
+++ b/skin/adminhtml/default/default/boxes.css
@@ -572,7 +572,7 @@ td.divider              { font-size:1px; line-height:0; }
 .container-collapsed { padding:1.8em 2.2em 1.8em 2em; padding-top:0; }
 
 div.side-col { float:left; width:280px; margin-right:-280px; padding-bottom:25px; }
-div.main-col { margin-left:280px; min-height:450px; padding:0 0 25px 25px; border-left: 2px solid #bdbdbd; }
+div.main-col { margin-left:280px; min-height:450px; padding:0 0 25px 25px; border-left: 2px solid #dedede; }
 div.main-col-inner { float:left; /* Fixes some inner clears in the liquid main-col */ width:100%; }
 
 .footer { clear:both; background:url(images/footer_bg.gif) repeat-x #e6e6e6; padding:105px 2.8em 2.8em 2.8em; font-size:.95em; text-align:center; }
--- a/skin/adminhtml/default/default/boxes.css
+++ b/skin/adminhtml/default/default/boxes.css
@@ -570,10 +570,9 @@ td.divider              { font-size:1px; line-height:0; }
 .middle { min-height:450px; background:url(images/simple_container_bg.gif) repeat-x #fff; padding:23px 27px 0 27px; }
 .middle-popup { border-bottom:3px solid #fff; background:url(images/middle_bg.gif) repeat-x 0 100% #fff; padding:0 0 0 0; background:yellow; }
 .container-collapsed { padding:1.8em 2.2em 1.8em 2em; padding-top:0; }
-.columns {background:url(images/side_col_bg.gif) repeat-y 217px 0; }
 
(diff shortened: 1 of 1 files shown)


loekvangool:
Before:

image

After:

image

jouriy:
In my opinion skin/adminhtml/default/openmage/scss/override.scss would be a better place for all new CSS improvements instead of direct modifications in fallback legacy theme.

ADDISON74:
Personally, I encourage solving bugs in the default theme rather than changes that look like and can be operated in a single file, in this case styles.css. This is because appearance changes are often a subjective matter and without a UX / UI team in the background a detailed analysis of the changes cannot be performed.

At this point there is a nice theme for the Magento backend that I tested and it offered several look options. I don't remember if it was available a SCSS file or not with this theme.

After 12 years of using the Magento interface, I came to learn about its pros and cons. I consider a real improvement for example the ascending sorting of the attributes or their search in the editing page of a set of attributes than a visual improvement.

loekvangool:
>I consider a real improvement for example the ascending sorting of the attributes or their search in the editing page of a set of attributes than a visual improvement.

I would love to have that.

ADDISON74:
@loekvangool: please take a look to this post: https://github.com/OpenMage/magento-lts/issues/1289.

fballiano (approved):
I made the border lighter but I think this is a good PR

ADDISON74 (approved):

ADDISON74:
If it were up to me I would have removed the dividing line between the category tree and the grid. It looks much better without it. Only if the width for the category tree container was allowed to be resized then I would have kept a line with a thickness of 2px (half of the header page bottom border). I guess I saw a PR which implements this future.

This is the vertical line:

```css
.columns {background:url(https://www.mydomain.tld/skin/adminhtml/default/default/images/side_col_bg.gif) repeat-y 217px 0;;}
```

I'm not a fan of the bottom borders under the header either, that has 4px. I consider this line useless, eliminating it would solve more visual problems. It is obvious that the interface was designed for the times when it was an obsession to separate the content as much as possible. Much like the gradient buttons, animated gifs, flash content.

category_page

luigifab:
There is something wrong.
It "destroy" the system configuration menu (before/after):

image image

ADDISON74:
If I am not mistaken there is a JavaScript-based solution that allows resizing the category tree block. That is the best option because you can adjust it as you wish. It won't work only from CSS because you will not be able to solve all situations, some will have too much space, others too small.

In this case I suggest reverting this PR. That right border looks ugly. I also don't like to see those left side tabs so wide. It is clear that I approved this PR without testing it enough. These are situations when things that are not at all complicated at first glance can cause problems.

ADDISON74:
I tested this PR and I confirm the issue reported by @luigifab . The change was made in a class that affects in the Backend all the left side columns. Here's what the change implemented in some sections of the Backend looks like:

**Categories**

categories

**Products**

products

**Configuration**

system

The width of the tabs changes and becomes unsightly, we have an extra vertical line, the right border of the container is thickened. This PR must be reversed at once. As I said in my previous post, who has the issue with the lack of space in categories, either makes a personal change or looks for that module which uses JavaScript in the resizing of the container by pulling the gray edge to its right.

fballiano:
I don't see what you see but I'm working on a fix

fballiano:
done, please check https://github.com/OpenMage/magento-lts/pull/2251

loekvangool:
Strange, I don't remember running into these issues. I'm not interested in pursuing this old issue though, let's leave it to rest.

fballiano:
It was merged so we can't leave it as it is

ADDISON74:
This extension seems to do the resize of the category tree in a better way than using fixed value in CSS files.

https://github.com/yireo-magento1/Yireo_CategoryTreeResizable

It can be customized by removing some JS files as long as JQuery v1 is shipped with OpenMage.

Anyway I will analyze this issue and I will come back later with a better default value for the width of the container.

PR #1320, opened 2020-11-24
Minor cron housekeeping

woutersamaey:
Catch Throwables, fixed possibly undefined var, use __DIR__ instead of dirname(__FILE), stricter type checking

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -25,7 +25,7 @@
  */
 
 // Change current directory to the directory of current script
-chdir(dirname(__FILE__));
+chdir(__DIR__);
 
 require 'app/bootstrap.php';
 require 'app/Mage.php';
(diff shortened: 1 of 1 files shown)


kkrieger85 (approved):
LGTM 
thanks

Flyingmana:
would it be ok to move this to version 20?
Its probably save, but its also a behavior change

woutersamaey:
Which behaviour change do you mean? With PHP7, we need to be using Throwables more often.
But okay to move to 20 if you feel this is warranted.

colinmollenhour:
Before: 'PHP Fatal error:  Uncaught Error: ...' is handled by PHP engine and logged according to `error_log` configuration in php.ini

After: Error is printed to stdout and is not logged to PHP `error_log` and not re-thrown.

I think the latter is probably better for simple installations, but this could be unexpected for sysadmins/devops and perhaps make it difficult to capture these errors in other tools like Sentry (e.g. with Firegento_Logger).

Suggestion: Catch the throwable but rethrow it if it is an error..

```php
try {
    //...
} catch (Exception $e) {
    Mage::printException($e);
    exit(1);
} catch (Error $e) {
    Mage::printException($e);
    throw $e;
}
```

I think without rethrowing it this will effectively break integrations that register additional error handlers, although I haven't tested this theory yet so could be wrong somehow.

woutersamaey:
What should we do @colinmollenhour  and @Flyingmana ? I prefer my version, but can compromise.

colinmollenhour:
@woutersamaey Here is an example, I use Sentry to log/report/alert unexpected errors and Sentry hooks into PHP by setting the exception handler which receives all uncaught Throwables: https://github.com/getsentry/sentry-php/blob/cd27630daa2491f18f826118a7a8860d0cc6e9f4/src/ErrorHandler.php

So I'm 99% sure this patch would break that capability and these errors would quit getting logged to Sentry or anywhere and also the return code would be a success instead of an error. If the exception is not caught or is caught and re-thrown then PHP will exit with a failure return code and also print out an "Uncaught Exception" error making it easy to detect errors and see and log the details and also not interfere with the custom exception handlers registered by tools like Sentry.

woutersamaey:
I'm using a tool similar to Senty, but my integration hooks into Mage::logException() and Mage::printException() using a core edit.
I've understood your point and have amended this PR. Should be OK.

I did change the last `catch(Error)` into a `catch(GThrowable)` as it is more generic. Should be fine, I believe.

colinmollenhour:
Throwing after `Mage::printException` will have no effect since `Mage::printException` calls `die()` so maybe it should be `Mage::logException()` and then re-throw.

kiatng (approved):

PR #1322, opened 2020-11-27
Move orig data to abstract model (according to Magento 2.x)

Sekiphp:
### Description (*)
Move `origData` property and related methods to `mage_Core_Model_Abstract` according to Magento 2.x.

Related M2 file: https://github.com/magento/magento2/blob/fbfac3e51730abe2b79c6072544bc9742fcf7606/lib/internal/Magento/Framework/Model/AbstractModel.php#L56

**These methods are not used in any direct child of `Varien_Object`:**
`$ grep -R -B10000 -A10000 -P 'OrigData|dataHasChangedFor' . | grep -F "extends Varien_Object"`

**Output:**
`./app/code/core/Mage/Core/Model/Abstract.php-abstract class Mage_Core_Model_Abstract extends Varien_Object`

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1325, opened 2020-12-02
Move orig data to abstract model (according to Magento 2.x)

Sekiphp:
### Description (*)
Move `origData` property and related methods to `mage_Core_Model_Abstract` according to Magento 2.x.

Related M2 file: https://github.com/magento/magento2/blob/fbfac3e51730abe2b79c6072544bc9742fcf7606/lib/internal/Magento/Framework/Model/AbstractModel.php#L56

**These methods are not used in any direct child of `Varien_Object`:**
`$ grep -R -B10000 -A10000 -P 'OrigData|dataHasChangedFor' . | grep -F "extends Varien_Object"`

**Output:**
`./app/code/core/Mage/Core/Model/Abstract.php-abstract class Mage_Core_Model_Abstract extends Varien_Object`

### DIFF
--- a/app/code/core/Mage/Core/Model/Abstract.php
+++ b/app/code/core/Mage/Core/Model/Abstract.php
@@ -59,6 +59,13 @@ abstract class Mage_Core_Model_Abstract extends Varien_Object
     protected $_eventObject = 'object';
 
     /**
+     * Original data that was loaded
+     *
+     * @var array
+     */
+    protected $_origData;
+
+    /**
      * Name of the resource model
      *
      * @var string
@@ -112,6 +119,51 @@ abstract class Mage_Core_Model_Abstract extends Varien_Object
     protected function _init($resourceModel)
     {
         $this->_setResourceModel($resourceModel);
+    }
+
+    /**
+     * Get object loaded data (original data)
+     *
+     * @param string $key
+     * @return mixed
+     */
+    public function getOrigData($key = null)
+    {
+        if (is_null($key)) {
+            return $this->_origData;
+        }
+        return isset($this->_origData[$key]) ? $this->_origData[$key] : null;
+    }
+
+    /**
+     * Initialize object original data
+     *
+     * @param string $key
+     * @param mixed $data
+     * @return $this
+     */
+    public function setOrigData($key = null, $data = null)
+    {
+        if (is_null($key)) {
+            $this->_origData = $this->_data;
+        } else {
+            $this->_origData[$key] = $data;
+        }
+        return $this;
+    }
+
+    /**
+     * Compare object data with original data
+     *
+     * @param string $field
+     * @return boolean
+     */
+    public function dataHasChangedFor($field)
+    {
+        $newData = $this->getData($field);
+        $origData = $this->getOrigData($field);
+
+        return $newData != $origData;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Sekiphp:
Opened PR to another branch @midlan @Flyingmana @kkrieger85 

#1322

midlan (approved):

kiatng (approved):

colinmollenhour:
While we're refactoring code why not put it in a trait so that if someone needs to use it in a class that extends Varien_Object but not Mage_Core_Model_Abstract it can be easily added?

midlan:
@colinmollenhour Why not? Let me explain.

From my view, the origData is strongly linked to object loading. Even the doc says it is linked to object loading:
> Get object loaded data (original data)

Nowadays I assume there is simply no programmer who would like reinvent new own object model/orm and use such trait, since there are plenty of object models already invented laying on github.

Magento 2 team is thinking the same (the didn't created the trait). And there does not exist such feature request in M2.

Another thing is that once you divide the class into class + trait, you must keep the back compatibility of both of them the whole life of openmage major version. I would divide it only in the case someone will send feature request that he/she needs it (what probably won't happen as I written above). In other words: do not make things complicated, only because someone _may_ need it.

I am doing the same. Every pull request I open is bugfix or thing that I really run into or used in my Magento/OpenMage instance.

joshua-bn:
@midlan makes a very good point... don't add features because someone _might_ want them.

colinmollenhour:
I'm fine with it as is, just thinking that if in the unlikely case that this does break BC for someone who used _origData on  an instance of Varien_Object that did not extend Mage_Core_Model_Abstract the remedy would be to simply add the trait rather than copy each method over to their custom model.

midlan:
@Flyingmana correctly changed the base branch to 20.x so the BC break is acceptable during major upgrade.

In the case it really breaks something, I would first recommend to change the parent to `Mage_Core_Model_Abstract` if applicable. If not, the trait is way.

Flyingmana (approved):

fballiano:
I think this was a mistake, how do we solve the phpstan error on https://github.com/OpenMage/magento-lts/pull/2206 https://github.com/OpenMage/magento-lts/runs/6803200822?check_suite_focus=true ? that are some parts of the core that are calling dataHasChangedFor() but they can expect Varien_Object objects.

What are we doing now with those?

kiatng:
There are 3 errors:

```
  ------ ----------------------------------------------------------------- 
  Line   Catalog/Model/Category/Indexer/Product.php                       
 ------ ----------------------------------------------------------------- 
  199    Call to an undefined method Varien_Object::dataHasChangedFor().  
  200    Call to an undefined method Varien_Object::dataHasChangedFor().  
 ------ ----------------------------------------------------------------- 
 ------ ----------------------------------------------------------------- 
  Line   Catalog/Model/Resource/Product/Attribute/Backend/Urlkey.php      
 ------ ----------------------------------------------------------------- 
  66     Call to an undefined method Varien_Object::dataHasChangedFor().  
 ------ ----------------------------------------------------------------- 
```

For the first two, https://github.com/OpenMage/magento-lts/blob/c83207b4bbe6a3a4ee79fa2db0179706567dd7dc/app/code/core/Mage/Catalog/Model/Category/Indexer/Product.php#L190-L202

Line 194 should be

```php
             /** @var Mage_Catalog_Model_Product $product */ 
```

For the third error, the file is deprecated: https://github.com/OpenMage/magento-lts/blob/c83207b4bbe6a3a4ee79fa2db0179706567dd7dc/app/code/core/Mage/Catalog/Model/Resource/Product/Attribute/Backend/Urlkey.php#L34-L36

That, I am not sure what can be done.

PR #1350, opened 2020-12-14
Trim values from XML so auto-formatting our XML does not break the autoloader.

woutersamaey:
If you auto-format an XML file (like in PHPStorm), the value/class name can end up on a new line. This breaks the loading of the class, because Magento tries to create a class that starts with a newline and spaces. This simple trim avoids formatting issues.

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1288,6 +1288,8 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
                 }
             }
         }
+        
+        $className = trim($className);
 
         // Second - if entity is not rewritten then use class prefix to form class name
         if (empty($className)) {


woutersamaey:
This PR fixes XML that looks like this:
```
<adminhtml>
   <rewrite>
      <catalog_product_edit_tab_inventory>
         Storefront_MultiStock_Block_Adminhtml_Catalog_Product_Edit_Tab_Inventory
      </catalog_product_edit_tab_inventory>
   </rewrite>
</adminhtml>
```

Before the XML would need to be 
```
<adminhtml>
   <rewrite>
<catalog_product_edit_tab_inventory>Storefront_MultiStock_Block_Adminhtml_Catalog_Product_Edit_Tab_Inventory</catalog_product_edit_tab_inventory>
   </rewrite>
</adminhtml>
```

But auto-formatting in PHPStorm and other IDEs puts the contents on different lines because the node name / value is too long.

colinmollenhour (approved):
I wish it was possible to apply a broader fix but I don't know of an elegant way to do that. Thanks for the PR!

Flyingmana:
some more information on this topic. XML has some informations related handling this in its standard and a special attribute to signal the intent for whitespace handling.
http://www.xmlplease.com/xml/xmlspace/

Why is this relevant? Because this means we dont violate any xml related standard by doing a trim.


For further going changes, there is some "normalize-space()" function mentioned, which can do this.
There are also mentioned XSLT methods to remove whitespaces `<xsl:strip-space element="*"/>`   `<xsl:output indention="no"/>`

In PHP there is DOMDocument::normalizeDocument ( ) which may be something related to this (but I did not check this further)

Anyway, for this specific case, Iam good with the solution for now.

Flyingmana (approved):

PR #1355, opened 2020-12-14
Removed unused variable

midlan:
### Description (*)

Just removed unused variable.

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Email/Template.php
+++ b/app/code/core/Mage/Adminhtml/Model/Email/Template.php
@@ -53,8 +53,6 @@ class Mage_Adminhtml_Model_Email_Template extends Mage_Core_Model_Email_Template
         }
         $paths = array();
 
-        $configSections = Mage::getSingleton('adminhtml/config')->getSections();
-
         // find nodes which are using $templateCode value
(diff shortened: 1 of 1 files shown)


kiatng (dismissed):

Flyingmana:
could have unintended side effects on internal initialization and also about when the event "adminhtml_init_system_config" is called

midlan:
@Flyingmana you are right about the event `adminhtml_init_system_config`, we should check if there is observers for it

About the init, seems safe to me. It is simple [lazy-loading pattern](https://en.wikipedia.org/wiki/Lazy_loading), I don't see how it may behave differently when lazy-loaded from different model.

luigifab (dismissed):
PayPal use the `adminhtml_init_system_config` event with `paypal/observer::loadCountryDependentSolutionsConfig`.
But I didn't see any emails added with `Mage_Paypal`.

The only configuration about email that I found is: `email_confirmation`, used in API request.

Flyingmana:
I moved it to the 20.0 branch, because even if it looks ok, I want to not risk issues, and its not a bugfix, so not necessarily needed in the 1.9.4.x branch.

PR #1359, opened 2020-12-18
Removed the word "Manage" from the Admin menu so all items are named are consistently

woutersamaey:
I never understood why Magento added the word "Manage" for some menu items and not for others. Isn't it obvious you're managing everything in the Admin?
i.e. "Manage Products" and "Manage Categories", but for pages it's just "Pages" and not "Manage Pages".
Yeah, not so exciting PR, but it has bothered me for a long time ;-)


sreichel (changes requested):
This would break all locale packs. Maybe fix this in translation files?

woutersamaey:
Not really, because the terms without the prefix “Manage “ should already be in the locale packs. 
I can’t imagine there is no translation for words like “Products”, “Customers”, etc. 

The thought crossed my mind, but didn’t think it would be a problem.

sreichel:
Can you please verify that all translations still work? Unused translations should be removed from csv.

woutersamaey:
@sreichel There's no need to change the CSV files, because these labels are still used in other places. Both labels with and without the "Manage " prefix are in use by Magento.

woutersamaey:
I did a few "search everywhere"s and this came up clean.

woutersamaey:
@sreichel can we can this approved?

luigifab:
Perhaps we can also rename for example _Manage Products_ to _Products_ like _Orders_?

image image

sreichel (approved):

ADDISON74:
@luigifab - From what I see he has already done that. All menu links starting with the word Manage have been changed.

@woutersamaey - Basically the change only targets the name of the links but the names create initially by the Magento team stay on inside the PHP files. Did you do a check what happens when switching to a language other than English? It may be mandatory to add some of them into the locale files.

luigifab:
Yes, sorry, not in Menu, but in page title (`h2.icon-head`).

ADDISON74:
@luigifab - You are right. The change should not be limited to the name of the links in menus but also in pages. All loaded pages that have the word "Manage" in the link have it also in page content (e.g. <h3 class = "head-system-store"> Manage Stores </h3>).

In conclusion, either we make all changes, including CSV translations, PHP files or leave it as it is now.

woutersamaey:
I made the changes throughout the code as well.

ADDISON74:
@woutersamaey - Please check this file too. I think you missed it.

https://github.com/OpenMage/magento-lts/blob/17128ba430b2d372da7901a22ee0f083b47d66e4/app/design/adminhtml/default/default/template/catalog/product/tab/inventory.phtml#L41

woutersamaey:
@ADDISON74 Nope. I have left the pharse "Manage Stock" untouched on purpose. It is a product attribute and does have meaning: Magento should "Manage" the stock for the product. It's not about the user managing something.
This is the only exception I have made.

ADDISON74:
@woutersamaey  - Thank you for your clarification.

As an observation, in OpenMage 20 the /app/design/frontend/default directory was removed but it was kept in version 19. It is good that this PR is wanted to be merged into OpenMage 19 considering that the change concerns PHTML files in this directory  (e.g. iphone).

woutersamaey:
@ADDISON74 out of curiosity. How would this PR get into 20.0?

ADDISON74:
As far as I know what is added to version 19 will be in version 20. If those who use a language other than English in Backend are dissatisfied with the changes, which in my opinion are harmless but which visually improve the menus, then it should be merged into OpenMage 20.

luigifab (approved):

tobihille (approved):

fballiano:
This PR now has conflicts and can't be modified by maintainers.

ADDISON74:
Maybe we need to discuss a new rule regarding these PRs, there will certainly be others in this situation. If the author does not respond within X days, and we don't have editing rights that PR is closed and another one is opened by a person active at that moment.

fballiano:
> Maybe we need to discuss a new rule regarding these PRs, there will certainly be others in this situation. If the author does not respond within X days, and we don't have editing rights that PR is closed and another one is opened by a person active at that moment.

agree 100%, I'd say a couple of weeks?

ADDISON74:
Two weeks is more than enough time to get a reply from the authors.

I will assign myself several PRs so that I can know about them at any time for testing. I hope we can manage to reduce their number below 100 by the end of the year. There would still be the Issues and Discussions/Ideas section to analyze.

ADDISON74:
I looked over this PR and it's fine, I approve it, but I noticed that it has conflicts that need to be resolved and the branch it comes from is read-only. What a pity.

sreichel:
This change breaks every locale pack. No!

E.g. https://github.com/riconeitzel/German_LocalePack_de_DE translates ...

```
"Manage Products","Artikel verwalten"
```

sreichel:
If you want to change wording, please in CSV only. Extensions may rely on that php string with custom translation.

ADDISON74:
I liked this PR if some rules were applied. Let's discussed more in detail if it was a good idea or not to close this PR here #2590.

luigifab:
I disagree. Like #2147, this PR reduce the number of translations of OpenMage. Moreover if your language pack translate "Mange Products" or "Products Section", it will probably also translate "Products" and "Product".

If not we must revert 2147.

luigifab:
:'(

ADDISON74:

PR #1361, opened 2020-12-20
Inventory saving improvement

boesbo:
### Description (*)
If I perform mass actions in the inventory, in addition to not wanting to trigger the reindex at each save I don't want an index process log to be saved, since a manual index will be done later.
This significantly improves performance during mass action.

Send a massaction on many products by updating inventory attributes.


boesbo:
The reindex works by looking at pending events, it doesn't index everything. So it is wrong to remove the event log.

You could think of changing the reindex logic after the mass action but for the moment I don't think it's appropriate.

PR #1380, opened 2021-01-08
Add check to avoid js error

luigifab:
### Description

This PR add a check to avoid a javascript fatal error when I merge and minify all backend javascript files into one single file.

screen

OpenMage 20.0.5 / PHP 7.4.11 - 8.0.0

### Contribution checklist

### DIFF
--- a/js/mage/adminhtml/product/composite/configure.js
+++ b/js/mage/adminhtml/product/composite/configure.js
@@ -54,7 +54,9 @@ ProductConfigure.prototype = {
      * Initialize object
      */
     initialize: function() {
-        this._initWindowElements();
+        if ($("product_composite_configure")) {
+            this._initWindowElements();
+        }
     },
 
     /**


colinmollenhour:
Does the init happen later or should it be added to an event observer like onload?

luigifab:
Function `_initWindowElements` via `initialize` is already triggered on page load (at the end of the file).
Perhaps it is better to add the check here:

```js
Event.observe(window, 'load',  function() {
    productConfigure = new ProductConfigure();
});
```

fballiano (approved):

Flyingmana:
in any case I would think it needs to wait for it to exist, and not just skipping it?
I think the current place is already ok to add the check, but maybe add a retry logic with setTimeout.

luigifab:
For my and standard usage, a `setTimeout` is useless. Because if `product_composite_configure` does not exist on page load, it will never exist.

Flyingmana (approved):

PR #1382, opened 2021-01-09
Remove help cell on calendar

luigifab:
### Description

This PR remove the help cell on calendar (on top left). I think it is useless. Moreover this change remove some translations.

date

OpenMage 20.0.5 / PHP 8.0.0

### Manual testing scenarios
Go to sales / order, and click on the date picker on purchased on column.

### Contribution checklist


Flyingmana:
https://www.gnu.org/licenses/lgpl-3.0.html
4 a)
```
a) Give prominent notice with each copy of the Combined Work that the Library is used in it and that the Library and its use are covered by this License.
```

so, we are very likely not allowed to remove this, as this is used to show license related information, which are apparently required.

PR #1386, opened 2021-01-10
Add some specific debugging hint to email queue recipient empty

seansan:
No idea what works best. Ideally I would want to log ID/SUBJECT/STORE so I know where to start looking if this happens. If you propose otherwise, then please adjust the pull request directly and push. Only small win here is a little more specific debugging.

https://stackoverflow.com/questions/53581543/exception-mage-core-exception-with-message-message-recipients-data-must-be-se

1. ...
2. ...


app/code/core/Mage/Core/Model/Email/Queue.php
```diff
@@ -96,7 +96,8 @@ protected function _afterSave()
     protected function _beforeSave()
     {
         if (empty($this->_recipients) || !is_array($this->_recipients)) {
-            Mage::throwException(Mage::helper('core')->__('Message recipients data must be set.'));
+            $error = Mage::helper('core')->__('Message recipients data must be set.');
+            Mage::throwException($error . " Msg Id:" . $this->getId());
```
kiatng, line 100: Were you able to get the ID? I got a feeling that it's probably `null`.

seansan:
In this case the ID and subject exist. The method checks for empty
recipients ... which is a good thing to do ... bt if it occurs you would
want to know "which" message entity has this problem

In this (rare) case we had 1 data row in queue but not corresponding data
in the linked recipients table

kiatng (dismissed):

fballiano:
I fixed conflicts and slightly changed the message so that doesn't create a problem with translations. I think this can be merged.

fballiano (approved):

ADDISON74 (approved):

fballiano:

PR #1391, opened 2021-01-12
Fixes core and lib issues for PHP 8.0 compatibility

lamskoy:
### Description (*)

1. Adds check for PHP 8 and phpseclib: if it's PHP 8.0+ then disallow inline crypt mode
2. Fixes each() function calls - not available in PHP 8.0 anymore
3. Fixes create_function() function calls - not available in PHP 8.0 anymore
4. Fixes lib/Zend/Xml/Security.php: version_compare doesn't support 'lt' 'gte' 'gt' compare operators anymore in PHP 8.0
5. Inside app/code/core/Mage/Catalog/Model/Product/Type/Configurable.php fixes wrong calls getProduct($product) for objects:

a) Mage_Sales_Model_Quote_Item_Option has getProduct() method which is not taking any parameters!
b) Mage_Catalog_Model_Product_Configuration_Item_Option is Varien_Object derivative, and "product" is just field containing one product data, not a collection inside. (It's causing issue under PHP 8.0 inside Varien_Object::getData() method passing instance of Catalog_Model_Product as $key argument)

1. Add to cart some configurable product on frontend under PHP 8.0

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php
+++ b/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php
@@ -124,7 +124,7 @@ class Mage_Adminhtml_Model_System_Config_Source_Admin_Page
 
         uasort($parentArr, array($this, '_sortMenu'));
 
-        while (list($key, $value) = each($parentArr)) {
+        foreach ($parentArr as $key => $value) {
             $last = $key;
         }
         if (isset($last)) {
--- a/lib/Zend/Config/Yaml.php
+++ b/lib/Zend/Config/Yaml.php
@@ -289,7 +289,7 @@ class Zend_Config_Yaml extends Zend_Config
     {
         $config   = array();
         $inIndent = false;
-        while (list($n, $line) = each($lines)) {
+        foreach($lines as $n => $line) {
             $lineno = $n + 1;
 
             $line = rtrim(preg_replace("/#.*$/", "", $line));
--- a/lib/Zend/Http/UserAgent/Features/Adapter/TeraWurfl.php
+++ b/lib/Zend/Http/UserAgent/Features/Adapter/TeraWurfl.php
@@ -88,7 +88,7 @@ class Zend_Http_UserAgent_Features_Adapter_TeraWurfl implements Zend_Http_UserAg
             if (!is_array($group)) {
                 continue;
             }
-            while (list ($key, $value) = each($group)) {
+            foreach($group as $key => $value) {
                 if (is_bool($value)) {
                     // to have the same type than the official WURFL API
                     $features[$key] = ($value ? 'true' : 'false');
--- a/lib/Zend/XmlRpc/Value.php
+++ b/lib/Zend/XmlRpc/Value.php
@@ -486,13 +486,15 @@ abstract class Zend_XmlRpc_Value
      */
     protected static function _extractTypeAndValue(SimpleXMLElement $xml, &$type, &$value)
     {
-        list($type, $value) = each($xml);
+        $value = reset($xml);
+        $type = key($xml);
 
         if (!$type and $value === null) {
             $namespaces = array('ex' => 'http://ws.apache.org/xmlrpc/namespaces/extensions');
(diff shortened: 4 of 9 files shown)


lamskoy:
Unit tests seems to be broken:
github.GithubException.GithubException: 403 {"message": "Resource not accessible by integration", "documentation_url": "https://docs.github.com/rest/reference/checks#create-a-check-run"}

Flyingmana:
sorry about the unittest, the action has a permission issue for sending the results as a comment to the PR. Have modified to allow this step to fail in the future without marking the run as failed.

app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php
```diff
@@ -124,7 +124,7 @@ protected function _buildMenuArray(Varien_Simplexml_Element $parent=null, $path=
 
         uasort($parentArr, array($this, '_sortMenu'));
 
-        while (list($key, $value) = each($parentArr)) {
+        foreach ($parentArr as $key => $value) {
```
kiatng, line 127: Since PHP7.3, instead of loop and 6 lines of code, it can be simplified with:
```php
if ($parentArr) {
    $parentArr[array_key_last($parentArr)]['last'] = true;
}

Comment (Flyingmana) [reply app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php:127]: Our composer file says, we still support 7.0.
And for migration reasons we should stay with compatibility for 7.2.

Comment (lamskoy) [reply app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php:127]: Yeah, PHP7.2 support is important

**Comment by jfksk** [comment]
It seems phpseclib [removed create_function](https://github.com/phpseclib/phpseclib/commit/cc3ad790589535a731d8cedb1d8b4aff808c20cd) altogether. Maybe we could simplify that a little and just follow phpseclib here until we eventually update it?

**Comment by lamskoy** [comment]
Have reverted phpseclib changes. Please review

**Review by kiatng — APPROVED** [review]
Thanks.

File: lib/Zend/Config/Yaml.php [hunk 2]
```diff
@@ -289,7 +289,7 @@ protected static function _decodeYaml($currentIndent, &$lines)
     {
         $config   = array();
         $inIndent = false;
-        while (list($n, $line) = each($lines)) {
+        foreach($lines as $n => $line) {
```
S0FTWEX, line 292: Add space after 'foreach' to meet PSR-4 standards:
```foreach ($lines as $n => $line) {```

File: lib/Zend/Http/UserAgent/Features/Adapter/TeraWurfl.php [hunk 3]
```diff
@@ -88,7 +88,7 @@ public static function getAllCapabilities(TeraWurfl $wurflObj)
             if (!is_array($group)) {
                 continue;
             }
-            while (list ($key, $value) = each($group)) {
+            foreach($group as $key => $value) {
```
S0FTWEX, line 91: Add space after 'foreach' to meet PSR-4 standards:
```foreach ($group as $key => $value) {```

File: lib/Zend/XmlRpc/Value.php [hunk 4]
```diff
@@ -486,13 +486,15 @@ protected static function _createSimpleXMLElement(&$xml)
      */
     protected static function _extractTypeAndValue(SimpleXMLElement $xml, &$type, &$value)
     {
-        list($type, $value) = each($xml);
+        $value = reset($xml);
```
Flyingmana, line 489: technically this is a logic change, as `each()` returns the current, and `reset()` the first.
Also `each()` moves the internal coursor.
But from the usage here it seems to not matter, and also makes the result more predictable, so Iam ok with leaving it as it is.

Flyingmana (approved):

jfksk (approved):

fballiano:
should we maintain ZF1 by ourselves? I've found this project https://github.com/Shardj/zf1-future/ which I think it's quite interesting, it would have its own challanges but...

PR #1396, opened 2021-01-14
Remove the need for filter callbacks to be in the same class as the grid. 

woutersamaey:
This opens up adding/changing grid columns from observers and having filter callbacks in other places (like in that observer). No need to override the grid each time so a filter callback can be put there.

Some background: I wanted to add a column to a grid from an observer.
The column is a special case that needs a `filter_condition_callback` so filtering can work.

Problem: The callback isn't used if it its not defined in the grid class itself, making it impossible to add this column and forces me to rewrite the grid block. If I have multiple of these columns, I need to extend the grid each time, causing rewrite hell.

Removing this simple check opens up callback to be put anywhere we want, and should not be a problem.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -464,7 +464,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
     {
         if ($this->getCollection()) {
             $field = ( $column->getFilterIndex() ) ? $column->getFilterIndex() : $column->getIndex();
-            if ($column->getFilterConditionCallback() && $column->getFilterConditionCallback()[0] instanceof self) {
+            if ($column->getFilterConditionCallback()) {
                 call_user_func($column->getFilterConditionCallback(), $this->getCollection(), $column);
             } else {
                 $cond = $column->getFilter()->getCondition();


sreichel:
Same as #887

sreichel:
Put you callback methods into a class, that extens `Mage_Adminhtml_Block_Widget_Grid` should fix it. No need for rewrites.

woutersamaey:
@sreichel this adds a lot of overhead. Any reason why my fix is so bad / unacceptable to you? Why can't we lift this restriction?

colinmollenhour:
@woutersamaey Are you not already overriding the grid class in order to add a new column?

woutersamaey:
No, I am adding the column from an observer and have the callback there as well. Is may not be the best place to put the callback, but whatever. That doesn’t affect this PR. 
The event that I’m using is custom though.

colinmollenhour:
@sreichel What is the objection to this PR, specifically? It seems like a valid need to complement the events which allow you to avoid using rewrites. Unless the column definition can be defined by users or is stored is some serialized format somewhere I don't see that this is a security risk.

woutersamaey:
My point exactly. Thanks @colinmollenhour

sreichel:
> Still not sure why this matters though since the callback just needs to be a method that accepts two arguments and returns an array.

I don't know why it has been added, but should we revert changes made via SUPEE, just to (better) fit custom code?

You still can add colums without rewrites.  Just move the callbacks to a propper class.

sreichel:
> 
> 
> @sreichel this adds a lot of overhead. Any reason why my fix is so bad / unacceptable to you? Why can't we lift this restriction?

Its no overhead to move your code to antother class. Nothing else. No rewrites. If this would avoid rewrites, I'd be first to approve ;)

woutersamaey:
Creating a class of the Grid type is more overhead no?

Basically you don’t want to make the change because of the SUPEE and SUPEE changes cary more weight than other code?
This sounds like a bad precedent.

woutersamaey:
Can’t we at least put this to a vote?

woutersamaey:
Also, if we don’t want to make this change, we should throw an exception stating that the callback should implement this class if it doesn’t. I lost quite a bit of time before I discovered that the callback has limitations.

sreichel:
> Creating a class of the Grid type is more overhead no?

We are talking about ONE single class that needs to be added. As you say `[Observer] Is may not be the best place to put the callback`.

(you can also do this .... `Your_Module_Model_Observer extends Mage_Adminhtml_Block_Widget_Grid` ... not nice, but should work)

> Basically you don’t want to make the change because of the SUPEE and SUPEE changes cary more weight than other code?
> This sounds like a bad precedent.

Basically i think 3rd party code should fit core, not visa verse.

This (official) change was a BC break indeed, but we shouldn't undo it just because we're too comfortable adjusting our code.

> Can’t we at least put this to a vote?

Haha, no, this is no democracy! :P (@Flyingmana @colinmollenhour, @kkrieger85, @tmotyl, ...)

sreichel (changes requested):
Better fix you code.

Flyingmana:
A few things...

As this additional check for the class was added by a SUPEE patch, we should look closer here, what kind of attack this prevents before removing it. The supee-11219 https://magento.com/security/patches/supee-11219 Patch fixes multiple Remote Code Executions, so its likely serious.
Maybe it is related to call_user_func() also allowing strings to be used for callbacks.
Limiting the Class to instances of `self` also feels provisionally and would lead to new security holes, if 3th party vendors improvise by just extending from Grid to implement base functionality which could be misused. (So we probably even need to give them ways to not create new classes for this)

But in general I support changes which reduce the number of needed classes. PHP supports closures/anonymous functions now for a while, and they are specifically designed to be used for callbacks.
So maybe instead of removing the limit to `self`, we could add another case which allows a Closure

something like (only an idea, not tested if the syntax is actually correct)
```
            if ($column->getFilterConditionCallback() && $column->getFilterConditionCallback()[0] instanceof self) {
                call_user_func($column->getFilterConditionCallback(), $this->getCollection(), $column);
            } else if ($column->getFilterConditionCallback() instanceof Closure ) {
                $column->getFilterConditionCallback()($this->getCollection(), $column);
            } else {
```

This should prevent the typical admin attacks which inject calls as strings somewhere. (maybe it was one of the layout.xml based RCEs?)
But we should still check this more rigorous, as this touches a security sensible area.


> Also, if we don’t want to make this change, we should throw an exception stating that the callback should implement this class if it doesn’t. I lost quite a bit of time before I discovered that the callback has limitations.

This is also good to keep in mind, maybe not an exception, but logging at least a warning when falling back despite a callback beeing set.


---

ok, now to @sreichel 
It was not ok to "just close" this one, it was also not ok for the referenced "duplicate" to be "just closed"
If you link to an alternative, and the contributor is happy with this alternative, closing would be ok.
But here you closed despite the contributor being unhappy with the alternative.

I want to cite here a part of our Code of Conduct, which Maintainers should follow as a good role model.

> * Using welcoming and inclusive language
> * Being respectful of differing viewpoints and experiences
> * Gracefully accepting constructive criticism
> * Focusing on what is best for the community
> * Showing empathy towards other community members

Referencing to the SUPEE was correct here. And its also ok to not be in favor of a Change.
You are also allowed to Post as a comment, that you dont like the change and you are against it being merged.

Its also ok to just step back from PullRequest you do not support. no Maintainer is bound to be responsible for all PullRequests.

> Basically i think 3rd party code should fit core, not visa verse.

Technically we dont have a written guideline here, and apparently we need one, because I see it the opposite way.
3th partys are the Heart💜 of OpenMage, they are the users, and the ones keeping it alive by helping and supporting us.
We should have an open ear for everything which makes their lives easier.

> > Can’t we at least put this to a vote?
> 
> Haha, no, this is no democracy! :P (@Flyingmana @colinmollenhour, @kkrieger85, @tmotyl, ...)

Such ridiculing commentary is not ok of you @sreichel 
Even if you did edit it afterwards, people see this, and its not funny.
Especially as we are definitely trying to act as democratic as possible.
We do this on a bigger level with our RFCs.
And we did it a few times on controversial PullRequests in the past.

> Better fix your code.

This sounds very dismissively, does not contain any explanation, no suggestions to improve the code.


Posting this here Public for transparency and to even out what was written here Public already.

woutersamaey:
@Flyingmana the Closure support looks awesome. It solves my issue and is a way forward at the same time.
I'm willing to change, test & finalize this PR if we're all in agreement.

Also thanks for stepping in. I got a very bad feeling and flashbacks to Magento 2 rudeness I was not expecting to see here.

joshua-bn:
I made this same change a while back. Been running in production for over a month.

sreichel:
> A few things...

**First of all ... @woutersamaey please excuse my statement about democracy or even the short answers. It was only meant as a joke! I noticed myself that it wasn't that good, so I edited it.** 

> It was not ok to "just close" this one, it was also not ok for the referenced "duplicate" to be "just closed"

In all this time, the two PRs are the only ones I have closed on my own authority. In the past, some PRs were also closed because it could be implemented without rewrities or adjustments to the code. (We should keep this.)

> > Basically i think 3rd party code should fit core, not visa verse.
> 
> Technically we dont have a written guideline here, and apparently we need one, because I see it the opposite way.
> 3th partys are the Heartpurple_heart of OpenMage, they are the users, and the ones keeping it alive by helping and supporting us.
> We should have an open ear for everything **which makes their lives easier**.

I'm all for anything that makes life easier. Don't ask how long I sat on DOCBlocks/CS, support for PhpStorm, or the still open PR regarding "event_prefixes". I think new "events" that save rewrites are great. Often requested features are nice too. Performance optimizations are great. But making a change here to make it fit your own code better is like the tail wagging the dog.

On topic ...

> >     Better fix your code.
> 
> This sounds very dismissively, does not contain any explanation, no suggestions to improve the code.

I think I have already made that suggestion. Use a class that extends `Mage_Adminhtml_Block_Widget_Grid `!
It was said that rewrites are needed. This is wrong.
It was said that there is a large overhead. That is also wrong in my opinion. Creating ONE file/class is not reasonable overhead?

> This is also good to keep in mind, maybe not an exception, but logging ...

In other places where `instanceof` is used, an exception is usually thrown. Wouldn't that be the simplest solution?

sreichel:
> I got a very bad feeling and flashbacks to Magento 2 rudeness I was not expecting to see here.

This was not intended and should not cast a bad light on the project. Sorry again too all.

PR #1404, opened 2021-01-19
[BUGFIX] Count doesn't work with group by columns. This fix keeps the group by

dbachmann:
Some extensions like the Amasty Product Alerts Reports didn't worked correct because they use some group by statements in their queries. I've attached two screenshots.

Here is a stock alerts report before the fix
without-fix

Here is a stock alert report after the fix
with-fix

The second commit contains a fix for the onepage checkout. The base template used a fieldset as wrapper for payment methods in checkout. That is not correct because payment methods are defined as dt and dd.

### DIFF
--- a/lib/Varien/Data/Collection/Db.php
+++ b/lib/Varien/Data/Collection/Db.php
@@ -242,8 +242,14 @@ class Varien_Data_Collection_Db extends Varien_Data_Collection
         $countSelect->reset(Zend_Db_Select::LIMIT_OFFSET);
         $countSelect->reset(Zend_Db_Select::COLUMNS);
 
-        $countSelect->columns('COUNT(*)');
-
+        if (count($this->getSelect()->getPart(Zend_Db_Select::GROUP)) > 0) {
+            $countSelect->reset(Zend_Db_Select::GROUP);
+            $countSelect->distinct(true);
+            $group = $this->getSelect()->getPart(Zend_Db_Select::GROUP);
+            $countSelect->columns("COUNT(DISTINCT " . implode(", ", $group) . ")");
+        } else {
+            $countSelect->columns('COUNT(*)');
+        }
         return $countSelect;
     }
 
--- a/lib/Varien/Data/Collection/Db.php
+++ b/lib/Varien/Data/Collection/Db.php
@@ -246,6 +246,9 @@ class Varien_Data_Collection_Db extends Varien_Data_Collection
             $countSelect->reset(Zend_Db_Select::GROUP);
             $countSelect->distinct(true);
             $group = $this->getSelect()->getPart(Zend_Db_Select::GROUP);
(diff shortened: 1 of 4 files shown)


Flyingmana:
Thank you, looks like a good addition.
But we should check if this has somehow the potential for an SQL injection (iam not good enough into the DB Abstraction to know if reusing the group part is safe there)

dbachmann:
The second commit contains a fix for the onepage checkout. The base template used a fieldset as wrapper for payment methods in checkout. That is not correct because payment methods are defined as dt and dd.

colinmollenhour:
Good observation, @Flyingmana, the group part is usually quoted (in `_renderGroup`):

```php
            $group = array();
            foreach ($this->_parts[self::GROUP] as $term) {
                $group[] = $this->_adapter->quoteIdentifier($term, true);
            }
            $sql .= ' ' . self::SQL_GROUP_BY . ' ' . implode(",\n\t", $group);
```

dbachmann:
@colinmollenhour What would you suggest or what would you change?

lib/Varien/Data/Collection/Db.php
```diff
@@ -242,8 +242,14 @@ public function getSelectCountSql()
         $countSelect->reset(Zend_Db_Select::LIMIT_OFFSET);
         $countSelect->reset(Zend_Db_Select::COLUMNS);
 
-        $countSelect->columns('COUNT(*)');
-
+        if (count($this->getSelect()->getPart(Zend_Db_Select::GROUP)) > 0) {
+            $countSelect->reset(Zend_Db_Select::GROUP);
+            $countSelect->distinct(true);
+            $group = $this->getSelect()->getPart(Zend_Db_Select::GROUP);
+            $countSelect->columns("COUNT(DISTINCT " . implode(", ", $group) . ")");
```
colinmollenhour, line 249: Untested...

```suggestion
            $group = array_map(function($token) {
                return $this->getSelect()->getAdapter()->quoteIdentifier($token, true);
            }, $group);
            $countSelect->columns("COUNT(DISTINCT " . implode(", ", $group) . ")");
```

dbachmann: Ok, another approach. I will test this one.

dbachmann: I've tested your suggestion and it works as expected. You never stop learning ;) Thank you @colinmollenhour

colinmollenhour:
Can you please remove the commits that are unrelated?

dbachmann:
@colinmollenhour Done

colinmollenhour (approved):
Looks good!

dbachmann:
@Flyingmana Could you take a look?

Flyingmana (approved):

sreichel:
@all-contributors add @dbachmann code

PR #1424, opened 2021-01-23
Ascending alphabetical sort for Attribute Set Name

ADDISON74:
A first modification allows the ascending alphabetical order of the list containing the attribute sets on the [Attrib. Set Name] column in Magento Backend -> Catalog -> Manage Products page.

It is useful for those who have a long list of sets of attributes. Magento sorts them by default as they were created, so it's very difficult to identify a particular set in a long list, especially since there is no possibility to filter the column.

Here is the result in production this store has almost 120 attributes sets:

screenshot_new

Another modification is useful when creating a New Product. The attribute sets list is sorted too. 

The third modification sorts Attribute Sets lists when creating a new attribute set:

And finally this modification sorts Unassigned Attributes list in Edit Attribute Page:

screenshot2


sreichel (commented):
Very useful! No time to test, but LGTM.

luigifab (approved):

sreichel (approved):

colinmollenhour (approved):

Issue #1428, opened 2021-01-26
Warnings when customer add configurable product to cart


luigifab:
With OpenMage 20.0.6 and PHP 7.4.6, when a customer add a configurable product to his cart, we encounter the following errors in logs:

_strpos() expects parameter 1 to be string, object given_

![image](https://user-images.githubusercontent.com/31816829/105856108-20f6dc80-5fe9-11eb-9098-9e03730952f7.png)

_Illegal offset type in isset or empty_

![image](https://user-images.githubusercontent.com/31816829/105856181-3409ac80-5fe9-11eb-9629-63e7b02eb102.png)

To fix the problem, [we have revert this change](https://github.com/OpenMage/magento-lts/commit/2fe965826a26bf6cfe36b4edba0d51d5e3f31491).

matteorossi-thespacesm:
I've got the same warning on OpenMage LTS 19.4.10. I've just made two pull requests #1431 and #1432 to fix it.

luigifab:
Also #1429

PR #1431, opened 2021-01-27
Update Interface.php

matteorossi-thespacesm:
Part of change to fix bug #1428 on OpenMage LTS 19.4.10

### Description (*)

Add method getProduct to interface

### Related Pull Requests

Add method getProduct to Mage_Catalog_Model_Product_Configuration_Item_Option class

1. Fixes OpenMage/magento-lts#1428 on OpenMage LTS 19.4.10

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Configuration/Item/Option/Interface.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Configuration/Item/Option/Interface.php
@@ -34,6 +34,13 @@
 interface Mage_Catalog_Model_Product_Configuration_Item_Option_Interface
 {
     /**
+     * Retrieve associated product
+     *
+     * @return Mage_Catalog_Model_Product
+     */
+    public function getProduct();
+    
+    /**
      * Retrieve value associated with this option
      *
      * @return mixed


kkrieger85 (approved):

fballiano:
modify the interface could create breaking changes?

kiatng:
> modify the interface could create breaking changes?

Yes, it would. Not only that, it will break existing classes which implement the interface, and we happen to have one such class:
https://github.com/OpenMage/magento-lts/blob/f71321b37a782bd167ef56de0c6e18e67645685d/app/code/core/Mage/Catalog/Model/Product/Configuration/Item/Option.php#L34-L44

This PR should not be merged.

Flyingmana (changes requested):
Blocked for reasons of BC incompatibility mentioned in the comments.

Alternatives:
- Add a new Interface for this method, which is used by the classes where this method is relevant
- not adding it (where is this method currently used from the outside, where we need it?)
- adding the interface only in later versions

fballiano:
@Flyingmana @kiatng the class was already fixed in https://github.com/OpenMage/magento-lts/pull/1432 but the problem kinda still exists if somebody extended the interface, would somebody do that to that component? I mean, if the getproduct method is necessary than we could add in https://github.com/OpenMage/magento-lts/pull/1432 without modifying the interface and case closed, right?

kiatng:
I agree with @fballiano. So, definitely not in branch 19. May be branch 20?

luigifab:
I vote for close them.

PR #1434, opened 2021-01-29
Better error handling for issues during bulk attribute update

woutersamaey:
If something goes wrong during mass product attribute update, the admin gets a generic message and there is no error in the log, making it harder to debug this issue.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
@@ -188,9 +188,11 @@ class Mage_Adminhtml_Catalog_Product_Action_AttributeController extends Mage_Adm
             $this->_getSession()->addError($attributeName . ': ' . $e->getMessage());
         }
         catch (Mage_Core_Exception $e) {
+            Mage::logException($e);
             $this->_getSession()->addError($e->getMessage());
         }
-        catch (Exception $e) {
+        catch (Throwable $e) {
+            Mage::logException($e);
             $this->_getSession()->addException($e, $this->__('An error occurred while updating the product(s) attributes.'));
         }
 
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
@@ -192,7 +192,7 @@ class Mage_Adminhtml_Catalog_Product_Action_AttributeController extends Mage_Adm
         }
         catch (Throwable $e) {
             Mage::logException($e);
-            $this->_getSession()->addException($e, $this->__('An error occurred while updating the product(s) attributes.'));
+            $this->_getSession()->addError($this->__('An error occurred while updating the product(s) attributes.'));
         }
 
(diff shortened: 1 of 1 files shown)


kiatng (dismissed):

jouriy (commented):
I just would suggest to keep Exception handling for backward compatibility, i.e.
```
        catch (Throwable $e) {
            Mage::logException($e);
            $this->_getSession()->addException($e, $this->__('An error occurred while updating the product(s) attributes.'));
        } catch (Exception $e) {
            $this->_getSession()->addException($e, $this->__('An error occurred while updating the product(s) attributes.'));
```

woutersamaey:
@jouriy your suggestion does not make much sense. Magento 1 was written at a time when Exception was the most generic thing to catch. Since PHP7 this has become Throwable, so I’m pretty sure this is what was originally meant. 
Things like TypeError did not exist at the time, but can now be a regular issue if you work in strict mode. This is why (IMHO) all mentions of Exception should be replaced with Throwable.

jouriy:
Thank you for the details. The reason why I suggested it was my impression that 1.9.4.x branch is supposed to be backward compatible with Magento 1.9.x (and drop-in replacement) which is still mostly on PHP 5.6, so removing Exception handling means no any error reporting for 5.x. 
> This is why (IMHO) all mentions of Exception should be replaced with Throwable.
Should not we create a new PR than to bring better error handing for all cases?

app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
```diff
@@ -188,9 +188,11 @@ public function saveAction()
             $this->_getSession()->addError($attributeName . ': ' . $e->getMessage());
         }
         catch (Mage_Core_Exception $e) {
+            Mage::logException($e);
```
colinmollenhour, line 191: I believe the convention is that the error is either a user-error (Mage_Core_Exception) and therefore displayed to the user, or an internal/unexpected error and therefore replaced with a generic error message and logged. So both logging and displaying the same error is not correct. I approve the `logException` in the next block but do not believe this one is necessary.

colinmollenhour (changes requested):

app/code/core/Mage/Adminhtml/controllers/Catalog/Product/Action/AttributeController.php
woutersamaey: I took your advice and removed this line.

woutersamaey:
I updated the code as per @colinmollenhour suggestion.

woutersamaey:
@jouriy running on PHP5.6 is a security risk, so I'm trying my best to modernize up to PHP7 without going overboard.
I'm contributing small things, as they affect my work.

colinmollenhour:
Ahh I just remembered that `addException` already logs the exception so this will double-log it.

woutersamaey:
@colinmollenhour I don’t follow. What are you trying to say?

woutersamaey:
I also created #1443 where I change the superfluous handing of the Exception in Mage_Core_Model_Session_Abstract::addException()

Magento seams to have forgotten about it's own Mage::logException(). Why they would ever format their own string and pass it to Mage::log() is beyond me.

woutersamaey:
Should I also move this to branch 20.0 ?

fballiano (approved):

kiatng (approved):

PR #1440, opened 2021-02-02
Allow changing of posted attribute values by event on mass attribute update

woutersamaey:
Passing vars by reference in an event does not work (anymore) in PHP7.
To provide a minimalistic solution, I wrapped the array in an ArrayObject, so not to break anything. Since this is an object, the contents can be changed by listeners.
The other vars could also be changed this way, but it's more invasive and wanted to start with just this.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Action.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Action.php
@@ -63,16 +63,18 @@ class Mage_Catalog_Model_Product_Action extends Mage_Core_Model_Abstract
      */
     public function updateAttributes($productIds, $attrData, $storeId)
     {
+        $attrData = new ArrayObject($attrData);
+
         Mage::dispatchEvent('catalog_product_attribute_update_before', array(
-            'attributes_data' => &$attrData,
+            'attributes_data' => $attrData,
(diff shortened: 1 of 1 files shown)


woutersamaey:
@colinmollenhour this is a tricky one, but works great. What do you think?

woutersamaey:
I needed this, because a client of mine has an image gallery in his "update attribute" page. This posts multiple values that come in as arrays, breaking the update attribute functionality which expects strings. By hooking into this event, I can do the needed steps separate and clean up the dirty data that was posted.

Seeing that Magento passed the vars by reference, it was clearly their intention to allow for this, but passing by reference is old skool, so we need to make changes.

Flyingmana (approved):

fballiano:
@sreichel what do you think about this one? I think this explanation is interesting regard what was being discussed in the past few days.

anyway I'll set it to draft since it has conflicts and it's not modifiable by anybody.

sreichel:
@fballiano i've tested the other PR .... passing by ref still works. I'll join in the late evening.

sreichel:
I wonder how this could be approved.

> I wrapped the array in an ArrayObject, so not to break anything

It could break observers that rely on `$attrData` being an array.

PR #1442, opened 2021-02-02
Replaced all catch Exception with Throwable

woutersamaey:
In PHP 5 the most generic thing to catch was Exception and Magento does this a lot. However with PHP7 and things like strict types, the most generic error to catch is now Throwable.

So, in this PR I've replaced 99% of the catch Exception with catch Throwable.
The only cases I didn't change are:
- The catch in cron.php (See my other PR #1320 for this)
- Any catches in Zend Framework

I hope this PR is appreciated, since it would greatly help the adoption of PHP7 and strict types, like I'm doing every day.


woutersamaey:
I've updated this PR by replacing the arguments of type Exception to Throwable.
And changes all Exception types to RuntimeException since this is common practise now. RuntimeException extends Exception with no changes, so this is a harmless change.

luigifab (dismissed):

woutersamaey:
Can we get this approved @Flyingmana @sreichel  @colinmollenhour ?

kiatng:
@woutersamaey Thanks for the effort. But should we follow the convention of naming the var as `$t`: ` } catch (Throwable $t) {`, instead of `$e`?

woutersamaey:
@kiatng What's the benefit? Some people like `$e` others use `$ex` now there is `$t`? Never seen it before, but I'm mostly in M1 still.
If this is truely desired, it would be an easy find & replace all...

kiatng:
The var `$t` was probably first used by the author of the [RFC on throwable](https://wiki.php.net/rfc/throwable-interface) in his [blog](https://trowski.com/2015/06/24/throwable-exceptions-and-errors-in-php7/):

```php
try {
    // Code that may throw an Exception or Error.
} catch (Throwable $t) {
    // Handle exception
}
```

I agree with @woutersamaey, we need to get this fixed ASAP, to be fully compliant with PHP 7, see comment in the said blog on _Writing Code to Support PHP 5.x and 7 Exceptions_:

```php
try {
    // Code that may throw an Exception or Error.
} catch (Throwable $t) {
    // Executed only in PHP 7, will not match in PHP 5.x
} catch (Exception $e) {
    // Executed only in PHP 5.x, will not be reached in PHP 7
}
```
I was wondering why some errors were not caught after upgrading to PHP7, now it made sense.

kiatng (approved):

Rahza:
+1 on this, missed several errors because of this

Flyingmana:
in any case, I would like to move the PR to the 20.0 branch then, if its ok for you

woutersamaey:
Good idea

kiatng:
Question: what is the difference between 

```php
throw new Exception($msg);
```

and 

```php
throw new RuntimeException($msg);
```

and why do you choose the latter?

woutersamaey:
It's a recommended practise. Some more details can be read here: https://webdevetc.com/blog/why-you-should-use-spl-exceptions-in-php-for-better-exception-handling/#section_runtimeexception

colinmollenhour:
Thanks for the proposal, @woutersamaey!

Using RuntimeException seems like a mostly subjective change, what will it improve in the case of OpenMage?
I'm afraid it will create an unnecessary burden when reviewing pull requests, backporting or forward-porting patches, etc. Also I would argue that in a way Mage_Core_Exception is basically equivalent to RuntimeException so while it may have been standardized since, it is not really a new concept and is instead a redundant one.

This probably belongs in an RFC.. Please see the RFC [template](https://github.com/OpenMage/rfcs/blob/main/draft/0000-template.md), I think this is the kind of discussion this change needs before it is accepted. I'm not so sure that a practice being recommended is a sufficient reason to make such sweeping changes. "They" also made PHP backwards compatible with the previous recommended practices so that existing projects would not be forced to adopt the new ones. :)

kkrieger85 (dismissed):

Flyingmana (changes requested):
semantically seen, it seems wrong to catch and log exceptions for non-exceptions.

luigifab:
Some news?

ADDISON74:
Before getting this merged we should discussed the PR #1434 too. At least we have the same way of handling exceptions adapted to the new variants of PHP.

luigifab (approved):
In production for > 12 months.
But I think there are new Exception added.

kiatng (dismissed):

fballiano:
> Using RuntimeException seems like a mostly subjective change, what will it improve in the case of OpenMage? I'm afraid it will create an unnecessary burden when reviewing pull requests, backporting or forward-porting patches, etc. Also I would argue that in a way Mage_Core_Exception is basically equivalent to RuntimeException so while it may have been standardized since, it is not really a new concept and is instead a redundant one.

I agree with @colinmollenhour on this one. To have these many differences between v19 and v20 will make it really hard in the future. But since v19 will have to be bumped to something higher than php7.0 sometimes soon, would't it be ok to have this PR on v19 too?

ADDISON74:
Let's rebase the PR to 19. Sury dropped this month the support for Debian 9 and you cannot install other PHP versions anymore. Debian 10 comes with 7.3 and Debian 11 with 7.4. We should not care about versions bellow 7.3.

kiatng:
I vote for v19. I just spent many minutes debugging an error because OM does not throw anything due to this:

https://github.com/OpenMage/magento-lts/blob/5de30bfc6239d9a4072b743b2dac29c079763176/app/code/core/Mage/Core/Block/Template.php#L263

The error is thrown once I changed it to `Throwable`.

The code in the block:
```php
    /**
     * @return bool
     */
    public function getCanWestEastTransfer()
    {
        return (bool) $this->getParentBlock()
            ->getCustomer()
            ->getCanWestEastTransfer();
    }
```

The page loaded without any error, but it is broken. After change to `Throwable`, error is thrown, which should be expected:

> Fatal error:  Uncaught Error: Call to a member function getCanWestEastTransfer() on null

colinmollenhour:
My take currently:

- I agree with changing `Exception` to `Throwable` where it is meant to catch the most generic error, but I don't think that is every location. For example there are places in controllers where the exception is caught, a somewhat generic error message added to the session and the exception is neither logged nor rethrown. In these cases the Throwable becomes harder to debug, not easier. Perhaps `Throwable` should only be caught in the platform code, not in models, blocks, helpers and controllers? This would let these truly unexpected errors bubble up to the platform code where it can always be logged.
- Throwing exceptions with messages that are translated strings should be `Mage_Core_Exception` rather than `RuntimeException`, in my opinion.
- I'm hesitant to approve this for v19 but will defer to others; as stated in another issue somewhere I would prefer the v19 discussion be a separate PR altogether after the v20 issue is merged to keep the considerations separate and avoid slowing what everyone agrees is good for v20 with the v19 discussion.

ADDISON74:
I confirm what @kiatng said. I recently faced the implode function in an old PHTML file and the error did not appear anywhere. When I changed Exception to Throwable the error appeared in the browser window. Although I solved the issue without modification, I lost time to find it.

I have a question, is it possible to display these error messages in the exceptions.log file instead of the browser window? If the error occurs in production, at least it is not displayed to visitors.

ADDISON74:
@kiatng - let's discuss it here #2606.

luigifab:
Any news?

fballiano:
I rebased on 1.9.4.x

ADDISON74:

luigifab:
New PR or not?

colinmollenhour:
> New PR or not?

I think a new PR is needed as a blanket replacement of all Exception with Throwable is probably not optimal and may not be an improvement. I think a new PR would need to spell out exactly what problems it is solving and why the specific approach was taken. That is, coder errors (e.g. type errors, calls to undefined functions) should not necessarily be handled the same as all other errors. Since there is already a "default error handler" registered, what needs to be fixed exactly?

PR #1448, opened 2021-02-04
Remove _isDownloader flag

luigifab:
This PR remove an useless flag about the downloader.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -104,13 +104,6 @@ final class Mage
     static private $_objects;
 
     /**
-     * Is downloader flag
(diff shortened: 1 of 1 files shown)


Flyingmana:
can we be sure, nobody is actually using this?
I would say we should mark it with a @ deprecated annotation only for now.

luigifab:
I kept the method and added deprecated.

Flyingmana (approved):

fballiano (approved):

kkrieger85 (approved):

PR #1451, opened 2021-02-05
Remove useless item options save

bastienlm:
Hello,

Because the `_afterSave()` function already calls `_saveItemOptions()` everytime a quote item is saved, the override of `save()` function to add `_saveItemOptions()` in case of data changes is useless.

What do you think about this modification?


colinmollenhour:
I'm not so sure, it looks like it could actually have an effect because in the case that `hasDataChanges` is false and `$this->_getResource()->hasDataChanged($this)` is also false then I think this would force `_saveItemOptions` to be called when otherwise it would not because  `_afterSave` is not called.

bastienlm:
I think you right, so i's maybe better to not remove this.

Thanks for your feedback

PR #1456, opened 2021-02-09
Correct password length message grammar

rjocoleman:
### Description (*)

This message has bad grammar, it's only in Adminhtml. Not a big issue but it's an easy fix.

1. Admin -> Create new customer, admin user, api user etc.

### Questions or comments

~~This will break existing translations - if this is a concern I can revert and change just the `en_US` translation replacement if that's preferred?~~ This PR updates the `en_US` translation only.

### DIFF
--- a/app/locale/en_US/Mage_Adminhtml.csv
+++ b/app/locale/en_US/Mage_Adminhtml.csv
@@ -716,7 +716,7 @@
 "Password","Password"
 "Password Confirmation","Password Confirmation"
 "Password confirmation must be same as password.","Password confirmation must be same as password."
-"Password must be at least of %d characters.","Password must be at least of %d characters."
+"Password must be at least of %d characters.","Password must be at least %d characters."
 "Password must include both numeric and alphabetic characters.","Password must include both numeric and alphabetic characters."
 "Password:","Password:"
 "Path:","Path:"


sreichel (changes requested):
This would break all language packs.  I'd just change the translation csv.

rjocoleman:
Thanks @sreichel this has been updated

sreichel (approved):

jouriy:
It will break widely used Ess/M2ePro and Aheadworks extensions (AW/Core, AW/Blog etc.), so likely hit 9/10 of stores..

kkrieger85 (approved):

ADDISON74:
As I see this was merged but I suggest changing the PHP code where it is used this phrase

```"Password must be at least of %d characters."```

to this

```Password must be at least %d characters.```

**Comment by rjocoleman** [comment]
@ADDISON74 that was the original change in this PR and "the correct fix" but it would break all of the translations/language packs that already exist and rely on that string.

So it's a backward incompatible change that doesn't fit the requirements of (at least) the 1.9.4.x release stream.

**Comment by ADDISON74** [comment]
Checking Magento code for the phrase "Password must be at least of %d characters." I found only two files:

```
/app/locale/en_US/Mage_Adminhtml.csv:
"Password must be at least of %d characters.","Password must be at least of %d characters."
```

```
/app/code/core/Mage/Admin/Model/User.php:
$errors[] = Mage::helper('adminhtml')->__('Password must be at least of %d characters.', self::MIN_PASSWORD_LENGTH);
```

I understood the change is for a better English and not breaking up other translations but it should be operated in both files to keep consistency. I never saw in a translation different values like this one: 

```
"Password must be at least of %d characters.",""Password must be at least %d characters."
```

Those who are using translations for Backend should update their translations file. Just switching from Magento vanilla I had to add a lot of translations in different languages.

sreichel:
> Those who are using translations for Backend should update their translations file.

You had to update your german translation for a change that did not even affect you. Does that make sense?

ADDISON74:
@sreichel: OpenMage ships with one translation US English. If we do a change in the code it should be in all places. I understood your point of view for compatibility with other languages already in use but I do not agree it.

sreichel:
Then, there is no need for en_US translations at all?

Flyingmana:
@ADDISON74 you have to see the texts inside the template just as "Keys", not actual texts. To use actual understandable texts is just a decision with certain up- and down- sides. One of them is to easier understand templates.
A grammar error in the template does not hurt for this. If it would be a logical error, that would be different story.

ADDISON74:
Thank you for your replies. I will respect my opinion related to this commit. You know better what to do.

PR #1457, opened 2021-02-09
Fix: Customer with empty DOB can't be saved

rjocoleman:
### Description (*)

If Date of Birth is enabled when admin creates a customer that has an empty DOB - when that record is saved the save fails without visible error. 

This appears to be because `Mage_Eav_Model_Form` validates and sets default data, including `''` for dates. 
 `Zend_Format`'s date validation doesn't like the default string `''`

This has been encountered in similar code in Magento2: 
https://github.com/magento/magento2/pull/14666
https://github.com/magento/magento2/issues/12146
https://github.com/magento/magento2/issues/13311

To be clear this issue is dates in the `Mage_Eav_Model_Form` and those which extend it - Customer DOB is an easily visible example.

This PR ports the solution from Magento 2 which basically fails fast and bypasses the lengthly validation, parsing and filtering if there is no value set.

1. Log in to admin
2. Enable Date Of Birth under Customer in System Configuration. It should not be a required value.
3. Create New Customer:
  * Change store to any other than admin (this was required for me to reproduce the issue)
  * Set no DOB
  * Add all other required fields
4. Save will fail without an error (server ajax response shows error with no message)

### Questions or comments

I am unsure if this is related to some changes around PHP 7.1+
The Magento 2 issues reference some "A non-numeric value" errors but I haven't seen them.
I am unsure if my locale not being `en_US` (it's `en_NZ`) is significant

I added this to Datetime as well, and output filters. 

### DIFF
--- a/lib/Varien/Data/Form/Filter/Date.php
+++ b/lib/Varien/Data/Form/Filter/Date.php
@@ -71,6 +71,10 @@ class Varien_Data_Form_Filter_Date implements Varien_Data_Form_Filter_Interface
      */
     public function inputFilter($value)
     {
+        if (!$value) {
+            return $value;
+        }
+
         $filterInput = new Zend_Filter_LocalizedToNormalized(array(
             'date_format'   => $this->_dateFormat,
             'locale'        => $this->_locale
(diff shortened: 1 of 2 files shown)


kiatng:
I cannot replicate the error with OM 20.0.6. I set DOB in

backend > System > Configuration > Customers / Customer Configuration > Name and Address Options > set Show Date of Birth to _Optional_

Then I create a customer in backend, associated the customer to _Main Website_, leave DOB empty and saved without issue. What did I miss?

rjocoleman:
hmm, thanks @kiatng it appears to be something in my locale. I can't see exactly what to reproduce it outside of the older DB I found it in. I'll dig deeper.

Marking this draft until I have clear and complete steps to reproduce.

PR #1466, opened 2021-02-18
Improve error message of product required option by including option …

kiatng:
…title.

### Description (*)

If a product has lots of custom options, and when server-side validation fail on a required option, it is useful to know which option failed the validation, like this:

image

The above error message was taken from frontend cart page after the change.

1. ...
2. ...


ADDISON74:
By default when there are no options selected in product page when you press the [Add to Cart] button Magento will highlight all options and displays a red message bellow. If there are options selected already only those not selected will be highlighted. This is coming from client side based on JS validation.

A few issues related to this commit:

- What happens when there are more than one option unselected? Will the error message list all unselected options?
- An new option is configuration is a must having the option to choose between the default behavior or this one.

kiatng:
> A few issues related to this commit:
> 
> * What happens when there are more than one option unselected? Will the error message list all unselected options?
> * An new option is configuration is a must having the option to choose between the default behavior or this one.

My intention is not to change the existing behavior on server-side validation, only provide better error message for users. There are 2 use cases:
1. When setting or adding new required options to an existing product. If the carts contain the product without the new options, the errors will be thrown and shown on frontend cart page. This is useful in my case where a product may have > 50 options.
2. For developers dealing with frontend product and cart pages, while the logic of the products are still being developed and tested. The error message with option title is very useful for me.

kkrieger85 (approved):

Sekiphp (approved):

Flyingmana (approved):

PR #1467, opened 2021-02-18
_exportIterateCollection performance optimization

midlan:
### Description (*)

In my opinion, there is no reason to export like:
1. hey database, tell me the count of the data!
2. if there is count > 0, give me the actual data

Instead of it, we can skip the step 1. and load the data immediately.

1. use export feature of any grid widget pointing to `getCsvFile` method, e.g. tax rates grid

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -963,7 +963,6 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
      *
      * @param string $callback
      * @param array $args additional arguments for callback method
-     * @return $this
      */
     public function _exportIterateCollection($callback, array $args)
     {
@@ -971,17 +970,13 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
         $count = null;
         $page  = 1;
 
-        while (true) {
+        do {
             $collection = clone $originalCollection;
             $collection->setPageSize($this->_exportPageSize);
             $collection->setCurPage($page);
             $collection->load();
 
             $count = $collection->count();
-
-            if ($count < $this->_exportPageSize) {
-                break;
-            }
 
             $page++;
 
@@ -990,9 +985,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
             }
             $collection->clear();
             unset($collection);
-        }
-
-        return $this;
+        } while($count == $this->_exportPageSize);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
```diff
@@ -970,29 +970,29 @@ public function _exportIterateCollection($callback, array $args)
...
+            $page++;
 
             foreach ($collection as $item) {
                 call_user_func_array(array($this, $callback), array_merge(array($item), $args));
             }
             $collection->clear();
             unset($collection);
         }
+
+        return $this;
```
midlan, line 995: I added this return to match the phpdoc

joshua-bn: I'd go the other way with it. Make the PhpDoc match what's actually happening.

There's no good reason why anyone should use this as a fluid interface. Too much `return $this` in Magento

app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
```diff
@@ -970,29 +970,29 @@ public function _exportIterateCollection($callback, array $args)
...
-            if (is_null($count)) {
-                $count = $collection->getSize();
-                $lPage = $collection->getLastPageNumber();
-            }
-            if ($lPage == $page) {
-                $break = true;
+
+            $count = $collection->count();
+
+            if ($count < $this->_exportPageSize) {
```
kiatng, line 982: I didn't test this. But assuming `$originalCollection` has 1001 items, given that `protected $_exportPageSize = 1000;`, it will require 2 iterations. On the second iteration, `$count` will be 1, but it won't get exported since it'll `break` here. So, we need `$break = true;` instead of `break;`. Or am I missing something.

kiatng: If I am right, `$count` can be init to `$this->_exportPageSize`:

```php
$originalCollection = $this->getCollection();
$count = $this->_exportPageSize;
//...
while ($count != $this->_exportPageSize) {
    //...
    $count = $collection->getSize(); // I prefer the standard method.
    // we can skip `if ($count < $this->_exportPageSize) {` and continue to export
}
```

midlan: Your thoughts seems right. My mistake. I swear I tested it, we are using the patched version in our Magento instance. I will test it again.

midlan: Ok I figured it out. There is two methods `getCsv` and `getCsvFile`. Only `getCsvFile` is using `_exportIterateCollection`. But in most grids, the `getCsv` is used, so when I was testing my code, I actually run the other method. The right place to test it is tax rates grid:

https://yourwebsite/admin/tax_rate/

midlan: It is fixed now and tested.

app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
midlan: Turned your note into code.

colinmollenhour (approved):
This is a good improvement, count is very slow on some collections.

kiatng (approved):
LGTM.

Sekiphp (approved):

LukasCesal (approved):

tmotyl:
regression fix: https://github.com/OpenMage/magento-lts/pull/1888

PR #1471, opened 2021-02-22
xmlrpc Calling parameters do not match signature

empiricompany:
Fix wrong datatype in xmlrpc webservice
return this error: "Calling parameters do not match signature" cause can't check datatypes properly

### Description (*)

This is an old bug in Zend XmlRpc library reported also from many third party players, like [zapier](https://zapier.com/help/doc/common-problems-magento), that suggest to comment out exception

my solution is to explicit gettype in condition to properly check
i'm run in php 7.2 env, i can't understand why '===" condition not properly work, but we have solved this way 

in some scenarios calling xmlrpc api return error "Calling parameters do not match signature"

### DIFF
--- a/lib/Zend/XmlRpc/Server.php
+++ b/lib/Zend/XmlRpc/Server.php
@@ -586,7 +586,7 @@ class Zend_XmlRpc_Server extends Zend_Server_Abstract
         $signatures = $info->getPrototypes();
         foreach ($signatures as $signature) {
             $sigParams = $signature->getParameters();
-            if ($sigCalled === $sigParams) {
+            if (gettype($sigCalled) === gettype($sigParams)) {
                 $matched = true;
                 break;
             }


colinmollenhour:
What is the value of `$sigParams` in this context?

kiatng:
I'm not familiar with `Zend_XmlRpc_Server`. Is this related to missing docblock reported [here](http://zend-framework-community.634137.n4.nabble.com/XmlRpc-quot-Calling-parameters-do-not-match-signature-quot-problem-td644401.html)? 

> Yes. You need to have docblocks in place:

    /**
     * Echo back a string
     *
     * @param string $textToEcho
     * @return string
     */
    function echoTest($textToEcho)
    {
        return "You said: " . $textToEcho;
    }

> The docblocks are used to provide type hinting so that the XmlRpc server
knows what variable types are allowed for a given parameter -- this is
both as a security measure as well as to ensure that method signatures
are properly generated.

The docblock convention can be found [here](https://web.archive.org/web/20070302190453/http://framework.zend.com/manual/en/zend.xmlrpc.server.html#zend.xmlrpc.server.conventions):

> Zend_XmlRpc_Server allows the developer to attach functions and class method calls as dispatchable XML-RPC methods. Via Zend_Server_Reflection, it does introspection on all attached methods, using the function and method docblocks to determine the method help text and method signatures.
>
> XML-RPC types do not necessarily map one-to-one to PHP types. However, the code will do its best to guess the appropriate type based on the values listed in @param and @return lines. Some XML-RPC types have no immediate PHP equivalent, however, and should be hinted using the XML-RPC type in the phpdoc.

empiricompany:
i went deeper, i tested in the official docker openmage environment and I have not encountered any problems

i can't say why in my installation and environment php can't properly compare types with '===' condition ($sigCalled and $sigParams in both environment is the same), maybe something wrong in my magento or some difference with php 7.2 and php 7.3, unfortunately i have not so much time to find the issue.

@colinmollenhour sorry for unforgivable mistake, comparing $sigCalled and $sigParams with gettype() is wrong because is always true, is always an array (of types).

@kiatng yes the correct implementation must declare in docblock @params and @return types
but i can say that work also in some third party implementations without a docblock 

the signature check of the requested and expected types work as follow:
for example api xmlrpc order.info expect only 3 types:
- string (token)
- string (order.info)
- array (order id)

if i add a 4° parameter with empty array cause the error of signature as excpected:
`<param><value><array><data></data></array></value></param>`
or if i pass order id as a string type like
`<param><value><string>145000004</string></value></param>`


```
curl -X POST -i 'http://www.openmage.test/api/xmlrpc' --data '<methodCall>
    <methodName>call</methodName>
    <params>
        <param>
        <value>
            <string>e20841d99d2e08ca868c7a73443ca154</string>
        </value>
        </param>
        <param>
        <value>
            <string>order.info</string>
        </value>
        </param>
        <param>
        <value>
            <array>
                <data>
                    <value>145000004</value>
                </data>
            </array>
        </value>
        </param>
        <param><value><array><data></data></array></value></param>
    </params>
</methodCall>'
```

```
/**
     * Handle an xmlrpc call (actual work)
     *
     * @param Zend_XmlRpc_Request $request
     * @return Zend_XmlRpc_Response
     * @throws Zend_XmlRpcServer_Exception|Exception
     * Zend_XmlRpcServer_Exceptions are thrown for internal errors; otherwise,
     * any other exception may be thrown by the callback
     */
    protected function _handle(Zend_XmlRpc_Request $request)
    {
        ....
        $signatures = $info->getPrototypes();
        foreach ($signatures as $signature) {
            Mage::log(__METHOD__);
            Mage::log('$signature');
            Mage::log($signature);
            $sigParams = $signature->getParameters();
            Mage::log('$sigCalled');
            Mage::log($sigCalled);
            
            Mage::log('$sigParams');
            Mage::log($sigParams);
            if ($sigCalled === $sigParams) {
            //if (gettype($sigCalled) === gettype($sigParams)) {
                $matched = true;
                break;
            }
        }
        ....
```


```
2021-02-23T07:48:20+00:00 DEBUG (7): Zend_XmlRpc_Server::_handle
2021-02-23T07:48:20+00:00 DEBUG (7): $signature
2021-02-23T07:48:20+00:00 DEBUG (7): Zend_Server_Method_Prototype Object
(
    [_returnType:protected] => struct
    [_parameterNameMap:protected] => Array
        (
            [sessionId] => 0
            [apiPath] => 1
            [args] => 2
        )

    [_parameters:protected] => Array
        (
            [0] => Zend_Server_Method_Parameter Object
                (
                    [_defaultValue:protected] => 
                    [_description:protected] => 
                    [_name:protected] => sessionId
                    [_optional:protected] => 
                    [_type:protected] => string
                )

            [1] => Zend_Server_Method_Parameter Object
                (
                    [_defaultValue:protected] => 
                    [_description:protected] => 
                    [_name:protected] => apiPath
                    [_optional:protected] => 
                    [_type:protected] => string
                )

            [2] => Zend_Server_Method_Parameter Object
                (
                    [_defaultValue:protected] => Array
                        (
                        )

                    [_description:protected] => 
                    [_name:protected] => args
                    [_optional:protected] => 1
                    [_type:protected] => array
                )

        )

)

2021-02-23T07:48:20+00:00 DEBUG (7): $sigCalled
2021-02-23T07:48:20+00:00 DEBUG (7): Array
(
    [0] => string
    [1] => string
    [2] => array
    [3] => array <=== 4° parameter not match with signature
)

2021-02-23T07:48:20+00:00 DEBUG (7): $sigParams
2021-02-23T07:48:20+00:00 DEBUG (7): Array
(
    [0] => string
    [1] => string
    [2] => array
)
```

so we can close this PR..

tmotyl:
closing as requested by author

PR #1476, opened 2021-02-26
Fix branch alias to make the branch importable on packagist.org

Seldaek:
The invalid branch alias causes the update to fail on packagist.org, and the branch can not be found as a version there https://packagist.org/packages/openmage/magento-lts


Flyingmana:
duplicate of https://github.com/OpenMage/magento-lts/pull/1464

Flyingmana:
but merged the other one now :)

PR #423, opened 2018-01-09
Added some event-prefixes, ref #249

sreichel:
- fixes wrong EAV prefixes from #249
- removed block prefixes (wasnt complete)

PR #1485, opened 2021-03-04
New event "init_form_values_after" after data set on a form

woutersamaey:
Magento does not have a common event after data is set on a form.
This new event allows us to change the form, the elements and the values once data is entered, giving full flexibility.
I needed this event in a special case where I needed to change the options of a select, to include the current value that was not in the existing options list.
There was no other way to do this without overriding many forms 1 by 1.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Form.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Form.php
@@ -143,6 +143,7 @@ class Mage_Adminhtml_Block_Widget_Form extends Mage_Adminhtml_Block_Widget
     {
         $this->_prepareForm();
         $this->_initFormValues();
+        Mage::dispatchEvent('init_form_values_after', ['form' => $this]);
         return parent::_beforeToHtml();
     }
 
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Form.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Form.php
@@ -143,7 +143,13 @@ class Mage_Adminhtml_Block_Widget_Form extends Mage_Adminhtml_Block_Widget
     {
         $this->_prepareForm();
         $this->_initFormValues();
-        Mage::dispatchEvent('init_form_values_after', ['form' => $this]);
+        Mage::dispatchEvent(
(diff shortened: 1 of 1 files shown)


colinmollenhour (dismissed):
Excellent addition!

If I was designing from scratch I'd probably add a second event with a default overridable prefix to it like `$this->_eventPrefix.'_form_before_html'` so that the eventPrefix can be added to classes making it easier to override the specific form using the event name (same as models). But with or without this should have been added years ago. :)

sreichel:
> 
> 
> Excellent addition!
> 
> If I was designing from scratch I'd probably add a second event with a default overridable prefix to it like `$this->_eventPrefix.'_form_before_html'` so that the eventPrefix can be added to classes making it easier to override the specific form using the event name (same as models). But with or without this should have been added years ago. :)

OT: Is it worth tackling #423 again? You had some conserns here https://github.com/OpenMage/magento-lts/pull/423#issuecomment-356660393 ...  but i'm still not sure how to fix it.

app/code/core/Mage/Adminhtml/Block/Widget/Form.php
```diff
@@ -143,6 +143,7 @@ protected function _beforeToHtml()
     {
         $this->_prepareForm();
         $this->_initFormValues();
+        Mage::dispatchEvent('init_form_values_after', ['form' => $this]);
```
kiatng, line 146: I'd like to change 2 things:
1. Event name should follow naming convention
2. The param `form` should be `$this->getForm()`, and `$this` is assigned to param `block`.

```suggestion
Mage::dispatchEvent(
    'adminhtml_block_widget_form_init_form_values_after',
    [
        'block' => $this,
        'form' => $this->getForm()
    ]
);
```

kiatng (changes requested):

woutersamaey:
Thanks for the great suggestions @kiatng

kiatng (approved):

woutersamaey:
@colinmollenhour and @sreichel can we get this merged?

colinmollenhour (approved):

sreichel (approved):

PR #1358, opened 2020-12-18
Changed event name `adminhtml_block_widget_tabs_html_before`, add 'block' for name consistency.

kiatng:
### Description (*)
Ref PR #879 on adding tab in pages in adminhtml where the event name was `adminhtml_widget_tabs_html_before`. But it hsould be `adminhtml_block_widget_tabs_html_before` for name consistency. The other event that can be used to add tab is `core_block_abstract_to_html_before`, but it's fired for every block in frontend and backend.

### Related Pull Requests

 PR #879.

### Fixed Issues (if relevant)

1. Fixes OpenMage/magento-lts#<issue_number>

### Manual testing scenarios (*)

1. ...
2. ...

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)

PR #1488, opened 2021-03-08
New event after tabs have been added

woutersamaey:
### Description (*)
A generic event after tabs have been added, allowing us to add more tabs, edit or remove them without complicating the active tab or sorting of tabs.

Overriding the block is not an option, because the active tab + sorting is handled in the parent class. There is no pretty way to add a tab without ugly copy pasting of code Magento code.

The case where I ran into this was when I wanted to add a tab to `Mage_Adminhtml_Block_Catalog_Product_Attribute_Edit_Tabs`


woutersamaey:
Similar to #1488 I also figured out we could use an event after tabs have been added, but before the active tab and sorting is handled. This event provides this convenience for the entire backend.

kiatng:
Duplicate with PR #1358.

PR #1489, opened 2021-03-08
Old inactive quotes are now actually purged from the database

FredericMartinez:
### Description (*)

Since the beggining of Magento, only quotes converted to orders can be deleted by `sales_clean_quotes` cron because quote collection filter on `is_active=0` ​

---
Show quotes group by year
```
select YEAR(updated_at), count('*')
from sales_flat_quote
GROUP BY YEAR(updated_at)
```

| YEAR\(updated\_at\) | count\('\*'\) |
| :--- | :--- |
| 2018 | 1014080 |
| 2019 | 404537 |
| 2020 | 159526 |
| 2021 | 46765 |

### DIFF
--- a/app/code/core/Mage/Sales/Model/Observer.php
+++ b/app/code/core/Mage/Sales/Model/Observer.php
@@ -38,17 +38,19 @@ class Mage_Sales_Model_Observer
     public function cleanExpiredQuotes($schedule)
     {
         Mage::dispatchEvent('clear_expired_quotes_before', ['sales_observer' => $this]);
-
         $lifetimes = Mage::getConfig()->getStoresConfigByPath('checkout/cart/delete_quote_after');
-        foreach ($lifetimes as $storeId => $lifetime) {
-            $lifetime = (int)$lifetime * 86400;
+
+        foreach ($lifetimes as $storeId => $day) {
+            $day = (int) $day;
+            $lifetime = 86400 * $day;
 
             /** @var Mage_Sales_Model_Resource_Quote_Collection $quotes */
-            $quotes = Mage::getModel('sales/quote')->getCollection();
-
+            $quotes = Mage::getResourceModel('sales/quote_collection');
             $quotes->addFieldToFilter('store_id', $storeId);
-            $quotes->addFieldToFilter('updated_at', ['to' => date("Y-m-d", time() - $lifetime)]);
-            $quotes->addFieldToFilter('is_active', 0);
+            $quotes->addFieldToFilter('updated_at', ['to' => date('Y-m-d', time() - $lifetime)]);
+            if ($day == 0) {
+                $quotes->addFieldToFilter('is_active', 0);
+            }
 
             foreach ($this->getExpireQuotesAdditionalFilterFields() as $field => $condition) {
                 $quotes->addFieldToFilter($field, $condition);
@@ -56,6 +58,7 @@ class Mage_Sales_Model_Observer
 
             $quotes->walk('delete');
         }
+
         return $this;
     }
 
--- a/app/code/core/Mage/Checkout/etc/system.xml
+++ b/app/code/core/Mage/Checkout/etc/system.xml
@@ -80,6 +80,7 @@
                     <fields>
                         <delete_quote_after translate="label">
                             <label>Quote Lifetime (days)</label>
+                            <validate>validate-not-negative-number</validate>
                             <sort_order>1</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>


app/code/core/Mage/Sales/Model/Observer.php
```diff
@@ -60,7 +60,7 @@ public function cleanExpiredQuotes($schedule)
 
             $quotes->addFieldToFilter('store_id', $storeId);
             $quotes->addFieldToFilter('updated_at', array('to'=>date("Y-m-d", time()-$lifetime)));
-            $quotes->addFieldToFilter('is_active', 0);
+            $quotes->addFieldToFilter('is_active', 1);
```
FredericMartinez, line 63: Note: We can also remove the line

app/code/core/Mage/Sales/Model/Observer.php
```diff
@@ -60,7 +60,7 @@ public function cleanExpiredQuotes($schedule)
...
 
             foreach ($this->getExpireQuotesAdditionalFilterFields() as $field => $condition) {
```
FredericMartinez, line 65: Note2: We can also remove this "useless" foreach

kiatng: Removing may break BC.

app/code/core/Mage/Sales/Model/Observer.php
kiatng: `is_active` fitlering should be removed, otherwise, non-active quotes would not be deleted.

rvelhote:
I was thinking about this issue last week.

Our `sales_flat_quote` table is at ~2.15 million records and there is a situation where the `trigger_recollect` flag might be [modified for all quotes](https://github.com/OpenMage/magento-lts/blob/18cb71ad5b543bb3b83314813415714580ae6a6f/app/code/core/Mage/Sales/Model/Resource/Quote.php#L185) due to a catalog rule or price modification (maybe more reasons?).

I believe that this was the cause for the dreaded `SQLSTATE[HY000]: General error: 1205 Lock wait timeout exceeded; try restarting transaction, query was: XYZ` we experienced last week. The table took 3 to 5 minutes fo fully update which stopped customers from adding products to the cart during that timeframe - it's possible/likely, however, that it was executed multiple times in succession rather than just once.

I'm still in the process of investigating this issue and mitigation and reducing the table would have a very positive effect.

fballiano (approved):

joshua-bn:
Yikes. If `checkout/cart/delete_quote_after` is empty for whatever reason, it will cause all active quotes to break. Please, whatever you do, put a default on that.

fballiano:
what are we doing on this @FredericMartinez?

fballiano (approved):

kiatng (dismissed):

luigifab:
> If `checkout/cart/delete_quote_after` is empty for whatever reason, it will cause all active quotes to break.
A check is needed no? For example, if empty, keep is_active filter?

kiatng:
> > If `checkout/cart/delete_quote_after` is empty for whatever reason, it will cause all active quotes to break.
> 
> A check is needed no? For example, if empty, keep is_active filter?

We cannot merge this until this is addressed: all active quotes will be deleted if `checkout/cart/delete_quote_after` is empty. 

A suggestion:

```php

    public function cleanExpiredQuotes($schedule)
    {
        Mage::dispatchEvent('clear_expired_quotes_before', array('sales_observer' => $this));
        $lifetimes = Mage::getConfig()->getStoresConfigByPath('checkout/cart/delete_quote_after');
        foreach ($lifetimes as $storeId => $day) {
            $day = (int) $day;
            $lifetime = 86400 * $day;
            /** @var Mage_Sales_Model_Mysql4_Quote_Collection $quotes */
            $quotes = Mage::getModel('sales/quote')->getCollection();

            $quotes->addFieldToFilter('store_id', $storeId);
            $quotes->addFieldToFilter('updated_at', array('to'=>date("Y-m-d", time()-$lifetime)));
            if ($day) {
                $quotes->addFieldToFilter('is_active', 0);
            }
            // ...
```

Also add a note in the backend, something like:

image

Please suggest better English.

luigifab:
With the comment, I also suggest to add in system.xml: `<validate>validate-not-negative-number</validate>`

fballiano (changes requested):
we've to take a look at what @kiatng is saying

luigifab:
I updated and merge with Github web editor.

app/code/core/Mage/Sales/Model/Observer.php
```diff
@@ -45,24 +45,29 @@ class Mage_Sales_Model_Observer
...
 
-            /** @var Mage_Sales_Model_Resource_Quote_Collection $quotes */
-            $quotes = Mage::getModel('sales/quote')->getCollection();
+        foreach ($lifetimes as $storeId => $day) {
 
+            $day = (int) $day;
+            $lifetime = 86400 * $day;
+
+            /** @var Mage_Sales_Model_Resource_Quote_Collection $quotes */
+            $quotes = Mage::getResourceModel('sales/quote_collection');
```
sreichel, line 56: Why inside loop?

luigifab (dismissed):

luigifab (dismissed):

app/code/core/Mage/Sales/Model/Observer.php
```diff
@@ -45,24 +45,28 @@ class Mage_Sales_Model_Observer
...
-            $lifetime *= 86400;
 
-            /** @var Mage_Sales_Model_Resource_Quote_Collection $quotes */
-            $quotes = Mage::getModel('sales/quote')->getCollection();
+        foreach ($lifetimes as $storeId => $day) {
+            $day = (int) $day;
+            $lifetime = 86400 * $day;
 
+            /** @var Mage_Sales_Model_Resource_Quote_Collection $quotes */
+            $quotes = Mage::getResourceModel('sales/quote_collection');
```
sreichel, line 55: Is it need to call it inside foreach loop?

elidrissidev: Yes, to respect the configuration of every store.

fballiano (approved):

fballiano:
I've rebased it and I think it's ready to be merged

luigifab (approved):

kyrena:
Ping @kiatng and @FredericMartinez for latest PR update 😸.

Issue #784, opened 2019-06-23
[Planned] Cleanup: Round 2 - clean code (phpstan update)


sreichel:
Next planned steps ...

## DOC updates

- [ ] add `@property` annotation for undefined properties
- [ ] complete/improve/fix doc updates from round 1
  - [ ] correct `object`/`mixed` annotations
  - [ ] return `Type[]` instead of `array` (should make some inline @var annotation obsolete)
  - [ ] add more  `getCollection()`, `getItems()` hints for better 3rd patry-code support
- [ ] add some DOCS to tempates *

## Code changes

- add typecasting and type checks to enforce correct parameter type where it it neccessary
- <s>replace `is_null($var)` with `$var === null` ... NOTE: seems useless, but from code analyte view it has a big impact ... same for return `false` instead of `bool`, where it is false.</s>
- years later ... better replace  `$var === null` with `is_null($var)`

This should be far far less changes the DOC update ...

# Goal

... get rid of the supressed erros from below and make it possible to run LTS against phpstan checks.

---

Excluded files:

- controllers .... fails due to autoloading
- all other (Adminhtml isnt complete) fails because of dynamic calls like `Mage::getModel($someVar)` (hast to be fixed in phpstan extension)

Ignored errors:

- most comparison checks are "addressed" with this isue ... but require some small codeshanges ...

Current setup fir doc changes ...

```
includes:
    - vendor/inviqa/phpstan-magento1/extension.neon
parameters:
    reportUnmatchedIgnoredErrors: false
    autoload_files:
        - %currentWorkingDirectory%/app/Mage.php
    excludes_analyse:
        - %currentWorkingDirectory%/app/code/core/Mage/*/controllers/*
        - %currentWorkingDirectory%/app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
        - %currentWorkingDirectory%/app/code/core/Mage/Admin/Model/Acl/Assert/Time.php
        - %currentWorkingDirectory%/app/code/core/Mage/Admin/Model/Config.php
        - %currentWorkingDirectory%/app/code/core/Mage/Admin/Model/User.php
        - %currentWorkingDirectory%/app/code/core/Mage/Admin/Model/Roles.php
        - %currentWorkingDirectory%/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tabs.php
        - %currentWorkingDirectory%/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Widget/Chooser.php
[...]

PR #1495, opened 2021-03-12
Removed unnecessary replacing of variable from parent

Sekiphp:
### Description (*)

`Mage_Directory_Model_Resource_Country_Collection` extending `Varien_Data_Collection` where is `$_items` already defined:
```
    /**
     * Collection items
     *
     * @var array
     */
    protected $_items = array();

```
And there was diferrent initial value.

### DIFF
--- a/app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
+++ b/app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
@@ -30,6 +30,8 @@
  * @category    Mage
  * @package     Mage_Directory
  * @author      Magento Core Team <core@magentocommerce.com>
+ *
+ * @property Mage_Directory_Model_Country[] $_items
  */
 class Mage_Directory_Model_Resource_Country_Collection extends Mage_Core_Model_Resource_Db_Collection_Abstract
 {
--- a/app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
+++ b/app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
@@ -34,11 +34,6 @@
 class Mage_Directory_Model_Resource_Country_Collection extends Mage_Core_Model_Resource_Db_Collection_Abstract
 {
     /**
-     * @var Mage_Directory_Model_Country[]
-     */
-    protected $_items;
-
-    /**
      * Define main table
      */
     protected function _construct()


app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
```diff
@@ -33,11 +33,6 @@
  */
 class Mage_Directory_Model_Resource_Country_Collection extends Mage_Core_Model_Resource_Db_Collection_Abstract
 {
-    /**
-     * @var Mage_Directory_Model_Country[]
-     */
-    protected $_items;
```
midlan, line 39: podle mě to tu bylo schválně kvůli phpdoc
dokážeš tam dostat phpdoc bez přepsání tý proměnný?

midlan, line 39: We will loose phpdoc (typehint) by this?

midlan (approved):

app/code/core/Mage/Directory/Model/Resource/Country/Collection.php
midlan: 👍  resolved in ea13c48a0bbe48dfdc0a786aa0e9ea8decec7173

colinmollenhour (approved):

sreichel:
> The last one in particular is helpful when iterating a collection so that the loop value has correct type hinting automatically, at least in PhpStorm.

Which phpStorm version do you use? I thought this would already work? 

Wait ... no time to check, but I solved in another way. I added something like `@method SomeClass[] getCollection()`, but not for all classes yet. See #784. I'd prefer my way, because all methods are covered. (if i'm right here)

colinmollenhour:
PhpStorm is smart enough to use the return type of `getIterator()` inside a loop to determine the iterator item type because `Varien_Data_Collection` implements `IteratorAggregate`. So if `getIterator()` hints `Mage_Directory_Model_Country[]` then this will work with no additional hints (`$country` will be detected as an instance of `Mage_Directory_Model_Country`):

```php
foreach (Mage::getResourceModel('directory/country_collection') as $country) {
    $country->getIso2Code(); // returns string
}
```

Also this:

```php
Mage::getResourceModel('directory/country_collection')->getFirstItem()->getIso2Code(); // returns string
```


My example with `@method` annotations is the only way I know to get this to work.

sreichel (approved):

sreichel:
@colinmollenhour you're right. `getIterator()` is something i missed for all doc block update :/

LukasCesal (approved):

colinmollenhour (approved):

PR #1497, opened 2021-03-13
Upgrade prototype to version 1.7.3

Sekiphp:
### Description (*)
Upgrade prototype from version `1.7` to `1.7.3`

http://prototypejs.org/

### DIFF
--- a/js/prototype/prototype.js
+++ b/js/prototype/prototype.js
@@ -1,4 +1,4 @@
-/*  Prototype JavaScript framework, version 1.7
+/*  Prototype JavaScript framework, version 1.7.3
  *  (c) 2005-2010 Sam Stephenson
  *
  *  Prototype is freely distributable under the terms of an MIT-style license.
@@ -8,7 +8,7 @@
 
 var Prototype = {
 
-  Version: '1.7',
+  Version: '1.7.3',
 
   Browser: (function(){
     var ua = navigator.userAgent;
@@ -49,7 +49,7 @@ var Prototype = {
     })()
   },
 
-  ScriptFragment: '<script[^>]*>([\\S\\s]*?)<\/script>',
+  ScriptFragment: '<script[^>]*>([\\S\\s]*?)<\/script\\s*>',
   JSONFilter: /^\/\*-secure-([\s\S]*)\*\/\s*$/,
 
   emptyFunction: function() { },
@@ -59,27 +59,6 @@ var Prototype = {
 
 if (Prototype.Browser.MobileSafari)
   Prototype.BrowserFeatures.SpecificElementExtensions = false;
-
-
-var Abstract = { };
-
-
-var Try = {
-  these: function() {
-    var returnValue;
-
-    for (var i = 0, length = arguments.length; i < length; i++) {
-      var lambda = arguments[i];
-      try {
-        returnValue = lambda();
-        break;
-      } catch (e) { }
-    }
-
-    return returnValue;
-  }
-};
-
 /* Based on Alex Arnell's inheritance implementation. */
 
 var Class = (function() {
@@ -141,8 +120,13 @@ var Class = (function() {
           return function() { return ancestor[m].apply(this, arguments); };
         })(property).wrap(method);
 
-        value.valueOf = method.valueOf.bind(method);
-        value.toString = method.toString.bind(method);
+        value.valueOf = (function(method) {
+          return function() { return method.valueOf.call(method); };
+        })(method);
+
+        value.toString = (function(method) {
+          return function() { return method.toString.call(method); };
+        })(method);
       }
       this.prototype[property] = value;
     }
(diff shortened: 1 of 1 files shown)


kiatng:
This update is long overdued. When I compare the official version [here](https://ajax.googleapis.com/ajax/libs/prototype/1.7.3.0/prototype.js) with this PR version, there are lots of discrepancies, all due to 2-spaces indent in the original versus the 4-spaces indent in the PR version. No biggie, but wouldn't it be better to use the official file as is? I think it'll help to verify the file easier.

colinmollenhour:
Just FYI I applied the same update to a different Magento-based project just now that has a fairly rigorous suite of functional tests using Cypress and it passed just fine.

Sekiphp:
@kiatng Fixed :-) I did not realise, that PHPStorm reformated file

kiatng (approved):
Thank you!

colinmollenhour (approved):

LukasCesal (approved):

Flyingmana (approved):
Compared old and new version with original source

empiricompany:
this update of prototype from 1.7 to 1.7.3 break many things with 3rd party extension like proxyblue recaptcha, firecheckout and  i also see many issues opened related to this change.  so...i can ask why this update? there was some security / performance reason?

maybe we can mantain original version in the 19.4.15 release?

**related issues 3rd party extensions: **
firecheckout order review feature (unrecognize input:tel)

immagine

immagine

also in proxyblue recaptcha
https://github.com/ProxiBlue/reCaptcha/issues/56

ADDISON74:
I would recommend reverting to the original file format and keeping this PR open for intense testing. Otherwise we risk even more errors.

kiatng:
Prototype.js 1.7 is not compatible with vuejs, but 1.7.3 is. That is the reason I needed the upgrade.

empiricompany:
@kiatng but vuejs is not used in magento (it would be great if it were 😅), it's for more specific 3rd party extension not compatibile with original magento?

kiatng:
I have 1.7.3 in production for more than a year now, there's no issue with OM so far. For BC to 3rd party extensions, perhaps we can revert prototype.js in ver 19? And keep 1.7.3 in ver 20?

empiricompany:
my opinion is to revert to original prototype.js version because there is no real improvment or security issues and have a big impact in many areas, even because we had to be rewritten original library for issues with serialize function but I leave the choice to the maintainers @colinmollenhour @Flyingmana @LeeSaferite @drobinson @tmotyl @sreichel

sreichel:
If it breaks something, please revert - at least for 19.x branch.

colinmollenhour:
@empiricompany What is the difference between `input:tel` and `input[type=tel]`? The latter works in all Prototype versions and seems to be the correct way anyway, but I'm probably missing something..

I think supporting Vue, a widely used library on the same playing field as jQuery, is a strong reason to keep the latest version. Neither Vue or `input:tel` are used in the core Magento but that doesn't mean there aren't valid uses for them in private code or third-party extensions.

If using `input[type=tel]` is a valid workaround for this issue then why not just go with that as a fairly simple fix? That said, I'm not opposed to keeping the old version for v19 and using the latest for v20+. Are there any other known issues with upgrading to 1.7.3?



The most obvious resolution would be to keep the old in v19 and the new in v20.

empiricompany:
yes it is more correct to use input[type=tel] and the problem with firecheckout is solved (I will report it to them).
The update to prototype 1.7.3 also fixes the reported problem with ProxiBlue/reCaptcha#56 (sorry I got a little confused), although I found another problem in configuration with this commit https://github.com/OpenMage/magento-lts/pull/317/commits for which I will respond in the inherent pull request.
For me, updating the library for vuejs which, unlike jquery, is not used by third party extensions is not correct but it may be my thought too "conservative".
At this point I agree to leave the updated version 1.7.3 for v19 as well.
Sorry for the discussion but I just wanted to go deeper on this choice.

kiatng:
@empiricompany Ref https://github.com/prototypejs/prototype/issues/337, v1.7.3 is not just for vuejs.

empiricompany:
> @empiricompany Ref [prototypejs/prototype#337](https://github.com/prototypejs/prototype/issues/337), v1.7.3 is not just for vuejs.

thanks @kiatng this seems to me a more correct reason to update from my point of view

ADDISON74:
After this PR was merged, comments appeared in which issues related to functionality were reported. There were also some requests for revert. Because Prototype was updated in Magento 2 to the same version, I think we should analyze if they used the version from this PR or a modified one. Here is the PR merged in Magento 2 https://github.com/magento/magento2/pull/33480

ADDISON74:
I made a comparison between the two files and went through it carefully. I didn't find any differences except for some of the shape, added/deleted spaces or lines. 

I recommend those who have issues with this PR to post them in the Issues section, also to follow the messages displayed in the browser console.

Issue #1486, opened 2021-03-04
Northern Ireland (after Brexit)


AlterWeb:
### Description (*)
Northern Ireland has special rules sins the Brexit. For what I understand, it is like for some tax rules they still belong to the EU and are not part of the UK. They got there own vat number starting with XI. For goods delivered to or from Europe they should not calculate or pay tax. This is not possible as of now in OpenMage because they belong to the UK.

So I think we should add Northern Ireland as an extra country to OpenMage. In that way we can apply different tax rules for it and add it to the list of European countries.

Before I will create the code for this in a pull request I have to say I'm not an tax expert and I don't want to break something or add other unexpected behavior. So if anyone can verify that adding a extra country is indeed the right thing to do then I will do so.

For more info see: [https://www.avalara.com/vatlive/en/vat-news/brexit-northern-ireland-vat-and-eoro--xi--number.html](https://www.avalara.com/vatlive/en/vat-news/brexit-northern-ireland-vat-and-eoro--xi--number.html)

### Expected behavior (*)
Companies from Northern Ireland buying in the EU with a valid XI vat number should not pay tax to the European store. 

### Benefits

The right tax rules could be applied.

jfksk:
I'm also not an expert on this matter but I'd like add how I understand it:

VAT/Customs:
NI <-> EU: intra community (for goods)
NI <-> GB: domestic
EU <-> GB: customs (also EU <-> GB via NI)

NI has a new EU VAT ID: XI. This VAT ID can be validated in VIES.
So NI stays in the EU customs union and the EU VAT system for the trade of goods. Normal rules apply. The VAT depends on the merchant and the merchant-customer-relationship (B2B with optional reverse charge or B2C (from July OSS, limit rule or direct)).
However, AFAIK this does not apply to digital services. Digital services sold to consumers in NI are taxable in the UK and not in the EU (OSS for e.g.) The destination country principle applies. Off the top of my head, I don't see any problem with that.

The UK VAT ID (GB) does not have any meaning in the EU anymore, but I dont know about NI <-> GB trade. However, we do not have a validator for it - not anymore. So it doesn't matter...
Customs rules apply for the trade of goods EU <-> GB. The new EORI (XU) is IMHO not of concern for OpenMage as EORI numbers aren't a feature, yet.

The basic B2C setup could probably be done by adding NI as a subdivision of the UK but the VIES validation would need to be adapted (actually quiet a bit) because its currently on country level. 

[...]

PR #1503, opened 2021-03-17
Improved VAT number validation

AlterWeb:
### Description (*)
Improved VAT number validation (VIES):
- added validation for Greece and Northern Ireland
- added validation option for accepting VAT numbers including country code

1. Fixes OpenMage/magento-lts#1486

1. Setup tax rules for Europe (for Northern Ireland use the "United Kingdom" as country and "BT*" for the postcode)
2. Make sure you have configured a VAT number for another EU country for your store and all settings for the Intra Union functionality to work (let me know if you need help with this).
3. Buy a product and use a valid VAT number for Greece (this is a VAT number starting with EL) or Northern Ireland (this is a VAT number starting with XI) but only submit the part after the EL or XI (this is how Magento always worked).
4. If you test for Northern Ireland you also have to provide a valid postcode for the Northern Ireland (starting with BT).
5. The tax should be set to 0.
6. Set the config value for customer/create_account/viv_vat_number_format to 1 (With country code)
7. Repeat step 3 to 5 but now you have to use the whole VAT number including the "country" code prefix.
9. Set the config value for customer/create_account/viv_vat_number_format to 2 (With and without country code)
10. Repeat step 3 to 5, now you can use VAT number with or without the "country" code prefix.

### Questions or comments
1. What do we want the copyright comment to be for newly created files? I think the Magento copyright notice is not right for this or is it? For now I added a simple OpenMage comment block.
2. I created a new model for this functionality because I think it is too much functionality to go in a generic helper. For backwards compatibility I changed the old helper method to use the new model. This however will not work for Northern Ireland because there should also be a postcode provided but this is not available in the helper method.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Customer/Vatnumberformat.php
+++ b/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Customer/Vatnumberformat.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * OpenMage
+ *
+ * @category    Mage
+ * @package     Mage_Adminhtml
+ * @copyright   Copyright (c) 2009-2021 OpenMage (https://www.openmage.org/)
+ */
+class Mage_Adminhtml_Model_System_Config_Source_Customer_Vatnumberformat
+{
+    /**
+     * Options getter
+     *
+     * @return array
+     */
+    public function toOptionArray()
+    {
+        return array(
+            array('value' => 0, 'label'=>Mage::helper('adminhtml')->__('Without country code')),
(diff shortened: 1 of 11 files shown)


ADDISON74:
This PR has conflicts. It needs translations too.

fballiano:
now that we already merged the greek VAT validation it will be more complex to fix this one

fballiano:
At this point I don't know how to resolve the conflicts.

fballiano:
I labeled this as "don't forget this PR" but close it because the conflict are too difficult to fix without the participation of the original author. Feel absolutely free to reopen it with the fixes.

PR #1505, opened 2021-03-22
Fix the deprecation problem of idn_to_ascii() since php7.2

amiriun:
idn_to_ascii() php function is deprecated as of php 7.2
However it is still available but you should use INTL_IDNA_VARIANT_UTS46 as a third argument. (see https://www.php.net/manual/en/function.idn-to-ascii.php)
Otherwise you will see the error because INTL_IDNA_VARIANT_2003 is deprecated.
image

### DIFF
--- a/app/code/core/Zend/Validate/Hostname.php
+++ b/app/code/core/Zend/Validate/Hostname.php
@@ -2512,7 +2512,12 @@ class Zend_Validate_Hostname extends Zend_Validate_Abstract
     protected function checkDnsRecords($hostName)
     {
         if (function_exists('idn_to_ascii')) {
-            $result = checkdnsrr(idn_to_ascii($hostName), 'A');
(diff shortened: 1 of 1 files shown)


kiatng:
There are 4 references to `idn_to_ascii` that may require attention, see [search result](https://github.com/OpenMage/magento-lts/search?q=idn_to_ascii).

See also similar issues in M2 raised in [stackoverflow](https://stackoverflow.com/questions/56661916/deprecated-functionality-idn-to-ascii-magento-2-3-1-with-php-7-2) and in [github](https://github.com/magento/magento2/issues/19795).

kkrieger85 (approved):

midlan (approved):

Sekiphp (approved):

luigifab:
I getting the same error with PHP 7.3.33 when I try to update base url from backend.
```
Deprecated functionality: idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated
  in app/code/core/Mage/Core/Helper/Url.php on line 183
```
😏

PR #1506, opened 2021-03-22
Fix typo in bundle (CACL_ to CALC_)

Sekiphp:
### Description (*)
- fixed typo


jouriy:
please note that this will affect all custom themes (Ultimo, Argento, Meigee blacknwhite, Emthemes, Sm market to name a few) previously used CACL_ prefix copied from defaults, so it will break bundle product prices in these themes.

kkrieger85 (approved):

Flyingmana (changes requested):
Please keep the old variable names active somehow, as this is possible breaking scripts and extensions relying on it. You can mark them with a `@deprecated` annotation

fballiano:
I marked this as draft since the comment from @Flyingmana should be checked

sreichel:
> so it will break bundle product prices in these themes.

It will break every theme that uses this vars.

Issue #379, opened 2017-11-17
Cm_RedisSession default to true


mpingu:
Affected file: /app/etc/modules/Cm_RedisSession.xml

Could we set the Default value for the module to be true?

When you just update the core with unzipping the tar.gz you overwrite this file and deactivate your Redis integration and Magento will write the sessions into the MySQL Database.

seansan:
​+1

haha did not even realize this - as a result, I see we are also using
core_session DB (instead of Redis)​

seansan:
https://github.com/OpenMage/magento-lts/pull/380

tomekjordan:
what if store is not using redis for sessions?

infabo:
It does not harm enabling Cm_RedisSession by default. But IMHO any module not being used, should be disabled. In my experience, only "bigger" shops/setups use Redis for session storage (I can see no reason for Redis on a simple single instance shop with SSD file-storage for example).
Anyone in need for Redis (e.g. on scaling setups) should already have a proper deployment configuration up-and-running, so it is easy to consider "enabling modules" like Cm_RedisSession automagically. Just my 2cents.

tomekjordan:
true. I use Redis only for FPC

colinmollenhour:
There are a couple things going on here:

1. If your method of updating Magento is to untar files into your production directory you're doing it wrong.. If you insist on doing it this way then definitely do not modify any files that are included in the tar file. Perhaps set the read-only flag on all core files to prevent it explicitly. Then, to enable/disable modules add your own `app/etc/modules/Z-local.xml` (Z to sort after all other modules) which will override the existing files without you modifying them.

2. The Magento core code does not allow for easily adding new backends so while it should not even be necessary for Cm_RedisSession to be a proper module (versus just an auto-loaded class), it is required in order to work without modifying core code. However as we actually can modify core code perhaps the best solution would be to allow for other backends so that Cm_RedisSession and the 'db' backend can coexist peacefully. Specifically, here:

https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php#L59

3. Enabling Cm_RedisSession causes a rewrite so I wouldn't say there is 0 impact:
https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/community/Cm/RedisSession/etc/config.xml#L42
This is related to (2) above so if that was fixed this rewrite could be removed and then it would be a non-issue.

[...]

PR #1513, opened 2021-03-29
Add redis as a valid option for global/session_save

rjocoleman:
This is a backwards incompatible change, so I don't think it's a good idea to merge it into 1.9.4.x, however 20.0 perhaps..?

So, this PR:

1. Adds `redis` as an entry in the Core session selection switch statement.
2. Removes the rewrite from `Cm_RedisSession` (to restate `db` when `Cm_RedisSession` is enabled)
3. Enables `Cm_RedisSession`

The impact of this is in `local.xml` `global/session_save` will accept both `db` and `redis` as valid values. 

The reason this is backwards incompatible is `Cm_RedisSession` used to rewrite and replace the `db` handler (if `Cm_RedisSession` was set to enabled/active in its module config.
Thus if merged `Cm_RedisSession` users will revert to DB for session storage.

### Description (*)

This was mentioned here (among other places): https://github.com/OpenMage/magento-lts/issues/379#issuecomment-345328667

1. change local.xml `<global><session_save>` to `redis`
2. configure `local.xml` `redis_session`
3. Create a session with something clearly in it, e.g. add items to the cart and verify they are there after a page load (session is valid)
4. use `redis-cli FLUSHALL` (etc) to drop clear the redis database (this will clear the whole redis db so don't run it on something you care about)
5. Verify session is gone e.g. cart is now empty

### DIFF
--- a/app/code/community/Cm/RedisSession/etc/config.xml
+++ b/app/code/community/Cm/RedisSession/etc/config.xml
@@ -38,11 +38,9 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
     </modules>
     <global>
         <models>
-            <core_mysql4>
-                <rewrite>
-                    <session>Cm_RedisSession_Model_Session</session>
-                </rewrite>
-            </core_mysql4>
+            <cm_redissession>
+                <class>Cm_RedisSession_Model</class>
+            </cm_redissession>
         </models>
     </global>
 </config>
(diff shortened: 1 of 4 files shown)


colinmollenhour:
Thanks for the PR! It would probably require some care in the future not to clobber your changes when updating the module but updates are also quite rare.. :) I'm in favor for 20.x to make it enabled out of the box and hence a little easier to configure.

colinmollenhour (approved):

fballiano (approved):

kkrieger85 (approved):

Issue #1518, opened 2021-04-02
Backend hints are now useless


fballiano:
The backend is showing a couple of "hints" linking to the old magento.com documentation, which is no longer available. Those links were useless when the doc was online, they're now more useless.

Screenshots:
<img width="247" alt="Schermata 2021-04-02 alle 17 07 36" src="https://user-images.githubusercontent.com/909743/113433083-7202c280-93d6-11eb-9c1d-a410b1f735e5.png">
<img width="305" alt="Schermata 2021-04-02 alle 17 07 33" src="https://user-images.githubusercontent.com/909743/113433089-7333ef80-93d6-11eb-80d6-5433bcbb68a4.png">

If you agree I'd create a PR removing all that code, for the v20 branch.

jouriy:
ref #1376

fballiano:
oops sorry! I searched for "hints" and didn't find anything, closing this one!

Issue #1376, opened 2021-01-08
(?) Get help for this page - Just remove it


ADDISON74:
**OpenMage v20.0.5**

Go to Backend. On every page there is a **(?) Get help for this page** link on the top right of the page. All links point to deleted resources on Magento website. This link and all the garbage code should be removed from OpenMage to avoid confusion.

![sceenshot1](https://user-images.githubusercontent.com/8360474/104006784-a720bf80-51af-11eb-8e1a-850e2f6a72ce.jpg)

colinmollenhour:
@Flyingmana do we have a copy of this documentation somewhere? It would be a shame for there to be no user docs at all.. I tried backing up dev docs at https://github.com/OpenMage/devdocs-m1 and https://github.com/OpenMage/devdocs but haven't figured out exactly what is in there..

ADDISON74:
@colinmollenhour, @Flyingmana: Do we have any idea related to this issue?

In case it cannot be implemented I recommend a complete remove from the code.

fballiano:
we have a copy of the documentation but... those hints were never used (also because the documentation was never translated and it really is not user friendly) so I'd like to create a PR to completely remove the "get help for this page" and this hint:

https://user-images.githubusercontent.com/909743/113433089-7333ef80-93d6-11eb-80d6-5433bcbb68a4.png

Flyingmana:
yeah, I think it can be removed then.

Should we want to do something with it, we can still bring it back.

Also linking related Issue here: https://github.com/OpenMage/magento-lts/issues/1374

Issue #1374, opened 2021-01-08
(?) - What is this? tooltip


ADDISON74:
**OpenMage v20.0.5**

1. Go to Backend -> System -> Configuration
2. On the left side in section "Current Configuration Scope" there a (?) tooltip. See the screenshot attached bellow
3. By clicking on the tooltip a new tab will open with Documentation and Resources for Magento 2.

The link for this tooltip is http://merch.docs.magento.com/ce/user_guide/configuration/scope.html

This should be removed from OpenMage. I did not check for other tooltips, probably there are more.

![screenshot](https://user-images.githubusercontent.com/8360474/103999668-db8f7e00-51a5-11eb-986e-147482e1d29b.jpg)

ADDISON74:
Here are the files:

```
./app/code/core/Mage/Adminhtml/Helper/Data.php: $url = "http://merch.docs.magento.com/{$this->getHelpTargetVersion()}/user_guide/";
./app/code/core/Mage/Adminhtml/etc/config.xml: <url><![CDATA[http://merch.docs.magento.com/ce/user_guide/configuration/scope.html]]></url>
```

Here are other links:

```
<a id="page-help-link" href="http://merch.docs.magento.com/ce/user_guide/order-processing/order-manage.html">Get help for this page</a>
<a id="page-help-link" href="http://merch.docs.magento.com/ce/user_guide/configuration/advanced/admin.html">Get help for this page</a>
<a id="page-help-link" href="http://merch.docs.magento.com/ce/user_guide/configuration/general/general.html">Get help for this page</a>
<a id="page-help-link" href="http://merch.docs.magento.com/ce/user_guide/search_seo/search-terms.html">Get help for this page</a>
```

ADDISON74:
@colinmollenhour, @Flyingmana: Do we have any idea related to this issue?

In case it cannot be implemented I recommend a complete remove from the code.

colinmollenhour:
I think there is a copy of the user docs somewhere but I don't know if they have a permanent home yet.. Will defer to @Flyingmana

Flyingmana:
Will check, but also not sure if this is covered by what we have.

in general I would want to keep it, even if we do not have the documentation at the moment. We could at least use it to check on our documentation page, where people more often look at, to extend the docs there with higher priority.

ADDISON74:
[...]

PR #1519, opened 2021-04-02
Remove the documentation Hint "(?)" functionality

fballiano:
This PR should solve https://github.com/OpenMage/magento-lts/issues/1374  https://github.com/OpenMage/magento-lts/issues/1376 and https://github.com/OpenMage/magento-lts/issues/1518 removing links to old magento.com documentation from the legacy backend theme.

I've divided the PR into 4 commits:
- delete the "get help for this page" link as per https://github.com/OpenMage/magento-lts/issues/1376
- delete the "what is this" hint as per https://github.com/OpenMage/magento-lts/issues/1374
- remove all unused PHP code related to the first 2 commits
- remove all unused PHP code related to the generation of "merch.docs.magento.com" links in the "system -> configuration" section of the backend

I've tested the backend after these modifications and it seems to work fine on all sections


kiatng:
Shouldn't this be merged to 1.9.4.x instead of 20.0?

Flyingmana:
Removing the functions has a small chance of beeing a BC incompatibility.
Also we can backport it later, if it proves to be safe.
Also while the current state is not nice, its not necessarily a bug.

Flyingmana (approved):

fballiano:
> Removing the functions has a small chance of beeing a BC incompatibility.
> Also we can backport it later, if it proves to be safe.
> Also while the current state is not nice, its not necessarily a bug.

That's absolutely true, that's why I split the commits.
At the same time removing the code from the templates leaving the "php" logic behind seemed kinda wrong.
Also, for sure nobody ever used/extended this feature ehehheh

:-)

kiatng (approved):

fballiano:
@colinmollenhour what do you think about this?

colinmollenhour (approved):

app/code/core/Mage/Adminhtml/Block/Store/Switcher.php
```diff
@@ -223,38 +223,4 @@ public function hasDefaultOption($hasDefaultOption = null)
...
-
-    /**
-     * Return url for store switcher hint
-     *
-     * @return string
-     */
-    public function getHintUrl()
-    {
-        if (null === $this->_hintUrl) {
-            $this->_hintUrl = Mage::helper('core/hint')->getHintByCode(self::XPATH_HINT_KEY);
```
Sekiphp, line 235: I think, that we should also remove constant `XPATH_HINT_KEY`, which is currently used nowhere

fballiano: you're totally right, I have a problem tho, I've deleted that forked repo cause I messed it up and I think I can't modify the PR anymore :-|

Flyingmana: its a bit "complicated", but you can checkout a PullRequest directly, see step5: https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally

fballiano: I tried but I think I would be able to do actually update the PR only if I had write access to this repo. If seems (https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/checking-out-pull-requests-locally) that I can checkout a PR but then I can only create another one... so if you agree I'd merge this one and I'll create another PR later, just to remove the constant, if that's ok.

Sekiphp: @fballiano I think, that you should create new PR now (with complete code) and link this pull request inside. This PR should be closed, if is not possible to edit it.

fballiano: it seems a bit overkill just to remove an unused constant... anyways... I'll do that, let's not get stuck on that.

fballiano:
I'm closing this PR, because it's been superseeded by https://github.com/OpenMage/magento-lts/pull/1536 (as requested by @Sekiphp)

PR #1523, opened 2021-04-05
don't redirect to "downloader" (which doesn't exists)

fballiano:
In the main index.php there's a check for the app/Mage.php (which is 99.9% useless) and a redirect to the "magento downloader".

Downloader was removed with https://github.com/OpenMage/magento-lts/pull/952 so it's useless to redirect to the downloader subdir. Also, I think the purpose of checkin if the MAGENTO_ROOT . '/app/Mage.php' is useless, one file check for every request while magento wouldn't work at all if it didn't exist, I think that check has to be removed, there's no point in executing it.

### DIFF
--- a/index.php
+++ b/index.php
@@ -42,15 +42,6 @@ define('MAGENTO_ROOT', getcwd());
 $mageFilename = MAGENTO_ROOT . '/app/Mage.php';
 $maintenanceFile = 'maintenance.flag';
 
-if (!file_exists($mageFilename)) {
-    if (is_dir('downloader')) {
-        header("Location: downloader");
-    } else {
-        echo $mageFilename." was not found";
-    }
-    exit;
-}
-
 if (file_exists($maintenanceFile)) {
     include_once dirname(__FILE__) . '/errors/503.php';
     exit;


kiatng (approved):

index.php
```diff
@@ -42,15 +42,6 @@
 $mageFilename = MAGENTO_ROOT . '/app/Mage.php';
 $maintenanceFile = 'maintenance.flag';
 
-if (!file_exists($mageFilename)) {
-    if (is_dir('downloader')) {
-        header("Location: downloader");
-    } else {
-        echo $mageFilename." was not found";
-    }
-    exit;
```
Flyingmana, line 51: this also removes the exit, if the Mage.php does not exist, which I think is not intended.

Flyingmana (changes requested):

index.php
fballiano: It was intentional, quoting my first comment:

Also, I think the purpose of checkin if the MAGENTO_ROOT . '/app/Mage.php' is useless, one file check for every request while magento wouldn't work at all if it didn't exist, I think that check has to be removed, there's no point in executing it.

kiatng (approved):
I agree that the check is not useful. For new install, it may be useful to inform that file is missing, but PHP would throw an error anyway. For development or production, it's totally useless.

fballiano:
I think this should really be merged, the downloader doesn't exist since some time and this check is just a waste of CPU clocks, let's close this and move on to other stuff no? ;-)

I know I'm a pain in the *** but I keep checking for my issues/PR and so I don't use the time to do other stuff for the project.

fballiano:
what should we do about this?

colinmollenhour (approved):

fballiano:
@Flyingmana @sreichel @kiatng I think this one could be easily merged, otherwise close it and let's forget it (but I can't see why)

kiatng:
I agree that this should be merged to v20.

kkrieger85 (approved):

Flyingmana (approved):

PR #1469, opened 2021-02-18
Delete RELEASE_NOTES.txt

fballiano:
RELEASE_NOTES.txt file was removed because unmaintained.

https://github.com/OpenMage/magento-lts/issues/1465

PR #1524, opened 2021-04-05
Delete RELEASE_NOTES.txt

fballiano:
This file was already deleted from 1.9.4 branch but not ported to the 20.0. Reference PR is https://github.com/OpenMage/magento-lts/pull/1469


colinmollenhour (approved):

Flyingmana:
This will happen automatically as part of the release process during the forward porting from 19 => 20

fballiano:
when will it happen? my previous PR was in february? wouldn't it be better for these kind of merges to happen right away? otherwise the more time passes the more the merge could be problematic.

fballiano:
So sorry, I've seen the 20.0.8 release now. closing this one

Flyingmana:
As for now, the merges have usually no conflict, so waiting till release did work out well.
Maybe we will automate it via a cron based github action at one point

Issue #466, opened 2018-03-09
Strange date time issues in customer view


seansan:
Whilst thinking to fix a small error I ran into several strange issues occuring in customer edit view in adminthtml (customer/edit). Several things are unexpected. Let me explain

- Local Time now is May 9 4:32 PM (UTC Now = May 9 3:32)
- Local Timezone Amsterdam/Netherlands
- Local Account creation date is July 27 2015 at 8:38 AM

Below are some strange things

1. The last login date is very strange. Year is incorrect. Time is minutes off.
2. The account creation date is not 7:38 AM it was 8:38 AM (i have also jus confirmed this by creating a new account. Even stranger , the order datetime is 4:32 PM but the account creation datetime - which occured at the same time - shows as 3:32 PM)

So there  seem to be 2 problems
1. getLastLoginDate / logindate date does not seem correct
(however the localized LastLoginStore is! correct)
2. but the same date getter for getStoreCreateDate is not! correct and is 1 hour off (here I would expect 2 lines als showing the alternative localized version)

I hope it is a little clear

![image](https://user-images.githubusercontent.com/652395/37215381-13986e3e-23b8-11e8-85eb-5a716f304c00.png)

seansan:
Possible solution for 1. 
- formatdate() in Mage_Core_Helper_Data has a check for instanceof _zend_date_ but the date passed in getLastLoginDate() in Mage_Adminhtml_Block_Customer_Edit_Tab_View is a member of _Mage_Core_Helper_Data_ and thus returns false
- As a result strtotime is performed on the date object [here ](https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Core/Helper/Data.php#L175)
- A possible fix is to add the following to getLastLoginDate() [here](https://github.com/OpenMage/magento-lts/blob/1.9.3.x/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php#L106)
```
        $date = $this->getCustomerLog()->getLoginAtTimestamp();
        $date = new Zend_Date($date, Zend_Date::TIMESTAMP);
```

ADDISON74:
I discovered this issue today in OpenMage 20.0.7 instance in production. A fix should be implemented.

PR #1525, opened 2021-04-06
Fixed incorrect datetime in block.

kiatng:
### Description (*)

This PR fixes 2 issues:
1. Incorrect datetime in backend customer page, see issue #466.
2. Enable the ability to `formatDate()` in the block without converting to store timezone. A use case is when a datetime is stored in the DB in local timezone (or a date such as birthdate, which is independent of timezone), the method `formatDate()` would convert the datetime and couldn't be used for rendering. This PR fixes that. Now, the same method can be used.

1. Fixes OpenMage/magento-lts#466

After:
image

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
@@ -78,11 +78,9 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_View
      */
     public function getCreateDate()
     {
-        if ( ! $this->getCustomer()->getCreatedAt()) {
-            return null;
-        }
-        return $this->_getCoreHelper()->formatDate($this->getCustomer()->getCreatedAt(),
-            Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true);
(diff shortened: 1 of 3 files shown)


ADDISON74:
I do not agree changing the code by replacing **if/else** sequences with ternary operator **?**. The code is just fine as it is the only necessary change should be fixing the issue.

Although I understand that you want a compression of the code but it is not beneficial for the reader. It is much easier for me personally to see if/else sequences then ?. If we agree to such changes then we should do so throughout Magento, otherwise we lose the rules initially set.

fballiano:
I also don't like the 1 line functions, but anyways we need this merged asap

colinmollenhour (approved):

sreichel:
Is this a BC breaking change?

kiatng:
> Is this a BC breaking change?

I think so, due to addition of new param `$useTimezone`  in `public function formatDate($date = null, $format = Mage_Core_Model_Locale::FORMAT_TYPE_SHORT, $showTime = false, $useTimezone = true)`. Existing method without `$useTimezone` that overrides it will run into PHP exceptions.

sreichel:
IMHO ... this change to `formatDate()` method should be reverted.

I'd add a new method that covers `$useTimezone = true` parameter and can be used to fix the issue.

Issue #134, opened 2016-11-05
Improve global search (note)


seansan:
This is a note to improve global search. If someone already has some working code then this is great. 

Things to improve, maybe there is more

- replace search with %STRING%, orders numbers like %STRING (often we only get last couple of digits xxx not the leading 300000xxxx)
- search for lastname, firstname, email, order number
- if ordernumber found then show orders first
- if customer data used to localize then show customer data first
- improve 2nd line small font span element (giving extra information)
- add store name to order number in results
- sort by last created orders first (this is probably what the customer is calling about)
- add some index because the search can be really really slow (try using it with someone on the telephone)

colinmollenhour:
We had to improve global search by completely disabling it.. On a large store it becomes completely unusable.. Maybe with some FULLTEXT indexes it could be made usable but in general using `LIKE %..` only works for smaller stores where full table scans are doable. On another installation it wasn't so bad when we replaced all of the `LIKE %...%` with `LIKE ...%` so that indexes are used.

seansan:
I agree - haha turning it off always makes it sound bad. There must be a working solution. Did you ever find alternatives? Or maybe we should add LIKE ...% & maybe even add an index here and there

My post was mainly to capture: 
- what do we find important to search for
- how do we speed this up (query + add. index)
- how can we better present it

colinmollenhour:
We have do not have an alternative for global search, we just quit using it in favor of improving grid search. I changed the default grid searching behavior to use `LIKE ...%` by default (to be faster) and added some magic to it so that it supports things like IN, REGEX, LIKE %...% (by specifying *...* for the lay-person), etc.

colinmollenhour:
I think ideally it would be replaced with something like elasticsearch but I think that it out of the scope of LTS project.

seansan:
@colinmollenhour I read that you did some manual changes to your installation. Is there something to move/push to LTS ... or should we rather close the issue?

(for now we dont use it either)

sreichel:
Any updates here? Make it configurable?

colinmollenhour:
I think just replacing all `LIKE %...%` with `LIKE ...%` makes sense because it is no good if it isn't performant. All of the other suggestions @seansan had generally sound good so long as they don't impact performance. In general the rule should be that global search only does queries that are able to effectively use an index.

seansan:
Not sure what to do here. We completely rewrote search based on above and specificlly for our site (and products). Great improvements can be made here though .... 

[...]

PR #1532, opened 2021-04-07
Possibility to disable global search in backend

fballiano:
Global search in the backend can be very resource heavy on big stores, there's an ACL to configure if a user can use it or not but I thought it would be handy to have a global flag to turn it off for everybody (most of the times a lot of backend users have "full on privileges" without a proper ACL configuration) and prevent backend users to potentially run extremely heavy queries.

This is the global search:
Schermata 2021-04-02 alle 21 57 27

This is the flag:
Schermata 2021-04-02 alle 21 57 49

Global search is enabled by default (after this PR), so no breaking change is added.

1. login to the backend
2. go to "system -> configuration -> advanced -> admin"
3. expand the global search section and disable global search
4. save the configuration and check that the global search isn't visible anymore

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/IndexController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/IndexController.php
@@ -104,7 +104,7 @@ class Mage_Adminhtml_IndexController extends Mage_Adminhtml_Controller_Action
         $searchModules = Mage::getConfig()->getNode("adminhtml/global_search");
         $items = array();
 
-        if (!Mage::getStoreConfig('admin/global_search/enable') || !Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
+        if (!Mage::getStoreConfigFlag('admin/global_search/enable') || !Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
             $items[] = array(
                 'id' => 'error',
                 'type' => Mage::helper('adminhtml')->__('Error'),
--- a/app/code/core/Mage/Adminhtml/controllers/IndexController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/IndexController.php
@@ -104,7 +104,7 @@ class Mage_Adminhtml_IndexController extends Mage_Adminhtml_Controller_Action
         $searchModules = Mage::getConfig()->getNode("adminhtml/global_search");
         $items = array();
 
-        if (!Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
+        if (!Mage::getStoreConfig('admin/global_search/enable') || !Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
             $items[] = array(
                 'id' => 'error',
                 'type' => Mage::helper('adminhtml')->__('Error'),
--- a/app/code/core/Mage/Core/etc/system.xml
+++ b/app/code/core/Mage/Core/etc/system.xml
@@ -1113,6 +1113,25 @@
                     </fields>
                 </design>
 
+                <global_search>
+                    <label>Global Search</label>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <sort_order>6</sort_order>
+                    <fields>
+                        <enable translate="label comment">
+                            <label>Enable global search</label>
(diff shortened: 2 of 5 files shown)


colinmollenhour:
No objection but really the global search just needs to be fixed. See #134:

> I think just replacing all LIKE %...% with LIKE ...% makes sense because it is no good if it isn't performant. ... In general the rule should be that global search only does queries that are able to effectively use an index.

colinmollenhour (approved):

fballiano:
totally agree about the fixes, but (on the stores I manage) I know for sure none uses it and I'd prefer to disable it altogether :-)

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1113,6 +1113,25 @@
...
                 </design>
 
+                <global_search>
+                    <label>Global Search</label>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <sort_order>6</sort_order>
+                    <fields>
+                        <enable translate="label comment">
```
Sekiphp, line 1123: Little suggestion - there is no comment :-)

fballiano: great eye! fixed

app/code/core/Mage/Adminhtml/controllers/IndexController.php
```diff
@@ -104,7 +104,7 @@ public function globalSearchAction()
         $searchModules = Mage::getConfig()->getNode("adminhtml/global_search");
         $items = array();
 
-        if (!Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
+        if (!Mage::getStoreConfig('admin/global_search/enable') || !Mage::getSingleton('admin/session')->isAllowed('admin/global_search')) {
```
luigifab, line 107: getStoreConfigFlag :)

fballiano: right, done!

fballiano:
I see a problem, when we need to modify a PR we lose traction and than we forget about it, reviews were already positive but now, because of a small change, they're stuck again :-(

app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -32,7 +32,7 @@
         <p class="super">
             <?php echo $this->__("Logged in as %s", $this->escapeHtml($this->getUser()->getUsername())) ?><span class="separator">|</span><?php echo $this->formatDate(null, 'full') ?><span class="separator">|</span><a href="<?php echo $this->getLogoutLink() ?>" class="link-logout"><?php echo $this->__('Log Out') ?></a>
         </p>
-        <?php if ( Mage::getSingleton('admin/session')->isAllowed('admin/global_search') ): ?>
+        <?php if ( Mage::getStoreConfig('admin/global_search/enable') && Mage::getSingleton('admin/session')->isAllowed('admin/global_search') ): ?>
```
luigifab, line 35: getStoreConfigFlag :)

app/design/adminhtml/default/openmage/template/page/header.phtml
```diff
@@ -49,7 +49,7 @@
 		        </ul>
 		    </li>
 		</ul>
-        <?php if ( Mage::getSingleton('admin/session')->isAllowed('admin/global_search') ): ?>
+        <?php if ( Mage::getStoreConfig('admin/global_search/enable') && Mage::getSingleton('admin/session')->isAllowed('admin/global_search') ): ?>
```
luigifab, line 52: getStoreConfigFlag :)

fballiano: @luigifab totally right, PR updated!

app/design/adminhtml/default/default/template/page/header.phtml
fballiano: @luigifab totally right, PR updated!

colinmollenhour (approved):

kkrieger85 (approved):

sreichel:
Is this a BC breaking change?

fballiano:
It shouldn’t be

sreichel:
Then it should be backported to 19.4.x, not?!?

Issue #569, opened 2018-11-20
add and advertise Firegento_Logger as suggested Module on website


Flyingmana:
with #568 and #556 we had two Requests, which already can get solved by the 3th party module Firegento_Logger.
As there is a general interest, we should make it easy to find and understand this modules.

sreichel:
Add it to composer suggest? :+1:

fballiano:
I think composer suggest is the best way at the moment. I wouldn't include a new module to the default install but suggesting it seems perfect.

fballiano:
with this PR: https://github.com/OpenMage/documentation/pull/3 on the documentation repo we should be able to close this issue (let's close it already, it's on the docu repo anyways)

PR #1533, opened 2021-04-07
advertise firegento_logger

fballiano:
As per https://github.com/OpenMage/magento-lts/issues/569 let's help people find a better logging solution.

It's not in composer suggest but I (now) think this is the best (and less intrusive, and more easily findable) way to advertise important 3rd party modules

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ There are some new or changed translations, if you want add them to your locale 
 - `Mage_Xmlconnect`
 - `Phoenix_Moneybookers`
 
+### Suggested Modules
+- [FireGento_Logger](https://github.com/firegento/firegento-logger)
+
 ## Development Environment with ddev
 - Install [ddev](https://ddev.com/get-started/)
 - Clone the repository as described in Installation -> Using Git


colinmollenhour:
One impediment to having this section grow will be the fact that it is part of the main repo that has requirements for multiple code reviews, etc. It creates noise for code reviewers and then the maintainer managing releases has to mess with porting commits between branches and the commit log gets cluttered with README updates.

What about instead starting a "OpenMage/awesome-openmage" project and letting a few people such as yourself maintain it much more freely and referencing this project prominently in the README? This will be less busy-work for the reviewers and maintainers and allow it to grow much more rapidly and keep non-code commits out of the main repo.

fballiano:
totally makes sense, when we'll have the new repo I'll change the PR ;-)

fballiano:
any chance we can create the repo so to close this and move it forward?

Flyingmana:
maybe thats more something for the (way to empty) documentation.
We have an area for Modules there, see https://openmage.readthedocs.io/en/latest/modules/index.html

The related repository is https://github.com/OpenMage/documentation/tree/master/modules

fballiano:
@Flyingmana you're totally correct, I'll PR on that repo and I'll link it in https://github.com/OpenMage/magento-lts/issues/569

PR #1536, opened 2021-04-09
Remove hints

fballiano:
This PR replaces https://github.com/OpenMage/magento-lts/pull/1519 because I couldn't add a commit to that one (because of a technical problem).

The only difference is this new commit:
https://github.com/OpenMage/magento-lts/commit/753f46137c14d39fbded64365b4496455e7d6b3a
which removes an unused constant.


Sekiphp (approved):
Thank you for new PR :-)

fballiano:
@colinmollenhour and @Flyingmana sorry to bother, we had your positive review on the original version of this PR, could you take a look at this new version? it's exactly the same thing, plus a single mega-easy commit:
https://github.com/OpenMage/magento-lts/pull/1536/commits/753f46137c14d39fbded64365b4496455e7d6b3a

thanks! :-)

Flyingmana (approved):
the other commits are identical to the referenced PR

PR #1544, opened 2021-04-11
Prevent backend_type from being changed automatically on attribute save without the admin realizing

woutersamaey:
### Description (*)

If an attribute is user defined, the backend type is automatically set every time you save the attribute. If another backend_type was in use before, this breaks the attribute without the admin realizing it. We cannot have a different backend_type unless we also make the attribute is_user_defined = 0. Making the attribute non-user defined, might seam as the solution, but if you do this, the attribute becomes locked in the set and it cannot be removed anymore, so you're stuck with either e locked attribute that must remain in the set, or an attribute that will change backend_type on every save.
My code change prevents the auto-changing of the backend_type. Only if there wasn't a backend_type before, will it now be auto-detected.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/AttributeController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/Product/AttributeController.php
@@ -275,7 +275,7 @@ class Mage_Adminhtml_Catalog_Product_AttributeController extends Mage_Adminhtml_
                 $data['is_filterable_in_search'] = 0;
             }
 
-            if (is_null($model->getIsUserDefined()) || $model->getIsUserDefined() != 0) {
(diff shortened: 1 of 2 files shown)


kiatng (approved):
A use case is for a user-defined attribute with the following properties:

1. `frontend_input` is `select` or `multiselect`
2. `source_model` is a custom model with value of type `varchar`

In this case, the `backend_type` should be `varchar` but it will be set as `int`. This commit will fix this.

woutersamaey:
Exactly. I'm working on a very big store and am
optimizing every single attribute for performance and storage. This "bug" has been very annoying as multiple people are opening and saving attributes.

colinmollenhour (approved):
Nice fix. Thanks!

fballiano (approved):

PR #1551, opened 2021-04-16
Removed "admin routing compatibility mode" for extensions

luigifab:
### Description

After #1470, this PR remove **Admin routing compatibility mode for extensions**.
Yes, it can break some old modules 💥 .

### Contribution checklist


fballiano:
would it be possible to also delete the admin/security/extensions_compatibility_mode entry from core_config_data?

Flyingmana:
is there a specific reason for this @fballiano? I ask, because this would break the ability to roll back to the previous version, while I not see a great advantage in return.

fballiano:
> is there a specific reason for this @fballiano? I ask, because this would break the ability to roll back to the previous version, while I not see a great advantage in return.

mmm I thought about not leaving an orphan record that will never be used again but yes, the rollback possibility is probably worth leaving it there :-)

sreichel (changes requested):
I'd not remove that code, but set default value to 0 in config.xml (and update README with link to information what this is about).

fballiano (approved):
I think this one should be merged as-is in `next`

elidrissidev:
I'm not very familiar with the old way of defining admin routing, but I did some searching within the codebase and it seems that `Mage_Adminhtml` is still using the old approach, not sure if that should be changed as part of this PR?

https://github.com/OpenMage/magento-lts/blob/7f73fa548d7bb79b9055d1c95b64a37f816ff127/app/code/core/Mage/Adminhtml/etc/config.xml#L82-L90

kyrena:
I think that for Mage_Adminhtml it's normal and not a problem.
But it might be a good idea to add an example somewhere: _bad and old way: ..., good way: ..._.

fballiano:
I've re-tested this PR today and all of the backend works fine, @elidrissidev do you still have doubts about it?

kiatng (approved):
+1 for security.

fballiano:
great, I'll wait a bit more and then merge it, it's in `next` anyway so..

elidrissidev (approved):

elidrissidev:
LGTM

fballiano:
oki then :-)

ADDISON74:
This one needs a note in the README file.

fballiano:
@ADDISON74 done in https://github.com/OpenMage/magento-lts/commit/5655b4b62108a08398ddb43b779b990cc75ae370

ADDISON74:
Very good, don't let anyone say that the modification was not made public ...

PR #1555, opened 2021-04-19
Performance improvement for catalog product images

luigifab:
### Description

This PR reduce calls for:
* `getSingleton` and `getBaseMediaPath`
* `_rgbToString`

OpenMage 20.0.8 / PHP 8.0.3

### Manual testing scenarios

Go to any pages where product images are visible.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Image.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Image.php
@@ -76,7 +76,8 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
      *
      * @var array
      */
-    protected $_backgroundColor  = array(255, 255, 255);
+    protected $_backgroundColor    = array(255, 255, 255);
+    protected $_backgroundColorStr = 'ffffff';
 
     /**
      * Absolute path to and original (full resolution) image
@@ -107,6 +108,11 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     protected $_watermarkWidth;
     protected $_watermarkHeigth;
     protected $_watermarkImageOpacity = 70;
+
+    /**
+     * @var string directory
+     */
+    protected static $_baseMediaPath;
 
     /**
      * @param int $width
@@ -213,6 +219,7 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     public function setBackgroundColor(array $rgbArray)
     {
         $this->_backgroundColor = $rgbArray;
+        $this->_backgroundColorStr = $this->_rgbToString($rgbArray);
         return $this;
     }
 
@@ -347,7 +354,11 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
         if (($file) && (0 !== strpos($file, '/', 0))) {
             $file = '/' . $file;
         }
-        $baseDir = Mage::getSingleton('catalog/product_media_config')->getBaseMediaPath();
+
+        if (empty(self::$_baseMediaPath)) {
+            self::$_baseMediaPath = Mage::getSingleton('catalog/product_media_config')->getBaseMediaPath();
+        }
+        $baseDir = self::$_baseMediaPath;
 
         if ('/no_selection' == $file) {
             $file = null;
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -76,7 +76,8 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
      *
      * @var array
      */
-    protected $_backgroundColor  = array(255, 255, 255);
+    protected $_backgroundColor    = array(255, 255, 255);
+    protected $_backgroundColorStr = 'ffffff';
```
tmotyl, line 80: is saving a single call to _rgbToString worth creating another property? From my POV its not.

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -108,6 +109,11 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     protected $_watermarkHeigth;
     protected $_watermarkImageOpacity = 70;
 
+    /**
+     * @var string directory
```
tmotyl, line 113: what directory? absolute or relative path? with or without trailing slash?

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -347,7 +354,11 @@ public function setBaseFile($file)
         if (($file) && (0 !== strpos($file, '/', 0))) {
             $file = '/' . $file;
         }
-        $baseDir = Mage::getSingleton('catalog/product_media_config')->getBaseMediaPath();
+
+        if (empty(self::$_baseMediaPath)) {
+            self::$_baseMediaPath = Mage::getSingleton('catalog/product_media_config')->getBaseMediaPath();
```
tmotyl, line 359: why make it static? 
I would prefer to have a getter with this logic.

luigifab:
This is not a single call, because `Mage_Catalog_Model_Product_Image` is not a singleton.

Before / After:
image image

For example, for 2592x `setBaseFile`:
* 5184x `getBaseMediaPath`
* 2592x _rgbToString
* 5184x `getSingleton`

tmotyl:
How many images are you processing ? is it some batch job or a caegory page?

luigifab:
It's a category with 81 products, 162 pictures tags (one picture for default, another picture on hover).
Each `<picture>` tag contains 8 `<source>` tag with 2 urls (1x and 2x).

colinmollenhour (dismissed):
Nice work @luigifab! I get the desire to reduce additional class properties and such but sometimes the benefits far outweigh the few lines of extra code and this looks like a very clean implementation to me. 156ms is a nice improvement

fballiano (approved):

app/code/core/Mage/Catalog/Model/Product/Image.php
Flyingmana: I would say its ok, but its a BC break if someone extended this class and relies on the `$_backgroundColor` to be changed.

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -108,6 +109,11 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     protected $_watermarkHeigth;
     protected $_watermarkImageOpacity = 70;
 
+    /**
+     * @var string directory
+     */
+    protected static $_baseMediaPath;
```
Flyingmana, line 115: this introduces an additional state to the Application which for some usecases needs a special cleanup.
see https://github.com/OpenMage/magento-lts/blob/d1c7b4bb8ad806b50e510c20db783861c6735f82/app/Mage.php#L258-L274

Usecases can be changes while doing unittests, or running the project inside an Application Server in a permanent process. Although this is are rather rare cases, so I just moved the target branch to 20.0.

Still we should think about a general solution, how to manage such a "cache" in a easy to cleanup way.

Flyingmana (approved):

Flyingmana:
Still merged it, because we already had 2 approvals before mine, and they were only dismissed because of the Branch change from 1.9.4.x to 20.0

PR #1557, opened 2021-04-20
Mark invoice as last when dummy items included

mattdavenport:
This commit fixes invoice generation when an order contains configurable or bundled items.

This is a [backported fix](https://github.com/magento/magento2/pull/31472) from Magento 2, as this bug exists in M1 as well. When a configurable or dummy item is included on an order, and custom tax collectors are used the tax total is off due to `$invoice->isLast()` being false and the code choosing the tax total of only items+shipping (in `Mage_Sales_Model_Order_Invoice_Total_Tax`). This has been testing with multiple-invoiced orders as well and works as expected.

Thanks!

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Invoice.php
+++ b/app/code/core/Mage/Sales/Model/Order/Invoice.php
@@ -726,6 +726,11 @@ class Mage_Sales_Model_Order_Invoice extends Mage_Sales_Model_Abstract
     public function isLast()
     {
         foreach ($this->getAllItems() as $item) {
+            $orderItem = $item->getOrderItem();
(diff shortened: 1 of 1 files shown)


colinmollenhour (approved):
Very nice!

Flyingmana (approved):

Flyingmana:
there is a comparable part in \Mage_Sales_Model_Order_Creditmemo::isLast https://github.com/OpenMage/magento-lts/blob/f37c26d8103d06f0fdae8cce33ba1edd521f6167/app/code/core/Mage/Sales/Model/Order/Creditmemo.php#L1023
Can we have dummy Items there, too?

PR #1568, opened 2021-04-23
Fixes for #1564 and #1289

### LINKED ISSUE #1564: Checkout - Agreement content when empty (Terms and Conditions)
With #1480 the Content field is no longer required which is a normal behavior. This field is not always used.

agreement_2

A strange behavior occurs in frontend if I don't fill in the textarea. 

agreement_1

As you can see although I did not enter a content for the first term I get an empty div. 

This template **/frontend/base/default/template/checkout/onepage/agreements.phtml** was not intended for such a situation so it should be changed. If there is no value for Content then do not display that div.

I would make another change in the Sales menu. "Terms and conditions" should become "Terms and Conditions" to keep the same uppercase format like for other links. 

If I have some time I will create this commit.

ADDISON74:
This PR fixes #1564 and #1289 . Please read bellow my comments for every commit.


ADDISON74:
I will close this issue. Please check #1568 for a fix.

app/code/core/Mage/Catalog/Model/Resource/Product/Compare/Item/Collection.php
```diff
@@ -236,13 +234,18 @@ public function getComparableAttributes()
...
+                        'ai.attribute_id = main_table.attribute_id'
                     )
                     ->where('additional_table.is_comparable=?', 1)
-                    ->where('main_table.attribute_id IN(?)', $attributeIds);
+                    ->where('ai.attribute_set_id IN(?)', $setIds)
+                    ->order(array('ai.attribute_group_id ASC', 'ai.sort_order ASC'));
                 $attributesData = $this->getConnection()->fetchAll($select);
                 if ($attributesData) {
-                    $entityType = Mage_Catalog_Model_Product::ENTITY;
+                    $entityType = 'catalog_product';
```
colinmollenhour, line 248: What is the purpose of ceasing to use a constant here?

kiatng:
The 2 issues are not related and should be split into 2 separate PRs. It will be easier to review.

ADDISON74:
@kiatng: I used the GitHub interface when I posted them and I'm not very familiar with it. I'm sorry for this inconvenience.

The only commit that needs attention is the one related to the order of the attributes in the comparison window. The code does not belong to me, but it works, I have been testing it for over 5 years. It may require optimization or another approach. In any case, this issue had to be solved for a long time.

app/code/core/Mage/Catalog/Model/Resource/Product/Compare/Item/Collection.php
ADDISON74: I borrowed the code here https://magento.stackexchange.com/questions/64983/attributes-order-in-comparison-window-is-wrong. I did a PR to discuss it. I tested both changes and they do not bring any changes. I will modify the code to keep the initial version. Although the code has been working for years without any issues, what I would be interested in is to see if the change has the format you are taught. Also, if the database query is optimized.

kiatng (approved):

fballiano (dismissed):

colinmollenhour (dismissed):

kkrieger85 (dismissed):

ADDISON74:
I did not find in the latest OpenMage the changes related to this #1289, even this PR was approved and fix it. 

Maybe I am wrong ...

ADDISON74:
In this file

/app/code/core/Mage/Catalog/Model/Resource/Product/Compare/Item/Collection.php

**line 209 function getComparableAttributes(**) was change again after this PR to the initial behavior. This is the reason the attributes in comparison window are not following the order set up the Backend.

ADDISON74:
This PR is responsible https://github.com/OpenMage/magento-lts/pull/2993. The function mentioned above has been radically modified, but the check whether the attributes are in order or not has not been done. fortunately, thanks to the migration of a store where I had custom changes, I was able to identify this problem. my opinion that that PR needs to be reviewed carefully.

@davidhiendl - may I ask why you changed the function getComparableAttributes() although it solved the problem related to the order of the attributes in the comparison window?

davidhiendl:
@ADDISON74 The aim of the PR was to reduce the countless DB queries maybe for fetching attributes for a single batch call to cache storage or database. `getComparableAttributes` however does not have any impact on the layered navigation plus I cannot reproduce the problem. Also the new function implements the same database logic from cached data: 
- fetch attributes for `setIds`
- filter by `is_comparable`
- sort order by position

ADDISON74:
My PR was related to the wrong positioning of the attributes in the comparison window. Your PR came after and modified my changes and now I am not able to see the attributes in the order I set up. 

It is not related to layering navigation.

Issue #1070, opened 2020-06-28
Change Github REPO description/about


seansan:
### Description (*)
This project is not listed at its best in google and this can be improved. We need to be found on search strings like "Magento", "Magento 1", "Magento 1.9" and "Magento LTS" or "Openmage"

### Expected behavior (*)
Search for Magento LS in Google

Instead of the current result
![image](https://user-images.githubusercontent.com/652395/85947791-1d18b400-b94d-11ea-94e1-5bfc91d3a7cc.png)

It could look like
![image](https://user-images.githubusercontent.com/652395/85947781-168a3c80-b94d-11ea-8112-3cd416458c13.png)

### Additional information
The information in Google comes from here
![image](https://user-images.githubusercontent.com/652395/85947825-476a7180-b94d-11ea-80a1-e234a3091264.png)

**Change to? (just suggesting)**

Before the DASH = title? => "Official OpenMage Codebase"

After the DASH = description? => "Download or contribute OpenMage. Empowering thousands of Magento 1 stores with the best eCommerce platform around. Community driven Long Term Support for Magento 1.9.x."

simbus82:
Another tip: to split concepts, you can use the | instead of the -. It takes up fewer pixels. The visual limit in the SERP of the title and description is not in characters, but in pixels. ;-)

colinmollenhour:
Thanks for the suggestions, description updated.

seansan:
> Another tip: if you have to split concepts and do something visually engaging (in description for example) , you can use the | instead of the -. It takes up fewer pixels because the visual limit in the SERP of the title and description is not in characters, but in pixels. ;-)
> 
> To make the repository easily found in Google, the Readme file is also very important.
> So an SEO optimization can also be done there, respecting the headings (h1, h2, etc.) and adding different information about "answers" to be given to users' questions (queries).
> 
> Do not worry if you go to compete with the same OpenMage site, 2 results in SERP are always better than one ;-)

Maybe you can make suggestions by editing the README? and create a pull request

simbus82:
> > Another tip: if you have to split concepts and do something visually engaging (in description for example) , you can use the | instead of the -. It takes up fewer pixels because the visual limit in the SERP of the title and description is not in characters, but in pixels. ;-)
> > To make the repository easily found in Google, the Readme file is also very important.
> > So an SEO optimization can also be done there, respecting the headings (h1, h2, etc.) and adding different information about "answers" to be given to users' questions (queries).
[...]

PR #1572, opened 2021-04-27
Update README.md

fballiano:
Proposed rebranding of the official repository title.

I think it's time for OpenMage to be called OpenMage, not Magento LTS.

At the moment this change is just about the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,11 @@
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/license.svg" alt="License"></a>
 </p>
 
-# Magento - Long Term Support
+# OpenMage
 
-This repository is the home of an **unofficial** community-driven project. It's goal is to be a dependable alternative
-to the Magento CE official releases which integrates improvements directly from the community while maintaining a high
-level of backwards compatibility to the official releases.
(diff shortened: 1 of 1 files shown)


colinmollenhour (changes requested):
I don't know that we want to say "that took over Magento 1" since although we may be the defacto or most popular continuation of the Magento 1 legacy, we did not acquire any special IP rights or license or blessing from Adobe. I think from a legal standpoint we should not make this claim or at least in these terms.

fballiano:
what about instead of "took over Magento 1" it was "is continuing the free/libre open source development of Magento 1"?

Flyingmana:
Generally I support the rebranding.

But also keep in mind how this affects "SEO", or better said, ways how people find us.
When choosing the wording, think of what would people search for. 
And especially as Adobe seems to phase out the usage of the Term "Magento" I would like to keep this special part.

image

kkrieger85 (approved):

Flyingmana:
todo: check SEO requirements from https://github.com/OpenMage/magento-lts/issues/1070 after change

fballiano:
I do not agree on the direction that being taken regarding the description of openmage and this repo, so I guess I'm not the right person to update the README.

Some PRs take ages to be checked and the description of the repo (which is quite important) gets changed in few hours... This project should have the same way of dealing with things for everything.

Issue #1590, opened 2021-05-03
Magento logo on Demo store notification


ProxiBlue:
### Preconditions (*)

1. 19.4.5

### Steps to reproduce (*)

1. Enable demo store notification

### Expected result (*)

1. Openmage logo in top bar with notification

### Actual result (*)

1. Magento logo is shown
2. 
![image](https://user-images.githubusercontent.com/4994260/116837525-684ec380-abfd-11eb-9475-ef40282586f7.png)

luigifab:
Can be closed :)

PR #1591, opened 2021-05-03
removed magento logo in demo notice

fballiano:
In issue https://github.com/OpenMage/magento-lts/issues/1590 @ProxiBlue shows that the demo message still has the magento logo.

This PR removes the logo (there's a logo literally 20px to the bottom and it was used only for that message) and centers the notice text.

This is the new style:
Schermata 2021-05-03 alle 19 10 31


colinmollenhour (approved):

fballiano:
@Flyingmana I think there's a bug in the unit testing, they fail on every PR since a few days

kkrieger85 (approved):

Issue #1595, opened 2021-05-04
Wrong Link Twitter Shared Broken


nihadrekany:
hi team ,,
i want to share shoes on my twitter but when i click to twitter icon i will redirected to home not to share the link is not check
Step
1- go to https://demo.openmage.org/accessories/shoes/barclay-d-orsay-pump-nude.html?___store=default
2- click to twitter share it is redirect to twitter home not shared .
vuln Link
http://twitter.com/home?status=Barclay+d%27Orsay+pump%2C+Nude+https%3A%2F%2Fdemo.openmage.org%2Fbarclay-d-orsay-pump-nude.html%3F___SID%3DU
removed this link and added this link
https://twitter.com/intent/tweet?url=https://demo.openmage.org/barclay-d-orsay-pump-nude.html?___SID=U
Impact
wrong twitter share

fballiano:
I've just posted PR https://github.com/OpenMage/magento-lts/pull/1598 to fix this.

sreichel:
@all-contributors add @nihadrekany bug

PR #1598, opened 2021-05-04
fixed twitter share link, issue 1595

fballiano:
This PR should fix https://github.com/OpenMage/magento-lts/issues/1595
It's done in a slightly different way (to keep the product name in the shared tweet) compared to the one suggested by the original reporter.

### DIFF
--- a/app/design/frontend/rwd/default/template/catalog/product/view/sharing.phtml
+++ b/app/design/frontend/rwd/default/template/catalog/product/view/sharing.phtml
@@ -44,7 +44,7 @@
         </a>
     </li>
     <li>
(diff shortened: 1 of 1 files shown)


Flyingmana (approved):

Flyingmana:
yeah, we can do the "proper" sending of the url via separate parameter as a separate task, if it is offering an improvement

fballiano:
I didn't think about doing that but makes sense, I separated the URL parameter already and updated the PR.

colinmollenhour (approved):

Flyingmana (approved):

PR #1600, opened 2021-05-05
Remove "default" themes from 20.x

fballiano:
This PR removes app/design/frontend/default and skin/frontend/default.

Those themes are very old, not good looking and nobody (who keeps magento somewhat updated) is using them.

With this PR we delete almost 580 (useless) files from the repository.

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/1037

Just browse a new installation and see that the RWD theme (the default since some time) doesn't need the old "default" ones to work.

### Questions or comments
- The only people who could have problems with this PR is somebody using composer. But really, who's using composer and this extremely old themes at the same time? If they're not using composer and just copying the new version on top of the other there will be no problem cause the old files with still be there.
- I've big stores in productions without these 2 folders since years without any problems (with commercial themes or custom made ones)


ADDISON74:
I agree. Relics from the time when the design was not responsive.

PS - /errors directory should be revised too.

luigifab:
This remove the legacy frontend theme:

image

I don't know what to think.

colinmollenhour (approved):
I approve since these themes are incredibly old (e.g. IE6 support) and in the off chance someone wants them they can just grab the files from the repo.

fballiano:
@luigifab was your frontend set to use the default? because the new installations have RWD by default if I didn't get everything wrong when I tested it.

luigifab (approved):
No, it was set to my dev website very long time ago.

kiatng (approved):

kkrieger85 (approved):

ADDISON74:
I see this PR as merged but I would like to clarify a potential issue that could appear. As you know OM is shipped with ```/app/design/frontend/default```. Into this folder there are the following directories:

- blank
- default
- iphone
- modern

Most of the themes on the market are requesting installations into this directory. If a file is missing from theme Magento fallback will solve the issue and will search in /frontend/default/default then in frontend/base.

If we remove /default/default directory are we creating any errors with fallback model?

Flyingmana:
Yes, these themes will possibly break then.

fballiano:
if they use themes so old for sure they're not using composer and so they
won't lose the directories and so won't have problems.
also, we can't think of having rhose ugly unmaintained themes forever,
we're talking about the 20.x branch here.

Il Dom 16 Mag 2021, 20:11 Daniel Fahlke ***@***.***> ha
scritto:

ADDISON74:
@Flyingmana: In this case we should remove only /blank, /modern and /iphone directories.

@flaliano: It is just a matter of fallback model in Magento. OM should keep backward compatibility related to default folder for old themes installed in /frontend/default directory. Please note that although OM wants to be a better Magento it is doomed to work with old products such as themes and extensions. What we do through OM is just to dress the code, to add facilities, but 99% of the people who are using the platform adapt it to their own needs using this "old stuff". 

An eloquent phrase says "If it ain't broke, don't fix it".

fballiano:
sorry but i don't agree, the default theme is ugly and has to go. backwards
compatibility up to a certain point (theres the 19.x branch for that).
otherwise we'd have to support php5.6

Il Dom 16 Mag 2021, 20:42 ADDISON ***@***.***> ha scritto:

tmotyl:
@ADDISON74 would a solution for you be to have a new package on openmage github with these old themes?
This would alow to easily install them by whoever still needs them.

ADDISON74:
@tmotyl: I think my original post was misunderstood. I did not mean to allow myself to continue using old themes from this directory ```/frontend/default``` which was delivered by Magento but to a completely different situation.

If I install a theme called Luna in the directory ```/frontend/default/luna``` then if a layout/template file is not in the theme directory then Magento looks for it in the ```/front/default``` directory on the same level, if it doesn't find it looks up one level in the ```/frontend/base```.

This PR deletes the default directory and I asked to evaluate if deleting it could affect this fallback. I did a simulation by renaming the directory and nothing happened. But I did not check if there were any errors in the logs and the page load time was longer.

Issue #1602, opened 2021-05-06
WSDL API does not work in PHP 8.0


Caprico85:
### Preconditions (*)
1. OpenMage 20.0.10
2. PHP 8.0

### Steps to reproduce (*)
1. Open http://example.com/index.php/api/v2_soap/index/?wsdl in browser

### Expected result (*)
1. URL can be opened

### Actual result (*)
The URL can't be opened. The error
```
Fatal error: Declaration of Mage_Api_Model_Wsdl_Config_Element::getChildren($source) must be compatible with SimpleXMLElement::getChildren() in [...]/app/code/core/Mage/Api/Model/Wsdl/Config/Element.php on line 179
```
is shown.

This works in PHP 7.4. It only fails since PHP 8.0. I'm not sure if this is caused because PHP 8.0 is more strict towards incompatible method signatures. Or if SimpleXMLElement has got a new method getChildren() in PHP 8.0.

Flyingmana:
Solved by #1613

Thank you a lot for finding this.

PR #1603, opened 2021-05-06
Fix getChildren must be compatible with SimpleXMLElement with PHP 8

luigifab:
### Description

This PR fix the following error with PHP 8 (#1602):
```
Fatal error: Declaration of Mage_Api_Model_Wsdl_Config_Element::getChildren($source)
 must be compatible with SimpleXMLElement::getChildren() in
 [...]/app/code/core/Mage/Api/Model/Wsdl/Config/Element.php on line 179
```

OpenMage 20.0.10 / PHP 8.0.5

### Manual testing scenarios

```php
$proxy = new SoapClient('https://.../api/v2_soap/?wsdl', ['user_agent' => 'Local/Test']);
$sessionId = $proxy->login('...', '...');
$products = $proxy->catalogProductList($sessionId, [], 1);
foreach ($products as $product)
	echo $product->product_id,' ',$product->name,"\n";
```

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Api/Model/Wsdl/Config/Element.php
+++ b/app/code/core/Mage/Api/Model/Wsdl/Config/Element.php
@@ -176,7 +176,7 @@ class Mage_Api_Model_Wsdl_Config_Element extends Varien_Simplexml_Element
      * @param Varien_Simplexml_Element $source
      * @return array
      */
-    public function getChildren($source)
+    public function getChildren($source = null)
     {
         $children = array();
         $namespaces = $source->getNamespaces(true);


fballiano (approved):

kiatng (approved):
More info on the error in [PHP doc](https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp). The mandatory parameter needs to be optional to have signature compatibility.

Flyingmana:
anyone found the documentation for the SimpleXMLElement::getChildren method?
Is this a new method?

jfksk:
@Flyingmana 
The documentation seems to be missing. According to [PHP src](https://github.com/php/php-src/commit/4730b06f1d026047c63980298d358e28e2183de6) iterator functionality was added to SimpleXMLElement in 8.0. The signature according to the [stub](https://github.com/php/php-src/blob/master/ext/simplexml/simplexml.stub.php) [(changes/added methods):](https://github.com/php/php-src/pull/5216/files#diff-d4e1b2d0453202a95ac12a7ecea6f1d1e5725f99f68de16b3ae4872b21e8a173) 

`SimpleXMLElement::getChildren() : ?SimpleXMLElement`

Flyingmana:
thank you. The return type could lead to major issues for us here.
Did someone already test if the return type will also cause a signature error, and if returning an array will cause issues?

kkrieger85 (approved):

colinmollenhour:
I can see how the this PR would fix the issue but making the parameter optional when it really is required is hard to stomach.. :)

I propose an alternative in #1613 which changes the method name so it doesn't conflict with the parent method. It doesn't use "$this" so I made it static as well.

Flyingmana:
closed in favor of #1613

Still, thank you a lot

PR #1604, opened 2021-05-07
Last login date in wrong in backend

fballiano:
In "adminhtml -> customers -> manage customers" if you click on a customer sheet and check the "last login" section, sometimes you get a weird situation like this:

Schermata 2021-05-07 alle 12 32 07

the "login date" for the store timezone is correct while the UTC one is not, that is because, in the code, the formatDate function expects a "date", not a "timestamp".

After the patch, this is the result:
Schermata 2021-05-07 alle 12 54 57

Insert "2021-05-07 07:12:18" as "login_at" column in the "log_customer" table, for the user you want to test.


ADDISON74:
I guess there is another commit that fixes this issue. Please check it here #1525.

fballiano:
🤦🏻‍♂️ i hope it would have been merged...

anyways, the usetimezone = false should be exactly utc, but anyways this doesn't make sense.

PR #1383, opened 2021-01-09
Remove latest occurrences of XmlConnect

luigifab:
I searched in all files XmlConnect, and removed the CSV and a comment.

PR #1616, opened 2021-05-15
Revert "cleanup - remove orphan js/jscolor/* from XmlConnect package (#1436)"

### LINKED ISSUE #1615: jscolor.js was removed by mistake from OpenMage
Yesterday in an OpenMage store I manage a content creator pointed out to me that the color picker hasn't worked for a while. This color picker is used by MageWorx Advanced Product Options.

Initially I analyzed the source code of the page and I did not find anything wrong, everything was unchanged. When I opened the console in the browser I saw 5 errors related to the color picker. Here is one of them:

```
Uncaught TypeError: Cannot read property 'showPicker' of undefined
    at Object.showColor (eval at <anonymous> (prototype.js:640), <anonymous>:574:27)
    at HTMLDivElement.onclick ((index):1)
```

Everything was as clear as possible a missing file javascript file. I checked the directory ```/js/jscolor``` and I did not find it in the file system. I looked for a previous version of OM 20.0.6 and took the directory from there. Everything worked as expected.

**Checking the merges this directory was removed by mistake based on this commit #1436.** 

Although at first glance this library was associated with XmlConnect and was considered an orphan one deleting it from OM creates issues in Magento extensions which are using the color picker. MageWorx is one of the developers who use it, more than likely there are others who use it.

One thing is obvious, we can't ask the developers to put this library in their extensions because all knew it existed in any Magento distribution. In addition, I am not sure that they can be contacted or interested to make changes.

**My proposal is as follows:**

1. This library should be brought back to OM yesterday if possible so as not to create issues for those who have made version upgrades after the date when the commit was merged.

2. An update of the library must be evaluated too. The version used in OM was 1.3.1 and now jscolor is version 2.4.5. For more information please check the developer page https://jscolor.com/

Flyingmana:
Revert "ref #1383, cleanup - remove orphan js/jscolor/* from XmlConnect package (#1436)

reverting, as jscolor is used by 3th party modules, so removing it is a big BC breaks.

Reverts #1436
Fixes #1615
This reverts commit e4c2c4c9


ADDISON74:
Thank you @Flyingmana for your quick reaction.

Do we need the demo.html file in this directory?

Flyingmana:
probably not needed, but just reverting was the fastest

colinmollenhour (approved):

Issue #1614, opened 2021-05-14
Readme file (All Contributors) is broken


jandolejs1999:
### Preconditions (*)

1. 0ca69ceec78f30c573e043d5fac2dcb3c112cd81

### Steps to reproduce (*)

1. Open [repository](https://github.com/OpenMage/magento-lts)
2. See first line after files (README.md file)

### Expected result (*)

1. ![image](https://user-images.githubusercontent.com/16374521/118249881-27c93280-b4a6-11eb-8808-24255c25cba4.png)

### Actual result (*)

1. ![image](https://user-images.githubusercontent.com/16374521/118249900-30ba0400-b4a6-11eb-8b2e-79b0b987792b.png)

julienloizelet:
It reminds me of something :) : https://github.com/OpenMage/magento-lts/pull/1088

jandolejs1999:
Well, i think its auto-generated, maybe thats why there is note that it should not be edited

fballiano:
shouldn't it just be something like

`<img src="https://img.shields.io/badge/all_contributors-101-orange.svg?style=flat-square" alt="All Contributors">`

I've always seen it broken and I thought there was something wrong with my browser or something heheheh

fballiano:
I think that also the .all-contributorsrc file is not updated automatically

fballiano:
I've run allcontributors on the REPO and it shows:

```
Missing contributors in .all-contributorsrc:
    daim2k5, fballiano, jouriy, digitalpianism, ADDISON74, aredridel, drwilliams, hhirsch, kdckrs, sicet7, justinbeaty, theroch, AlterWeb, Caprico85, davidwindell, drashmk, lamskoy, ferdiusa, himanshu007-creator, idziakjakub, JosephMaxwell, joshua-bn, KBortnick, elidrissidev, MrGekko, nikkuexe, psdevhh, PatrickCronin, petrsvamberg, rafaelstz, mageconsult, sunel, ktomk, ToonSpin, wexo-team, wilfriedwolf, akrzemianowski, andthink, eetzen, lemundo-team, mdlonline
Unknown contributors found in .all-contributorsrc:
loekvangool, samsteele, nintenic
```

it seems there's no way of knowing who contributed to what (at the contributors table in the README is not well aligned/formatted I suggest we add all the missing contributors and remove the "code/doc" specification, so that the table will be more beautiful to see and everything will be much easier to update in the future.

Flyingmana:
[...]

PR #1627, opened 2021-05-20
Fix README.md contributors badge

mehdichaouch:
Fix README.md contributors badge, causing the display of badge url code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 <p align="center">
+<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
 <a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-101-orange.svg?style=flat-square" alt="All Contributors"></a>
+<!-- ALL-CONTRIBUTORS-BADGE:END -->
 <a href="https://travis-ci.org/openmage/magento-lts"><img src="https://travis-ci.org/openmage/magento-lts.svg" alt="Build Status"></a>
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/d/total.svg" alt="Total Downloads"></a>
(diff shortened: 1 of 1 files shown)


colinmollenhour:
I don't know much about this badge but I think the "101" part of the url is basically meant to be dynamically replaced by the All Contributors bot so this PR would break that. Granted, it is already broken! 😄 #1614

mehdichaouch:
> I don't know much about this badge but I think the "101" part of the url is basically meant to be dynamically replaced by the All Contributors bot so this PR would break that. Granted, it is already broken! 😄 #1614

Oh I understand, by fixing this I maybe brake the bot in charge of changing this dynamically.

mehdichaouch:
**_DON'T MERGE MY PULL REQUEST._**

I read more about **_all-contributors-bot config_**, I think [installation](https://allcontributors.org/docs/en/bot/installation) have to be checked and workflow to [add contributor](https://allcontributors.org/docs/en/bot/usage) was not respected.

Flyingmana:
@mehdichaouch yes there is somewhere a related ticket. We did not "install" the all contributors bot, because it needs write access to the repository, which could be used in a possible supply-chain-attack

mehdichaouch:
So what about the badge, do you have a solution do fix it?
Should I close my PR or not?

colinmollenhour:
> Means, it needs to be manually run and used to create PullRequests

Is this something you could document somewhere very briefly like an internal wiki?

fballiano:
what are we doing with this? it's really ugly to have a broken README

Flyingmana:
Iam ok with merging it for now to fix the obvious error and take care of the automatism separately

Flyingmana (approved):

Flyingmana (approved):

sreichel:
@all-contributors add @mehdichaouch doc

sreichel:
@all-contributors add @mehdichaouch doc

PR #1629, opened 2021-05-20
Removed non-nonsensical local var

woutersamaey:
### Description (*)
Removed a non-nonsensical local var


Flyingmana:
This may be used in classes extending from this one. Also property access is in general preferred compared static calls

woutersamaey:
There are no classes extending this one (but you may have custom modules, that's true).
The use of a local property isn't something I was aware of. It's not used consistently by Magento.

Guess we can drop the PR. It's not very helpful.

PR #1630, opened 2021-05-20
Replaced isAvailable() with isSalable() so events are considered

woutersamaey:
### Description (*)
Replaced isAvailable() with isSalable() so event are considered.

I have a store that has more complex logic for determining "salability". This block does not consider event-driven salability.

I have this code running in production for 2 years, now I'm contibuting it.

### DIFF
--- a/app/code/core/Mage/ProductAlert/Block/Product/View.php
+++ b/app/code/core/Mage/ProductAlert/Block/Product/View.php
@@ -49,7 +49,7 @@ class Mage_ProductAlert_Block_Product_View extends Mage_Core_Block_Template
      */
     public function prepareStockAlertData()
     {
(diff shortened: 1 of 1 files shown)


colinmollenhour:
Thanks @woutersamaey ! Is there any scenario you can think of where this would not be the behavior expected by someone? Is there any difference on a vanilla installation?

woutersamaey:
Not really. This change should be harmless, but it's essential if you're changing salability using events.
Changing salability using events works well, except in this single instance.

colinmollenhour:
I think I would agree with the change, but perhaps it belongs in the 20.0 branch.

woutersamaey:
Moved to 20 branch. Can we get this merged?

Flyingmana (approved):

Sekiphp (approved):
I investigating the diferrence between `isAvailable` and `isSalable` and It looks, that this should work properly.

PR #1632, opened 2021-05-21
 Support saving static attributes

woutersamaey:
### Description (*)
Magento's resource models only allow saving of EAV attributes that are non-static. However, static attributes make sense in a lot of ways to increase performance + to have an actual unique key on DB level.

My client has a huge catalog and is moving attributes around to increase storage efficiency and speed.

All unique + global attributes benefit from becoming static + receiving a unique DB key.

Magento's own uniqueness is a joke and not really unique.

This PR allows static attributes to be saved as you would expect any attribute to save through the resource layer.
Existing logic is not effected, so there should be no danger.

Upon merge, please squash these commits.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
@@ -238,24 +238,38 @@ abstract class Mage_Catalog_Model_Resource_Abstract extends Mage_Eav_Model_Entit
             ));
         }
 
-        $data = new Varien_Object(array(
-            'entity_type_id'    => $attribute->getEntityTypeId(),
-            'attribute_id'      => $attribute->getAttributeId(),
-            'store_id'          => $storeId,
-            'entity_id'         => $object->getEntityId(),
-            'value'             => $this->_prepareValueForSave($value, $attribute)
-        ));
+        $isStatic = $attribute->getBackendType() === 'static';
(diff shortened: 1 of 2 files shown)


kiatng:
I implemented custom eav object with many static attributes and added a custom static attribute to table `customer_entity`, when models are saved, the static attributes are updated in the entity tables in both cases. This is done without touching the core. I have not done anything with product, but I would imagine it should work the same. The static values are prepares as an array `$entityRow` in the function `_collectSaveData()`:
https://github.com/OpenMage/magento-lts/blob/9b04ad2f30a7fcb092052597787cd6e1d4ab4f8d/app/code/core/Mage/Eav/Model/Entity/Abstract.php#L1202-L1205

and then saved in function `_processSaveData()`:
https://github.com/OpenMage/magento-lts/blob/9b04ad2f30a7fcb092052597787cd6e1d4ab4f8d/app/code/core/Mage/Eav/Model/Entity/Abstract.php#L1334-L1351

If your static values aren't saved, `$entityRow` may be empty, you can take a look at function `_collectSaveData()` to figure out why.

fballiano:
I'm closing this because:
- from @kiatng's comment it seems there's no need for the PR (if I'm reading it correctly) and there was no feedback from the author
- there are conflicts on both of the files of the PR

@woutersamaey please feel free to reopen it if you want to continue it

Issue #1634, opened 2021-05-23
Product not visible on front after updating stock status to "In Stock" through SOAP API V2 and vice-versa


elidrissidev:
### Preconditions (*)
Reproduced in OpenMage 1.9.4.5, but should be reproducible in many other versions below and above that. Additionally, the `Display Out of Stock Products` option should be set to `No` under `System -> Configuration -> Catalog -> Inventory -> Stock Options`.

### Steps to reproduce (*)
1. Have any product that is visible in the catalog and is Out of Stock.
2. Update the stock status of that product to "In Stock" with the SOAP API V2. It will not show up.
3. Save the product in the admin panel (even with no changes). It will now be visible in front.
4. Update the stock status of the same product to "Out of Stock" with the SOAP API V2. It will still be visible in front. 

### Expected result (*)
1. After updating the stock status to "In Stock"/"Out of Stock" from the SOAP API V2, the product should be visible/not visible respectively.

### Actual result (*)
1. When the stock status is updated from "Out of Stock" to "In Stock" from the SOAP API V2, the product will not be visible in front.
2. To make the product visible again, you need to save it from the admin panel (with, or without changes).
3. When the stock status is updated from "In Stock" to "Out of Stock" from the SOAP API V2, the product will still be visible in front.

PR #1635, opened 2021-05-23
Fixed stock status update from SOAP API V2

elidrissidev:
### Description (*)
This change will fix an issue where updating the stock status from "Out of Stock" to "In Stock" from the SOAP API V2, with `Display Out of Stock Products` set to `No`, does not make the product show in front and vice-versa.
This is because in the method `Mage_CatalogInventory_Model_Stock_Item_Api_V2::update` the stock item is obtained from the call to `$product->getStockItem()` and so `$stockItem->getProduct()` returns the associated product. Because the stock indexer is expecting `$stockItem->getProduct()` to be `null` (see snippet below) this leads the product prices to not be indexed and therefore causes issues with product visibility in front.

https://github.com/OpenMage/magento-lts/blob/d078657d88e7758f4b6bd1b7046751197f063bc1/app/code/core/Mage/CatalogInventory/Model/Indexer/Stock.php#L272

### Fixed Issues
1. Fixes OpenMage/magento-lts#1634

Can be found in the linked issue.

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Stock/Item/Api/V2.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Stock/Item/Api/V2.php
@@ -35,14 +35,15 @@ class Mage_CatalogInventory_Model_Stock_Item_Api_V2 extends Mage_CatalogInventor
         $idBySku = $product->getIdBySku($productId);
         $productId = $idBySku ? $idBySku : $productId;
 
-        $product->setStoreId($this->_getStoreId())
-            ->load($productId);
+        /** @var Mage_CatalogInventory_Model_Stock_Item $stockItem */
+        $stockItem = Mage::getModel('cataloginventory/stock_item')
+            ->setStoreId($this->_getStoreId())
+            ->loadByProduct($productId);
 
-        if (!$product->getId()) {
+        if (!$stockItem->getId()) {
             $this->_fault('not_exists');
         }
 
-        $stockItem = $product->getStockItem();
         $stockData = array_replace($stockItem->getData(), (array)$data);
         $stockItem->setData($stockData);
 


Flyingmana:
I understand the issue, but this change probably causes a big drop in performance for this endpoint.

also the save should trigger the afterCommitCallback, which should trigger a reindex for this entity. Does this not work?

elidrissidev:
@Flyingmana `afterCommitCallback` on the stock item does not reindex the product, which is what's causing this issue AFAIK.

elidrissidev:
This is where the product prices get indexed. It is only getting triggered by the `catalog_product_save_commit_after` event

https://github.com/OpenMage/magento-lts/blob/d078657d88e7758f4b6bd1b7046751197f063bc1/app/code/core/Mage/CatalogRule/Model/Rule.php#L352-L356

Flyingmana:
can you check if this `_registerStockItemSaveEvent` call is triggered?
It only uses the `logEvent()` call, so the other indices need to be processed in a separate process I think. Can you check if this entry is created in `index_event`? And if it indirectly also triggers the price index?

https://github.com/OpenMage/magento-lts/blob/d078657d88e7758f4b6bd1b7046751197f063bc1/app/code/core/Mage/CatalogInventory/Model/Indexer/Stock.php#L241-L284

elidrissidev:
@Flyingmana  I reverted my changes and checked, that method is called but it's not getting inside the `if` because `$object->getProduct()` is truthy. I'm gonna keep checking why.

https://github.com/OpenMage/magento-lts/blob/d078657d88e7758f4b6bd1b7046751197f063bc1/app/code/core/Mage/CatalogInventory/Model/Indexer/Stock.php#L272

elidrissidev:
I think I've come to a closed road here. So If anyone is willing to give this issue another try, feel free.

elidrissidev:
@Flyingmana I updated the PR and the description with a new fix. Kindly check and review so we get this fixed asap.

elidrissidev:
Can we get this reviewed and merged if all is good, please @Flyingmana

fballiano (dismissed):

fballiano:
I think this one could be merged already

fballiano:
@elidrissidev could you rebase it to "main", I would try to re-get some approvals and merge it

fballiano (approved):

ADDISON74 (approved):

PR #1643, opened 2021-05-26
Grid range filter replaced span with placeholder

Sekiphp:
### Description (*)
Design improvment for adminhtml grid - replaced `span` to `placeholder`.

#### Before changes
image

#### After changes
image

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Date.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Date.php
@@ -50,16 +50,16 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Filter_Date
         $htmlId = $this->_getHtmlId() . microtime(true);
         $format = $this->getLocale()->getDateStrFormat(Mage_Core_Model_Locale::FORMAT_TYPE_SHORT);
         $html = '<div class="range"><div class="range-line date">'
-            . '<span class="label">' . Mage::helper('adminhtml')->__('From').':</span>'
-            . '<input type="text" name="'.$this->_getHtmlName().'[from]" id="'.$htmlId.'_from"'
+            . '<input type="text" name="' . $this->_getHtmlName() . '[from]" id="' . $htmlId . '_from"'
+                . ' placeholder="' . $this->quoteEscape(Mage::helper('adminhtml')->__('From')) . '" aria-label="' . $this->quoteEscape(Mage::helper('adminhtml')->__('From')) . '"'
                 . ' value="'.$this->getEscapedValue('from').'" class="input-text no-changes"/>'
             . '<img src="' . Mage::getDesign()->getSkinUrl('images/grid-cal.gif') . '" alt="" class="v-middle"'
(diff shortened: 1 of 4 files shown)


LukasCesal (dismissed):

FreeWall (approved):

midlan (dismissed):

kiatng:
Good idea! But in the original theme v20 (client still prefer original theme), this is the after-look that I have:
image

It'll look better if the date-range can be made into placeholder as well and the width of the number-range can be widen to fill the gap.

ADDISON74:
Even I appreciate the efforts for the OM Backend theme in all my stores the content editors want to use the Magento default theme. As I see I am not the only one who confronts with such of request.

**I am sorry to say but I do not agree this change.** If this will be merged all Magento Backend sections must be changed according.  At least for Orders section not only G.T. (Base) and G.T. (Purchased) columns should be changed but also Purchased On. Removing the text in front of the input text boxes creates a new issue the blue empty space which is not good looking at all. Unfortunately it is a change in the core file (Range.php) not in a template which could be related to the new Backend theme.

orders

midlan:
I agree that the blue gap is ugly.

Flyingmana (changes requested):
While I appreciate improvements on the design, this fast becomes a controversial topic. And Iam no expert in this area ( and also have no overview, who among us is experienced in this topic)
By moving the label into the placeholder you may save space, but it also hides the label if the field is filled.
Also Iam not sure how this affects accessibility and screen readers.

Therefore I will block this for a while, so we can find a proper strategy first how to handle such changes/cases.

midlan:
About the screen readers I contacted my friend who is optimizing pages for accessibility. He told me that `aria-label` is needed in this case to help readers properly read it.

@Flyingmana I read that article: https://www.nngroup.com/articles/form-design-placeholders/
I would say that none of them apply to this case. In this case it is only two fields, you won't forgot which one is which. The other filter fields in Magento grids have no placeholder or label at all and everybody knows how to use them.

Sekiphp:
I found this document from w3c: https://www.w3.org/TR/html-aria/

ADDISON74:
I repeat the observation that if this change is merged, it must be made at the level of the entire interface, not only in one place. Also not without prior testing before by those who can still do it and leaving an opinion here what benefits brings. 

As for me, I will retain this change and if it is merged I will remove it in the new OM versions to avoid discussions with those who have been using the default Magento interface for a long time. Maybe if I was the only user I would have been more flexible about this change.

OT - there are a lot of issues still in Magento core reported in Issue section we should focus on.

Sekiphp:
I added same behaviour to all filter classes which have more than 1 field. I used `placeholder` & `aria-label`

midlan:
@Sekiphp What about the blank space that @kiatng mentioned above?

sreichel:
_This branch has conflicts that must be resolved_

ADDISON74:
I would close this PR. Using placeholders can reduce the width of the column, but input boxes with that text will look very ugly. We have already done the modification to keep outside From, To, In and I think it looks very good.

sreichel:
Just asked to resolve conflicts. 

Having placeholders and aria-label is not wrong.

Issue #1666, opened 2021-05-30
Remove orphan directory and code


kiatng:
* Directory `includes` in the root is used by the compiler, which was removed in PR #534.
* Code that reference it should be removed

https://github.com/OpenMage/magento-lts/blob/4d2ce083a9b7a620f085c89183789d890b7eb5ab/index.php.sample#L34-L40

* Line 492
https://github.com/OpenMage/magento-lts/blob/4d2ce083a9b7a620f085c89183789d890b7eb5ab/app/code/core/Mage/Core/etc/config.xml#L486-L498

* Line 492 above also can be removed, since downloader was removed in PR #952.

ADDISON74:
You may close this one if there is already a PR for it.

PR #1667, opened 2021-05-30
Removed orphan directory and code of compiler and downloder.

kiatng:
### Related Pull Requests

PR #534 and PR #952.

1. Fixes issue #1666.

1. ...
2. ...


fballiano (dismissed):

fballiano:
I love these kinds of PRs!

ADDISON74:
I strongly recommend not removing ```/includes``` directory. Inside of it there is an important file named **config.php** that can help you to achieve some unknown features. For example, if you are using this combination HAProxy (for SSL Offloading) + Varnish + Apache in Magento Backend -> Orders page you will get a chain of IP's addresses, more precisely 3 IP's because the offloading part of Magento was never pushed to a final shape. Recently Colin Mollenhour managed to offer a correct solution for this. In order to get only one IP address listed in Orders page I am using config.php. Here is an example from production:

```
if(isset($_SERVER['HTTP_X_REAL_IP'])) {
    $_SERVER['REMOTE_ADDR'] = $_SERVER['HTTP_X_REAL_IP'];
    $_SERVER['HTTP_X_FORWARDED_FOR'] = '';
}
```
HTTP_X_REAL_IP comes from .htaccess where I am processing the IP's to get the remote one.

fballiano:
> I strongly recommend not removing `/includes` directory. Inside of it there is an important file named **config.php** that can help you to achieve some unknown features. For example, if you are using this combination HAProxy (for SSL Offloading) + Varnish + Apache in Magento Backend -> Orders page you will get a chain of IP's addresses, more precisely 3 IP's because the offloading part of Magento was never pushed to a final shape. Recently Colin Mollenhour managed to offer a correct solution for this. In order to get only one IP address listed in Orders page I am using config.php. Here is an example from production:
> 
> ```
> if(isset($_SERVER['HTTP_X_REAL_IP'])) {
>     $_SERVER['REMOTE_ADDR'] = $_SERVER['HTTP_X_REAL_IP'];
>     $_SERVER['HTTP_X_FORWARDED_FOR'] = '';
> }
> ```

it's interesting but in my opinion it seems the wrong place to put it (although I totally see your point). If IP chains are necessary to handle, then let's put it in the index.php. includes/config.php is 99% unused (while causing I/O) and had a totally different purpose.

ADDISON74:
I have nothing against as long as what I said above can be obtained in another way. Magento needs to be able to deal with such server configurations that when it was born in 2008 were not possible or accessible to anyone. I will even propose in the Discussions section on this topic.

colinmollenhour:
I think the `include` for `includes/config.php` was already removed from `index.php` but it was missed from `index.php.sample` somehow.. Should this sample file not just be deleted as well? So already to use `config.php` you would have to have a modified `index.php`. But I agree this file is not the right place to add custom code so as a project we should consider it dead code that is safe to remove.

One minor problem with this PR is if existing installations have important files in `includes` directory (no idea why this would be the case) these will be potentially exposed by Apache because the `.htaccess` file is removed.. So maybe the `.htaccess` file should remain in the 1.9.4.x branch and only be removed in 20.0?

kiatng:
As pointed out by @ADDISON74, this PR may break BC. I tried to edit the base branch to 20.0 but that didn't work as it brought along some 35 other commits. Not sure how to do this, do I close this PR and make a new one for branch 20?

Flyingmana:
you can wait till next release, then all of these commits are already part of 20.0, so changing the branch will not bring them up anymore

luigifab (approved):

fballiano (approved):

PR #1698, opened 2021-06-24
Do not load product when it is already loaded

luigifab:
### Description

When `review/form` block is added to the product page, without this PR, the product is loaded a second time.

### Contribution checklist


Flyingmana (approved):

fballiano (approved):

luigifab:
I noticed that `($product->getId() == $this->getRequest()->getParam('id'))` doesn't work with [Turpentine+Varnish](https://github.com/nexcess/magento-turpentine), because esi URL doesn't contain the id parameter. Simple solution: remove it.

PR #1254, opened 2020-10-09
Fix undefined offset on redis session

spinsch:
### Description
In most cases it will never be a problem, but in test environments with the highest error level it will not work.
list() wants numeric arrays

### Manual testing scenarios
1. Visit a page

### Questions or comments
**Prerequisite**
- Run php with error_reporting = E_ALL
- Configure redis as session backend

**ENV**
- PHP 7.3.22
- Redis 6.05

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1705, opened 2021-06-26
Cm_RedisSession: Fix Notice: Undefined offset: 0

justinbeaty:
### Description (*)
After upgrading my local environment to v20.0.12 and setting `<global><session_save>` to `redis`, I could not log in to admin anymore, and would just get invalid form key. On my development box, I have developer mode enabled, and I get this error:
```
Notice: Undefined offset: 0  in /home/www-data/magento-root/htdocs/app/code/community/Cm/RedisSession/Model/Session.php on line 349
```

Previously, I was using this in my composer.json: `"colinmollenhour/magento-redis-session": "^2.3"`, but I removed that and am using the version in this repo.

I am not sure if it is related to the upgrade in particular, or maybe there has always been this problem. I am on PHP 7.4.3. The problem is that hMGet returns an assoc array like `array( 'data' => '', 'writes' => '')` but list is expecting numerical keys.

@colinmollenhour Tagging you just for visibility.

### Related Pull Requests
#1513

**Edit:** This is the same PR: #1254 -- probably was not reproducible for everyone because it requires `Cm_Cache_Backend_Redis` to be installed, see next comment.

1. Upgrade to v20.0.12
2. Change local.xml <global><session_save> to redis
3. Configure local.xml redis_session
4. Attempt to login, receive invalid form key
5. Enable developer mode
6. Attempt to login, will receive the notice above

### DIFF
--- a/app/code/community/Cm/RedisSession/Model/Session.php
+++ b/app/code/community/Cm/RedisSession/Model/Session.php
@@ -339,7 +339,7 @@ class Cm_RedisSession_Model_Session extends Mage_Core_Model_Mysql4_Session
         if ($this->_logLevel >= Zend_Log::DEBUG) {
             $timeStart = microtime(true);
         }
-        list($sessionData, $sessionWrites) = $this->_redis->hMGet($sessionId, array('data','writes'));
+        list($sessionData, $sessionWrites) = array_values($this->_redis->hMGet($sessionId, array('data','writes')));
         Varien_Profiler::stop(__METHOD__);
         if ($this->_logLevel >= Zend_Log::DEBUG) {
             $this->_log(sprintf("Data read for ID %s in %.5f seconds", $sessionId, (microtime(true) - $timeStart)));


justinbeaty:
I think I figured out what is going on here, it looks like it is because also use `Cm_Cache_Backend_Redis`.

In my composer.json, I put `"colinmollenhour/cache-backend-redis": "^1.14",`, which requires `"colinmollenhour/credis": "*"`.

However the version in my vendor folder is different than the one in this repo. The one in this repo has the following at line 964:

```
case 'hmget':
    $response = array_values($response);
    break;
```

Thus, it is already returning `array_values`. So this PR won't hurt anything since running an array through `array_values` is the same result, but is there anything else I should be aware of when using both this repo's Cm_RedisSession and Cm_Cache_Backend_Redis?

Flyingmana:
Interesting.

can you create a separate issue about the conflict of both versions and how the suggested way of handling it is?
Cant answer this spontanously, but its definitely something we should handle. And be it by making sure its not installed via composer.json

Flyingmana (approved):

justinbeaty:
Hi @Flyingmana thanks for merging the other PR. I actually had made this PR before finding #1254, so I sort of laughed when I found the old one and it was the same exact fix. I'll make a new issue to discuss the conflict.

Closing this PR as resolved.

PR #1714, opened 2021-07-01
Reload admin ACL

luigifab:
### Description

This PR allow to reload admin acl without logout/login:
- for associated users when a role is saved
- when admin user is saved

OpenMage 20.0.13 / PHP 7.4.6 + 8.0.18.

### Manual testing scenarios

* Create a new restricted role (for example, only with access to _Customers_).
* Create a new admin user, use the previous created role.
* **Open a private window** and login into backend with the previous created user.
* a/ Update the role of the new user: add access to _Sales_.
* Press F5 on the private window, and go to _Sales / Orders_.
* b/ Update the role of the new user: remove access to _Sales_.
* Press F5 on the private window.
* c/ Go to the Administrator role: associate the new user.
* Press F5 on the private window.
* d/ Go to the restricted role: associate the new user.
* Press F5 on the private window.
* e/ Update the new user: set role to Administrators.
* Press F5 on the private window.

You can add the following line in app/code/core/Mage/Admin/Model/Session.php on line 221 to add a success message when ACL is reloaded: `Mage::getSingleton('adminhtml/session')->addSuccess('ACL reloaded.');`

```php
    public function refreshAcl($user = null)
    {
        if (is_null($user)) {
            $user = $this->getUser();
        }
        if (!$user) {
            return $this;
        }
        if (!$this->getAcl() || $user->getReloadAclFlag()) {
            $this->setAcl(Mage::getResourceModel('admin/acl')->loadAcl());
            Mage::getSingleton('adminhtml/session')->addSuccess('ACL reloaded.');
        }
        if ($user->getReloadAclFlag()) {
            $user->getResource()->saveReloadAclFlag($user, 0);
        }
        return $this;
    }
```

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Permissions/RoleController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Permissions/RoleController.php
@@ -83,7 +83,6 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
 
     /**
      * Show grid with roles existing in systems
-     *
      */
     public function indexAction()
     {
@@ -92,13 +91,11 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
              ->_title($this->__('Roles'));
 
         $this->_initAction();
-
         $this->renderLayout();
     }
 
     /**
      * Action for ajax request from grid
-     *
      */
     public function roleGridAction()
     {
@@ -108,7 +105,6 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
 
     /**
      * Edit role action
-     *
      */
     public function editRoleAction()
     {
@@ -143,7 +139,6 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
 
     /**
      * Remove role action
-     *
      */
     public function deleteAction()
     {
@@ -177,9 +172,9 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
 
         try {
             $role->delete();
-
             Mage::getSingleton('adminhtml/session')->addSuccess($this->__('The role has been deleted.'));
         } catch (Exception $e) {
+            Mage::logException($e);
             Mage::getSingleton('adminhtml/session')->addError($this->__('An error occurred while deleting this role.'));
         }
 
@@ -188,7 +183,6 @@ class Mage_Adminhtml_Permissions_RoleController extends Mage_Adminhtml_Controlle
 
     /**
      * Role form submit action to save or create new role
-     *
      */
     public function saveRoleAction()
     {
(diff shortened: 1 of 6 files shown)


colinmollenhour:
What is the benefit of this? The user's role cannot be changed on that page so the ACL shouldn't need to be updated. Also the response will just be sent next and nothing else until the next request.

colinmollenhour:
Ahh, I see. The stock ACL reload mechanism is rather broken as you noticed.. For one it doesn't make sense to store the ACL in the session, it should be stored in cache. Also it *should* be possible for one ACL to be cached for all users and any time any role is updated the cache can be invalidated so the next user's request will refresh the cache for all users. That would solve this problem as well as some others and be faster and less demand on the session backend.

luigifab:
I found this solution on stackexchange...
Now I checked the login method, and it store ACL data in session:

https://github.com/OpenMage/magento-lts/blob/f71321b37a782bd167ef56de0c6e18e67645685d/app/code/core/Mage/Admin/Model/Session.php#L179

I searched and I found `Mage_Admin_Model_Session::refreshAcl` called from `controller_action_predispatch` event.
So I can `$user->getResource()->saveReloadAclFlag($user, 1);` for all users associated to a role when it is updated.

woutersamaey:
@colinmollenhour This would also help cases where you have a dynamic menu in Admin (i.e. we have a report section that is auto-populated with more SQL reports, depending on some settings and the user).
ACK from me.

app/code/core/Mage/Adminhtml/controllers/Permissions/RoleController.php
```diff
@@ -232,30 +225,35 @@ public function saveRoleAction()
...
 
             foreach ($roleUsers as $nRuid) {
                 $this->_addUserToRole($nRuid, $role->getId());
             }
 
-            $rid = $role->getId();
+            $users = Mage::getResourceModel('admin/user_collection')
+                ->addFieldToFilter('user_id', ['in' => Mage::getModel('admin/roles')->setId($role->getId())->getRoleUsers()]);
+
+            foreach ($users as $user) {
```
Flyingmana, line 239: in \Mage_Admin_Model_Resource_Roles::_updateRoleUsersAcl there is a more performant way to update the flag for a big list of users.
This may work better for Shops which may have a lot of Admin Users.


https://github.com/OpenMage/magento-lts/blob/f71321b37a782bd167ef56de0c6e18e67645685d/app/code/core/Mage/Admin/Model/Resource/Roles.php#L166-L179

not sure if its better to open up this method, or to copy the code into a private method here. Probably the Copy for now.

Flyingmana (changes requested):

app/code/core/Mage/Adminhtml/controllers/Permissions/RoleController.php
luigifab: You are so right, moreover, my `foreach` is so useless, because the previous `$role->save()` run `_updateRoleUsersAcl`. I updating and testing again.

luigifab:
Tested again, still works. I'm surprised that "a/" is already working...

fballiano (dismissed):

fballiano:
I think this PR is a great addition, having to logout/login it's simply not a very nice thing to ask users.

colinmollenhour (approved):
All good changes. 👍

fballiano (approved):

PR #1721, opened 2021-07-06
Update getNextAutoincrement to not use showTableStatus

jonashrem:
### Description (*)
update getNextAutoincrement to not use showTableStatus for performants reasons.

### Related Pull Requests
https://github.com/magento/magento2/pull/33433
#1712

/

test import/export

### Questions or comments
/


colinmollenhour:
Looks great, what is the oldest version of MySQL where this code will work?

colinmollenhour (approved):

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -61,12 +61,19 @@ public function getMaxDataSize()
...
-        if (empty($entityStatus['Auto_increment'])) {
-            Mage::throwException(Mage::helper('importexport')->__('Cannot get autoincrement value'));
+        $sql = sprintf(
+            'SELECT AUTO_INCREMENT FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = %s AND TABLE_SCHEMA = DATABASE()',
+            $this->quote($tableName)
+        );
+        $entityStatus = $connection->fetchRow($sql);
+        if (empty($entityStatus['AUTO_INCREMENT']))
+        {
+            throw new \Magento\Framework\Exception\LocalizedException(__('Cannot get autoincrement value'));
```
fballiano, line 73:
```suggestion
            Mage::throwException(Mage::helper('importexport')->__('Cannot get autoincrement value'));
```

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -61,12 +61,19 @@ public function getMaxDataSize()
...
+        );
+        $entityStatus = $connection->fetchRow($sql);
+        if (empty($entityStatus['AUTO_INCREMENT']))
+        {
             Mage::throwException(Mage::helper('importexport')->__('Cannot get autoincrement value'));
         }
-        return $entityStatus['Auto_increment'];
+
+        return $entityStatus['AUTO_INCREMENT'];
+
```
fballiano, line 77:
```suggestion
```

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -61,12 +61,18 @@ public function getMaxDataSize()
...
+        $connection = $this->_getReadAdapter();
 
-        if (empty($entityStatus['Auto_increment'])) {
+        $sql = sprintf(
+            'SELECT AUTO_INCREMENT FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = %s AND TABLE_SCHEMA = DATABASE()',
+            $this->quote($tableName)
+        );
+        $entityStatus = $connection->fetchRow($sql);
+        if (empty($entityStatus['AUTO_INCREMENT']))
+        {
```
fballiano, line 72:
```suggestion
        if (empty($entityStatus['AUTO_INCREMENT'])) {
```

fballiano:
was this code tested on OM?

colinmollenhour:
I confirmed that the "SELECT AUTO_INCREMENT FROM INFORMATION_SCHEMA.TABLES" query works on MySQL 5.6 and up and MariaDb 10.3 and up. (probably some earlier versions, these are just the oldest I checked)

I've not tested an import in Magento, @jonashrem did you test an import with the latest version of this PR?

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -54,12 +54,17 @@ public function getMaxDataSize()
...
     public function getNextAutoincrement($tableName)
     {
-        $adapter = $this->_getReadAdapter();
-        $entityStatus = $adapter->showTableStatus($tableName);
+        $connection = $this->_getReadAdapter();
 
-        if (empty($entityStatus['Auto_increment'])) {
+        $sql = sprintf(
+            'SELECT AUTO_INCREMENT FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_NAME = %s AND TABLE_SCHEMA = DATABASE()',
+            $this->quote($tableName)
```
colinmollenhour, line 61: Should this be like so?
```suggestion
            $connection->quote($tableName)
```

colinmollenhour (changes requested):

fballiano:
I've rebased this and tested it briefly creating new customer/addresses/products, the autoincrements work fine.

I've no idea if this is really a performance improvement or if it could create a problem.

But now it's ready to be merged, or it should be closed.

jonashrem:
Thanks

colinmollenhour (approved):
Approve, but I didn't perform a end-to-end test.

@jonashrem Please confirm this has been tested in it's current state. Thanks!

empiricompany (approved):

jonashrem:
> 

sorry this got under.

I did not test the current variant with up to date OpenMage.

Do you have some specific test szenario in mind?

fballiano:
I also will test this out as soon as possible

kiatng (approved):
Tested. It is faster by about 50% to sub mini-sec from 1.5ms.

```php
Mage::getResourceHelper('importexport')->getNextAutoincrement('sales_flat_order_item');
```

fballiano:
is it possible that the result of this call gets cached somewhere?

on my DB I've an autoincrement value of 65:
Screenshot 2023-09-13 alle 15 03 34

but when I run this:
Screenshot 2023-09-13 alle 15 03 57

I get 63:
Screenshot 2023-09-13 alle 15 03 54

colinmollenhour:
> is it possible that the result of this call gets cached somewhere?

Very odd, I don't see how that is possible.. Are you 100% certain the two tests were run against the same database?

elidrissidev:
> is it possible that the result of this call gets cached somewhere?

I encountered this issue the first time I tested this PR but forgot to post it. It seems `performance_schema` is not guaranteed to contain the most up-to-date value, sometimes `ANALYZE TABLE` is necessary to get accurate results. Maybe just a simple `MAX() + 1` query would do the job here.

fballiano:
@colinmollenhour I'm sure it is the same database

@elidrissidev AFAIK `MAX+1` is not safe in some cases, I don't remember exactly on top of my head, zf1 autoincrement function uses the "last insert id", would that be better at that point? or just leave everything is it is now?

colinmollenhour:
Ahh, I still don't see how essentially the same query via the console and via PHP can have different results, but it appears the information schema data is indeed cached within MySQL for an entire 24 hours by default. This can be controlled by [information_schema_stats_expiry](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_information_schema_stats_expiry). This can be set as a session variable just before the query and yield correct results which isn't a terrible approach, but not sure if that is really an improvement in performance at that point.

Perhaps your ID is setting this session variable behind the scenes and that is why the results are different?

https://stackoverflow.com/a/65996809/187780

The `MAX(id)` seems to be just as fast. It might not be accurate if there were deletions or if there were simultaneous transactions that incremented the auto increment but have not been committed yet or were rolled back (I'm not 100% sure on this but somewhat certain). However, that may not be an issue or even a regression - is this method going to be used in a multi-threaded environment?

fballiano:
the problem I see is when I run from the PHP CLI, not from the Mysql client (I use SequelAce) which anyway has an expire of 0

Screenshot 2023-09-14 alle 17 47 19

fballiano:
but actually if I run php cli I get an expire of 24h
Screenshot 2023-09-14 alle 17 53 26

weird... I'll try to find where is this set...

colinmollenhour:
It's probably just a feature of SequelAce so it makes sense now. :)

fballiano:
If I do something like this:

Screenshot 2023-09-14 alle 18 01 34

then it works ok.

not sure about all of this :-\

colinmollenhour:
In my tests `information_schema` method takes about half the time of `show table status`,  but considering you have to set a session variable first it requires two round trips at least on the first use. And then there is possibly a performance penalty of disabling the information schema cache which I have no idea how to measure. The `MAX(id)+1` method is the fastest by far. This is on a very big and active table from a production database. 
image

So I think the `MAX()+1` method is a very good suggestion as long as this method isn't used with tables that have deleted rows. If that is not a safe assumption to make (probably isn't) then I think the old method should be kept and no change made since there is not that much benefit for increased complexity and uncertainty.

Perhaps the better change would be to update where this method is used to only call the method once or use `MAX(id)+1` instead but leave getNextAutoincrement alone?

colinmollenhour:
I don't think stats cache expiry should be set globally if it is not absolutely necessary - that should be up to the DBA to decide ideally. It could be set just before querying information schema, perhaps using a static variable so that it is only done once.

fballiano:
sure, that's why I was testing it in the initStatement of the openmage connection, if we remove the GLOBAL it still works

Screenshot 2023-09-14 alle 18 49 22

Screenshot 2023-09-14 alle 18 49 44

fballiano (changes requested):
I have to block the merge because of https://github.com/OpenMage/magento-lts/pull/1721#issuecomment-1719734451

what are we doing about this?

kiatng:
I think we should close it.

fballiano:
me too, unless something completely different comes out, this is too risky

Issue #1163, opened 2020-08-22
Set log level from config


sreichel:
### Description (*)

Currently you just can enable/disable logging (system.log/exception.log -  or force logging), without any controll about log level. 

`Mage::log()` uses `Zend_Log` ...

>     const EMERG   = 0;  // Emergency: system is unusable
>     const ALERT   = 1;  // Alert: action must be taken immediately
>     const CRIT    = 2;  // Critical: critical conditions
>     const ERR     = 3;  // Error: error conditions
>     const WARN    = 4;  // Warning: warning conditions
>     const NOTICE  = 5;  // Notice: normal but significant condition
>     const INFO    = 6;  // Informational: informational messages
>     const DEBUG   = 7;  // Debug: debug messages

It would be nice to set a minimum level to get rid of debug infos, but still get warnings/errors logged ...

### Expected behavior (*)

Only log what you're interessted in.

### Benefits

Reduce logs to minimum.

elidrissidev:
I'm taking this!

ADDISON74:
This would be a helpful addition to OM. If anyone will implement it I hope the levels will be changed in Backend.

elidrissidev:
@ADDISON74 Yes, of course! I'm adding them with the other log settings.

PR #1621, opened 2021-05-16
Drop LTS csv files

luigifab:
### Description

This is a suggestion, to drop LTS csv files, to use only original csv files.
I also removed some unused translations (with the Magento word).

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1727, opened 2021-07-10
Set log level from config

elidrissidev:
This PR will add support for setting a max log level from the Admin.

### Description (*)
This will add a new config value called `dev/log/max_level`, it holds an integer representing the maximum log level that will be logged. Any logs with a higher level will not be logged.
The log levels are based on the ones defined in `Zend_Log` and are provided to the Admin area by a custom config source class: `Mage_Adminhtml_Model_System_Config_Source_Log_Level`.
This option is set to 7 (Debug) by default, meaning –if logs are enabled– everything will be logged. Additionally, if developer mode is active, log level check is ignored.
As a side note, this will also load `Mage_Core_LTS.csv` translations in adminhtml.

1. Fixes OpenMage/magento-lts#1163

1. In the Admin, go to `Configuration > Advanced > Developer > Log Settings` and set `Maximum Log Level` to the maximum level you want to log.
2. Log something with a level lower and higher than the max level selected to test: `Mage::log('Test max log levels', Zend_Log::INFO)`

### Questions or comments
Reviews and further discussion are welcome.

### ToDo
- [x] Update README.md to include the new config


sreichel (commented):
`validate-digits-range` could be added. (untested)

app/Mage.php
```diff
@@ -865,7 +865,14 @@ public static function log($message, $level = null, $file = '', $forceLog = fals
 
         static $loggers = array();
 
+        $maxLogLevel = (int) self::getStoreConfig('dev/log/max_level');
+
         $level  = is_null($level) ? Zend_Log::DEBUG : $level;
+
+        if ($level > $maxLogLevel) {
```
Flyingmana, line 872: we should still respect developerMode here as higher priority, means $isDeveloperMode === true => log everything

Flyingmana (changes requested):

Flyingmana (approved):

kiatng:
TLDR: 3 issues for me:

1. Need to review `Mage::log()` in OM and make sure the level is set correctly, but what about 3rd-party extensions?
2. May be it's useless, explain below.
3. This is a new feature, may be branch 20?

Take an example: https://github.com/OpenMage/magento-lts/blob/c2abc3faa77e42643037e0df4b00f98a762a8dad/app/code/core/Mage/Adminhtml/controllers/System/BackupController.php#L141-L146

(The above should be `Mage::logException($e)`. Since the param `$level` is omitted, it is default to `Zend_Log::DEBUG`. 

https://github.com/OpenMage/magento-lts/blob/c2abc3faa77e42643037e0df4b00f98a762a8dad/app/Mage.php#L868-L868

Because if  `dev/log/max_level` is set to another level in production where developer mode is off, it could cause pain to developer who is unaware of this setting. So, to be save, the level should be set to DEBUG, which make this PR not very useful. 

But if the community finds it useful, I'm not against it.

elidrissidev:
Thanks for the review @kiatng. To address your points:

> 1. Need to review Mage::log() in OM and make sure the level is set correctly, but what about 3rd-party extensions?

You're right, adding appropriate levels to the existing log statements is definitely something that should be addressed in a separate PR if this one is merged. About third-party extensions, in my company's codebase we have a few extensions that do use proper log levels but the majority don't, and because most M1 extensions are no longer being actively maintained, that's something that would need to be taken care of manually unfortunately.

> Because if `dev/log/max_level` is set to another level in production where developer mode is off, it could cause pain to developer who is unaware of this setting. So, to be save, the level should be set to DEBUG, which make this PR not very useful.

I don't understand exactly what you mean. `max_level` config is set by default to DEBUG to maintain backwards-compatibility, so if someone changes it, they must know what it does.

> 3. This is a new feature, may be branch 20?

If no one needs this in `1.9.4.x` then sure we can switch it to `20.x`.

kiatng:
> I don't understand exactly what you mean. max_level config is set by default to DEBUG to maintain backwards-compatibility, so if someone changes it, they must know what it does.

My own experience informs me that it's best not to assume the users always know what they do.  It's always good idea to explain the risks. If an admin user select an option other than DEBUG, and if the system encountered some issues, the developer who is unaware of this new feature may get very confuse.

Given that OM doesn't have a user guide, it's good idea to add more info in the note:
```xml
<comment>Log Levels are ordered from highest to lowest priority</comment>
```

Personally, I do not find this PR useful. But if others think is useful, I am not against it.

elidrissidev:
Ahh ok I get what you mean now. I can improve the comment to better communicate what the option does.

I requested more reviews and if the majority thinks it's not very useful then I'm completely fine with closing the P :) Thanks once again.

luigifab (commented):
Why not, but for I think I will keep debug level in config.
You can remove the `<validate>` because this is now a select.
And there is a "." missing at the end of the added comment :)

Perhaps you can also enable by default the log, or in another PR, or not.

elidrissidev:
> Open PR https://github.com/OpenMage/magento-lts/pull/1621 is removing all LTS CSV files but two of them are used in this PR (Mage_Adminthml_LTS.csv, Mage_Core_LTS.csv). 

Yeah I'm not sure if that PR is going to be merged as I see it has different opinions.

elidrissidev:
Rebased, updated config translation on Admin, and removed the translations I added from *LTS.csv files to the core ones.

ADDISON74 (dismissed):

luigifab (approved):

luigifab:
Edit: the `<validate>` can be removed.

luigifab (approved):

ADDISON74 (approved):

fballiano:
I'm merging since it's approved

app/Mage.php
```diff
@@ -863,7 +863,14 @@ public static function log($message, $level = null, $file = '', $forceLog = fals
 
         static $loggers = array();
 
+        $maxLogLevel = (int) self::getStoreConfig('dev/log/max_level');
```
tmotyl, line 866: this should be wrapped in try / catch because getStoreConfig can throw several exceptions (especially is Mage::log is used during early  stage of the request flow.

elidrissidev: Thanks! PR created.

Issue #1056, opened 2020-06-23
Unable to change tax rule name


sreichel:
### Preconditions (*)

1. latest 1.9.4x 
2. php7.3
3. mysql8

### Steps to reproduce (*)

1. got to Admin -> Sales -> Manage Tax Rules
2. save tax rule with only change name

### Expected result (*)

1. rule name change saved

### Actual result (*)

1. Error: `Rules (Standard) already exist for the specified Tax Rate, Customer Tax Class and Product Tax Class combinations`

PR #1728, opened 2021-07-10
Fix Tax Rule name update

elidrissidev:
Fix Tax Rule name update from Admin

### Description (*)
This PR will fix a bug where changing the name of a Tax Rule will trigger an error. This was caused by a check that is supposed to prevent creating multiple rules with the same Tax Rate/Customer Tax Class/Product Tax Class combination, but the issue was that it was checking the original name of the rule against the newly submitted one from the`POST` request, which is not gonna match if you submit a new name. All I did here is `load` the existing rule and use the original rule name instead of the submitted one to perform the check.

1. Fixes OpenMage/magento-lts#1056

Go to `Sales > Tax > Manage Tax Rules` and choose or create a rule:
1. You should be able to set a new name for any rule and save it.
2. If you try to create a new rule with the same Tax Rate/Customer Tax Class/Product Tax Class combination, it should show an error.

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php
@@ -110,7 +110,8 @@ class Mage_Adminhtml_Tax_RuleController extends Mage_Adminhtml_Controller_Action
             return $this->getResponse()->setRedirect($this->getUrl('*/tax_rule'));
         }
 
-        $ruleModel = $this->_getSingletonModel('tax/calculation_rule');
+        $ruleId = (int)$this->getRequest()->getParam('tax_calculation_rule_id');
+        $ruleModel = $this->_getSingletonModel('tax/calculation_rule')->load($ruleId);
         $ruleModel->setData($postData);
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Adminhtml/controllers/Tax/RuleController.php
```diff
@@ -110,7 +110,8 @@ public function saveAction()
             return $this->getResponse()->setRedirect($this->getUrl('*/tax_rule'));
         }
 
-        $ruleModel = $this->_getSingletonModel('tax/calculation_rule');
+        $ruleId = (int)$this->getRequest()->getParam('tax_calculation_rule_id');
+        $ruleModel = $this->_getSingletonModel('tax/calculation_rule')->load($ruleId);
```
Flyingmana, line 114: Using a singleton here to load by ID seems a bit dangerous. Is the singleton needed here?
Not an argument against the PR, as its happening limited to this PR, and a Singleton was already used before the change.
But wanted to at least note it.

Flyingmana (approved):

sreichel (approved):

luigifab (approved):

PR #1735, opened 2021-07-15
Fixed origData is empty in a quote

woutersamaey:
### Description (*)

`_origData` is empty when quotes are loaded though 
- loadByIdWithoutStore()
- loadActive()
- loadByCustomer()

When loading a quote with load(), then `_origData` is filled, but only in that case.

This fixed change detection and is important for my module because origData is always NULL on the cart page and several other locations.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -416,6 +416,7 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
         }
         $this->_getResource()->loadByCustomerId($this, $customerId);
         $this->_afterLoad();
+        $this->setOrigData();
         return $this;
     }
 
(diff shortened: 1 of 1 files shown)


kiatng (approved):

woutersamaey:
Can we get this moving along @Flyingmana @colinmollenhour @sreichel ?

elidrissidev (approved):

Flyingmana:
@woutersamaey did you check what would happen, if any data changes during the collectTotals() which is called in the `_afterLoad``?
If I see it right, this would currently lead to them not get persisted if something changes there.
Thats probably ok, as anything which does a relevant change, would call the collectTotals again, also the normal load() would cause the same.

Flyingmana (approved):

ADDISON74 (approved):

woutersamaey:
@Flyingmana this does not matter. Like you said yourself, it's the same as with the regular `load()` so no issue here. I'm just following what Magento itself does.

PR #1736, opened 2021-07-16
Also change the default when calling setTheme

Asfolny:
### Description (*)
When a custom controller is mailing multiple orders, first in a store where the design/theme/default is set (to "fr") and then in a store where it isn't set, we ran into problems where skinurls would refer to the first store used.

This is because calling `getTheme("skin")` the package would first look in `$this->_theme["skin"]`, and if it found nothing (as emulation calls `setTheme('')`), the packager will defer to call `getTheme("default")`, which returns "fr", rather than the expected "default" for the store. 

0. Create an extra store view (if you only have 1)
1. Add an extra folder to skin/frontend/base/
2. Add an image to both skin/frontend/base/default/images/ and skin/frontend/base/$FOLDER_STEP_2
3. Change the theme in one store view under Configuration > General > Design > Themes > Default
3. Create, change the hardcoded store ids and run the following shell script under shells/

```
<?php
require_once 'abstract.php';

class Storefront_TestStoreEmulation extends Mage_Shell_Abstract
{
    public function run()
    {
        // For testing purposes only, change if need be
        foreach (['3', '1'] as $store) {
            $appEmulation = Mage::getSingleton('core/app_emulation');
            $initialEnvironmentInfo = $appEmulation->startEnvironmentEmulation($store, 'frontend');

            // Must be an existing image
            echo Mage::getDesign()->getSkinUrl('images/logo.png') . "\n";

            $appEmulation->stopEnvironmentEmulation($initialEnvironmentInfo);
        }

        echo 'DONE!';
    }
}

$shell = new Storefront_TestStoreEmulation();
$shell->run();

```


Flyingmana:
From the code this looks ok and makes sense. But this should really be tested by one of the Reviewers to make sure it has no odd side effects.

app/code/core/Mage/Core/Model/Design/Package.php
```diff
@@ -248,7 +248,7 @@ public function setTheme()
     {
         switch (func_num_args()) {
             case 1:
-                foreach (array('layout', 'template', 'skin', 'locale', 'default') as $type) {
+                foreach (['layout', 'template', 'skin', 'locale'] as $type) {
```
kiatng, line 251: The PR change is lost!

```suggestion
                foreach (['layout', 'template', 'skin', 'locale', 'default'] as $type) {
```

kiatng (dismissed):
Tested.

sreichel:
Should this be added to `Mage_Adminhtml_Controller_Action::preDispatch()` too?

luigifab (approved):
In production for two weeks, seems good.

fballiano (approved):

fballiano:
I think this is now ready to be merged

kiatng (approved):

PR #1749, opened 2021-07-30
Removed _isMemoryLimitReached() nonsense

woutersamaey:
### Description (*)
If we don't have enough memory, we'll know soon enough. No need to do these computations every time.
### Questions or comments
- Less time/resources wasted on calculating memory needed.


luigifab (commented):
There is also a `_checkMemory` in `Mage_Catalog_Model_Product_Image`. On our store we have added `return true;`.

woutersamaey:
Good catch @luigifab, I removed these as well.

app/code/core/Mage/Catalog/Model/Product/Image.php
```diff
@@ -244,84 +244,6 @@ public function setSize($size)
         return $this;
     }
 
-    /**
-     * @param null $file
-     * @return bool
-     */
-    protected function _checkMemory($file = null)
```
Flyingmana, line 251: also removing the function may break modules using it. Please only remove the usages, and optionally mark the method as deprecated.
Please also do not remove the functionality, yet.

Flyingmana (changes requested):

Flyingmana:
Generally support the removal of the check. The PHP way of reporting reaching the memory limit is actually better and more verbose.
Although, this check allows for a more gracefully fail, as an exception allows handling it, while the php way lets the script die if I remember correctly.

woutersamaey:
@Flyingmana you are correct, but Magento does nothing with the exception that is worth our consideration.
Any script can run out of memory, so it's up to the developer to watch for memory issues in the PHP log.

Flyingmana:
@woutersamaey yes, thats why I support the change :)
But we need to keep the protected functions, in case other modules/shops use them on purpose.

woutersamaey:
@Flyingmana who in their right mind is going to use this on purpose?
I say we clean up a bit and move forward, not stay in place. This is the v20 branch, right?

Flyingmana:
It is not ours to judge others decision, and I want to please ask you to not talk ill of others, especially  not knowing which circumstances might have lead to these decisions.

Allowing BC breaks in the v20 branch is not an excuse to act reckless in this regard.

woutersamaey:
Okay, fair enough. I didn't want to be rude, just can't see what possible purpose it could serve.
Is OpenMage ever planning on breaking away or will things stay forever as they are? Would be nice to see a roadmap or something like that...

fballiano:
I support the cleaning tasks for v20 (some would say too much ehehehe), in this case, seeing you already discussed it, I'd just remove the calls to the "checkmemory" methods and mark all related methods as deprecated, so we can move this PR forward as we're all wasting computing cycles now :-)

elidrissidev:
I support this. Let's just remove the usages and mark the method as deprecated for the v19 branch.

tobihille (approved):

tmewes (dismissed):

tmewes:
@fballiano Can you please push the magic button (update the branch with the current version of 20.0) that we get rid of the conflicts and get this PR done? :)

fballiano:
I've restored the original methods marking them as deprecated, so it should be ready to merge.

But, I can't switch this PR to 1.9.4.x cause it's breaking for some reasons (it generates 90+ commits and 5000 modified files), if you're ok with it I'd close this one and recreate it as a new PR on v19.

tmewes:
Sounds like a good plan. 👍

fballiano:
Closing this in favour of https://github.com/OpenMage/magento-lts/pull/2471 which is based on v19 and non-BC :-)

PR #1750, opened 2021-07-30
Imagick adapter, handles color profiles better

woutersamaey:
### Description (*)
Proposal PR for an Imagick adapter.
This one has better image quality and can handle embedded color profiles.
I had a store where the image colors changed because GD could not properly process the ICC profile.
This adapter is better.

I'm just contributing my code for this 1 file and don't want to make a big deal out of it. If someone wants to help complete and integrate further, feel free.

### DIFF
--- a/lib/Varien/Image/Adapter/Imagick.php
+++ b/lib/Varien/Image/Adapter/Imagick.php
@@ -1,0 +1,223 @@
+<?php
+declare(strict_types=1);
+
+class Varien_Image_Adapter_Imagick extends Varien_Image_Adapter_Abstract
+{
+
+    /**
+     * @var Imagick
+     */
+    protected $_imageHandler;
+
+    /**
+     * @var string[]
+     */
+    protected $_requiredExtensions = ["imagick"];
+
+    /**
+     * Whether image was resized or not
+     *
+     * @var bool
+     */
+    protected $_resized = false;
+
+    public function __construct()
+    {
+        register_shutdown_function(array($this, 'destruct'));
+    }
+
+    /**
+     * Destroy object image on shutdown
+     */
+    public function destruct()
+    {
+        $this->_imageHandler->destroy();
+    }
+
+    /**
+     * Opens image file.
+     *
+     * @param string $filename
+     */
+    public function open($filename)
+    {
+        $this->_fileName = $filename;
+        $this->getMimeType();
+        $this->_getFileAttributes();
+
+        $this->_imageHandler = new Imagick();
+        $this->_imageHandler->readImage($filename);
+    }
+
+    public function save($destination = null, $newName = null)
+    {
+        if ($destination && $newName) {
+            $fileName = $destination . "/" . $newName;
+        } elseif (isset($destination) && !isset($newName)) {
+            $info = pathinfo($destination);
+            $fileName = $destination;
+            $destination = $info['dirname'];
+        } elseif (!isset($destination) && isset($newName)) {
+            $fileName = $this->_fileSrcPath . "/" . $newName;
+        } else {
+            $fileName = $this->_fileSrcPath . $this->_fileSrcName;
+        }
+
+        $destinationDir = (isset($destination)) ? $destination : $this->_fileSrcPath;
+
+        if (!is_writable($destinationDir)) {
+            try {
+                $io = new Varien_Io_File();
+                $io->mkdir($destination);
+            } catch (Exception $e) {
+                throw new Exception("Unable to write file into directory '{$destinationDir}'. Access forbidden.");
+            }
+        }
+
+//        if ($this->_fileType == IMAGETYPE_JPEG) {
+//            $threshold = (int)Mage::getStoreConfig('catalog/product_image/progressive_threshold');
+//            if ($threshold && $threshold <= (imagesx($this->_imageHandler) * imagesy($this->_imageHandler) / 1000000)) {
+//                imageinterlace($this->_imageHandler, 1);
+//
+//                TODO support progressive JPG? Like this?
+//                $this->_imageHandler->setInterlaceScheme();
+//            }
+//        }
+
+        // set quality param for PNG file type
+        $quality = $this->quality();
+        if ($quality !== null) {
+            if ($quality < 1) {
+                throw new RuntimeException('Image compression quality cannot be < 1');
+            } elseif ($quality > 100) {
+                throw new RuntimeException('Image compression quality cannot be > 100');
+            }
+            $this->_imageHandler->setCompressionQuality($quality);
+        }
+        $this->_imageHandler->writeImage($fileName);
+    }
+
+    public function display()
+    {
+        header("Content-type: " . $this->getMimeTypeWithOutFileType());
+        echo $this->_imageHandler->getImageBlob();
+    }
+
+    /**
+     * Change the image size
+     *
+     * @param int $frameWidth
+     * @param int $frameHeight
+     */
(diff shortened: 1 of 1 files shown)


tmotyl:
Nice idea!
Please also provide which IM version you have tested it with (different IM versions have different quircks). Maybe a check in the code would also be useful?
Do I get it right that to use this adapter, some additional changes/configuration is required? Can you describe it?

woutersamaey:
@tmotyl unfortunately I don't have the time, but you can look at `Varien_Image_Adapter::factory()`. This needs to change + some core files + maybe create a config setting somewhere.

I'm not aware of any Imagick quirks to be honest.

woutersamaey:
This adapter now also supports "exif:Orientation" which can cause an image to appear rotated in Magento when it is not rotated on your computer.

fballiano:
I always use https://github.com/colinmollenhour/Perfect_Watermarks cause the GD adapter never works for me

fballiano:
I've added the PerfectWatermarks module to https://openmage.readthedocs.io/en/latest/modules/more.html#images-optimization

fballiano:
@colinmollenhour what do you think about this PR? Asking you cause you're the author of the PerfectWatermarks extension that I normally use in my projects.

woutersamaey:
I have 1 huge volume site using this in production. They have a lot of custom code, but that should not affect the image handling. I'm very confident is it stable.

woutersamaey:
I just noticed that my class is still called `Storefront_MediaStorage_Model_Image_Adapter_Imagick`. Guess I need to change this to `Varien_Image_Adapter_Imagick`? It's been a while...

colinmollenhour:
@woutersamaey Yes the class name will need to match the pattern in the file path. :)

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,252 @@
+<?php
+declare(strict_types=1);
+
+class Storefront_MediaStorage_Model_Image_Adapter_Imagick extends Varien_Image_Adapter_Abstract
```
colinmollenhour, line 4:
```suggestion
class Varien_Image_Adapter_Imagick extends Varien_Image_Adapter_Abstract
```

fballiano (dismissed):

colinmollenhour (commented):
Seems that either the old class `Varien_Image_Adapter_Imagemagic` should be deleted or a new constant added for it such as `ADAPTER_IM_LEGACY`.

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,252 @@
...
+                    break;
+                case 3:
+                    $this->_imageHandler->rotateImage('#000000', 180);
+                    break;
+
+                case 6:
+                    $this->_imageHandler->rotateImage('#000000', 90);
+                    break;
+
+                case 8:
```
fballiano, line 69:
```suggestion
                    break;
                case 6:
                    $this->_imageHandler->rotateImage('#000000', 90);
                    break;
                case 8:
```

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,250 @@
+<?php
+declare(strict_types=1);
```
fballiano, line 2: should we have the strict_types? there's only in one single file in the whole repo, isn't it better just not to use it?=

fballiano:
@colinmollenhour Varien_Image_Adapter_Imagemagic doesn't even exist in our repo :-\

lib/Varien/Image/Adapter/Imagick.php
colinmollenhour, line 2:
```suggestion
```

colinmollenhour (dismissed):

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,252 @@
...
+
+//        if ($this->_fileType == IMAGETYPE_JPEG) {
+//            $threshold = (int)Mage::getStoreConfig('catalog/product_image/progressive_threshold');
+//            if ($threshold && $threshold <= (imagesx($this->_imageHandler) * imagesy($this->_imageHandler) / 1000000)) {
+//                imageinterlace($this->_imageHandler, 1);
+//
+//                TODO support progressive JPG? Like this?
+//                $this->_imageHandler->setInterlaceScheme();
+//            }
+//        }
```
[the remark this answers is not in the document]
colinmollenhour: I don't think anyone really cares about progressive anymore so probably better to delete but not a hard requirement IMHO in this case since it could potentially be tested and used someday..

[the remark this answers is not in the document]
fballiano: mmm but it seems to handle the whole JPEG processing too, maybe there's no need for anything for jpegs?

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,249 @@
...
+
+        // set quality param for PNG file type
```
fballiano, line 111: why is quality only for PNGs? I don't see any filter on the image type anyway

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,249 @@
...
+                throw new RuntimeException('Image compression quality cannot be < 1');
+            } elseif ($quality > 100) {
+                throw new RuntimeException('Image compression quality cannot be > 100');
+            }
+            $this->_imageHandler->setCompressionQuality($quality);
+        }
+        $this->_imageHandler->writeImage($fileName);
+    }
+
+    public function display()
```
fballiano, line 124:
```suggestion
    /**
     * @return string
     */
    public function display()
```

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,249 @@
...
+    {
+        header("Content-type: " . $this->getMimeTypeWithOutFileType());
+        echo $this->_imageHandler->getImageBlob();
+    }
+
+    /**
+     * Change the image size
+     *
```
fballiano, line 132:
```suggestion
```

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,250 @@
...
+            $offsetY = (int)round(($frameHeight - $dstHeight) / 2);
+            $canvas->compositeImage($this->_imageHandler, \Imagick::COMPOSITE_OVER, $offsetX, $offsetY);
+            $this->_imageHandler = $canvas;
+        }
+
+        $this->refreshImageDimensions();
+        $this->_resized = true;
+    }
+
+    public function rotate($angle)
```
fballiano, line 202: what type is the $angre?

lib/Varien/Image/Adapter/Imagick.php
```diff
@@ -0,0 +1,250 @@
...
+    {
+        $this->_imageHandler->rotateImage($this->imageBackgroundColor, $angle);
+        $this->refreshImageDimensions();
+    }
+
+    public function watermark($watermarkImage, $positionX = 0, $positionY = 0, $watermarkImageOpacity = 30, $repeat = false)
```
fballiano, line 208: what types are these params?

fballiano (changes requested):
- needs to implement gif, jpg, png, xbm, bmp, webp
- needs to implement the watermark method

PR #1751, opened 2021-07-30
Allow for easier adding of general order information on the order view page

woutersamaey:
### Description (*)
Allow for easier adding of general order information on the order view page.

In this screenshot I added "Some Item, Some Value" at the top of the order page using an event observer.

The data is collected as a key-value array using a new event: `adminhtml_block_order_view_collect_additional_info`
The value can be HTML.

I mentioned the new event in the README as well.

Screenshot from 2021-07-30 10-24-25

1. Fixes how hard it is to add generic order information without overriding the block or template or adding yet another block to the page.

###

### DIFF
--- a/app/design/adminhtml/default/default/template/sales/order/view/info.phtml
+++ b/app/design/adminhtml/default/default/template/sales/order/view/info.phtml
@@ -98,7 +98,7 @@ $orderStoreDate = $this->formatDate($_order->getCreatedAtStoreDate(), 'medium', 
                 <td class="label"><label><?php echo Mage::helper('sales')->__('%s / %s rate:', $_order->getOrderCurrencyCode(), $_order->getBaseCurrencyCode()) ?></label></td>
                 <td class="value"><strong><?php echo $_order->getBaseToOrderRate() ?></strong></td>
             </tr>
-            <?php endif 
+            <?php endif;
             
             foreach($this->getAdditionalInfo() as $label => $htmlFragment){
                 ?>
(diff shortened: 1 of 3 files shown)


app/design/adminhtml/default/default/template/sales/order/view/info.phtml
```diff
@@ -98,7 +98,17 @@ $orderStoreDate = $this->formatDate($_order->getCreatedAtStoreDate(), 'medium',
...
                 <td class="value"><strong><?php echo $_order->getBaseToOrderRate() ?></strong></td>
             </tr>
-            <?php endif ?>
+            <?php endif;
+            
+            foreach($this->getAdditionalInfo() as $label => $htmlFragment){
+                ?>
+                <tr>
+                    <td class="label"><label><?=$this->escapeHtml($label) ?></label></td>
+                    <td class="value"><?=$htmlFragment ?></td>
```
Flyingmana, line 107: As we directly use this as a html fragment here, I would like to make the array structure a bit more complex, so its already clear on populating the array, the value is used in an unescaped way.

so filling it like this:

```
$array['label']  = [
    "htmlFragment" => "<span>my html</span>",
];
```

Iam worried else its to easy people populate it with unexcaped data, which opens it up to xss attacks.

Flyingmana (changes requested):

fballiano:
I'll close this, as Daniel pointed out there is a potential for a vulnerability and we have zero feedback on this PR anyway since 2 years.

PR #1754, opened 2021-07-30
Fix (Adminhtml-Customer-Edit): allowed countries of selected website aren't considered when editing existing addresses

DeineTuer-GmbH:
### Description (*)

This PR enables the editing of customers with addresses using countries that aren't allowed in the default config (`General > Countries Options > Allow Countries [general/country/allow]`) but are in `Website` scope, without the need to change the country first.

This is achieved by taking the customer's store id into account when determining the allowed countries to fill the select options. Previously the input was always filled with the allowed country of the default store (admin, store id 0). Setting the customer's store id in select and multiselect attributes is necessary because the way in which values are set in `Mage_Adminhtml_Block_Widget_Form::_setFieldset()` line 208 and 210 respectively.

### Related Pull Requests

* None

* #1762 

1. Set system config `General > Countries Options > Allow Countries [general/country/allow]` so that at least one country is allowed in `Website` scope that is not allowed in default config scope. 
2. Have a customer for this website, log into it and create an address that uses a country that is only allowed in `Website` scope.
3. Navigate to Adminhtml Customer/Edit, edit the customer and select addresses.
4. See the created address with the country selected in the country input select.

### Questions or comments

This PR can't fix cases in which one decides to set the allowed countries in `Store View` scope because a customer is assigned to a website (and within the database to a store id it was created in). This is true in some circumstances and additionally depends on the configured store scenario (how many websites/stores/store views and in what distribution).

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
@@ -131,6 +131,9 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Addresses extends Mage_Adminhtml_Bl
             /* @var $attribute Mage_Eav_Model_Entity_Attribute */
             $attribute->setFrontendLabel(Mage::helper('customer')->__($attribute->getFrontend()->getLabel()));
             $attribute->unsIsVisible();
+            
+            // sets store id to attribute, is evaluated in customer country input fields to provide correct allowed countries
+            $attribute->setStoreId($customer->getStore()->getId());
         }
         $this->_setFieldset($attributes, $fieldset);
 
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
@@ -127,13 +127,25 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Addresses extends Mage_Adminhtml_Bl
             Mage::helper('adminhtml/addresses')
                 ->processStreetAttribute($attributes['street']);
         }
+
+        $customerStoreId = null;
+        if ($customer->getId()) {
+            $customerStoreId = Mage::app()->getWebsite($customer->getWebsiteId())->getDefaultStore()->getId();
+        }
+
         foreach ($attributes as $attribute) {
             /* @var $attribute Mage_Eav_Model_Entity_Attribute */
             $attribute->setFrontendLabel(Mage::helper('customer')->__($attribute->getFrontend()->getLabel()));
             $attribute->unsIsVisible();
             
-            // sets store id to attribute, is evaluated in customer country input fields to provide correct allowed countries
-            $attribute->setStoreId($customer->getStore()->getId());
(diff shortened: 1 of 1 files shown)


ADDISON74:
I think this PR should be better explained. Magento allows in the Backend to select the countries in which the store is available (General/Countries Options). If there is only one store you will not find that there is an issue because both in Frontend in editing the customer's account and in Backend in editing the customer's in the drop-down list there will be only the allowed countries set in the configuration.

The situation changes when you have several stores and each store has its own countries selection. Magento will show you only those in the general configuration, not those specific to the store. That's why it is necessary to use the store ID in order to get the correct selection of countries.

deinetuer:
@ADDISON74 Thanks for the feedback! We overhauled the explanation of the PR and created an accompaning issue (#1762) to explain the problem even further.

Sdfendor (approved):

tobihille (approved):

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
```diff
@@ -127,10 +127,25 @@ public function initForm()
             Mage::helper('adminhtml/addresses')
                 ->processStreetAttribute($attributes['street']);
         }
+
+        $customerStoreId = null;
+        if ($customer->getId()) {
+            $customerStoreId = Mage::app()->getWebsite($customer->getWebsiteId())->getDefaultStore()->getId();
+        }
```
kiatng, line 134: I see that the code block is moved from line 163. But if a website has more than one store, with mutually exclusive country list, e.g., the website's default store is store1, which has only France, store2 has only Germany, etc. Wouldn't non-French customers get incorrect country list? Is this case tested?

```suggestion
        $customerStoreId = $customer->getStoreId(); // null if new customer
```

Not tested, but I think the above should work for `suffix` and `prefix` later on, lines 180 and 194.

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Addresses.php
```diff
@@ -127,10 +127,25 @@ public function initForm()
...
             $attribute->unsIsVisible();
+            
+            /* Sets store id to attribute if select or multiselect because values are loaded store specific.
+               If not set values of default store are loaded which can be undesirable (e.g. country list in customer
+               addresses). */
+            /** @see Mage_Adminhtml_Block_Widget_Form::_setFieldset() lines 208 & 210 */
+            $frontendInputType = $attribute->getFrontend()->getInputType();
+            if ($customerStoreId !== null && ($frontendInputType === 'select' || $frontendInputType === 'multiselect')) {
+                $attribute->setStoreId($customerStoreId);
+            }
```
kiatng, line 148:
```suggestion
            $attribute->setStoreId($customerStoreId); // For country list
```
I think it's unnecessary to check for `null` and input type.

kiatng (commented):
Suggestions are note tested.

fballiano:
@DeineTuer-GmbH can you rebase to `main` branch and check the comments to the code?

fballiano:
This PR could hide a problem because of https://github.com/OpenMage/magento-lts/pull/1754#discussion_r950628781 and since it's now on a wrong branch and with no feedback since many month, sorry but I'll close it.

Hope it will be reopened by the original author with answers to the feedbacks :-)

PR #1758, opened 2021-07-31
Zend_Locale_Format::isNumber throws notice for float values

### LINKED ISSUE #1134: Zend_Locale_Format::isNumber throws notice for float values
### Preconditions (*)

1. OpenMage 1.9.4.5
2. PHP 7.4
3. adding 

### Steps to reproduce (*)

1. call Neuca_Base_Wishlist_IndexController cartAction, which calls `_processLocalizedQty` and then `$this->_localFilter->filter((float)$qty);`

### Expected result (*)

1. OpenMage doesn't throw a notice on PHP 7.4

On one hand Zend should check input variable type and don't process it if its already a number,
on the other hand OpenMage should not pass a float to `filter` in line:
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Wishlist/Controller/Abstract.php#L62

The casting to float was added in Magento 1.7. This cast made the whole zend filtering obsolete, so it can be removed in my POV.

### Actual result (*)

1. OpenMage throws error when float value is being passed to Zend_Locale_Format::is_number
```
Notice: Trying to access array offset on value of type float  in htdocs/lib/Zend/Locale/Format.php on line 523

#0 htdocs/lib/Zend/Locale/Format.php(523): mageCoreErrorHandler(8, 'Trying to acces...', '..', 523, Array)
#1 htdocs/lib/Zend/Filter/LocalizedToNormalized.php(100): Zend_Locale_Format::isNumber(1, Array)
#2 
```
The notice is being thrown because Zend tries to access float value as array in line:
```
if (!empty($input) && ($input[0] == $symbols['decimal'])) {
```

fballiano:
Everything is explained in issue https://github.com/OpenMage/magento-lts/issues/1134

Fixes #1134

### DIFF
--- a/app/code/core/Mage/Wishlist/Controller/Abstract.php
+++ b/app/code/core/Mage/Wishlist/Controller/Abstract.php
@@ -51,15 +51,11 @@ abstract class Mage_Wishlist_Controller_Abstract extends Mage_Core_Controller_Fr
      *
      * @param string $qty
      * @return float|int|null
+     * @deprecated
      */
     protected function _processLocalizedQty($qty)
     {
-        if (!$this->_localFilter) {
-            $this->_localFilter = new Zend_Filter_LocalizedToNormalized(
-                array('locale' => Mage::app()->getLocale()->getLocaleCode())
-            );
-        }
-        $qty = $this->_localFilter->filter((float)$qty);
+        $qty = (float)$qty;
         if ($qty < 0) {
             $qty = null;
         }
--- a/app/code/core/Mage/Wishlist/controllers/IndexController.php
+++ b/app/code/core/Mage/Wishlist/controllers/IndexController.php
@@ -508,8 +508,8 @@ class Mage_Wishlist_IndexController extends Mage_Wishlist_Controller_Abstract
                 $qty = 1;
             }
         }
-        $qty = $this->_processLocalizedQty($qty);
-        if ($qty) {
+        $qty = (float)$qty;
+        if ($qty && $qty>0) {
             $item->setQty($qty);
         }
 


Flyingmana:
the \Zend_Filter_LocalizedToNormalized::filter method is used in a lot of places related to qty, so Iam worried, this is potentially causing followup issues, as the qty might be stored in the database as a localized value...

I think the better and more stable solution would be to fix the \Zend_Locale_Format::isNumber method by casting the $input to a string in the first line.

fballiano:
the cast was already in the original method:

`
-        $qty = $this->_localFilter->filter((float)$qty);
`

that's why I thought it was safe to use it anyway, shouldn't change anything in the actual behavior right?

Flyingmana:
yeah, but the return value is not a float, but a string, a localized string.

we only input the locale, lets see for which parts it is used.
For the return value we are more interested in `Zend_Locale_Format::getNumber` then `Zend_Locale_Format::isNumber`.
there we have somewhere in the beginning
```
        if (!is_string($input)) {
            return $input;
        }
```
So a float is supported, and yes, it seems we would not have any difference in behavior then.
 
Anyway, the locale is used as `$symbols = Zend_Locale_Data::getList($options['locale'],'symbols');`
This is used to determine the minus, group and decimal symbol inside the string, and then to replace them with the usual ones.
So yes, if it is provided with a float, it results in the same value, but as a float.


So yeah, change is ok, but we should keep in mind, this is not in line with other usages, like \Mage_Checkout_CartController::_updateShoppingCart

Flyingmana (approved):

PR #1769, opened 2021-08-04
Array to string conversion when adding reviews from backend

fballiano:
This PR fixes issue https://github.com/OpenMage/magento-lts/issues/741 simply forcing a unique ID for the review_products_grid in order to avoid conflicts with the main products grid

1. https://github.com/OpenMage/magento-lts/issues/741

All is explained in the issue

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Review/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Review/Product/Grid.php
@@ -37,6 +37,7 @@ class Mage_Adminhtml_Block_Review_Product_Grid extends Mage_Adminhtml_Block_Cata
     public function __construct()
     {
         parent::__construct();
+        $this->setId('reviewProductGrid');
         $this->setRowClickCallback('review.gridRowClick');
         $this->setUseAjax(true);
     }


elidrissidev (approved):

ADDISON74 (approved):

luigifab (approved):

midlan (approved):

luigifab:
Shame on us! This break the possibility to add new comment from backend (v19 and v20).
After choosing a product, the review form appear and at the end of the page, products list is still present.

v20.0.13
image

v20.0.14
image

ADDISON74:
I confirm the issue reported by @luigifab. It is obvious that this implementation was not tested enough although the list of those who approved about a year ago is unusually long (which includes me as well). It can be reverted or fixed in another PR.

ADDISON74:
I confirm the issue reported by @luigifab. It is obvious that this implementation was not tested enough although the list of those who approved about a year ago is unusually long (which includes me as well). It can be reverted or fixed in another PR.

The issue reported in #741 was that if you set a filter in the product grid, for example Price From 160 to 160 and you go to add a new review, after pressing the [Add New Review] button, the grid from which you select products was identical to the one previously filtered in Catalog > Products, and on the Price column, filtered, were the price values set at 160.

This PR was supposed to return the full grid, unfiltered, which it does, but after selecting the product, the changes on the page can no longer be saved because there is only the [Back] button.

I have analyzed whether a Review can be added to at least one of the filtered products in the Catalog > Products grid and I confirm that it is not possible. Practically all products are affected by the new issue.

ADDISON74:
The modification not only removes the buttons but adds the product grid at the bottom

screenshot

fballiano:
mmm the buttons are there but get hidden via some javascript...

fballiano:
ok I think I got it

ADDISON74:
Good observation. The two buttons have inline css "display:none". If I make them visible I can save the review. The product grid at the bottom of the screen would also have to be removed. I don't understand how it appeared there, it was probably hidden before this PR.

fballiano:
done: https://github.com/OpenMage/magento-lts/pull/2415

PR #1779, opened 2021-08-12
Fix newsletter subcriber loading by customer

mmenozzi:
### Description (*)

IMHO the implementation of `\Mage_Newsletter_Model_Resource_Subscriber::loadByCustomer` is wrong. Currently, if there are 2 customers on 2 different websites with the same email address, if one customer subscribe himself to the newsletter from one website it will then appear subscribed even from the other website. This is wrong especially if the two websites are synchronized with two different audiences on the mail marketing platform (for example MailChimp). Moreover, if Magento is configured to split customers between websites this implementation could lead to have inconsistent data in the `newsletter_subscriber` table: you could have the subscriber assigned to a `store_id` which is not the the one of its linked customer.

For these reasons I propose this change.

1. Create 2 different websites
2. In Magento admin `System -> Configuration -> Customers -> Customer Configuration -> Account Sharing Options`, set `Share Customer Accounts` to `Per Website`
3. Create 2 different customer account one for each website **with the same email address**
4. Login with one customer account on one website
5. Subscribe to the newsletter from the `/newsletter/manage` page
6. Login with the other customer account on the other website
7. Go to the `/newsletter/manage` page
8. You'll see that the customer is already subscribed and **this is wrong**. The second customer never subscribed to the newsletter.
9. On this page, with the second customer, unsubscribe from the newsletter.
10. Look at the `newsletter_subscriber` table. You'll see that the `store_id` is the one of the second website but the `customer_id` is the one of the first website which is inconsistent.

### DIFF
--- a/app/code/core/Mage/Newsletter/Model/Resource/Subscriber.php
+++ b/app/code/core/Mage/Newsletter/Model/Resource/Subscriber.php
@@ -126,9 +126,13 @@ class Mage_Newsletter_Model_Resource_Subscriber extends Mage_Core_Model_Resource
 
         $select = $this->_read->select()
             ->from($this->getMainTable())
-            ->where('subscriber_email=:subscriber_email');
+            ->where('subscriber_email=:subscriber_email')
+            ->Where('store_id=:store_id');
 
-        $result = $this->_read->fetchRow($select, array('subscriber_email'=>$customer->getEmail()));
+        $result = $this->_read->fetchRow(
+            $select,
+            array('subscriber_email'=>$customer->getEmail(), 'store_id' => $customer->getStoreId())
+        );
 
         if ($result) {
             return $result;


ADDISON74:
Thank you for your contribution. I have never used the same installation for several websites but reading the forums over time I found that there are problems with the store id's. We may have other issues that cannot be identified unless they are tested intensively as you did in this case.

app/code/core/Mage/Newsletter/Model/Resource/Subscriber.php
```diff
@@ -126,9 +126,13 @@ public function loadByCustomer(Mage_Customer_Model_Customer $customer)
 
         $select = $this->_read->select()
             ->from($this->getMainTable())
-            ->where('subscriber_email=:subscriber_email');
+            ->where('subscriber_email=:subscriber_email')
+            ->Where('store_id=:store_id');
```
fballiano, line 130: wouldn't it be better to filter by website_id instead of store_id? cause customers can be global or per-website so I think it could make more sense to filter by website_id no?

mmenozzi: It would be better. But there's no website_id column on the newsletter_subscriber table as far as I remember.

fballiano: oh right I'm so sorry, I was thinking about the customer_entity and not the newsletter_subscriber!

mmenozzi: No problem @fballiano! Thank you for your review!

fballiano (approved):

luigifab (approved):

Issue #1781, opened 2021-08-14
20.0.12 Wrong field specified


pauldpauld:
### Preconditions (*)

new fresh download 20.0.12 no extensions added.
PHP 7.4.22
Fedora 33
old DB from 1.7.0.2 (which seems to have been updated correctly to work with 20.0.12) products, customers, sales all look correct.

### Steps to reproduce (*)

go to System Configuration Design
Don't even have to change anything, just save; error appears

I checked previous bugs and noted the adding of bool to a file in Mage, my file has that update. 
(ref: #1411 #1417 #1418 I think)

### Expected result (*)

1. page should save with my changes
2.

### Actual result (*)

1. ![image](https://user-images.githubusercontent.com/1549372/129441790-fe1f45f8-5e9a-46e9-ae18-9060281d2fd8.png)
2. pages does not save with changes
3. page does not save even if you leave everything unchanged.

kiatng:
May be it's helpful to find out which field is throwing the error. Can you change the following line 

https://github.com/OpenMage/magento-lts/blob/02c01c1e12dd5705963e7dbb6fe7d54c0ee7a994/app/code/core/Mage/Adminhtml/Model/Config/Data.php#L438-L438

to

```php
Mage::throwException(Mage::helper('adminhtml')->__('Wrong field specified in %s/%s/%s', $section, $group, $field)); 
```

Then, go to your backend and repeat your steps to see the error. Hopefully, this time, it'll show the responsible field.

pauldpauld:
*please insert Embarrassed emoji here*

Thanks Kiatng, I did as you suggested, it showed an error with Amazon_image (which is not part of OpenMage), I had an attribute with the special size and white background added to my image selection in product images config area. Which of course was brought over with my Db, but not added into my new Openmage.
As soon as I realised this, I deleted the Amazon_image attribute, and Openmage now operates correctly.
Sorry for wasting your time, but you were VERY helpful, thanks.
Please close this Bug report.

P.S. Would adding your suggested code change be helpful in the next release of Openmage? Just a thought.
Regards, Paul

kiatng:
Thanks for your feedback. I'm glad you find it helpful. I'll attempt a PR.

sreichel:
@all-contributors add @pauldpauld bug

PR #1784, opened 2021-08-15
Add field info to exception when thrown in saving System Configuration.

kiatng:
### Description (*)

The exception message "Wrong field specified." is not very helpful to the users. This PR add the field info that caused the exception. Now, the user can take appropaite action.

1. Fixes OpenMage/magento-lts#1781

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Config/Data.php
+++ b/app/code/core/Mage/Adminhtml/Model/Config/Data.php
@@ -435,7 +435,8 @@ class Mage_Adminhtml_Model_Config_Data extends Varien_Object
                     }
                 }
                 if (($groupConfig ? !$groupConfig->dynamic_group : true) && !$this->_isValidField($fieldConfig)) {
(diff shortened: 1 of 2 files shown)


fballiano:
Hi @kiatng, I think this PR is a good idea, but doesn't it break existing language packs (I know... admin translations suck... but still) because of the new string?

I know it's not the best but wouldn't it be possible to output the field details without changing the translation string?

kiatng:
@fballiano What would be a better way to handle the translation? Like the following?

```php
$message = Mage::helper('adminhtml')->__('Wrong field specified.')) . ' ' . Mage::throwException(Mage::helper('adminhtml')->__('See %s/%s/%s', $section, $group, $field));
Mage::throwException($message);
```

fballiano:
your original way was the best for sure but just not to have any troubles with translations how do you see something like this?

```php
$message = Mage::helper('adminhtml')->__('Wrong field specified.') . ' ' . Mage::helper('adminhtml')->__('(%s/%s/%s)', $section, $group, $field);
```

luigifab (approved):

elidrissidev (approved):

PR #1802, opened 2021-08-27
PHPDoc fix for Mage_Core_Model_Config

Sdfendor:
### Description (*)

The PHPDoc return type of method `getTablePrefix()` was wrong (_string_ vs. _Mage_Core_Model_Config_). The method is used one time in core but could be and - in my case is - used more frequently by 3rd party modules. The using line casts the value correctly to string. An IDE that parses PHPDoc and considers the return types assumed the cast as redundant though: 

old
image

fixed
image
 
* Check your IDE (with active PHPDoc support)

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1558,7 +1558,7 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     /**
      * Get DB table names prefix
      *
-     * @return string
+     * @return Mage_Core_Model_Config_Element
      */
     public function getTablePrefix()
     {


kiatng (approved):

luigifab (approved):

elidrissidev (approved):

midlan:
From my view we should rather match the code to the PHP doc. I'm assuming that the purpose of the method was to get the string table prefix. **Not the object that can be manipulated by reference.**

```php
    /**
     * Get DB table names prefix
     *
     * @return string
     */
    public function getTablePrefix()
    {
        return (string) $this->_xml->global->resources->db->table_prefix;
    }
```

Sekiphp:
@Sdfendor Fox example `Mirasvit_Rewards` using at one place this method.
@midlan I agree, 3rd party modules uses output of this method like a string. But if we add casting to `string` to method directly, we should change base of this branch to `20.x`

Sdfendor:
@Sekiphp Thanks for the example. 

I agree with your assessment that we should change the base branch to 20.x in case the cast is added. Maybe the current PHPDoc adjustment could be the solution for 1.9.4.x though. I could provide an accompanying PR for 20.x with the change proposed by @midlan.
All other occurrences I encountered with 3rd party modules have the string cast present as seen in the core so I don't see any harm in only changing the type hint (for 1.9.4.x).

Flyingmana (approved):

PR #1804, opened 2021-09-03
Allow to cancel orders like registerCancellation

luigifab:
### Description

This little change allow user to cancel order with a comment.

### Manual testing scenarios

For example:
```php
$order->registerCancellation('Message')->save();
$order->cancel('Message')->save();
```

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -1248,15 +1248,14 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
 
     /**
      * Cancel order
-     *
+     * @param string $comment
      * @return $this
      */
-    public function cancel()
+    public function cancel($comment = '')
     {
         if ($this->canCancel()) {
             $this->getPayment()->cancel();
-            $this->registerCancellation();
-
+            $this->registerCancellation($comment);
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Sales/Model/Order.php
```diff
@@ -1248,14 +1248,15 @@ public function unhold()
 
     /**
      * Cancel order
-     *
+     * @param string $comment
+     * @param bool $graceful
      * @return $this
      */
-    public function cancel()
+    public function cancel($comment = '', $graceful = true)
```
kiatng, line 1255: I think the second param `$graceful` should be removed as it has no meaning in this context, because `$this->canCancel()` can only be `true` in

https://github.com/OpenMage/magento-lts/blob/4885b20d86bba7aa200007a76620575488e25846/app/code/core/Mage/Sales/Model/Order.php#L1273-L1276

and this line 
https://github.com/OpenMage/magento-lts/blob/4885b20d86bba7aa200007a76620575488e25846/app/code/core/Mage/Sales/Model/Order.php#L1304
will never execute.

luigifab: You are so right!

kiatng (approved):

fballiano (approved):

PR #1814, opened 2021-09-07
Update Mage_Eav_Model_Entity_Collection_Abstract::joinTable()

cebe:
### Description (*)

allow passing a `Zend_Db_Expr` object.

Zend_Db_Select::join*() methods accept `Zend_Db_Expr` for a table name, which allows joining a subquery instead of a table by name.

When passing such an object through `Mage_Eav_Model_Entity_Collection_Abstract::joinTable()` there is an edge case where it fails on this line when the subquery contains a `/`. Adding the `is_object()` check makes this more robust.

### Related Pull Requests

none.

none.

unfortunately this is a very specific edge case so there is no manual testing.
Here is an example call on how a subquery can be used:

```php
        $subQuery = new Zend_Db_Expr("(" .
            "SELECT request_path, category_id FROM core_url_rewrite WHERE is_system = 1 ".
            " AND " . $this->_connection->quoteInto('store_id = ?', $storeId).
            " AND " . $this->_connection->quoteInto('id_path LIKE ?', 'category/%').
        ")");
        $collection->joinTable(
            array('url_rewrite' => $subQuery),
            'category_id=entity_id',
            array('request_path'),
            null,
            'left'
        );
```

If you need more information, I am happy to provide it.


elidrissidev:
I'd also update the method's docblock to better indicate it supports `Zend_Db_Expr` as a table name

cebe:
> I'd also update the method's docblock to better indicate it supports `Zend_Db_Expr` as a table name

done.

elidrissidev (approved):

ADDISON74 (approved):

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -772,7 +772,7 @@ public function joinTable($table, $bind, $fields = null, $cond = null, $joinType
         }
 
         // validate table
-        if (strpos($tableName, '/') !== false) {
+        if (!is_object($tableName) && strpos($tableName, '/') !== false) {
```
kiatng, line 775: I prefer a stricter check with `instanceof Zend_Db_Expr`

fballiano: actaully I don't understand why the original code was `if (strpos($tableName, '/') !== false)`, I wouldn't even check for anything, just that `$tableName` is a string

cebe: @kiatng @fballiano like this? checking for string before making a string comparision seems to be the most resonable thing to do:

```suggestion
        if (is_string($tableName) && strpos($tableName, '/') !== false) {
```

elidrissidev: @fballiano If the table name contains a `/` it means it's a resource name like `core/url_rewrite`, so it should be converted to the raw table name `core_url_rewrite`.

elidrissidev:
Can you also resolve the conflicts, please.

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
fballiano: right!

kiatng (dismissed):

fballiano:
> Can you also resolve the conflicts, please.

I've done that

fballiano (approved):

ADDISON74 (approved):

elidrissidev (approved):

PR #1816, opened 2021-09-08
Update phpdoc in Collection.php

rubanooo:
Matched the phpdoc with the typehint from call_user_func

### DIFF
--- a/lib/Varien/Data/Collection.php
+++ b/lib/Varien/Data/Collection.php
@@ -449,7 +449,7 @@ class Varien_Data_Collection implements IteratorAggregate, Countable
      *
      * Returns array with results of callback for each item
      *
-     * @param string $callback
+     * @param closure $callback
      * @param array $args
      * @return array
      */
@@ -470,7 +470,7 @@ class Varien_Data_Collection implements IteratorAggregate, Countable
     }
 
     /**
-     * @param string $obj_method
+     * @param closure $obj_method
      * @param array $args
      */
     public function each($obj_method, $args = array())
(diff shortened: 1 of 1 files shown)


elidrissidev (dismissed):

Sdfendor (approved):

kiatng:
Question: `Closure` is an object, line 459 tests for string:

https://github.com/OpenMage/magento-lts/blob/4885b20d86bba7aa200007a76620575488e25846/lib/Varien/Data/Collection.php#L452-L460

isn't this inconsistent? Ref https://www.php.net/manual/en/language.types.declarations.php, `callable` is a type, but I am not sure if it is better like this:

```php
     * @param callable $callback
```

Sdfendor:
I'am with @kiatng on this one and will change my review accordingly. Callable is the type that is referenced in https://www.php.net/manual/de/function.call-user-func-array.php which is the method used in `walk` function.

Sdfendor (changes requested):
Callable is the type that is referenced in https://www.php.net/manual/de/function.call-user-func-array.php and https://www.php.net/manual/de/function.call-user-func.php which are the methods used in walk and each functions respectively. I would use `callable` instead of `closure`.

rubanooo:
I also had `callable` in my local branch, but forgot to change this in my fork. I changed it to `callable` now :)

elidrissidev (approved):

Sdfendor (approved):

woutersamaey (approved):

kiatng (approved):

luigifab (approved):

midlan (approved):

Flyingmana (approved):

PR #1832, opened 2021-09-23
Replace zend with zf1-future 1.21.1

luigifab:
### Description

This PR replace **zend 1.12.16** with **[zf1-future 1.20.0](https://github.com/Shardj/zf1-future)** (suggested by @kiatng).

I removed all files except one in _lib/Zend_, copy/pasted, and executed _lib/Zend/replace_recursive.php_.

Not fully tested (I just checked that all admin pages of all my modules works with this PR and PHP 8.0), not checked if changes made by magento/openmage contributors are still present.

### Contribution checklist


tmotyl:
@luigifab please keep in mind that zend was modified in Magento and openmage. Also m2 guys have their own fork of zend, not sure what are the changes there

aterjung:
aren`t these changes placed at 

app/code/core/Zend ?
 
At least comparing these files to know what was changed and what was the intension is a pain, cause these files were from different Zend1 releases.

For example, app/code/core/Zend/Db/Statement.php is (c) 2005-2010, the Select.php in the same folder is (c) 2005-2015 

With some luck, we can guess the what was the reason if we find the Magento Release where these files were added?

ADDISON74:
Magento Team has adapted ZF1 but a detailed comparison can highlight the changes. As you proposed the removal from the source code of the part related to MySQL4 this change is a substantial one too.

Out of curiosity did you take a test? Any errors reported?

luigifab:
Yes, I did a quick test, I checked 30 admin pages, no errors. But it's clearly not enough.
I have started a detailed comparison, and on the first 106/1288 files modified:

95% of changes are:
- `array` => `[]`
- `intval` => `(int)`
- new empty lines
- docblock
- no else after return/throw/continue (`if (...) { ... return; } else { ... return; }` => `if (...) { ... return; } ... return;` )
- `$a = ...; return $a;` => `return ...;`

other changes are:

_lib/Zend/Application.php_
line 333, I think the comment of `require_once` is an **error** of the _lib/Zend/replace_recursive.php_ script
but this does not crash my admin pages

_lib/Zend/Auth/Adapter/Http.php + lib/Zend/Auth/Adapter/Ldap.php + lib/Zend/Barcode/Object/Code128.php + lib/Zend/Barcode/Object/Upca.php + lib/Zend/Barcode/Object/Upce.php + lib/Zend/Barcode/Renderer/Pdf.php + lib/Zend/Cache/Backend/Apc.php + lib/Zend/Cache/Backend/Sqlite.php + lib/Zend/Cache/Backend/Static.php + lib/Zend/Cache/Backend/TwoLevels.php + lib/Zend/Cache/Backend/ZendPlatform.php + lib/Zend/Cache/Core.php + lib/Zend/Captcha/Image.php + lib/Zend/Cloud/DocumentService/Adapter/SimpleDb.php_
strict comparison
`if ($i == 0 || $i == 11) {`  =>  `if ($i === 0 || $i === 11) {`
`strcasecmp($value, 'true') == 0)`  =>  `strcasecmp($value, 'true') === 0)`
`if ($currentCharset != 'C')`  =>  `if ($currentCharset !== 'C')`
...

_lib/Zend/Cache/Frontend/Page.php_
`if (count($var) > 0) {`  =>  `if (is_array($var) && count($var) > 0) {`

_lib/Zend/Barcode/Renderer/Image.php_
a new thing
`if (gettype($image) != 'resource' || get_resource_type($image) != 'gd') {`
replaced by
`if ((gettype($image) === 'resource' && get_resource_type($image) === 'gd') || get_class($image) === 'GdImage') {`

_lib/Zend/Captcha/Word.php_
a new thing
added Zend/Crypt/Math.php

woutersamaey:
What about deleting unneeded Zend files to reduce the attack surface? Any thoughts on that?
We could script something that lists all classes and scans the codebase for used and unused Zend files...

luigifab:
Yeeessssss! But I don't know how to do that.

woutersamaey:
1. Write a PHP script that collects a list of all Zend classes, just by looking at the file structure
2. Write script that searches all Magento code and extracts all mentions of `new Zend_` meaning a Zend class is created.
3. Search the Zend files again for any Zend classes that are used in the files we want to keep. These are dependancies and also need to stay.
4. Compare the 2 lists (files needed vs files on disk)
5. Delete the unused files

luigifab:
Um, why not.
I'll read the remaining 889 changed files first.

Flyingmana:
> What about deleting unneeded Zend files to reduce the attack surface? Any thoughts on that?

that would be an immense BC break for all module providers

luigifab:
I reverted [Short Array Syntax](https://github.com/thomasbachem/php-short-array-syntax-converter#revert) to reduce number of changes.

When hiding [white space changes](https://github.com/OpenMage/magento-lts/pull/1832/files?diff=unified&w=1), there **513 files** changed:
- 60 deleted (lib/Zend/Gdata/YouTube*)
- 2 renamed (Zend/Validate/Barcode/ Intelligentmail.php → IntelligentMail.php, Zend/Tool/Framework/Provider/ DocblockManifestable.php → DocblockManifestInterface.php)
- 167 whitespace-only changes

So 284 files left.
I check them carefully.

luigifab:
I copy/pasted some of _app/code/core/Zend/*.php_ files to compare changes from OpenMage/Magento to zf1-future.
Well, I would be unable to say who is right and who is wrong.

- [Zend/Xml/Security.php](https://github.com/luigifab/zf1-future/commit/8f3f4e1e82c8fe4a11c9fb1b392adec27e62483c), Magento/OpenMage is better?
- [Zend/Controller/Request/Http.php](https://github.com/luigifab/zf1-future/commit/ec58319739cce983d380c061180dbcd5d0685033), zf1-future is better? except for `@return` and `HTTP_X_FORWARDED_PROTO`?
- [Zend/Validate/Hostname.php](https://github.com/luigifab/zf1-future/commit/8016d9e964fef72b1c44133902d3917ac5e13cff), Magento/OpenMage added `checkDnsRecords` method
- [Zend/Validate/File/Upload.php](https://github.com/luigifab/zf1-future/commit/b7155691d80d88aa3b6a374147b6d7c2e453b998), Magento/OpenMage is better with `PHP7.2 fix`? or zf1-future is better with `$this->_messages = []`?
- [Zend/Serializer/Adapter/PhpCode.php](https://github.com/luigifab/zf1-future/commit/bdf052fea0ebb9830f6a1595253a54d83d28ea35), Magento/OpenMage added `if (is_array($opts)) {`, but `$opts` parameters must be an array
- [Zend/Validate/EmailAddress.php](https://github.com/luigifab/zf1-future/commit/b0a44ff2f6263469661a62f6c8ab971b813a1c9d), Magento/OpenMage is better?
- [Zend/Locale/Math/PhpMath.php](https://github.com/luigifab/zf1-future/commit/db895fe11128c50e634c00136263f75ec082b223), Magento/OpenMage is better?

Remaining files:
- app/code/core/Zend/Date.php
- app/code/core/Zend/Db/Select.php
- app/code/core/Zend/Db/Statement.php
- app/code/core/Zend/Filter/PregReplace.php
- app/code/core/Zend/Form/Decorator/Form.php
- app/code/core/Zend/Mime.php et lib/Zend/Mime.php

Command used to preview changes: `diff -I "    . @return .*" -r --ignore-all-space --ignore-blank-lines app/code/core/Zend/ lib/Zend/ | grep -v "Only in lib/Zend" | most`

fballiano:
I'd totally dig the original idea of this PR (a total replacement, not the removal of unused files) (I think I also suggested it sometime in the past), it would remove a big chunk of code that we've to maintain!

fballiano:
But I think it shouldn't be done in the 1.9 branch.

luigifab:
I had chosen 1.9 branch because I think (well I hope) that this update will not change anything.

Flyingmana:
I would prefer to have this in the 1.9 branch, as this also moves the responsibility of security vulnerabilities of this into the external project and it makes it easier for us, as long as the class names are the same.

Although, pure technically, I would include it as a composer dependency, so we can actually remove the code out of the repository and make it easier exchangeable. But this requires finishing the building of release downloadables as I did start here: https://github.com/Flyingmana/magento-lts/blob/release-actions/.github/workflows/release.yml#L49

fballiano:
I've no problem in having it in v19, I thought it was a bit more risky so maybe the general consensus would be more "safe" :-)

Also, totally agree on the composer dependency and removal of that code from the repo.

what's missing from the build process?

Sdfendor:
As far as I understand it, in case https://github.com/Shardj/zf1-future/pull/191 is approved, the plan to replace Zend 1 with ZF1-Future would possibly be affected in the sense that minimal version of OpenMage (19 and 20) is currently PHP 7.0.

luigifab:
We can also propose to update to keep PHP 7.0 as minimal version.

fballiano:
Could we update this PR with zf1 1.21.1 and try to merge it?

Sdfendor:
To update on the comment above: min version of PHP would then be 7.1 (from 7.0). I'm totally fine to increase minimum version also in general not only in the context of this PR.

fballiano:
> To update on the comment above: min version of PHP would then be 7.1 (from 7.0). I'm totally fine to increase minimum version also in general not only in the context of this PR.

I'd be too, 7.1 is out of support since two years now

Sdfendor:
At least the readme doesn't reflect though: 
image

fballiano:
@Sdfendor https://www.php.net/supported-versions.php

Sdfendor:
Sorry I misread. I though you talked about OpenMage not PHP EOL. Yes of course you are right about 7.0 and 7.1.

fballiano:
Ideally lib/Zend (but also lib/Cm and lib/Credis) should be removed from this repo and added as a composer dependancy. This repo would be cleaner, with better responsibilities separation and we wouldn't have people doing PRs agains the zf1 code (which creates problems with updates of the whole zf1 library).

I tried to check the magento-composer-installer but I can't get a grasp on how it works (never used in any of my projects), we should keep the installation of zf1 in the lib/Zend directory and I've no idea on how to make that happen.

As @Flyingmana said, the release downloadables builder should be fixed (but I think that's easier than the first part).

fballiano:
I did a quick test with zf1-future 1.21.1 and worked great!

ADDISON74:
Even though a package like PHP 7.0 has been the official EOL for some time there are Linux distributions where patches for it continue to be provided. For example Debian 9 Stretch. The last update of PHP 7.0 if I'm not mistaken was in January 2022. But you must not lose sight of the fact that Debian 9 has LTS until mid-2022.

If I had Magento1 and switched to OM20 I would not agree to maintain a PHP 5.6 or 7-0-7.3 but I would opt for a solution at this time that would ensure me 3-5 years in a secure configuration. I would be practically obliged to install a minimum of PHP 7.4 I would not mind this aspect at all. This is about the fact that many do not want changes. Right, do not change what works but adapt if you want to use it for a long time. For Magento1/OM they have already knocked on the door for a long time PHP 8.1, MySQL 8.0.

As for ZF1-Future I agree with this change and many others as long as they don't "hurt". It is good there are parts of Magento 1 that can still be maintained by communities and we can benefit the work. There are a few PRs that change hundreds or over a thousand files. Maybe in their case the approvals should be up to 5 - 10 people until a merged is made.

luigifab:
Some days/weeks ago I tried to remove `lib/Zend` and to `composer require shardj/zf1-future`.
This is working, **but** files in `app/code/core/Zend` are ignored.

fballiano:
> Some days/weeks ago I tried to remove `lib/Zend` and to `composer require shardj/zf1-future`. This is working, **but** files in `app/code/core/Zend` are ignored.

I'd prefer if the composer install could somehow install the files in lib/Zend so that nothing really changes :-\

kiatng:
> I'd prefer if the composer install could somehow install the files in lib/Zend so that nothing really changes :-\

Does it mean that for those who do not use composer, there won't be any update to Zend?

fballiano:
> Does it mean that for those who do not use composer, there won't be any update to Zend?

the release zip/tar should be exactly like the ones we have now, with zend (updated) and everything.

fballiano:
we've to remove zf1 from our repo instead of just replacing it with zf1-future. I love the idea of this PR but I voto for closing it and working in that other direction instead, like what I'm trying to do with the new release builder.

ADDISON74:
Before removing ZF1 from OpenMage shouldn't we make a repository to keep the latest version? Who knows, maybe someone will need it and can download it at any time.

PR #1851, opened 2021-10-05
Fix Core Bug: To make "01/01/1970" a valid birthday

1Shiroyuuki1:
When parsing "01/01/1970" as date to setDate() the output is '0'.
Because default value of strtoftime() is "01/01/1970".
```
    public function setDate($date)
    {
        $this->setTime($date ? strtotime($date) : false);
        $this->setData('date', $date);
        return $this;
    }
```

### Description (*)
According to this bug means that Magento 1.X.X never allowed a birthday of "01/01/1970".
A simple null check in the getter will prevent it.

1. https://www.tehplayground.com/auf4jhP1T5thJDtK
2. Text form:
```
<?php
// example code

//setDate() simulation
$customerBirthday = '01/01/1970';
$isDate             = $customerBirthday ? strtotime($customerBirthday) : false;

//getDate() simulation
$isDate ? print('Yeaj 01/01/1970 is valid') : print('but it´s not...');

//fix
($isDate !== null) ? print('Yeaj 01/01/1970 is valid') : print('but it´s not...');

?>
```

 - [Having Issues] Add yourself to contributors list

### DIFF
--- a/app/code/core/Mage/Customer/Block/Widget/Dob.php
+++ b/app/code/core/Mage/Customer/Block/Widget/Dob.php
@@ -77,7 +77,7 @@ class Mage_Customer_Block_Widget_Dob extends Mage_Customer_Block_Widget_Abstract
      */
     public function getDay()
     {
-        return $this->getTime() ? date('d', $this->getTime()) : '';
+        return ($this->getTime() !== null) ? date('d', $this->getTime()) : '';
     }
 
     /**
@@ -85,7 +85,7 @@ class Mage_Customer_Block_Widget_Dob extends Mage_Customer_Block_Widget_Abstract
      */
     public function getMonth()
     {
-        return $this->getTime() ? date('m', $this->getTime()) : '';
+        return ($this->getTime() !== null) ? date('m', $this->getTime()) : '';
     }
 
     /**
@@ -93,7 +93,7 @@ class Mage_Customer_Block_Widget_Dob extends Mage_Customer_Block_Widget_Abstract
      */
     public function getYear()
     {
-        return $this->getTime() ? date('Y', $this->getTime()) : '';
+        return ($this->getTime() !== null) ? date('Y', $this->getTime()) : '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kiatng (approved):
I tested setting the dob to 01/01/1970 in both frontend and backend. Actually, the dob is set correctly in both cases. However, in frontend, the dob was empty even though its value is 01/01/1970. This PR fixes it.

image

elidrissidev (approved):

ADDISON74 (dismissed):

1Shiroyuuki1:
I have fixed a following bug after this fix.
@kiatng strtotime returns via default a value of "01/01/1970". If you create a new customer and leave birthday input empty you will get this value as default birthday.

Steps to Reproduce:
(Frontend)
1. Create a new Customer
2. Leave Birthday input Empty
3. Save
4. Try to edit this customer again and you will see the default value of strtotime()

Fix: I have improved the settime() and getTime() method and replaced strtotime with a Datetime. There is absolutely no reason to parse a unix timestamp for a birthday date. 
Since strtotime was very often used in php 4.x.x i think adobe never cares to improve their Date handling.
With us now supporting PHP 7++ we should consider to clean up this as well.

Awaiting new Review :)

fballiano (approved):

app/code/core/Mage/Customer/Block/Widget/Dob.php
```diff
@@ -62,38 +62,54 @@ public function isRequired()
     }
 
     /**
-     * @param string $date
+     * @param $date
```
kiatng, line 65:
```suggestion
     * @param string $date
```

kiatng (changes requested):

kiatng:
@1Shiroyuuki1 I tested your commit and it worked as expected. One last request from me:

https://github.com/OpenMage/magento-lts/blob/c48b3669ebc3c73edaf3965707abcd8bc1beac15/app/code/core/Mage/Customer/Block/Widget/Dob.php#L28-L28

The docblock needs to be updated:

```php
  * @method DateTime getTime() 
 ```

Please do another commit.

1Shiroyuuki1:
@kiatng will do it this evening 👍

kiatng (approved):

fballiano (approved):

PR #1852, opened 2021-10-05
make fulltext search separator (OR/AND) configurable from backend

sreichel:
### Description (*)

https://magento.stackexchange.com/a/84312/46249

> The biggest inexplicably nonsensical move by Magento is to use "OR" search instead of "AND". If you search "red cars" you will get everything red (including cars, dogs, forks, mountainsides, etc), and every kind of car (including red, blue, green, yellow, etc). With "AND", you get only items that contain "red" AND "car", which is how search should work!

Just added a config option to change from OR to AND.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ Most important changes will be listed here, all other changes since `19.4.0` can
 - `admin/design/use_legacy_theme`
 - `admin/emails/admin_notification_email_template`
 - `catalog/product_image/progressive_threshold`
+- `catalog/search/search_separator`
 
 ### New Events
 - `adminhtml_block_widget_form_init_form_values_after`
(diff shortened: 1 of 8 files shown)


ADDISON74:
Nice work Sven. I am glad to see you posting again.

fballiano (approved):
tested and working, nice PR and needed feature!

fballiano:
@ADDISON74 would you add a code review?

Flyingmana (approved):

luigifab:
I wake up today, but new configuration values use beginning and ending space.
I dislike 😢. Not sure but perhaps this is the first configuration value with required space.

fballiano:
hehehe you're the master of syntax, it makes sense tho and we can always change it :-)

Flyingmana:
personally I also think the config without spaces would be better, and adding them where the config is applied.
Still possible to change, its merged but not inside a release yet.

sreichel:
Whats wrong with that spaces?

Just tried to make it as simple as possible.

luigifab:
I think the fate of these spaces is to be trimmed.

PR #1862, opened 2021-10-15
Fix broken admin theme

Imaginaerum:
Fix broken admin theme when design/theme/layout config is not NULL

### DIFF
--- a/app/code/core/Mage/Core/Model/Layout/Update.php
+++ b/app/code/core/Mage/Core/Model/Layout/Update.php
@@ -492,6 +492,9 @@ class Mage_Core_Model_Layout_Update
         if (null === $storeId) {
             $storeId = Mage::app()->getStore()->getId();
         }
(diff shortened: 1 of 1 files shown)


elidrissidev:
Can you please elaborate on this issue and the fix you applied?

Imaginaerum:
When you change theme layout config, admin theme is broken.

image

Result without patch is:

image

kkrieger85 (approved):

elidrissidev:
Upon reviewing older pending PRs I found that this PR is a duplicate of #1216 which seems to be addressing the same issue. Sorry.

ADDISON74:
@elidrissidev - From what I checked you closed this PR and made a reference to a reported issue. There is indeed PR #1229, but it is a draft. It has to be seen which of them solves the issue. This one here had an approval.

elidrissidev:
I mistakenly linked to the issue instead of the PR. I will review the first PR this weekend since I've also been dealing with this issue for a long time.

PR #1872, opened 2021-10-31
Improve getScheme on Zend_Controller_Request_Http

mmenozzi:
### Description (*)
Some web servers (for example the [local web server of the Symfony CLI](https://symfony.com/doc/current/setup/symfony_server.html#getting-started)) puts the value `On` (and not `on`)  in the `$_SERVER['HTTPS']` super global.

With this PR the `Zend_Controller_Request_Http::getScheme` properly returns the HTTPS scheme in such situations.

1. Install a clean copy of OpenMage as usual
2. Install Symfony CLI and its [SSL certificates](https://symfony.com/doc/current/setup/symfony_server.html#enabling-tls)
3. Start the Symfony CLI local web server with `symfony server:start`
4. Configure OpenMage to use the `https://127.0.0.1:8000/` base URL for both secure and unsecure
5. Try to open the home page at https://127.0.0.1:8000/
6. Without this PR you'll get "Too many redirects" error, with this PR you'll properly get the home page

### DIFF
--- a/app/code/core/Zend/Controller/Request/Http.php
+++ b/app/code/core/Zend/Controller/Request/Http.php
@@ -1034,7 +1034,7 @@ class Zend_Controller_Request_Http extends Zend_Controller_Request_Abstract
      */
     public function getScheme()
     {
-        return (strtolower($this->getServer('HTTPS')) == 'on') || $this->getServer('HTTP_X_FORWARDED_PROTO') == 'https' ?
+        return (strtolower((string)$this->getServer('HTTPS')) == 'on') || $this->getServer('HTTP_X_FORWARDED_PROTO') == 'https' ?
             self::SCHEME_HTTPS :
             self::SCHEME_HTTP;
     }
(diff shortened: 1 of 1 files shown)


elidrissidev (dismissed):

sreichel (dismissed):

ADDISON74 (dismissed):

luigifab (commented):
With PHP 8.1, if `$this->getServer('HTTPS')` return null, it will crash.
`strtolower(null)`: `PHP Deprecated:  strtolower(): Passing null to parameter #1 ($string) of type string is deprecated`

Not sure if this case can happen.

Flyingmana (changes requested):
strtolower requires a string, but we could get a null here as of https://github.com/OpenMage/magento-lts/blob/b6cc7b24b078b721d6daa191a40bf5b1324e72df/app/code/core/Zend/Controller/Request/Http.php#L354

Flyingmana (approved):

elidrissidev (approved):

PR #1874, opened 2021-11-03
Fix attribute model has data changes was true right after loading

woutersamaey:
### Description (*)
Immediately after loading an attribute, it already has the "has data changes" flag set to true, even though it was just loaded and could not possibly have been changed yet.

1. Saving attributes does faster if they are not really changed.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Resource/Entity/Attribute.php
+++ b/app/code/core/Mage/Eav/Model/Resource/Entity/Attribute.php
@@ -521,7 +521,7 @@ class Mage_Eav_Model_Resource_Entity_Attribute extends Mage_Core_Model_Resource_
                 $object->addData($result);
             }
         }
-
+        $object->setDataChanges(false);
         return $this;
     }
 


elidrissidev (dismissed):

kiatng:
https://github.com/OpenMage/magento-lts/blob/8b9bfae16c35f2740b7e066864ee07a5db64218a/app/code/core/Mage/Core/Model/Abstract.php#L234-L242

As you can see in line 240, `$this->_hasDataChanges = false;`, so this PR is redundant?

elidrissidev:
@kiatng But not in all cases, at least not with this snippet:
```php
$attribute = Mage::getModel('eav/entity_attribute')->loadByCode('catalog_product', 'color');
echo $attribute->hasDataChanges(); // true
```

woutersamaey:
The issue was indeed after a model was loaded via `loadByCode()`

kkrieger85 (approved):

fballiano:
Isn't it possible, with this implementation, that somebody hooks into an event which runs between load and afterLoad, changes the data, and then the afterload reset the hasDataChange?

Wouldn't it be better to search for why the hasDataChange is marked as true although it should be false, instead of forcing it to false?

fballiano:
@woutersamaey could you check my last comment? sorry to bother.

fballiano (changes requested):
please provide a code sample to reproduce the problem, I'm blocking the merge since it's unclear

elidrissidev:
> please provide a code sample to reproduce the problem

@fballiano Check my reply above, it includes a snippet to reproduce this issue. 

Also, you are right in that this change might override the `hasChanges` flag if a code actually modifies the model in `afterLoad` event, but the same thing applies to the core `load()` method:

https://github.com/OpenMage/magento-lts/blob/1641eab640cd1254aa01aea3bd69723e4eec96e1/app/code/core/Mage/Core/Model/Abstract.php#L237-L245

As for the reason the `hasDataChanges` flag is set to `true` directly after loading the model, it's simply because the method `loadByCode` in the model is missing the line that initializes `$this->_hasDataChanges` to false after loading, in fact it's even missing the code to fire the `beforeSave` event and set `origData`!

https://github.com/OpenMage/magento-lts/blob/1641eab640cd1254aa01aea3bd69723e4eec96e1/app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php#L134-L152

app/code/core/Mage/Eav/Model/Resource/Entity/Attribute.php
```diff
@@ -521,7 +521,7 @@ protected function _afterLoad(Mage_Core_Model_Abstract $object)
                 $object->addData($result);
             }
         }
-
+        $object->setDataChanges(false);
```
elidrissidev, line 524: This change should preferably be done in `Mage_Eav_Model_Entity_Attribute_Abstract::loadByCode`. You may also `setOrigData` since it seems it was also forgotten.

```diff
    public function loadByCode($entityType, $code)
    {
        Varien_Profiler::start('_LOAD_ATTRIBUTE_BY_CODE__');
        if (is_numeric($entityType)) {
            $entityTypeId = $entityType;
        } elseif (is_string($entityType)) {
            $entityType = Mage::getModel('eav/entity_type')->loadByCode($entityType);
        }
        if ($entityType instanceof Mage_Eav_Model_Entity_Type) {
            $entityTypeId = $entityType->getId();
        }
        if (empty($entityTypeId)) {
            throw Mage::exception('Mage_Eav', Mage::helper('eav')->__('Invalid entity supplied.'));
        }
        $this->_getResource()->loadByCode($this, $entityTypeId, $code);
        $this->_afterLoad();
+       $this->setOrigData(); 
+       $this->_hasDataChanges = false;
        Varien_Profiler::stop('_LOAD_ATTRIBUTE_BY_CODE__');
        return $this;
    }
```

PR #1837, opened 2021-09-27
Add phpstan configuration file

tmotyl:
This PR provides basic configuration file to run phpstan on openMage code base.
Right now it just whitelists few core modules to show that it's working.
allowing more modules requires validation and probably some fixes, so should be tackled in separate PR's.

TODO:
Help is needed in providing a github action which will run phpstan on every PR.

### Description (*)
Provides basic configuration for running phpstan

### Fixed Issues (if relevant)

Fixes: 
1. Fixes https://github.com/OpenMage/magento-lts/issues/365

### Manual testing scenarios (*)

1. pull
2. composer install
3. run  `vendor/bin/phpstan analyze -c .github/phpstan.neon`

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [ ] Add yourself to contributors list

PR #1875, opened 2021-11-07
Add phpstan configuration, dependencies and github action

Flyingmana:
* takes changes from #1837
* Adds phpstan as direct dev dependency
* executes phpstan as part of our github actions


Flyingmana:
merging as biggest part was already approved in the linked branch
and mostly only additional logic for the workflows was added

PR #1888, opened 2021-12-01
Fix infinite loop caused by #1467

drashmk:
### Description (*)
The infinite loop happens when you try to export collection that has number of items divisible by 1000(`$this->_exportPageSize`).

In this case the Varien collection will reset the page number to `lastPageNumber` when `$page` is higher.
`$page` is higher than `lastPageNumber` because the current logic assume that there is always one more page if the current page count is the same as page export size which is not true in case when count is divisible by page size.

### Related Pull Requests

https://github.com/OpenMage/magento-lts/pull/1467

1. Create sales rule without coupon code
2. Generate 1000 coupons
3. Export them as CSV

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -969,7 +969,6 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
     public function _exportIterateCollection($callback, array $args)
     {
         $originalCollection = $this->getCollection();
-        $count = null;
         $page  = 1;
 
         do {
@@ -978,7 +977,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
             $collection->setCurPage($page);
             $collection->load();
 
-            $count = $collection->count();
+            $lastPageNumber = $collection->getLastPageNumber();
 
             $page++;
 
@@ -987,7 +986,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
             }
             $collection->clear();
             unset($collection);
-        } while($count == $this->_exportPageSize);
+        } while ($page <= $lastPageNumber);
     }
 
     /**


tmotyl:
@midlan plese take a look at this patch, as this is a patch to your change.
cc @Flyingmana @colinmollenhour @Sekiphp

midlan:
Nice find @drashmk !

@tmotyl 
Calling getLastPageNumber will ruin performance since it is calling database `COUNT(*)` inside. I would stick just with:

```
} while ($count <= $this->_exportPageSize);
```

If it fixes the issue.

drashmk:
Unfortunately it won't solve the issue. 
`getLastPageNumber()` is still called on the collection during sql query generation when `OFFSET` needs to be added if `curPage` is more than 1. https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Varien/Data/Collection.php#L212

midlan:
I looked into it. To keep optimum performance the only solution is to avoid limiting the collection by its internal paging, rather use limit directly on select object. Method will look like this:

```php
    public function _exportIterateCollection($callback, array $args)
    {
        $originalCollection = $this->getCollection();
        $page  = 1;

        do {
            $collection = clone $originalCollection;

            //reset collection internal paging
            $collection->setCurPage(1);
            $collection->setPageSize(false);

            //use select paging to avoid counting collection
            $collection->getSelect()->limitPage($page, $this->_exportPageSize);

                call_user_func_array(array($this, $callback), array_merge(array($item), $args));
            }

            $page++;
            $count = $collection->count();

            $collection->clear();
            unset($collection);
        } while($count == $this->_exportPageSize);
    }
```
Kinda hack solution, not respecting return type of `Mage_Adminhtml_Block_Widget_Grid::getCollection` which is `Varien_Data_Collection` that is lack of `getSelect()` method.

I can improve the code above with such condition:

`if ($collection instanceof Varien_Data_Collection_Db)`

to respect return type of `Mage_Adminhtml_Block_Widget_Grid::getCollection`. What do you think other openmage developers @Flyingmana @Sekiphp @tmotyl ?

drashmk:
@midlan that should prevent calling `getLastPageNumber()` and yes it is hacky solution. 
IMO the improvement should be applied in the Varien Collection class rather that here so all collections would benefit from it, not just when exporting.

henryhayes:
> IMO the improvement should be applied in the Varien Collection class rather that here so all collections would benefit from it, not just when exporting.

+1. I agree in principle.

However, we still don't have any meaningful PHPUnit tests for the Varien classes; to ensure backwards compatability. And therefore need to be very careful about modifying the Varien classes.

midlan:
I agree, I know about this bottleneck method. I am author of the fix that it is not calling `getLastPageNumber` at least on first page #852

 Changing `getCurPage` behavior might break many things.
 
 For now, we have bug in `_exportIterateCollection` method. Not sure how limiting is this bug for magento other openmage users. We can:
 1. Use code I posted above
 2. Keep it open and close it when `getCurPage` is fixed
 3. Rollback it to previous state

drashmk:
@midlan after checking I can see that `getLastPageNumber ` does not call php `count` on the collection items, it is making count sql call as you can see here: https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Varien/Data/Collection/Db.php#L221

midlan:
@drashmk Yes I know, thats the performance issue I wanted avoid by #1467 . It passed my unit tests, but I used Varien_Data_Collection as mock, but there is different implementation of `getLastPageNumber` and `getSize` methods.\

What about the fix? In fact we can merge your code as the performance would be same. Or I can improve the code I posted above.

drashmk:
@midlan IMO in most cases I've encountered the collections are `Varien_Data_Collection_Db` and `getLastPageNumber ` and `getSize` are optimized, The hacky solution is not needed as the performance will stay the same.

midlan:
I strongly disagree. `Varien_Data_Collection_Db::getSize` performance is pretty bad in many collections. E.g. my catalog size is 500k products. Counting such large collections is really slow.

drashmk:
@midlan I agree that the performance is not the best, but that is not the aim of this PR. 
At the moment you can't export collection with 1k, 2k, 10k, 500k,... items because it goes into infinite loop. This PR fixes that issue.

midlan:
I would rather go and rollback #1467 to previous code that was in magento core for years.

tmotyl:
Please prepare a patch to revert the change. Then the potential performance impr can be discussed separately

midlan:
Revert PR #1888 opened

PR #1889, opened 2021-12-02
fix bug with applying catalog price rules due to wrong timestamp

ioweb-gr:
### Description (*)
This is a bug that's been annoying us for quite some time. We always had to patch it in installations using catalog price rules because our country keeps changing time due to daylight saving.
It seems magento is not taking into account the locale and because the rule runs at midnight, it doesn't apply our catalog prices.

The issue is documented very well here as well as the solution 

https://stackoverflow.com/questions/25280095/magento-catalog-price-rule-disappears-at-night

1. You need to have an installation on a server where your server and website are using a timezone like Athens and a few catalog rules. The rules won't be applied at midnight


elidrissidev:
If this is bug, it should probably target the `1.9.4.x` and then it can be merged to `20.0` but I'll leave the decision to the maintainers. Can you also please set your git credentials and amend your commit so that it gets proper authorship?

ioweb-gr:
OK I amended the commit and force pushed to my fork. It should reflect here

luigifab (approved):

fballiano:
Hi @ioweb-gr, could you switch the branch to 1.9.4.x please? I'd approve the PR right away. Thank you!

fballiano:
Closing since it became duplicate of https://github.com/OpenMage/magento-lts/pull/2189

PR #1890, opened 2021-12-03
Fix incompatible declaration error

holozaen:
Fix incompatible declaration in method items of Mage_Catalog_Model_Product_Option_Api_V2. 

### Comments
Declaration of this method in parent class Mage_Catalog_Model_Product_Option_Api allows the parameter $store to be null. Php 8.1 throws a PHP Fatal error "Declaration of Mage_Catalog_Model_Product_Option_Api_V2::items($productId, $store) must be compatible with Mage_Catalog_Model_Product_Option_Api::items($productId, $store = null)."

This fix has no side effects because the child method passes the $store parameter directly on to the parent method which explicitly allowes null as the $store parameter value.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Option/Api/V2.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Option/Api/V2.php
@@ -69,7 +69,7 @@ class Mage_Catalog_Model_Product_Option_Api_V2 extends Mage_Catalog_Model_Produc
      * @param int|string|null $store
      * @return array
      */
-    public function items($productId, $store)
+    public function items($productId, $store = null)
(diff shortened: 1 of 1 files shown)


luigifab (approved):

Sdfendor (approved):

elidrissidev (approved):
LGTM. You may wanna configure your email with git and amend your commit before the PR is merged so that it's properly attributed to you.

holozaen:
@elidrissidev 
Thank you...I just tried this but it looks like I messed up things... I am a complete noob in this area....sorry.

elidrissidev (approved):

PR #1891, opened 2021-12-03
PHP 8.1 Fix deprecation notice in mcryptcompat lib

Sdfendor:
### Description (*)

Open Mage contains a mcrypt polyfill to provide extension functionality after PHP 7.1 removed the extension. 
This polyfill contains a class extending [php_user_filter](https://www.php.net/manual/en/class.php-user-filter.php), an internal class. 

PHP 8.1 introduced *tentative return types* for internal classes that will throw a deprecation notice instead of fail in case an extending child class omits a return type set in the internal parent class (In PHP 9.0 those notices will become errors). 
As a consequence executing an arbitrary request in OpenMage results in two deprecation notices: 
```
Deprecated: Return type of phpseclib_mcrypt_filter::filter($in, $out, &$consumed, $closing) should either be compatible with php_user_filter::filter($in, $out, &$consumed, bool $closing): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
Deprecated: Return type of phpseclib_mcrypt_filter::onCreate() should either be compatible with php_user_filter::onCreate(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice 
```

This PR sets the return type of those two methods as they are defined internally. Because OpenMage requires at least PHP 7.0 those return types have no implications regarding backwards compatibility and it's unnecessary to use the mentioned workaround via attribute.

### DIFF
--- a/lib/mcryptcompat/mcrypt.php
+++ b/lib/mcryptcompat/mcrypt.php
@@ -1087,7 +1087,8 @@ if (!function_exists('phpseclib_mcrypt_list_algorithms')) {
          * @return int
          * @access public
          */
-        public function filter($in, $out, &$consumed, $closing): int
+        #[\ReturnTypeWillChange]
+        public function filter($in, $out, &$consumed, $closing)
         {
             $newlen = 0;
             $block_mode = phpseclib_mcrypt_module_is_block_mode($this->cipher->mcrypt_mode);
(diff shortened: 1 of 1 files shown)


Flyingmana:
please use the same solution the original took, see https://github.com/phpseclib/mcrypt_compat/issues/32

Sdfendor:
I can do that. But by doing so it will be a temporary solution in regards to PHP 9.0. It is however - like you mentioned - the original's route so I will change it accordingly.

Flyingmana (approved):

PR #1898, opened 2021-12-14
Revert #1355 for transactional email

luigifab:
This PR revert #1355.
And it can be also applied to 1.9.4.x.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Email/Template.php
+++ b/app/code/core/Mage/Adminhtml/Model/Email/Template.php
@@ -51,7 +51,9 @@ class Mage_Adminhtml_Model_Email_Template extends Mage_Core_Model_Email_Template
         if (!$templateCode) {
             return array();
         }
+
         $paths = array();
+        Mage::getSingleton('adminhtml/config')->getSections();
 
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Adminhtml/Model/Email/Template.php
```diff
@@ -80,13 +82,13 @@ public function getSystemConfigPathsWhereUsedCurrently()
         if (!$templateId) {
             return array();
         }
-        $paths = array();
 
+        $paths = array();
```
ADDISON74, line 86: There is another $paths at line 55. Are you sure this should be two times in a file?

luigifab: Not sure what do you mean, but at line 55 it's another method.

ADDISON74: You are right. I did not check the whole file.

elidrissidev (approved):

kiatng (approved):

ADDISON74 (approved):

Flyingmana (approved):

PR #1896, opened 2021-12-12
Bugfix #1885 - Add to cart only accepts 1 item

gdemarsico:
### Fixed Issues

1. Fixes OpenMage/magento-lts#1885

PR #1901, opened 2021-12-16
Fix syntax in lib/Zend/Locale/Data/es_419.xml

elidrissidev:
Fixes a syntax error introduced in OpenMage/magento-lts#1896 as spotted by @luigifab 

### Related Pull Requests
OpenMage/magento-lts#1896

### DIFF
--- a/lib/Zend/Locale/Data/es_419.xml
+++ b/lib/Zend/Locale/Data/es_419.xml
@@ -194,7 +194,7 @@ For terms of use, see http://www.unicode.org/copyright.html
 			<decimalFormatLength>
 				<decimalFormat>
 					<pattern>#,##0.###</pattern>
-				</decimalFormat>>
+				</decimalFormat>
 			</decimalFormatLength>
 			<decimalFormatLength type="short">
 				<decimalFormat>


ADDISON74 (approved):

luigifab (approved):

Flyingmana (approved):

Flyingmana:
anyone knows if this is causing an actual error in any case?
I would like to add a check for it.

elidrissidev:
It's weird because I was only able to reproduce the original issue after changing the locale to `Spanish (Colombia)`, but not `Spanish (Argentina)`, and even the merged fix doesn't fix it for me, cleared caches and everything. Has anyone tested this fix or am I missing something...

luigifab:
For me, by changing products quantity in cart, it does not work without initial PR for:
- es_CL Spanish Chile
- es_CO Spanish Colombia
- es_CR Spanish Costa Rica
- es_PA Spanish Panama
- es_PE Spanish Peru
- es_VE Spanish Venezuela

luigifab:
With the initial fixed PR, and after removed `/tmp/zend_cache---*`, it works for all es locales.
I checked _es_419.xml_ but not (yet) tested _Zend/Locale/Data/*.xml_ files from [zf1-future](https://github.com/Shardj/zf1-future/blob/master/library/Zend/Locale/Data/es_419.xml), the proposed solution is not present.

luigifab:
I using this _.git/hooks/pre-commit_ hook to check syntax of php and phtml and xml files for `git commit`:

```bash
#!/usr/bin/env bash
# sudo aptitude install libxml2-utils
 
PROJECT=$(php -r "echo dirname(dirname(dirname(realpath('$0'))));")
STAGED_FILES_CMD=$(git diff --cached --name-only --diff-filter=ACMR HEAD | grep '\.php\|\.phtml\|\.xml' )

# Determine if a file list is passed
if [ "$#" -eq 1 ]
then
	oIFS=$IFS
	IFS='
	'
	SFILES="$1"
	IFS=$oIFS
fi
SFILES=${SFILES:-$STAGED_FILES_CMD}

if [[ $SFILES ]]
then
	echo "Checking PHP/PHTML/XML..."
	for FILE in $SFILES
	do
		if [[ $FILE =~ ".xml" ]]
		then
			xmllint --noout $PROJECT/$FILE
		else
			php -l $PROJECT/$FILE
		fi

		if [ $? != 0 ]
		then
			echo "STOP"
			exit 1
		fi
		FILES="$FILES $PROJECT/$FILE"
	done
fi

exit $?
```

It's not perfect, and yes this is only local.

elidrissidev:
@luigifab does the additional `>` cause an error or not?

luigifab:
No there are no errors, even if I forgot to close a tag in any line of _es_419.xml_.
I suppose if the XML is invalid, it is ignored.

ADDISON74:
Ideally all XML files should have tags that close. Otherwise it is not known what could come out of the concatenation of all the files. Most likely Magento ignores the error although we should have known about it. I will do a test in some XML files to see what happens. If anyone knows where to add some code to log the error in a particular XML file it would be perfect.

fballiano (approved):

Sekiphp (approved):

Flyingmana:
I created #1918 as followup to add a detection for this specific error, and xml syntax errors in general.

PR #1902, opened 2021-12-16
Allows to make creditmemo on payment type "free"

spinsch:
# Description

With this PR allows the merchant to create a creditmemo on orders with payment method "free". 

We ourselves require this option when an order is completely covered by a voucher code and the payment method "Free" is used. It is mainly so that we can cancel items and release them for another customer in the inventory as well as for the administrative overview.

# Manual testing scenarios

a) Activate "Free" Payment Method

1. Goto Backend > System > Configuration > [SALES] Payment Methods > Zero Subtotal Checkout
2. Set Enabled => Yes
3. Set Automatically Invoice All Items => Yes
4. Save

b) Create a Coupon Code

1. Goto Promotions > Shopping Cart Price Rules > Add New Rule
2. Fill all the form to use with your own code
3. Set under Actions Apply => Fixed amount discount for whole cart
4. Set Discount Amount => [amount to trigger the "free" payment method]

c) Make an order as customer with single item

1. Add a product to the cart
2. Use your coupon code
3. Check if payment method "free" is using
4. Complete the order

d) Make an order as customer with multiple item

1. Add a product to the cart
2. Add another product to the cart and set amount to 2
2. Use your coupon code
3. Check if payment method "free" is using
4. Complete the order

e) Make creditmemo with the single item order

f) Make creditmemo with the multiple item order

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -751,7 +751,8 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
          * for this we have additional diapason for 0
          * TotalPaid - contains amount, that were not rounded.
          */
-        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001) {
+        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001
+            && $this->getPayment()->getMethod() !== 'free') {
             return false;
         }
 
--- a/app/code/core/Mage/Sales/Model/Order.php
+++ b/app/code/core/Mage/Sales/Model/Order.php
@@ -735,7 +735,11 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
     public function canCreditmemo()
     {
         if ($this->hasForcedCanCreditmemo()) {
-            return $this->getForcedCanCreditmemo();
+            foreach ($this->getAllItems() as $item) {
+                if ($item->canRefund()) {
+                    return $this->getForcedCanCreditmemo();
+                }
+            }
         }
 
         if ($this->canUnhold() || $this->isPaymentReview()) {
@@ -751,8 +755,7 @@ class Mage_Sales_Model_Order extends Mage_Sales_Model_Abstract
          * for this we have additional diapason for 0
          * TotalPaid - contains amount, that were not rounded.
          */
-        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001
-            && $this->getPayment()->getMethod() !== 'free') {
+        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001) {
             return false;
         }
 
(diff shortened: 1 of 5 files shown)


spinsch:
i forget one file in my commit so i made a force push.

fballiano (approved):

elidrissidev:
It seems that I'm able to refund the free order multiple times, the "Credit Memo" button is always shown, though only one creditmemo is ever created. Not sure if this is the intended behavior or is it?

Screen Shot 2021-12-22 at 12 49 08 PM

app/code/core/Mage/Sales/Model/Order.php
```diff
@@ -751,7 +751,8 @@ public function canCreditmemo()
          * for this we have additional diapason for 0
          * TotalPaid - contains amount, that were not rounded.
          */
-        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001) {
+        if (abs($this->getStore()->roundPrice($this->getTotalPaid()) - $this->getTotalRefunded()) < .0001
+            && $this->getPayment()->getMethod() !== 'free') {
```
fballiano, line 755: This line causes the problem highlighted by @elidrissidev. I think it should be remove and the previous line should be changed. Instead of checking for TotalPaid and TotalRefunded we should count how many non-refunded items are still in the sales_flat_order_item, what do you think?

fballiano (changes requested):

spinsch:
I have found code fragments in my research that suggest that this feature was already worked on but never finished.

I found the flag "forced can creditmemo" in the order class. 
```php
 * @method bool hasForcedCanCreditmemo()
 * @method bool getForcedCanCreditmemo()
 * @method $this setForcedCanCreditmemo(bool $value)
```

This is updated in the payment module via an observer.
```php
class Mage_Payment_Model_Observer
{
    /**
     * Set forced canCreditmemo flag
     *
     * @param Varien_Event_Observer $observer
     * @return $this
     */
    public function salesOrderBeforeSave($observer)
    {
        /** @var Mage_Sales_Model_Order $order */
        $order = $observer->getEvent()->getOrder();

        if ($order->getPayment()->getMethodInstance()->getCode() != 'free') {
            return $this;
        }

        if ($order->canUnhold()) {
            return $this;
        }

        if ($order->isCanceled() || $order->getState() === Mage_Sales_Model_Order::STATE_CLOSED) {
            return $this;
        }

        /**
         * Allow forced creditmemo just in case if it wasn't defined before
         */
        if (!$order->hasForcedCanCreditmemo()) {
            $order->setForcedCanCreditmemo(true);
        }
        return $this;
    }
```

The problem with this, it is never saved to the database so this attribute is useless.

I have now changed my PR so that the original idea of this code works.

ADDISON74 (approved):

fballiano:
there are also conflicts to solve, temporary moving to draft

fballiano:
@elidrissidev you checked this PR in https://github.com/OpenMage/magento-lts/pull/1902#issuecomment-999518855, do you think it should be saved or should we close it?

NOTE: it has conflicts and needs rebase...

spinsch:
hi @fballiano 
sorry i didn't respond. for some reason i had deactivated the notification. can you please open this pull request again so i can finalise it? thank you and sorry again for not responding to your messages.

fballiano:
@spinsch for some reason I can't reopen it, probably because the 1.9.4.x branch was deleted in the meanwhile, the only way it to create a new one, thanks!

PR #1915, opened 2021-12-21
Prevent string * int multiplification issue in PHP 8.0+

lamskoy:
### Description (*)

I faced with fatal errors while upgrading native magento 1.9.x to PHP8.0 locally:

After debugging I found the issue 1:

`echo '' * 1;`

gives:

`Fatal error: Uncaught TypeError: Unsupported operand types: string * int`

Issue 2:

`round('')`

gives:

`Fatal error: Uncaught TypeError: round(): Argument #1 ($num) must be of type int|float, string given`

So patched core files with missing methods which should be casted to float.

Wanna share fix with you. Hope this helps to make whole code better.

This is potential issue with PHP8, to be checked with auto tests

### Questions or comments

This stuff gives small performance improvement coz we get rid of slow **preg_replace**, 2 **substr** and 1 **strtolower** calls after changes. Coz we use _getData() instead of magic method __call now

Varien_Object:
```
   protected function _underscore($name)
    {
...
        $result = strtolower(preg_replace('/(.)([A-Z])/', "$1_$2", $name));
  ...
    }

    public function __call($method, $args)
    {
        switch (substr($method, 0, 3)) {
            case 'get' :
...
                $key = $this->_underscore(substr($method,3));
                $data = $this->getData($key, isset($args[0]) ? $args[0] : null);
```

So on quote calculation we can save much calls to preg_match (even if count that _underscore caches calls)
If there are N calls to different methods getBlabla during quote recalculation, we save
N * preg_replace, N*2 substr, N strtolower calls
Not bad if N > 20 :)


sreichel:
Should we use type hinging for newly introduced methods?

lamskoy:
> Should we use type hinging for newly introduced methods?

You mean this
`public function setSomething(float $value)`
`public function getSomething(): float`

instead of this?

`public function setSomething(float $value)`
`public function getSomething()`

luigifab:
Interesting changes!

I'm sorry, but you have forgotten a new empty line here before the comment:
image

lamskoy:
> Interesting changes!
> 
> I'm sorry, but you have forgotten a new empty line here before the comment: image

Here you go with updated empty lines

sreichel:
> If yes, then we can have issues with 3rd party extensions overriding getSomething()

I'm pretty sure we will have issues ... but most of M1-extensions are not maintained anymore. IMHO ... i'd prefer this and patch/fix extensions to match method declaration.

Nevertheless +1 for the PR.

kiatng:
Does it make sense to fix the type at the source, make PDO returns numeric types from numeric columns? Ref https://write.corbpie.com/how-to-return-integers-and-floats-from-mysql-with-php-pdo/ and https://stackoverflow.com/questions/1197005/how-to-get-numeric-types-from-mysql-using-pdo

lamskoy:
> Does it make sense to fix the type at the source, make PDO returns numeric types from numeric columns? Ref https://write.corbpie.com/how-to-return-integers-and-floats-from-mysql-with-php-pdo/ and https://stackoverflow.com/questions/1197005/how-to-get-numeric-types-from-mysql-using-pdo

Some of fields are set/get dynamically by observers/tax calculators, think there's no big sense to make PDO returns in this case

Flyingmana:
with this amount of new methods I wonder if its not easier, to extend the getData, setData logic to force types for specific fields instead of overwriting the methods. So in the end we just have a configurable array.

Flyingmana (approved):

tmotyl:
Dedicated methods have a big advantage of being static analisys friendly. So im against of putting more logic in magic get data

fballiano (approved):

fballiano:
PHPStan is telling us that the modification to app/code/core/Mage/Wishlist/controllers/IndexController.php has some problem:

Call to an undefined method Mage_Wishlist_IndexController::_getData().

Schermata 2022-05-10 alle 11 02 40

we've to fix it before release

constancyGmbH:
Hi @fballiano , after applying this release, all of my prices drop to zero in cart and Checkout. In Catalog and Product view everything is fine.
Any Idea why?

constancyGmbH:
I broke it down to the File /app/code/core/Mage/Sales/Model/Quote/Item/Abstract.php and here the function 
```
/**
     * @return $this
     */
    public function setCalculationPrice($value)
    {
        return $this->setData('calculation_price', (float) $value);
    }
```

If I comment this function out, it works and Prices are there. I'm not sure why.

fballiano:
Thank you very much @constancyGmbH, I'll try to take a look at this ASAP

constancyGmbH:
I think it has something to do with setting this value to NULL.
with this workaround it is working, but I'm not sure if this will work in PHP 8:

```
/**
     * @return $this
     */
    public function setCalculationPrice($value)
    {

        if(is_null($value)){
            return $this->setData('calculation_price', $value);
        }else{
            return $this->setData('calculation_price', (float) $value);
        }
    }
```

fballiano:
mmmm it doesn't happen to me on a vanilla openmage installation :-\

constancyGmbH:
ok, so maybe this is a Probleme with a custom module or an adjustment on our Magento installation....

constancyGmbH:
Have you tried your Code in PHP 7 oder 8 (or Both)?

fballiano:
> ok, so maybe this is a Probleme with a custom module or an adjustment on our Magento installation....

that's what i was thinking, I tried it on php7.4

kiatng:
I haven't tested the finding of @constancyGmbH, but I highly suspect it is related to https://github.com/OpenMage/magento-lts/blob/dfb6fe83d51d304d33b7c9d16195440d882faa89/app/code/core/Mage/Sales/Model/Quote/Item/Abstract.php#L609-L613.

This PR added a bunch of setters: https://github.com/OpenMage/magento-lts/blob/dfb6fe83d51d304d33b7c9d16195440d882faa89/app/code/core/Mage/Sales/Model/Quote/Item/Abstract.php#L1114-L1477

Just eyeballing without testing, we know that `setField(null)` is another way to unset the field `unsField()`, which is not the same as `setField(0)`, which is what the bunch of methods do above.

I do not know how best to address the issue, but may be something like:

```php
// Field is generic field name.
function setField($value)
{
    if ($value === null) {
        unset('field');
    } else {
        $this->setData('field', (float) $field);
   }
   return $this;
}
```

kiatng:
I think this is reason the price is zero in the cart:

https://github.com/OpenMage/magento-lts/blob/dfb6fe83d51d304d33b7c9d16195440d882faa89/app/code/core/Mage/Sales/Model/Quote/Item/Abstract.php#L434-L446

`$price` above is zero always when the setters value is `null`. So, another fix is change line 437 to `if (empty($price)) {`.

kiatng:
+1 to revert this PR.

PHP 8:
```php
echo '1' + 1; // OK, result int(2)
echo null + 1; //OK, result int(1)
echo '1.1' + 1; //OK, result float(2.1)
echo round('1.1'); //OK, result float(1)
echo '' + 1; // Not OK, PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string + int 
echo round(''); // Not OK, PHP Fatal error:  Uncaught TypeError: round(): Argument #1 ($num) must be of type int|float, string given
```
However, getter method returns `null` or numeric string (for number type) and should not return empty string. Is there a case where a getter return an empty string for field of type int or float? Also, prior to this PR, OM seems to work just fine with PHP 8.

fballiano:
I've created a PR to revert this one, altough I can't merge it without 2 other maintainers to approve

PR #1916, opened 2021-12-21
Config cache for system.xml loader: Speed improvement

lamskoy:
### Description (*)
System.xml loading is not cached anywhere. So each time you open backend page in Magento it scans *ALL* system.xml inside all modules, merges them and doesn't cache this big XML structure.

Header menu is rendered from system.xml contents too, not only config fields and sections!

My PR implements caching of this stuff.

It decreases any admin page load time in about 1.5 sec on second load on my local machine :)

Open any magento config page twice and measure time of page load

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Config.php
+++ b/app/code/core/Mage/Adminhtml/Model/Config.php
@@ -34,6 +34,15 @@
  */
 class Mage_Adminhtml_Model_Config extends Varien_Simplexml_Config
 {
+    /**
+     * @var string
+     */
+    protected $_cacheId = 'mage_adminhtml_config_system_xml';
+
+    /**
+     * @var Mage_Core_Model_Config_Base
+     */
+    protected $_config;
 
     /**
      * Enter description here...
@@ -81,15 +90,61 @@ class Mage_Adminhtml_Model_Config extends Varien_Simplexml_Config
     }
 
     /**
+     * @param array $params
+     */
+    public function __construct(array $params = array())
+    {
+        $this->_cacheChecksum = null;
+        $this->setCache(Mage::app()->getCache());
+        $this->setCacheTags([Mage_Core_Model_Config::CACHE_TAG]);
+        if (!$this->loadCache()) {
+            /** @var Mage_Core_Model_Config_Base $config */
+            $this->_config = Mage::getConfig()->loadModulesConfiguration('system.xml')
+                ->applyExtends();
+            $this->saveCache();
+        }
+    }
+
+    /**
+     * @param $tags
+     * @return $this|Mage_Adminhtml_Model_Config
+     */
+    public function saveCache($tags=null)
+    {
+        if ($this->getCacheSaved()) {
+            return $this;
+        }
+        if (is_null($tags)) {
+            $tags = $this->_cacheTags;
+        }
+        $xmlString = $this->_config->getXmlString();
+        $this->_saveCache($xmlString, $this->getCacheId(), $tags, $this->getCacheLifetime());
+        $this->setCacheSaved(true);
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function loadCache()
+    {
+        $xmlString = $this->_loadCache($this->getCacheId());
+        $class = Mage::getConfig()->getModelClassName('core/config_base');
+        $this->_config = new $class();
+        if ($this->_config->loadString($xmlString)) {
+            return true;
+        }
+        return false;
+    }
+
+    /**
      * Init modules configuration
      *
      * @return void
      */
     protected function _initSectionsAndTabs()
     {
-        $config = Mage::getConfig()->loadModulesConfiguration('system.xml')
-            ->applyExtends();
-
+        $config = $this->_config;
         Mage::dispatchEvent('adminhtml_init_system_config', array('config' => $config));
         $this->_sections = $config->getNode('sections');
         $this->_tabs = $config->getNode('tabs');


luigifab:
From 550/650 ms to 350/450 ms at localhost with PHP 7.4 when only CONFIG cache is enabled.
From ~1700 ms to 1350/1450 ms at localhost with PHP 7.4 when all cache are disabled.

luigifab:
I think that this error is due to this PR (exception.log):
```
Exception: "$string" parameter for simplexml_load_string is not a string in
  lib/Varien/Simplexml/Config.php:517
```

Flyingmana:
> From ~1700 ms to 1350/1450 ms at localhost with PHP 7.4 when all cache are disabled.

sounds like there is still a check missing, to not cache it when the cache is disabled?

elidrissidev:
Agree with @Flyingmana, I don't see any code that checks if the cache is enabled. Example from `Mage_Api_Model_Config`:
https://github.com/OpenMage/magento-lts/blob/4e7a8e134ab62f5b5532bec50890e3abf3809b6a/app/code/core/Mage/Api/Model/Config.php#L58-L62

lamskoy:
> Agree with @Flyingmana, I don't see any code that checks if the cache is enabled. Example from `Mage_Api_Model_Config`:
> 
> https://github.com/OpenMage/magento-lts/blob/4e7a8e134ab62f5b5532bec50890e3abf3809b6a/app/code/core/Mage/Api/Model/Config.php#L58-L70

Yeah, you're right. Pushed new commit with fix. Bound to 'config' cache

elidrissidev:
Since you're including it with the `CONFIG` tag, can you also update the description to include `system.xml`? Something like:
```diff
-System(config.xml, local.xml) and modules configuration files(config.xml).
+System(config.xml, local.xml) and modules configuration files(config.xml, system.xml).
```

Screen Shot 2021-12-31 at 12 59 57 PM

app/code/core/Mage/Adminhtml/Model/Config.php
```diff
@@ -80,16 +89,67 @@ public function getTabs()
...
+        $this->setCacheTags([Mage_Core_Model_Config::CACHE_TAG]);
+        $usesCache = Mage::app()->useCache('config');
+        if (!$usesCache || !$this->loadCache()) {
+            /** @var Mage_Core_Model_Config_Base $config */
+            $this->_config = Mage::getConfig()->loadModulesConfiguration('system.xml')
+                ->applyExtends();
+            if ($usesCache) {
+                $this->saveCache();
+            }
+        }
```
elidrissidev, line 108: Missing call to parent constructor

elidrissidev (changes requested):

app/code/core/Mage/Adminhtml/Model/Config.php
lamskoy: We don't need call to Varien_Simplexml_Config as it loads and parses string inside object itself.
We use external config object. That's a trick

lamskoy:
> 

Done

app/code/core/Mage/Core/etc/config.xml
```diff
@@ -122,7 +122,7 @@
             <types>
                 <config translate="label,description" module="core">
                     <label>Configuration</label>
-                    <description>System(config.xml, local.xml) and modules configuration files(config.xml).</description>
+                    <description>System(config.xml, local.xml) and modules configuration files(config.xml, system.xml).</description>
```
luigifab, line 125: I suggest to add the missing spaces:
`System (config.xml, local.xml) and modules configuration files (config.xml, system.xml).`

luigifab: :'(

lamskoy: Will you merge my stuff? :)

luigifab: I can't, but someone perhaps.
I will test it again... with the missing spaces? :)

lamskoy: Added whitespaces. Welcome to merge

luigifab:
I still gets the following error (I didn't searched why):
```
2022-01-28T11:09:27+00:00 ERR (3): 
Exception: "$string" parameter for simplexml_load_string is not a string in lib/Varien/Simplexml/Config.php:517
Stack trace:
#0 app/code/core/Mage/Adminhtml/Model/Config.php(138): Varien_Simplexml_Config->loadString()
#1 app/code/core/Mage/Adminhtml/Model/Config.php(101): Mage_Adminhtml_Model_Config->loadCache()
#2 app/code/core/Mage/Core/Model/Config.php(1380): Mage_Adminhtml_Model_Config->__construct()
#3 app/Mage.php(532): Mage_Core_Model_Config->getModelInstance()
#4 app/Mage.php(546): Mage::getModel()
#5 app/code/core/Mage/Adminhtml/controllers/System/ConfigController.php(86): Mage::getSingleton()
#6 app/code/core/Mage/Core/Controller/Varien/Action.php(437): Mage_Adminhtml_System_ConfigController->editAction()
#7 app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(262): Mage_Core_Controller_Varien_Action->dispatch()
#8 app/code/core/Mage/Core/Controller/Varien/Front.php(192): Mage_Core_Controller_Varien_Router_Standard->match()
#9 app/code/core/Mage/Core/Model/App.php(381): Mage_Core_Controller_Varien_Front->dispatch()
#10 app/Mage.php(754): Mage_Core_Model_App->run()
#11 /var/www/cellublue/public/index.php(44): Mage::run()
#12 {main}
```

After clearing cache and after refresh any configuration pages.

tmotyl:
I think this chnage deservces an comment in readme/changelog, as it changes an expected behaviour developers have -> now they will need to clear the cache after changing system.xml

app/code/core/Mage/Core/etc/config.xml
elidrissidev: I just realized that the description is translated, so changing it will require changing the translation as well, and that'll break language packs.

app/code/core/Mage/Adminhtml/Model/Config.php
elidrissidev: Feel free to mark this as resolved if you're sure about it.

app/code/core/Mage/Adminhtml/Model/Config.php
```diff
@@ -80,16 +89,67 @@ public function getTabs()
...
+    /**
+     * @return bool
+     */
+    public function loadCache()
+    {
+        $xmlString = $this->_loadCache($this->getCacheId());
+        $class = Mage::getConfig()->getModelClassName('core/config_base');
+        $this->_config = new $class();
+        libxml_use_internal_errors(true);
+        if ($this->_config->loadString($xmlString)) {
```
elidrissidev, line 138: Like @luigifab mentioned, the call to `loadString` fails when the cache is empty: `"$string" parameter for simplexml_load_string is not a string`. You'll have to handle that here and load an empty config xml string otherwise (`<?xml version="1.0"?><config></config>`).

elidrissidev (changes requested):

app/code/core/Mage/Core/etc/config.xml
luigifab: It seems that the original text is missing into _Mage_Core.csv_. Can you add the new one into the CSV?

app/code/core/Mage/Adminhtml/Model/Config.php
luigifab: or with `!empty() && ...` ?

fballiano: @lamskoy could you fix this last comment so that we can approve the PR? Thank you!

lamskoy: > @lamskoy could you fix this last comment so that we can approve the PR? Thank you!
A bit later. Have war in my country - Ukraine

fballiano: @lamskoy sorry, I didn't know.

luigifab: I added a check for `$xmlString`.

app/code/core/Mage/Core/etc/config.xml
luigifab: Text added.

luigifab (approved):

fballiano:
is phpstan right about this?

Schermata 2022-06-07 alle 17 03 35

elidrissidev:
Yes, because that parameter is not passed to the parent class.

fballiano (dismissed):

app/code/core/Mage/Adminhtml/Model/Config.php
```diff
@@ -80,16 +89,64 @@ public function getTabs()
...
+            $this->_config = Mage::getConfig()->loadModulesConfiguration('system.xml')
+                ->applyExtends();
+            if ($usesCache) {
+                $this->saveCache();
+            }
+        }
+    }
+
+    /**
+     * @param $tags
```
elidrissidev, line 109:
```suggestion
     * @param array|null $tags
```

elidrissidev (commented):
Two more changes to please phpstan

fballiano (dismissed):

fballiano:
@elidrissidev added info to README ;-)

elidrissidev (approved):
Tested once again and everything works well for me, and also a big improvement in loading speed across the Admin area!

PR #1923, opened 2021-12-25
Change default sort in Manage Attributes grid

ADDISON74:
This changes the default sort order in Manage Attributes grid from column 1 (Attribute Code) to column 2 (Attribute Label)


ADDISON74:
I will close this PR and I will include one more modification.

PR #1925, opened 2021-12-25
Sort ascending CMS Block list in Categories

ADDISON74:
This PR fixes the issue reported here #1678.


fballiano (approved):

sreichel:
Can't add review ...

```
POSThttps://github.com/OpenMage/magento-lts/pull/1925/review_comment/create
[HTTP/2 500 Internal Server Error 321ms
```

sreichel (changes requested):
```suggstion
                ->setOrder('title', Varien_Data_Collection::SORT_ORDER_ASC)
```

ADDISON74:
@sreichel - The proposal that you made by using Varien_Data_Collection :: SORT_ORDER_ASC instead of 'ASC' I did not find it in the OpenMage code. Through proposed PR I sought not to have any difference of approach between what exists and what I propose. Eventually both variants produce the same result.

Flyingmana (approved):

Issue #1584, opened 2021-04-30
Product Alerts Unsubscribe - Wrong Messages


ADDISON74:
Simulate Product Alerts. When you receive the email click the links to unsubscribe for a product and for all products and check the messages. My personal opinion the messages wrong. 

Bellow I fixed them (the modified lines have [* !!! *] at the end).

**file: /app/code/local/Mage/ProductAlert/controllers/UnsubscribeController.php
line: 57**
```
        $session    = Mage::getSingleton('catalog/session');

        /* @var $session Mage_Catalog_Model_Session */
        $product = Mage::getModel('catalog/product')->load($productId);
        if (!$product->getId() || !$product->isVisibleInCatalog()) {
            /* @var $product Mage_Catalog_Model_Product */
            Mage::getSingleton('customer/session')->addError($this->__('The product was not found.')); [* !!! *]
            $this->_redirect('customer/account/');
            return ;
        }

        try {
            $model  = Mage::getModel('productalert/price')
                ->setCustomerId(Mage::getSingleton('customer/session')->getCustomerId())
                ->setProductId($product->getId())
                ->setWebsiteId(Mage::app()->getStore()->getWebsiteId())
                ->loadByParam();
            if ($model->getId()) {
                $model->delete();
            }

            $session->addSuccess($this->__('You will no longer receive price alerts for this product.')); [* !!! *]
        }
        catch (Exception $e) {
            $session->addException($e, $this->__('Unable to update the alert subscription.'));
        }
        $this->_redirectUrl($product->getProductUrl());
    }

    public function priceAllAction()
    {
        $session = Mage::getSingleton('customer/session');
        /* @var $session Mage_Customer_Model_Session */

        try {
            Mage::getModel('productalert/price')->deleteCustomer(
                $session->getCustomerId(),
                Mage::app()->getStore()->getWebsiteId()
            );
            $session->addSuccess($this->__('You will no longer receive price alerts.')); [* !!! *]
        }
```

colinmollenhour:
HI @ADDISON74 thanks for your contributions! However, your chances of getting a thorough review and merge would be greatly increased by submitting a PR instead of an issue.

ADDISON74:
@colinmollenhour it would be good for 2-3 people to check this issue before and if it is confirmed then I will create a commit.

fballiano:
hey @ADDISON74, I agree with @colinmollenhour, the problem is that it's really difficult to check the differences with the previous code this way :-(

[...]

PR #1927, opened 2021-12-25
Price Alerts unsubscribe - Wrong messages 

ADDISON74:
Wrong messages when you click to unsubscribe from receiving price alerts for that product/all products.

The issue was discussed here #1584.

### DIFF
--- a/app/code/core/Mage/ImportExport/Block/Adminhtml/Export/Filter.php
+++ b/app/code/core/Mage/ImportExport/Block/Adminhtml/Export/Filter.php
@@ -55,7 +55,7 @@ class Mage_ImportExport_Block_Adminhtml_Export_Filter extends Mage_Adminhtml_Blo
 
         $this->setRowClickCallback(null);
         $this->setId('export_filter_grid');
-        $this->setDefaultSort('attribute_code');
+        $this->setDefaultSort('frontend_label');
         $this->setDefaultDir('ASC');
         $this->setPagerVisibility(false);
         $this->setDefaultLimit(null);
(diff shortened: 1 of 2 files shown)


fballiano:
is the modification to app/code/core/Mage/ImportExport/Block/Adminhtml/Export/Filter.php really needed for this PR?

ADDISON74:
You are right. By mistake it got caught from another PR that I posted before. The approval of this PR will not affect the code in any way because that change in export file is also in another PR from today.

fballiano:
yes but that PR may not be approved, would it be possible to change this PR removing that modification?

sorry!!

ADDISON74:
Today I worked far away from the office directly on GitHub. I deleted that branch so I can't make any changes in this PR. If a maintainer can do the change it will be perfect, Otherwise I will close this PR next week and post it again. I could change its description as I mentioned before it solves another reported issue which has already been posted in another PR.

ADDISON74:
The change in Filter.php file is also in this PR #1928. The approval of the two PR's will keep the changes.

elidrissidev:
@ADDISON74 Tip: when you make a commit directly from GitHub, there is an option to create it in a new branch, that way you separate your work.

elidrissidev:
Fun fact: even GitHub errors out when trying to review this PR, I think because of the deleted branch.

elidrissidev:
Can you update the translation of the lines you modified? I'm unable to create a review for some reason.

ADDISON74:
I will close this PR and create a new one which will correct all issues, including translations if the strings are missing.

fballiano:
actually, I've been thinking about it and it's probably not a good idea to change the messages in the source code, it would break every language pack already out there, and we can't fix those.

so in my opinion it would be much better to leave the source as it is and "translate" those messages via language CSV file, in this case https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/locale/en_US/Mage_ProductAlert.csv

ADDISON74:
Don't worry. I've fixed the issue. The correct PR is here #1929. 

I only added one line in Mage_ProductAlert.csv related to price. For stock there was one already. It seems the Magento team missed it.

PR #1928, opened 2021-12-25
Change the default sort column - Export page

ADDISON74:
This PR changes the default sort column on the Export page in backend from **Attribute Code** to **Attribute Label**.

Based on this change it is so easy to check those rows you want to skip from export.

sort_export


Flyingmana:
how does this interact with different language?

ADDISON74:
No worries about. It does not affect the local language anyway. It simply changes the default sort column as mentioned in the description. 

It is a visual improvement and a real help when you want to skip some rows from the export file. Instead of clicking on the top of the Attribute Label column to sort ASC or DESC you already have it sorted.

luigifab (changes requested):
I think you need to remove `'sortable' => false,` for `frontend_label` on line 362 of the same file.
Because with the change I got:

image

PR #1929, opened 2021-12-26
Price Alerts unsubscribe - Wrong messages

ADDISON74:
Wrong messages when you click to unsubscribe from receiving price alerts for that product/all products. It adds also a missing one line in translation too.

The issue was discussed here #1584.

### DIFF
--- a/app/code/core/Mage/ProductAlert/controllers/UnsubscribeController.php
+++ b/app/code/core/Mage/ProductAlert/controllers/UnsubscribeController.php
@@ -60,7 +60,7 @@ class Mage_ProductAlert_UnsubscribeController extends Mage_Core_Controller_Front
         $product = Mage::getModel('catalog/product')->load($productId);
         if (!$product->getId() || !$product->isVisibleInCatalog()) {
             /* @var Mage_Catalog_Model_Product $product */
-            Mage::getSingleton('customer/session')->addError($this->__('The product is not found.'));
+            Mage::getSingleton('customer/session')->addError($this->__('The product was not found.'));
             $this->_redirect('customer/account/');
             return ;
         }
@@ -75,7 +75,7 @@ class Mage_ProductAlert_UnsubscribeController extends Mage_Core_Controller_Front
                 $model->delete();
             }
 
-            $session->addSuccess($this->__('The alert subscription has been deleted.'));
+            $session->addSuccess($this->__('You will no longer receive price alerts for this product.'));
         } catch (Exception $e) {
             $session->addException($e, $this->__('Unable to update the alert subscription.'));
         }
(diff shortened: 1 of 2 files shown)


fballiano:
IMHO we shouldn't change the PHP files and change only the CSV, with the PR as it is we'd break the language packs.

ADDISON74:
@fballiano - Test it and you will see that there is nothing wrong with this PR. Basically for the two actions you do in the email you receive about the price change, when you click on the links on the store page, the messages are not the correct ones, but simply reversed. Practically if you already have a translation it will not be affected in any way because it is assumed that they already exist. The only translation to be done is for "You will no longer receive price alerts." This string was missed from the file by the Magento team. For the stock alerts it exists.

I agree that this line should be translated, but it is assumed that those who manage OpenMage stores and update them regularly to new version should ensure proper operation. There are PR's in which new strings have been added and I don't think it has been a problem so far with translations.

fballiano:
no man i dont agree, we can't break all yhr translation packs just to
change a couple of messages, instead of changing the php, "translate" the
messages in the en_us and, for all other languages, everything will still
continue to work as before

Il Dom 26 Dic 2021, 12:45 ADDISON ***@***.***> ha scritto:

ADDISON74:
OK man in this case I will close this comment.

fballiano:
you can do the same stuff in the csv... you already modified one...

Il Dom 26 Dic 2021, 15:42 ADDISON ***@***.***> ha scritto:

ADDISON74:
@fballiano - Half a year ago I reported an issue here #1584 related to the confusing messages that a Magento store displays when in the received Price Alert email an user clicks on a link to delete the notification on that product or for all of them. You asked me to create a PR that would solve this issue and I did it these days.

You are welcome to propose your own PR as you wish. I closed my proposal but the issue is still there.

fballiano:
please don't take it personally, but i don't want to have my italian customers with broken messages when it's not necessary, the change can be made in the csv -> same result, no problem, so i don't understand why you g
ot upset.

ADDISON74:
Do not worry about. I am a professional person and I do not take anything personal. In my case in English in order to fix this issue I have to change two strings in Mage_PriceAlerts.csv as follows:

```
"The alert subscription has been deleted.","You will no longer receive price alerts for this product."
"You will no longer receive price alerts for this product.","You will no longer receive price alerts."
```

My opinion that if you are Italian or English the change must be done. And then why not do it directly in the PHP file because that's where the confusion comes? It is obvious that this is the source of the issue due to a programming bug.

Honestly speaking I don't like to have different strings in English translations. Just see above how ugly they look like. However, let's try to see a little beyond the convenience of changing something.

PR #1930, opened 2021-12-26
Better messages for product alerts

fballiano:
In PR https://github.com/OpenMage/magento-lts/pull/1929 was proposed a change to some messages that made sense but I think they should be done in CSV instead of the PHP file, to avoid breaking the languages pack already out there.

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/1929
https://github.com/OpenMage/magento-lts/issues/1584


fballiano:
@ADDISON74 I created my version of the PR, the only thing I didn't modify is the "You will no longer receive stock alerts for this product." message cause I think that "for this product" seems to be needed (you're not unsubscribing from every alert, I think).

ADDISON74:
Thank you for your effort but on this PR I already have my own implementation. 

I don't like having two different strings in English in the same row. I have had this discussion with someone else in the past and I still maintain my opinion. If it is English then it should be English and nothing else.

fballiano:
that is exactly the purpose why we have an en_US CSV, to be able to change translations without breaking all the languages pack.

ADDISON74:
The change bellow is missing for PR to be complete:

```
"You will no longer receive price alerts for this product.","You will no longer receive price alerts."
```

If you do a test you will see why I claimed that the messages displayed by Magento are confusing. I made the change in the PHP file because it's simpler and I don't want to leave source code with logic errors.

ADDISON74:
Let's analyze a little bit what is happening when there is a subscription for a price alert.

**1. The customer receives the message as bellow.** 

email_received

There are two important links in the message content: 

**Click here not to receive alerts for this product** having this URL
https://mydomain.tld/index.php/productalert/unsubscribe/price/product/456/?___store=admin

**Unsubscribe from all price alerts** having this URL
https://mydomain.tld/index.php/productalert/unsubscribe/priceAll/?___store=admin

**2. Click here not to receive alerts for this product**. Here is what we get in frontend after clicking on this link (after you log into the customer's account)

alert_product

My comment: The message is not clear. It should be "You will no longer receive price alerts for this products." or "The alert subscription for this product has been deleted.". I prefer the first one.

**3. Unsubscribe from all price alerts**. Here is what we get in frontend

alert_allproducts

My comment:  This time the message is completely wrong. It should be "You will no longer receive price alerts." However, we talked about all the products where I subscribed to notifications.

**4. If the product is not in the database**. Here is what we get in frontend

alert_product_not_found

My comment: Here we are dealing with a faulty translation into English. It should be "The product was not found" or "You don't have a subscription for this product."

My initial PR that I closed solves these issue in the controller and in English translation. I keep the option to change directly in these files and not a translation of an English sentence into another sentence as proposed by this PR. There is a logic bug in that controller PHP file.

elidrissidev:
@ADDISON74 Your points are completely valid, the current lines are very confusing. However, @fballiano as I understood is coming from backwards-compatibility point of view, even though it's a rather small change it's still breaking. Now whether it's worth it or not that's a debatable thing 😄. Personally, I'm not a fan of translation keys that don't match their value. It's a shame we don't have control over the language packs.

Flyingmana (approved):

luigifab (approved):
Because #1929 is closed, I approve #1930.

elidrissidev (approved):
Not a big fan of un-matching keys and values, but because it's a small change I approve.

PR #1941, opened 2021-12-30
Fix #1924 non-existing field

### LINKED ISSUE #1924: block_identifier issue in Widget Chooser
In this file app/code/core/Mage/Adminhtml/Block/Cms/Block/Widget/Chooser.php **block_identifier** doesn't exist. It should be **block_id**. 

```
class Mage_Adminhtml_Block_Cms_Block_Widget_Chooser extends Mage_Adminhtml_Block_Widget_Grid
{
    /**
     * Block construction, prepare grid params
     *
     * @param array $arguments Object data
     */
    public function __construct($arguments=array())
    {
        parent::__construct($arguments);
        $this->setDefaultSort('block_identifier');
        $this->setDefaultDir('ASC');
        $this->setUseAjax(true);
        $this->setDefaultFilter(array('chooser_is_active' => '1'));

### PR DESCRIPTION
**Pull request by woutersamaey**
### Description (*)
Fixes #1924

1. Fixes OpenMage/magento-lts#1924

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Cms/Block/Widget/Chooser.php
+++ b/app/code/core/Mage/Adminhtml/Block/Cms/Block/Widget/Chooser.php
@@ -41,7 +41,7 @@ class Mage_Adminhtml_Block_Cms_Block_Widget_Chooser extends Mage_Adminhtml_Block
     public function __construct($arguments=array())
     {
         parent::__construct($arguments);
-        $this->setDefaultSort('block_identifier');
+        $this->setDefaultSort('block_id');
         $this->setDefaultDir('ASC');
         $this->setUseAjax(true);
         $this->setDefaultFilter(array('chooser_is_active' => '1'));

### CODE REVIEW

**Review by ADDISON74 — APPROVED** [review]

**Review by elidrissidev — APPROVED** [review]

File: app/code/core/Mage/Adminhtml/Block/Cms/Block/Widget/Chooser.php [hunk 1]
```diff
@@ -41,7 +41,7 @@ class Mage_Adminhtml_Block_Cms_Block_Widget_Chooser extends Mage_Adminhtml_Block
     public function __construct($arguments=array())
     {
         parent::__construct($arguments);
-        $this->setDefaultSort('block_identifier');
+        $this->setDefaultSort('block_id');
```
Flyingmana, line 44: Iam sure this should be 'identifier'

Flyingmana (changes requested):

app/code/core/Mage/Adminhtml/Block/Cms/Block/Widget/Chooser.php
ADDISON74: Clearly 'block_identifier' does not exist as a column in the database or in Magento source code . If we want to fix this issue in a first phase we can correct the error to 'block_id' which exists as column. 

If in time we decide to sort by default using the 'identifier' column we can change it but we should do some tests before. After it is merged I will review this in the future and I will give you my feedback between 'block_id' and 'title' I prefer the last one which is the easiest to find and understand. The block number or its identifier. It doesn't help me at all. I always click to sort by the meaning of the block.

Flyingmana: this is going onto the table `cms_block`, right?
so we have 
```

block_id | smallint(6) AI PK
title | varchar(255)
identifier | varchar(255)
content | mediumtext
creation_time | timestamp
update_time | timestamp
is_active | smallint(6)
```

ADDISON74: This is correct. In the future I will propose to do the default sorting by **title**. No one wants to see a table sorted by number or code.

Flyingmana: I would, in part because there is an existing index on identifier, but also because as a Developer I usually do not care about the title, only the identifier.
But maybe its better to make this generally configurable somehow.

ADDISON74: Not worth it. I'll come up with some screenshots when we have to determine which column the default sort is. I would say to correct only the error for the moment merging it and remaining at the block ID.

Flyingmana (approved):

luigifab (changes requested):
Can you update sorting direction (`DESC`)? To get the latest first?

luigifab (approved):

kiatng (approved):

ADDISON74 (approved):

PR #1946, opened 2022-01-02
adding total variable to available variables for order email templates

f1-outsourcing:
### Description (*)

I am not entirely sure why I had to add this in the past. I think that currency formatting is not available in the order email templates, that is why I put this in the variable.
I am using the total to highlight what needs to be paid more at the top of the email without any other amounts close to it. So eg. blind people do not accidentally pay without VAT.

1. ...
2. ...


luigifab (approved):

kiatng (approved):

fballiano:
mmm I thought I commented this already (but actually I didn't because the comment is not here) but isn't it as easy to just use {{order.getGrandTotal()}} in the template?

kiatng:
> mmm I thought I commented this already (but actually I didn't because the comment is not here) but isn't it as easy to just use {{order.getGrandTotal()}} in the template?

Yes, it can. But that doesn't have currency formatting.

fballiano:
right, but I still don't know about adding just one formatted grand total, what about the other properties of the order?

for example, in another template I've found this:

`{{var order.billing_address.format('html')}}`

which would be more correct as an approach, or maybe something like `{{ var order.formatPriceTxt('grand_total') }}`

ADDISON74:
This PR has 2 green approves. Convert it to draft if you have any doubts.

kiatng:
> which would be more correct as an approach, or maybe something like `{{ var order.formatPriceTxt('grand_total') }}`

From this:
https://github.com/OpenMage/magento-lts/blob/bd0cb12cb1c084e714e8e282f8d4f6a5431bb089/app/code/core/Mage/Sales/Model/Order.php#L2014-L2023

To this?
```php
    /**
     * Retrieve text formated price value includeing order rate
     *
     * @param   float|string $price
     * @return  string
     */
    public function formatPriceTxt($price)
    {
        $price = (float) (is_numeric($price) ? $price : $this->_getData($price));
        return $this->getOrderCurrency()->formatTxt($price);
    }
```

fballiano:
@kiatng :-) that's great! yes I meant something like that, this way it's more general and it can be used for other prices too, what do you think?

PR #1835, opened 2021-09-25
Use short array syntax

luigifab:
### Description

I found this great tool: [php-short-array-syntax-converter](https://github.com/thomasbachem/php-short-array-syntax-converter)
So I converted to short array syntax some directories with:

```
find app/ dev/ errors/ shell/ lib/Varien/ lib/Magento/ lib/Mage/ lib/Zend/ lib/Unserialize/ ./*.php -name "*.php" -exec php convert.php -w "{}" \;
find app/ dev/ errors/ shell/ lib/Varien/ lib/Magento/ lib/Mage/ lib/Zend/ lib/Unserialize/ -name "*.phtml" -exec php convert.php -w "{}" \;
```

_I tested this tool with [zf1-future](https://github.com/Shardj/zf1-future), already converted to short array syntax.
I reverted library/Zend/ directory to traditional syntax, then I converted to short array syntax, result: no differences._

_I added our lib/Zend directory to reduce number of changes of #1832._

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1960, opened 2022-01-05
Fix count(): Parameter must be an array or an object that implements Countable

Sekiphp:
### Description (*)
Unfortunatelly I found new bug in my favourite obscure method `getAttributeRawValue`. 

When `fetchRow` called there `$attributesData = $adapter->fetchRow($select, array('entity_id' => $entityId))` returned `false` method ends with `E_WARNING`: `count(): Parameter must be an array or an object that implements Countable` on line 694
image

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
@@ -634,7 +634,7 @@ abstract class Mage_Catalog_Model_Resource_Abstract extends Mage_Eav_Model_Entit
         if ($staticAttributes) {
             $select = $adapter->select()->from($staticTable, $staticAttributes)
                 ->where($this->getEntityIdField() . ' = :entity_id');
-            $attributesData = $adapter->fetchRow($select, array('entity_id' => $entityId));
+            $attributesData = $adapter->fetchRow($select, array('entity_id' => $entityId)) ?: [];
         }
 
         /**


kiatng (approved):

LukasCesal (approved):

ADDISON74 (approved):

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -634,7 +634,7 @@ public function getAttributeRawValue($entityId, $attribute, $store)
         if ($staticAttributes) {
             $select = $adapter->select()->from($staticTable, $staticAttributes)
                 ->where($this->getEntityIdField() . ' = :entity_id');
-            $attributesData = $adapter->fetchRow($select, array('entity_id' => $entityId));
+            $attributesData = $adapter->fetchRow($select, array('entity_id' => $entityId)) ?: [];
```
midlan, line 637: using old array syntax `array` and new `[]` on the same line looks weird

luigifab: [ads]But #1835 will spank you if you use old array syntax.[/ads]

midlan: I'm not telling it is wrong or good, just that it looks weird with hope there may exists instructions which syntax to use :)

Flyingmana (approved):

PR #1269, opened 2020-10-19
Fix Varien_Io_File ignored errors with error control operator

luigifab:
### Description

This fix some (perhaps not all) `Varien_Io_File` ignored errors with error control operator.
- `fclose supplied resource is not a valid stream resource`
- `chdir:  No such file or directory (errno 2)`
- `chmod: No such file or directory`

I hope it's good.
OpenMage 20.0.10 - 20.0.13 / PHP 7.4.6 and 8.0.5 - 8.0.14

### Manual testing scenarios
- Install OpenMage
- Install https://github.com/getsentry/magento-amg-sentry-extension
- Configure the installed module, go to System → Configuration → Advanced → Developer → AMG Sentry
- Go to a CMS page or a CMS block, and add a new directory, and upload an image
- Go to your Sentry instance, and read errors:
image

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #1963, opened 2022-01-06
Fixed error in Varien_Io_File if $this->_iwd is null

woutersamaey:
### Description (*)

The following code breaks because `$this->_iwd` is `NULL` and the line `chdir($this->_iwd)` cannot work.

```
$io = new Varien_Io_File();
$io->write('/tmp/test.txt', 'body');
```

This happens because `$io->open()` was never called, and it shouldn't really. The use case is valid as `Varien_Io_File` is the class to use for all disk IO. Problem is: sometimes Magento uses it to present a single file, somethimes Magento uses it in a generic way.

I want to use it in a generic way too, but my 2 lines break, so I propose this simple fix.

### Questions or comments
I am using `Varien_Io_File` for all disk IO stuff, but I cannot because it breaks.

### DIFF
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -557,7 +557,7 @@ class Varien_Io_File extends Varien_Io_Abstract
     {
         @chdir($this->_cwd);
         $result = @file_put_contents($filename, $src);
-        chdir($this->_iwd);
+        @chdir($this->_iwd);
 
         return $result;
     }


luigifab:
New achievement unlocked: #200th pending PR!
I also did something many months ago in #1269.

woutersamaey:
Then let's merge and get on with it pls.

fballiano:
I'll close this one cause it's a duplicate of https://github.com/OpenMage/magento-lts/pull/1269 (which is more complete IMHO)

Issue #1965, opened 2022-01-07
My product reviews are not my but all


luigifab:
From My account / My product reviews, I can access all product reviews, even if the current customer account is not the author of the review.

This is OpenMage 20.0.13 with PHP 8.0.

**Screenshot of the review when customer account is the author:**
![image](https://user-images.githubusercontent.com/31816829/148529536-9e9a32f7-09b3-4a59-a9d5-1cbd5d764a6b.png)

**Screenshot of the review when customer account is NOT the author:**
![image](https://user-images.githubusercontent.com/31816829/148529670-1e86bbd4-4dcb-4264-9d31-9e1bf7792d21.png)

This is not critical, but any customer can get all reviews of any Magento 1 and OpenMage stores by trying all ids from the current url.

elidrissidev:
Can't reproduce this in v1.9.4.x nor v20.0, here is the snippet where the data is coming from and I can only reproduce it if I remove the customer filter line
https://github.com/OpenMage/magento-lts/blob/ff6189ae82f141c13224318241be64d512555c0d/app/code/core/Mage/Review/Block/Customer/List.php#L49-L53

luigifab:
Yes, but to to the detail page, not the list.

elidrissidev:
Do you have to type the URL manually?

luigifab:
Yes, change the id from the url with any existing review.

aterjung:
I am not shure if it is a real issue, cause the reviews are open to everyone via the regular frontend controller:

review/product/view/id/xxx/

but i would prevent it anyway. Probably someone extends the page to edit the review etc. and its no harm.

I have prepared a commit to fix this. But it is a bit bigger. I would expect it to respond with a 404 error - thrown in the controller. The actual review model is loaded in the block class where it should not. This way its too late to throw a 404 error. Bad design :-( 

Intrestlingly its done better in the normal frontend controller.  So i changed the customer review controller and block to match that and added the customer check:

https://github.com/OpenMage/magento-lts/commit/28aceafcc02f8743fb594bdfd3b79535a7550544

I did not tested this - i will do this tomorrow and would create the PR. Would somewone expect this to be a breaking change? Cause it could be security related - i would expect it to go into the 19 branch...

ADDISON74:
[...]

PR #1970, opened 2022-01-08
prevent customer review view from loading foreign reviews 

aterjung:
### Description (*)
Review View in the customers menu and load any review changing the id in the url. Even not approved reviews can be loaded. This PR moves the Review loading logic to the controller and prevents it from loading reviews not created by the customer loading it.

1. Fixes OpenMage/magento-lts#1965

Go to customer menu, select "My Product Reviews" and select an review. In the view, change the id to a review id not belonging to your customer. 

### Questions or comments
Some more changes were needed to move the loading logic to the controller. Could this be a breaking change?
This opens the ability to view not approved foreign reviews, this should be merged into the LTS branch.

### DIFF
--- a/app/code/core/Mage/Review/Block/Customer/View.php
+++ b/app/code/core/Mage/Review/Block/Customer/View.php
@@ -51,7 +51,6 @@ class Mage_Review_Block_Customer_View extends Mage_Catalog_Block_Product_Abstrac
         parent::__construct();
         $this->setTemplate('review/customer/view.phtml');
 
-        $this->setReviewId($this->getRequest()->getParam('id', false));
     }
 
     /**
@@ -74,10 +73,7 @@ class Mage_Review_Block_Customer_View extends Mage_Catalog_Block_Product_Abstrac
      */
     public function getReviewData()
     {
-        if ($this->getReviewId() && !$this->getReviewCachedData()) {
-            $this->setReviewCachedData(Mage::getModel('review/review')->load($this->getReviewId()));
-        }
-        return $this->getReviewCachedData();
+        return Mage::registry('current_review');
     }
 
     /**
--- a/app/code/core/Mage/Review/Block/Customer/View.php
+++ b/app/code/core/Mage/Review/Block/Customer/View.php
@@ -51,6 +51,7 @@ class Mage_Review_Block_Customer_View extends Mage_Catalog_Block_Product_Abstrac
         parent::__construct();
         $this->setTemplate('review/customer/view.phtml');
 
+        $this->setReviewId($this->getRequest()->getParam('id', false));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Review/controllers/CustomerController.php
```diff
@@ -66,6 +91,12 @@ public function indexAction()
 
     public function viewAction()
     {
+        $review = $this->_loadReview((int) $this->getRequest()->getParam('id'));
+        if (!$review) {
+            $this->_forward('noroute');
```
luigifab, line 96: I suggest to redirect to reviews list instead of 404 error page.

aterjung: would't you expect an additional information message when redirected to the list view? In which cases this could ever happen to regular users?

app/code/core/Mage/Review/Block/Customer/View.php
```diff
@@ -74,10 +73,15 @@ public function getProductData()
...
-        return $this->getReviewCachedData();
+        return Mage::registry('current_review');
+    }
+
+    /**
+     * @return int
+     */
+    public function getReviewId()
+    {
+        return $this->getReviewData()->getId();
```
tmotyl, line 84: this will throw error when `current_review` is not set. Previously calling getReviewId() returned null 
.
What about `getData('review_id')` -> this will not work any more, right?

So the change is a breaking change, but it doesn't have to be.

aterjung: What would you suggest, setting the reviewId in the constructor?

sreichel:
> prevents it from loading reviews not created by the customer loading it.

I've tested it and it does not harm to load reviews from other customers -  as long they are approved (only this should be fixed).

app/code/core/Mage/Review/controllers/CustomerController.php
luigifab: Yes, why not. This can happen to regular users when a review is deleted.

luigifab: And when you are redirected to 404 error page, you loose the customer menu, I think it's not good.

fballiano:
Hi everybody, what are we doing with this one?

fballiano:
I've modified this PR, reverting back the block and changing the controller to redirect instead of forward, it's cleaner on the frontend in my opinion.

app/code/core/Mage/Review/controllers/CustomerController.php
```diff
@@ -47,6 +47,28 @@ public function preDispatch()
...
+     */
+    protected function _loadReview($reviewId)
+    {
+        if (!$reviewId) {
+            return false;
+        }
+
+        /* @var Mage_Review_Model_Review $review */
+        $review = Mage::getModel('review/review')->load($reviewId);
+        if (!$review->getId() || $review->getCustomerId() != Mage::getSingleton('customer/session')->getCustomerId()){
```
luigifab, line 65: A space is missing! But I agree.

fballiano: fixed it

fballiano (approved):

luigifab (approved):

PR #1982, opened 2022-01-15
[PHP 8.0] Check if track `progressdetail` is an array before calling `count`

elidrissidev:
### Description (*)
In shipment tracking popup there is a use of `sizeof` on a nullable array which can end up throwing an error on PHP 8.0+, this PR fixes it by adding a simple `is_array` check before (and changing `sizeof` to `count` for clarity).

1. Create shipment for a Sales Order and select a built-in carrier like UPS for example.
2. Clicking "Track Order" will open the tracking popup which will display a warning on PHP 7.2+, and an error starting from PHP 8.0.


Sdfendor (approved):
@elidrissidev Good find 👍.

Flyingmana (approved):

Issue #1958, opened 2022-01-05
PHP 8.1 Issue with the way deleting cookies is done in Mage_Core_Model_Cookie


Sdfendor:
### Summary (*)

There is a `delete` method in the cookie model (`Mage_Core_Model_Cookie`) where the deletion is done by setting the cookie - given by its name (in addition with other optional parameters) - to null via the `set` method within the same model. In this `set` method the built-in function `setcookie` is used.
Since PHP 8.1 passing null to a non-nullable parameter emits a deprecation error. The 2nd parameter `value` of the `setcookie` function has the type string and is non-nullable ([php.net reference](https://www.php.net/manual/en/function.setcookie.php)).
Regarding cookie deletion the reference states: 
```Cookies must be deleted with the same parameters as they were set with. If the value argument is an empty string, and all other arguments match a previous call to setcookie, then the cookie with the specified name will be deleted from the remote client. This is internally achieved by setting value to 'deleted' and expiration time in the past.```

The deprecation error is only thrown and not ignored if one follows the steps described in #1812 ("_disabled Suppress deprecation warnings in app/code/core/Mage/Core/functions.php_"). ~Either way the cookie isn't deleted when calling the `delete` method. Using `setcookie` with a value of null has no effect anymore!~ Currently the cookie is still deleted regardless of the ignore status of deprecation errors. 

The problem exists in both OM 19.4.5 or 20.0.13 since 19.4.9 in which the `delete` method code has been refactored to use the `set` method.

### Examples (*)

After logging into the admin interface (adminhtml try to instantly logout by clicking the appropriate button. The `Mage_Adminhtml_IndexController` tries to delete the session cookie (by using the cookie model) which in turn leads to the deprecation error in the system log and the rest of the action is performed. When opening the admin login dialog again there is still the session cookie for adminhtml the action tried to delete prior.  

### Proposed solution

Instead of setting a cookie to null set it to empty string and for good measure directly to a point in time in the past to let it expire instantaneously. The `set` method in the cookie model should already support that by using value = '' and period = -1.

PR #1984, opened 2022-01-20
PHP 8.1 Fix cookie deletion behavior

Sdfendor:
### Description (*)

As described in the corresponding issue internal function `setcookie` only excepts string as value parameter and PHP 8.1 emits decription error (if not ignored by OpenMage) in case _null_ is passed. 
While the prior way of using _null_ still works (but with deprecation error) the_ correct way to delete the cookie is using an empty string as value. PHP internally changes the value to `deleted` which is sent in the response header. It's not necessary to pass anything else like a period that lies in the past (using negative value when calling set method of cookie model). 

1. Fixes OpenMage/magento-lts#1958

1. Log into Admin area
2. Immediately log out and observe the response header
3. Find a Set-Cookie field with `adminhtml=deleted`

### DIFF
--- a/app/code/core/Mage/Core/Model/Cookie.php
+++ b/app/code/core/Mage/Core/Model/Cookie.php
@@ -264,7 +264,7 @@ class Mage_Core_Model_Cookie
         if (PHP_VERSION_ID >= 70300) {
             setcookie(
                 $name,
-                $value,
+                (string)$value,
                 [
                     'expires'  => $expire,
                     'path'     => $path,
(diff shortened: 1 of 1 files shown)


elidrissidev (approved):
LGTM. PHP 8.1 seems to have deprecated passing `null` to non-nullable parameters for all internal functions, so I guess this error may be occurring in a lot of other areas.

Flyingmana (changes requested):

elidrissidev (approved):

Flyingmana (approved):

PR #1987, opened 2022-01-26
Add docblock to app\code\core\Mage\Adminhtml\Block\Widget\Grid\Column…

kiatng:
…\Renderer\Abstract.php

### Description (*)
I work with grid a lot, this commitI added docblock to one file to remove warning in my IDE.

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
@@ -38,12 +38,19 @@ abstract class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Abstract
     protected $_defaultWidth;
     protected $_column;
 
+    /**
+     * @param Mage_Adminhtml_Block_Widget_Grid_Column $column
+     * @return $this
+     */
     public function setColumn($column)
     {
         $this->_column = $column;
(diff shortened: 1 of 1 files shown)


Sdfendor (approved):

ADDISON74 (approved):

fballiano (approved):

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
```diff
@@ -77,6 +84,10 @@ public function renderExport(Varien_Object $row)
         return $this->render($row);
     }
 
+    /**
+     * @param Varien_Object $row
+     * @return string|null
```
Sekiphp, line 89: I'm not sure about return type of method `_getValue`. If getter is callable, is returned result of called method `return call_user_func($getter, $row)`.

kiatng: I have in fact made use of the `getter` param, yes, it can be either a method or a callback, but the return of both only make sense if the return type (null, string, or number) can be rendered in the cell of the grid; ie, the type should not be an object or array, for example. Therefore, string is the expected return of this function.

elidrissidev (approved):

PR #1999, opened 2022-02-01
Add missing attributes to `salesOrderInvoiceInfo` SOAP method

elidrissidev:
### Description (*)
This PR will add 3 missing attributes to be returned in the `salesOrderInvoiceInfo` SOAP call:
- `order_created_at`
- `billing_firstname`
- `billing_lastname`

These attributes are mentioned in the official [Magento documentation](https://r-martins.github.io/m1docs/guides/m1x/api/soap/sales/salesOrderInvoice/sales_order_invoice.info.html) but are missing from the response.

1. Run the following code and you should see the above attributes after the change:
```php
$client = new SoapClient('<magento_host>/api/v2_soap/?wsdl');
$session = $client->login('<username>', '<password>');
var_dump($client->salesOrderInvoiceInfo($session, '<increment_id>');
```

### Questions or comments

I assumed the billing address will never be `null` so I didn't do any extra checks. Correct me if I'm wrong.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Invoice/Api.php
+++ b/app/code/core/Mage/Sales/Model/Order/Invoice/Api.php
@@ -98,6 +98,9 @@ class Mage_Sales_Model_Order_Invoice_Api extends Mage_Sales_Model_Api_Resource
 
         $result = $this->_getAttributes($invoice, 'invoice');
         $result['order_increment_id'] = $invoice->getOrderIncrementId();
+        $result['order_created_at'] = $invoice->getOrder()->getCreatedAt();
+        $result['billing_firstname'] = $invoice->getBillingAddress()->getFirstname();
+        $result['billing_lastname'] = $invoice->getBillingAddress()->getLastname();
 
         $result['items'] = array();
         foreach ($invoice->getAllItems() as $item) {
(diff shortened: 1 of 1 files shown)


Sdfendor (approved):

app/code/core/Mage/Sales/Model/Order/Invoice/Api.php
```diff
@@ -98,6 +98,9 @@ public function info($invoiceIncrementId)
 
         $result = $this->_getAttributes($invoice, 'invoice');
         $result['order_increment_id'] = $invoice->getOrderIncrementId();
+        $result['order_created_at'] = $invoice->getOrder()->getCreatedAt();
+        $result['billing_firstname'] = $invoice->getBillingAddress()->getFirstname();
```
fballiano, line 102: could we also add the middlename?

fballiano (changes requested):

app/code/core/Mage/Sales/Model/Order/Invoice/Api.php
elidrissidev: Hey, I added these attributes to conform with the official M1 documentation which doesn't include middlename.

Happy to see you became a collaborator btw ;)

fballiano: Thanks @elidrissidev!

Regarding the API, if the middlename is an available field I'd add it anyway, I had to do the same "fix" on another part of the API cause the middlename was used in the database and not returned by the APIs ;-)

elidrissidev: Done. I will keep looking for other inconsistencies between the docs and the api and open following PRs.

fballiano (approved):

kiatng (approved):
Tested.

Issue #2001, opened 2022-02-03
EAV Datetime attribute can't save Time when using different Locale


Hanmac:
### Preconditions (*)

1. OpenMage LTS 19.4.13
2. Backend Locale "de-DE", timezone "Europe/Berlin"

### Steps to reproduce (*)

1. add a Product EAV Attribute with type=datetime, input=**datetime** and backend and frontend set to the correct datetime classes
2. set the Backend Locale to something else, in my case "de-DE", timezone "Europe/Berlin"
3. use DateTimePicker to set the value for the time, for example today: "03.02.2022 17:00"
4. try to save the product
5. Mage_Eav_Model_Entity_Attribute_Backend_Datetime will fail to parse the Datetime value, try to use Locale but uses getDateFormat instead of getDateTimeFormat.
6. now in the Database there is "2022-02-03 00:00:00"

### Expected result (*)

1. [Screenshots, logs or description]
2. in the Database is "2022-02-03 17:00:00"   (or maybe "2022-02-03 16:00:00" when it would fallback to UTC? )

### Actual result (*)

1. [Screenshots, logs or description]
2. now in the Database there is "2022-02-03 00:00:00"

Also the Frontend should probably differ if it is date or datetime

kiatng:
See PR #2003.

PR #2003, opened 2022-02-07
Fixed issue #2001, EAV datetime attribute can't save time.

kiatng:
### Description (*)
See issue #2001.

1. Fixes OpenMage/magento-lts#2001

1. Add a catalog attribute with property `frontend_input` set to `datetime`.
2. Add the attribute to a Attribute Set `default`.
3. In backend product page, try saving the attribute with date and time value.
4. Without this commit, the time is not saved.

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Attribute/Backend/Datetime.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Attribute/Backend/Datetime.php
@@ -83,7 +83,7 @@ class Mage_Eav_Model_Entity_Attribute_Backend_Datetime extends Mage_Eav_Model_En
         else {
             $date = Mage::app()->getLocale()->date(
                 $date,
-                Mage::app()->getLocale()->getDateFormat(Mage_Core_Model_Locale::FORMAT_TYPE_SHORT),
+                Mage::app()->getLocale()->getDateTimeFormat(Mage_Core_Model_Locale::FORMAT_TYPE_SHORT),
                 null,
                 false
             );


Hanmac:
there two things related to this Issue:
* `Mage_Eav_Model_Entity_Attribute_Frontend_Datetime` doesn't output the time, maybe make it depend on the input type?
* in Backend Attribute Configuration you can't choose the Frontend Input, should probably be added to Mage_Eav_Model_Adminhtml_System_Config_Source_Inputtype too

kiatng:
> there two things related to this Issue:
> 
> * `Mage_Eav_Model_Entity_Attribute_Frontend_Datetime` doesn't output the time, maybe make it depend on the input type?
> * in Backend Attribute Configuration you can't choose the Frontend Input, should probably be added to `Mage_Eav_Model_Adminhtml_System_Config_Source_Inputtype` too

I think these can be in separate PRs. The broader the PR, the more time is required to get reviewed and merged.

elidrissidev:
@Hanmac If the `datetime` input type isn't available while creating an attribute from the admin panel, then how did you come across this issue the first time? Or did you modify it manually from the database?

fballiano (approved):

fballiano:
I checked this topic and, because in the catalog_product_entity_datetime the value is a datetime (with 00:00:00) saving a real datetime value doesn't introduce any BC so... approved

ADDISON74 (approved):

fballiano:
> I checked this topic and, because in the catalog_product_entity_datetime the value is a datetime (with 00:00:00) saving a real datetime value doesn't introduce any BC so... approved

mmm unless somebody is expecting to have "00:00:00" in the database :-\ mah

kiatng:
> > I checked this topic and, because in the catalog_product_entity_datetime the value is a datetime (with 00:00:00) saving a real datetime value doesn't introduce any BC so... approved
> 
> mmm unless somebody is expecting to have "00:00:00" in the database :-\ mah

That somebody should check that the attribute's `frontend_input` is `datetime`. If it's a `date`, then his/her expectation is valid.

I tested again for `frontend_input` = `date`, it was saved correctly:
image
Saved as:
```
  ["test_at"] => string(19) "2022-06-08 03:06:00"
  ["test_on"] => string(19) "2022-06-08 00:00:00"
```

PR #2011, opened 2022-02-14
Fixed the unused parameter in assignProduct

rubanooo:
The defined parameter is never passed to the intended method.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Category/Api.php
+++ b/app/code/core/Mage/Catalog/Model/Category/Api.php
@@ -464,7 +464,7 @@ class Mage_Catalog_Model_Category_Api extends Mage_Catalog_Model_Api_Resource
     {
         $category = $this->_initCategory($categoryId);
         $positions = $category->getProductsPosition();
-        $productId = $this->_getProductId($productId);
+        $productId = $this->_getProductId($productId, $identifierType);
         $positions[$productId] = $position;
         $category->setPostedProducts($positions);
 
--- a/app/code/core/Mage/Catalog/Model/Category/Api.php
+++ b/app/code/core/Mage/Catalog/Model/Category/Api.php
@@ -404,11 +404,14 @@ class Mage_Catalog_Model_Category_Api extends Mage_Catalog_Model_Api_Resource
     }
 
     /**
-     * Get prduct Id from sku or from product id
+     * Get the product ID from a product ID or SKU. When $identifierType is left empty the helper will try to
+     * automatically parse the given $productId and determine if it is a SKU or ID value
      *
-     * @param int|string $productId
-     * @param  string $identifierType
+     * @param int|string  $productId      The product ID or SKU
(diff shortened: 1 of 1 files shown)


tmotyl:
@rubanooo can you please describe an issue this patch is solving, so its easier to test the patch.

rubanooo:
> @rubanooo can you please describe an issue this patch is solving, so its easier to test the patch.

The `$identifierType` parameter was not used in the function `assignProduct`, the `$identifierType` should be passed to `_getProductId` so you can assign products to categories based on sku.

tmotyl:
@rubanooo can you outline steps needed to reproduce the issue? Just by reading, the change looks legit, but in order to merge it, it needs to be tested.

rubanooo:
@tmotyl In my case I was writing a custom method to assign categories to products. I found when I tried to sync a product that has a numeric SKU, that the `assignProduct` method would fail because when the `_getProductId` method is ran with a numeric ID, the method parses it as a product ID instead of a product SKU. 

You can test `assignProduct` with a numeric SKU. In this case the method will fail because no product can be found with an ID of the SKU value.

Sdfendor:
Could you add the information that "sku" is the expected value of identifier type (in case not null)? As I understand the called helper within the method, no other value is possible (except the default null).
Besides that good fix.

rubanooo:
@Sdfendor I have added some explanation in the phpdoc

Sdfendor:
Thanks 👍.

Sdfendor (dismissed):

sreichel (dismissed):

app/code/core/Mage/Catalog/Model/Category/Api.php
```diff
@@ -464,7 +467,7 @@ public function assignProduct($categoryId, $productId, $position = null, $identi
     {
         $category = $this->_initCategory($categoryId);
         $positions = $category->getProductsPosition();
-        $productId = $this->_getProductId($productId);
+        $productId = $this->_getProductId($productId, $identifierType);
```
kiatng, line 470: Can you fix the docblock for `$identifierType` at line 462?

```
// from
* @param null $identifierType
// to
* @param null|string $identifierType Should be 'sku' when working with SKU's. Leave null when using ID's
```

rubanooo: Added 😃

kiatng (approved):

Sdfendor (approved):

Sekiphp (approved):

elidrissidev (approved):
Tested and it works fine. To reproduce:
```php
$client = new SoapClient('<magento-host>/api/v2_soap/?wsdl');
$session = $client->login('<username>', '<password>');
$client->catalogCategoryAssignProduct($session, <category-id>, '1209', null, 'sku');
```
Assuming you have a product with numeric sku like `1209`, this throws a `SoapFault` with message `Product not exists.` before this change.

fballiano (approved):

ADDISON74 (approved):

PR #2014, opened 2022-02-14
Add polyfill for PHP8 native string functions

tmotyl:
- str_ends_with
- str_starts_with
- str_contains

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -420,3 +420,48 @@ if (version_compare(PHP_VERSION, '7.0.0', '<') && !function_exists('random_int')
         return mt_rand($min, $max);
     }
 }
+
+/**
+ * polyfill for PHP 8.0 function "str_contains"
+ */
+if (!function_exists('str_contains')) {
+    /**
+     * @param string $haystack
+     * @param string $needle
+     * @return bool
+     */
+    function str_contains($haystack, $needle)
+    {
+        return '' === $needle || false !== strpos($haystack, $needle);
+    }
+}
+
+/**
+ * polyfill for PHP 8.0 function "str_starts_with"
+ */
+if (!function_exists('str_starts_with')) {
+    /**
+     * @param string $haystack
+     * @param string $needle
+     * @return bool
+     */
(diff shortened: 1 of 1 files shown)


sreichel (approved):

Sdfendor:
The introduced methods in PHP 8.0 are core functionality and not part of an extension. so in case PHP is in or above 8.0 the methods are present. Therefore I suggest to introduce a version compare conditional around the three polyfills. Potential other 8.0 related polyfills could be placed within this conditional as well.

tmotyl:
@Sdfendor checking function_exists helps in cases, when somebody introduced the methods in other modules, or has a composer package with polyfill installed.
So we can add a version check but function_exists should stay to avoid fatal error.

Sdfendor:
@tmotyl I would keep the function_exists but nonetheless add the PHP version check. That is a good plan!

joshua-bn:
There's no need for a version check. Just check if the function exists.

joshua-bn:
Why not? Do you think it's some kind of micro optimization? Adding version
compare is a bad code smell.

Sekiphp:
Why do not use composer package `symfony/polyfill-php80`?

Flyingmana (approved):

Flyingmana:
checking by version seems to be slower by factor 10 with 7.4 checked it with the code from https://stackoverflow.com/questions/41568048/function-exists-performance-cost

But we can still change it later if we have an improved version, or decide to use an external polyfill

tmotyl:
> Why do not use composer package `symfony/polyfill-php80`?

Because not everybody uses composer unfortunatelly and we havent changed this requirement.

Sdfendor:
@tmotyl I valid point to consider. 

@Flyingmana The version compare code mentioned in Stack Overflow is quite long-winded, unoptimizied and don't even utilize the built-in `version_compare` (the method is used in our code base in some occurrences). 
The mentioned `symfony/polyfill-php80` has the following:
```php
if (\PHP_VERSION_ID >= 80000) {
    return;
}
```
I can't imagine this slower by factor 10 compared against three `function_exists` calls.

PR #2019, opened 2022-02-15
Patch to fix vulnerability APSB22-12

Sekiphp:
### Description (*)
Hello, 

I prepared fix for M2 vulnerability which should be also in OpenMage. I find the metod in M1 and apply patch. Unfortunatelly I'm not experienced with this code snippet. Can someone help to test / review the code?

Vulnerability info: https://helpx.adobe.com/security/products/magento/apsb22-12.html
Patch info: https://gist.github.com/wigman/171f9314d692d23330591d20cec3a9fd
Composer patch for M2: https://github.com/magento/knowledge-base/blob/main/src/troubleshooting/known-issues-patches-attached/assets/MDVA-43395_EE_2.4.3-p1_COMPOSER_v1.patch.zip?raw=true

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Core/Model/Email/Template/Filter.php
+++ b/app/code/core/Mage/Core/Model/Email/Template/Filter.php
@@ -375,8 +375,16 @@ class Mage_Core_Model_Email_Template_Filter extends Varien_Filter_Template
         $parts = explode('|', $construction[2], 2);
         if (2 === count($parts)) {
             list($variableName, $modifiersString) = $parts;
-            return $this->_amplifyModifiers($this->_getVariable($variableName, ''), $modifiersString);
+
+            $text = $this->_getVariable($variableName, '');
+            $pattern = '/{{.*?}}/';
+            do {
+                $text = preg_replace($pattern, '', (string)$text);
+            } while (preg_match($pattern, $text));
+
+            return $this->_amplifyModifiers($text, $modifiersString);
         }
+
         return $this->_getVariable($construction[2], '');
     }
 


Flyingmana:
did you test, if the variable parsing still works, or it removed it fully now? Also if this makes a difference for active or inactive email queue? I think the issue in M2 appeared by double parsing the template.
Although this whole part is only a very basic protection against the RCE(which does not apply to OpenMage) and the sql Injection (which seemed to not even happen in M2 core but a 3th party module?)

Still, double parsing would still be an issue which we should fix, but first we would need to validate if it even happens. Should propbably not be hard, just adding a variable reference into a username, and seeing if it is correctly resolved

dford-avb:
I have verified currently {{var store.getFrontendName()}} does get parsed but haven't been able to get any other variables to parse.

Screenshot from 2022-02-16 09-55-46
Screenshot from 2022-02-16 09-56-58

luigifab:
I tried to apply the patch, but something is wrong...

* Apply the PR.
* Go to app/locale/en_US/template/email/password_new.html, replace the h1 by `<h1>firstname: {{var customer.firstname}} / lastname: {{var customer.lastname}}</h1>`.
* Go to `Mage_Core_Model_Email_Template::getProcessedTemplate()`, before the last return, add `echo $processedResult;`.
* Go to any customer in OpenMage backend, set customer firstname to `{{var customer.lastname}}`, set customer lastname to `hello`.
* Check the _Send Auto-Generated Password_ checkbox.
* Save the customer.

You will get the email template displayed with a fatal error `Cannot send headers; headers already sent...`.
But the h1 is: `firstname: hello / lastname: hello`

I didn't understood the vulnerability or the fix, or my test is bad?

fballiano:
after reading all comments I think this is closable

PR #2032, opened 2022-03-03
Fix warnings on undefined user agent

Caprico85:
### Description (*)
There can be HTTP requests without an user agent (not just empty, but not set at all). In my case they are coming from an internal system talking to Magento's SOAP API. Depending on PHP's error settings, this leads to warning in the system.log.

```
Warning: Undefined array key "HTTP_USER_AGENT"  in .../app/code/core/Mage/Core/Model/Design/Package.php on line 675
```

### Related Pull Requests
None

None

Run `curl -H "User-Agent:" https:/example.com`. This sends a request without an user agent. Depending on your configuration you will see the warnings in system.log

### Questions or comments
None

### DIFF
--- a/app/code/core/Mage/Core/Model/Design/Package.php
+++ b/app/code/core/Mage/Core/Model/Design/Package.php
@@ -672,15 +672,16 @@ class Mage_Core_Model_Design_Package
     public static function getPackageByUserAgent(array $rules, $regexpsConfigPath = 'path_mock')
     {
         foreach ($rules as $rule) {
-            if (!empty(self::$_regexMatchCache[$rule['regexp']][$_SERVER['HTTP_USER_AGENT']])) {
+            $userAgent = $_SERVER['HTTP_USER_AGENT'] ?? '';
+            if (!empty(self::$_regexMatchCache[$rule['regexp']][$userAgent])) {
                 self::$_customThemeTypeCache[$regexpsConfigPath] = $rule['value'];
                 return $rule['value'];
             }
 
             $regexp = '/' . trim($rule['regexp'], '/') . '/';
 
-            if (@preg_match($regexp, $_SERVER['HTTP_USER_AGENT'])) {
-                self::$_regexMatchCache[$rule['regexp']][$_SERVER['HTTP_USER_AGENT']] = true;
+            if (@preg_match($regexp, $userAgent)) {
+                self::$_regexMatchCache[$rule['regexp']][$userAgent] = true;
                 self::$_customThemeTypeCache[$regexpsConfigPath] = $rule['value'];
                 return $rule['value'];
             }


app/code/core/Mage/Core/Model/Design/Package.php
```diff
@@ -672,15 +672,16 @@ protected function _checkUserAgentAgainstRegexps($regexpsConfigPath)
     public static function getPackageByUserAgent(array $rules, $regexpsConfigPath = 'path_mock')
     {
         foreach ($rules as $rule) {
-            if (!empty(self::$_regexMatchCache[$rule['regexp']][$_SERVER['HTTP_USER_AGENT']])) {
+            $userAgent = !empty($_SERVER['HTTP_USER_AGENT']) ? $_SERVER['HTTP_USER_AGENT'] : '';
```
Sdfendor, line 675: It would be shorter (and without warning because ?? performs an isset check) and in my opinion more readable to use: 
```suggestion
            $userAgent = $_SERVER['HTTP_USER_AGENT'] ?? '';
```

Sdfendor (changes requested):

app/code/core/Mage/Core/Model/Design/Package.php
Caprico85: empty() does an isset check too. See https://www.php.net/manual/en/function.empty.php: "Determine whether a variable is considered to be empty. A variable is considered empty if it does not exist or if its value equals false. **empty() does not generate a warning if the variable does not exist.**"

But yes, it's more readable with ??. Thank you!

ADDISON74 (approved):

Sdfendor (dismissed):

tmotyl:
please rebase the patch on top of current 1.9.4.x to see if the build will be green

Caprico85:
Rebased and squashed my commits. All green now.

Sdfendor (approved):

tmotyl (approved):

luigifab (commented):
I never used that, so after 11 years, it's time to try!
Configuration:

image

cat app/design/frontend/rwd/chrome/template/page/html/header.phtml 
```Chrome!!!<?php exit;```
 cat app/design/frontend/rwd/firefox/template/page/html/header.phtml 
```Firefox!!!<?php exit;```

Then `curl -H "User-Agent:" -k https://me/test1/index.php`.
No error :)

So... this is because:
image
After comment these lines, `Undefined array key` is here, and after apply the PR, no error.

---

But, if the user agent is empty, the function must `return false` like `_checkUserAgentAgainstRegexps`, no?

Caprico85:
> But, if the user agent is empty, the function must `return false` like `_checkUserAgentAgainstRegexps`, no?

Not necessarily. Somebody could want to write a regex that matches on empty user agents. Returning false on empty user agents prevents this.

luigifab (approved):

Issue #1087, opened 2020-07-02
Detect undefined functions on PullRequests


Flyingmana:
With the PR #966 we introduced a function call, to a non defined function by accident.
The error was reported/fixed in #1084

It should be possible to detect this case with a static analyses, and have this triggering an error as one of the Checks running on a PR

possible tools?
https://github.com/vimeo/psalm
https://github.com/phan/phan
https://github.com/phpstan/phpstan

Flyingmana:
Phan is outputting a lot, but it contains:

```
lib/Varien/Io/File.php:490 PhanUndeclaredFunction Call to undeclared function \char() (Did you mean \chr())
```

my config I used:
https://gist.github.com/Flyingmana/484003b1887ae71ce6a96146cc4a73fa

not sure yet how I limit the results to only specific errors

Also it does not know context, so also has false positives like in Mage_PageCache_Model_Control_Zend

Also it finds some cases where I think its not an issue?
```
lib/Cm/Cache/Backend/Redis.php:866 PhanUndeclaredFunction Call to undeclared function \snappy_compress()
lib/Cm/Cache/Backend/Redis.php:868 PhanUndeclaredFunction Call to undeclared function \lz4_compress() (Did you mean \gzcompress() or \gzuncompress())
lib/Cm/Cache/Backend/Redis.php:888 PhanUndeclaredFunction Call to undeclared function \snappy_uncompress()
lib/Cm/Cache/Backend/Redis.php:890 PhanUndeclaredFunction Call to undeclared function \lz4_uncompress() (Did you mean \gzuncompress())
lib/Mage/Connect/Validator.php:411 PhanUndeclaredFunction Call to undeclared function \version_compate() (Did you mean \version_compare())
```

sreichel:
Maybe you want to test phpStan too .... https://phpstan.org/r/940d1f01-5b1e-43e5-9a28-3ec530e1cd9f

- you also need https://github.com/inviqa/phpstan-magento1
- there is my old config (#784) .... After most DOC updates merged it should give a nice result (for now)
PhpStan works too ... https://phpstan.org/r/940d1f01-5b1e-43e5-9a28-3ec530e1cd9f

Flyingmana:
Ahh nice, I did not find thisnrule when I initially searched, so tried out the other first.
Can I limit phpstan to only do this one check?

sreichel:
> 
> 
[...]

Issue #365, opened 2017-11-03
Add ability to run PHPStan 


akira28:
I'm trying to figure out how to run [PHPStan](https://github.com/phpstan/phpstan) on Magento LTS.
After dozens of tries I have a somewhat working instance.
These are the files I had to modify to make it run:
```	modified:   app/Mage.php
	modified:   app/code/core/Mage/Admin/Model/Acl/Assert/Ip.php
	modified:   app/code/core/Mage/Admin/Model/Acl/Assert/Time.php
	modified:   app/code/core/Mage/Adminhtml/Block/Widget/Grid/Block.php
	modified:   app/code/core/Mage/Adminhtml/Model/Extension.php
	modified:   app/code/core/Mage/Api/Model/Acl/Assert/Ip.php
	modified:   app/code/core/Mage/Api/Model/Acl/Assert/Time.php
	modified:   app/code/core/Mage/Authorizenet/controllers/Adminhtml/Authorizenet/Directpost/PaymentController.php
	modified:   app/code/core/Mage/Bundle/controllers/Adminhtml/Bundle/Product/EditController.php
	modified:   app/code/core/Mage/Bundle/controllers/Product/EditController.php
	modified:   app/code/core/Mage/Bundle/controllers/SelectionController.php
	modified:   app/code/core/Mage/Catalog/Model/Product/Option/Api/V2.php
	modified:   app/code/core/Mage/Centinel/Model/Api/Client.php
	modified:   app/code/core/Mage/Compiler/controllers/ProcessController.php
	modified:   app/code/core/Mage/Core/Model/Mysql4/Design/Theme/Collection.php
	modified:   app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php
	modified:   app/code/core/Mage/Dataflow/Model/Convert/Iterator.php
	modified:   app/code/core/Mage/Dataflow/Model/Convert/Iterator/File/Csv.php
	modified:   app/code/core/Mage/Dataflow/Model/Convert/Iterator/Http.php
	modified:   app/code/core/Mage/Dataflow/Model/Session/Adapter/Http.php
	modified:   app/code/core/Mage/Downloadable/controllers/Adminhtml/Downloadable/Product/EditController.php
	modified:   app/code/core/Mage/Downloadable/controllers/FileController.php
	modified:   app/code/core/Mage/Downloadable/controllers/Product/EditController.php
	modified:   app/code/core/Mage/GoogleBase/controllers/ItemsController.php
[...]

PR #2035, opened 2022-03-08
Analyze most of the source code with phpstan. Provide baseline.

tmotyl:
This patch makes phpstan analyze almost whole OpenMage source code.
To make the phpstan green despite not all issues are yet solved,
a baseline file is introduced, which contains "known issues".
Thanks to this approach we can make sure new changes
are not introducing new issues, and its easy to fix issues from baseline
one by one.

This patch also fixes some issues along the way:

1. deprecate _getHelper method from Mage_Adminhtml_Controller_Rss_Abstract
as it had incompatible interface with parent class, and was used only in one place.
2. Fix non existing class name Mage_Catalog_Model_Entity_Product_Collection
with correct one Mage_Catalog_Model_Resource_Product_Collection
3. Correct casing of the class name Zend_Pdf_Color_Rgb
4. few minor phpdocs changes
5. update macopedia/phpstan-magento1 to newest version which brings
performance improvements
6. add #[\ReturnTypeWillChange] in some places to avoid errors on php8.1

1. Fixes OpenMage/magento-lts#365
2. Fixes OpenMage/magento-lts#1087

1. ...
3. ...

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Attribute/Api.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Attribute/Api.php
@@ -217,10 +217,10 @@ class Mage_Catalog_Model_Product_Attribute_Api extends Mage_Catalog_Model_Api_Re
             $model->save();
             // clear translation cache because attribute labels are stored in translation
             Mage::app()->cleanCache(array(Mage_Core_Model_Translate::CACHE_TAG));
-            return true;
         } catch (Exception $e) {
             $this->_fault('unable_to_save', $e->getMessage());
         }
+        return true;
     }
 
     /**
--- a/.github/phpstan-baseline.neon
+++ b/.github/phpstan-baseline.neon
@@ -1,0 +1,1157 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Method Mage_Adminhtml_Block_Backup\\:\\:_prepareLayout\\(\\) should return \\$this\\(Mage_Adminhtml_Block_Backup\\) but return statement is missing\\.$#"
+			count: 1
+			path: ../app/code/core/Mage/Adminhtml/Block/Backup.php
+
+		-
+			message: "#^Method Mage_Adminhtml_Block_Backup_Dialogs\\:\\:_prepareLayout\\(\\) should return \\$this\\(Mage_Adminhtml_Block_Backup_Dialogs\\) but return statement is missing\\.$#"
+			count: 1
+			path: ../app/code/core/Mage/Adminhtml/Block/Backup/Dialogs.php
+
+		-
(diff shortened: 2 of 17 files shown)


tmotyl:
interestingly, here on github phpstan finds more issues then running locally. I'm investigating

tmotyl:
mystery solved, its due to php version. locally I run it with 7.4 and github runs 8.1

tmotyl:
looks good now

sreichel:
I'd keep _getHelper() method. It could be used in 3rd party code.

Maybe mark it as deprecated and ignore it from sniffs?

tmotyl:
> I'd keep _getHelper() method. It could be used in 3rd party code.
> 
> Maybe mark it as deprecated and ignore it from sniffs?

This issue cant be ignored afaik, only whole file can be skipped.
Alternatively i can keep the method but remove the param and make it always return rss helper? This would make it less breaking

tmotyl:
@sreichel I've excluded the RSS files from analysing and deprecated the method. We might want to remove this method in v20 branch. but this will be a different patch.

tmotyl:
FYI, once we have this one merged, I have a bunch of patches ready on top of it, which lowers down number of issues and rise the phpstan level to 1.
A little teaser:
image

sreichel (approved):

sreichel:
> FYI, once we have this one merged, I have a bunch of patches ready on top of it,

NICE! Thanks for your effort. :heart: 

Not ready for PR, but i already tried to bring it to level 2,3 or 4 as where it is possible :)

app/code/core/Mage/Catalog/Model/Product/Attribute/Api.php
```diff
@@ -217,10 +217,10 @@ public function update($attribute, $data)
             $model->save();
             // clear translation cache because attribute labels are stored in translation
             Mage::app()->cleanCache(array(Mage_Core_Model_Translate::CACHE_TAG));
-            return true;
         } catch (Exception $e) {
             $this->_fault('unable_to_save', $e->getMessage());
         }
+        return true;
```
Flyingmana, line 223: thats a very strong behavior change to now return true even in case of an error?

tmotyl: Is it? `$this->_fault('unable_to_save', $e->getMessage());` throws an exception internally, so return is not reached.

tmotyl: see https://github.com/OpenMage/magento-lts/blob/d19f0e13d3d2fcbcc2d8fa6073568d5f95442470/app/code/core/Mage/Api/Model/Resource/Abstract.php#L97

sreichel: @Flyingmana agree to _return is not reached_.

Flyingmana (approved):

PR #2040, opened 2022-03-14
Fix supplied resource is not a valid stream resource for Varien_Io_File

eneiasramos:
bugfix: HP Fatal error:  Uncaught TypeError: fclose(): supplied resource is not a valid stream resource in /opt/lampp/htdocs/lib/Varien/Io/File.php:255

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Option/Type/Select.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Option/Type/Select.php
@@ -54,7 +54,7 @@ class Mage_Catalog_Model_Product_Option_Type_Select extends Mage_Catalog_Model_P
         if (!$this->_isSingleSelection()) {
             $valuesCollection = $option->getOptionValuesByOptionId($value, $this->getProduct()->getStoreId())
                 ->load();
-            $valueCount = empty($value) ? 0 : count($value);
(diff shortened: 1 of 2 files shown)


tmotyl:
@eneiasramos what php version are you running?

eneiasramos:
@tmotyl PHP 8.0.15 the error happens when I try to export a CSV

luigifab (approved):

iamniels:
Just ran into this issue. Happy to see the MR is already here :+1:

fballiano:
> Just ran into this issue. Happy to see the MR is already here 👍

Can you let us know if it the fix works for you?

iamniels:
Yes it works. 

In general for PHP 8, the use of the error suppression operator is problematic since it doesn't work anymore AFAIK.

fballiano (approved):

luigifab:
My bad, there is a problem with `is_countable`! See #2162.

ADDISON74:
It is interesting how this PR reached the OM, although the function used is_countable() required PHP version 7.3 as a minimum to work. This would no longer be an issue today, because the new OM versions require at least version 7.3 and probably there are few who still use this version (Debian 10 default).

PR #2047, opened 2022-04-20
Pass delimiter char to preg_quote

MarcinNowakMacopedia:
### Description (*)
The most common used delimiters in PHP are /@#~,
which are not escaped by default
(by default preg_quote escapes .\+*?[^]$(){}=!<>|:-).

This avoids hard to detect issues when the content passed to regex contains char being used as a delimiter.

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [ ] Add yourself to contributors list

PR #2048, opened 2022-04-20
Pass delimiter char to preg_quote

MarcinNowakMacopedia:
### Description (*)
The most common used delimiters in PHP are /@#~,
which are not escaped by default
(by default preg_quote escapes .\+*?[^]$(){}=!<>|:-).


Sdfendor:
There is #2047 and there is this PR. Which one is the correct one and which is the duplicate? Can you please abandon one of the PRs @MarcinNowakMacopedia?

fballiano:
duplicate of https://github.com/OpenMage/magento-lts/pull/2047 (which is on v19 branch), there's no need for a PR on both branches because v19 is always merged into v20 :-)

MarcinNowakMacopedia:
> duplicate of #2047 (which is on v19 branch), there's no need for a PR on both branches because v19 is always merged into v20 :-)
sorry I didn't know that :-)

fballiano:
not a problem don't worry :-)

PR #2050, opened 2022-04-20
Prevent Zend_Date from checking locale for know date codes

MarcinNowakMacopedia:
Add U, X and c, as we know the codes are not real locales.

### DIFF
--- a/app/code/core/Zend/Date.php
+++ b/app/code/core/Zend/Date.php
@@ -544,12 +544,12 @@ class Zend_Date extends Zend_Date_DateObject
         }
 
         if (($part !== null) && !defined($part)
-            && ($part != 'ee') && ($part != 'ss') && ($part != 'GG') && ($part != 'MM') && ($part != 'EE') && ($part != 'TT')
+            && ($part !== 'ee') && ($part !== 'ss') && ($part !== 'GG') && ($part !== 'MM') && ($part !== 'EE') && ($part !== 'TT')
+            && ($part !== 'U') && ($part !== 'X') && ($part !== 'c')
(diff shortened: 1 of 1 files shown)


joshua-bn:
With this many, I would use in_array() or isset() so it's more readable. But, looks good.

fballiano:
can we know a little bit more about this?

tmotyl:
@fballiano sure, this change is a perfromance optimization that we run on prod since more than a year.
Without it, Zend_Date tries to detect Locale for a given date even though the date is not locale aware.
The logic of fetching Locale is expensive at its requires a cache lookup.
So basically this change prevents multiple calls to Zend cache during request. Its especially painful/visible on requests with many date formatting e.g. reports.

fballiano:
Thanks @tmotyl!
How did you find these "UXc" parts that you've added? Also, where there specific reasons to move from "!=" to "!=="?

tmotyl:
@fballiano 
See constants in https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Zend/Date.php#L94-L94 and then look for their usage in the code.
And also 
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Zend/Date.php#L800-L800

I have found the issue when investigating (profiling with blackfire) why some magento reports are taking ages to load / timeout.

> Also, where there specific reasons to move from "!=" to "!=="?
Its a best practice, as both sides are strings, it doesn't make sense to have "!="

fballiano (approved):

ADDISON74 (approved):

Sdfendor (approved):

PR #2051, opened 2022-04-21
Use vanillaJS instead of prototype in Wishlist add to cart column

MarcinNowakMacopedia:
Use vanillaJS instead of prototype in Wishlist add to cart column


fballiano (approved):

fballiano:
beautiful, I wish one day we'd be able to remove prototype...

ADDISON74:
Removing Prototype is like "That's one small step for the code, one giant leap for OpenMage."

I agree that if we could detach ourselves from Prototype it would be an important step. but even if we clean the code there are many extensions that use the library. I didn't feel comfortable with this library so when I had something to implement I resorted to either vanilla JS or jQuery depending on the complexity.

fballiano:
removing prototype will probably never be possible, but this small refactor is probably a small performance improvement for the frontend so I think that's good ;-)

ADDISON74:
Let's not give ideas because @luigifab can hear us and for sure he will change all OM code from Prototype to vanilla JS :))

Flyingmana (approved):

PR #2052, opened 2022-04-21
Use vanillaJS instead of encodeURI in wishlist add to cart

MarcinNowakMacopedia:


ADDISON74 (dismissed):

Sdfendor:
If I understand correctly you inline the function call defined in `js.js` which is part of varien thus part of OpenMage, not any of the 3rd party frameworks like prototype or jQuery. To replace the 3rd party frameworks is a concern I fully understand. But why not use a function that is defined in the core. Isn't it cleaner to do so? Especially because the inlined code still uses another function from `js.js`. I'm not aware of an effort to replace and remove the varien js code. The change in this PR doesn't hurt though. 
Can you please elaborate on the motivation for this change?

tmotyl:
@Sdfendor this change and similar one https://github.com/OpenMage/magento-lts/pull/2051#issuecomment-1105634079 is are just  small steps towards reducing custom JS in openmage 

I know that single change alone doesnt make an impact, but its still a step into good direction. One place less to tackle.

in one of our projects we managed to get rid of libraries like js.js and this patch was a part of this effort

Sdfendor:
Alright so if removal of as much custom JS regardless of first or third party is the goal I naturally have nothing against it :-).

Sdfendor (approved):

tmotyl (dismissed):

tmotyl:
@MarcinNowakMacopedia please rebase the patch, to make the tests go green

tmotyl:
rebased

tmotyl (approved):

luigifab (approved):

Sdfendor (approved):

fballiano (approved):

Issue #268, opened 2017-05-24
sales_order_create_sidebar_wishlist breaks if wishlist module is disabled


Flyingmana:

sreichel:
Again ... please add some code to reproduce ;)

Flyingmana:
I will, just wanted to fast note this issue to not forget about it

fballiano:
I'll close this issue just because I've created the PR and the description is much better on the PR than here.

PR #2059, opened 2022-05-01
Issue #268: sales_order_create_sidebar_wishlist breaks if wishlist module is disabled

fballiano:
The block "sales_order_create_sidebar_wishlist" must say it can't be displayed if the Mage_Wishlist module is disabled, otherwise the order creation (from the backend) will break.

### Description (*)

Disable Mage_Wishlist, try to create an order from backend, the mask will break with a fatal error before you can finish the order

### Fixed Issues

https://github.com/OpenMage/magento-lts/issues/268


app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Sidebar/Wishlist.php
```diff
@@ -109,6 +109,19 @@ public function getIdentifierId($item)
     {
         return $item->getId();
     }
+    
+    /**
+     * Retrieve display block availability
+     *
+     * @return bool
```
kiatng, line 116: Not sure if this is better:
```suggestion
     * @inheritDoc
 ```

**Review by Sdfendor — APPROVED** [review]

File: app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Sidebar/Wishlist.php [hunk 2]
```diff
@@ -109,6 +109,17 @@ public function getIdentifierId($item)
     {
         return $item->getId();
     }
+
```
luigifab, line 112: It seems that there are useless spaces here.

image

luigifab:
I tried to disable Mage_Wishlist, then I tried to create a customer from backend before creating a new order, but I can't:

https://github.com/OpenMage/magento-lts/blob/f71321b37a782bd167ef56de0c6e18e67645685d/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View/Accordion.php#L73

`Fatal error: Uncaught Error: Call to a member function addCustomerIdFilter() on bool`

I tried to create an order and a new customer at same time, but the page is displayed without error.

app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Sidebar/Wishlist.php
fballiano: fixed it! :-)

fballiano:
Thanks @luigifab, I've fixed the customer creation too.

luigifab (dismissed):

kiatng:
The check should be `Mage::helper('wishlist')->isAllow()`, which checks for 3 settings:

1. Wishlist module is enabled in the xml
2. Wishlist module is enabled in backend > System > Configuration > Advance
3. Wishlist is set to active in  backend > System > Customers > Wishlist > General Options > Enabled

If any one of the above is disabled, it returns `false`.

fballiano:
But the isAllow method only does this:

```
    public function isAllow()
    {
        if ($this->isModuleOutputEnabled() && Mage::getStoreConfig('wishlist/general/active')) {
            return true;
        }
        return false;
    }
```

it doesn't check if the module is enabled, it checks if it's active and has output enabled, wouldn't it be better to have my check plus this one together? Also because if somebody completely removes the wishlist module (i know...) the `Mage::helper('wishlist')` wouldn't work at all right?

kiatng:
> it doesn't check if the module is enabled, it checks if it's active and has output enabled, wouldn't it be better to have my check plus this one together? 

`Mage::helper('wishlist')->isAllow()` does check if the module is enabled, I tested it and it works. You can see why here at line 144:

https://github.com/OpenMage/magento-lts/blob/90d017eb2eab000b99ae0c520faa45402a367343/app/code/core/Mage/Core/Helper/Abstract.php#L138-L152

Line 144: `if (!$this->isModuleEnabled($moduleName)) {` which call the sasme method as `Mage::helper('core')->isModuleEnabled('Mage_Wishlist')`.

> Also because if somebody completely removes the wishlist module (i know...) the `Mage::helper('wishlist')` wouldn't work at all right?

I simulated this case by calling `Mage::helper('wishlist1')`, it throws:

**Fatal error:  Uncaught Error: Class 'Mage_Wishlist1_Helper_Data' not found**

I believe this is not that bad as the description is quite clear. They can then trace it to the offending code and remove all traces of Wishlist module.

Alternatively, to play nice in this case, we need to do the same check as in `Mage::helper('wishlist')->isAllow()`:

`Mage::helper('core')->isModuleOutputEnabled('Mage_Wishlist') && Mage::getStoreConfig('wishlist/general/active')`

But, that not ideal just to take care of a fringe case.

fballiano:
@kiatng Fair enough, I switched the check

kiatng (dismissed):
Tested and worked as expected.

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View/Accordion.php
```diff
@@ -69,16 +69,18 @@ protected function _prepareLayout()
...
-            ->addCustomerIdFilter($customer->getId())
-            ->addStoreData()
-            ->getSize();
-        // add wishlist ajax accordion
-        $this->addItem('wishlist', array(
-            'title' => Mage::helper('customer')->__('Wishlist - %d item(s)', $wishlistCount),
-            'ajax'  => true,
-            'content_url' => $this->getUrl('*/*/viewWishlist', array('_current' => true)),
-        ));
+        if (Mage::helper('core')->isModuleEnabled('Mage_Wishlist')) {
```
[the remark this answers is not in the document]
fballiano: I don't know why you're seing this but that's what I've already done

fballiano:
it's really weird cause that method calls this code:

Schermata 2022-05-16 alle 12 59 47

which doesn't make any sense... i'll add another check to the method

elidrissidev:
@fballiano The problem is with `$this->getCustomerWishlists()` inside the if statement which calls the wishlist collection

fballiano:
the code I posted is the isMoveToWishlistAllowed method, which is anyway called before that

fballiano:
changed the isMoveToWishlistAllowed method ;-)

app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Items/Grid.php
```diff
@@ -466,7 +466,7 @@ public function getItemExtraInfo($item)
      */
     public function isMoveToWishlistAllowed($item)
     {
-        return $item->getProduct()->isVisibleInSiteVisibility();
+        return Mage::helper('wishlist')->isAllow() and $item->getProduct()->isVisibleInSiteVisibility();
```
luigifab, line 469: `and` !? oh no!

luigifab (changes requested):

app/code/core/Mage/Adminhtml/Block/Sales/Order/Create/Items/Grid.php
fballiano: :-D fixed it

luigifab (approved):

elidrissidev (approved):

kiatng (approved):

Issue #828, opened 2019-10-11
Quickwin/Idea: Add item id to list as info in front of row (or behind)


seansan:
Quickwin/Idea: Add item id to list as info in front of row (or behind)

![image](https://user-images.githubusercontent.com/652395/66657409-0fe15b00-ec40-11e9-9db1-de171ec9cc74.png)

sreichel:
Pleas update title ;)

fballiano:
Closing because I've created https://github.com/OpenMage/magento-lts/pull/2060

PR #2060, opened 2022-05-01
Added database ID to EAV options table

fballiano:
Everything is exaplained in https://github.com/OpenMage/magento-lts/issues/828

### Tech info

I've added a "new" prototype variable called "_id" because, when adding a new attribute option, the template would have printed "option_1", "option_2" etc for the not-yet-saved options, which seemed wrong. This way only the IDs that are already saved in the database are actually shown.

https://github.com/OpenMage/magento-lts/issues/828

From the backend, go to "catalog -> attributes -> manage attributes", open an attribute with multiple options.

### DIFF
--- a/app/design/adminhtml/default/default/template/eav/attribute/options.phtml
+++ b/app/design/adminhtml/default/default/template/eav/attribute/options.phtml
@@ -75,7 +75,6 @@
         <div class="hor-scroll">
             <table class="dynamic-grid" cellspacing="0"  cellpadding="0">
                 <tr id="attribute-options-table">
-                    <th style="width:1px" class="a-right">ID&nbsp;&nbsp;</th>
                     <?php foreach ($this->getStores() as $_store): ?>
                         <th><?php echo $this->escapeHtml($_store->getName()); ?></th>
                     <?php endforeach ?>
@@ -88,7 +87,6 @@
                         </th>
                     </tr>
                     <tr class="no-display template" id="row-template">
-                        <td class="a-right">{{_id}}&nbsp;&nbsp;</td>
                         <?php foreach ($this->getStores() as $_store): ?>
                         <td><input name="option[value][{{id}}][<?php echo $_store->getId() ?>]" value="{{store<?php echo $_store->getId() ?>}}" class="input-text<?php if($_store->getId()==0): ?> required-option<?php endif ?>" type="text" <?php if ($this->getReadOnly()):?> disabled="disabled"<?php endif ?>/></td>
                         <?php endforeach ?>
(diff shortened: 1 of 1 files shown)


app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -75,6 +75,7 @@
         <div class="hor-scroll">
             <table class="dynamic-grid" cellspacing="0"  cellpadding="0">
                 <tr id="attribute-options-table">
+                    <th style="width:1px">ID</th>
```
elidrissidev, line 78: Works flawlessly, but how about increasing the width a little as it looks too tight, something like `25px`?

Screen Shot 2022-05-17 at 4 32 54 PM

fballiano: I'd changed the PR, with a better alignment and spacing (I think) also when numbers became bigger

Example:
Schermata 2022-05-17 alle 18 49 59

ADDISON74:
What are the benefits of this PR for a Magento user? I personally don't like to see that column there. ID is a number that I do not use elsewhere in the store, for example in the Frontend. There will be situations when the ID has more or less digits and the string on the column will look bad not like in the picture. If it has to be there in order to fix something why not hide it from the display like a hidden column?

luigifab (commented):
Why not, not sure if it's really useful. Perhaps without the `&nbsp;` or in gray?
I don't know.

image

fballiano:
mmm I wouldn't make it less readable actually.

about the `&nbsp;` I put it because I couldn't simply add a "padding-right" only to the `th` and I didn't want to introduce another DOM element (like a `span`) just for a margin.

I'm not a super fan of this PR (although mine) but I wanted to implement the issue that was opened by another user and I can see some use for it (sometimes users need the ID of an attribute value for various reasons and they've to search for it in the source code, which is very uncomfortable.

ADDISON74:
@fballiano - Here's what I suggest, add a new configuration value in a specific section in Backend to show or hide that column. So whoever wants to use it can show it immediately. I tell you honestly I didn't see its usefulness and no one ever asked me for it. If it will be integrated then I will have to hide it myself.

fballiano:
mmm I think a configuration for that it's kinda overkill.

If none cares we can close this PR.

I know one of my customer would use this but doesn't matter.

ADDISON74:
It doesn't matter to the store user, as I said I never used this ID somewhere, but it could help the developers in a certain situation. Instead of looking for values in the database they will have them displayed. That's why I recommended that this feature be enabled somewhere in the Backend. We reconcile both goat and cabbage.

kiatng:
Why not use the title attribute to show the ID when mouse over? No need for additional column.

fballiano:
interesting idea, the title attribute on which element? can it be used on an input field?

luigifab:
You can add the title on tr tag.

fballiano:
Thank you guys, I've rewritten the PR, it was a good thoughts process and I think we got the best result :-)

kiatng (approved):

ADDISON74 (approved):

PR #2061, opened 2022-05-02
Remove allcontributors

luigifab:
### Description

This is an old idea/bug. Each day I go to the [home](https://github.com/OpenMage/magento-lts) of the repository, my browser download 10 MB of avatars images. This is not acceptable.

I suggest to redirect to [pulse](https://github.com/OpenMage/magento-lts/pulse).

avatars

### Contribution checklist


elidrissidev:
An alternate solution could be to [load images lazily](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading) by overriding `contributorTemplate` in `.all-contributorsrc` file.

Flyingmana:
how about moving it out of the readme onto the website?

fballiano:
> An alternate solution could be to [load images lazily](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-loading) by overriding `contributorTemplate` in `.all-contributorsrc` file.

it seems a great idea, I think I found the original template here: https://allcontributors.org/docs/en/cli/configuration and it seems to be   `"contributorTemplate": "<a href=\"<%= contributor.profile %>\">\" width=\"<%= options.imageSize %>px;\" alt=\"\"/><br /><sub><b><%= contributor.name %></b></sub></a>"` right?

elidrissidev:
Yes, that's true @fballiano

luigifab:
So I closing and create a new one with `contributorTemplate` update.

Issue #2017, opened 2022-02-15
add required PHP modules to composer.json


Flyingmana:
as we absolutely require some php extensions like mbstring, we should make sure this is visible in the composer.json

luigifab:
Here is what PHPStorm propose for app directory:
```
    "ext-simplexml": "*",
    "ext-json": "*",
    "ext-curl": "*",
    "ext-iconv": "*",
    "ext-gd": "*",
    "ext-libxml": "*",
    "ext-dom": "*",
    "ext-soap": "*",
    "ext-mbstring": "*",
    "ext-zlib": "*",
    "ext-ftp": "*",
    "ext-pdo": "*"
```

Flyingmana:
now we need to find out, which of them are optional :/
Iam pretty sure the ftp one is not generally needed?

elidrissidev:
Mage_Dataflow is using `ext-ftp`, and I think Mage_Backup too but it's disabled by default.

luigifab:
I updated my anwser by adding php package.
Not sure who provides "ext-libxml", perhaps php-xml.
I checked my pcakages, php-iconv is not installed.
I don't remember if php-soap is required or not.

elidrissidev:
> I don't remember if php-soap is required or not.
Yes it is, it's used by Mage_Api and I some other core modules.

fballiano:
I've created a PR for this

PR #2064, opened 2022-05-03
Added required PHP modules to composer.json

fballiano:
Everything is explained in ISSUE https://github.com/OpenMage/magento-lts/issues/2017


fballiano:
@tmotyl could you help me updating the composer lock?

tmotyl:
`composer update —lock` should be enough

fballiano:
to be able to run composer update --lock I had to update the composer.json file again adding a few things, do you think it's ok?

Sdfendor (approved):

tmotyl:
I can't say whether the change of ""magento-root-dir": "."" is correct or not, I'm not sure what it does.

fballiano:
> I can't say whether the change of ""magento-root-dir": "."" is correct or not, I'm not sure what it does.

it seems to be where composer installs magento, it's weird but without that any composer command wouldn't work

luigifab:
I think that the good composer command is: `composer update --lock --no-install`
With it, `magento-root-dir` changes are not required.

fballiano:
Did as @luigifab and worked flawlessly, thank you so much! :-)

luigifab:
If we add [that](https://github.com/OpenMage/magento-lts/issues/2017#issuecomment-1048620586), peoples need to install **php-ftp** and **php-iconv**, right?
I never installed them, and I don't want to install them (ok I use `--ignore-platform-reqs`, but...).

fballiano:
> If we add [that](https://github.com/OpenMage/magento-lts/issues/2017#issuecomment-1048620586), peoples need to install **php-ftp** and **php-iconv**, right? I never installed them, and I don't want to install them (ok I use `--ignore-platform-reqs`, but...).

I think so. bBut you sent me the list of extensions :-)

luigifab:
Remove ext-ftp and ext-iconv, zlib? I don't know, and let's go.

fballiano:
mmm but it seems to me that the ConfigurableSwatches module uses iconv and it is required by app/code/core/Mage/Install/etc/install.xml

composer.json
```diff
@@ -46,7 +58,6 @@
   "extra": {
     "branch-alias": {
       "dev-main": "1.9.4.x-dev"
-    },
```
Flyingmana, line 49: unrelated change which needs to be avoided

fballiano: removed

Flyingmana:
so following https://magento.stackexchange.com/questions/61377/minimum-number-of-php-modules-needed-to-run-magento it seems we got all the commonly needed exceptions.
Only ftp seems to not be commonly needed, so if we remove this one I think the PR is ready.

Thanks for all the work already :)

fballiano:
ftp extension removed

ADDISON74 (approved):

Flyingmana (approved):

Issue #1208, opened 2020-09-15
75f13c9eea changed the function prototype for Mage_GiftMessage_Model_Observer->checkoutEventCreateGiftMessage breaking Ebizmarts Sagepaysuite


rbrown:
Hi, 75f13c9eea includes the following changes to the method `Mage_GiftMessage_Model_Observer->checkoutEventCreateGiftMessage`

```diff
      * Operate with gift messages on checkout proccess
      *
-     * @param Varieb_Object $observer
+     * @param c $observer
      * @return $this
      */
-    public function checkoutEventCreateGiftMessage($observer)
+    public function checkoutEventCreateGiftMessage(Varien_Event_Observer $observer)
     {
         $giftMessages = $observer->getEvent()->getRequest()->getParam('giftmessage');
```

Ebizmarts Sagepaysuitepro includes code similar to

```php
        $event = new Varien_Object;
        Mage::getModel('giftmessage/observer')->checkoutEventCreateGiftMessage($event);
```

This was the correct object to create according to the docblock, but now causes php to die because Varien_Object is not a child of Varien_Event_Observer.

I guess my questions are:
1. Should the prototype have been changed in 19.4.x
2. If it should, should I be contacting Ebizmarts  to request they update their code to use. a Varien_Event_Observer?

colinmollenhour:
I've seen code like this before as well.  A good compromise would be to use `Varien_Object` in the type declaration since `Varien_Event_Observer` extends `Varien_Object`.

sreichel:
Sorry for introducing this ... 

It may be a BC breaking change, but i would prefer a change in 3rd party code an keep/make core code consistent.

rbrown:
I'm a big fan of consistency, but with strict typing this change is breaking an interface that, according to the history in this repo, was created in 2009 in v1.1.1 d19f0e13d 

When this interface is broken like this it stops customers being able to place orders using this module

fballiano:
Hi everybody, just to check, is this issue still valid? should we do something about it?

rbrown:
Yes, there's been no change in the downstream module, and openmage should (I think)  change the type declaration to Varien_Object as Colin says in https://github.com/OpenMage/magento-lts/issues/1208#issuecomment-693113573 to preserve the historical contract.

fballiano:
I've created a PR for this, let's see if people will approve it

fballiano:
[...]

PR #2068, opened 2022-05-04
Changed type to Varien_Object (issue #1208)

fballiano:
Everything is explained in issue https://github.com/OpenMage/magento-lts/issues/1208

### DIFF
--- a/app/code/core/Mage/GiftMessage/Model/Observer.php
+++ b/app/code/core/Mage/GiftMessage/Model/Observer.php
@@ -103,10 +103,10 @@ class Mage_GiftMessage_Model_Observer extends Varien_Object
     /**
      * Operate with gift messages on checkout proccess
      *
-     * @param Varien_Event_Observer $observer
+     * @param Varien_Object $observer
      * @return $this
      */
-    public function checkoutEventCreateGiftMessage(Varien_Event_Observer $observer)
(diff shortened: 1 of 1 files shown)


Sdfendor (approved):

ADDISON74 (approved):

kiatng:
The file https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/GiftMessage/Model/Observer.php contains many more methods with `Varien_Event_Observer` in the param. Shouldn't we change all of them?

Also, I think this is better `@param Varien_Event_Observer|Varien_Object $observer`:

```php
/**
     * Set gift messages to order item on import item
     *
     * @param Varien_Event_Observer|Varien_Object $observer
     * @return $this
     */
    public function salesEventConvertQuoteItemToOrderItem(Varien_Object $observer)
    {
```

fballiano:
actually, the more I think about it, the more this PR is wrong. Checking all of the source code I can see that only Varien_Event_Observer is used. we shouldn't have changed the "interface" back in 2019 but since it's done 3 years ago it's not right to remove it now. We should have done in in v20 but now it's too late anyway.

A fix to any imcompatible module would be extremely easy to do so... I'll close this PR and the issue.

PR #2069, opened 2022-05-05
Update References to Documentation/Magento Website - Issue #988

fballiano:
This PR aims to close https://github.com/OpenMage/magento-lts/issues/988 with a few last small changes.

I've follow @colinmollenhour last comment https://github.com/OpenMage/magento-lts/issues/988#issuecomment-1118658721 because it made a lot of sense.

Also, some deprecated and unused code was removed (I did it in this same PR because it is part of the same topic).

### DIFF
--- a/app/code/core/Mage/AdminNotification/Helper/Data.php
+++ b/app/code/core/Mage/AdminNotification/Helper/Data.php
@@ -34,22 +34,6 @@
  */
 class Mage_AdminNotification_Helper_Data extends Mage_Core_Helper_Abstract
 {
-    const XML_PATH_POPUP_URL    = 'system/adminnotification/popup_url';
-
-    /**
-     * Widget Popup Notification Object URL
-     *
-     * @var string
-     */
-    protected $_popupUrl;
-
-    /**
-     * Is readable Popup Notification Object flag
-     *
-     * @var bool
-     */
-    protected $_popupReadable;
-
     /**
      * Last Notice object
      *
@@ -89,45 +73,5 @@ class Mage_AdminNotification_Helper_Data extends Mage_Core_Helper_Abstract
             $this->_unreadNoticeCounts = Mage::getModel('adminnotification/inbox')->getNoticeStatus();
         }
         return isset($this->_unreadNoticeCounts[$severity]) ? $this->_unreadNoticeCounts[$severity] : 0;
-    }
-
-    /**
-     * Retrieve Widget Popup Notification Object URL
-     *
-     * @param bool $withExt
-     * @return string
-     */
-    public function getPopupObjectUrl($withExt = false)
-    {
-        if (is_null($this->_popupUrl)) {
-            $this->_popupUrl = 'https://' . Mage::getStoreConfig(self::XML_PATH_POPUP_URL);
-        }
-        return $this->_popupUrl . ($withExt ? '.swf' : '');
-    }
-
-    /**
-     * Check is readable Popup Notification Object
-     * @deprecated after 1.4.2.0
-     *
-     * @return bool
-     */
-    public function isReadablePopupObject()
-    {
-        if (is_null($this->_popupReadable)) {
-            $this->_popupReadable = false;
-            $curl = new Varien_Http_Adapter_Curl();
-            $curl->setConfig(array(
-                'timeout'   => 2
-            ));
-            $curl->write(Zend_Http_Client::GET, $this->getPopupObjectUrl(true));
-            if ($curl->read()) {
-                if ($curl->getInfo(CURLINFO_HTTP_CODE) == 200) {
-                    $this->_popupReadable = true;
-                }
-            }
-
-            $curl->close();
-        }
-        return $this->_popupReadable;
     }
 }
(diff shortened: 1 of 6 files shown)


Sdfendor (dismissed):

colinmollenhour (dismissed):

app/code/core/Mage/AdminNotification/Helper/Data.php
```diff
@@ -90,44 +74,4 @@ public function getUnreadNoticeCount($severity)
...
         return isset($this->_unreadNoticeCounts[$severity]) ? $this->_unreadNoticeCounts[$severity] : 0;
     }
-
-    /**
-     * Retrieve Widget Popup Notification Object URL
-     *
-     * @param bool $withExt
-     * @return string
-     */
-    public function getPopupObjectUrl($withExt = false)
```
luigifab, line 100: This method is used by `Mage_Adminhtml_Block_Notification_Window::getObjectUrl()`. Not sure how.

kiatng (dismissed):

app/code/core/Mage/Tax/etc/config.xml
```diff
@@ -298,9 +298,6 @@
                 <full_summary>0</full_summary>
                 <zero_tax>0</zero_tax>
             </sales_display>
-            <notification>
-                <url>http://www.magentocommerce.com/knowledge-base/entry/magento-ce-18-ee-113-tax-calc</url>
```
luigifab, line 302: This break the source link here:

image

I suggest this url: https://web.archive.org/web/20150315064836/http://www.magentocommerce.com/knowledge-base/entry/magento-ce-18-ee-113-tax-calc
Or perhaps we can copy the page to openmage.org.

luigifab (changes requested):

app/code/core/Mage/Tax/etc/config.xml
ADDISON74: Perhaps the phrase "Please see source for more details" should be removed from the warning message.

fballiano: @luigifab I've modified the PR adding your link, great idea! I looked on google's cache but coulnd't find anything, I didn't think about archive.org!

app/code/core/Mage/AdminNotification/Helper/Data.php
fballiano: You're right as always, I was looking into removing the "window" class altogether (I wrote it in the original issue) then I thought it was too much and this part slipped from me.

Now I'm not sure what to do the `getObjectUrl` method looks for a "swf", which now nobody can run so... it's more useless than ever... should I remove the `getObjectUrl` method altogether? it seems nobody is calling it.

luigifab: If you drop them, @Flyingmana will says that's a BC.
Perhaps drop content and `return null or false` and mark functions as deprecated for 19.x, and drop functions for 20.x.

fballiano: I've restored the methods, returning nothing, adding a deprecation notice. One method was deprecated since magento 1.4 thus I've removed it completely ;-)

colinmollenhour (approved):

ADDISON74:
Referring to this post https://github.com/OpenMage/magento-lts/issues/988#issuecomment-1118740255, is there a possibility that certain extensions that are still maintained by developers may have issues with Mage_Adminhtml_Block_Notification_Window removal? Is this class still used?

fballiano:
@ADDISON74, yes there's the possibility, although it would be an extremely bothering thing if an extension open an overlay when you access the backend, also the majority of vendors don't even support M1 extensions anymore so you know... anyway I'd remove the object only in the v20 branch, seems fair

this PR doesn't remove that class anyway

kiatng (approved):

ADDISON74 (approved):

elidrissidev (approved):

app/code/core/Mage/AdminNotification/Helper/Data.php
```diff
@@ -96,38 +80,10 @@ public function getUnreadNoticeCount($severity)
...
-            if ($curl->read()) {
-                if ($curl->getInfo(CURLINFO_HTTP_CODE) == 200) {
-                    $this->_popupReadable = true;
-                }
-            }
-
-            $curl->close();
-        }
-        return $this->_popupReadable;
+        return '';
```
luigifab, line 87: `getPopupObjectUrl return ''` and `isReadablePopupObject return false`, no?

fballiano: I'm not sure I understood the question correctly. The getPopupObjectUrl method was returning a URL so I kept it returning a string (empty one), while the isReadablePopupObject method is now completely removed.

luigifab: oh, ok, no problem!

luigifab (approved):

PR #2089, opened 2022-05-17
fix wrong tier price percentage rounding

empiricompany:
### Description (*)
tier prices display wrong percentage saving due to rounding issues

create simple product 
with base price = $ 57
and tier prices:
qty = 3 and above price  = $ 55.29
qty = 5 and above price  = $ 54.15
qty = 10 and above price  = $ 51.30

frontend percentage saving is wrong
immagine

the issues is related to rounding in ceil function explained here:
https://www.php.net/manual/en/function.ceil.php#122724

correct values
immagine

100 - ((100/57) * 55,29) = 3 %
**100 - ((100/57) * 54,15) = 5 %**
**100 - ((100/57) * 51,3) = 10 %**

tested on latest v20.0.13

### DIFF
--- a/app/code/core/Mage/Catalog/Block/Product/Abstract.php
+++ b/app/code/core/Mage/Catalog/Block/Product/Abstract.php
@@ -426,7 +426,7 @@ abstract class Mage_Catalog_Block_Product_Abstract extends Mage_Core_Block_Templ
                 }
 
                 if ($price['price'] < $_productPrice) {
-                    $price['savePercent'] = ceil(100 - ((100 / $_productPrice) * $price['price']));
+                    $price['savePercent'] = ceil(100 - round((100 / $_productPrice) * $price['price']));
 
                     $tierPrice = Mage::app()->getStore()->convertPrice(
                         Mage::helper('tax')->getPrice($product, $price['website_price'])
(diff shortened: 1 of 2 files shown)


ADDISON74 (approved):

elidrissidev:
I had to apply this fix to `app/code/core/Mage/Catalog/Block/Product/Price.php` to get the correct percentage to show up in the product page.

ADDISON74 (approved):

elidrissidev (approved):
Tested it and it works correctly.

elidrissidev:
Not sure if it's OK to merge this or until a release is done @fballiano ?

empiricompany:
> I had to apply this fix to `app/code/core/Mage/Catalog/Block/Product/Price.php` to get the correct percentage to show up in the product page.

maybe this block is used for configurable product?
i have add also this file to the patch

elidrissidev:
I tested with a simple product and editing only the abstract class method didn't make any difference in the savings percentage.

empiricompany:
> I tested with a simple product and editing only the abstract class method didn't make any difference in the savings percentage.

mm seems getTierPrices() that i have patched first in Abstract is used by this old but gold extension
https://github.com/organicinternet/magento-configurable-simple

I would keep both files patched

PR #2132, opened 2022-05-18
Allow to rename api.php

luigifab:
### Description

This PR allow you to rename your api.php to something else to avoid kevin's attack on it.
Like you didn't use /admin for backend.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Api/Helper/Data.php
+++ b/app/code/core/Mage/Api/Helper/Data.php
@@ -380,7 +380,7 @@ class Mage_Api_Helper_Data extends Mage_Core_Helper_Abstract
         $uri = Zend_Uri_Http::fromString($url);
         $uri->setHost($request->getHttpHost());
         if (!$urlModel->getRouteFrontName()) {
-            $uri->setPath('/' . trim($request->getBasePath() . '/api.php', '/'));
+            $uri->setPath('/' . trim($request->getBasePath() . '/' . basename(getenv('SCRIPT_FILENAME')), '/'));
         } else {
             $uri->setPath($request->getBaseUrl() . $request->getPathInfo());
         }
--- a/app/code/core/Mage/Api2/Model/Route/ApiType.php
+++ b/app/code/core/Mage/Api2/Model/Route/ApiType.php
@@ -35,6 +35,7 @@ class Mage_Api2_Model_Route_ApiType extends Mage_Api2_Model_Route_Abstract imple
 {
     /**
      * API url template with API type variable
+     * @deprecated
      */
     const API_ROUTE = 'api/:api_type';
 
@@ -56,6 +57,6 @@ class Mage_Api2_Model_Route_ApiType extends Mage_Api2_Model_Route_Abstract imple
         Zend_Translate $translator = null,
         $locale = null
     ) {
-        parent::__construct(array(Mage_Api2_Model_Route_Abstract::PARAM_ROUTE => self::API_ROUTE));
+        parent::__construct(array(Mage_Api2_Model_Route_Abstract::PARAM_ROUTE => str_replace('.php', '', basename(getenv('SCRIPT_FILENAME'))).'/:api_type'));
     }
 }


elidrissidev:
> to avoid kevin's attack on it.

What's kevin's attack? Can you provide a related link?

luigifab:
Yes, at my office, we have a problem with some "developers" that we call Kévin, that copy/paste stackoverflow, and who don't understand that they are doing. This change will prevent brute force attacks on API if you rename your file, because only you and your providers know API url.

elidrissidev:
Not only in your office, unfortunately.

fballiano:
should we document somewhere how to use a different url for /api?

luigifab:
Yes, in my to do list.

ADDISON74:
This appears in .htaccess 

```php
############################################
## rewrite API2 calls to api.php (by now it is REST only)

    RewriteRule ^api/rest api.php?type=rest [QSA,L]
```

If you rename the api.php file what do we do with the .htaccess setting?

luigifab:
Yes you do (at htaccess level, or web server config level).

fballiano (changes requested):
There's a typo: "Don't forgot to also update server configuration" should be " Don't forget to also update server configuration"

ADDISON74:
"Don't forget to update the server configuration too." or "Don't forget to update the server configuration." 

Correct forms compared to the initial one.

ADDISON74:
I propose the following message:

```
Do not use common file names like api.php for OpenMage API URLs to prevent attacks. Do not use the new file name in _robots.txt_ and keep it secret with your partners. After renaming the file you must update the webserver configuration as follows:

* Apache .htaccess: `RewriteRule ^api/rest api.php?type=rest [QSA,L]`
* Nginx: `rewrite ^/api/(\w+).*$ /api.php?type=$1 last;`
```

fballiano (approved):

ADDISON74:
Thank you for taking my suggestion into account, but you had to make the change in the other paragraph as well to keep the same addressing mode. I allowed myself to do it. 

I personally have no other comments. We can merge this PR.

ADDISON74 (approved):

elidrissidev:
It should be note that this change will only affect the REST API as SOAP/Xmlrpc are accessed via the controller directly and not through `api.php`. If it's not clear enough to everyone then it should be amended in the README.

ADDISON74:
@elidrissidev - I only commented on the existing paragraphs. If new changes are needed please make a proposal and we implement it immediately. This PR allows us editing.

elidrissidev:
> Sorry deleted my last comment by accident. This is what it said:
> 
> > It should be noted that this change only affects the REST API as SOAP/Xmlrpc APIs are accessed directly from a controller (through `index.php`) and not through `api.php`. If it's not clear enough to everyone then it should be amended in the README.

Actually I retract what I said 😅, it is indeed possible to use `api.php` as a secondary way to access SOAP/Xmlrpc APIs, it's just not possible by default. So with the below change in `.htaccess` you will be able to access all types of APIs through `api.php` from `http://<magento_host>/api/<code>/`, where code is either `rest`, `soap`, or `xmlrpc`!
```diff
## uncomment next line to enable light API calls processing

-#    RewriteRule ^api/([a-z][0-9a-z_]+)/?$ api.php?type=$1 [QSA,L]
+    RewriteRule ^api/([a-z][0-9a-z_]+)/?$ api.php?type=$1 [QSA,L]

############################################
## rewrite API2 calls to api.php (by now it is REST only)

-    RewriteRule ^api/rest api.php?type=rest [QSA,L]
+#    RewriteRule ^api/rest api.php?type=rest [QSA,L]
```

So this PR looks good to me.

elidrissidev (approved):

elidrissidev:
Another discovery I made is that updating the filename alone is not enough because you'll get "Request does not match type route." 404 error with the REST Api. To fix this you'll also need to update the URL in `.htaccess`:
```diff
############################################
## rewrite API2 calls to api.php (by now it is REST only)

-    RewriteRule ^api/rest api.php?type=rest [QSA,L]
+    RewriteRule ^ipa/rest ipa.php?type=rest [QSA,L]
```

ADDISON74:
In the OpenMage documentation the author mentioned the following:

```
Don't use common file names like api.php for OpenMage API URLs to prevent attacks. Don't use the new file name in _robots.txt_ and keep it secret with your partners. After renaming the file you must update the webserver configuration as follows:

* Apache .htaccess: `RewriteRule ^api/rest api.php?type=rest [QSA,L]`
```

He practically indicated where the new file name for the two webservers should be changed.

PR #2092, opened 2022-05-18
Remove end class comment

luigifab:
### Description

This remove all? useless `// Class xyz End` comments.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2134, opened 2022-05-18
Revert "Remove end class comment"

ADDISON74:
I am reverting PR #2092 for being squash and merge. Sorry for merging without squash.


justinbeaty:
#2092 looks like it only had one commit anyway, so no need to squash.

ADDISON74:
Indeed #2092 has a single commit. It would not require squash. In this case I leave it to the discretion of a maintainer if he keeps it or closes it.

fballiano:
If the original one was a single commit you don’t need a squash 😊 let’s leave it as is

Issue #1722, opened 2021-07-06
Possibility for multiple Credis versions to conflict


justinbeaty:
### Preconditions (*)
1. OpenMage 20.12 (but I think it will be reproducible in any version)
2. `colinmollenhour/cache-backend-redis` installed, and using the version of `Cm_RedisSession` in this repo.

### Steps to reproduce (*)
1. Install preconditions as noted above, before #1254 was merged, it was impossible to log in

### Expected result (*)
1. Log in in should work as normal

### Actual result (*)
1. With `colinmollenhour/cache-backend-redis` installed, credis was being loaded from vendor instead of `lib/Credis`

### Discussion
Since #1254 is now merged, I think the more urgent issue is resolved. I've been running the patch from that PR for the past week, and things are okay. However, I am running as headless Magento, and only the admin session creates a redis session. It's possible there are other incompatibilities between the different versions of Credis, but I haven't run into any as far as I know.

Here's the possible different versions that could be installed as I understand it:

* The version in this repo has a copyright of 2011, so pretty old
* The version (as a git submodule) in the external [Cm_RedisSession](https://github.com/colinmollenhour/Cm_RedisSession/tree/master/app/code/local) is from about 4 years ago
* The version required in `Cm_Cache_Backend_Redis` (through composer.json) is set to `"*"`, so the latest.

So this issue is more of a discussion on how small things like this should be prevented. Perhaps instead of `lib/Credis`, you would need to install Credis via composer, and the Cm_RedisSession in this repo is updated to support the latest version? Or maybe Cm_RedisSession is removed from this repo and we only should use from composer? Unfortunately, Magento 1.x comes from a time before when composer was popular, and we also have to support people who do not use composer. But, I don't think composer could even have solved this problem, except that it might just warn that you are trying to use two different versions.

justinbeaty:
I also just re-read Flyingmana's comment in #1705 where it would be best to not involve composer in fixing this, likely because there are still many non-composer users. I think the best fix then is probably updating this repo's version of `Cm_Cache_Backend_Redis` to work with the latest Credis, and also updating `lib/Credis` to the latest as well. That way, if one did install Credis via composer (via `Cm_Cache_Backend_Redis`), then there is no incompatibility. But we would have to wonder if any 3rd party modules rely on the old Credis lib, and updating it would break those modules? I don't have a really good answer yet to all of this so I'm interested to hear other opinions.

Flyingmana:
[...]

PR #2138, opened 2022-05-19
Move phpseclib, mcrypt_compat, net_idna2, Cm_RedisSession and Cm_Cache_Backend_Redis to composer

fballiano:
I think it should be time to start cleaning our repo from 3rd party modules (zend framework anyone?) but I'd like to start from a small a simpler one (at least to me).

Well, we could remove Cm_RedisSession and leave it up to the user to install it (as any other module) from @colinmollenhour's repo.

Or, with the PR, I'm removing it from the repo and moving it to composer.json.

For better discussing the matter I'll start a comment with some observation.


composer.json
```diff
@@ -47,6 +48,11 @@
     "branch-alias": {
       "dev-main": "1.9.4.x-dev"
     },
-    "magento-root-dir": "root"
+    "magento-root-dir": "."
+  },
+  "config": {
+    "allow-plugins": {
+      "magento-hackathon/magento-composer-installer": true
+    }
```
fballiano, line 56: I can't find a way to install the module (and update the composer.lock) without these lines. I also think that magento-root-dir should really be "."

What do you guys think? Is it ok? Could you give me a hand here?

fballiano:
Another important thing, we can't release something like this if we don't have a "release builder" that creates the "release.tgz" downloading the module and adding it to "app/code", right? how could we approach that? I could create a script for that.

I've searched github documentation but it seems to me that there's no automatic way to include some build binaries into a newly created release, am I missing something?

justinbeaty:
I’m a fan of this as well as the general direction of using composer modules and a build step.

If this direction is taken, we should remove lib/credis and have that added via the build script as well. There is some more details in #1722 that I encountered last year.

fballiano:
I've also moved zf1 (introducing zf1-future) and pelago_emogrifier to composer.

From what I've tested it work flawlessly

tmotyl:
The problem with that is that many people still runs magento in non-composer mode.
For them we need to provide an easy to use way of downloading magento with dependencies.
This can be achieved by building and attaching a zip with full openmage to every release.

fballiano:
> The problem with that is that many people still runs magento in non-composer mode. For them we need to provide an easy to use way of downloading magento with dependencies. This can be achieved by building and attaching a zip with full openmage to every release.

I also run every customer without composer, it's fundamental to have a release builder, do you know some specific github instruments to do so? cause actually i didn't find much.

fballiano:
I was thinking, why don't we commit the composer autoloader patch (the one to app/Mage.php) as part of our repo?

fballiano:
Also, why there's a composer patch when a few lines down there's another support for includes?

Schermata 2022-05-20 alle 14 18 25

colinmollenhour:
I think the release file could be built with the combination of a composer action like https://github.com/owenvoke/composer-action with https://github.com/actions/create-release.

You're right that the "includes" directory is redundant to the autoloader patch, I think it predates it. Not sure how widely used it is but would say it is ok to remove it in a future release.

composer.json
Flyingmana: I recommend to test it very good what a root-dir of "." will do with an install of this project. Iam not sure what kind of unintended consequences we can have.

Flyingmana:
> Another important thing, we can't release something like this if we don't have a "release builder" that creates the "release.tgz" downloading the module and adding it to "app/code", right? how could we approach that? I could create a script for that.
> 
> I've searched github documentation but it seems to me that there's no automatic way to include some build binaries into a newly created release, am I missing something?

have a look at this action here https://github.com/Flyingmana/magento-lts/blob/release-actions/.github/workflows/release.yml#L45

Flyingmana:
> I've also moved zf1 (introducing zf1-future) and pelago_emogrifier to composer.
> 
> From what I've tested it work flawlessly

dont mix multiple things in a single PR, please try to do it step by step, and ideally spread it over multiple releases, so we can first gauge the feedback we get, especially if we move required core files out of the repository

Flyingmana:
> Also, why there's a composer patch when a few lines down there's another support for includes?
> 
> Schermata 2022-05-20 alle 14 18 25

I cant really look at the diff, because its to many changes for my browser to process it correctly, but the installer adds this patch, because its also used for non-openmage projects.

fballiano:
> > Also, why there's a composer patch when a few lines down there's another support for includes?
> > Schermata 2022-05-20 alle 14 18 25
> 
> I cant really look at the diff, because its to many changes for my browser to process it correctly, but the installer adds this patch, because its also used for non-openmage projects.

I know that, my question is that a few lines after that there's an include that mention composer... so I think we should have a native support for the composer autoloader and avoid composer patching the app/Mage.php which seems some kind of workaround from a million years ago :-D

composer.json
fballiano: I don't understand (and never found any doc) how you can have a composer json/lock without those lines. composer requires them. How was the composer json/lock generated if composer doesn't work without those?

fballiano: I've manually removed the line, although everything seems very weird about this composer.json

luigifab:
zf1-future doesn't include some changes made by magento/openmage contributors. I didn't have the full list.

fballiano:
> zf1-future doesn't include some changes made by magento/openmage contributors. I didn't have the full list.

I know and I think it was wrong to do those changes in our repo that way. Anyway I'll move the zf1-future in a separare PR, as requested by @Flyingmana

fballiano:
I modified this PR to process only Cm_RedisSession and Pelago_Emogrifier, not zend framework. Also I've remove the magento-root-dir from the composer.json

composer.json
Flyingmana: > I don't understand (and never found any doc) how you can have a composer json/lock without those lines. composer requires them. How was the composer json/lock generated if composer doesn't work without those?

simply said, you dont. The composer.json was intended for having OpenMage as a dependency, not as an installable project. So all this questions were not relevant till now.

fballiano:
I've updated this PR, which I think it's now not that bad, to complete the activity
- I've created a release builder workflow which I'll publish in a separate PR https://github.com/fballiano/magento-lts/blob/test_release_workflow/.github/workflows/release.yml
- This is a test release I've created with the release builder workflow https://github.com/fballiano/magento-lts/releases/tag/vprova123.1.6 and you can see the ZIP file is correct (with Cm_RedisSession, Cm_Cache_Backend_Redis and Pelago_Emogrifier in the original places in /lib or /app/code/community and app/etc/modules)
- This is a test run of the release building workflow https://pipelines.actions.githubusercontent.com/serviceHosts/3e020c0a-0aab-49ba-a131-e89b3e602b32/_apis/pipelines/1/runs/476/signedlogcontent/2?urlExpires=2022-05-29T15%3A25%3A34.3520780Z&urlSigningMethod=HMACV1&urlSignature=%2B3obw%2BRu0x1oPlo9qf6%2FtwfNPVPyyNgkRLxicYShNPI%3D

ADDISON74:
If people are not runing composer they must be pointed to download the files from these repos and extract them where they were before.

fballiano:
> If people are not runing composer they must be pointed to download the files from these repos and extract them where they were before.

there's no need, as I explained in my previous comment and in https://github.com/OpenMage/magento-lts/pull/2165 the release builder restore those files and directories exactly where they are/were.

fballiano:
I had to go back on my steps about the Pelago_Emogrifier lib since none of their versions supports the PHP versions we need, all of them were generating incompatibilities... that's something we'll have to discussion in a separate topic

fballiano:
at the moment i've removed the composer.lock from this PR since it has to be synced right before merging

fballiano:
I'm checkin phpstan but it seems to me that I can't work if composer installation dir is set to "root" since the directory doens't exist so everything kinda falls apart. But I lack knowledge here and would need some help.

justinbeaty:
@fballiano I don't see why it can't change to `"magento-root-dir": "."` 

Anyone who is using `aydin-hassan/magento-core-composer-installer` must set their own `magento-root-dir` value else they get an error:

```
  [InvalidArgumentException]
  magento-root-dir must be specified in root package
```

fballiano:
Thanks @justinbeaty for the info, in the past I changed the magento-root-dir but, if I remember correctly, I was told not to :-\

sreichel:
^^^root^^^ is default value from composer-installer? I'd like to see this changed to ^^^.^^^ too.

justinbeaty:
@fballiano I somewhat remember that coming up in a previous issue, but I don’t know why it couldn’t change…? I thought that maybe if a user didn’t specify their own root dir that this repo’s value would be used (and that would be bad if we changed it to “.”) but I tested it and that doesn’t happen. Maybe an old version of the core-installer did that? Either way I agree with the change.

fballiano:
actually my fault, I found the discussion but it wasn't a "let's not change it", it was more of a "I'm not sure what it does", so let's change it :-)

phpstan still complains so I'll check asap!

fballiano:
First error:
Error: Call to method setSaveHandler() on an unknown class Mage_Cm_Redissession_Model_Session.

I think it comes from here:
Schermata 2022-08-03 alle 20 53 36

what if we change it to this:
```php
            case 'redis':
                /* @var Cm_RedisSession_Model_Session $sessionResource */
                $sessionResource = Mage::getSingleton('cm_redissession/session');
                if ($sessionResource) {
                    $sessionResource->setSaveHandler();
                    break;
                }
```
would you like it?

fballiano:
In the meanwhile I wrote to the maintainer or https://github.com/pear/Net_IDNA2 to see if he can tag a release for us, since the fix to our phpstan error is already merged in their repo (in 2019) but unreleased. let's see.

justinbeaty:
@fballiano For the Redis issue, I did some digging and I think this is not the solution.

Cm_Redissession used to work by overriding `Mage_Core_Model_Resource_Session` and we used it not by selecting `redis` in `local.xml` but instead by selecting `db`.

This was changed in #1513 and instead of overriding, we simply allow the new model to be called, `Cm_RedisSession_Model_Session`. But in Colin's upstream module it is still wanting to re-write and thus `Mage::getSingleton('cm_redissession/session');` fails.

Hopefully that makes sense... We can fork Colin's module and while we're at it remove the submodule.

sreichel:
> EDIT: doesn't seem to solve, ideas?

Just an idea ...

Files are symlinked by default and (maybe) can't be resolved? Change deploy strategy to `copy`?

justinbeaty:
Maybe this helps explain it:

image

fballiano:
Thanks for the support guys, I've switched the deployment strategy to copy but doesn't fix it, looking closely:

```
  Line   Core/Model/Session/Abstract/Varien.php                                                   
 ------ ----------------------------------------------------------------------------------------- 
  80     Call to method setSaveHandler() on an unknown class Mage_Cm_Redissession_Model_Session.  
```

`Mage_Cm_Redissession_Model_Session` doesn't exist, the correct class would be `Cm_RedisSession_Model_Session`... why the class name generation is wrong? :-\

fballiano:
@justinbeaty sorry I saw you last comment just now (I'm a little tired now), mmmm at this moment I don't know how to solve.

@colinmollenhour could you help us here?

justinbeaty:
@fballiano I tried pushing a fork which we can try to install with composer:

```
{
  "name": "openmage/magento-lts",
  "description": "A fork of Magento-1 that is accepting bug fixes (backward compatible, drop in replacement for official Magento)",
  "license": [
    "OSL-3.0",
    "AFL-3.0"
  ],
  "type": "magento-source",
  "repositories": [
    { "type": "vcs", "no-api": true, "url": "https://github.com/justinbeaty/Cm_RedisSession.git" }
  ],
  "require": {
    "php": ">=7.0 <7.5 || >=8.0 <8.1",
    "ext-curl": "*",
    "ext-gd": "*",
    "ext-json": "*",
    "ext-mbstring": "*",
    "ext-pdo": "*",
    "ext-simplexml": "*",
    "ext-dom": "*",
    "ext-libxml": "*",
    "ext-soap": "*",
    "ext-iconv": "*",
    "ext-zlib": "*",
    "pear/net_idna2": "^0.2.0",
    "colinmollenhour/cache-backend-redis": "^1.14",
    "colinmollenhour/magento-redis-session": "dev-master"
  },
  "require-dev": {
    "macopedia/phpstan-magento1": "^1.0.4",
    "phpstan/phpstan": "^1.7.1"
  },
  "authors": [
    {
      "name": "Lee Saferite",
      "email": "lee.saferite@openmage.org",
      "role": "Maintainer"
    },
    {
      "name": "Daniel Fahlke aka Flyingmana",
      "email": "flyingmana@googlemail.com",
      "role": "Maintainer"
    },
    {
      "name": "David Robinson",
      "email": "david.robinson@aoe.com",
      "role": "Maintainer"
    },
    {
      "name": "Tymoteusz Motylewski",
      "role": "Maintainer"
    },
    {
      "name": "Sven Reichel",
      "email": "github-sr@hotmail.com",
      "role": "Maintainer"
    }
  ],
  "autoload-dev": {
    "psr-4": { "OpenMage\\Tests\\Unit\\": "dev/tests/unit" }
  },
  "extra": {
    "branch-alias": {
      "dev-main": "1.9.4.x-dev"
    },
    "magento-root-dir": "."
  },
  "config": {
    "allow-plugins": {
      "magento-hackathon/magento-composer-installer": true,
      "cweagans/composer-patches": true
    }
  }
}
```

I think the only way to solve this is to either keep this module in our repo or use a fork (or Colin releases a new 3.0 tag)

fballiano:
@justinbeaty I've transformed your fork into https://github.com/colinmollenhour/Cm_RedisSession/pull/187 for @colinmollenhour, let's see :-)

colinmollenhour:
@fballiano @justinbeaty I merged and tagged as 3.0.0

fballiano:
thanks so much @colinmollenhour 

@justinbeaty I think we messed up something

```
fab@MBP-Fab ~/P/openmage (remove_cm_redissession)> /usr/local/opt/php@8.0/bin/php /usr/local/bin/composer update
you may want to add the packages.firegento.com repository to composer.
add it with: composer.phar config -g repositories.firegento composer https://packages.firegento.com
Loading composer repositories with package information
Updating dependencies
Nothing to modify in lock file
Installing dependencies from lock file (including require-dev)
Nothing to install, update or remove
Generating autoload files
14 packages you are using are looking for funding.
Use the `composer fund` command to find out more!

In DeploystrategyAbstract.php line 382:
                                                                                                                         
  Source /Users/fab/Projects/openmage/vendor/colinmollenhour/magento-redis-session/app/code/local/Credis does not exist  
```

if we've to fix it in Colin's repo we should also update his config.xml with a new version number

justinbeaty:
@fballiano it looks like I forgot to remove Credis directory from the modman file. Was this a fatal error or just a warning? If it’s just a warning it should be okay until tomorrow and I’ll submit another PR with the module version update too

justinbeaty:
Actually one more thing, I don’t think Cm_RedisSession includes Credis via composer at all, so removing that submodule would in fact break things (except we’re saved by including Cm_Cache_Backend_Redis via composer). 

But, this is a good thing because of issue #1722. So I’ll also PR to add Credis via composer into Cm_RedisSession and we should be all good.

justinbeaty:
(Maybe actually one more thing) - I need to also PR the fix in #1254 / #1705 (assuming it’s not already there) because when we are now using the latest Credis, we will run into the same problem.

Also as a note to myself check to see if any other changes were made to Cm_RedisSession in this repo.

fballiano:
> Actually one more thing, I don’t think Cm_RedisSession includes Credis via composer at all, so removing that submodule would in fact break things (except we’re saved by including Cm_Cache_Backend_Redis via composer).

I was wondering that, could it be included in Colin's repo's composer.json?

justinbeaty:
Pushed a few more commits to https://github.com/justinbeaty/Cm_RedisSession but let's test this before submitting the PR.

- Fixed the modman file
- Added Credis to composer.json
- Regarding the fixes from #1254 / #1705 it is already in the repo here: colinmollenhour/php-redis-session-abstract#16
- Bumped the version in config.xml

I'll test installing it to my local env via composer to make sure it's okay

fballiano:
it seems that Credis only lives in the "vendor" folder now. At the end of the day it could be ok since with autoload there shouldn't be no need to move it anywhere else :-)

justinbeaty:
@fballiano Yes it's definitely okay in just vendor, and solves the issue of three possible different versions as reported in #1722.

For the release builder however we should move it to lib. Note that there's a breaking change from the current version of Credis in lib to the one now installed via composer so it may be prudent to mention it somewhere in the release notes.

fballiano:
Can I ask you what's the breaking change? I couldn't find it

justinbeaty:
I explained it a bit over here: https://github.com/OpenMage/magento-lts/pull/1705#issuecomment-869054124

Basically at one point Magento developers likely modified lib/Credis to add this:

```php
case 'hmget':
    $response = array_values($response);
    break;
```

But never submitted it as an upstream patch. So the breaking change is what `$redis->hMGet()` returns in Credis. If someone was relying on the lib/Credis in this repo (which had a copyright of 2011, so pretty old) then us upgrading to the latest Credis is breaking. All they need to do to fix their code is run the output of `hMGet` through `array_values`, like in #1705.

fballiano:
Can we add that part of code to the PR to Colin's repo or it doesn't make sense these many years later?

justinbeaty:
> Can we add that part of code to the PR to Colin's repo or it doesn't make sense these many years later?

I don't think it makes sense, because then we are making a breaking change in Credis, and I assume much more people use Credis than just for OpenMage. Plus, Cm_RedisSession has already been updated to work with the latest Credis version anyway so no need to make that change other than for the possibility that some 3rd party developer relies on that functionality for their custom Magento module.

fballiano:
ok understood, thank you! :-)

justinbeaty:
Pushed a few more commits to the fork. Notably, I moved the code back to the `community` code pool. This will avoid that step in your release builder where you have to move it manually to that folder. I think this change should be okay since (1) we are removing the module in the community folder in this repo, but also (2) composer install seemed to have no problem with either copy or symlink deploy strategy.

I tested installing locally with composer, that seems all good. I also tested forking this branch and requiring the updated fork of Cm_RedisSession to see if all the actions pass: https://github.com/justinbeaty/magento-lts/actions?query=branch%3Aremove_cm_redissession (just one action is queued).

fballiano:
I'm not sure if Colin would agree on moving to the community pool but I'm absolutely pro this choice, I think FLOSS modules should be in the community pool :-)

justinbeaty:
@fballiano I am not sure either, but I can revert the commit if he disagrees. The reason it's in the local pool I suppose is to not conflict with the version that was bundled in Magento. But if we're removing it, then I don't see why we can't move it. His Cm_Cache_Backend_Redis module installs via composer / modman to the community folder as well.

As mentioned, `magento-composer-installer` seems to be fine overwriting files in community, as does modman (even though the 3.0 branch cannot be installed via modman):

```console
~/magento-root/htdocs$ ~/bin/modman clone git@github.com:justinbeaty/Cm_RedisSession.git
Cloning into 'Cm_RedisSession'...
remote: Enumerating objects: 930, done.
remote: Counting objects: 100% (53/53), done.
remote: Compressing objects: 100% (34/34), done.
remote: Total 930 (delta 11), reused 31 (delta 4), pack-reused 877   
Receiving objects: 100% (930/930), 234.95 KiB | 1.12 MiB/s, done.
Resolving deltas: 100% (350/350), done.         
WARNING: Removing conflicting directory: /home/www-data/magento-root/htdocs/app/code/community/Cm/RedisSession
 Applied: app/code/community/Cm/RedisSession  app/code/community/Cm/RedisSession
WARNING: Removing conflicting regular file: /home/www-data/magento-root/htdocs/app/etc/modules/Cm_RedisSession.xml
 Applied: app/etc/modules/Cm_RedisSession.xml  app/etc/modules/Cm_RedisSession.xml
Successfully create symlink new module 'Cm_RedisSession'
```

fballiano:
Agree on everything

justinbeaty:
I get a PHPStan error on my branch: https://github.com/justinbeaty/magento-lts/runs/7674837674?check_suite_focus=true

But it doesn't seem related to anything in this PR and we get past the "Install dependencies" stage. Let me try and rebase that branch to see if this was fixed elsewhere.

justinbeaty:
Hm, I pushed a new branch with commits squashed and rebased on 20.0 and still the same PHPStan error: https://github.com/justinbeaty/magento-lts/tree/remove_cm_redissession_clean

I do not know why that is.

colinmollenhour:
Merged your latest PR as 3.0.1, no objections to any of the changes. Thanks!

fballiano:
Thanks everybody :-)

phpstann now complains about:

```
Error: Ignored error pattern #^Binary operation "\*" between string and 0\.3 results in an error\.$# in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Catalog/Model/Resource/Product/Indexer/Price.php is expected to occur 2 times, but occurred only 1 time.
Error: Ignored error pattern #^Binary operation "\*" between string and 0\.3 results in an error\.$# in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/CatalogInventory/Model/Resource/Indexer/Stock.php is expected to occur 2 times, but occurred only 1 time.
Error: Ignored error pattern #^PHPDoc tag @param has invalid value \(\$item\)\: Unexpected token "\$item", expected type at offset 18$# in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Checkout/Block/Multishipping/Addresses.php was not matched in reported errors.
Error: Ignored error pattern #^PHPDoc tag @param has invalid value \(\$a\)\: Unexpected token "\$a", expected type at offset 18$# in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Payment/Helper/Data.php was not matched in reported errors.
Error: Ignored error pattern #^PHPDoc tag @param has invalid value \(\$b\)\: Unexpected token "\$b", expected type at offset 35$# in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Payment/Helper/Data.php was not matched in reported errors.
 ------ ----------------------------------------------------------------------- 
  Line   Catalog/Model/Resource/Product/Indexer/Price.php                       
 ------ ----------------------------------------------------------------------- 
  227    Ignored error pattern #^Binary operation "\*" between string and 0\.3  
         results in an error\.$# in path                                        
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Catalog/  
         Model/Resource/Product/Indexer/Price.php is expected to occur 2        
         times, but occurred only 1 time.                                       
 ------ ----------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   CatalogInventory/Model/Resource/Indexer/Stock.php                      
 ------ ----------------------------------------------------------------------- 
  181    Ignored error pattern #^Binary operation "\*" between string and 0\.3  
         results in an error\.$# in path                                        
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/CatalogI  
         nventory/Model/Resource/Indexer/Stock.php is expected to occur 2       
         times, but occurred only 1 time.                                       
 ------ ----------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   Checkout/Block/Multishipping/Addresses.php                             
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^PHPDoc tag @param has invalid value            
         \(\$item\)\: Unexpected token "\$item", expected type at offset 18$#   
         in path                                                                
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Checkout  
         /Block/Multishipping/Addresses.php was not matched in reported         
         errors.                                                                
 ------ ----------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   Payment/Helper/Data.php                                                
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^PHPDoc tag @param has invalid value \(\$a\)\:  
         Unexpected token "\$a", expected type at offset 18$# in path           
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Payment/  
         Helper/Data.php was not matched in reported errors.                    
         Ignored error pattern #^PHPDoc tag @param has invalid value \(\$b\)\:  
         Unexpected token "\$b", expected type at offset 35$# in path           
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Payment/  
         Helper/Data.php was not matched in reported errors.                    
 ------ ----------------------------------------------------------------------- 
```

but nothing seems related to this task, maybe we should only regenerate the baseline?

justinbeaty:
@fballiano yes that's the same error I got on my own fork even when I rebased on latest 20.0. I tried to figure out why this is now just happening but I don't know.

sreichel:
This makes no sense to me ... just for the last error in Payment/Helper/Data.php ...

The igorned error belongs to this DOCblock, where the parameter type is missing ...

```
    /**
     * @param $a
     * @param $b
     * @return int
     */
    protected function _sortMethods($a, $b)
```

This is fixed in unmerged PR #2336

https://github.com/OpenMage/magento-lts/blob/01dc0a0230d3ccb12b20c25f7ddf297bd934dfb0/app/code/core/Mage/Payment/Helper/Data.php#L86-L97

sreichel:
This is somehow related to composer.lock/php version ...

With composer install on php8 the errors does not pop up. Downgrade to php7.x - that requires a composer update, because of the .lock file - makes this appaer.

sreichel:
Phpstan is locked to 1.7.1 ... running `composer update` updates to 1.8.2 ...

justinbeaty:
`"phpstan/phpstan": "^1.7.1"` in our composer.json allows any version up to but not including 2.0.0.

We should change it to `~1.7.1` to allow only 1.7.x

sreichel:
> We should change it to `~1.7.1` to allow only 1.7.x

No. If possible, we should use the latest version and raise to `^1.8`. (currently testing)

Reason for this error/message is a change in phpstan (did not found the PR), that moved this type of error to level 6.

https://phpstan.org/r/55d76fc8-9154-4098-86bf-3c6b0791e9de

Updating phpstan + baseline should be fine.

sreichel:
Tests still running ... Just want to mention that OM officilly supports php7.0+ .... phpstan requires 7.2. Maybe we should addthis to readme (or drop support below 7.2 :rofl: )

fballiano:
should I cherry pick https://github.com/OpenMage/magento-lts/pull/2367 into this PR?

justinbeaty:
> No. If possible, we should use the latest version and raise to `^1.8`. (currently testing)

I misunderstood and thought we had to stay with 1.7.x branch. I agree newer is better. Same with PHP 7.0 -- I'm thought we already require 7.2 but looks like it's just not reflected in the README.

---

For the `composer.lock` mess, the problem is that we should avoid running `composer update` and updating all packages when we only mean to update a few. So I've again forked this PR, reverted `composer.lock` to the version in `20.0` and ran just:

```
composer update --ignore-platform-reqs phpseclib/phpseclib phpseclib/mcrypt_compat pear/net_idna2 colinmollenhour/cache-backend-redis colinmollenhour/magento-redis-session
```

All seems okay: https://github.com/justinbeaty/magento-lts/actions/runs/2804030203

---

> should I cherry pick #2367 into this PR?

It's not necessary. Instead this is a good example of when two PRs which modify `omposer.lock` will have to be handled in a merge. So when both #2367 and this PR are merged into 20.0, the `content-hash` will at least conflict. Solution is shown in the following resources:

- https://mikemadison.net/blog/2020/8/18/resolving-composerlock-git-merge-conflicts
- https://stackoverflow.com/questions/46185777/is-content-hash-a-mandatory-part-of-composer-lock

fballiano:
@justinbeaty I reverted the composer.lock and run your update command :-) let's see

justinbeaty:
@fballiano If you want, you can update this PR to my branch where I just re-organized the commits on top of latest 20.0. It might help to keep a cleaner git log and in case any one of them needs to be reverted.

https://github.com/justinbeaty/magento-lts/commits/remove_cm_redissession

```console
# make sure you're on correct branch
git checkout remove_cm_redissession

# make backup just in case
git branch remove_cm_redissession_backup remove_cm_redissession

# add my repo as remote
git remote add justin https://github.com/justinbeaty/openmage.git
git fetch justin

# reset your branch to mine
git reset --hard justin/remove_cm_redissession

# force push
git push -f
```

Let's just wait until the actions are completed: https://github.com/justinbeaty/magento-lts/actions/runs/2804361532

fballiano:
if we squash as usual it shouldn't be necessary, if I'm not missing anything. I also rebased lately.

justinbeaty:
It's up to you. Normally I wouldn't want to squash a PR like this because there are many changes that should be separate commits for both git history as well as the ability to revert if needed.

sreichel:
Should we merge #2367 into 20.0 first?

fballiano:
IMHO it's better to have 1 PR = 1 commit

@sreichel that's why I asked if I should cherry pick it because I don't know how to merge it in 20.0 otherwise.

justinbeaty:
> IMHO it's better to have 1 PR = 1 commit

I don't always agree, but it's offtopic to this PR anyways.

> @sreichel that's why I asked if I should cherry pick it because I don't know how to merge it in 20.0 otherwise.

I don't really think it's needed. We would avoid a merge conflict if #2367 is merged into 20.0 and this PR is rebased, but in the future I think we are going to have more instances of composer.lock conflicting and need to be able to deal with it. For example we may in the future open several PRs to update each library and we either have to wait until one is merged, rebase the second, etc or deal with the merge conflict. Hopefully in most cases it will just be the `content-hash` that conflicts and is easy to resolve.

fballiano:
but to merge it in 20.0 we've to create a pr just for that and it may take days or more to have to merged.

justinbeaty:
> but to merge it in 20.0 we've to create a pr just for that and it may take days or more to have to merged.

This is one of the reasons why I don't think it's necessary to include #2367 in here at all. What if that PR stayed opened for a while, should it block this PR from being merged? I don't see the reason to make one PR dependent on the other in this case because they are not related in any way except that they both modify composer.lock.

fballiano:
rebased and all green

fballiano:
question: should we do this kind of change to v19 too since we will need a release builder anyway?

sreichel:
Yes, please.

Any ideas to build different packages for different php versions? I don't think it's required here, but for Emogriefer (and php <7.3)

fballiano:
when I rebase it breaks everything so I'll close this one and create a new one on v19

fballiano:
> Any ideas to build different packages for different php versions? I don't think it's required here, but for Emogriefer (and php <7.3)

@sreichel With https://github.com/shivammathur/setup-php#matrix-setup we should be able to do it kinda easily, I'm already porting https://github.com/fballiano/openmage/blob/first_release_builder_workflow/.github/workflows/release.yml but it's difficult to test without having the composer modules actually in the composer.json

PR #2140, opened 2022-05-20
Version bump in preparation for next release

fballiano:
This is just a version bump for next release, I'm doing this at the beginning of the development cycle instead of right before release, in order not to forget at last minute.

Also, with this PR, I propose to have the same "minor" number for both v19 ad v20.
I know it's a bit weird to jump a couple of numbers but it's a bit uncomfortable, during release time, to have manage very similar but different numbers. What do you think about that?

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -222,7 +222,7 @@ final class Mage
             return array(
                 'major'     => '20',
                 'minor'     => '0',
-                'patch'     => '17',
+                'patch'     => '15',
                 'stability' => '', // beta,alpha,rc
                 'number'    => '', // 1,2,3,0.3.7,x.7.z.92 @see https://semver.org/#spec-item-9
             );
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -222,7 +222,7 @@ final class Mage
             return array(
                 'major'     => '20',
                 'minor'     => '0',
-                'patch'     => '14',
+                'patch'     => '17',
                 'stability' => '', // beta,alpha,rc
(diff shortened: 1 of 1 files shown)


Flyingmana:
minor versions have a meaning, and if we would have a system to take them more serious, having them in sync would not make sense, and older major versions would always have higher minor versions, while they likely do not increase when the one for newer major versions do increase. Also its maybe not a good idea to start such a discussion in a PR

fballiano:
I agree to the theory but I think whatever we can make to make our job easier we should do.

Anyway I'll switch it back

Flyingmana:
in the best case we have automated it via a script anyway at one point as planed with the automated release process

Flyingmana:
oh, also, as you were talking about minor version number, not patch-version, patch-version is even more likely to be asyncron, as soon as we have a bugfix which is only needed for the higher major version. Like sometimes it happens a release is broken. Should we then also for the other major version do a new (empty)release? do we skip one?

fballiano:
ye sorry i was meaning patch number your right!

let's keep it as it is, it's not a big problem, but I noticed that when preparing the release I was like 19, 4, 16, 20, 0, 14 all the time and I was getting confused :-D

luigifab:
Next release Next month? 😹

fballiano:
> Next release Next month? 😹

why not :-)

fballiano:
since @luigifab note about the phpstorm metadata i'm closing this PR, it will be done before release altogether

PR #2144, opened 2022-05-24
Cast params receiving int to string

holozaen:
Mage_Oauth_Model_Server uses php's hash_equals(string1, string2) to check the matching of tokens.

Php's hash_equals(string1, string2) expects strings as params and throws a warning if it receives numbers. In concreto, we compare the identity of the consumerId, which is an integer. We should cast the params to string here to avoid breaking the oauth authorization.

While it would (in principle) be possible to change Mage_Oauth_Model_Token and Mage_Oauth_Model_Consumer to return a string as the id/consumerId we should not do this IMHO because while we understand that a token usually is represented by a string value this is not the case for an id in models based on Mage_Core_Model_Abstract.


elidrissidev (approved):

elidrissidev:
This would also throw a TypeError in PHP 8

PR #2162, opened 2022-05-26
PHP 8 fix null provided to count() in tax calculation class

iamniels:
This is a fix for a type error which occurs when PHP 8 is used.

    Fix Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in Mage/Sales/Model/Quote/Address/Total/Tax.php:212

This error is very hard to reproduce. I saw it in my production logs and this PR fixes it.

The fix uses `is_countable`, so is working from PHP 7.3. The Magento-LTS readme states that it is compatible with PHP 7.0. Is this actually the case? In that case `is_countable` can be replaced with `!== null` for PHP 7.0 compatibility.

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote/Address/Total/Tax.php
+++ b/app/code/core/Mage/Sales/Model/Quote/Address/Total/Tax.php
@@ -209,7 +209,7 @@ class Mage_Sales_Model_Quote_Address_Total_Tax extends Mage_Sales_Model_Quote_Ad
     protected function _saveAppliedTaxes(Mage_Sales_Model_Quote_Address $address, $applied, $amount, $baseAmount, $rate)
     {
         $previouslyAppliedTaxes = $address->getAppliedTaxes();
-        $process = count($previouslyAppliedTaxes);
+        $process = is_countable($previouslyAppliedTaxes) ? count($previouslyAppliedTaxes) : 0;
 
         foreach ($applied as $row) {
             if (!isset($previouslyAppliedTaxes[$row['id']])) {


luigifab:
I checked at [php.net](https://www.php.net/is_countable), and it require: PHP 7 >= 7.3.0, PHP 8.
We are wrong already wrong!

https://github.com/OpenMage/magento-lts/blob/2611b6b8d80a25829e7f07ffcfa24df1b96eede2/app/code/core/Mage/Catalog/Model/Product/Option/Type/Select.php#L57

https://github.com/OpenMage/magento-lts/blob/2611b6b8d80a25829e7f07ffcfa24df1b96eede2/composer.json#L10

🆘

fballiano:
there's this polyfill we could introduce: https://php.watch/versions/7.3/new-is_countable-function

justinbeaty:
With the possible direction that we are using composer more, how about considering requiring these packages:

```
"symfony/polyfill-php71": "*",
"symfony/polyfill-php72": "*",
"symfony/polyfill-php73": "*",
"symfony/polyfill-php74": "*",
"symfony/polyfill-php80": "*",
"symfony/polyfill-php81": "*",
"symfony/polyfill-php82": "*",
```

eneiasramos:
@luigifab @fballiano is there a possibility to add this [Polyfill](https://php.watch/versions/7.3/new-is_countable-function) in app/bootstrap.php ?

luigifab (approved):
Not tested, but it looks good... at same time `getAppliedTaxes` is supposed to return an array:

https://github.com/OpenMage/magento-lts/blob/5112767714cefbc9d001131cce4f2fed8c013d6b/app/code/core/Mage/Sales/Model/Quote/Address.php#L1176-L1189

iamniels:
> Not tested, but it looks good... at same time `getAppliedTaxes` is supposed to return an array:
> 
> https://github.com/OpenMage/magento-lts/blob/5112767714cefbc9d001131cce4f2fed8c013d6b/app/code/core/Mage/Sales/Model/Quote/Address.php#L1176-L1189

It is possible to unserialize null from 'applied_taxes', so probably count not accepting null has uncovered an error somewhere else in the code. Nice example of the usefulness of type hinting in PHP :-)

kiatng (approved):
`Mage::helper('core/unserializeArray')->unserialize()` can return `null`. See https://github.com/OpenMage/magento-lts/blob/e6c127c11a2dcff29212a323a4f4f2db3e03bdca/lib/Unserialize/Reader/Arr.php#L71-L71

PR #2163, opened 2022-05-26
Introducing polyfill for is_countable()

fballiano:
PR https://github.com/OpenMage/magento-lts/pull/2040 introduced an is_countable() call, which doesn't exist in PHP<7.3. Also PR https://github.com/OpenMage/magento-lts/pull/2162 is introducing the same problem.

So I thought about adding a polyfill

### DIFF
--- a/app/bootstrap.php
+++ b/app/bootstrap.php
@@ -33,7 +33,7 @@ if ((LIBXML_VERSION < 20900) && function_exists('libxml_disable_entity_loader'))
     libxml_disable_entity_loader(false);
 }
 
-// Polyfill for PHP<7.4
+// Polyfill for PHP<7.3
 if (!function_exists('is_countable')) {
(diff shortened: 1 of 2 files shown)


luigifab:
In this file _app/code/core/Mage/Core/functions.php_ with the other ones?

fballiano:
I mean, in bootstrap.php there's some sort of polyfill too

fballiano:
moved it

eneiasramos:
perfect!

luigifab:
It seems that PHPStan says the truth: _Class [SimpleXMLElement](https://www.php.net/SimpleXMLElement) referenced with incorrect case: SimpleXmlElement._

fballiano:
done!

luigifab (approved):

kiatng (approved):

PR #2171, opened 2022-05-31
Backend Orders grid - More space for columns Order #, Purchased On, S…

ADDISON74:
The purpose of this PR is to solve a visual issue in the Orders grid. By default, the "**Order #**", "**Purchased On**", and "**Status**" columns are not wide enough. Thus the content uses instead of 1 row, 2 or even 3 rows and as an effect the grid increases vertically. Please note that for the "Status" column I used the longest string in the list in order to find out the proper width (PayPal Canceled Reversal).

This is how the Orders grid in OpenMage is now:

orders-grid-before

This will be the Orders grid if the PR is merged:

orders-grid

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/Grid.php
@@ -66,7 +66,7 @@ class Mage_Adminhtml_Block_Sales_Order_Grid extends Mage_Adminhtml_Block_Widget_
 
         $this->addColumn('real_order_id', array(
             'header' => Mage::helper('sales')->__('Order #'),
-            'width'  => '80px',
+            'width'  => '100px',
             'type'   => 'text',
             'index'  => 'increment_id',
             'escape' => true,
@@ -87,7 +87,7 @@ class Mage_Adminhtml_Block_Sales_Order_Grid extends Mage_Adminhtml_Block_Widget_
             'header' => Mage::helper('sales')->__('Purchased On'),
             'index' => 'created_at',
             'type' => 'datetime',
-            'width' => '100px',
+            'width' => '150px',
         ));
 
         $this->addColumn('billing_name', array(
@@ -118,7 +118,7 @@ class Mage_Adminhtml_Block_Sales_Order_Grid extends Mage_Adminhtml_Block_Widget_
             'header' => Mage::helper('sales')->__('Status'),
             'index' => 'status',
             'type'  => 'options',
-            'width' => '70px',
+            'width' => '150px',
             'options' => Mage::getSingleton('sales/order_config')->getStatuses(),
         ));
 


kiatng (dismissed):
I do not think you can satisfy everyone, but it seems this is reasonable.

Sdfendor (dismissed):

ADDISON74:
@kiatng - I appreciate you approved this PR and your observation. These two grids are most often used in the stores I manage and I received recommendations to make them look better. When I set the values I took into account real situations, including also an evaluation of the Magento Sample package for 1.9 versions. In my scenario the Orders grid shows a custom order number (ORD-XXXXXX) based on an Amasty extension. I also evaluated the other columns. It is obvious that they could be bigger, but the PR presents this issue and it can be modified over a while to cover other situations, in fact it is a harmless change. These values could be configured in the Backend in a special section too, but I haven't really encountered anything like this in Magento (with one exception in an extension I've contributed to).

ADDISON74:
@luigifab - what screen resolution are you using? It looks fine for 1920px even I shrink the window.

ADDISON74:
Here is what happens to the Product grid at various screen resolutions. The database used in testing is Magento Sample 1.9. The values in columns Name, Attrib. Set Name, SKU, are not short at all but I found that their sizes cover the real situation in some of the stores I manage and where I did the implementation already.

**Screen Resolution Width: 960px**

960px

**Screen Resolution Width: 1280px**

1280px

**Screen Resolution Width: 1437px** - starting with this width all the rows occupy a single row.

1437px

Please note for some columns the values cannot be removed, nor can smaller widths be set (eg ID, Price, Qty).

These results were obtained by adding the width of 300px (missing from OM code and this PR) for the Name column, which is absolutely necessary in this situation. After doing the test I wait for your feedbacks and move on to its implementation .

Anyone is welcome to play with the width values for these columns to cover the real situation in their stores. At this point I think the one used in this PR would be the right values.

ADDISON74:
... and to make a visual comparison. Below are screenshots of the Product grid in OpenMage default for the resolutions used in the previous test. I personally consider this PR brings significant visual improvements.

**Screen Resolution Width: 960px**

960px-default

**Screen Resolution Width: 1280px**

1280px-default

**Screen Resolution Width: 1437px**

1437px-default

ADDISON74:
The only issue that should be discussed for Orders grid is the width of the Status column. It was set by me to 150px for covering all the situations in the dropdown list but I consider a value of 100px may be just as appropriate. As a positive result the column would not be so large.

luigifab:
I using 1280x1024.

ADDISON74:
@luigifab - based on your screen resolution the last commit should fix your issue related to Name column. Give it a try and I am looking forward for your feedback.

Please check also what would be the proper width for Status column in Orders grid. 150px or 100px?

luigifab (approved):
Ok better! 150 is good for status in orders grid.

ADDISON74:
@luigifab - The only visual aspect is that when you don't have 2 or 3 words in a string on the Status column it will remains a free space. For this reason I also proposed the 100px as variant. For example, when you have the status only set to Pending, Processing, Complete or Canceled, the width of 175px provided a space is almost double than necessary.

A little clarification of why I created this PR which at first glance may seem useless, easy to implement. In the stores I manage there are content administrators with a certain well-established role who deal with orders, products, customers. I have often been asked to improve these frequently used grids because using them daily and not exposing the information in an orderly manner causes visual issues. I made a change for some and then I brought it back to get reactions. I can't tell you how quickly I was blamed for changing the grid to the original version which means that the new widths were pleasing to the eye :)

ADDISON74:
I closed the PR and I created two separated PR's so that opinions could be expressed on each of them:

#2217 - Backend > Orders grid - More space for columns
#2218 - Backend > Manage Products grid - More space for columns

PR #2178, opened 2022-06-01
Fix the calculation of bundle saleable status

aredridel:
### Description

`Mage_Bundle_Model_Product_Type` expects the parent::isSaleable() to return null in the case of a bundle that is not known to be out of stock, but the parent was flattening that into a boolean false. This returns the parent to returning null for that case so the specific type can add its own logic, in the case of bundles taking selection saleable status into account.

### Manual testing scenarios

1. Create a bundle, with one required option and two children
2. set all of its children out of stock.
3. Verify that `isSaleable()` is false
4. Set one of the children in stock
5. Verify that `isSaleable()` is true

### Contribution checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-146-orange.svg?style=flat-square)](#contributors-)
+<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-146-orange.svg?style=flat-square" alt="All Contributors"></a>
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/d/total.svg" alt="Total Downloads"></a>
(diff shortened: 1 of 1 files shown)


elidrissidev (dismissed):

elidrissidev (approved):

ADDISON74 (approved):

kiatng (approved):
Besides this reference https://github.com/OpenMage/magento-lts/blob/27407244502611a467627fdd8271e31dc1e56c7b/app/code/core/Mage/Bundle/Model/Product/Type.php#L483-L486

We have also https://github.com/OpenMage/magento-lts/blob/27407244502611a467627fdd8271e31dc1e56c7b/app/code/core/Mage/Catalog/Model/Product/Type/Grouped.php#L248-L251

So I think this is a correct fix.

kiatng:
Ooops, I approve too soon, how about this reference: https://github.com/OpenMage/magento-lts/blob/27407244502611a467627fdd8271e31dc1e56c7b/app/code/core/Mage/Catalog/Model/Product/Type/Configurable.php#L470-L488

It may return `null`, what effect would that be?

kiatng (commented):
More review required.

elidrissidev:
I double checked to make sure everything is correct as per @kiatng's comment. This change looks correct to me because `Mage_Catalog_Model_Product_Type_Configurable::isSalable` is supposed to return a `boolean` only if it _can_ determine if the product is salable, otherwise in the case of a composite product, it should offloads that responsibility to a specific type instance (configurable/grouped...etc) by returning `null`.

PR #2184, opened 2022-06-01
Commits on 1.9.3.x that are not on 1.9.4.x

fballiano:
Since github doesn't allow to dismiss the "compare & pull request" button I've to create this PR. I will close it right away since no commit in the 1.9.3.x branch should be ported to 1.9.4.x at this point in time.

We should probably freeze old branches if possible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ php:
   - 7.1
   - 7.2
   - 7.3
-  - nightly
 
 matrix:
   allow_failures:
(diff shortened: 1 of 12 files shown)


ADDISON74:
Nice find but you closed the PR. Are those commits merged in latest OM? Did we miss others?

ADDISON74:
Can we revert and leave them open for discussions? Or shall we recreate them?

fballiano:
anyone can cherry-pick one any comment into another branch

ADDISON74:
I will let someone more advanced in Git to do this task. If you can do it would be great because you know what it's about. The most important thing is not to lose them and to capitalize the efforts made in the past if they are still relevant.

PR #2206, opened 2022-06-08
Merge v19 to v20

fballiano:
I'm creating this PR to merge 1.9.4.x into 20.0 for next release. Also if it doesn't get approved I'll update it every once in a while to have a ready made branch to merge into v20 ;-)


fballiano:
I'm not sure why phpstan is failing since the `dataHasChangedFor` should be available in `Mage_Core_Model_Abstract`, which is inherited by `Mage_Catalog_Model_Product`, for sure this problem comes from https://github.com/OpenMage/magento-lts/pull/1325

@Sekiphp could you help me here? Since you managed https://github.com/OpenMage/magento-lts/pull/1325

tmotyl:
I believe the problem is that `$product` object is marked few lines before as ` /** @var Varien_Object|Mage_Catalog_Model_Product $product */`

```
  if ($eventType == Mage_Index_Model_Event::TYPE_SAVE) {
            /** @var Varien_Object|Mage_Catalog_Model_Product $product */
            $product = $event->getDataObject();
 ```
 And phpstan complains that Varien_Object doesn't have this method.
 If we're sure that this event always gets an product and never varien_object, then the type annotation should be fixed.

**Comment by fballiano** [comment]
> I believe the problem is that `$product` object is marked few lines before as ` /** @var Varien_Object|Mage_Catalog_Model_Product $product */`
> 
> ```
>   if ($eventType == Mage_Index_Model_Event::TYPE_SAVE) {
>             /** @var Varien_Object|Mage_Catalog_Model_Product $product */
>             $product = $event->getDataObject();
> ```
> 
> And phpstan complains that Varien_Object doesn't have this method. If we're sure that this event always gets an product and never varien_object, then the type annotation should be fixed.

I've commented on https://github.com/OpenMage/magento-lts/pull/1325 because I think it's more complicated than that, we can't just simply remove Varien_Object, the code is very nested in the indexers and it's really hard to check.

@Sekiphp @midlan @kiatng @colinmollenhour

**Comment by colinmollenhour** [comment]
If the code works in v19 it should work in v20, the only difference is the phpstan error, right? So if the indexer is working it is safe to change the vardoc. Changing the vardoc will have no effect on execution it will just allow phpstan to pass.

**Comment by colinmollenhour** [comment]
Why is my PR (#2131) so different from this one? Seems there were some commits missed somewhere along the way?

**Comment by fballiano** [comment]
I've no idea what's happening now:

```
------ ----------------------------------------------------------------------- 
  Line   Catalog/Model/Resource/Product/Attribute/Backend/Urlkey.php            
 ------ ----------------------------------------------------------------------- 
         Ignored error pattern #^Call to an undefined method                    
         Varien_Object\:\:formatUrlKey\(\)\.$# in path                          
         /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Catalog/  
         Model/Resource/Product/Attribute/Backend/Urlkey.php was not matched    
         in reported errors.                                                    
 ------ ----------------------------------------------------------------------- 
```

**Comment by tmotyl** [comment]
@fballiano phpstan complains that you have fixed an issue, but an ignore rule related to it is still present in the ignore list (baseline file). Just regenerate baseline and it should be gone.

**Comment by fballiano** [comment]
> @fballiano phpstan complains that you have fixed an issue, but an ignore rule related to it is still present in the ignore list (baseline file). Just regenerate baseline and it should be gone.

I don't know what it mean sorry! 😅

**Comment by tmotyl** [comment]
@fballiano just run 
```
vendor/bin/phpstan  analyse -c .github/phpstan.neon --generate-baseline .github/phpstan-baseline.neon
```
and there should be one error in .github/phpstan-baseline.neon gone (which is ok).
Commit this change to phpstan-baseline.neon and the job will be green

fballiano:
I regenerated the baseline but I'm not sure about the result actually

fballiano:
@tmotyl I've regenerated with php8 (before it was php7.4), dunno if that makes a difference

tmotyl:
looks ok now, only labbeler fails "Validation Failed. Issues cannot have more than 100 labels: {"value":" so it's not a blocker for a merge

tmotyl (dismissed):

fballiano:
we can maybe keep this PR open until release, I'll keep it updated :-)

tmotyl:
@fballiano lets merge it, so people can start testing the v20 branch before release!

tmotyl:
In the long run, I would love to have every commit merged to 1.9.4.x being merged to v20 right away

luigifab:
There is something wrong on branch 1.9.4.x, at commit 412bee01.
When I press "Reset filter", Sort order of any grid is lost. It's me?

ADDISON74:
@luigifab - that commit is harmless and aims to solve the issue of translating the interface in the Backend in case of Import/Export operations. Please open a new issue and provide more details about reproducing the problem so I can test it. If it is confirmed we can find a fix. In this way we keep this PR without too many comments.

fballiano:
> There is something wrong on branch 1.9.4.x, at commit [412bee0](https://github.com/OpenMage/magento-lts/commit/412bee01b9cc77f50b4333b098e25d116bf9e820) (since when? I don't know). When I press "Reset filter", Sort order of any grid is lost. It's me?

I can see the problem and it is happening on 20.0.14 too (so it was introduced in a previous release)

fballiano:
I've created https://github.com/OpenMage/magento-lts/pull/2252 to solve this

tmotyl (approved):

kiatng (approved):

tmotyl:
In my pov, Ideally we would cherry pick commits from 1.9 to this branch, after each merge to 1.9 branch.
second best option is to merge to 20 (no squashing) as we still want to be able to git bisect to find regressions

fballiano:
ok i merged with no squash and I'll recreate the branch with the same name in order to re-update it in a new PR.

Anyway in my opinion we should have the v20 branch as the default one and backport to v19 upon release.

tmotyl:
Im ok with merging first to 20 and then to 19 but i strongly disagree that the merge should happen only just before release.
Changes should get backported minute after they were merged to 20 (merger might be responsible for doing it)

only then we allow people to use pre-release version and find regressions before the official release

fballiano:
i also would prefer to have every commit immediately merged to the "other" branch, but I don't know how to do that, as far as I know github doesn't have a way to do that, and do a PR for every other PR is not viable so... i don't know..

elidrissidev:
It may be possible to use the [`push` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) or similar to cherry-pick the new commits to `20.0`, worth trying.

midlan:
what about dropping support of `v19` branch? :) seems it is lot of work to keep two active branches

ADDISON74:
I don't use in my installation v1.9.4.x. I would drop all the branches and keep two named openmage and openmage-bc. Who wants to maintain 1.9.4.x can do it the last one. When I make a clone of the magento-lts repository and give the command to show me all the branches it makes me dizzy at how many there are. OM-20 should become the default branch. It is obvious that over time OM-19 has been maintained in order not to break the window with the elephant and OM-20 has incorporated all the news. but the effort seems more than enough.

I know it's a tough decision to make but I notice that we don't have any feedback from the users of this project who don't get their hands dirty with code. If BC will appear when using OM-20 only I am pretty sure it will be solved. We need a platform that doesn't drag the ballast of the past after it, here I have in mind replacing ZF1 with ZF1-Future that I have been using in production for almost 1 month and it generated only a few issues for me.

fballiano:
There are only 4 branches now, I cleaned everything I could already

Schermata 2022-07-01 alle 12 57 27

ADDISON74:
Hallelujah! RIP forever those branches.

fballiano:
I think some of the maintainers told me that the 20.0.x-dev was used for some continous integration or something like that so that we shouldln't remove it without an announcement. which i've no idea how we can announce something, where? how?

Flyingmana:
as a note for the approvers: @kiatng @tmotyl 
this merge accidentally lost the changes from https://github.com/OpenMage/magento-lts/blob/3ba2eadfbdd77db5b66ec42bb8964c97a47d0a60/skin/adminhtml/default/default/boxes.css#L746-L747

did not happen when I reproduced the merge, so not a general git problem

Issue #2196, opened 2022-06-05
OM allows new newsletter subscriptions even though the visitor has been confirmed as a subscriber.


ADDISON74:
OM allows new newsletter subscriptions even though the user has been confirmed as a subscriber. The issue relates to the subscription form, I did not check what if the address is subscribed already and a new account is created in which the subscription box is checked by creator.

**Step 1**: In OM-20.14 Backend I set the following values for Customer Newsletter
- Allow Guest Subscription = Yes
- Need to Confirm = Yes

![config](https://user-images.githubusercontent.com/8360474/172069323-13636bef-1f25-49ff-b91a-102bbe20b878.jpg)

**Step 2**: In OM-20.14 Frontend I filled in an email address in Newsletter input box. After pressing the [Subscribe] button page reloads with the following message "Confirmation request has been sent."

![confirmation_sent](https://user-images.githubusercontent.com/8360474/172069446-d6b2766a-92fc-441d-a338-6949cadec9cb.jpg)

**Step 3**: In guest Inbox I received the newsletter subscription confirmation email. 

![confirmation_received](https://user-images.githubusercontent.com/8360474/172069494-85882de9-3864-403e-8c21-ebda9f257267.jpg)

**Step 4**: After pressing [Confirm Subscription] the Frontend page loads with the following message "Your subscription has been confirmed". Here follows step 4.1 which is solved in PR #2195 which sends the confirmation of the successful subscription.

**Now comes the funny part.**

**Step 5**: I filled in again the same email address in Newsletter input box. After pressing the [Subscribe] button page reloads with the following message "Thank you for your subscription.". This is completely wrong because the email address is already subscribed.

![confirmation_thank_you](https://user-images.githubusercontent.com/8360474/172069532-431c3fcd-b0c5-41b8-8959-7a51da086299.jpg)

**Step 6**: In guest Inbox I received again the newsletter subscription confirmation email.

![confirmation_resent](https://user-images.githubusercontent.com/8360474/172069559-d6d728d3-937e-4d3e-9c0b-63bb3eaccaf0.jpg)

**Step 7**: After pressing [Confirm Subscription] the Frontend page loads with the following message "Your subscription has been confirmed".

![confirmation_confirmed_2](https://user-images.githubusercontent.com/8360474/172069569-9559d0e0-191a-4ef3-86f2-45a6b09c5e12.jpg)

[...]

PR #2195, opened 2022-06-05
Added support for sending subscribe confirm email

fballiano:
In case "newsletter subscribe confirmation" is enabled, it seems that no "subscription success" email is sent.

This code was provided by @hejhog in issue https://github.com/OpenMage/magento-lts/issues/822, I'm just converting it to a PR, which I've not tested.

### Fixed Issues
1. https://github.com/OpenMage/magento-lts/issues/822

### Manual Testing
1. enable newsletter confirmation
2. subscribe to newsletter and check no success email is sent

PR #2214, opened 2022-06-12
Fixed issue #2196, throws error if an email is already a newsletter s…

kiatng:
…ubscriber.

### Description (*)

My test display: 
image

### Related Pull Requests

PR #2195 

1.  Issue #2196

### DIFF
--- a/app/code/core/Mage/Newsletter/Model/Subscriber.php
+++ b/app/code/core/Mage/Newsletter/Model/Subscriber.php
@@ -354,6 +354,8 @@ class Mage_Newsletter_Model_Subscriber extends Mage_Core_Model_Abstract
                 $this->setStatus(self::STATUS_SUBSCRIBED);
             }
             $this->setSubscriberEmail($email);
+        } elseif ($this->getStatus() == self::STATUS_SUBSCRIBED) {
+            Mage::throwException(Mage::helper('newsletter')->__('This email address is already registered.'));
         }
 
         if ($isSubscribeOwnEmail) {
--- a/app/locale/en_US/Mage_Newsletter.csv
+++ b/app/locale/en_US/Mage_Newsletter.csv
@@ -128,6 +128,7 @@
 "There was a problem with the subscription: %s","There was a problem with the subscription: %s"
 "There was a problem with the un-subscription.","There was a problem with the un-subscription."
 "This email address is already assigned to another user.","This email address is already assigned to another user."
+"This email address is already registered.","This email address is already registered."
 "Type","Type"
 "Unconfirmed","Unconfirmed"
 "Unsubscribe","Unsubscribe"


ADDISON74:
Regarding the issue reported by me here #2196. We can also have the following scenario:

1. I fill in the email address to subscribe to the newsletter and send the form

2. I receive a message to confirm the subscription but I choose not to

3. I send again the same email address to subscribe to the newsletter

Obviously I should receive the same message to confirm the subscription (point 2 above). As long as I do not confirm the subscription the store cannot consider it a subscribed one, even if the address exists in the database.

@kiantg - your proposal make changes to Model, @fballiano in controller file. You should agree on the best option.

I remain of the opinion that PR #2195 - Added support for sending subscribe confirm email - should be strictly limited to that issue, namely that no successful subscription confirmation message is sent and this PR should solve the issue reported in #2196, related to the first part of action, subscribing an email address **even it exists in the newsletter_subscriber table but it was not confirmed**. Why separating the PR's? Because it allows us to discuss each of them and make the proper changes for approval., We don't have to make a PR like a Christmas tree, especially when we have now the opportunity to analyze and make a quick decision.

kiatng:
My reasons to perform the check in the model:

1. There is a status validation for new **and existing** subscriber in the `subscribe()` method. If the check for `STATUS_SUBSCRIBED` is required, then it should be here and not in the controller to avoid doing status validation in 2 places.
2. If it is in the controller and it is a new subscriber, then the model is loaded twice, once in the controller, and again in the `subscribe()` method. This can be avoided by performing the status validation in the same method, the model is loaded only once.

ADDISON74 (approved):

kiatng:
> @kiatng - What is happening when I am a registered customer, subscribed already to newsletter, then logout from account and try to subscribe. In this scenario it cannot be considered a Guest subscription. OpenMage should throw an error message not matter it is a registered customer or guest.

Yes, I just tested it to be sure, it throws error if the email exists in the newsletter table regardless it is a registered customer or guest.

Reviewing the code again, the algorithm is:

1. If the configuration of guest subscription is not allowed, it throws error in the controller: https://github.com/OpenMage/magento-lts/blob/dd2c97d4fc3eea08a2ee9b7d60f43bcf4b3f98f3/app/code/core/Mage/Newsletter/controllers/SubscriberController.php#L61-L64
2. If the guest email belongs to another registered customer,  it throws error in the controller: https://github.com/OpenMage/magento-lts/blob/dd2c97d4fc3eea08a2ee9b7d60f43bcf4b3f98f3/app/code/core/Mage/Newsletter/controllers/SubscriberController.php#L66-L72
3. With this PR, if the email exists in newsletter table **and** the status is `STATUS_SUBSCRIBED`, it throws error: _This email is already registered._

fballiano (approved):

PR #2218, opened 2022-06-12
Backend > Manage Products grid - More space for columns

ADDISON74:
The purpose of this PR is to solve a visual issue in the Manage Products grid. By default, the "**Type**", "**Attrib. Set Name**", "**SKU**" and "**Visibility**" columns are not wide enough. Thus the content uses instead of 1 row, 2 or even 3 rows and as an effect the grid increases vertically. 

Please note that for the "Type" and "Visibility" columns I used the longest strings in the dropdown lists in order to find out the proper width, these are "Downloadable Product" and "Not Visible individually".

An image is worth a thousand words. Below are screenshots of the Manage Products grid in OpenMage 20-14 for a few resolutions. The database used in testing is Magento Sample 1.9. The values in columns "Name", "Attrib. Set Name", "SKU", are not short at all but I found that their sizes cover the real situation in some of the stores I manage and where I did the implementation already.: 

**Screen Resolution Width: 960px**

960px-default

**Screen Resolution Width: 1280px**

1280px-default

**Screen Resolution Width: 1437px**

1437px-default

Bellow is what happens if this PR is merged. I personally consider it brings significant visual improvements:

**Screen Resolution Width: 960px**

960px

**Screen Resolution Width: 1280px**

1280px

**Screen Resolution Width: 1437px** - starting with this width all the rows occupy a single row.

1437px

Please note for some columns the values cannot be removed, nor can smaller widths be set (e.g. ID, Price, Qty).

Also, these results were obtained by adding the width of 300px (missing from OM code) for the "Name" column, which is absolutely necessary in this situation. 

Anyone is welcome to play with the width values for these columns to cover the real situation in their stores.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
@@ -154,6 +154,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
         $this->addColumn('name',
             array(
                 'header'=> Mage::helper('catalog')->__('Name'),
+                'width' => '300px',
                 'index' => 'name',
         ));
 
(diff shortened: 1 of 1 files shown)


elidrissidev (approved):
Looks okay for me in 1600px wide screen.

ADDISON74:
It still needs one approval. Who offers to give it?

fballiano (approved):

Issue #2200, opened 2022-06-06
Missing bottom borders in Orders page with Gift Messages allowed


ADDISON74:
Go to Backend > System > Sales. In "Gift Options" section set as follows:

- Allow Gift Message on Order = Yes
- Allow Gift Message for Order = Yes

Create an order in Backend and place it. Take a look to "Items Ordered" grid. Both rows do not have bottom borders. 

![borders](https://user-images.githubusercontent.com/8360474/172254123-a2bb7e7b-abc5-48e8-963d-5510324ad910.jpg)

Let's inspect these rows. As you may see with Gifts allowed <tr> tag doesn't have any class. 

![inspector-1](https://user-images.githubusercontent.com/8360474/172254982-25400aa6-16ac-428e-a619-76eda2f13817.jpg)

When "Allow Gifts Message on/for Order" are set to No, both rows has ```class="border"```. Just create a new order after you set the values to No and see what is happening.

Let's add this class in browser Inspector for the first row and look what happened. It got border bottom.

![inspector-2](https://user-images.githubusercontent.com/8360474/172255284-9a3c2774-a928-4e77-98f4-abb570eb7127.jpg)

In conclusion, when "Allow Gifts" is set to Yes there is a template used in this case but ```class="border"``` is missing from it. I did not have time to identify the template but I will do it soon. It took me some time to figure out what is going on, initially I was convinced it was a missing template or CSS issue but it wasn't.

ADDISON74:
The template file responsible for this issue I found it here

/app/design/adminhtml/default/default/template/sales/order/view/items/renderer/default.phtml

From line 27:

```php
<?php $_item = $this->getItem() ?>
<?php $this->setPriceDataObject($_item) ?>
<tr<?php if (!$this->canDisplayGiftmessage()): ?> class="border"<?php endif ?>>
    <td>
        <?php if ($this->canDisplayContainer()): ?>
        <div id="<?php echo $this->getHtmlId() ?>" class="item-container">
        <?php endif ?>
        <div class="item-text">
            <?php echo $this->getColumnHtml($_item, 'name') ?>
        </div>
        <?php if ($this->canDisplayContainer()): ?>
        </div>
        <?php endif ?>
    </td>
```

As I anticipated border class is not added when Allow Gift Message is set to Yes.

```php
<tr<?php if (!$this->canDisplayGiftmessage()): ?> class="border"<?php endif ?>>
```

[...]

PR #2220, opened 2022-06-12
Missing bottom borders in Orders page with Gift Messages allowed

ADDISON74:
A full report of the issue can be found here #2200.

Go to **Backend> System> Config> Sales> Gift Options**. If the values for "Allow Gift Messages on Order" and "Allow Gift Messages for Order Items" are Yes the if statement from line 27 
```php 
<tr<?php if (!$this->canDisplayGiftmessage()): ?> class="border"<?php endif ?>>
```
excludes border class and the row in the items grid will have no bottom border. If the value is No all rows have bottom border. 

Since the condition in the statement could not be reversed because the effect was inverse affecting the situation when No value is used, I have to remove the if statement completely so that every row in the items grid always has the border class.


ADDISON74:
From what I see this PR that I did directly on GitHub and in a certain branch took over two commits from another PR of mine that was already merged yesterday. The approval in its current state does not create trouble in case someone raises this issue.

fballiano (approved):
I reverted the files that were not part of the PR

kiatng (approved):

Sdfendor (approved):

PR #2250, opened 2022-06-24
Replace deprecated global now() calls with Varien_Date::now()

Sdfendor:
### Description (*)

Global function `now()` and static method `Varien_Date::now()` are equivalent. #1991 accommodated that by deprecating the global function. This PR does two things: 
* Replacing all calls to `now()` with `Varien_Date::now()`
* Where possible a variable is used instead of multiple immediately subsequent calls of the method.

Additionally there is a reformat for an if-conditional in which the method is used.  

### Related Pull Requests

#1991 


fballiano (approved):
LGTM

ADDISON74:
The PR idea is fine with one condition, to have consistency regarding the source code. I notice that either a $now variable or Varien_Date :: now() are used. My proposal is to give up declaring the $now variable everywhere.

Sdfendor:
@ADDISON74 `$now` is used only in cases where the method is called multiple times immediately one after each other (e.g. to fill a 'created_at' field and a 'updated_at' field in the same array to be saved to DB). That made no sense to me.

ADDISON74 (approved):

elidrissidev (approved):

PR #2251, opened 2022-06-25
Fixed problems introduced with PR 1314

fballiano:
PR https://github.com/OpenMage/magento-lts/pull/1314 made the side column of the categories management page (in adminhtml) wider, but doing so broke other pages (like the system->configuration page).

This PR fixes this situation

Screenshot of category page after the patch:
Schermata 2022-06-25 alle 16 00 04

Screenshot of configuration page after the patch:
Schermata 2022-06-25 alle 15 59 58

PS: I've also remove two abandoned and unused gif files (previously used as backgrounds just for a minor shadow effect).


ADDISON74:
Please take a look how it was before and after this change:

before_after

I don't agree removing line 260 and adding "border-left" in line 581 because it doesn't look good:

- the new border doesn't have the same color as before (the one used in the gif file)
- an active tab now has right border. 
- some right margins become bigger, just see the circle areas in the image. I tried to change the border size from 2px to 1px, it looks better, but I don't like the result.

If you get the same result as the initial one I approve it.

fballiano:
This is what I could do and it is consistent with https://github.com/OpenMage/magento-lts/pull/1314 which you also approved. I don't have time to rethink the whole thing.

ADDISON74:
In that PR page I asked for a revert. I agree I approved it but I would not agree changes I don't like.

fballiano:
I'll approve a revert PR if you do that.

PR #2252, opened 2022-06-25
Fix for "reset filter" breaking grid sorting

fballiano:
Pressing "reset filters" on a backend grid also loses the sorting, this PR fixes is so that pressing "reset filters" resets all filters but keeps the sorting (both field and direction).

Go to product grid, search for something, reset filters and note that the sorting is no longer there.

### DIFF
--- a/js/mage/adminhtml/grid.js
+++ b/js/mage/adminhtml/grid.js
@@ -296,8 +296,6 @@ varienGrid.prototype = {
         }
     },
     resetFilter : function(){
-        this.addVarToUrl(this.sortVar, '');
-        this.addVarToUrl(this.dirVar, '');
         this.addVarToUrl(this.pageVar, 1);
         this.reload(this.addVarToUrl(this.filterVar, ''));
     },


ADDISON74:
In OM-20.13 there is no such problem, but in OM-20.14 I managed to reproduce it. If any column in a Backend grid is sorted, pressing the [Reset Filters] button in 20.13 does not happen, but in 20.14 the effects of sorting the column are canceled.

This is from the file modified in 20.13

```js
    resetFilter : function(){
        this.reload(this.addVarToUrl(this.filterVar, ''));
    },

```

Do you think that this line should be deleted to make the code the same as in 20.13 ```this.addVarToUrl(this.pageVar, 1);```?

Do you have any idea how this change was made between versions?

fballiano:
pageVar tells OM to go to the first page, which I think it's correct so I'd leave it :-)

ADDISON74:
I tested and the issue reported by @luigifab is solved. I have to find out what PR after 20.13 changed this file in bad.

The only aspect I am not yet convinced of is related to this return to the first page by resetting the filters. I start from the idea that sorting and navigation in pages are the result of user-set actions. If he resets the filters then I should leave both the sorting and the position on a page. We only leave sorting.

As the code was in 20.13 I think this aspect was taken into account. I would be of the opinion not to return to the first page, but let's listen other opinions too.

fballiano:
If you remove the filters it wouldn't make sense to stay on the same page since the data will be totally different, to me it's better to go to the first page (as it is already)

kiatng (approved):
I agree that the page is reset to 1 when filter is reset. Currently, I always need to go back to page 1.

ADDISON74:
The issue occurred in PR #821 in the 1cfc542779238996d64fd5bf9f6122923b35c6 commit where the first 2 lines were added to the resetFilter function, which resets the sorting and direction of a column bringing the grid to its original shape. In my opinion [Reset Filter] should be strictly limited to resetting the filters, so these two lines were pointless. Regarding the return to the first page that was also added in #821, considering that there are 3-1 opinions for this feature I will approve the change.

ADDISON74 (approved):

Issue #2255, opened 2022-06-25
Store scope hint points to unexisting old M1 documentation


fballiano:
In the backend, the store switcher has a hint:
<img width="357" alt="Schermata 2022-06-26 alle 00 21 19" src="https://user-images.githubusercontent.com/909743/175793323-8f1eb089-b65c-4a38-bae0-a92886c5a664.png">

and the `?` points to http://merch.docs.magento.com/ce/user_guide/configuration/scope.html which doesn't exist anymore.

Do we have a copy of that url somewhere? I was unable to find it.

kiatng:
[Internet archive](https://web.archive.org/web/20170529192915/http://docs.magento.com/m1/ce/user_guide/configuration/scope.html) to the rescue.

ADDISON74:
In what section are seen this? Here is what I am seeing in Manage Products (OpenMage 20.14):

![help_magento](https://user-images.githubusercontent.com/8360474/175813410-576beb06-ba15-448d-b3a0-8c2c30dde172.jpg)

There were discussions before removing those [?] here [#137](https://github.com/OpenMage/magento-lts/issues/1374) and [here https://github.com/OpenMage/magento-lts/issues/1374](https://github.com/OpenMage/magento-lts/issues/1376). 

Finally they were removed from OpenMage by this PR https://github.com/OpenMage/magento-lts/pull/1536. Are we bringing back the help in OpenMage based on Internet archive? I thought we agreed to get rid off it.

fballiano:
I remembered something like that but I see it in my dev environment on 1.9.4.x branch

Sdfendor:
@fballiano The PR mentioned is for the 20.0 branch so it's expectable to see it still in 1.9.4.x.

fballiano:
@Sdfendor yes I later remembered and written https://github.com/OpenMage/magento-lts/pull/2256#issuecomment-1166518605

PR #2256, opened 2022-06-26
Fixed documentation link for "store scope" in store switcher (backend)

fballiano:
Everything is explained in ISSUE https://github.com/OpenMage/magento-lts/issues/2255, this PR only fixes the URL of the documentation that got broken during the years

### DIFF
--- a/app/code/core/Mage/Adminhtml/etc/config.xml
+++ b/app/code/core/Mage/Adminhtml/etc/config.xml
@@ -226,7 +226,7 @@
         </general>
         <hints>
             <store_switcher>
-                <url><![CDATA[http://merch.docs.magento.com/ce/user_guide/configuration/scope.html]]></url>
+                <url><![CDATA[https://web.archive.org/web/20170529192915/http://docs.magento.com/m1/ce/user_guide/configuration/scope.html]]></url>
                 <enabled>1</enabled>
             </store_switcher>
         </hints>


ADDISON74:
All [?] should be removed by now from OpenMage. I posted a comment here https://github.com/OpenMage/magento-lts/issues/2255#issuecomment-1166514972

fballiano:
I don't see it on  20.0.14 but I see in on my dev environment from 1.9.4.x branch, could you confirm?

fballiano:
As you can see in https://github.com/OpenMage/magento-lts/pull/1536 the functionality was only removed in v20 branch, not in v19 :-(

ADDISON74:
I never used those links. Probably like me many others. When Adobe shut down the Magento help website a discussion ensued to establish either the removal of those links or a new website to link to. In my opinion, the effort to maintain this help is not justified, so the solution to remove those links from OM was the best solution. 

Indeed it was only wanted to be merged in OM-20 but I don't see the point in OM-19 either. If this PR goes into OM-19 it will have to be excluded from OM-20. A personal opinion would be to remove those [?] from the OM-19 too, as long as they no longer have valid links, To make an effort to update even the Internet Archive is not worth it. If someone still wants to get help they can download the Magento 1 User Manual, but after so many years I don't think there are any novices in using the interface.

fballiano:
I guess it was decided to merge in v20 only cause it could be a small breaking change.

It's one like, so it won't be a hard conflict to solve (also I'll solve it in my v19->v20 merge PR anyway)

ADDISON74:
I do not use OM-19, as long as this feature does not return in OM-20 does not affect me in any way. If several decide to keep Help links in OM-19, then this PR must be approved.

If we got that website from Magento.com that had all the help sections and hosted it somewhere, it wouldn't have been necessary to remove it from the OM-20. We were replacing the URL and the functionality still exists today.

fballiano:
It won't come back to v20, the patch is a single line of xml, can't add anything also if it wanted to :-D At the same time we can't just leave v19 broken.

You know I'm all about disruptive change, if it was for me many more things would be removed...

ADDISON74:
We should ask those who read our posts if there is still the possibility to get this website http://merch.docs.magento.com/ce/user_guide/. Last year we removed the feature without analyzing this aspect. If we get it we would talk to @Flyingmana to host it on openmage.org. Once this is done we can do a revert in OM-20 of that PR then change the URL in file /app/code/core/Mage/Adminhtml/etc/config.xml.

If we can't get it, the Internet Archive is not a solution because provides an incomplete version of the user guide. Rather than using it remove it from 19 seems to be a fair solution.

fballiano:
I wouldn't re-enable this thing in v20, if a user needs a guide it reads the guide, we don't need to have links to the guide all over the backend.

ADDISON74:
This is a screeshot from the latest OM-19 version:

help

In my opinion it is useless to add that link to the Internet Archive to solve the documentation for "Store Scope" since in OM-19 there is still a major issue related to the documentation of each section. In the absence of that website identical to the one offered in the past by Magento, keeping this online help is useless. It was removed from the OM-20 and I understand that it is not wanted to be brought back even if we find the original Magento website, so it must be removed from the OM-19 without any resentment.

luigifab (approved):

ADDISON74:
Why are you approving this PR? On one hand it fixes the store scope with a link to IA and the loaded page doesn't look good. But on the other hand all help links are still there without proper links.

The help feature must be removed from OM-19 as we did for OM-20. We should close this PR.

fballiano:
I think it should be fixed anyway or removed but it wasn't removed because it was BC (if you check the original thread). so at the moment let's merge it at least we don't have a broken link in all v19 installations :-)

ADDISON74:
We worry too much about BC which obviously involves a lot of effort to actually maintain a custom version of the OM-20.

This PR solves the help issue for the store scope but the issue of the other help links remains and it cannot be solved because there is no Internet Archive pages for them and no clone of the Magento website. This is why I said that the change is useless. I am wondering how many will click on that icon at least out of curiosity. They will most likely cry because they do not see the icon they are used to.

fballiano:
as you know i would remove a lot of things from v19 too, but not eveybody things the same, what can i tell you.

ADDISON74:
Today someone asked me to get rid of the Magento help links because none of them work. I remembered our discussions from here, but considering that I was not allowed to upgrade to OpenMage and I did not have time to search through files, I edited the CSS file and made a display: none for all. Obviously the code for them remained but I solved his visual issue quickly. In addition he was glad that he got rid of them :)

PR #2258, opened 2022-06-26
Removed module Mage_PageCache

fballiano:
The Mage_PageCache was a half baked first attempt at pagecache for community edition.
As far as I can see it lacks basic features like holepunching and support for modern technologies like redis.
It only supports Zend_Server which probably doesn't even exist anymore.

PS: I also found out that, just having it enabled, broke my varnish/turpentine.


ADDISON74:
This module is related to "External Full Page Cache Settings" section in Backend. It is Enable by default, but it works only if you enable an external cache. See the image bellow:

page_cache

The only "External Cache Control" option available is "Zend Full Page Cache". Then you are able to flush the external page cache from "Cache Storage Management" page:

external page cache

I never use this module and I consider it far inferior to the Varnish-based Turpentine extension. If anyone is using it please let us know. If there is no feedback in 2 weeks I will approve it on my behalf.

ADDISON74:
You have to mention the removal of this module from OpenMage in the README.md file. There is a section named "Removed Modules".

ADDISON74:
From what I notice we have no opinions to keep this module in OpenMage. Therefore I don't have to wait two weeks to approve the PR. I clicked the button. 

There is a conflict related to an XML that must be resolved.

ADDISON74 (dismissed):

fballiano:
fixed the conflict and a typo ;-)

ADDISON74 (approved):

tmewes (dismissed):

Flyingmana:
would need to check, but I think a php based fullPage cache I build 7y ago could have used this as base.

but guess removing it ks ok, if no modern module uses parts of it.

fballiano:
I'd just suggest to use Lesti :-)

Flyingmana:
> I'd just suggest to use Lesti :-)

mine might have been less flexible, but I got constantly ~40-60ms response times on cached pages. Is it comparable in performance?

fballiano:
No they're much higher (at least on the production website I'm using it in) but it's probably because of the holepunching dynamic blocks

fballiano:
fixed phpstan

ADDISON74 (approved):

fballiano:
fixed conflict in README

colinmollenhour (approved):
Agreed, this page cache when I reviewed it appeared to be almost unusable.

ADDISON74 (approved):

PR #2259, opened 2022-06-26
Removed Mage_Backup module

fballiano:
Mage_Backup was disabled by "the old management" years ago, this PR removes it and clean all its files (and related ones).

Note that this PR also solves a big problem, when deleting a website there's no confirmation prompt ask, press the button and the website is gone. This PR adds the confirmation prompt back.


sreichel:
Some people still use it (re-enabled).

fballiano:
In my opinion this module is useless in 2022, with big catalogs it WILL crash your servers (and I saw it happen multiple times) and it was deactivated for a reason, it was inactive for years and nobody complained so... ;-)

justinbeaty:
Very nice work @fballiano! I haven’t tested the PR, but I agree with its removal.

Perhaps a compromise is that we put it into a separate repo with a modman file for it to be installable. We could even install it into the release zip with the build tool so it stays available unless you use composer and don’t specifically require it.

fballiano:
It is hard to make it an external module, it had parts that were hardcoded in the store management masks.

It was disabled during a security patch, probably because they didn't have time to remove it completely (because of the hardcoded pieces in the core). I don't know precisely why but it seems to me that the original authors evaluated that it wasn't safe or something.

Flyingmana:
> In my opinion this module is useless in 2022, with big catalogs it WILL crash your servers (and I saw it happen multiple times) and it was deactivated for a reason, it was inactive for years and nobody complained so... ;-)

this is an opinion, but reality sets facts, and the fact is there are people out there using it.
You might find some of them who complained in this project already about it being disabled.

fballiano:
people use every kind of things, that doesn't make it right and a project should have a direction that doesn't necessarily need to accomodate every kind of weirdness

fballiano:
And I think it's profoundly wrong to keep it in the core, disabled.
It has to be enabled or out of the way.

fballiano:
ok it's broken, i'll close it

ADDISON74:
Copied from here https://github.com/OpenMage/magento-lts/discussions/2024#discussioncomment-3066042.

I tried to understand why this module was disabled in 1.9.4.0, given that backing up this way from Backend was a handy solution for small and medium-sized stores. The Magento team made only a brief statement: "It no longer meets compliance requirements. Saved backups should be encrypted".

Instead of finding a solution to this issue they chose to disable it. Over time starting with November 2018 there have been discussions that if it is enabled it creates a security hole in Magento. This created a false impression about this module, when in fact the issue was completely different. If someone gets backups they can use the information, some of them confidential about sales, customers.

If what is saved by this module is secured and protected from being accessed from outside then there would be no issues. Whoever has access to a server will not be interested in the files located in /var/backup when he can access any information in the file system and database.

This module can be left alone in OpenMage and if data security over backups is ensured it can be enabled by those who need it. Reading the forums of the time there were enough who missed it. If you choose to remove it please offer in README.md alternative solutions for making backups. At that time Magento showed a couple but the link is no longer available and it doesn't exist in the Internet Archive.

**Addendum**: What I wrote above refers to a small store where Backup/Restore feature can be done through Backend. I have never tested the performance of this module in the case of a large store. Every time I used the command line when I manually saved data or restored them. If the database is corrupted there is little chance that the Backend will still work.

fballiano:
I kinda broke the PR trying to fix the conflicts and, since the consensus was that nobody wants it removed I'll leave it as it is.

PR #2261, opened 2022-06-28
Flush Magento Cache - button and message changes

ADDISON74:
This PR replaces Magento word with OpenMage for [Flush Magento Cache] actions in Backend > System > Cache Management. 

Here is the result:

flush_cache

PS - I know that some will say that it is better to make the change only in the translation file. This time let's do it as it should by modifying the PHP files.


fballiano (approved):
Since this is a quite important string I agree it should be changed in the PHP files too ;-)

kiatng (approved):

PR #2264, opened 2022-06-29
GUI to edit "Category Attributes"

fballiano:
This PR give backend users the possibility to create/edit/delete "category attributes".

This feature was developed mostly using the code of the "product attributes" edit feature, that's why I've copied the license in the file header.

Notes:
- It's a big PR so suggestions/fixes/additions are welcome.
- We'd probably need to block the user from editing/deleting "system" attributes.
- At the moment the newly created attributes will not appear in the "category edit" masks nor on the frontend.

1. open the backend
2. open the menu "Catalog -> Attributes -> Manage Category Attributes"
3. Add, edit or delete category attributes

### Screenshots

The new menu element
Schermata 2022-06-29 alle 13 32 38

Category attributes grid
Schermata 2022-06-29 alle 13 32 52

Add a new attribute
Schermata 2022-06-29 alle 13 33 00

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
@@ -1,28 +1,4 @@
 <?php
-/**
- * Magento
- *
- * NOTICE OF LICENSE
- *
- * This source file is subject to the Open Software License (OSL 3.0)
- * that is bundled with this package in the file LICENSE.txt.
- * It is also available through the world-wide-web at this URL:
- * http://opensource.org/licenses/osl-3.0.php
- * If you did not receive a copy of the license and are unable to
- * obtain it through the world-wide-web, please send an email
- * to license@magento.com so we can send you a copy immediately.
- *
- * DISCLAIMER
- *
- * Do not edit or add to this file if you wish to upgrade Magento to newer
- * versions in the future. If you wish to customize Magento for your
- * needs please refer to http://www.magento.com for more information.
- *
- * @category    Mage
- * @package     Mage_Adminhtml
- * @copyright  Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
- * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
- */
 
 class Mage_Adminhtml_Block_Catalog_Category_Attribute extends Mage_Adminhtml_Block_Widget_Grid_Container
 {
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * Magento
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@magento.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade Magento to newer
+ * versions in the future. If you wish to customize Magento for your
+ * needs please refer to http://www.magento.com for more information.
+ *
+ * @category    Mage
+ * @package     Mage_Adminhtml
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+class Mage_Adminhtml_Block_Catalog_Category_Attribute extends Mage_Adminhtml_Block_Widget_Grid_Container
+{
+    public function __construct()
+    {
+        $this->_controller = 'catalog_category_attribute';
+        $this->_headerText = Mage::helper('catalog')->__('Manage Category Attributes');
+        $this->_addButtonLabel = Mage::helper('catalog')->__('Add New Attribute');
+        parent::__construct();
+    }
+}
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
@@ -60,7 +60,7 @@ class Mage_Adminhtml_Block_Catalog_Category_Attribute_Edit extends Mage_Adminhtm
         $this->_updateButton('save', 'label', Mage::helper('catalog')->__('Save Attribute'));
         $this->_updateButton('save', 'onclick', 'saveAttribute()');
 
-        if (! Mage::registry('entity_attribute')->getIsUserDefined()) {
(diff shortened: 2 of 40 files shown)


ADDISON74:
... another important feature missing from Magento. I like such modules that improve functionality. Don't stop your PR's here :)

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade Magento to newer
+ * versions in the future. If you wish to customize Magento for your
+ * needs please refer to http://www.magento.com for more information.
+ *
+ * @category    Mage
+ * @package     Mage_Adminhtml
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
```
Flyingmana, line 23: you should leave out the copyright and license note, or maybe just the whole file level docblock

fballiano: done

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
```
luigifab, line 2: You have forgotten the copy comment (you have added it elsewhere).

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
```diff
@@ -0,0 +1,93 @@
...
+{
+
+    public function __construct()
+    {
+        $this->_objectId = 'attribute_id';
+        $this->_controller = 'catalog_category_attribute';
+
+        parent::__construct();
+
+        if($this->getRequest()->getParam('popup')) {
```
luigifab, line 37: You have forgotten a space.

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
```diff
@@ -0,0 +1,93 @@
...
+                    'class'     => 'save'
+                ),
+                100
+            );
+        }
+
+        $this->_updateButton('save', 'label', Mage::helper('catalog')->__('Save Attribute'));
+        $this->_updateButton('save', 'onclick', 'saveAttribute()');
+
+        if (! Mage::registry('entity_attribute')->getIsUserDefined()) {
```
luigifab, line 63: I think you have added a extra space.

app/design/adminhtml/default/default/template/catalog/category/attribute/js.phtml
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * @category    design
+ * @package     default_default
+ * @copyright   Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
+ * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+ */
+?>
+<script type="text/javascript">
+//<![CDATA[
+function saveAndContinueEdit(){
```
luigifab, line 29: You have forgotten a space.

luigifab (changes requested):
Thank you so much for this missing feature. I will test it at production level and deploy it!

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
fballiano: fixed!

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute/Edit.php
fballiano: fixed (it's in the original catalog module, that's why ended up here)

app/design/adminhtml/default/default/template/catalog/category/attribute/js.phtml
fballiano: reformatted the whole file, it was a mess, but it was copied from the catalog module so... I thought it would be better to have the same file instead of different ones

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Attribute.php
fballiano: are you referring about the copyright? I was told by @Flyingmana (it's the comment above yours) to remove it because new files (it they're not copied from old ones) shouldn't have the old copyright

fballiano: are you referring about the copyright? I was told by @Flyingmana (it's the comment above yours) to remove it because new files (it they're not copied from old ones) shouldn't have the old copyright

luigifab: oh... ok.

justinbeaty:
Very interesting feature! Pulled into local env and playing around with it.

justinbeaty:
I added some changes to your branch @fballiano here: https://github.com/fballiano/openmage/pull/2

Mainly, I copy pasted the Attribute Set editor so you can actually see the custom attributes on the category edit page.

image

image

image

image

Still probably needs work and testing before this feature hits a release, but I found it interesting enough to help working on. For example, should users be able to even create a different attribute set for categories? If so, then they should be able to set a category to a different set, but what does this accomplish?

One more note, maybe consider that we should update the database with this:

```sql
UPDATE eav_attribute SET frontend_input = 'boolean' WHERE source_model = 'eav/entity_attribute_source_boolean' AND entity_type_id = 3;
```

because right now many system attributes are using select when it should be boolean. So, it shows the option to "Manage Options" of those attributes when it should not be possible.

fballiano:
Thanks!! I'll check it out ASAP!

justinbeaty:
I kind of wonder if we can abstract this code and re-use for all EAV entities?

```
mysql> select entity_type_id,entity_type_code from eav_entity_type;
+----------------+------------------+
| entity_type_id | entity_type_code |
+----------------+------------------+
|              3 | catalog_category |
|              4 | catalog_product  |
|              7 | creditmemo       |
|              1 | customer         |
|              2 | customer_address |
|              6 | invoice          |
|              5 | order            |
|              8 | shipment         |
+----------------+------------------+
8 rows in set (0.00 sec)
```

fballiano:
a generic EAV module would be possibile and nice, I didn't want to use the "catalog" classes for that not to couple this new functionality to the catalog module, but if we do it from scratch in a generic way and remove the current implementation from the catalog module... that'd be awesome (althought removing the current one should be considered BC so v20...)

fballiano:
also customer attributes have the same things although different, like to show an attribute in a particular form (subscribe in checkout or others), so I that may be needed for all the different EAV components.

anyway I was referring to extending the "catalog module" classes, that would be kinda wrong because it would make the EAV dependant on catalog.

justinbeaty:
Agreed on not Eav should not depend on Catalog, that's definitely wrong. I think I have a pretty good abstraction which I'll push as a commit on my personal branch soon today and link it here.

Also yes about used_in_forms part. What I think is best then is to have the base code in Mage_Eav and then extend it in Mage_Customer to add such functionality. I probably won't tackle making the products attribute code in Mage_Catalog extend these new base classes in Mage_Eav, but I suppose that could be done.

justinbeaty:
@fballiano I just pushed my progress as a new PR #2317 to make it easier for comments or reviews. I *think* this general purpose solution is the way to go, but am interested in your feedback.

PR #2271, opened 2022-07-01
Removed lib/flex

fballiano:
I've searched the whole source tree but I couldn't find any place where these files were used. They seem to be ActionScript files used in the very old file uploader that was replaced years ago, although I'm not sure.

From my tests, removing this folder didn't have any impact.


luigifab (approved):

luigifab:
Already removed at dev env since some years.

tmewes (approved):

midlan (approved):

PR #1024, opened 2020-06-04
Add robots.txt.sample

simbus82:
The old magento has never provided a robots.txt to guide search engine through the multiple urls that a magento installation can generate.

This robots.txt was born from a some years of experience in SEO and from several readings in the most famous "blogs" on magento (and SEO too).

The goal is to ensure that the search engine does not consume all the crawling budget on pages that do not make sense being indexed.
In the same way it avoids that numerous CMS assets are still crawled and indexed.

The search engine is very smart, therefore it is unlikely that once a assets of the core of a cms has been scanned it will be proposed in the SERP.

But even better, we will prevent the search engine from going to waste time on files or urls which will then be rejected anyway.

E-commerce sites with a large number of products and complex layered navigation systems should benefit, reducing the time to index their pages and allowing the search engine to crawl the site more often and more easily.

Bonus: the robots.txt file is the ideal file where to enter site's sitemap path.
Almost all search engines consider the indications entered in robots.txt, so the presence of sitemap.xml will be very important for let search engines crawl first all our most important pages.

PS: PR as Draft, because i don't know if i need to write elsewhere the presence of two new files...

PR #2272, opened 2022-07-02
Update robots.txt.sample

ADDISON74:
This is an update to the robots.txt.sample file that was recently added to OpenMage (PR #1024). 

A short brief of the changes:

- added the Crawl-delay option for the situation in which a good crawler could create traffic issues.
- added a few URL paths that I found in frontend pages such as checkout, sales
- added the option to disallow indexing the media for known bots

The file can be used in production without any issues. Both the initial version and this updated version have been tested by me for a long time. You have to take in consideration that only a good bot will take this file into account. For bad bots, those which skip the robots.txt file content and crawling everything, I will post a new PR these days.

### DIFF
--- a/robots.txt.sample
+++ b/robots.txt.sample
@@ -1,8 +1,8 @@
 ############################################
-## For more information about the robots.txt standard, see:
+## For more information about the robots.txt standard visit:
 ## http://www.robotstxt.org/orig.html
 ##
-## For syntax checking, see:
+## For syntax checking visit:
 ## http://tool.motoricerca.info/robots-checker.phtml
 
 ############################################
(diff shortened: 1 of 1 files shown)


fballiano:
I've added some comments above

ADDISON74:
@fballiano - I don't see any comments. Could you please point me where to look for?

fballiano:
Schermata 2022-07-04 alle 16 23 16

ADDISON74:
On this PR page your comments are not displayed, at least that is the content displayed to me in 2 different browsers. If you didn't show me the screenshot I would never have found them.

Those good and well-known crawlers only collect images, not text content. All lines are commented so that they do not apply. If someone wants to activate any of them, he can modify each line according to the own needs. By default it does not index anything, but you can set for /media or any other directory where there are images.

fballiano (approved):
I'm not really happy about the "disallow /" for media bots, anyway..

fballiano (approved):

PR #2273, opened 2022-07-03
JSON-RPC API support

fballiano:
This PR adds support for JSON RPC protocol for API calls.

- The endpoint for API calls is https://openmagehost/api/jsonrpc
- It uses the same username/roles you'd use for SOAP
- It uses the same "calls" you'd use in SOAP v1 or XML-RPC like explained here https://devdocs-openmage.org/guides/m1x/api/soap/introduction.html

I tested these APIs with postman using this "body" for the login method
Schermata 2022-07-03 alle 15 06 35

and this "body" for the "call" method
Schermata 2022-07-03 alle 15 09 28


fballiano:
PHPStan fails for reasons unrelated to this PR.

app/code/core/Mage/Api/Model/Server/Adapter/Jsonrpc.php
```diff
@@ -0,0 +1,110 @@
+<?php
+/**
+ * Magento
+ *
+ * NOTICE OF LICENSE
```
shyim, line 5: btw you removed that line in one PR for all files. maybe don't add it here 🤔

fballiano: hahahaa right, this PR was much older than the other one :-) I'll change this one if the other gets approved :-) but thanks for the comment so that I won't forget!

btw I'm positive this PR will stay unapproved forever...

FredericMartinez (approved):

empiricompany (approved):

empiricompany (approved):

kiatng (dismissed):
I have used a variation of this which enabled soap_v2 style calling.

app/code/core/Mage/Api/controllers/JsonrpcController.php
kiatng, line 1: I think this controller is not used since it is taken care of by `api.php` in the root: https://github.com/OpenMage/magento-lts/blob/main/api.php

kiatng: I have been using this in nginx:

```bash
  location /api {
    rewrite ^api/([a-z][0-9a-z_]+)/?$ api.php?type=$1 last;    
    #rewrite ^/api/v2_soap /api.php?type=v2_soap last;
    #rewrite ^/api/soap /api.php?type=soap last; 
    #rewrite ^/api/rest /api.php?type=rest last;
    #rewrite ^/api/xmlrpc /api.php?type=xmlrpc last;
    rewrite ^/api/jsonrpc /api.php?type=jsonrpc last;
  }
```
But I do not know why the standard way is to route api calls to `api.php` instead of `index.php`.

kiatng (approved):
My test script:
```php
    public function jsonrpcAction()
    {
        $result = [];
        $url = Mage::getStoreConfig('muze/stars/url');
        $client = new Krixon_JsonRpc_Client($url);

        $sessionId = Mage::getSingleton('customer/session')->getData('jsonsessionKey');
        if (!$sessionId) {
            $apiUser = Mage::getStoreConfig('muze/stars/user');
            $apiKey = Mage::getStoreConfig('muze/stars/key');
            $session = $client->call('login', [$apiUser, $apiKey]);
            if ($session instanceof stdClass) {
                if (property_exists($session, 'result')) {
                    $sessionId = $session->result;
                    Mage::getSingleton('customer/session')->setData('jsonsessionKey', $sessionId);
                } else {
                    Mage::getSingleton('customer/session')->unsetData('jsonsessionKey');
                    return $this->_echo($session, 'cannot login');
                }
            } else {
                return $this->_echo($session, 'unexpected response');
            }
            $result['new_session'] = $sessionId;
        } else {
            $result['session'] = $sessionId;
        }

        /** @var stdClass $std */
        $std = $client->call('call', [$sessionId, 'magento.info', []]);
        if (property_exists($std, 'error')) {
            if ($std->error->code === -32002) {
                Mage::getSingleton('customer/session')->unsetData('jsonsessionKey');
            }
        }
        $result['magento.info'] = $std;

        $this->_echo($result);
    }
```

Result:
```
 array(2) {
  ["new_session"] => string(32) "a3253324b025fcd5789659c2f75aaa06"
  ["magento.info"] => object(stdClass)#168 (3) {
    ["result"] => object(stdClass)#81 (3) {
      ["magento_edition"] => string(9) "Community"
      ["magento_version"] => string(7) "1.9.4.5"
      ["openmage_version"] => string(7) "20.0.16"
    }
    ["id"] => string(13) "64da3c21b96ef"
    ["jsonrpc"] => string(3) "2.0"
  }
}
```

empiricompany (approved):

ADDISON74 (approved):

medeirosleandro:
Please credit the source repository @gamuzatech

eneiasramos:
This implementation is very similar to my module:

https://github.com/gamuzatech/gamuza_jsonapi-magento

ADDISON74:
My opinion is that your PR solves an issue reported in OpenMage. It was suggested by a person I have not met so far in discussions that another author should be credited for this PR. The source shown does not resemble your implementation, I doubt that you analyzed it because you had to spend time to see if it solves the issue, then adapted to OpenMage not as a module but directly in the existing code. From a legal perspective, if we look at the copyright applied to the source, the situation is resolved.

Before coming with legal arguments, I ask kindly the person @medeirosleandro who made the accusation that the code is taken from another repository to come with clear evidence in this regard. If it is only about the fact that there is someone else with an implementation, then the initial statement was wrong because we have to discuss about stealing code. Then we decide whether to credit the author or not.

fballiano:
this PR contains 3 files:
1. https://github.com/gamuzatech/gamuza_jsonapi-magento/blob/master/app/code/local/Gamuza/JsonApi/controllers/JsonController.php doesn't really seem like the one in this PR, there's 1 line which looks the same and that's because it's exactly equal to the one in the core https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Api/controllers/XmlrpcController.php
2.  https://github.com/gamuzatech/gamuza_jsonapi-magento/blob/master/app/code/local/Gamuza/JsonApi/Model/Server/Adapter/Json.php, it's obviously similar to mine because, again, it's exactly the same as https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Api/Model/Server/Adapter/Xmlrpc.php
3. 3rd file is a 3 lines xml config file, which again, the 3 lines are exactly copied from the 3 lines right above.

ADDISON74:
@fballiano - You don't have to explain, but the one who made the accusation, which seems serious to me.

colinmollenhour:
I don't see any evidence that the gamuzatech repo was used as a source - same idea, simple feature, same outcome. Given that you're implementing an interface there is basically only one way to do it and JSONRPC is an open standard..

If @medeirosleandro or @eneiasramos want to push it further I think they need to make a stronger case than just having done it first. I made a JSONRPC adapter myself for a private project back in February of 2013 which pre-dates the Gamuza one - it really wasn't hard, will gladly share it if needed but I think that is not necessary.

elidrissidev:
I'd say it resembles the core `Xmlrpc` code more than the other repo, the yoda conditions and long array syntax refactoring was done after this PR was created.

Besides, there's only so many ways to implement such functionality, so having some resemblance between them isn't out of the ordinary.

PR #2289, opened 2022-07-08
Create/Update .htaccess file in skin and media directories

ADDISON74:
1. I updated the file format for /media/.htaccess.

2. I created .htaccess file in /skin directory. This change doesn't permit executing CGI scripts like in /media and /js (OM-20) directories.

### DIFF
--- a/media/.htaccess
+++ b/media/.htaccess
@@ -4,15 +4,11 @@ Options All -Indexes
 ## Disable PHP code execution
 
 <IfModule mod_php7.c>
-
     php_flag engine 0
-    
 </IfModule>
 
 <IfModule mod_php8.c>
-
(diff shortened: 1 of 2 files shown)


fballiano:
I think (I guess @luigifab would too) that we shouldn't have all those empty lines :-D

ADDISON74:
I agree in other files such as css, js, php, but here we need visibility. I kept the same format as in the .htaccess located in /root where I will propose a larger change these days. The new PR will be on the same format

fballiano:
I see, although I still have a different opinion

fballiano (approved):

elidrissidev (approved):

PR #2294, opened 2022-07-08
Merge v19 to v20

fballiano:
Forward port all 1.9.4.x commits into v20.

Note: I fixed a small conflict generated by https://github.com/OpenMage/magento-lts/pull/2256

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,29 @@ git add -A && git commit
 
 [More Information](http://openmage.github.io/magento-lts/install.html)
 
+## Secure your installation
+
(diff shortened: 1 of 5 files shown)


elidrissidev:
The PHPStan error is originating from my last PR 🙈 but it didn't show up because the composer error made the workflow not run fully.

elidrissidev:
The merge commits make the history look really ugly. Maybe we could instead start rebasing PRs that sync 19 with 20 branches if that won't cause another problem?

fballiano:
this is what github does when:

Schermata 2022-07-15 alle 22 27 55

fballiano:
mmm i kinda don't know how to avoid the "merge commits" so I'll close this and reopen it when I'll be able to fix that properly.

PR #2302, opened 2022-07-10
add ReturnTypeWillChange to various Files catched by code style checker

Flyingmana:

1. ...
2. ...

### DIFF
--- a/lib/Mage/Backup/Filesystem/Iterator/File.php
+++ b/lib/Mage/Backup/Filesystem/Iterator/File.php
@@ -46,6 +46,7 @@ class Mage_Backup_Filesystem_Iterator_File extends SplFileObject
      *
      * @return string
      */
+    #[ReturnTypeWillChange]
     public function current()
     {
         return $this->_currentStatement;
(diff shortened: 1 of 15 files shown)


fballiano:
IMHO we should stop touching lib/Zend

Flyingmana:
its either this or accepting we cant reach compatibility with future php versions without replacing it fully with an alternative.

fballiano:
zf1-future is php8.1 compatible. actually i'm testing OM19 on php8.1 with no problems

ADDISON74:
I confirm that ZF1-Future is compatible with PHP8.1. In my case I had to keep the 2 files in /app/code/core/Zend/Db so I could run OM.

Not only this compatibility should interest us, but also other aspects that are maintained by those who strive to continue with this framework. For example, in OM the TLS protocol is supported up to version 1.0 while ZF1-Future has support for version 1.2. I tested extensions that no longer work in the our current configuration, but that work without issues with ZF1-Future.

Sdfendor (approved):

tmewes (approved):

PR #2307, opened 2022-07-10
Update robots.txt.sample

ADDISON74:
This is an update to the robots.txt.sample file that was recently added to OpenMage (PR #1024). 

A short brief of the changes:

- added the Crawl-delay option for the situation in which a good crawler could create traffic issues.
- added a few URL paths that I found in frontend pages such as checkout, sales
- added the option to disallow indexing the media for known bots

The file can be used in production without any issues. Both the initial version and this updated version have been tested by me for a long time. You have to take in consideration that only a good bot will take this file into account. For bad bots, those which skip the robots.txt file content and crawling everything, I will post a new PR these days.

### DIFF
--- a/robots.txt.sample
+++ b/robots.txt.sample
@@ -34,10 +34,9 @@ Disallow: */catalog/product_compare/
 Disallow: */catalog/category/view/
 Disallow: */catalog/product/view/
 Disallow: */catalog/product/gallery/
-Disallow: */catalogsearch/advanced
-Disallow: */catalogsearch/result
-Disallow: */catalogsearch/seo_sitemap
-Disallow: */catalogsearch/term/popular
+Disallow: */catalogsearch/
+#Allow: */catalogsearch/seo_sitemap
+#Allow: */catalogsearch/term/popular
 Disallow: */checkout/
 Disallow: */control/
 Disallow: */contacts/
--- a/robots.txt.sample
+++ b/robots.txt.sample
@@ -1,12 +1,12 @@
 ############################################
-## For more information about the robots.txt standard, see:
+## For more information about the robots.txt standard visit
 ## http://www.robotstxt.org/orig.html
 ##
-## For syntax checking, see:
+## For syntax checking visit
 ## http://tool.motoricerca.info/robots-checker.phtml
 
 ############################################
-## Crawl the Sitemap. Set the correct url before uncomment
+## Crawl the Sitemap. Set the correct URL before uncomment
 
 # Sitemap: https://example.com/sitemap.xml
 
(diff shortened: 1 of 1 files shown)


fballiano:
mmmm I'm not sure about these ones

Schermata 2022-07-11 alle 00 33 39

I think I preferred how it was before

ADDISON74:
I'll add it and let you know.

ADDISON74:
```Disallow: */catalogsearch/``` matches the next four green lines. Let's analyze each line and decide if we keep any of them.

- **advanced and result** - definitely we do not want the indexing of these pages by good bots. I deactivated advanced search in my stores.
- **term/popular** - although in all my stores I have deactivated this section too, there are others that have used the page and want it to be indexed.
- **seo_sitemap** has two sections categories and products. In my case I disabled it as long as a sitemap is set in robots.txt, but there could be advantages if the sections are indexed.

If I keep only the red line I have no control over the others. I would keep the red line and the last two green ones. If I want access to the term/popular and seo_sitemap pages I can uncomment. The proposed variant is the following:

```php
Disallow: */ catalogsearch /
#Allow: */catalogsearch/seo_sitemap
#Allow: */catalogsearch/term/popular
```

fballiano:
This PR puts in "dissallow" also "catalogsearch/seo_sitemap" which seems to be seo related. but anyway I'd prefer to leave all of catalogsearch/ out of the search engines

robots.txt.sample
```diff
@@ -15,26 +15,38 @@
...
 
 Disallow: */index.php/
 Disallow: */catalog/product_compare/
 Disallow: */catalog/category/view/
 Disallow: */catalog/product/view/
 Disallow: */catalog/product/gallery/
-Disallow: */catalogsearch/
+Disallow: */catalogsearch/advanced
+Disallow: */catalogsearch/result
+Disallow: */catalogsearch/seo_sitemap
```
Flyingmana, line 39: this file is needed for google (and others) to easy find crawl-able URLs. Preventing them to crawl this removes the whole purpose of this endpoint.

robots.txt.sample
```diff
@@ -59,3 +71,32 @@ Disallow: /*?q=*
...
+# Disallow: */media/
+# Disallow: */skin/
+
+# User-agent: msnbot-media
+# Disallow: /
+# Disallow: */media/
+# Disallow: */skin/
+
+# User-agent: Googlebot-Image
+# Disallow: /
```
Flyingmana, line 95: some people might rely on their product and other images being available via image search

robots.txt.sample
fballiano: tested on 2 of my stores and that endpoint goes to 404... i'm not sure it exists

Flyingmana (approved):
approving, because in the end its a sample file and in the end there is nothing changed which is clearly wrong

robots.txt.sample
Flyingmana: which brings up the question, what links to it :/

fballiano: exactly, that's why I should stay `Disallow: */catalogsearch/` instead of the 3 split lines

robots.txt.sample
ADDISON74: This is the main reason those line are commented from the very beginning. But there are situations when you have a website and maybe you don't want your images to be indexed even by Googlebot-Image. in this case you uncomment on those lines.

robots.txt.sample
ADDISON74: to make the change by removing those 4 lines is no problem. but keep in mind that so far OM has not provided a robots.txt file. who has Frontend links to / catalogsearch / seo_map and / catalogsearch / term / popular when using this file in its previous form with the only option `` `Disallow: / * / catalogsearch /` `` will remove the indexing of these two pages. obviously we are talking about bots that take into account the existence and content of this file.

informatively, I analyzed for a month all the bots that went through my websites and I can say that very few come in contact with robots.txt. GoogleBot ignores Crawl-delay.

ADDISON74:
I made the change by removing two lines. The other two remain under discussion.

fballiano (approved):

Flyingmana (approved):

PR #2308, opened 2022-07-10
Addendum to PR 2087

ADDISON74:
@luigifab made a few remarks here https://github.com/OpenMage/magento-lts/pull/2087#pullrequestreview-979979388 after PR #2087 was merged.  I also requested a few small changes but which did not happen.

This PR adds the missing translation strings, removes word Magento, changes the label format with capitalize letters as it is throughout the Backend. It also adds a more explicit comment for novices. 

As an observation, the value for never run (0 0 30 2 0) is for February 30, a day that does not exist in calendar. I mentioned this in case anyone wonders why if it was not better to leave the input box empty. if it is empty the default value will be taken.

This is how the Backend implementation looks like if this PR is merged:

advanced_cache

### DIFF
--- a/app/code/core/Mage/Core/etc/system.xml
+++ b/app/code/core/Mage/Core/etc/system.xml
@@ -61,20 +61,20 @@
                         </use_form_key>
                     </fields>
                 </csrf>
-                <cache>
-                    <label>Advanced Magento cache settings</label>
-                    <sort_order>10000</sort_order>
+                <cache translate="label" module="core">
+                    <label>Advanced Cache Settings</label>
+                    <sort_order>1000</sort_order>
                     <show_in_default>1</show_in_default>
                     <show_in_website>0</show_in_website>
                     <show_in_store>0</show_in_store>
                     <fields>
-                        <flush_cron_expr>
-                            <label>Cron expression for regular cache flushing</label>
+                        <flush_cron_expr translate="label comment">
+                            <label>Cron Expression for Flushing Cache</label>
                             <sort_order>1</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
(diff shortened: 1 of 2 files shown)


ADDISON74:
In PR https://github.com/OpenMage/magento-lts/pull/2087 there is an initial value set that appears in the input text box "30 2 * * *". If I delete this value and I don't set one the box remains empty after saving the configuration.

**Question: What happens in this case with an empty value?**
Does the cron run with the initial value or not? If it doesn't run then we no longer need to specify the value "Never Run" (30 3 * * 6). We can only mention to leave the box empty as not to run the cron.

I guess @luigifab asked the same question without getting an answer.

ADDISON74:
I will do tests to analyze in more detail what happens when the input box is empty. Most likely the default value will be used, but I will change it to see how the cron table updates. If the results are as I anticipate the message in the comment I will have to modify it.

ADDISON74:
The default value used in OpenMage is "30 2 * * *" inherited from Magento. This value is displayed in the box but if the configuration is not saved it will not be recorded as a row in the core_config_data table.

Once it's saved in the table the default value will no longer be used, but the established one. If the box is left empty the cron will not run. If the value in the table is deleted the default value will be used again.

I made the necessary changes in the comment. I used the expression "no longer run" because until then the cron was running. The "never run" variant is not suitable because this cron can run in a certain situation.

fballiano (approved):

justinbeaty:
I would maybe suggest something like:

```
The default value is "30 2 * * *". Enter an empty value to disable automatic cache flushing.
```

But I think this line has seen many suggested revisions, so it doesn't matter too much in the end.

ADDISON74:
It should be fine now.

luigifab (approved):

elidrissidev (approved):

sreichel:
Does translations still work with locale packs?

(IMHO only output should be changed)

elidrissidev:
@sreichel The translations modified in this PR were added by #2087 which didn't include translations.

PR #2312, opened 2022-07-12
Fix #815 - add getAttributeRawText function

### LINKED ISSUE #815: Idea: Add helper for multiselect product attribute label
See here

https://stackoverflow.com/a/30519730/1173670

justinbeaty:
### Description (*)

Trying to clear out an old issue, this PR adds a new function to `app/code/core/Mage/Catalog/Model/Resource/Abstract.php` called ~~`getAttributeRawLabel`~~ `getAttributeRawText` with the same signature as `getAttributeRawValue`.

Issue #815 listed a stackoverflow answer here: https://stackoverflow.com/a/30519730/1173670 that suggested to put this function somewhere in a helper, but I think the abstract resource model will work better, especially as PR #2264 is developed.

1. Fixes OpenMage/magento-lts#815

You can try something like this. It works on both products and categories.

```php
$prod = Mage::getModel('catalog/product')->load(1);
print_r($prod->getResource()->getAttributeRawValue($prod->getId(), 'some_select_attr', 1));
print_r($prod->getResource()->getAttributeRawText($prod->getId(), 'some_select_attr', 1));

$cat = Mage::getModel('catalog/category')->load(1);
print_r($cat->getResource()->getAttributeRawValue($cat->getId(), 'some_select_attr', 1));
print_r($cat->getResource()->getAttributeRawText($cat->getId(), 'some_select_attr', 1));
```

### Questions or comments

I think this may be useful, but I wonder how do people know about these new methods we add to OpenMage?

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Abstract.php
@@ -700,6 +700,45 @@ abstract class Mage_Catalog_Model_Resource_Abstract extends Mage_Eav_Model_Entit
     }
 
     /**
+     * Retrieve attribute's raw value from DB using its source model if available.
+     *
+     * @param int $entityId
+     * @param int|string|array $attribute atrribute's ids or codes
+     * @param int|Mage_Core_Model_Store $store
+     * @return bool|string|array
+     */
+    public function getAttributeRawText($entityId, $attribute, $store)
+    {
+        if (!$entityId || empty($attribute)) {
+            return false;
+        }
+
+        if ($store instanceof Mage_Core_Model_Store) {
+            $store = $store->getId();
+        }
+
+        $store = (int)$store;
+        $attribute = is_array($attribute) ? $attribute : [$attribute];
+        $value = $this->getAttributeRawValue($entityId, $attribute, $store);
+
+        if (!$value) {
+            return false;
+        }
+
+        // Ensure we have an associative array of attribute => values
+        $values = is_array($value) ? $value : array_combine($attribute, [$value]);
+
+        foreach ($values as $_attribute => &$_value) {
+            $_attribute = (clone $this->getAttribute($_attribute))->setStoreId($store);
+            if ($_attribute->getSourceModel() || $_attribute->getFrontendInput() === 'select' || $_attribute->getFrontendInput() === 'multiselect') {
+                $_value = $_attribute->getSource()->getOptionText($_value);
+            }
+        }
+
+        return count($values) === 1 ? reset($values) : $values;
+    }
+
+    /**
      * Reset firstly loaded attributes
      *
      * @inheritDoc


elidrissidev:
I wonder what's the difference between this method and calling `$product->getAttributeText('color');`, besides the store param?

justinbeaty:
@elidrissidev I guess that you can call `getAttributeRawValue` or `getAttributeRawLabel` without loading the entire product

justinbeaty:
@kiatng i won’t be at my desk until tomorrow morning (PST) but it looks like it might just return the admin label. If so then I’ll either see if there’s an easy fix or close the PR (and suggest closing rhe original issue).

justinbeaty:
Fixed a few other issues when passing $store as Mage_Core_Model_Store or $attribute as an array. 

Everything should work now, I've done several tests like this:

```php
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), 171, $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), [171], $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), [171,169], $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), 'color', $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), ['color'], $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), ['color','size'], $store));
print_r($prod->getResource()->getAttributeRawLabel($prod->getId(), ['size'], $store));
```

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -699,6 +699,43 @@ public function getAttributeRawValue($entityId, $attribute, $store)
...
 
+    /**
+     * Retrieve attribute's label from DB.
+     *
+     * @param int $entityId
+     * @param int|string|array $attribute atrribute's ids or codes
+     * @param int|Mage_Core_Model_Store $store
+     * @return bool|string|array
+     */
+    public function getAttributeRawLabel($entityId, $attribute, $store)
```
kiatng, line 710: The name `getAttributeRawLabel()` doesn't seem correct to me, how about

```suggestion
    public function getAttributeText($entityId, $attribute, $store)
```

justinbeaty: I had a similar thought. How about `getAttributeRawText` to differentiate between `$product->getAttributeText('color');`. 

The "Raw" portion makes sense to me to match `getAttributeRawValue` since they both get data straight from the DB instead of going through the product model.

justinbeaty:
Pushed one more fix & renamed the function. I'm now realizing that original StackOverflow answer was very much not complete. I think this function is a good addition, I'd definitely use it in a few places where I've had to get attribute options manually.

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -699,6 +699,45 @@ public function getAttributeRawValue($entityId, $attribute, $store)
...
+        if (!$value) {
+            return false;
+        }
+
+        // Ensure we have an associative array of attribute => values
+        $values = is_array($value) ? $value : array_combine($attribute, [$value]);
+
+        foreach ($values as $_attribute => &$_value) {
+            $_attribute = (clone $this->getAttribute($_attribute))->setStoreId($store);
+            if ($_attribute->getFrontendInput() === 'select' || $_attribute->getFrontendInput() === 'multiselect') {
```
kiatng, line 733: To take care of `boolean` and custom frontend input, e.g., I have `select_country`:
```suggestion
            if ($_attribute->getSourceModel() || $_attribute->getFrontendInput() === 'select' || $_attribute->getFrontendInput() === 'multiselect') {
```

justinbeaty:
@kiatng Pushed your change, thanks for the suggestion.

I wonder if Mage_Catalog_Model_Product should have a similar check here:

https://github.com/OpenMage/magento-lts/blob/33dfa269fa60d56f4d7af19a393179e3ae927fab/app/code/core/Mage/Catalog/Model/Product.php#L1622-L1628

Because if I try and do `$product->getAttributeText('description')` for example, I get back `bool(false)`. Or if I do `$product->getAttributeText('some_nonexistent_attr_code')` I get an exception.

app/code/core/Mage/Catalog/Model/Resource/Abstract.php
```diff
@@ -699,6 +699,45 @@ public function getAttributeRawValue($entityId, $attribute, $store)
         return $attributesData ? $attributesData : false;
     }
 
+    /**
+     * Retrieve attribute's raw text value from DB.
```
fballiano, line 703: Hi @justinbeaty, I wanted to test this but actually I couldn't understand (from a first read) what text value the function is returning, and maybe the docblock could be a little bit more verbose?

Cause at the end of the day "Retrieve attribute's raw text value from DB" it's the almost the same of the function name so it doesn't help me much understanding what the function is doing

:-)

justinbeaty: I updated the docblock to "Retrieve attribute's raw value from DB using its source model if available." Not sure if that's still good enough, so if there's any suggestions I'm all ears.

elidrissidev (approved):
Tested with multiple attribute types and all looks good to me.

fballiano (approved):

PR #2319, opened 2022-07-15
Mage Factory Methods: Direct "_registry" Access Without Method Call

Sdfendor:
### Description (*)

Four factory methods in the main Mage class use the `registry()` method to retrieve object instances they created prior. To use this method in those cases is unnecessary and counterproductive. This PR adjusts those methods to use `self::$_registry` directly. 
In addition small formatting fixes are applied and one null coalescing replaces a "if/isset" construct.  Due to a changed return type of `getSingleton()` a few type hints (with explanation) are added as well to satisfy PHPStan static analysis.  

### Questions or comments

1. Why can the `registry()` calls be replaced?
    * The method accesses `self::$_registry` while checking if given `key` exists and is not null. In each adjusted case however it is guaranteed by the code executed prior that the used `$registryKey` is not null. Take `helper()` for example:
    
       image
       The value of `self::$_registry['registryKey']` is either an object instance (the helper) or a fatal error is emitted because class name in `$helperClass` is not found. Null isn't possible and the registry value is written. 
2. Why is replacement useful?
    * A call to another method is saved which only additional use is another (in this case unnecessary) isset check. More importantly though there are inspections (namely in the popular plugin Php Inspections ​(EA Extended))​ that look at return values of methods and check if null is possible. In those cases (and if the returned value is an object and its methods are called) the inspection warns and suggests usage of optional chaining (a feature of PHP 8.1). Those false positives can be avoided by this change.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -524,7 +524,7 @@ final class Mage
      *
      * @param   string $modelClass
      * @param   array $arguments
-     * @return  Mage_Core_Model_Abstract|false
+     * @return  Mage_Core_Model_Abstract
      */
     public static function getSingleton($modelClass = '', array $arguments=array())
     {
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -527,15 +527,15 @@ final class Mage
      *
      * @param   string $modelClass
      * @param   array $arguments
-     * @return  Mage_Core_Model_Abstract
+     * @return  Mage_Core_Model_Abstract|false
      */
-    public static function getSingleton($modelClass='', array $arguments=array())
+    public static function getSingleton($modelClass = '', array $arguments=array())
     {
         $registryKey = '_singleton/'.$modelClass;
         if (!isset(self::$_registry[$registryKey])) {
             self::register($registryKey, self::getModel($modelClass, $arguments));
         }
-        return self::registry($registryKey);
+        return self::$_registry[$registryKey];
     }
 
     /**
@@ -577,7 +577,7 @@ final class Mage
         if (!isset(self::$_registry[$registryKey])) {
             self::register($registryKey, self::getResourceModel($modelClass, $arguments));
         }
-        return self::registry($registryKey);
+        return self::$_registry[$registryKey];
     }
 
     /**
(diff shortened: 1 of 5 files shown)


app/Mage.php
```diff
@@ -527,15 +527,15 @@ public static function getModel($modelClass = '', $arguments = array())
      *
      * @param   string $modelClass
      * @param   array $arguments
-     * @return  Mage_Core_Model_Abstract
+     * @return  Mage_Core_Model_Abstract|false
```
Sdfendor, line 527: PHPStan complains about this change but I did it because the method is using `getModel` and it is possible that this method returns false (in case the given `$modelClass` is unknown). In turn it is possible that "false" is registered and `getSingleton` returns false as well. What can be done about this?

elidrissidev: There's some "magic" taking place because `Mage::getModel()` also returns `Mage_Core_Model_Abstract|false` yet phpstan doesn't complain about its usage. I don't see it in the baseline file either.

elidrissidev: I know what's happening, the error is only triggered for `getSingleton` calls with a variable as `$modelClass`.

Sdfendor: I'm not sure how to proceed, I don't like the possibility of `getModel()` and in turn `getSingleton()`  to return false to begin with (in contrast to something like `helper()`) but changing that behavior shouldn't be in the scope of this PR and would constitute a change for 20. I'm fully aware that returning false is somewhat of an edge case. 
I could revert the added `|false`. Otherwise the objected lines by PHPStan would need an added type check I guess.

elidrissidev: [macopedia/phpstan-magento1](https://github.com/macopedia/phpstan-magento1) has some extension code for PHPStan that tells it whether `getModel` or `getSingleton` will return `false` depending if the model actually exists. These errors are only from cases where a dynamic model class is given (PHPStan can't resolve that so it assumes false is a possible value), I'm not sure if adding a `@var` doc line above those calls would solve this..

Sdfendor: I tried to fix the errors with annotations but it solved only half the errors. I therefore have decided to revert my change to the return annotation of `getSingleton()`. While I still think that the return type must be different, this and subsequent changes (possibly to the PHPStan plugin as well) is not part of scope of this PR.

fballiano (approved):
tested briefly and LGTM

elidrissidev (approved):

PR #2320, opened 2022-07-15
Fix phpdoc for saveAttribute()

Caprico85:
### Description (*)
`saveAttribute()` in Mage_Sales_Model_Resource_Order_Abstract only specifies `string` as type for `$attribute`. As you can see a few lines later, it also takes string arrays.

```php
public function saveAttribute(Mage_Core_Model_Abstract $object, $attribute)
{
    if ($attribute instanceof Mage_Eav_Model_Entity_Attribute_Abstract) {
        $attribute = $attribute->getAttributeCode();
    }

    if (is_string($attribute)) {
        $attribute = array($attribute);
    }

    if (is_array($attribute) && !empty($attribute)) {                    // <-- see the is_array() here
        $this->beginTransaction();
```

### DIFF
--- a/app/code/core/Mage/Sales/Model/Resource/Order/Abstract.php
+++ b/app/code/core/Mage/Sales/Model/Resource/Order/Abstract.php
@@ -323,7 +323,7 @@ abstract class Mage_Sales_Model_Resource_Order_Abstract extends Mage_Sales_Model
      * Perform actions after object save
      *
      * @param Mage_Core_Model_Abstract $object
-     * @param string $attribute
+     * @param string|string[]|Mage_Eav_Model_Entity_Attribute_Abstract $attribute
      * @return $this
      */
     public function saveAttribute(Mage_Core_Model_Abstract $object, $attribute)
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Sales/Model/Resource/Order/Abstract.php
```diff
@@ -323,7 +323,7 @@ protected function _afterSaveAttribute(Mage_Core_Model_Abstract $object, $attrib
      * Perform actions after object save
      *
      * @param Mage_Core_Model_Abstract $object
-     * @param string $attribute
+     * @param string|string[] $attribute
```
elidrissidev, line 326: We should also add `Mage_Eav_Model_Entity_Attribute_Abstract` since it seems it's an accepted type.

elidrissidev:
This change also happens to solve a phpstan error that was previously ignored so we need to remove it from the baseline file below:

https://github.com/OpenMage/magento-lts/blob/0d997174c676cdb154162bd945f75bb37bb0b2b2/.github/phpstan-baseline.neon#L7013-L7017

Caprico85:
You're right. I updated my PR to include `Mage_Eav_Model_Entity_Attribute_Abstract` as additional type and remove the phpstan ignore.

elidrissidev (approved):

ADDISON74 (approved):

PR #2317, opened 2022-07-13
GUI to Edit EAV Attributes & Sets - Base Code + Category

justinbeaty:
### Description (*)

This PR gives backend users the possibility to create/edit/delete any EAV attribute. In other words, all of these:

```sql
mysql> SELECT entity_type_id,entity_type_code FROM eav_entity_type ORDER BY entity_type_id;
+----------------+------------------+
| entity_type_id | entity_type_code |
+----------------+------------------+
|              1 | customer         |
|              2 | customer_address |
|              3 | catalog_category |
|              4 | catalog_product  |
|              5 | order            |
|              6 | invoice          |
|              7 | creditmemo       |
|              8 | shipment         |
+----------------+------------------+
```

I have only done light testing on this, we probably need more work before this can be merged as a useful feature.

### Related Pull Requests

This PR builds on top of the work done by @fballiano. It is meant to supersede these PRs if the community accepts.
1. #2260
2. #2264

### Manual testing scenarios (*)

1. Open the backend
2. See new options in Sales > Attributes, Catalog > Attributes, Customer > Attributes
3. Try to create new attributes or attribute sets

### Questions or comments

Todo:
- [x] Check translation strings
- **For another PR**: Modify customer attributes to add `use_in_forms` option?
- **For another PR**: Modify Customers > Manage Groups to allow us to select a attribute set for each customer group?
- **For another PR**: Block the user from editing/deleting "system" attributes. Use `adminhtml_[entityCode]_attribute_scopes` event to disallow changing scope perhaps?
- [x] Check ACL permissions, and make sure they didn't break BC with new Catalog > Attributes > Products > Attributes path.
- [x] Fit icon-head for `admin/XXX_attribute/new/` and `admin/XXX_set/add/`
- [ ] Figure out Copyright headers -- most of the code is copy/pasted from other core code, so I think we still need Magento, LLC copyright notice (in addition to ours in another PR).
- [x] Fix array short syntax
- [ ] Remove all prototypejs code
- [x] merge at least customer or category management PR into this, otherwise this is untestable
- **For another PR**: Review the SQL commands at the end of [this comment](https://github.com/OpenMage/magento-lts/pull/2264#issuecomment-1181890693). Should we add an upgrade script?

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2321, opened 2022-07-15
Allow custom acl path in adminhtml.xml by using <acl> tag

justinbeaty:
### Description (*)

I have not fully tested this change yet, but wanted to seek feedback on if this is a good idea or not.

This PR will let you set a custom ACL `resource_id` in `adminhtml.xml` that is different than the XML structure. For example:

```xml
<acl>
    <resources>
        <admin>
            <children>
                <catalog translate="title" module="catalog">
                    <title>Catalog</title>
                    <sort_order>30</sort_order>
                    <children>
                        <attributes translate="title">
                            <title>Attributes</title>
                            <children>
                                <product translate="title">
                                    <title>Products</title>
                                    <children>
                                        <attributes translate="title">
                                            <title>Attributes</title>
                                            <acl>admin/catalog/attributes/attributes</acl>
                                        </attributes>
                                        <sets translate="title">
                                            <title>Attribute Sets</title>
                                            <acl>admin/catalog/attributes/sets</acl>
                                        </sets>
                                    </children>
                                </product>
                                <category translate="title">
                                    <title>Categories</title>
                                    <children>
                                        <attributes translate="title">
                                            <title>Attributes</title>
                                            <acl>admin/eav/catalog_category/attributes</acl>
                                        </attributes>
                                        <sets translate="title">
                                            <title>Attribute Sets</title>
                                            <acl>admin/eav/catalog_category/sets</acl>
                                        </sets>
                                    </children>
                                </category>
                            </children>
                        </attributes>
                    </children>
                </catalog>
            </children>
        </admin>
    </resources>
</acl>
```

Note the lines:
```xml
<acl>admin/catalog/attributes/attributes</acl>
<acl>admin/catalog/attributes/sets</acl>
<acl>admin/eav/catalog_category/attributes</acl>
<acl>admin/eav/catalog_category/sets</acl>
```

~~I think this could be a useful addition because in PR #2317 we may want to change the menu structure, but do not want to break current ACL permissions. Also, that PR expects to find generic EAV permissions at `admin/eav/$ENTITY/attributes` and `admin/eav/$ENTITY/sets` but we don't want to force the user to conform to our menu or ACL structure.~~

Edit: this PR is actually not necessary for #2317 but I will leave this PR open in case there might be other use-cases for changing the ACL path.

### Related Pull Requests
#2317 

1. Create a role with `Catalog > Attributes > Attributes` & `Catalog > Attributes > Sets` allowed
2. Update the `Mage/Catalog/etc/adminhtml.xml` to reflect a different menu and ACL structure
3. Apply this patch and make sure the permissions are still allowed

I also should test to make sure nothing else breaks with the API/API2 module. I honestly don't use either of those API modules nor do I use ACL often, so if anyone can see where this might break something, please let me know. I do not expect this PR to be merged quickly.

### DIFF
--- a/app/code/core/Mage/Admin/Model/Roles.php
+++ b/app/code/core/Mage/Admin/Model/Roles.php
@@ -152,7 +152,7 @@ class Mage_Admin_Model_Roles extends Mage_Core_Model_Abstract
             $level = -1;
         } else {
             $resourceName = $parentName;
-            if (!in_array($resource->getName(), array('title', 'sort_order', 'children', 'disabled'))) {
+            if (!in_array($resource->getName(), array('title', 'sort_order', 'children', 'disabled', 'acl'))) {
                 $resourceName = (is_null($parentName) ? '' : $parentName . '/') . $resource->getName();
 
                 //assigning module for its' children nodes
@@ -160,16 +160,22 @@ class Mage_Admin_Model_Roles extends Mage_Core_Model_Abstract
                     $module = (string)$resource->getAttribute('module');
                 }
 
+                if ($resource->acl) {
+                    $aclpath = (string)$resource->acl;
+                } else {
+                    $aclpath = $resourceName;
+                }
+
                 if ($rawNodes) {
-                    $resource->addAttribute("aclpath", $resourceName);
+                    $resource->addAttribute("aclpath", $aclpath);
                     $resource->addAttribute("module_c", $module);
                 }
 
                 if (is_null($represent2Darray)) {
-                    $result[$resourceName]['name']  = Mage::helper($module)->__((string)$resource->title);
-                    $result[$resourceName]['level'] = $level;
+                    $result[$aclpath]['name']  = Mage::helper($module)->__((string)$resource->title);
+                    $result[$aclpath]['level'] = $level;
                 } else {
-                    $result[] = $resourceName;
+                    $result[] = $aclpath;
                 }
             }
         }


app/code/core/Mage/Admin/Model/Roles.php
```diff
@@ -152,24 +152,30 @@ protected function _buildResourcesArray(
...
-                    $resource->addAttribute("aclpath", $resourceName);
+                    $resource->addAttribute("aclpath", $aclpath);
                     $resource->addAttribute("module_c", $module);
                 }
 
                 if (is_null($represent2Darray)) {
-                    $result[$resourceName]['name']  = Mage::helper($module)->__((string)$resource->title);
-                    $result[$resourceName]['level'] = $level;
+                    $result[$aclpath]['name']  = Mage::helper($module)->__((string)$resource->title);
+                    $result[$aclpath]['level'] = $level;
```
justinbeaty, line 176: I am not positive it's okay to change the array key here. Before I had something like:
```php
$result[$resourceName]['name']  = Mage::helper($module)->__((string)$resource->title);
$result[$resourceName]['level'] = $level;
$result[$resourceName]['aclpath'] = $aclpath; // New array value
```

But that will require more code changes in core. For example, in the admin role edit screen: https://github.com/OpenMage/magento-lts/blob/0d997174c676cdb154162bd945f75bb37bb0b2b2/app/code/core/Mage/Adminhtml/Block/Permissions/Tab/Rolesedit.php#L103-L111

We only check `array_key_exists`, but would need to also check for `aclpath` sub-array key.

PR #2074, opened 2022-05-10
Add a message calling for an end to the Russian invasion of Ukraine

luigifab:
### Description

This is a basic text for README and backend.

def

om

image

### Related Pull Requests

🇺🇦 #2038

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2325, opened 2022-07-17
Revert "Add basic text for Ukraine (#2074)"

fballiano:
Our customers' backends don't need to have an undismissable message like the one that was introduces with #2074.
That same message could have been delivered in the notification feed, instead of hardcoding it and adding an image file to our repository (while at the same time we create BC PRs to remove a few characters from file names).

#2074 will not help that cause in any way and it's only creating a disservice to the users of OM's backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,6 @@
 <img src="https://github.com/openmage/magento-lts/actions/workflows/unit-tests.yml/badge.svg" alt="Unit Tests workflow badge" />
 </p>
 
-<img src="skin/adminhtml/default/default/images/flag-ua.svg?raw=true" width="20" align="top" alt=""> Stop russian war. **Free Ukraine!** ([Read more](https://en.wikipedia.org/wiki/2022_Russian_invasion_of_Ukraine))
-
 # Magento - Long Term Support
 
 This repository is the home of an **unofficial** community-driven project. It's goal is to be a dependable alternative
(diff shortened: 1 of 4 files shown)


Flyingmana:
I see it differently and think its at maximum a minor disturbance and acceptable to keep it this way till the territory of the Ukraine got restored.

ADDISON74:
I approve this PR not because I am not in solidarity with the Ukrainians. The PR #2074 doesn't provide the option to dismissed the message. It would stay there until the user changes the source code or we revert it to the end of the war.

The variant that used the OpenMage feed would have been best in this case, as there would have been a message in the Notifications section of the Backend which could then be marked as read or deleted.

ADDISON74 (approved):

fballiano:
It shouldn't even be necessary to say but obviously this PR has nothing to do with being against this or any other war.
I just think that a message in a software does nothing for the cause, there are many ways of actually doing something and I think this is not one of them.

midlan:
100% agree

The right solution has been mentioned:
> That same message could have been delivered in the notification feed

midlan (approved):

sreichel:
@Flyingmana what's required to move this to feed? Just update OM/Web_Notification?

Flyingmana:
> @Flyingmana what's required to move this to feed? Just update OM/Web_Notification?

yes, adding a post analog to https://github.com/OpenMage/Web_Notifications/blob/master/_posts/2019-06-24-OpenMage.md

sreichel:
Anyone? I agree that this harcdoded message was wrong, but in don't wont to merge as long the feed isnt updated :)

fballiano:
Although I don't agree with these messages in software I created the web notification post with https://github.com/OpenMage/Web_Notifications/pull/4

This way we don't even have to release (and people don't even need to upgrade) in order to receive the message.

sreichel:
@Flyingmana can you please update OM/Web_Notification?

PR #2328, opened 2022-07-19
Fix for modphp8

fballiano:
Without this with modphp won't read the php8 instruction in js/.htaccess


elidrissidev (approved):

Flyingmana:
updated variant of https://github.com/OpenMage/magento-lts/pull/2329

Flyingmana (approved):

PR #2326, opened 2022-07-18
Fix mod_php8 wrong notation in .htaccess files

fballiano:
Lately "mod_php8" instructions were added to some htaccess but they don't work, the right notation is mod_php:
https://github.com/nextcloud/server/issues/26569

I also did a bit of reformatting, hope you like it.

Note: the diff looks complicated (dunno why) but if you check the 2 files:
- https://github.com/fballiano/openmage/blob/fix/skin/.htaccess
- https://github.com/fballiano/openmage/blob/fix/media/.htaccess
you'll see they're pretty straightforward

### Manual testing scenarios (*)

Enable apache's mod_php for php8 and test some php_value directives present in htaccess files, they won't work now, they work after this patch.

PR #2329, opened 2022-07-19
Fix for modphp8 

fballiano:
Without this with modphp won't read the php8 instruction in media/.htaccess and skin/.htaccess


Flyingmana (approved):

Flyingmana:
reasoning and more details are mentioned in https://github.com/OpenMage/magento-lts/pull/2326

Issue #2239, opened 2022-06-20
Inconsistent column width in Backend grids.


ADDISON74:
In the Grid.php files that add columns to the Backend grids I found a mismatch between the width values ​​used for some types of columns that should be fixed. For example, columns in which filter by date should have somewhere at 140px so that the longest date to fit without going to the second row. But there are grids where either the width value is not specified or it is larger than necessary. I personally think that a revision of these grid is necessary.

Here is an example below:

![data_range](https://user-images.githubusercontent.com/8360474/174557132-89ecf5a2-3332-447d-8d8b-d1f63b6beea7.jpg)

As you can see the filter by date column is ridiculously wide. I checked and it should have a minimum width value of 140px. Once this value is set up in boxes.css it is not necessary to specify it in the Grid.php files, except where it is strongly desired not to take up too much space and making the layout look ugly.

After checking all grids I came to the conclusion there is not logic in determining the right widths. There is an inconsistency in setting these values and each programmer chose what he thought would suit, but could choose much better.

I will deal with this issue because it is a visual one and it bothers and becomes tiring when there is no harmony in the presentation of the information in the grids.

fballiano:
closing since the linked PR is merged

Issue #2198, opened 2022-06-06
Dates columns in Backend grids need visual improvement


ADDISON74:
In Backend grids have columns where you can filter the data by date.

This is how the columns in the "Catalog Price Rules" section look like now (OpenMage 20.14):

![before](https://user-images.githubusercontent.com/8360474/172199752-9a0ee911-e398-46c1-b51e-00685ea5d5ce.png)

This is how the columns would look like based on my future change:

![after](https://user-images.githubusercontent.com/8360474/172199723-fab41c0f-8361-4b0d-ba2b-aeb2788aba10.png)

**What are the benefits?**

- Those colons used after words To, From and In don't look good at all.

- The icon for calendar popup is placed immediately after the text when in fact it should be placed after the input box.

- The input boxes are a little bit increased. As you can see the date you fill in does not appear completely andit is truncated due to the width limitation set to 50px. In my proposal I changed it to 60px to cover all situations.

This is a visual enhancement. Any opinions are welcome.

elidrissidev:
I agree with this change, it's more natural to have the date selection button to the right of the input.

ADDISON74:
I will take care of this issue. Two PR's will be ready to fix visual issues in filter columns separately.

PR #2330, opened 2022-07-20
Backend grid heads - Align the elements for range and date types

ADDISON74:
This PR is the last part of the implementation of the solution that solves the issue reported here #2198. The first part has already been merged into 1.9.4.x here #2267.

What changes does it bring?

- the text labels (From, To, In) are aligned to the right to avoid the annoying space up to the input box.
- the inputs have been increased in width in order to make the selected date visible in the box.
- the new widths are optimal so that no grid in the Backend is affected.
- 3px padding around the input is identical to that set by the Magento team (see Reports> Orders section for calendar image). a 4px padding will break the arrangement and affect all grids.

I visited one by one all the pages in Backend where there are grids twice and I didn't identify any issue. Here you may see the visual results:

date_range

products

Unfortunately the issue reported here #2239 will have to be solved, that inconsistency of the column width, which had to be solved in CSS first and only if it was necessary to add values in the Grid.php files. I reopened that post because it was not resolved by PR #2267.

Please do not approve this PR without testing it. If necessary changes can be made by anyone because it allows editing by those who have this right. Also, if a better implementation is requested the PR can be closed as well. I will not participate in discussions and I will not make changes if requested. Thank you for understanding.

### DIFF
--- a/skin/adminhtml/default/default/boxes.css
+++ b/skin/adminhtml/default/default/boxes.css
@@ -206,10 +206,10 @@ table.actions td                { vertical-align:top; }
 .grid tr.filter select           { width:100%; }
 .grid tr.filter .range .range-line { margin-bottom:3px; width:100px; }
 .grid tr.filter .range div.date  { min-width:121px; }
-.grid tr.filter .range input     { float:right; width:50px !important; margin-top:0; }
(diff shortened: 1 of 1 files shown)


luigifab (approved):
At first view seems good.

elidrissidev (approved):
Checked most grids, they all look good.

PR #2295, opened 2022-07-08
Add OpenMage Contributors Copyright

justinbeaty:
### Description (*)

From discussion #2276

Added `@copyright` header to 4,687 files that have been modified by the OpenMage project since 2015. The OpenMage project has done great work and deserves its copyright notice.

Instead of a generic 2015-2022 year range, each file has its own specific year, or year range, based on the git history of that file. I've included the script that I used to generate this commit.

The script has comments explaining what it does, but here's an overview:

1. Read all the entire commit history of the repo using `git log --pretty='%H %cs %s'`
2. Loop through each commit and try and determine if this was upstream changes from Magento, LLC or if it was a commit from OpenMage contributors. To do this, the script looks for commit messages such as:
	* "Import Magento"
	* "Merge ... magento-1.*"
	* "SUPEE"
	* Also, some hardcoded commit hashes that don't fit a pattern
3. Loop through each file in the entire repo that contains `@copyright` and find the first and last commit hash from our results in step 2
4. Add or update the `@copyright The OpenMage Contributors` line in each file

To see a list of commits attributed to either Magento or OpenMage, run `php shell/update-copyright.php dump`. I will post the results of that in the next comment so you can see the output here, too.

The rationale behind not using a generic date range is for ease of `git diff` between versions as suggested by @fballiano. Instead, if the script is run before a version is tagged, then it will only be modifying files that were already changed by a commit in that version.

Here is the output from when I ran the script:

```console
$ time php shell/update-copyright.php 
Updated 4687 file(s)

real    3m13.286s
user    2m31.269s
sys     0m48.590s
```

It did take a few minutes, maybe the reading of `git log` on each of the 4,000+ files can be optimized.

Also, this PR is made to 1.9.4.x. I can also make a smaller PR to 20.0 since there will be discrepencies. This could create some merge conflicts in the future, but I don't think it will be too bad. For example assume some file has '2015-2019' in 1.9.4.x and '2015-2020' in 20.0 (because there was a 20.0 branch only commit in 2020 to that file). If the file was then changed in 1.9.4.x in 2022, it would conflict I think when v19 is merged into v20. I ask the maintainers to weigh in on that.

### Manual testing scenarios (*)

Check out this branch and run:

```console
git diff -U0 HEAD~1 | grep '^[+-]' | grep -Ev '^(--- a/|\+\+\+ b/)' | sort -u
```

That will print all unique changes made in the second commit of this PR. You should only see copyright changes. I will post the output in a follow up comment.

### Questions or comments

[...]

PR #2332, opened 2022-07-20
Add translation helper shell script

justinbeaty:
### Description (*)

Because in #2317 I lost track of which translation strings I introduced, I wrote a quick helper script for me to find strings used in the `__` function or in XML files but not defined in a CSV file. I thought it would be useful so I fixed it up a bit and added it as `shell/translations.php`.

It has two modes, `missing` which will tell you which strings you must add to a CSV file, and `unused` which will print strings defined in CSV files but not apparently used anywhere in the source. Additionally, it lets you pipe in a list of files to only check, which is useful to test PRs.

```
Usage:  php -f translations.php -- [options]
        php -f translations.php -- missing --verbose

  missing           Display used translations strings that are missing from csv files
  unused            Display defined translations strings that are not used in templates
  --verbose         Include filename with output
  --lang <lang>     Specify which language pack to check in app/locale, default is en_US
  deprecated        Find deprecated usage of the global __() function
  --fix             Overwrite files to fix deprecated usage DO NOT RUN IN PRODUCTION!
  help              This help

Note: By default, this script will check all files in this repository. However,
      you can pipe a list of files to check for missing translations strings.
      This is useful for checking a specific commit. For example:

      # Check if last two commits may have introduced missing translations
      git diff --name-only HEAD~2 | php -f translations.php -- missing
```

1. Run `php shell/translations.php missing --verbose` to see php, phtml, or xml files that used non-defined translations
1. Run `php shell/translations.php unused --verbose` to see CSV files that have unused translations

### Questions or comments
```
# Returns 422 unused strings
php shell/translations.php unused | wc -l

# Returns 85 missing strings
php shell/translations.php missing | wc -l
```

It could be possible that the code to find usage of the `__` function is not perfect, such as `__('Total: ' . $total)` would only pick up `'Total: '`, but I don't think that's used often, or shouldn't be, as the `%s` usage is preferred.


elidrissidev:
Wow nice work! Will review when I get the time.

justinbeaty:
I don't know much about github actions, but maybe this could also be used for one to check that translation strings have been included if necessary.

shell/translations.php
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * Retrieve Usage Help Message
+     *
+     */
+    public function usageHelp()
+    {
+        return <<<USAGE
+Usage:  php -f translations.php -- [options]
+        php -f translations.php -- missing --verbose
```
fballiano, line 211: would it be possibile to have "--missing" and "--unused" instead of "-- missing"? would you like it that way?

tmotyl:
have you checked if n98-magerun has sth similar already?

justinbeaty:
@tmotyl I did look but I don't see anything like it:
```
  dev:translate:admin                Toggle inline translation tool for admin          
  dev:translate:export               Export inline translations                         
  dev:translate:set                  Adds a translation to core_translate table. Globally for locale
  dev:translate:shop                 Toggle inline translation tool for shop
```

justinbeaty:
As mentioned in #2333, I added a new mode to find deprecated usage of the global `__()` function:

```console
# list deprecated usage only
php shell/translations.php deprecated

# fix deprecated usage, DO NOT RUN IN PRODUCTION
php shell/translations.php deprecated --fix
```

ADDISON74:
This is an interesting PR and thank you for your contribution @justinbeaty. If I look at its description in the Questions or Comments section, I see that there are 422 unused strings and 85 missing strings. That is a problem. I'll take a look to see what it's all about.

For the deprecated option there is also a fix option. Running the command what is done cannot be reversed. We should evaluate if it is possible to make a backup of the files beforehand. If it is not possible, then a warning message must be mentioned to make a backup. We have to check where these changes are and find the parent directory of the files. If there are issues, either put all the files back in place, and if he uses Git he can do a revert.

justinbeaty:
> For the deprecated option there is also a fix option. Running the command what is done cannot be reversed. We should evaluate if it is possible to make a backup of the files beforehand. If it is not possible, then a warning message must be mentioned to make a backup. We have to check where these changes are and find the parent directory of the files. If there are issues, either put all the files back in place, and if he uses Git he can do a revert.

Ideally someone should not just blindly run the `--fix` option but should run `php shell/translations.php deprecated` first to see if there are any occurrences of the deprecated usage. Depending on the 3rd party module authors used, there could be zero occurrences. I think the warning is in the `--help` usage instructions where it says:

```
  --fix             Overwrite files to fix deprecated usage DO NOT RUN IN PRODUCTION!
```

I am not sure if we are responsible for backing up files for someone not using git and deploying directly to production. But if there's a better warning or other way to prevent accidental changes we can add it.

justinbeaty:
As mentioned in #2429 I think a new mode can be added such as:
```
php shell/translations.php upgrade
```

This mode would allow us to change the first column in en_US csv files when we change a translation string. It would be required to add an entry to this shell script so that we can simply find/replace the changes.

In other words, take this example:

"Password must be at least of %d characters."

We can instead fix that in both columns and also put in this shell script something like:

```php
$upgrades = array(
  "Password must be at least of %d characters." => "Password must be at least %d characters.",
);
```

And when the user runs `php shell/translations.php upgrade` it will search all other language packs for the first string to replace with the second.

Thoughts? Feedback from others who use non-english language packs is appreciated. This would be a new step users must take during upgrades.

ADDISON74:
Column 1 is what is used in PHP/PHTML files.
Column 2 is what is used in CSV files and is displayed in Backend/Frontend.

It is obvious that we must pay attention to column 2. And then the modification should occur on column 1 of the CSV, but once we modified it must have a correspondent in the PHP/PHTML file.

PHP/PHTML FILE => COLUMN 1 CSV => COLUMN 2 CSV

To remain Column 2 then the process must be reversed.

justinbeaty:
Let me try to re-explain what I propose.

When we want to change a string, we can do so in the PHTML files and also in **both** columns in the en_US csv files. This keeps everything consistent in our source code.

At the same time, we add an entry to translations.php shell script such that it will find/replace in all translation packs the old string with the new string.

So when a user upgrades to a new OpenMage version, they just have to run the script to fix what we broke in their translation files. The only negative is that they need to be aware of having to run that script.

ADDISON74:
I understand what you mean. I haven't tested the script yet, but I will, including in destructive mode to see what happens and how such a "mistake" can be fixed.

sreichel:
> have you checked if n98-magerun has sth similar already?

I'm pretty sure something like this already exists. If not in n98-core, maybe as addon?

However, I'll test this later :)

fballiano (approved):

sreichel:
```
 ------ ---------------------------------------------------------------------------------- 
  Line   update-copyright.php                                                              
 ------ ---------------------------------------------------------------------------------- 
  118    Strict comparison using === between '' and 'dump' will always evaluate to false.  
 ------ ---------------------------------------------------------------------------------- 

```

sreichel (dismissed):

sreichel:
@fballiano @Flyingmana 

I'd like to have 3 further repos at OpenMage ...

- [ ] `openmage/dev-fix-translations` (for this PR)
- [ ] `openmage/dev-update-copyright` (#2295)
- [ ] `openmage/dev-meta-package`

fballiano:
created the repos and set you as admin

sreichel (approved):
We can approve this first (for better history) and then apply #2853

fballiano (approved):

PR #2340, opened 2022-07-26
Update lib\Pelago\Emogrifier (install w/ composer)

sreichel:
### Description (*)

@fballiano tried to install this with composer, but it didnt work because of php version conflicts or it does not fit to OM code anymore. Updated OM code to work with latest releases of Pelago\Emogrifier.

- v4 => PHP 7.0 support
- v5 => PHP 7.1
- v6 => up to PHP 8.1

### Related Pull Requests

See OpenMage/magento-lts#2138

1. Preview newsletter template from admin section

### Questions or comments

Todo: add to upcoming release builder?


fballiano:
precisely what I did here: https://github.com/fballiano/openmage/blob/mcommerce/app/code/core/Mage/Core/Model/Email/Template/Abstract.php so I like this PR.

I like the removal of the composer.lock, it makes everything difficult :-)

Problem, if we move this library to composer then we need the release builder https://github.com/OpenMage/magento-lts/pull/2165 or we can't do a release.

sreichel:
@fballiano ```disableInlineStyleAttributesParsing()``` or ```disableStyleBlocksParsing()```?

fballiano:
Reading the documentation here: https://github.com/MyIntervals/emogrifier:

> disableStyleBlocksParsing() - By default, CssInliner will grab all <style> blocks in the HTML and will apply the CSS styles as inline "style" attributes to the HTML. The <style> blocks will then be removed from the HTML. If you want to disable this functionality so that CssInliner leaves these <style> blocks in the HTML and does not parse them, you should use this option. If you use this option, the contents of the <style> blocks will not be applied as inline styles and any CSS you want CssInliner to use must be passed in as described in the [Usage section](https://github.com/MyIntervals/emogrifier#usage) above.

I think disableStyleBlocksParsing shouldn't be called in openmage, the style blocks should be parsed.

For disableInlineStyleAttributesParsing (that I had in my code) instead:
> disableInlineStyleAttributesParsing() - By default, CssInliner preserves all of the "style" attributes on tags in the HTML you pass to it. However if you want to discard all existing inline styles in the HTML before the CSS is applied, you should use this option.

I don't really understand the documentation, I think inline styles should be left as they are no?

fballiano:
I think the `disableStyleBlocksParsing` should be removed, `<style>` blocks should be parsed by the library.

About `disableInlineStyleAttributesParsing`, checking the source code used I would leave it as default (not calling the `disableInlineStyleAttributesParsing` method)

sreichel:
See #2411

Issue #1911, opened 2021-12-21
Contact Form - Some bots can use the controller behind the form


ADDISON74:
Although I use the HoneySpam extension and the form has by default a hidden value (hideit) there are days when I receive a burst of 5 messages. In the web server log I found the following records:
```156.146.63.148 - - [21/Dec/2021:09:22:28 +0200] "GET /contacts/ HTTP/1.0" 200 80153 "https://www.mydomain.com/contacts/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36"
156.146.63.148 - - [21/Dec/2021:09:22:29 +0200] "POST /contacts/index/post/ HTTP/1.0" 302 549 "https://www.mydomain.com/contacts/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36"
156.146.63.148 - - [21/Dec/2021:09:22:29 +0200] "GET /contacts/index/ HTTP/1.0" 200 80182 "https://www.mydomain.com/contacts/index/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36"
156.146.63.148 - - [21/Dec/2021:09:22:30 +0200] "POST /contacts/index/post/ HTTP/1.0" 302 549 "https://www.mydomain.com/contacts/index/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36"
156.146.63.148 - - [21/Dec/2021:09:22:31 +0200] "GET /contacts/index/ HTTP/1.0" 200 80182 "https://www.mydomain.com/contacts/index/" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/90.0.4430.85 Safari/537.36"
[...]

PR #1866, opened 2021-10-20
[Theme BC] Add formkey validation to Newsletter subscribe action

elidrissidev:
Validate formkey when subscribing to newsletter to prevent CSRF and spam.

### Description (*)
Magento 1 lacks formkey validation when trying to subscribe to newsletter, this can cause bots or malicious scripts to issue many subscription requests and fill the database with junk.

Detailed discussion is in the first attached issue.

### Fixed Issues (if relevant)
1. Fixes OpenMage/magento-lts#1287
2. Fixes OpenMage/magento-lts#1738

### Manual testing scenarios (*)
1. Assuming you didn't disable `Mage_Newsletter` module or Guest Subscription from Configuration, run:
```sh
curl '<magento_host>/newsletter/subscriber/new' -d 'email=bot@example.org'
```
This will add `bot@example.org` to subscribers before applying the change, but not after.

### Questions or comments
I'm not a a security expert, so If you have anything to add feel free.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

### Todo
- [x] Disable formkey validation by default for BC (?)

Issue #869, opened 2019-12-13
Default setting for validate_formkey_checkout => 1


kkrieger85:
After new installation there's a warning in Backend. 

```
    <div class="notification-global notification-global-warning">
        <strong style="color:red">Important: </strong>
        <span>Formkey validation on checkout disabled. This may expose security risks.
        We strongly recommend to Enable Form Key Validation On Checkout in
        <a href="<?php echo $this->getSecurityAdminUrl(); ?>">Admin / Security Section</a>,
        for protect your own checkout process. </span>
    </div>
```

Is there a reason why the default setting should not be 1 (for yes, validate formkey)?

colinmollenhour:
Probably just so that upgrades for existing users with modified template files don't get broken checkouts.

One way around this could be to add some code to one of the initial `install-1.0.php` files so that new installations get the value overridden and old installations do not.

kkrieger85:
Agree with using an install script. 

In general, I think, it would improve shop security. So we should force user to be more secure.

PR #2347, opened 2022-07-30
[Theme BC] Add form key validation to Contacts form

elidrissidev:
Add form key validation when submitting Contacts form.

### Description (*)

Similar to #1866, this PR adds form key validation to the Contacts form to prevent bots from sending spam requests to your email address. This option is disabled by default as it is a breaking change for custom themes.

With this change, the `hideit` field in contacts form is now obsolete, so it may be removed in a future PR if it's safe to do so.

1. Fixes OpenMage/magento-lts#1911

1. Go to System -> Configuration -> General -> Contacts -> Security and enable form key validation.
2. Try submitting the form by issuing an HTTP request directly, if everything is fine the request should not go through. 
```sh
$ curl '<magento_host>/contacts/index/post' -d 'name=John+Doe&email=bot@example.org&comment=test'
```


app/code/core/Mage/Contacts/etc/system.xml
```diff
@@ -89,6 +89,23 @@
...
                     </fields>
                 </email>
+                <security translate="label">
+                    <label>Security</label>
+                    <sort_order>60</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>1</show_in_website>
+                    <show_in_store>1</show_in_store>
+                    <fields>
+                        <enable_form_key translate="label comment">
```
fballiano, line 99: Since in "core" we have use_form_key:
Schermata 2022-07-30 alle 18 40 52
wouldn't it be better to use the same?

I was also thinking that we had a single flag to enable csfr for all openmage:
Schermata 2022-07-30 alle 18 41 59
shouldn't we use that single one instead of creating a flag for each module?

elidrissidev: Yes it would be better to have one global flag, but a separate one is added for backwards-compatibility because if users have `system/csrf/use_form_key` enabled, their contacts form won't work if they don't apply the change to their custom themes.

fballiano: I thought so but at the same time now they update and they don't have the new security feature enabled so I don't know. Also the idea of having new configuration options for the same thing I don't know :-\

elidrissidev: I'm not a big fan of it either but I can't think of another way to fix this issue. We could remove these separate flags in `20.0`(or superior versions) and keep only the global one. There's also a separate flag for the checkout I think.

elidrissidev: I think a major reason why many features, especially security related, go unnoticed is because users don't have an easy way of knowing about them. It is easy for such changes to be lost in the changelog next to tens of other changes. Using the notifications for this purpose would be good, but I remember seeing opposing opinions about that in some thread.

ADDISON74:
The chunk bellow was introduced by the Magento team in the past in /app/code/core/Mage/Checkout/etc/system.xml

```xml
        <admin>
            <groups>
                <security>
                    <fields>
                        <validate_formkey_checkout translate="label">
                            <label>Enable Form Key Validation On Checkout</label>
                            <sort_order>4</sort_order>
                            <comment><![CDATA[<strong style="color:red">Important!</strong> Enabling this option means
                            that your custom templates used in checkout process contain form_key output.
                            Otherwise checkout may not work.]]></comment>
                        </validate_formkey_checkout>
                    </fields>
                </security>
            </groups>
        </admin>
```

Here is the visual result in Backend

formkey

These are my comments:

1. For this PR but also for the Newsletter (#1866) I would have insert the options in the Admin/Security section and not in Contacts and Newsletter. It is a personal opinion for grouping them in one place and better visibility.

2. I would have used validate_formkey_checkout, validate_formkey_contacts, validate_formkey_newsletter as fields and Enable Form Key Validation On Checkout, On Newsletter, On Contacts as labels. It is a personal opinion to keep the Magento team's initial implementation way.

I haven't tested yet if I can send the form using the curl command, in both situations Enabled/Disabled.

ADDISON74:
In addition to the comments shown previously I agree with the use of **use_form_key** also to keep the initial Magento format.

ADDISON74:
Perhaps we should also inform through a notification that formkeys should be activated for Newsletter and Contact, as happened for Checkout. If all OpenMage stores are not activated, if they do not use captcha, they are exposed to becoming spam deliverers. If they are not enabled in stores that use OpenMage, without a reCaptcha protection, they are exposed to becoming SPAM deliverers. The HoneySpam extension does not prevent abuse of the controller action, this extension only introduces a bait field in the form and if it is filled out it considers the visitor as bot (in the contact form there is already a field named hideit, but absolutely useless in my opinion).

More details here https://github.com/OpenMage/magento-lts/issues/869 and https://github.com/OpenMage/magento-lts/pull/871.

ADDISON74:
@elidrissidev - Thank you for work. PR works without issues and blocks the abusive use of the POST action in the controller. I implemented it in production on a server where 50 messages were received daily through the contact form, all with sexual content.

1. Before approving it I would suggest small format changes to respect the previous implementations of the Magento team. ```validate_formkey``` should replace ```enable_form_key```. The changes are in the following 3 files.

In the file /app/code/core/Mage/Contacts/controller/**IndexController.php**

```php
const XML_CSRF_USE_FLAG_CONFIG_PATH = 'contacts/security/validate_formkey';
```

In the file /app/code/core/Mage/Contacts/etc/**config.xml**

```xml
            <security>
                <validate_formkey>0</validate_formkey>
            </security
```

In the file /app/code/core/Mage/Contacts/etc/**system.xml**

```xml
                <security translate="label">
                    <label>Security</label>
                    <sort_order>60</sort_order>
                    <fields>
                        <validate_formkey translate="label comment">
                            <label>Enable Form Key Validation</label>
                            <sort_order>1</sort_order>
                            <comment><![CDATA[<strong style="color:red">Important!</strong> Enabling this option means that your custom templates used for contact form must contain <code>form_key</code> block output. Otherwise the contact form will not work.]]></comment>
                        </validate_formkey>
                    </fields>
                </security>
```

2. In my previous comment I referred to CSRF for Checkout and I suggested placing the recent Backend options for Newsletter and Contacts in /Admin/System/Security. Analyzing in detail I can say that the Magento team made a mistake by placing the option in this section, it should have been placed in Checkout section. In conclusion, you chose to place the options in the correct sections. Maybe in the future we will move that option to the section it belongs to.

3. Perhaps we should also inform through a notification that formkeys must be activated for Newsletter and Contact, as happened for Checkout in the past. If administrators do not activate these options in OpenMage stores, without a reCaptcha protection, they become SPAM deliverers.  The OM team discussed about this issue and enabled it by default. More details here https://github.com/OpenMage/magento-lts/issues/869 and https://github.com/OpenMage/magento-lts/pull/871.

4. The HoneySpam extension does not prevent abuse of the controller action, this extension only introduces a bait field in the form and if it is filled out it considers the visitor as bot (in the contact form there is already a field named hideit, but absolutely useless in my opinion when the formkey is used). The fishing percentage of HoneySpam is up to 40%, therefore it is useful and it should be installed, but most of the newsletter subscriptions and requests for information through contact form are made by bots that use the controller directly.

ADDISON74:
Maybe the formkey check in the controller should be better placed here:

```php
        if ($this->getRequest()->isPost() && $this->getRequest()->getPost('email')) {
            $session            = Mage::getSingleton('core/session');
            $customerSession    = Mage::getSingleton('customer/session');
            $email              = (string) $this->getRequest()->getPost('email');

            try {
		if (!$this->_validateFormKey()) {
	            Mage::throwException($this->__('Validation has failed.'));
        	    $this->_redirectReferer();
	            return;
        	}

                if (!Zend_Validate::is($email, 'EmailAddress')) {
                    Mage::throwException($this->__('Please enter a valid email address.'));
                }
```

In this way, when the formkey validation is not done, a line is written in the log file. You can find out how many unsuccessful attempts there were and you can move on with the banning of IPs.

elidrissidev:
Thank you for sharing your remarks @ADDISON74, I'll add them in the following days.

ADDISON74:
@elidrissidev - So far so good. What do you think about my last post above? The one in which I suggested that the formkey check should also generate an exception in the log. I think it is a welcome information to periodically check abuses on the contact controller. I also made you a code proposal.

ADDISON74:
@elidrissidev - After you will analyze whether it is beneficial to write the exception in the log file, this PR #1866 must be modified to follow the same format like this one in PHP/XML files.

fballiano:
I don’t know much about fail2ban and this kind of softwares buy is it possible to write a log file that’s already readable by fail2ban for an easy config?

ADDISON74:
Fail2Ban reads log files and based on the filters you set it will create jails for IPs. The two log files in OpenMage cannot be processed as long as there is no information about the IP, but in the case of this PR it is useful to find out when the CSRF was not validated and you can easily find the IP, or you can have an evaluation of the non-validations. If the level is high, then you need solutions such as reCaptcha. If you are interested I can give you an example of how Fail2Ban works with HoneySpam, which has a log and blocks all IPs that fill in the bait field, but we do it in Discussions.

ADDISON74:
@elidrissidev - if you don't have time just let me know. I can make the changes in the files. 

After we merge this PR we will have to make some changes in #1866 as well.

elidrissidev:
@ADDISON74 Feel free. I'm on vacation for another week.

app/code/core/Mage/Contacts/controllers/IndexController.php
```diff
@@ -57,6 +62,11 @@ public function indexAction()
 
     public function postAction()
     {
+        if (!$this->_validateFormKey()) {
+            $this->_redirectReferer();
+            return;
+        }
+
         $post = $this->getRequest()->getPost();
```
kiatng, line 70: In edge cases where the session has expired, it'll be nice to notify the user "Invalid Form Key. Please resubmit your request." In this case, it's necessary to repopulate the form. As an example:

```suggestion
        $post = $this->getRequest()->getPost();
        if (!$this->_validateFormKey()) {
            Mage::getSingleton('customer/session')->addError(Mage::helper('contacts')->__('Invalid Form Key. Please resubmit your request.'));
            Mage::getSingleton('customer/session')->setContactFormData($post); // To repopulate the form in block.
            $this->_redirectReferer();
            return;
        }
```

But it's even better to validate the form key within the `try catch` block; this way, the form can be repopulated whenever there is an error.

elidrissidev: Good idea!

elidrissidev:
I rebased and applied requested changes.

ADDISON74:
I have to test it, but the various situations that can occur are much better analyzed now. If they are found in the log file then it is a useful information. An increasing number of reports means that the website is under attack and then it is necessary to implement a better level of protection, such as reCaptcha.

fballiano:
since we were talking about https://github.com/OpenMage/magento-lts/discussions/2557 will we already merge the configuration key?

fballiano:
If we've to break something, let's brake it only once :-D

elidrissidev:
I agree, but this PR is meant to be backwards-compatible, hence why targeting 1.9.4.x.

fballiano:
> I agree, but this PR is meant to be backwards-compatible, hence why targeting 1.9.4.x.

also v19 will have a major because we're introducing php7.3 requirement, so it will happen for both v19 and v20

elidrissidev:
> > I agree, but this PR is meant to be backwards-compatible, hence why targeting 1.9.4.x.
> 
> also v19 will have a major because we're introducing php7.3 requirement, so it will happen for both v19 and v20

I think you mean minor version (the middle number), because I was thinking v21 😅

elidrissidev:
What are we doing with this one guys?

ADDISON74:
@elidrissidev - I left my comments in the code.  We need a log exception error written in the file.

elidrissidev:
I already added that @ADDISON74

app/code/core/Mage/Contacts/controllers/IndexController.php
```diff
@@ -96,24 +105,36 @@ public function postAction()
                     );
 
                 if (!$mailTemplate->getSentSuccess()) {
-                    throw new Exception();
+                    Mage::throwException($this->__('Unable to submit your request. Please, try again later'));
                 }
```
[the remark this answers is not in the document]
elidrissidev: I will add the period directly in the CSV without touching the key to not break existing translations.

app/code/core/Mage/Contacts/controllers/IndexController.php
```diff
@@ -63,6 +68,10 @@ public function postAction()
             /** @var Mage_Core_Model_Translate $translate */
             $translate->setTranslateInline(false);
             try {
+                if (!$this->_validateFormKey()) {
+                    Mage::throwException($this->__('Invalid Form Key. Please resubmit your request again'));
+                }
```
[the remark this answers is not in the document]
elidrissidev: I'll re-phrase it to be "Invalid Form Key. Please submit your request again."

elidrissidev:
Rebased and applied the changes to translations.

ADDISON74:
Thank you. Please press the [Resolve conversation] button on the conversations initiated by me and which were resolved by you.

kiatng (approved):

ADDISON74:
I have several PRs to test, including this one. This weekend I will make time for this task.

luigifab (commented):
There is a little error in CSV.
This is not: `"<strong style="color:red">Important!</strong>...`
This is: `"<strong style=""color:red"">Important!</strong>...`
Fully working for us.

elidrissidev:
Good catch @luigifab! will fix soon.

luigifab (dismissed):

ADDISON74:
After this PR is merged it must be modified according to the same format as https://github.com/OpenMage/magento-lts/pull/1866. The value in the config_data table will remain orphaned, but at least we have the same format for both forms.

fballiano:
since we're in a RC phase I'd much rather prefer using the global config value instead of creating another one, this is the same approach that was taken for https://github.com/OpenMage/magento-lts/releases/tag/v19.4.22 and I think we should do the same here.

elidrissidev:
Quoting from the new release [RFC](https://github.com/OpenMage/rfcs/blob/71e1251a1e49479af2928d9a71e0ecf058352ecd/proposed/0002-release-schedule.md#terminology):

> PATCH:
> ...
> - security/stability: if it doesn't warrant a CVE, it doesn't get a PATCH

fballiano:
That only means that we don’t have to part 19.4 and we can be good on 19.5, that’s why i say we can enable it with the global flag instead of a custom one

matteotestoni:
This changes are very welcome.
When is possible merge changes and integrated anti spam for contacts?

fballiano:
@matteotestoni agree, but I think we shouldn't have a specific flag to enable this feature because otherwise I think many people won't enable it.
and in a previous release we treated the same kind of thing as a security concerned and enabled by default for everybody, which i think we should do the same here.

luigifab (approved):

elidrissidev:
Since the new release RFC is accepted, I guess we should leave this until the next major version so we consolidate Contacts form and Newsletter under the global CSRF config.

matteotestoni:
thank you. By now i have spam from contacts without a solution in OpenMage

fballiano:
I'd convert this one to use the global flag and release it for 20.1.0 or we directly release 21.0, which would be about time anyway.

fballiano:
actually when we break compatibility with the current (which would be v20) we should release a new major v21...

elidrissidev:
> Thank you for closing. By now openmage followed by me have spam from contacts without a solution in OpenMage. I think i need to find another platform
> 
> Amasty google recaptcha is not compatible with openmage and other extension are no longer supported by many vendors such as mageworx, amasty and many others.

My only concern is breaking the rules defined by the new RFC this early on since this issue doesn't have a proper CVE to justify merging it in a MINOR or PATCH release.

fballiano:
we may not have a CVE but it's still a security concern

PR #2349, opened 2022-07-31
Block access to all dot files.

colinmollenhour:
This isn't meant to be a full audit of the .htaccess file or security of access to non-public files but currently a vanilla installation exposes the .git directory which if nothing else throws up a red flag for common scanners.

Rather than blocking specific files and directories like .git, .github, etc this will block all dot files (edit: only in the base path) which I think is a very sane out of the box configuration.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -201,6 +201,10 @@
     Allow from all
 
 ###########################################
+## Deny access to all files starting with a .
+    RedirectMatch 404 /\.
+
+###########################################
 ## Deny access to release notes to prevent disclosure of the installed Magento version
 
     <Files RELEASE_NOTES.txt>
--- a/.htaccess
+++ b/.htaccess
@@ -100,6 +100,12 @@
     #RewriteBase /magento/
 
 ############################################
+## Prevent serving "hidden" files like git repository
+
+    RewriteCond %{REQUEST_URI} /\.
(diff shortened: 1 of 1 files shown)


Flyingmana (approved):

tmewes (dismissed):

ADDISON74:
The RedirectMatch directive belongs to module Alias. If it is not enabled it will be a problem because it will not work and not all users have root access.

I would choose to use the RewriteCond/RewriteRule combination because the Rewrite module is obviously enabled, being mandatory for OpenMage to function properly.

If you keep it surround the directive with <IfModule mod_alias.c>.

colinmollenhour:
@ADDISON74 I changed it as you suggested so that mod_alias is not a dependency - the dot files will not be protected if someone is not using `mod_rewrite` which is a bit hazardous but I would venture to guess that almost nobody runs Magento without `mod_rewrite` so this is the best way to add protection without risking a BC break.

fballiano (approved):

ADDISON74 (approved):

ADDISON74:
I did not have time to check what is the best, using / or - in RewriteRule.

colinmollenhour:
> **Which variant would be better?**
I think it doesn't matter because both will match any string and the goal is simply to short-circuit the request and cause a 404 response.

PR #2260, opened 2022-06-28
GUI to edit "Customer Attributes" and "Customer Address Attributes"

fballiano:
Since the beginning of time we always had to rely on commercial modules (or developers) to add/edit/delete "customer attributes" or "customer address attributes". This PR aims to fix this and give the users the power to manage "customer attributes" the same way they do with "product attributes".

This feature was developed mostly using the code of the "product attributes" edit feature, that's why I've copied the license in the file header.

It's a big PR so suggestions/fixes/additions are welcome.

### Manual testing scenarios (*)
1. open the backend
2. open the menu "Customers -> Attributes"
3. Use the two new masks to add, edit or delete customer attributes or customer address attributes

### Screenshots

The new menus
Schermata 2022-06-28 alle 23 50 12

Customer attributes grid
Schermata 2022-06-28 alle 23 51 34

Add/Edit customer attribute
Schermata 2022-06-28 alle 23 52 02

For customer address it's the same thing so I'll not post screenshots of that.

PR #2352, opened 2022-08-01
GUI to Edit EAV Attributes & Sets - Customer

justinbeaty:
### Description (*)
[WIP] This commit was previously part of #2317. It is broken out so that we can discuss ways to integrate the customer EAV attributes into customer groups, and customer forms. Discussions about this are in that PR.

### Related Pull Requests
#2317
#2260 

### Questions or comments

Todo:

- [ ] Check translation strings
- [x] Modify customer attributes to add use_in_forms option
- [x] Modify Customers > Manage Groups to allow us to select an attribute set for each customer group (both customer and address set)


justinbeaty:
In testing the `use_in_forms` function, it seems like adding attributes to any of the frontend forms does not work, and would involve editing template files as mentioned by fballiano here: https://github.com/OpenMage/magento-lts/pull/2317#issuecomment-1201662652

However it does seem to work perfectly in all of the adminhtml areas.

fballiano:
I can confirm you that there's some work on the frontend that needs to be done for the use_in_forms to work :-)

app/code/core/Mage/Customer/Model/Config/Forms.php
```diff
@@ -0,0 +1,72 @@
...
+            ),
+        );
+    }
+
+    /**
+     * Get options in "key-value" format
+     *
+     * @return array
+     */
+    public function toArray()
```
kiatng, line 64: To be consistent with https://github.com/OpenMage/magento-lts/blob/a252af84f7f0ca2c4def48474351d53a137894f8/lib/Varien/Data/Collection.php#L716-L716
```suggestion
   public function toOptionHash()
```

app/code/core/Mage/Customer/Model/Config/Forms.php
```diff
@@ -0,0 +1,72 @@
...
+ */
+class Mage_Customer_Model_Config_Forms
+{
+
+    /**
+     * Options getter
+     *
+     * @return array
+     */
+    public function toOptionArray()
```
kiatng, line 33: I have custom form types, is it possible to define these in `app\code\core\Mage\Customer\etc\config.xml`?

justinbeaty: Would it maybe be better to set up an event to hook into? I'm thinking about the best way a module could define their own form types.

justinbeaty:
I've started support for `use_in_forms` to work on the frontend without the user having to edit templates:

image

Todo:
- Compile the SCSS to make the date field show up correctly.
- Add changes in `app/design/frontend/rwd/default/template/customer/form/edit.phtml` to `app/design/frontend/base` as well
- Support the other frontend forms

fballiano:
wow!

justinbeaty:
Working on the other frontend forms, `checkout_register` is a bit crowded, but it's just how Magento is with checkout & register I guess. I think a store owner should decide to not add too many fields to this form.

image

justinbeaty:
I still have address forms to do tomorrow.

I am also thinking it might be nice to group these fields on the frontend based on the attribute set > groups. This way there is some more organization and the user can label the fieldset. For example in Account Information they could have a group called "Social Media" and let the user enter urls to their pages (just an example.)

But also, I am going to take a stab at allowing you to define attribute set per customer group. Thus it would be important that not only do you define `use_in_forms` but also put that attribute in a attribute set. Otherwise you cannot control what attributes show up for each customer group.

app/code/core/Mage/Customer/Model/Config/Forms.php
kiatng: Using event is ok, but I like the config method. An example is product types:
https://github.com/OpenMage/magento-lts/blob/35acb46f0385484ec45397cc53b379ecf1434ea6/app/code/core/Mage/Catalog/etc/config.xml#L479-L487

and product custom option:
https://github.com/OpenMage/magento-lts/blob/35acb46f0385484ec45397cc53b379ecf1434ea6/app/code/core/Mage/Catalog/etc/config.xml#L531-L537

which I have used to add more options in the HTML select element.

justinbeaty: @kiatng thanks for the examples, I was thinking about it more and agree the config.xml method is indeed better.

kiatng: Ref https://github.com/OpenMage/magento-lts/blob/35acb46f0385484ec45397cc53b379ecf1434ea6/app/code/core/Mage/Adminhtml/Model/System/Config/Source/Product/Options/Type.php#L34

I made a sample code:
```php
class Mage_Adminhtml_Model_System_Config_Source_Customer_Form
{
    const CUSTOMER_FORM_PATH = 'global/customer/form';

    {
        $helper = Mage::helper('customer');
        foreach (Mage::getConfig()->getNode(self::CUSTOMER_FORM_PATH)->children() as $form) {
            $labelPath = self::CUSTOMER_FORM_PATH . '/' . $form->getName() . '/label';
            $forms[] = [
                'label' => $helper->__((string) Mage::getConfig()->getNode($labelPath)),
                'value' => $form->getName()
            ];
        }

        return $forms;
    }
}
```

The `config.xml` looks like this:
```xml
    <global>
        <customer>
            <form>
                <adminhtml_checkout translate="label" module="customer">
                    <label>Adminhtml Checkout</label>
                </adminhtml_checkout>
                <adminhtml_customer translate="label" module="customer">
                    <label>Adminhtml Customer</label>
                </adminhtml_customer>
                <!-- more forms -->
            </form>
        </customer>
    </global>
```

kiatng: That code sample doesn't consider the module attribute. Corrected version:
```php
class Mage_Adminhtml_Model_System_Config_Source_Customer_Form
{
    const CUSTOMER_FORM_PATH = 'global/customer/form';

    {
        foreach (Mage::getConfig()->getNode(self::CUSTOMER_FORM_PATH)->children() as $form) {
            $labelPath = self::CUSTOMER_FORM_PATH . '/' . $form->getName() . '/label';
            $moduleName = $form->getAttribute('module') ?? 'customer';
            $forms[] = [
                'label' => Mage::helper($moduleName)->__((string) Mage::getConfig()->getNode($labelPath)),
                'value' => $form->getName()
            ];
        }

        return $forms;
    }
}
```
Note that `$form->getAttribute('module')` can return `null`.

justinbeaty: @kiatng thanks for the implementation! 

Do you think this XML is better? Since we need to also define `customer_address` forms it keeps all forms under one node, `<form>`.

```xml
<global>
   <form>
      <customer>
         <adminhtml_checkout translate="label" module="customer">
            <label>Adminhtml Checkout</label>
         </adminhtml_checkout>
         <adminhtml_customer translate="label" module="customer">
            <label>Adminhtml Customer</label>
         </adminhtml_customer>
         <!-- more forms -->
      </customer>
      <customer_address>
         <adminhtml_customer_address translate="label" module="customer">
            <label>Adminhtml Customer Address</label>
         </adminhtml_customer_address>
         <customer_address_edit translate="label" module="customer">
            <label>Customer Address Edit</label>
         </customer_address_edit>
         <!-- more forms -->
      </customer_address>
   </form>
</global>
```

I could also move the source models to:
```
Mage_Adminhtml_Model_System_Config_Source_Form_Customer
Mage_Adminhtml_Model_System_Config_Source_Form_Customer_Address
```

to match.

justinbeaty:
@kiatng I'm seeking your opinion on the following ([originally posted here](https://github.com/OpenMage/magento-lts/pull/2317#issuecomment-1184563135))

---

> I use attribute sets to organize the attributes for different customer groups. The project has over 200 customer attributes, each customer group having very different set of attributes.
> 
> My concern is that changes to EAV could make it very difficult to merge OM to existing projects with custom EAV.

The PR as it now should definitely not conflict with anything since it's all new code, but I see your point if we hook up a customer group -> customer attribute set relation. Still, I think it could be worth it maybe just waiting to merge this PR until the next major version.

How did you correlate attribute sets with customer groups? My first thought is to modify the `customer_group` table:
```
mysql> describe customer_group;
+---------------------+-------------------+------+-----+---------+----------------+
| Field               | Type              | Null | Key | Default | Extra          |
+---------------------+-------------------+------+-----+---------+----------------+
| customer_group_id   | smallint unsigned | NO   | PRI | NULL    | auto_increment |
| customer_group_code | varchar(32)       | NO   |     | NULL    |                |
| tax_class_id        | int unsigned      | NO   |     | 0       |                |
+---------------------+-------------------+------+-----+---------+----------------+
3 rows in set (0.00 sec)
```

To add something like `customer_attribute_set_id` and maybe even `customer_address_attribute_set_id`. Probably both are useful in the case you want to control both attribute sets per group.

If we can do it in a way that is maybe compatible with what you've done, then that's at least one site that wouldn't break.

---

In any case, I'll set up an observer so that users can change this logic.

justinbeaty:
I have some customer group -> attribute set integration. Here's screenshots:

image

image

image

So basically I made a new customer attribute set "Special Customer" and then I can add or remove certain attributes to that set. If I then set a group to use that attribute set, attributes will show or hide based on the set in adminhtml and frontend forms.

I will have some extra care to not BC break but not implemented yet.

I think this is a really nice feature because you can easily show or hide certain fields for customers on their edit account page.

justinbeaty:
Also in working on this, I realize there's a lot of `eav_form_*` tables which I assume were to implement similar functionality into EE. This PR doesn't use those, but it's a whole 'nother can of worms to try and implement that functionality into CE.

app/code/core/Mage/Adminhtml/Model/Customer/Observer.php
```diff
@@ -0,0 +1,148 @@
...
+
+        // frontend properties fieldset
+        $fieldset = $form->addFieldset('front_fieldset', array('legend'=>Mage::helper('adminhtml')->__('Extra Properties')));
+
+        $fieldset->addField('use_in_forms', 'multiselect', array(
+            'name'   => 'use_in_forms',
+            'label'  => Mage::helper('adminhtml')->__('Use in Forms'),
+            'title'  => Mage::helper('adminhtml')->__('Use in Forms'),
+            'values' => Mage::getModel('customer/config_forms')->toOptionArray(),
+            'value'  => Mage::getResourceModel('customer/form_attribute')->getFormTypesByAttribute($attribute)
```
justinbeaty, line 57: Just looking at #2260 again, not sure why I implemented a new method instead of using `$attribute->getUsedInForms()`

kiatng:
> @kiatng (and @fballiano) I'm seeking your opinion on the following ([originally posted here](https://github.com/OpenMage/magento-lts/pull/2317#issuecomment-1184563135))
> 
> How did you correlate attribute sets with customer groups? My first thought is to modify the `customer_group` table:
> 
>...
> 
> To add something like `customer_attribute_set_id` and maybe even `customer_address_attribute_set_id`. Probably both are useful in the case you want to control both attribute sets per group.
> 
> If we can do it in a way that is maybe compatible with what you've done, then that's at least one site that wouldn't break.
> 
> In any case, I'll set up an observer so that users can change this logic.

What I did 10 years ago was to add a column `apply_to` to the table `eav_attribute_set`. `apply_to` sets comma-delimited customer groups. I needed to also drop the unique index on `array('entity_type_id', 'attribute_set_name')`. It looks like this: 
image
image

Since then, I also use the same `apply_to` for a custom EAV, which is not for customer group. 

I do not see how you can make it compatible without using the same approach above.

app/code/core/Mage/Customer/Model/Config/Forms.php
kiatng: Ref https://github.com/OpenMage/magento-lts/blob/18e68a223dffc3d8dec6bb3e7ac485a6338baacd/app/code/core/Mage/Customer/Model/Form.php#L35-L35

and https://github.com/OpenMage/magento-lts/blob/18e68a223dffc3d8dec6bb3e7ac485a6338baacd/app/code/core/Mage/Catalog/etc/config.xml#L620-L621

Use the module name like `<catalog>`:

```xml
<global>
   <eav>
     <form>
        <customer>
           <!-- more forms -->
        </customer>
        <customer_address>
           <!-- more forms -->
        </customer_address>
     </form>
   </eav>
</global>
```
or continue with the format `<eav_attributes>`:
```xml
<global>
   <eav_forms>
      <customer>
         <!-- more forms -->
      </customer>
      <customer_address>
         <!-- more forms -->
      </customer_address>
   </eav_forms>
</global>
```
Which do you prefer? I'd go with the latter.

justinbeaty: I like the latter as well :+1:

justinbeaty:
@kiatng So are you able to create two attribute sets that apply to the same customer group, but differ in the form setting? In other words, you can show different fields for the same customer group on different forms?

If so that's a nice feature that my PR doesn't have, but I am not sure it's something to introduce to the core this way. I would rather try and reverse engineer the eav_form tables.

So no this PR wouldn't be compatible with your way, but I think I can make it at least not interfere with what you've done.

kiatng:
> @kiatng So are you able to create two attribute sets that apply to the same customer group, but differ in the form setting? In other words, you can show different fields for the same customer group on different forms?

Yes, in fact sets 19 and 23 in the screenshot I attached are of the same customer group. The use case is for different backend users to view/edit a different set of attributes. At its height, there were a couple hundreds of backend users in this project.

There are other advantages to add `apply_to` at the attribute set level. Because it is upstream (EAV level), there is less coding downstream (customer level and other custom entity). 

Your approach to match an attribute set at the customer group is similar to the product's `attribute_set_id`. However, number of customer groups (I have seen max 30 groups) is much less than the number of products (can be hundreds of thousands). It's easier to configure `apply_to` at the attribute set level for customer groups, but impractical for products.

But what I have is customized to a very particular use case. I only developed what is required of the project, I do not have anything on addresses and categories. Developing for general use cases is a lot more challenging.

matteotestoni (approved):

luigifab:
Will be ready for RC4?

app/design/frontend/rwd/default/template/customer/address/edit.phtml
```diff
@@ -153,6 +153,18 @@
             </li>
         </ul>
     </div>
+    <div class="fieldset">
```
empiricompany, line 156: wrap into 
```
<?php if($this->extraFields()): ?>
...
<?php endif ?>
```

app/design/frontend/base/default/template/eav/widget/text.phtml
```diff
@@ -0,0 +1,26 @@
...
+ * @package     base_default
+ * @copyright   Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
+ * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
+ */
+?>
+<?php $_attribute = $this->getAttribute() ?>
+<?php $_code = $_attribute->getAttributeCode() ?>
+<label for="<?php echo $this->getFieldId($_code)?>"<?php if ($this->isRequired()) echo ' class="required"' ?>><?php if ($this->isRequired()) echo '<em>*</em>' ?><?php echo $this->__($_attribute->getStoreLabel()) ?></label>
+<div class="input-box">
+    <input type="text" id="<?php echo $this->getFieldId($_code)?>" name="<?php echo $this->getFieldName($_code)?>" value="<?php echo $this->escapeHtml($this->getObject()->getData($_code)) ?>" title="<?php echo Mage::helper('core')->quoteEscape($this->__($_attribute->getStoreLabel())) ?>" class="input-text" <?php echo $this->getFieldParams() ?> />
```
empiricompany, line 25: add validation classes
`class="input-text <?php echo $_attribute->getFrontendClass() ?>"`

empiricompany (changes requested):
check if extrafileds exists and add validation classes to text input
I wasn't able to send a commit

empiricompany:
i've made some test on this PR:

- for each type of attribute (custom/address), I can't set the scope value. It's always stuck on 'store view', and in the grid column, it's always null.
- in the frontend checkout, the extra fields are not being printed

fballiano:
there were some bad conflicts, I did what i could, not sure 100% was correct

PR #2358, opened 2022-08-03
Move phpseclib, mcrypt_compat, net_idna2 to composer

fballiano:
In the effort of cleaning our repo from libraries that are managed and updated elsewhere I worked on moving phpseclib, mcrypt_compat and net_idna2 to composer.


fballiano:
At the moment I didn't commit the composer lock since we have quite some PRs that modify the composer.json.
That's why some checks fail (also phpstan).

fballiano:
on 2nd though, I'll move everything to https://github.com/OpenMage/magento-lts/pull/2138

PR #2359, opened 2022-08-03
Block access to various project files

colinmollenhour:
Note, this blocks access to files of these types in directories other than the root directory as well.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -201,12 +201,17 @@
     Allow from all
 
 ###########################################
-## Deny access to release notes to prevent disclosure of the installed Magento version
+## Deny access to other project files to prevent disclosure of the installed Magento version
+## Only robots.txt should be allowed
 
-    <Files RELEASE_NOTES.txt>
+    <FilesMatch (?<!robots)\.txt$>
         order allow,deny
         deny from all
-    </Files>
+    </FilesMatch>
+    <FilesMatch \.(json|lock|md|sample|yml)$>
+        order allow,deny
+        deny from all
+    </FilesMatch>
 
 ############################################
 ## If running in cluster environment, uncomment this
--- a/.htaccess
+++ b/.htaccess
@@ -208,7 +208,11 @@
         order allow,deny
         deny from all
     </FilesMatch>
-    <FilesMatch \.(json|lock|md|sample|yml)$>
+    <FilesMatch (?<!manifest)\.json$>
+        order allow,deny
+        deny from all
+    </FilesMatch>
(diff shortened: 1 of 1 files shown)


fballiano (approved):

ADDISON74:
If we keep modifying .htaccess let's insert the missing piece for PHP8 as it has already been done in other files, less in this one. That it is in this PR or another, I have no problem with that.

colinmollenhour:
> If we keep modifying .htaccess let's insert the missing piece for PHP8 as it has already been done in other files, less in this one. That it is in this PR or another, I have no problem with that.

Sorry, wasn't trying to interfere with the other .htaccess file PR, just addressing some minor security issues that were reported/noticed. I think it is a lot easier to make changes when broken up into separate PRs though with specific fixes instead of major overhauls. Surely there is no reason a PR for the PHP8 update alone would not be accepted.

ADDISON74:
@colinmollenhour - Please check my comments to this PR. What I posted before and you answered was just a reminder of another unresolved issue.

elidrissidev:
Not sure about JSON files (other than composer.json of course), I've seen many extensions with that provide such files that are loaded from the frontend.

fballiano:
@elidrissidev is totally right, we should block every json, we could have a manifest.json for pwa support.

colinmollenhour:
> we should block every json

I assume you meant *not* block?
I see no problem with allowing common files like `manifest.json`

colinmollenhour:
I added a negative lookbehind for `manifest.json` specifically. Is that good or would you prefer to explicitly block only certain files?

I think in a production environment it is likely the user will need to modify this file anyway...

ADDISON74:
1) I would add **sh** to the list of extensions. So that it can no longer check the existence of the file cron.sh, used internally within the cronjob. Other extensions that can be found in OpenMage should also be analyzed, for example csv (exports), log. Even if some directories have protections in their own .htaccess files, an extra measure is welcome.

2) And a harmless proposal **O**rder and **D**eny with a capital letter. The initial Magento file complied with this requirement, but the RELEASE_NOTES chunk was added later in OpenMage and someone introduced lowercases.

fballiano:
> I assume you meant _not_ block? I see no problem with allowing common files like `manifest.json`

yes sorry for the mistake.

fballiano (dismissed):

colinmollenhour:
Changed capitalization and added .sh and .yaml to list of blocked extensions.

ADDISON74 (approved):

fballiano (approved):

PR #2363, opened 2022-08-04
Version bump only for v20

fballiano:
Since we found a problem with release 20.0.15 we'll need to re-release it soon.

This PR bumps the release number, only for v20, on branch 1.9.4.x (I know it may seem confusing).

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -222,7 +222,7 @@ final class Mage
             return array(
                 'major'     => '20',
                 'minor'     => '0',
-                'patch'     => '15',
+                'patch'     => '16',
                 'stability' => '', // beta,alpha,rc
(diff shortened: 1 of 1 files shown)


elidrissidev:
In this case since only a new v20 version will be released, is it not appropriate to update this directly in `20.0` branch?

fballiano:
Usually it's updated on 1.9.4.x and merged in 20.0 at release time, so I wouldn't want to create a problem in the next release when maybe who's preparing the release does a "+1" without recognizing that's been modified in the 20.0 branch.
Let's say I didn't want to change the workflow that's always been used.
Although, thinking about the future... maybe every branch should have its own version number and not the way it's handled now...

elidrissidev (approved):

fballiano:
since i guess at this point this hotfix for v20 will not be release I'm closing this.

PR #972, opened 2020-05-13
Changed sizeof to count

sreichel:
- 3k occurrences of `count` and 119 of `sizeof`
- simplified `count(x) > 0`
- simplified `count(x) == 0`
- strict type compare to `int`

PR #2369, opened 2022-08-05
Replace Remaining sizeof Alias Calls 

Sdfendor:
### Description (*)

`sizeof()` is an alias of `count()` and should be replaced in favor of consistency. In 2020 there has been already an effort to adjust the majority of those cases. 
This PS is a followup and does the following: 

1. Replace remaining occurrences of `sizeof` with `count` **except** those in *lib/Zend* because there is currently an effort to replace *Zend1* with *ZF1-Future* (#2303) and I don't want to interfere here in any way. 
2. Introduce variables for replacement cases where counting is done within a loop.
3. Moved two variable declaration into prior if statement to avoid double count.
4. Make comparision strict (===) because count is always integer.
5. Apply formatting fixes in the immediate surrounding of the replaced code places.

### Related Pull Requests
* #972

### DIFF
--- a/app/design/frontend/rwd/default/template/catalog/product/list.phtml
+++ b/app/design/frontend/rwd/default/template/catalog/product/list.phtml
@@ -48,7 +48,7 @@
     <?php $_iterator = 0; ?>
     <ol class="products-list" id="products-list">
     <?php foreach ($_productCollection as $_product): ?>
-        <li class="item<?php if(++$_iterator === count($_productCollection)): ?> last<?php endif ?>">
+        <li class="item<?php if(++$_iterator === $_productCollectionCount): ?> last<?php endif ?>">
             <?php // Product Image ?>
             <a href="<?php echo $_product->getProductUrl() ?>" title="<?php echo $this->stripTags($this->getImageLabel($_product, 'small_image'), null, true) ?>" class="product-image">
                 <?php /* Based on the native RWD styling, product images are displayed at a max of ~400px wide when viewed on a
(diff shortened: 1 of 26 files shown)


kiatng (approved):
LGTM, not tested.

app/design/frontend/rwd/default/template/catalog/product/list.phtml
```diff
@@ -47,7 +48,7 @@
     <?php $_iterator = 0; ?>
     <ol class="products-list" id="products-list">
     <?php foreach ($_productCollection as $_product): ?>
-        <li class="item<?php if( ++$_iterator == sizeof($_productCollection) ): ?> last<?php endif ?>">
+        <li class="item<?php if(++$_iterator === count($_productCollection)): ?> last<?php endif ?>">
```
elidrissidev, line 51: `count($_productCollection)` can be replaced by `$_productCollectionCount` declared above.

Sdfendor: Thanks, I missed that one.

elidrissidev (dismissed):

lib/Varien/Simplexml/Element.php
```diff
@@ -453,7 +453,7 @@ public function setNode($path, $value, $overwrite=true)
         foreach ($arr1 as $v) {
             if (!empty($v)) $arr[] = $v;
         }
-        $last = sizeof($arr)-1;
+        $last = count($arr)-1;
```
fballiano, line 456: since you spaced out the other "ifs" maybe we can make this one `count($arr) - 1;`? sorry for the nitpicky comment :-)

Sdfendor: fixed

elidrissidev (approved):

fballiano (approved):

PR #2377, opened 2022-08-07
Fixed bug on mini login form.

kiatng:
### Description (*)
The mini login form should only be used on SSL enabled pages. But nowadays, the entire production site should be SSL enabled. Mini login form makes it easier to login from any page.

This PR does 3 things 

1. Fixed broken mini login form due to missing form key.
2. Do not show the mini login if customer is logged in.
3. Set redirect URL after login.
~4. Add forgot password link.~

To test, show the mini login with layout update in any CMS page, I set it on `home`, page ID 2:

```xml
<reference name="right">
        <block type="customer/form_login" name="customer_form_mini_login" before="-" template="customer/form/mini.login.phtml"/>
</reference>
```
It will show like this on the right column:
image

Now try to login, it will redirect to `customer/account/login`, due to missing form key.

With this PR, we can now login. It will also redirect to the page where it was login or to My Account depending on the configuration:
image

### DIFF
--- a/app/code/core/Mage/Customer/Block/Form/Login.php
+++ b/app/code/core/Mage/Customer/Block/Form/Login.php
@@ -95,6 +95,7 @@ class Mage_Customer_Block_Form_Login extends Mage_Core_Block_Template
 
     /**
      * Can show the login form?
+     * For mini login, which can be login from any page, which should be SSL enabled.
      *
      * @return bool
      */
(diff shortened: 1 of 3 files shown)


kiatng:
I should add that the mini login form should only be used on SSL enabled pages.

app/code/core/Mage/Customer/Block/Form/Login.php
```diff
@@ -105,9 +106,25 @@ public function canShowLogin()
...
         // Set redirect URL after login
-        $url = Mage::getStoreConfigFlag(Mage_Customer_Helper_Data::XML_PATH_CUSTOMER_STARTUP_REDIRECT_TO_DASHBOARD)
-            ? Mage::helper('customer')->getDashboardUrl()
-            : Mage::getUrl('*/*/*', ['_current' => true]);
+        if (Mage::getStoreConfigFlag(Mage_Customer_Helper_Data::XML_PATH_CUSTOMER_STARTUP_REDIRECT_TO_DASHBOARD)) {
+            $url = Mage::helper('customer')->getDashboardUrl();
+        } else {
+            /** @var string $pathInfo 'something.html'|'/'|'/path0'|'/path0/path1/'|'/path0/path1/path2'|etc */
+            $pathInfo = $this->getRequest()->getOriginalPathInfo();
+            if (strtolower(substr($pathInfo, -5)) == '.html') {
```
sreichel, line 114:
```suggestion
            if (strtolower(substr($pathInfo, -5)) === '.html') {
```

sreichel (commented):
Tested, LGTM.

sreichel (approved):

fballiano (approved):

Issue #2399, opened 2022-08-10
SOAP heavily broken in 19.4.17


leissbua:
### Preconditions (*)

1. 19.4.17

### Steps to reproduce (*)

1. Do Login via https://www.domain.com/api/v2_soap  WS-I Compliant YES or NO

### Expected result (*)

 Login Result given

### Actual result (*)

WS-I Compliant YES

PHP Fatal error: Procedure 'loginParam' not present in /home/cloudpanel/htdocs/[www.domain.com/releases/2022_08_10_4_10_06_branch_master_soapfix/lib/Zend/Soap/Server.php](http://www.gigaset.com/releases/2022_08_10_4_10_06_branch_master_soapfix/lib/Zend/Soap/Server.php) on line 889

WS-I Compliant NO

SOAP-ERROR: Parsing WSDL: Unexpected WSDL element <schema> in /home/cloudpanel/htdocs/www.domain.com/lib/Zend/Soap/Server.php on line 814

ADDISON74:
Does it work in 19.4.16?

leissbua:
Yes - we did a rollback to 19.4.16 on several live instances that failed to handle SOAP. I guess its related to several XML fixes/checks coming in in the recent 19.4.17. Need to do more research tomorrow, was a long day.

ADDISON74:
I'm sorry I can't help you because I can't test SOAP, but if we have confirmation from you that it works in the previous version, we can identify the PR that generated the problem and revert it. On the SOAP side there were not many changes in the code in the last half year. There are good chances that it is the one mentioned by you.

fballiano:
I'll test tomorrow morning and try to find out the problem.

fballiano:
I did a quick test with postman on a vanilla v19 (actually the 1.9.4.x branch) and it worked normally:
<img width="882" alt="Schermata 2022-08-10 alle 19 43 47" src="https://user-images.githubusercontent.com/909743/183992251-f0eb26bf-3434-4b93-969d-c19293e0789b.png">

I could get the login token as usual

fballiano:
and it seems to be working also on the checkout of the 19.4.17 tagged release.

can't check any further at this specific moment.

ADDISON74:
I don't think that PR is to blame. The changes in it would not generate a situation like this. @leissbua could you please revert that PR and see what happens? If not possible just change the content of the file with that from 19.4.16 and flush the cache.

elidrissidev:
[...]

PR #2242, opened 2022-06-20
Fix xml validation problems

fballiano:
We're trying to introduce XML validation as a github workflow, the commonly used technique is running xmllit with something like:

`find . -type f -iname '*.xml' | xargs -I '{}' xmllint --noout '{}'`

problem is, xmllint is broken because our XMLs have a "name" variable in some places.

Since that variable was always replaced with "Magento" I figured to:
- change the Magento name in OpenMage
- hardcode the OpenMage name in the XMLs so that xmllint works without problems

### Fixed Issues (if relevant)
This PR is part of the fix for https://github.com/OpenMage/magento-lts/issues/1918

### Manual testing scenarios (*)

You've the command line in the task description.
Also check that APIs work.

PR #2405, opened 2022-08-11
Fix targetNamespace in WS-I Compliant

leissbua:
Fixing bad targetNamespace for WS-I Soap

### Description (*)

Bad targetNamespace was used in Mage/Api/etc/wsi.xml

Fixes OpenMage/magento-lts#2399


ADDISON74 (approved):

ADDISON74:
I checked all the modified files in PR #2242 and indeed there is one issue ```urn:OpenMage```. Those files need to be checked by others to be sure nothing is left behind.

fballiano (approved):

fballiano:
I did a search and it seems there is only one file with that mistake

PR #2407, opened 2022-08-11
Error_Processor Refactoring

Sdfendor:
### Description (*)

The `Error_Processor` class in `errors/processor.php` is an old component (last significant changes are from 2016). With this PR I want to give the class a basic refactor and refresh with our current max language level PHP 7.0 (PHPStorm is in PHP 7.0 feature level mode to avoid complications here) and its features in mind. In part this is a preliminary PR for some feature additions I would like to introduce to `Error_Processor` in the future. 
In detail this PR changes: 
* Due to the way `_prepareConfig()` worked, settings from `local.xml` were loaded as `SimpleXMLElement` but some PHPDoc comments implied differently. Subsequent comparisons relied on type juggeling and the fact that `SimpleXMLElement` has a `__toString()` method. The defaults for the config were set as string directly though. 
Config parameters are set as string directly now in the config object. This allows strict comparison and PHPDoc comments match the real program state. 
* PHPDoc is replaced in favor of typing for parameters and return types wherever possible. At the same time redundant PHPDoc comments are removed completely (e.g. Method descriptions containing only the words in the method name or paraphrase the verb). 
* Previously Dynamically used members are introduced, one unused member is removed. 
* General modernization (null comparison instead of `is_null`, Null-Coalescing operator etc.) and formatting

### DIFF
--- a/errors/default/report.phtml
+++ b/errors/default/report.phtml
@@ -18,9 +18,7 @@
  * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 
-/**
- * @var Error_Processor $this
- */
+/** @var Error_Processor $this */
 ?>
 <div class="col-main">
     <div class="page-title">
--- a/errors/default/report.phtml
+++ b/errors/default/report.phtml
@@ -17,6 +17,10 @@
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
  * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
+
+/**
+ * @var Error_Processor $this
+ */
 ?>
 <div class="col-main">
     <div class="page-title">
@@ -85,11 +89,11 @@
         </div>
         <p class="required">* Required Fields</p>
     </form>
-    <?php elseif ('' == $this->reportAction): ?>
+    <?php elseif ($this->reportAction === ''): ?>
         <p><em>Exception printing is disabled by default for security reasons.</em></p>
     <?php endif ?>
 
-    <?php if ($this->reportAction == 'print'): ?>
+    <?php if ($this->reportAction === 'print'): ?>
     <div class="trace-container">
         <div class="trace">
             <pre><?php echo $this->reportData[0] ?><br />
--- a/errors/processor.php
+++ b/errors/processor.php
@@ -247,19 +247,19 @@ class Error_Processor
             $this->_setSkin($skin, $config);
         }
         if ($local !== null) {
-            if (($action = (string)$local->report->action)) {
+            if ($action = (string)$local->report->action) {
                 $config->action = $action;
             }
-            if (($subject = (string)$local->report->subject)) {
+            if ($subject = (string)$local->report->subject) {
                 $config->subject = $subject;
             }
-            if (($emailAddress = (string)$local->report->email_address)) {
+            if ($emailAddress = (string)$local->report->email_address) {
                 $config->email_address = $emailAddress;
             }
-            if (($trash = (string)$local->report->trash)) {
+            if ($trash = (string)$local->report->trash) {
                 $config->trash = $trash;
             }
-            if (($localSkin = (string)$local->skin)) {
+            if ($localSkin = (string)$local->skin) {
                 $this->_setSkin($localSkin, $config);
             }
         }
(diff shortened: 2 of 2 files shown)


tobihille (dismissed):

fballiano:
couldn't we merge it on v19 too?

Sdfendor:
@fballiano I don't see much that speaks against it tbh.

Sdfendor:
After trying it I thought about it: because I started from v20 branch with this a simple base branch change will not work here. I would have to recreate the PR starting from 1.9.4.x, am I right?

fballiano:
did it give you a million commits when you switched to 1.9.4.x? if so I don't know any way to fix it but recreating the PR :-(

fballiano:
it's super easy tho, download this PR as a diff https://github.com/OpenMage/magento-lts/pull/2407.diff and simply apply it on the newly created branch based on 1.9.4.x ;-)

Sdfendor:
I tbh would have cherry picked all the commits one by one from this to the new branch. So applying a diff is definately easier.

fballiano:
oh right that's also possible (didn't think about it :-D) although with many commits may be easier with the diff, whatever methods helps us :-)

justinbeaty:
@Sdfendor FYI, you could fix PRs like this:
```
# Create new branch from 1.9.4.x
git checkout 1.9.4.x -b my_new_error_processor_refactor_1.9.4.x_branch

# Cherry pick or apply diff 

# Force push to branch error_processor_refactor
git push -u origin my_new_error_processor_refactor_1.9.4.x_branch:error_processor_refactor -f
```

Or:

```
# Rename the old 20.0 branch
git branch -m error_processor_refactor my_old_error_processor_refactor_20.0_branch

# Create new branch from 1.9.4.x
git checkout 1.9.4.x -b error_processor_refactor

# Cherry pick or apply diff 

# Force push 
git push -u origin error_processor_refactor -f
```

I usually do option 2 if I am changing a PR from 20.0 to 1.9.4.x or vice versa.

Sdfendor:
Ok thank you for the tip, I will try one of the options :-).

Sdfendor:
@justinbeaty @fballiano Thank you two, option 2 with the applied diff seems to have worked flawlessly. Now with 1.9.4.x as target only the two files that I touched show up as changed.

errors/processor.php
```diff
@@ -291,31 +239,31 @@ protected function _prepareConfig()
...
+                $config->email_address = $emailAddress;
             }
-            if ((string)$local->report->trash) {
-                $config->trash = $local->report->trash;
+            if (($trash = (string)$local->report->trash)) {
+                $config->trash = $trash;
             }
-            if ((string)$local->skin) {
-                $this->_setSkin((string)$local->skin, $config);
+            if (($localSkin = (string)$local->skin)) {
```
fballiano, line 262: Do we need the double parenthesis?

errors/default/report.phtml
```diff
@@ -17,6 +17,10 @@
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (http://www.magento.com)
  * @license     http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
+
+/**
+ * @var Error_Processor $this
```
fballiano, line 22: Can we maybe make this 1 line instead of the 3? Not sure if it would look better. Not the most important thing obviously

fballiano:
> @justinbeaty @fballiano Thanks to you two, option 2 with the applied diff seems to have worked flawlessly. Now with 1.9.4.x as target only the two files that I touched show up as changed.

great! sorry for my 2 new comments, could have done that before :-(

errors/processor.php
Sdfendor: No we do not, good catch. Fixed.

fballiano (approved):

fballiano (approved):

sreichel:
LGTM.

Maybe in another PR, we should explain how to customize error pages. It is not mentioned in official M1 guide.

We can copy this part https://devdocs.magento.com/guides/v2.4/install-gde/install/cli/install-cli-subcommands-maint.html#multi-store-setups and add it to `docs/` or re-enble github-wikipages (@Flyingmana)

sreichel (approved):
Tested.

fballiano (approved):

PR #2411, opened 2022-08-11
Moved phpseclib, mcrypt_compat, Cm_RedisSession, Cm_Cache_Backend_Redis and Pelago_Emogrifier to composer

fballiano:
This PR is the start of the migration of 3rd party modules to composer dependencies.
This PR continues from https://github.com/OpenMage/magento-lts/pull/2138 which I couldn't rebase from 20.0.

Notes:
- Zend Framework is left out because it needs a dedicated tasks
- Once we merge this I'll work on the release builder to create a release tarball file for each supported PHP version


app/Mage.php
```diff
@@ -47,8 +47,13 @@
...
 
-include_once "phpseclib/bootstrap.php";
-include_once "mcryptcompat/mcrypt.php";
+/** AUTOLOADER PATCH **/
+if (file_exists($autoloaderPath = BP . DS . 'vendor/autoload.php') ||
+    file_exists($autoloaderPath = BP . DS . '../vendor/autoload.php')
+) {
+    require $autoloaderPath;
+}
+/** AUTOLOADER PATCH **/
```
sreichel, line 56: Is thi required for this PR?

composer.json
```diff
@@ -65,7 +70,8 @@
   },
   "config": {
     "allow-plugins": {
-      "magento-hackathon/magento-composer-installer": true
+      "magento-hackathon/magento-composer-installer": true,
+      "cweagans/composer-patches": true
```
sreichel, line 74: Seems not required?

app/Mage.php
fballiano: I added it since now composer is required so also the patch so it seems that we can directly have the autoload built in in our core instead of having composer to constantly patch the app/Mage.php

composer.json
fballiano: Don't we need it for recursive patch install if defined by a dependency?
I think it was added automatically by requiring some modules and (since it would keep asking to allow the plugin, which would break the workflow) I left it it.
I'll try to rebuild it to check if it keep being added.

fballiano: you were right, removed

sreichel (dismissed):

app/Mage.php
justinbeaty: Double check this because I think it might break someone who has set a custom vendor-dir:

https://getcomposer.org/doc/06-config.md#vendor-dir

I think @Flyingmana mentioned this in another PR or issue but I’m on mobile now and can’t find it.

sreichel: @justinbeaty thanks. Never used a custom vendor dir ...

Tested with ...
```
{
    "config": {
        "vendor-dir": "test"
    }
}
```

Patched Mage.php looks like ...

```
/** AUTOLOADER PATCH **/
if (file_exists($autoloaderPath = BP . DS . 'test/autoload.php') ||
    file_exists($autoloaderPath = BP . DS . '../test/autoload.php')
) {
    require $autoloaderPath;
}
/** AUTOLOADER PATCH **/
```

:+1:

fballiano: ah! didn't even know this possibility existed. I'll remove the patch then.

kiatng:
There is always this check `if (function_exists('idn_to_ascii')) {` before `Net_IDNA2` is used:

https://github.com/OpenMage/magento-lts/blob/e6f08517af7e8ce1b166905737874da721e4427e/app/code/core/Mage/Core/Helper/Url.php#L174-L179

also `if (function_exists('idn_to_utf8')) {`:
https://github.com/OpenMage/magento-lts/blob/e6f08517af7e8ce1b166905737874da721e4427e/app/code/core/Mage/Core/Helper/Url.php#L196-L201

When I checked the PHP doc, `idn_to_ascii` and `'idn_to_utf8` are functions from PHP `intl` extension. So, if we make `intl` as required, can we remove `Net_IDNA2` entirely?

fballiano:
@kiatng wow this is a great find, I'm checking it out more in detail now :-)

fballiano:
I've removed Net_IDNA and switched to a proper polyfill which is allowing us to not have a "if funcion exists" in our code, I kinda prefer this solution, I hope you too :-)

justinbeaty:
@fballiano nice work! The polyfill definitely seems too big to just add to functions.php so I like it being separate (and not maintained by us).

I think it will however make the release builder more complicated. But if we can figure out a way to bundle any composer package in lib we could unlock a lot of possibilities like the php8 polyfill package or also removing a lot of our own polyfills from this repo.

fballiano:
mmm I think there's no way to move everything in lib because of dependancies, but if we have the vendor folder in the release it should be solved.

I know... it will make the release bigger but... i don't know any other way...

maybe set the composer install dir as "lib"? mmmmmmmmmmm seems dangerous

justinbeaty:
@fballiano Bundling vendor would be okay IMO. I basically do the same thing for the sites I manage. I only use composer locally and then build it by copying vendor to htdocs.

There’s one issue I don’t like about it though, Magento source code is duplicated there as well as the code for all modules. I have it on my roadmap to eventually see if it’s possible to avoid that.

justinbeaty:
Maybe this could help: https://github.com/liborm85/composer-vendor-cleaner

Not sure if it could remove the entire openmage/magento-lts dir in vendor, but maybe that can just be rm -rf’d

My plans were to always do something like loop through all packages and remove any that had magento-source or magento-module in their composer.json

fballiano:
yes i totally agree that it's really not nice to have the duplicated magento code, i think the composer plugin (i'll check it out asap) is a really good idea, not only for the production servers but also for local environment (like when I try to search on phpstorm i get duplicated results because of the vendor folder).

I'll take a look at it asap, thanks so much!

fballiano:
actually it seems OM is not in the vendor folder here

Schermata 2022-08-16 alle 12 48 30

it would be only in a "final" project, but not here

justinbeaty:
Ah, you're right since this repo doesn't require itself. But for example the `colinmollenhour/magento-redis-session` directory could go. Anything with `"type":"magento-module",` in its `composer.json` file should be safe to remove I think.

fballiano:
> Maybe this could help: https://github.com/liborm85/composer-vendor-cleaner

it seems this plugin runs BEFORE deploy, so we can't use it otherwise the deployment of modules brakes :-(

justinbeaty:
> > Maybe this could help: https://github.com/liborm85/composer-vendor-cleaner
> 
> it seems this plugin runs BEFORE deploy, so we can't use it otherwise the deployment of modules brakes :-(

Hm, what if we don't use it to clean up mage modules but just things like bin or other stuff to remove bloat from the dist? We could have a second manual step after composer install completes to remove any mage modules from vendor.

Alas this is also a discussion for the release builder PR but I thought to mention it because of the polyfill.

fballiano:
Checking the documentation I think the only way to do that is that the https://github.com/Cotya/magento-composer-installer/ actually does it, but it's not a feature it has at the moment.

So maybe I can modify the release builder for that, let's see if it gets accepted, but it will be done after this one is merged.

fballiano:
we can have something like this in the composer.json

```
  "scripts": {
    "post-install-cmd": [
      "rm -rf vendor/colinmollenhour/cache-backend-redis",
      "rm -rf vendor/colinmollenhour/magento-redis-session",
      "find vendor -iname '*.md' -delete"
    ]
  }
```

but I don't know... it's not possible to get the vendor directory name (in case it's a custom one) and it would work only on linux/mac

sreichel:
> we can have something like this in the composer.json
> 
> ```
>   "scripts": {
>     "post-install-cmd": [
>       "rm -rf vendor/colinmollenhour/cache-backend-redis",
>       "rm -rf vendor/colinmollenhour/magento-redis-session",
>       "find vendor -iname '*.md' -delete"
>     ]
>   }
> ```
> 
> but I don't know... it's not possible to get the vendor directory name (in case it's a custom one) and it would work only on linux/mac

There is a variable for composer vendor dir ... https://getcomposer.org/doc/03-cli.md#composer-vendor-dir

fballiano:
yep but if I do something like

```
  "scripts": {
    "post-install-cmd": [
      "rm -rf {$COMPOSER_VENDOR_DIR}/colinmollenhour/cache-backend-redis",
      "rm -rf '{$COMPOSER_VENDOR_DIR}/colinmollenhour/magento-redis-session'",
      "find '{$COMPOSER_VENDOR_DIR}' -iname '*.md' -delete"
    ]
  }
```

that variable doesn't get interpolated so I can't use it

justinbeaty:
We might be able to do more in a PHP script as well as make it cross platform. I'll try some things.

fballiano:
we can do this:

```
  "extra": {
    "branch-alias": {
      "dev-main": "1.9.4.x-dev"
    },
    "magento-root-dir": ".",
    "magento-deploystrategy": "copy",
    "magento-force": true
  },
  "scripts": {
    "post-install-cmd": [
      "rm -rf vendor/colinmollenhour/cache-backend-redis",
      "rm -rf vendor/colinmollenhour/magento-redis-session",
      "find vendor -iname '*.md' -delete"
    ]
  }
```

- with copy strategy we can remove stuff from vendor (otherwise, with symlink everything would be broken)
- only scripts from the root composer.json are executed, dependecies are never executes (safe)
- it's better to have the copy strategy for a release builder anyway

what do you think? should I add those lines?

justinbeaty:
Yes to those lines. Definitely copy method.

justinbeaty:
I am not too sure about using the `post-install-cmd` event, though. I think that's going to run when someone else requires openmage, and we shouldn't modify their vendor dir. Also, this should be done only in --no-dev mode.

fballiano:
noup they're not run https://getcomposer.org/doc/articles/scripts.md
Schermata 2022-08-16 alle 19 11 12

justinbeaty:
@fballiano Thanks for clarifying that. I still have one proposal I'm working on for the script block.

sreichel:
> what do you think? should I add those lines?

"copy" strategy is okay for production, but when developing I prefer symlinks. (can be set separately)

I don't like the script ... please do not touch vendor.

justinbeaty:
My suggestions:

1) Use
```
"magento-deploystrategy": "copy",
"magento-deploystrategy-dev": "symlink"
```

2) Change `post-install-cmd` to `build-release` and call `composer run-script build-release` in release builder workflow.

fballiano:
used this

"magento-deploystrategy": "copy",
"magento-deploystrategy-dev": "symlink"

and removed the scripts

sreichel:
Was something wrong with having these commands in release-builder workflow? If not I'd prefer this.

fballiano:
> Was something wrong with having these commands in release-builder workflow? If not I'd prefer this.

exactly, they can be there and no need to be here

sreichel:
> I've removed Net_IDNA and switched to a proper polyfill which is allowing us to not have a "if funcion exists" in our code, I kinda prefer this solution, I hope you too :-)

I do, but for release-builder ...?!?

Just an idea, write a magento module that requires `symfony/polyfill-...` with path mapping (a normal extension) and require it in composer.json.

**OR**

`composer require "ext-intl:*"` 😎

fballiano:
Sorry but i dont understand.
The release builder will kinda have to have the vendor folderc if that was the question

sreichel:
My idea was not to have vendor folder in built releases.

However ... my preferred solution is to add `ext-intl`.

justinbeaty:
I don't mean to duplicate this conversation over several PRs, but I wanted to share a different take on the release builder script: https://github.com/fballiano/openmage/pull/3/files

I think we have no other choice other than to bundle the vendor dir for the releases, because it's for people who don't use composer. The script I linked above would also make an accurate composer.json that could be used to recreate the vendor dir.

Do we maybe move the release builder talk to #2165 ?

fballiano:
The "cleanup" plugin works now (it was a misconfiguration that the developer explained to me) to I've readded it

fballiano:
At the moment I've removed the composer lock in this PR.

fballiano:
It should be ok now, do you want me to readd composer.lock?

sreichel:
Current composer.lock works for 7.3 upwards. As long we have no dependencies that require different versions, we can keep it. (???)

fballiano:
readded composer.lock generated with php73

sreichel:
I think emogrifier can be readied too. Different tasks were only required for 7.1/7.2.

fballiano:
good note, I've readded it to this PR

fballiano:
huston, we've green lights 😄

sreichel:
Tested a bit ... does `liborm85/composer-vendor-cleaner` really belongs to this PR?

fballiano:
> Tested a bit ... does `liborm85/composer-vendor-cleaner` really belongs to this PR?

I think it's useful but, at the end of the day, it's now only removing the .md files so I'll remove it altogether.

I'm copying here the configuration for my future reference:

```
{
  "name": "openmage/magento-lts",
  "description": "A fork of Magento-1 that is accepting bug fixes (backward compatible, drop in replacement for official Magento)",
  "license": [
    "OSL-3.0",
    "AFL-3.0"
  ],
  "type": "magento-source",
  "require": {
    "php": ">=7.3 <8.2",
    "ext-ctype": "*",
    "ext-curl": "*",
    "ext-dom": "*",
    "ext-gd": "*",
    "ext-hash": "*",
    "ext-iconv": "*",
    "ext-intl": "*",
    "ext-json": "*",
    "ext-libxml": "*",
    "ext-mbstring": "*",
    "ext-pdo": "*",
    "ext-simplexml": "*",
    "ext-soap": "*",
    "ext-zlib": "*",
    "colinmollenhour/cache-backend-redis": "^1.14",
    "colinmollenhour/magento-redis-session": "^3.0",
    "liborm85/composer-vendor-cleaner": "^1.7",
    "pelago/emogrifier": "^6.0",
    "phpseclib/phpseclib": "^3.0.14"
  },
  "require-dev": {
    "dealerdirect/phpcodesniffer-composer-installer": "^0.7.2",
    "friendsofphp/php-cs-fixer": ">=3.4",
    "macopedia/phpstan-magento1": "^1.0.4",
    "magento-ecg/coding-standard": "^4.5",
    "phpcompatibility/php-compatibility": "^9.3",
    "phpmd/phpmd": "^2.12",
    "phpstan/phpstan": "^1.8.5",
    "phpunit/phpunit": "^9.5",
    "squizlabs/php_codesniffer": "^3.7"
  },
  "conflict": {
    "n98/n98_layouthelper": "*"
  },
  "authors": [
    {
      "name": "Lee Saferite",
      "email": "lee.saferite@openmage.org",
      "role": "Maintainer"
    },
    {
      "name": "Daniel Fahlke aka Flyingmana",
      "email": "flyingmana@googlemail.com",
      "role": "Maintainer"
    },
    {
      "name": "David Robinson",
      "email": "david.robinson@aoe.com",
      "role": "Maintainer"
    },
    {
      "name": "Tymoteusz Motylewski",
      "role": "Maintainer"
    },
    {
      "name": "Sven Reichel",
      "email": "github-sr@hotmail.com",
      "role": "Maintainer"
    }
  ],
  "autoload-dev": {
    "psr-4": { "OpenMage\\Tests\\Unit\\": "dev/tests/unit" }
  },
  "extra": {
    "branch-alias": {
      "dev-main": "1.9.4.x-dev"
    },
    "magento-root-dir": ".",
    "magento-deploystrategy": "copy",
    "magento-deploystrategy-dev": "symlink",
    "dev-files": {
      "/": [
        "*.md"
      ]
    }
  },
  "config": {
    "allow-plugins": {
      "magento-hackathon/magento-composer-installer": true,
      "dealerdirect/phpcodesniffer-composer-installer": true,
      "liborm85/composer-vendor-cleaner": true
    },
    "dev-files": {
      "remove-empty-dirs": false
    },
    "sort-packages": true
  }
}
```

sreichel:
Just need to test emogrifier ... (do you have a sample mail template, with all possibilities to test against?)

fballiano:
no :-( I tested it with a template that I created on the fly

shaughnlgh (approved):

sreichel (approved):

fballiano:
fixed conflict with 1.9.4.x

fballiano:
I didn't notice the conflict but other than that there's nothing more to do on this one I think

sreichel:
As composer user i'm happy with it, but is release builder ready to merge this one for next release?

(Discussion? Composer as requirement?)

fballiano:
> As composer user i'm happy with it, but is release builder ready to merge this one for next release?
> 
> (Discussion? Composer as requirement?)

ah ok ok, I would have worked on that right after merging this one, resuming the other "release builder workflow" PR :-)

fballiano:
Fixed conflict regenarating composer.lock

will we ever merge this?

sreichel:
> will we ever merge this?

I hope so. Maybe its better to split that PR and start with one replacement? (Cm_Cache?)

sreichel:
@fballiano what is needed to get it merged? (I'd like to help, but how?)

fballiano:
@sreichel I think it's completed, we just will need a release building which is here: https://github.com/OpenMage/magento-lts/pull/2165 but I'll work on that when this one will be merged ;-)

sreichel (approved):

colinmollenhour (approved):

fballiano:
wow, let's go, time has come 🤞🤞🤞

PR #2413, opened 2022-08-12
Make php7.3 minimum requirement

sreichel:
### Description (*)

https://github.com/OpenMage/magento-lts/discussions/2402


Sdfendor (approved):

tmewes (approved):

Sdfendor:
After this PR anyone is able to add code that is no longer compatible with 7.0, The automated syntax check for 7.0 should therefore be removed, shouldn't it? When/in which PR is the right place to do so?

tobihille (dismissed):

fballiano:
once we change this should we release new major releases instead of minor?

sreichel:
> once we change this should we release new major releases instead of minor?

I'm just here for code ... not versioning ... :rofl:

tobihille (dismissed):

Sdfendor (dismissed):

tmewes (dismissed):

elidrissidev:
> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> 1. MAJOR version when you make incompatible API changes

@fballiano So yes, technically such a change should be in a new major version, but it's not set on stone.

tobihille (dismissed):

tmewes (dismissed):

Sdfendor (approved):

fballiano:
Let's wait a few days (since we just published a new release) and then I'm pro merging this one

fballiano:
why are these done on push and pull?
Schermata 2022-09-05 alle 14 10 24

also, if we fix the conflict I'd approve it right away, it's time for this.

ADDISON74:
The other day I was reading an article in which the called Docker one of the IT wonders of the last 10 years. Those who already use Docker can take a step towards DDEV. I would encourage using DDEV for OpenMage testing. I would mention this link https://ddev.readthedocs.io/en/stable/users/quickstart/#openmagemagento-1 in README file.

I would also provide all the necessary steps for installation. In Windows OS, the ```mutagen_enable: true``` option is mandatory, otherwise running OpenMage it is as slow as on XAMPP/WAMP.

PS - please write DDEV in uppercase, not lowercase.

README.md
```diff
@@ -19,21 +19,26 @@ level of backwards compatibility to the official releases.
...
 that will be independently maintained with upstream patches and community bug fixes for as long as it makes sense
-to do so (based on available resources). For example, Magento version `1.9.3.4` was merged into the `1.9.3.x` branch.
+to do so (based on available resources). For example, Magento version `1.9.4.5` was merged into the `1.9.4.x` branch.
 
 ## Requirements
 
-- PHP 7.0+ (PHP 7.3 with OpenSSL extension strongly recommended and verified compatible) (PHP 7.4 and 8.0 are supported)
-- MySQL 5.6+ (8.0+ recommended)
-- (optional) Redis 5+ (7.x recommended, latest verified compatible 7.0.4 with 20.x)
+- PHP 7.3+ (PHP 8.0 is supported)
```
fballiano, line 31: should we add also PHP 8.1?

sreichel:
> why are these done on push and pull?

I think one for newly created PR and one for added commits?!?

fballiano:
but it shouldn't be that way, the other PRs don't have all the workflow duplicated, one run per the full PR should be enough

fballiano:
maybe it's my fault in reading the workflow execution, dunno

README.md
sreichel: I have removed because i think OM is not completly tested with 8.1

E.g. https://github.com/OpenMage/magento-lts/pull/1198#issuecomment-1236416427

Maybe we should discuss it there: https://github.com/OpenMage/magento-lts/discussions/2253

fballiano: hmm ok, i said it since in composer.json we already have "php": ">=7.0 <8.2",

sreichel:
Maybe because i've created the branch a OM repo? (Yea, i already changed it for other PRs)

fballiano (approved):

README.md
sreichel: hmmm, .... this should be discussed. (composer.lock is for php8.1.)

ADDISON74 (approved):

Sdfendor (approved):

README.md
elidrissidev: I'm already working on a multi-part PR for PHP 8.1 compatibility (continuing from @luigifab's previous PR), part 1 should be done this week.

fballiano:
@sreichel should we merge?

PR #2415, opened 2022-08-12
Product reviews backend: fixed grid ID typo preventing save button to show

fballiano:
It was reported (https://github.com/OpenMage/magento-lts/pull/1769#issuecomment-1212864929) that because of https://github.com/OpenMage/magento-lts/pull/1769 it's not possible to create a product review from backend anymore.

This is because we changed grid ID but we didn't notice that the formed ID was used in a javascript.

With this PR we're fixing this ID too so that the original functionality is restored.


ADDISON74 (approved):

Sdfendor (approved):

Sdfendor:
Thanks for the quick fix!

ADDISON74:
This time I tested it locally and the issue was solved.

elidrissidev (approved):
Thanks man.

PR #2421, opened 2022-08-14
phpstan: some fixes

sreichel:
### Description (*)

- 1st commit ...
- 2nd/3rd call helper/model instead of wrapper method, that cant be resolved
- 4th added variable + DOCblocks that can be resolved


app/code/core/Mage/Catalog/Block/Widget/Link.php
```diff
@@ -68,18 +68,15 @@ public function getHref()
             $idPath = explode('/', $this->_getData('id_path'));
 
             if (isset($idPath[0]) && isset($idPath[1]) && $idPath[0] == 'product') {
-
-                /** @var Mage_Catalog_Helper_Product $helper */
-                $helper = $this->_getFactory()->getHelper('catalog/product');
+                $helper = Mage::helper('catalog/product');
```
kiatng, line 71: Is this BC? Suppose there is an override of `_getFactory()` or `getHelper()`?

sreichel:
@kiatng i'll revert mentioned lines, because change was not necessary, but one question remains ... is not using `getHelper($someName)` a BC breaking change?

Why would you override ...

```
    public function getHelper($helperClass)
    {
        return Mage::helper($helperClass);
    }
```

.. instead of rewriting the helper class itself?

sreichel:
I'll split in in samller chuncks.

app/code/core/Mage/Catalog/Block/Widget/Link.php
```diff
@@ -79,7 +78,7 @@ public function getHref()
                 $categoryId = $idPath[1];
                 if ($categoryId) {
                     /** @var Mage_Catalog_Helper_Category $helper */
-                    $helper = $this->_getFactory()->getHelper('catalog/category');
+                    $helper = $this->_getFactory()->getHelper('catalog/product');
```
fballiano, line 81: was this correct to change the model from category to product?

Schermata 2022-08-19 alle 00 01 47

sreichel: No, this was not correct. Mistake when reverting ... sorry.

kiatng:
> Why would you override ...

I do no such thing. But we know there are all sorts of extensions, some of which do crazy things.

PR #2424, opened 2022-08-14
Updated validation for modelName in Mage_Shipping_Model_Config::_getCarrier() and Mage_Payment_Model_Config::_getMethod()

eneiasramos:

1. ...
2. ...


app/code/core/Mage/Shipping/Model/Config.php
```diff
@@ -102,6 +102,11 @@ protected function _getCarrier($code, $config, $store = null)
...
         $modelName = $config['model'];
 
+        $className = Mage::getConfig()->getModelClassName($modelName);
+        if (!mageFindClassFile($className)) {
+            return false;
+        }
+
         /**
          * Added protection from not existing models usage.
          * Related with module uninstall process
```
kiatng, line 112: **Discussion**
If the model exists, it will call `getModelClassName` twice. `getModel()` check for `class_exists`:
https://github.com/OpenMage/magento-lts/blob/59629098f42ce7d4e2052c989f3fbd5523fb2ace/app/code/core/Mage/Core/Model/Config.php#L1382-L1385
 `mageFindClassFile` checks for `file_exists`:
https://github.com/OpenMage/magento-lts/blob/59629098f42ce7d4e2052c989f3fbd5523fb2ace/app/code/core/Mage/Core/functions.php#L90-L96. 

But I think in this case, lines 105 to 120 (including `try catch` block) can be simplified to

```suggestion
            /** @var Mage_Shipping_Model_Carrier_Abstract $carrier */
            $carrier = Mage::getModel($modelName);
            if (!$carrier) {
                return false;
            }
```

The same can be applied here too:
https://github.com/OpenMage/magento-lts/blob/59629098f42ce7d4e2052c989f3fbd5523fb2ace/app/code/core/Mage/Payment/Model/Config.php#L92-L97

What do you think?

eneiasramos: I made the proposed changes :)

kiatng (dismissed):
Thanks.

fballiano:
@kiatng did you understand what's this about? it is safe to remove those checks?

kiatng (approved):
@fballiano This PR fixes the round-about way to load the model. Let's take a look why I say so:

https://github.com/OpenMage/magento-lts/blob/f9d29b9557da4ab2eb1ff0d57d86d1c05f9089b0/app/Mage.php#L513-L516

This is the innards of `getModelInstance()`:

https://github.com/OpenMage/magento-lts/blob/f9d29b9557da4ab2eb1ff0d57d86d1c05f9089b0/app/code/core/Mage/Core/Model/Config.php#L1443-L1454

Note on line 1445, it invokes  `getModelClassName()`. 

Now look at the PR and the changes in question: 

```php
        $className = Mage::getConfig()->getModelClassName($modelName);
        if (!mageFindClassFile($className)) {
            $method = Mage::getModel($modelName);
```
`getModelClassName()` will be called twice, and this `if (!mageFindClassFile($className)) {` is redundant, because it is validated again, even though differently: `if (class_exists($className)) {` but the effect is the same.

Therefore, this PR is valid, it makes the code much cleaner.

eneiasramos:
@fballiano @kiatng thank you!

fballiano (approved):

Issue #2432, opened 2022-08-16
Image uploader broken?


justinbeaty:
### Preconditions (*)

1. OpenMage 20.0.16 - clean install
2. Chrome Version 104 / Windows 10
3. Firefox Version 103 / Ubuntu 20.04

### Steps to reproduce (*)

1. Edit a product, go to images tab
4. Press browse button and select image

![image](https://user-images.githubusercontent.com/51970393/184871216-455be9f5-e401-4113-8ae9-7cd8dce9073b.png)

### Expected result (*)

1. Image uploads

### Actual result (*)

1. Browser hangs and becomes unresponsive. Eventually tab crashes with out of memory.

Is this happening to anyone else? I don't use the image in manage products so I haven't checked it in a while.

fballiano:
with 20.0 branch i could upload no problem
<img width="1496" alt="Schermata 2022-08-16 alle 12 52 40" src="https://user-images.githubusercontent.com/909743/184873154-e3f85c6d-bcee-4ffd-844b-801f78dd973b.png">

fballiano:
also with legacy theme

justinbeaty:
Thanks for the confirmation. It could be my local server config perhaps. I'll close this for now and investigate.

justinbeaty:
Getting to the bottom of this, it looks like this happens if either `post_max_size` or `upload_max_filesize` are set to `0` (which should mean unlimited.)

It creates an infinite loop in JS that crashes the browser. I'll reopen this for now but am preparing a PR to fix it.

sreichel:
@all-contributors add @justinbeaty bug

PR #2433, opened 2022-08-16
Flow.js updates

justinbeaty:
### Description (*)

We use `flow.js` to upload images in the backend. However, the version in this repo is from Feb 01, 2015. 

In this PR:
- Updated flow.js to [v2.14.1](https://github.com/flowjs/flow.js/releases/tag/v2.14.1)
- Removed [fusty-flow.js](https://github.com/flowjs/fusty-flow.js) -- A JavaScript library which extends Flow.js and allows to use not resumable uploads for older browsers, such as IE7, IE8 and IE9.
- Fixed an error where setting post_max_size or upload_max_filesize to 0 causes the admin dashboard to crash when uploading a file. Ref #2432
- Fixed a typo "Disalollowed" -> "Disallowed". This was done in the phtml file but I can change that pending talks in #2429

1. Fixes OpenMage/magento-lts#2432

1. Pull this PR into your source code
2. Make sure to clear cache and hard refresh the edit product page
3. Try to upload some files, hopefully all should work


ADDISON74:
LGTM but before I approve it I will test it.

Have you looked in the browser console to check if there are any errors? Also, did you make a comparison between the versions so that there are no surprises that the removed version was one customized by the Magento team?

justinbeaty:
I do not see any browser errors with the new version.

I did a basic check between versions, you can see the commits [here](https://github.com/flowjs/flow.js/compare/v2.9.0...v2.14.1
). It mostly looks like bugfixes, a few new features, and a lot of documentation.

There's one thing I want to check here, but I think it's still maybe a local server issue.

justinbeaty:
Actually, I'm really confused *why* Magento developers are using flow.js. The point of it is that it can upload files in chunks, but they don't use that functionality at all. Instead they set the chunkSize to the max upload filesize, so flow.js is pointless. They probably used it as a quick upload library when Flash was deprecated.

In any case, the error again seems to come from me setting post_max_size and upload_max_filesize to 0, but I will make the fix better in this PR.

ADDISON74:
That comparison is valid only if the file used by us has the same content. I will still do a check like I did for ZF1-Future, when I came to the conclusion that even in /lib/Zend the Magento team got their hands up to the elbows.

justinbeaty:
> That comparison is valid only if the file used by us has the same content. I will still do a check like I did for ZF1-Future, when I came to the conclusion that even in /lib/Zend the Magento team got their hands up to the elbows.

Sorry I misunderstood what you asked. It looks like no modifications were made:

```
$ sha1sum *
bfcbe640f2c3ec26d11c9445b138292d7dc0898a  git-flow.min.js
bfcbe640f2c3ec26d11c9445b138292d7dc0898a  om-flow.min.js
```

Here is sha1sum of the file in our repo, and the 2.9.0 release from here: https://github.com/flowjs/flow.js/releases/tag/v2.9.0

justinbeaty:
Fixed the error caused by my config, changes here: https://github.com/OpenMage/magento-lts/compare/33cb0666acfbc2c990951395c82e1a4779e3a37d..9cc750f2a41f54c1f8fafe841723801b33308cb1

Again I am just left scratching my head wondering why Magento developers used flow.js in the first place...

fballiano:
> Again I am just left scratching my head wondering why Magento developers used flow.js in the first place...

let's drop it altogether :-D
maybe for big files?

justinbeaty:
> > Again I am just left scratching my head wondering why Magento developers used flow.js in the first place...
> 
> let's drop it altogether :-D maybe for big files?

I would like to drop it too, but that's for another PR because it needs much more changes.

For large files I think that's the point of flow.js, but I don't see anywhere the multiple chunks are read in Magento... And Magento is simply setting the chunkSize to the max upload file size, so there would be no chunks anyway (probably because the PHP code doesn't read them...)

ADDISON74:
Many things have not received a response from the Magento team. From the moment the attention and care moved to the M2 until the grand finale, solutions taken at the corner of the table were preferred. I can't forget the Stackoverflow discussions when bugs appeared for each version and the community reacted. It was solved on one side and changed in place to another version. Boredom was clearly seen starting with 1.9.4.0. Openmage was the place where many of the issues found a solution because those who maintained the project gathered information from the Forums, listened to the opinions. I was brought here by Sven who advised me not to post in the Magento Bugtracker so that no one would notice me. To conclude, it is obvious that we only now realize that the replacement of the uploader based on Flash Player was not an afterthought, but was a solution to quickly solve a issue at that moment. I consider that the Magento team knew most of the issue and felt relieved of the burden when the attention was moved to the new version which fails to impress in many chapters, more and more people complain about the complexity of the platform and the speed, but that is another discussion. I appreciate all of you who make efforts to make this project better.

ADDISON74:
I tested so far

1) Catalog > Manage Product > Edit > Image tab. Upload Files works as expected.

2) CMS > Manage Pages > Edit > Content tab > [Insert Image...] button. Upload Files works as expected.

Do I need to test other places in Backend?

ADDISON74:
We will have to periodically update this file or link it to Composer. That repository has a large number of reported issues. As if today you were discussing the issue of uploading large files. Here you go https://github.com/flowjs/flow.js/issues/291.

justinbeaty:
Eventually this library can go away. But I don’t think there’s any bugs in the new version that weren’t in our version of flow.js to begin with.

ADDISON74 (approved):

fballiano (dismissed):

fballiano:
should we add something in the README (in the v19-v20 differences section)?

Sdfendor (approved):

justinbeaty:
> should we add something in the README (in the v19-v20 differences section)?

I'm not sure what the criteria is for us to mention the change. Honestly I don't think anyone will even notice the library version bump because it seems backwards compatible.

sreichel:
If it is backwards compatible, we could merge this into 19.x.

ADDISON74:
The only issue I see is related to those two deleted JS files. If a developer has created an extension based on them then there will be a problem. 

Otherwise, it can be integrated just as well in version 19.

justinbeaty:
It's *maybe* not backwards compatible because of the removal of `fusty-flow.js` which was a compatibility library for < IE9 and we left MSIE support in 1.9.4.x.

But due to the way Magento team used `flow.js`, I'm not sure if `fusty-flow.js` was ever needed, and I don't have access to any install of IE 9 to test.

I can change it to 1.9.4.x though if others agree.

js/mage/adminhtml/uploader/instance.js
```diff
@@ -104,11 +104,16 @@
...
 
+            // Fix error where setting post_max_size or upload_max_filesize to 0
+            // causes the flow.js to make infinite chunks and crash the browser
+            if(config.uploaderConfig.chunkSize === 0) {
+                config.uploaderConfig.chunkSize = Number.POSITIVE_INFINITY;
+            }
+
             this.uploaderConfig = config.uploaderConfig;
             this.browseConfig = config.browseConfig;
             this.miscConfig =  config.miscConfig;
```
tobihille, line 115: I would suggest to add this line roughly here to enforce orderly updates:
`this.uploaderConfig.simultaneousUploads = 1;`

What happened in my case: the contend department uploaded a bunch of previously ordered images which ended up being shuffeled by the uploader because of the multiple requests.
I did not test the performance thoroughly but I noticed no slowdown of the upload.

If you feel not comfortable to add this line I will create my own PR, but I thougt to myself that a PR with one line is overkill 😅

tobihille (approved):
I was working on flow.js this week, good to see that you came to the same realization that fusty-flow.js is not used anymore.
I did not review the lib itself but your changes are looking good to me.

js/mage/adminhtml/uploader/instance.js
```diff
@@ -104,11 +104,16 @@
             this.elementsIds = config.elementIds;
             this.elements = this.getElements(this.elementsIds);
 
+            // Fix error where setting post_max_size or upload_max_filesize to 0
+            // causes the flow.js to make infinite chunks and crash the browser
+            if(config.uploaderConfig.chunkSize === 0) {
+                config.uploaderConfig.chunkSize = Number.POSITIVE_INFINITY;
+            }
+
             this.uploaderConfig = config.uploaderConfig;
```
tmewes, line 113:
```suggestion
            this.uploaderConfig = config.uploaderConfig;
            this.uploaderConfig.simultaneousUploads = 1;
```

tmewes (changes requested):

js/mage/adminhtml/uploader/instance.js
tmewes: I would highly recommend adding this line. Especially nowadays, due to better internet connections, it shouldn't be a problem to upload the files one after the other and get a reliable/expected order of the files.

justinbeaty:
I reduced the simultaneous uploads to 1 which I think is fine because it would take longer for a store admin to fix the messed up sorting of images than it would to wait a bit longer for uploads. Thanks @tobihille @tmewes 

However, I moved the config changes to `app/code/core/Mage/Uploader/Model/Config/Uploader.php`. Originally I didn't do this because `PHP_INT_MAX` is less than `Number.POSITIVE_INFINITY` but on a 32-bit system it's still 2.1 GB and on a 64-bit system it's practically infinite. This would allow someone to overwrite the Model class to change `simultaneousUploads` back if they wanted to.

tobihille (approved):

tmewes (approved):

justinbeaty:
The LGTM Useless conditional is because I've included the unminified flow.js in case anyone is wondering...

justinbeaty:
@fballiano @Sdfendor @ADDISON74 thoughts on changing this to 1.9.4.x branch? I guess it's okay but it is somewhat breaking.

ADDISON74:
I am also testing today's new changes and I will let you know an opinion. I hope you didn't limit uploading just one image, but I can still select as many as I want and upload them all at once. In the past I had a discussion on this topic with the MageWorx team who didn't want/didn't know how to modify the image uploader so that I can add multiple images to an option at once, instead we upload them one by one. Using that extension became a pain and gave up on it.

Regarding the change in 19, honestly speaking, I think we consume too much energy analyzing what would happen if we did one or the other. We don't really have feedback from those who still use it and when issues were reported, the references were mostly to version 20. This is also the version I use in production. If the PR will be accepted by majority in version 19 and we break something, I am sure we will find out. Good that we know more about the existence of this PR.

I must appreciate the contribution of the two today who had very good suggestions and which were adopted. OM needs as many devoted people as possible to get involved.

justinbeaty:
> I am also testing today's new changes and I will let you know an opinion. I hope you didn't limit uploading just one image,

It shouldn't be limited to just one upload. Instead it limits the _concurrency_ to one upload at a time. The issue reported was that several images are uploaded at once and whichever ones got there first would determine the image order.

ADDISON74:
Thank you for the explanation, everything is clear now.

ADDISON74:
Maybe we should analyze another issue related to the file loader. If we have 5 images with the names 1.jpg, 2.jpg, ..., 5.jpg and we upload them all at once, the images in the list are not positioned according to the name but according to another criterion. In order to solve the positioning issue I had to allocate time to establish a number for each one. The naming of the files should have solved this issue, an established order of the images. When we have a few it is no misfortune, but when there are 20 we lose 5 minutes scrolling the window content looking for each image.

tobihille:
@ADDISON74:
Your issue with the order of 1.jpg to 5.jpg is already adressed by setting `setSimultaneousUploads(1)`.
In general your OS will sort the images (at least in my testcases this was true) but the order is shuffeled again by uploading the files all at once.
The request that is handled first will add the first image and so on.
By Setting simultaneous uploads to 1 the issue is resolved because it will only update one image at a time so 1.jpg is uploaded, then (when 1.jpg is *finished*) 2.jpg and so on...

fballiano:
Do we have any idea if, without the 2 removed js, it would still work on older IE versions (also if doesn't support resumable downloads etc)?

fballiano:
so I guess it has to stay on v20 then

fballiano:
i can't test it right now cause I'm in the middle of another PR, will do ASAP

ADDISON74 (approved):

PR #2435, opened 2022-08-16
Removed unused boxover.js

fballiano:
I was checking js/lib/boxover.js file, it's an extremely old js library (no documentation exists anymore online, nor in the code itself) and, most of all, it seems totally unused:

Schermata 2022-08-16 alle 18 12 38

it's not even linked by any template or layout xml...


justinbeaty:
Here's the only documentation I can find: https://bytes.com/topic/javascript/answers/148855-boxover-easy-use-floating-captions

Let's get rid of it!

justinbeaty (approved):

ADDISON74:
Looking at the date of the posts in that link they are 17 years ago. Very cool.

The more you analyze the js and lib directories, the more you will uncover the garbage. So you won't be too surprised by what you will find. However, it is a project started somewhere in the mid-2000s.

fballiano:
that js is old enough to drive in some countries 😅

ADDISON74:
I took a look in a Magento archive for this file and I found the only reference here /var/package/Lib_Js_Mage-1.9.4.5.xml.

I doubt that someone has used this JS in any extension so that if we remove it we will create trouble. When /js/jscolor.js was deleted I came up with the proposal to put it back because it is used in MageWorx Advanced Product Options.

I approve and I hope no one calls for the file.

ADDISON74 (approved):

tmewes (approved):

elidrissidev (approved):

PR #1631, opened 2021-05-20
Used transliterator_transliterate to generate url_key

luigifab:
### Description

When a product or a category url key use special characters, they are substituted with **strstr**.
With this PR, **transliteration** is used (with [PHP INTL](https://www.php.net/transliterator_transliterate)).

Tested with the following locales: fr-FR, es-ES, de-DE, it-IT, pt-PT, nl-NL, pl-PL, cs-CZ, sk-SK, ru-RU, uk-UA.
OpenMage 20.x / PHP 7.4.6 and 8.0.x and 8.2.x

### Manual testing scenarios

- Set default locale to _French_.
- Set first store locale to _Russian_.
- Set second store locale to _French_.
- Create a product and category with the following name: _Массажный Аппарат для Похудения A_
- Set the product and category name on first store to:  _Массажный Аппарат для Похудения B_
- Set the product and category name on second store to:  _Массажный Аппарат для Похудения C_

Run:
```sql
DELETE FROM catalog_category_entity_varchar WHERE attribute_id IN (
    SELECT attribute_id FROM eav_attribute WHERE attribute_code LIKE 'url_key');
DELETE FROM catalog_product_entity_varchar WHERE attribute_id IN (
    SELECT attribute_id FROM eav_attribute WHERE attribute_code LIKE 'url_key');
```
Run:
```
rm -rf var/cache/ ; php shell/indexer.php reindexall ; rm -rf var/cache/
```

Check url keys generated:
- all store views: `massazhnyj-apparat-dlja-pohudenija-b`
- first store: `massazhnyj-apparat-dlja-pohudenija-b`
- second store: `massazhnyj-apparat-dlja-pohudenija-b`

That's wrong, and not good.
So apply the PR, clear existing url keys (first run), reindex and clear cache (second run).

Check url keys generated:
- all store views: `massaznyj-apparat-dla-pohudenia-a`
- first store: `massazhnyy-apparat-dlya-pokhudeniya-b`
- second store: `massaznyj-apparat-dla-pohudenia-c`

That's better.

Extra: for CMS / Pages, when an user create a new page without url key, an url key is generated.

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2437, opened 2022-08-16
Added php-extension "intl" as requirement, updated composer 

sreichel:
### Description (*)

- Magento installer checks for various php exentsion ... added `ext-intl` (and `ext-libxml`)
- added missing extensions from `install.xml` to `composer.json`

### Related Pull Requests

1. See OpenMage/magento-lts#2411
2. See OpenMage/magento-lts#1631


justinbeaty:
I like the simplification in the code and the removal of Net_IDNA, but I am not sure I agree with `ext-intl` being a required module. I think we should also have `symfony/polyfill-intl-idn` required and put `ext-intl` as a suggested module.

The reason is that if I check on my server, I don't have `php-intl` installed by default on Ubuntu 20.04. It's no problem to add for me, but it might catch people off guard.

fballiano:
I tend to agree with @justinbeaty, seems something for a major release, I'll think about it

sreichel:
> The reason is that if I check on my server, I don't have `php-intl` installed by default on Ubuntu 20.04. It's no problem to add for me, but it might catch people off guard.

I understand your concern, but then we should not even consider to raise minimum php version to 7.3 ... "because it could break something" (_does not fit server setup anymore_). Even on shared hosts (with Plesk/cPanel/...) you can enable/disable extension with some clicks.

fballiano:
Yes but it shouldnt just be a patch level release. And I’m confused now, for some things we’re extremely conservative and for some other we’re the opposite.

justinbeaty:
Well I agree we need to move forward and drop old PHP versions, but two things:

1) ext-intl was never required before
2) this is different than a PHP version that’s EOL.

But either way, there’s a simple fix which is to include the polyfill and is why I think it’s better to just include it for now.

sreichel:
> 1. ext-intl was never required before

PHP7 was never required before. ;)

> And I’m confused now, for some things we’re extremely conservative and for some other we’re the opposite.

It seems to be opinion based. I'd like to avoid BC breaking changes in code, but require a higher php-version or in this case another php-extionsion should not be a problem. I'd just (contact my hoster to) update requirements.

sreichel:
> Yes but it shouldnt just be a patch level release.

We have no semantic versioning. PHP5-support has been dropped with v19.4.9 (if i'm right).

justinbeaty:
> > 1. ext-intl was never required before
> 
> PHP7 was never required before. ;)

I still don’t think it’s a good comparison because when we supported PHP 5, PHP 7 was not required, it was optional. Only when PHP 5 was EOL did we require PHP 7.

It would be more like this situation if there was a php-intl2 extension that we now require because php-intl1 is deprecated.

In the end it doesn’t bother me either way as I follow this GitHub and am a composer user. I just think it’s a cheap fix to not make this breaking change by including the polyfill.

fballiano:
> We have no semantic versioning. PHP5-support has been dropped with v19.4.4 (if i'm right).

in issue 14 of the internal organization repository it seems to me that we're using semver, also https://github.com/OpenMage/magento-lts/blob/9b8a2cd1ce04b0c3d257b4ba47b604d86fcd3ef0/app/Mage.php#L176 links to semver.org website.

I didn't decide that, to me semver is too strict and everybody here knows i'm more for the disruptive changes, but since i've been blocked more than once because of BC, now I really am getting confused and I don't know what I should suggest anymore.

sreichel:
I got your point, but again ... if you want to use latest OM release, you should update your environment. Another example ... Magento2 has dropped MySQL-based search with 2.4.3 and everyone had to upgrade their servers (to install elastic-search) OR stay on previous release-line, that get only security fixes for another year(?).

We have no different release-lines, so we have to make a cut somewhere ....

(Just my point of view)

sreichel:
> seems to me that we're using semver

No. We never had major/minor/hotfix-releases ... even README says, we do NOT follow semver.

> now I really am getting confused and I don't know what I should suggest anymore.

Why? You`re doing great. :) We`re just about dicussing details :)

justinbeaty:
Agreed that we have no semblance of semver. I wish we would at least bump the minor version on larger releases (like magento team did) instead of BC breaks in patch versions. But anyway that’s another topic I guess.

I also get your point @sreichel, we’ve done plenty of breaking changes much larger than this. If you feel strongly about it, then I can agree with your changes. One less package in vendor anyway.

sreichel:
> I wish we would at least bump the minor version on larger releases

Agree. It should maybe combined with #2413 and release version could change to 1.9.5.x / 20.1.x ...

@Flyingmana?

fballiano:
> Agree. It should maybe combined with #2413 and release version could change to 19.5.x / 20.1.x ...

that would be great, I'd just do a release first since the last one had a couple of issues, then go on with this one :-)

tobihille (dismissed):

elidrissidev:
I do think such PRs should be left for new major versions, or at least v20, until the RFC for versioning is set in stone. Not against the change, just my personal take.

Flyingmana:
> > I wish we would at least bump the minor version on larger releases
> 
> Agree. It should maybe combined with #2413 and release version could change to 19.5.x / 20.1.x ...
> 
> @Flyingmana?

We dont have a process or rule for minor releases is all I can say to this

> I got your point, but again ... if you want to use latest OM release, you should update your environment. Another example ... Magento2 has dropped MySQL-based search with 2.4.3 and everyone had to upgrade their servers (to install elastic-search) OR stay on previous release-line, that get only security fixes for another year(?).
> 
> We have no different release-lines, so we have to make a cut somewhere ....
> 
> (Just my point of view)

we do have different release lines at the moment, v19 and v20 with v19 which should primarily only get bugfixes. and v20 for most of the other changes at the moment (although a v22 is justified by now)
Or did I misunderstood what you mean here?

sreichel:
@Flyingmana question was ro raise minor version or not, when we make php7.3/intl required.

fballiano (dismissed):

ADDISON74 (approved):

fballiano:
Let’s fix conflicts and merge

sreichel:
I'll wait for #2413

README.md
```diff
@@ -108,6 +108,10 @@ Most important changes will be listed here, all other changes since `19.4.0` can
 - search for "NULL" in backend grids [#1203](https://github.com/OpenMage/magento-lts/pull/1203)
 - removed modules `Mage_Compiler`, `Mage_GoogleBase`, `Mage_Xmlconnect`, `Phoenix_Moneybookers`
 
+### Between OpenMage 19.4.18 / 20.0.15 and 19.4.19 / 20.0.16
```
fballiano, line 111: this is now wrong, but i think we should go for 19.5.0 and 20.1.0 or something like that

sreichel: Can we fix this in another PR?, when we made decission to change minor-release?

fballiano: yes but i'd at least use a number that doesn't correspond to a previous release, just no to confuse people

ADDISON74 (approved):

README.md
sreichel: For me it does not matter if it is called 19.4.123 or 19.5.0. I'll changed README.md when decession is made :)

fballiano: I understand and I agree, but 20.0.16 is already released, we can't say that in 20.0.16 php7.3 is needed since it's simply not true.

sreichel: Ah ... got it.

sreichel: Can you please make a "github suggestion" to this line? (currently workin on ... phpstan ...)

fballiano, line 111:
```suggestion
### Between OpenMage 19.4.18 / 20.0.16 and 19.4.19 / 20.0.17
```

fballiano (approved):

ADDISON74 (approved):

PR #2438, opened 2022-08-17
DHL International: PHP8: Cast $dimension before calling round()

fballiano:
On PHP8.0, with DHL module enabled I get this error when getting to checkout:

```
<b>Fatal error</b>:  Uncaught TypeError: round(): Argument #1 ($num) must be of type int\|float, string given in /Users/fab/Projects/aaaaa/app/code/core/Mage/Usa/Model/Shipping/Carrier/Dhl/International.php:758
--
  | Stack trace:
  | #0 /Users/fab/Projects/aaaaa/app/code/core/Mage/Usa/Model/Shipping/Carrier/Dhl/International.php(758): round('', 3)

```

so I just added a cast to float


ADDISON74 (approved):

Flyingmana (changes requested):
we have another round a few lines above, we should convert it there, too

fballiano:
totally right, done it now

ADDISON74 (approved):

Flyingmana (approved):

Issue #2440, opened 2022-08-17
Product custom options always selected to "Yes" in PHP8.1


alexherbs:
The custom options are always selected to yes, even if "No" is selected. The reason is the since PHP 8.1 the JSON data.is_requiere is 0 instead of "0" and the code below isn't processed any more, because the condition is false.

https://github.com/OpenMage/magento-lts/blob/cc051f2f1807fef5762a5913e88115e40cbe46ac/app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml#L76

The line just needs to be changed to 

` if (data.is_require != undefined) {`

fballiano:
Hi and thanks for the report, I've created https://github.com/OpenMage/magento-lts/pull/2441 to fix the problem, it's a slightly different implementation (to be more consistent with the past codebase) but it should work.

kiatng:
I am testing with PHP 8.1.8, but I cannot replicate the issue:
![image](https://user-images.githubusercontent.com/1106470/185269779-077f4658-714b-4983-94d1-edf841894074.png)
The _Is Required_ dropdown is correctly set after saved. See screenshot above. What did I miss?

kiatng:
I take a look at the js:

https://github.com/OpenMage/magento-lts/blob/cc051f2f1807fef5762a5913e88115e40cbe46ac/app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml#L75-L80

My analysis:
line 76: It is checking if the property `is_require` exists; string `"0"` or `"1"` is truthy, but int 0 is falsy; in my case, it is a string, that's why it works for me. However, it is no guarantee. 
line 77: It iterates the options in the _Is Required_ dropdown
line 78: Here it sets the `selected` attribute, it is correctly set in my test shown in the screenshot above

OT, to step through the js in Firefox, I needed to add `//# sourceURL=option-panel` before `</script>`, see [stackoverflow](https://stackoverflow.com/a/65750872/488609).

But it is a good idea to have a proper check on line 76.

alexherbs:
@kiatng @fballiano  Hi, thanks for your fast testing and your reply. I've two enviroments, one with PHP7 and one with PHP8.1 and in PHP7 I get "0" and in PHP8.1 i get INT 0 for "is_requiere" which caused lots of problems with my products. Maybe some PHP settings causes different behaviour in JSON enconding ?!? But, its truly fair to check by an proper way. Thanks at all.

[...]

PR #2441, opened 2022-08-17
Fixed "Is Required" custom option attribute always set "Yes" in PHP8.1

fballiano:
Explained in https://github.com/OpenMage/magento-lts/issues/2440, in PHP8.1, when editing a product (from backend) with custom options, the "is required" was always shown as "yes".

After this PR the "is required" works as expected.

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml
@@ -73,7 +73,7 @@ var productOption = {
         }
 
         //set selected is_require
-        if (typeof data.is_require !== 'undefined') {
+        if (data.hasOwnProperty('is_require')) {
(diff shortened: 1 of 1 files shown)


app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml
```diff
@@ -73,7 +73,7 @@ var productOption = {
         }
 
         //set selected is_require
-        if (data.is_require) {
+        if (typeof data.is_require !== 'undefined') {
```
kiatng, line 76: How about this:
```suggestion
        if (data.hasOwnProperty('is_require')) {
```
Because we want to know if the property exists, if it does, the check `if (option.value==data.is_require) option.selected = true;` will work even if the value of `data.is_require` is `undefined`.

fballiano: ok done

kiatng (approved):

ADDISON74 (approved):

PR #2442, opened 2022-08-17
Removed unused jQuery 1.10.2, updated jQuery 1.12.1 to 1.12.4

fballiano:
If you search for "jquery-1.10.2" you'll see that there are no occurrencies, since version 1.12 is present and used in RWD theme

Schermata 2022-08-17 alle 21 54 18

Maybe we could target this to v20, in case we think some extension uses the old version of the javascript, what do you think?


ADDISON74:
I just wanted to propose this change. I would go ahead and bring the last version from branch 1, which is 1.12.4. so the JQuery in OpenMage will be up to date.

fballiano:
should we update it in this same PR?

ADDISON74:
> Maybe we could target this to v20, in case we think some extension uses the old version of the javascript, what do you think?

We will faced this question from now on, if any developer has used the files we want to delete or update, but we will never know if we don't step forward. I think it is more important to have all the code up to date than to leave old, outdated. Surely if there will be an issue in the future those who use OM will report it.

> should we update it in this same PR?

I have nothing against.

fballiano:
ok i guess we can leave this PR in the v19 branch.

and i guess it also makes sense to update the jquery 1.12 to 1.12.4 in this same PR, will to it asap (working on another issue at the moment)

ADDISON74:
Please upload both uncompressed and minified versions https://releases.jquery.com/jquery/. If possible the map file.

justinbeaty:
I say remove both versions, add `1.12.4`, and update `app/design/frontend/rwd/default/layout/page.xml`

Modules often just bundled their own version of jQuery, but in the case it was actually needed, it's an easy fix. Plus many old-school Magento installs probably just extract the releases on top of the old install anyway, and thus wouldn't have it removed.

fballiano:
i don't think we can't remove it easily no?

Schermata 2022-08-17 alle 22 42 56

ADDISON74:
I don't think a new version will ever be released in branch 1.

Leave the modification only in OM-20 and modify the XML file according to the new version.

justinbeaty:
> i don't think we can't remove it easily no?

I meant to add in 1.12.4 as a replacement, so that 1.12.4 is the only version left. Or were you thinking keep both 1.12.1 and 1.12.4?

fballiano:
> I don't think a new version will ever be released in branch 1. Leave the modification only in OM-20 and modify the XML file according to the new version.

yes I agree, but I thought we said to keep this PR in the v19 branch too no?

fballiano:
> I meant to add in 1.12.4 as a replacement, so that 1.12.4 is the only version left. Or were you thinking keep both 1.12.1 and 1.12.4?

we'll only have the last one ;-)

ADDISON74:
> yes I agree, but I thought we said to keep this PR in the v19 branch too no?

It can be used in OM-19 as well. Basically if someone uses the classic RWD theme in Frontend he will be just fine.

fballiano:
I think everything should be done now

ADDISON74 (approved):

justinbeaty (dismissed):

justinbeaty:
I have some potential feedback / ideas to test tomorrow if we can wait to merge this.

justinbeaty:
I was trying to see if it was feasible to install jQuery via composer. There are a few ways to do it as you can see here: https://github.com/justinbeaty/magento-lts/commit/ab05de9369abc9e640986c5410aa72b40ff0fbf5 (inspired by [this guide](https://lightning.acquia.com/blog/round-your-front-end-javascript-libraries-composer))

But it doesn't seem to work when installing openmage using `magento-core-composer-installer`. Maybe eventually we can modify `magento-hackathon/magento-composer-installer` to let us install JS libraries too.

So I'm back to approving this PR as it is. :)

tmewes (approved):

ADDISON74:
Even if I have approved it, I will check if errors appear in the browser console when I visit pages in the Frontend. If there are errors they are not from the new version anyway, but maybe we will fix them in the future. I hope it won't be necessary.

ADDISON74 (approved):

ADDISON74 (approved):

tmewes (approved):

kiatng (approved):

ADDISON74 (approved):

Issue #874, opened 2019-12-19
[RFC] Review mergeFiles logic in case of missing file


mmenozzi:
Hi guys,
given this line in the `Mage_Core_Helper_Data::mergeFiles()` method:

https://github.com/OpenMage/magento-lts/blob/143b03dd51bbdf92bd165942fa3c5a14a2d9b178/app/code/core/Mage/Core/Helper/Data.php#L752

In case one of the source file doesn't exists on filesystem the merge is performed repeatedly at every request.

I agree that this should never happen because if a file is not on filesystem it should not be used by the application.

But if by mistake a CSS is removed from filesystem and the developer forgot to remove the related layout XML rule, we have other CSS files merged again and again at every request.

What do you think? It's worth to find a better way to handle merging in case of missing files or simply we don't want to "support" this kind of mistake.

colinmollenhour:
Good observation. I think it would be appropriate to not re-merge the files every request as you stated, and an error should be logged when a file is missing.

mmenozzi:
I agree.

ADDISON74:
If this is an issue in OpenMage please create a PR for being tested. Thank you.

PR #2445, opened 2022-08-17
An Exception is thrown and logged if file is not found while merging

fballiano:
If css/js merging is enabled but one of the files doesn't exist on the file system, the merging was re-done at every request, as explained in https://github.com/OpenMage/magento-lts/issues/874

This should be avoided and an exception should be logged.

On line 837 of app/code/core/Mage/Core/Helper/Data.php all the exceptions are logged and I've added that message (with the comment above) to be consistent with the rest of the code.

### DIFF
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -774,7 +774,10 @@ XML;
                 } else {
                     $targetMtime = filemtime($targetFile);
                     foreach ($srcFiles as $file) {
-                        if (!file_exists($file) || @filemtime($file) > $targetMtime) {
+                        if (!file_exists($file)) {
+                            // no translation intentionally
+                            throw new Exception(sprintf('File %s not found.', dirname($file)));
+                        } elseif (@filemtime($file) > $targetMtime) {
                             $shouldMerge = true;
                             break;
                         }
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -776,7 +776,7 @@ XML;
                     foreach ($srcFiles as $file) {
                         if (!file_exists($file)) {
                             // no translation intentionally
-                            throw new Exception(sprintf('File %s not found.', dirname($file)));
+                            throw new Exception(sprintf('File %s not found.', $file));
                         } elseif (@filemtime($file) > $targetMtime) {
                             $shouldMerge = true;
                             break;


Sdfendor (approved):

ADDISON74 (approved):

tmewes (approved):

tobihille (dismissed):

Sdfendor (dismissed):

tmewes (dismissed):

ADDISON74 (approved):

sreichel:
What i have to to, to trigger this error?

I've eanbled css/js-merging, deleted a view files (that should be merged?!?), cleared cache ,,,

fballiano:
I don't know, I've just put that issue into code 😅

ADDISON74 (approved):

Sdfendor (dismissed):

app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -774,7 +774,10 @@ public function mergeFiles(
                 } else {
                     $targetMtime = filemtime($targetFile);
                     foreach ($srcFiles as $file) {
-                        if (!file_exists($file) || @filemtime($file) > $targetMtime) {
+                        if (!file_exists($file)) {
+                            // no translation intentionally
+                            throw new Exception(sprintf('File %s not found.', dirname($file)));
```
elidrissidev, line 779: If I delete the file `js/prototype/window.js`, in the log file I get the an incomplete path: `File /usr/local/var/www/magento/js/prototype not found` because `dirname` returns the parent dir's name, so there's no need to use it:
```suggestion
                            throw new Exception(sprintf('File %s not found.', $file));
```

elidrissidev (approved):

ADDISON74 (approved):

ADDISON74:
Lately there are more and more tests that fail for PHP Stan - .github/phpstan.neon. For this PR the error is the following

```
Error: Cannot call method addTitleToResult() on Mage_Core_Model_Resource_Db_Collection_Abstract|false.
 ------ -------------------------------------------------------- 
  Line   Catalog/Model/Resource/Product/Option/Collection.php    
 ------ -------------------------------------------------------- 
  152    Cannot call method addTitleToResult() on                
         Mage_Core_Model_Resource_Db_Collection_Abstract|false.  
 ------ --------------------------------------------------------
```

Is it a problem that we merged it without fixing this error?

elidrissidev:
> Is it a problem that we merged it without fixing this error?

Already fixed in another pending PR.

fballiano:
I think it started with https://github.com/OpenMage/magento-lts/pull/2562, https://github.com/OpenMage/magento-lts/pull/2572 should fixes it but I'm also trying to understand why it generated, which i still don't know

fballiano:
found it now :-)

PR #2461, opened 2022-08-21
Removed deprecation notice for Mage_Catalog_Model_Resource_Product_Attribute_Backend_Urlkey

fballiano:
Mage_Catalog_Model_Resource_Product_Attribute_Backend_Urlkey was deprecated in M1 1.7.0.2 but
- it was never removed or replaced
- there's no mention of why this was done in the release notes https://web.archive.org/web/20120825222617/http://www.magentocommerce.com/download/release_notes#Release%20Notes%20-%20Magento%201.7.0.2%20(Jul%205,%202012)
- it's still used in some places:
Schermata 2022-08-21 alle 15 32 02
and
Schermata 2022-08-21 alle 15 36 20

So I just think this class will never go away and, in that case, it shouldn't be marked as deprecated.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
@@ -25,6 +25,7 @@
  * @category    Mage
  * @package     Mage_Catalog
  * @author      Magento Core Team <core@magentocommerce.com>
+ * @deprecated
  */
 class Mage_Catalog_Model_Resource_Eav_Mysql4_Product_Attribute_Backend_Urlkey extends Mage_Catalog_Model_Resource_Product_Attribute_Backend_Urlkey
 {
--- a/app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
@@ -25,7 +25,6 @@
  * @category    Mage
  * @package     Mage_Catalog
  * @author      Magento Core Team <core@magentocommerce.com>
- * @deprecated
  */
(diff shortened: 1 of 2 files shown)


ADDISON74 (approved):

app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
```diff
@@ -25,7 +25,6 @@
  * @category    Mage
  * @package     Mage_Catalog
  * @author      Magento Core Team <core@magentocommerce.com>
- * @deprecated
```
sreichel, line 28: All MySql4-classes are deprecated.

sreichel (changes requested):

app/code/core/Mage/Catalog/Model/Resource/Eav/Mysql4/Product/Attribute/Backend/Urlkey.php
fballiano: should I just remove the change to this file then?

sreichel: Justed marked that line because i'm not sure to keep/remove the other.

Btw ... _there's no mention of why this was done_  ... maybe its not used in core anymore and is just there for BC. For MySql4 classes we could add a notice why they're deprecated.

fballiano: It is used in core (if I'm not mistaked), attribute_code url_key (for entity_type "product") is set to use this class. But also in other places (in the PR description).

fballiano: @sreichel you're right, since all mysql4 classes are deprecated they should stay that way.

Sdfendor (approved):

ADDISON74 (approved):

sreichel (approved):

PR #2467, opened 2022-08-21
Use composer PHP script for PHP_CS setup

justinbeaty:
### Description (*)

There were a few issues with the merge of #2400 
1. Hardcoded vendor dir (minor issue, [ref](https://github.com/OpenMage/magento-lts/pull/2400#discussion_r947325609))
2. Bash syntax caused an error on Windows ([ref](https://github.com/OpenMage/magento-lts/pull/2400#issuecomment-1221534161))
3. Global PHP version is called even if you do something like `/usr/local/opt/php@8.1/bin/php composer install` ([ref](https://github.com/OpenMage/magento-lts/pull/2400#issuecomment-1221546475)) 

I fixed this by making a PHP script that should be platform independent and fix all three issues.

### Related Pull Requests

#2400

1. Make sure it installs correctly on Linux as it did before -- @sreichel please check this, I don't know how to verify
2. Make sure `composer install` doesn't throw an error on Windows -- @ADDISON74 please check this
3. Make sure `composer install` works correctly when calling with a non-system php install -- @fballiano please check this

### Questions or comments
I added the `dev/scripts` directory and an autoload entry in `composer.json`. The new directory could be used for other composer scripts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,7 @@
   },
   "scripts": {
     "set-phpcs-paths": [
-      "[ $COMPOSER_DEV_MODE -eq 0 ] || vendor/bin/phpcs --config-set installed_paths vendor/phpcompatibility/php-compatibility,vendor/magento-ecg/coding-standard",
-      "[ $COMPOSER_DEV_MODE -eq 0 ] || vendor/bin/phpcs -i"
+      "OpenMage\\Scripts\\Devel::setupPhpCs"
     ],
     "post-install-cmd": [
       "@set-phpcs-paths"
@@ -73,6 +72,9 @@
   ],
   "autoload-dev": {
     "psr-4": { "OpenMage\\Tests\\Unit\\": "dev/tests/unit" }
+  },
+  "autoload": {
+    "psr-4": { "OpenMage\\Scripts\\": "dev/scripts" }
   },
   "extra": {
     "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
   },
   "scripts": {
     "set-phpcs-paths": [
-      "OpenMage\\Scripts\\Devel::setupPhpCs"
+      "OpenMage\\Scripts\\Composer\\SetupPhpCs::run"
     ],
     "post-install-cmd": [
       "@set-phpcs-paths"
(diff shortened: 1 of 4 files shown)


sreichel:
Nice! For me it works.

Only downside is that `exec()` is possibly disabled on shared hosts.

justinbeaty:
> Nice! For me it works.
> 
> Only downside is that `exec()` is possibly disabled on shared hosts.

I put a check for `exec` so we don't terminate the script if it's not available (for example if we did other stuff in that PHP file).

The only thing we could do to get around exec being disabled is if PHP_CS has a PHP API we could call. Not sure it's needed unless people will be running dev tools on shared hosting, though.

sreichel (dismissed):

sreichel:
Just an idea ...

- rename class `Devel` to `SetupPhpCs`
- change method name to `run(Event $event)` and maybe add an interface

```
interface ScriptInterface
{
    public static function run(Event $event);
}
```

justinbeaty:
@sreichel The interface is a good idea. I had named it `Devel` in case we wanted to do other things there without making tons of PHP files, but I'm okay with either way. You'll probably know more about what sort of tools might need to be initialized in the future.

It's getting late here so I'll make the changes tomorrow (or a maintainer is free to push).

sreichel:
> in case we wanted to do other things there without making tons of PHP files, but I'm okay with either way.

This is what i really like from M2 ... _Single Responsibility_ for each class.

justinbeaty:
+1 on all the changes!

sreichel (dismissed):

fballiano:
This scripts will for sure be used for some other tasks but I think now we should go with https://github.com/OpenMage/magento-lts/pull/2470 :-)

sreichel:
> This scripts will for sure be used for some other tasks

Definitly! :+1:

ADDISON74:
@justinbeaty's work must be appreciated. Since my finding yesterday and until today there have been constructive discussions. The approach in this PR is interesting and shows that the use of Composer is extremely powerful for this project.

justinbeaty:
#2470 is better :)

At least we have a framework set up in the case we need custom composer scripts that aren't handled by a plugin.

PR #2471, opened 2022-08-22
Removed memory_limit checks for better performance

fballiano:
All credits go to @woutersamaey since this PR was his idea. It was mostly implemented in https://github.com/OpenMage/magento-lts/pull/1749 but I couldn't rebase it to 1.9.4.x without breaking it so I opened this one instead, hopefully implementing everything that was discussed in the original thread.

Now the PR just removes checking for memory limit but keeps all the methods marking them as deprecated.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Image.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Image.php
@@ -232,6 +232,7 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     }
 
     /**
+     * @deprecated
      * @param null $file
      * @return bool
      */
@@ -241,6 +242,7 @@ class Mage_Catalog_Model_Product_Image extends Mage_Core_Model_Abstract
     }
 
     /**
+     * @deprecated
      * @return float|int|string
      */
     protected function _getMemoryLimit()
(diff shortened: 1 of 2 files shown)


ADDISON74 (approved):

Sdfendor (approved):

ADDISON74:
I am curious how a situation could be simulated in which the memory limit is exceeded in order to evaluate what happens then. I set in php.ini memory_limit=8M but the website relays without problems. I checked if the value is used and it is.

ADDISON74:
I tested  this PR. I set in php.ini memory_limit=8M and opcache.enable=0 (if opcache is enabled, the website works without problems.) 

This appears in the browser window based on the current PR:

```
Fatal error: Allowed memory size of 8388608 bytes exhausted (tried to allocate 655360 bytes) in /home/openmage/public_html/app/code/core/Mage/Persistent/Model/Observer.php on line 476
```

This appears in the browser window as it was until now:

```
Fatal error: Allowed memory size of 8388608 bytes exhausted (tried to allocate 655360 bytes) in /home/openmage/public_html/app/code/core/Mage/Persistent/Model/Observer.php on line 476
```

Practically with check and without check the error appears anyway, the advantage is that the extra evaluation is not done every time, I don't know how much is gained by omitting it, but it is OK to be merged.

kiatng (approved):

PR #2473, opened 2022-08-22
Upgraded ExtJS to 1.1.1

justinbeaty:
### Description (*)

Thanks to @fballiano for pointing this out on discord. `ext-tree.js` is version 1.0.1 in our source. I've updated to 1.1.1 which was the last 1.x release (apparently 2.x was BC-breaking). I also updated the bundled resources (CSS and images)

It turns out `ext-tree.js` is not just the tree component from ExtJS, but a custom build of ExtJS. I had to recreate this from the `ext-1.1.1.zip` archive. 

Source is available here: https://web.archive.org/web/20130401000000*/http://dev.sencha.com/deploy/ext-1.1.1.zip

1. Test the following pages:
2. https://store.example.com/index.php/admin/permissions_role/
3. https://store.example.com/index.php/admin/catalog_product_set/edit/id/4/
4. https://store.example.com/index.php/admin/catalog_category/

### Questions or comments
There are the files `fix-defer-before.js` and `fix-defer.js`. The former looks like an MSIE7 fix. The latter I'm not sure, but possibly some polyfill or browser fix. It's probably worth looking into what those both do.

There is also the `ext/css` dir which contains files that look like they're from ExtJS 2.0. I'm not sure what they're doing there and I don't see usage of them.


justinbeaty:
In an attempt to reduce the size of the bundle, here's an updated build script. I am basically seeing what can be removed from the files array. 224KB -> 168KB.

```console
#!/bin/bash

files=(
  build/core/Ext-min.js
  build/adapter/prototype-bridge-min.js
  build/core/DomHelper-min.js
  # build/core/Template-min.js
  # build/core/DomQuery-min.js
  build/util/Observable-min.js
  build/core/EventManager-min.js
  build/core/Element-min.js
  # build/core/Fx-min.js
  # build/core/CompositeElement-min.js
  build/core/UpdateManager-min.js
  # build/util/DelayedTask-min.js
  # build/util/TaskMgr-min.js
  build/util/MixedCollection-min.js
  # build/util/JSON-min.js
  # build/util/Format-min.js
  # build/util/CSS-min.js
  # build/util/ClickRepeater-min.js
  # build/util/KeyNav-min.js
  # build/util/KeyMap-min.js
  # build/util/TextMetrics-min.js
  build/dd/DDCore-min.js
  build/dd/ScrollManager-min.js
  build/dd/Registry-min.js
  build/dd/StatusProxy-min.js
  build/dd/DragSource-min.js
  build/dd/DropTarget-min.js
  build/dd/DragZone-min.js
  build/dd/DropZone-min.js
  # build/data/SortTypes-min.js
  # build/data/Record-min.js
  # build/data/Store-min.js
  # build/data/Connection-min.js
  # build/data/DataField-min.js
  # build/data/DataReader-min.js
  # build/data/DataProxy-min.js
  # build/data/MemoryProxy-min.js
  # build/data/HttpProxy-min.js
  # build/data/ScriptTagProxy-min.js
  # build/data/JsonReader-min.js
  # build/data/ArrayReader-min.js
  build/data/Tree-min.js
  build/widgets/Component-min.js
  build/widgets/BoxComponent-min.js
  build/widgets/Layer-min.js
  build/widgets/Shadow-min.js
  build/widgets/Editor-min.js
  build/widgets/tree/TreePanel-min.js
  build/widgets/tree/TreeSelectionModel-min.js
  build/widgets/tree/TreeNode-min.js
  build/widgets/tree/AsyncTreeNode-min.js
  build/widgets/tree/TreeNodeUI-min.js
  build/widgets/tree/TreeLoader-min.js
  build/widgets/tree/TreeFilter-min.js
  build/widgets/tree/TreeSorter-min.js
  build/widgets/tree/TreeDropZone-min.js
  build/widgets/tree/TreeDragZone-min.js
  build/widgets/tree/TreeEditor-min.js
  build/widgets/form/Field-min.js
  build/widgets/form/TextField-min.js
  )


# Build License Header
printf '/**\n' > ext-tree.js
printf ' * Ext JS Library 1.1.1\n' >> ext-tree.js
tail -n +2 resources/license.txt | sed -e 's/^/ \* /' >> ext-tree.js
printf '\n */\n' >> ext-tree.js

# Concat files
for file in "${files[@]}"; do
    tail -n +8 $file >> ext-tree.js
    printf '\n\n\n' >> ext-tree.js
done

dos2unix ext-tree.js
```

ADDISON74:
In the repository can be the normal and the reduced size. I don't agree with only one reduced versions. Such a format is supposed to be final, for production, free of issues and we all know that this is impossible. To check the code we have to make another step but when I propose a PR in a minified version it will be difficult for others to follow. Classic coding is preferred for the project (my opinion).

ADDISON74:
I did not find any references in the code related to files:

- **fix-defer-before.js** (this one has the code commented and with references to IE7 and IE8) **=> it could be deleted.**
- **fix-defer.js** (this one seems to defer something) **=> I don't know what is the impact if it is deleted.**

justinbeaty:
I see `fix-defer-before.js` included [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/design/adminhtml/default/default/layout/main.xml#L53)

```
$ grep -R 'fix-defer-before.js' ~/magento-public/htdocs/app/
/home/www-data/magento-public/htdocs/app/design/adminhtml/default/default/layout/main.xml:                <action method="addItem"><type>js</type><name>extjs/fix-defer-before.js</name><params/><if/><condition>can_load_ext_js</condition></action>
/home/www-data/magento-public/htdocs/app/design/adminhtml/default/default/layout/oauth.xml:            <action method="removeItem"><type>js</type><name>extjs/fix-defer-before.js</name></action>
```

I'm not sure yet either the reason for `fix-defer.js`, but I'll get to the bottom of it.

---

> In the repository can be the normal and the reduced size. I don't agree with only one reduced versions. Such a format is supposed to be final, for production, free of issues and we all know that this is impossible. To check the code we have to make another step but when I propose a PR in a minified version it will be difficult for others to follow. Classic coding is preferred for the project (my opinion).

Normally I agree. In this case I don't want to include the entire ExtJS library since we only use a very small subset, but I should be able to include a non-minified source of each file I'm concating together.

justinbeaty:
I removed a lot of files that I don't think are needed at all. Maybe too much, but I left only files that are used by Mage as far as I can tell.

There's a slight chance that some 3rd party extension used some of these JS components, but I really doubt it. Because if they did they would have been limited to the ones Magento team chose to bundle into `ext-tree.js`. It's more likely a module developer would have bundled their own ExtJS files into their module. Plus ExtJs 1.x is from 2007, and was replaced by ExtJS 2.x around the same time too.

In any case, I say to get rid of all this old JS. One day when we want to modernize the JS stack, we will be thankful that there is less crap to go through.

fballiano:
do you think this is ready for test?

justinbeaty:
@fballiano yes I think it’s ready. I forgot it was marked draft, I’ve been out of town last few days.

fballiano (dismissed):
tested both categories and products mask work fine.

elidrissidev (approved):
All good after testing the mentioned pages

ADDISON74:
This PR seems fine to me and it can be approved, but first the conflict that has arisen must be resolved.

fballiano:
fixed conflict, could you re-test just to be sure?

ADDISON74:
I will test this PR again.

justinbeaty:
> fixed conflict, could you re-test just to be sure?

Hi Sorry, I missed this comment earlier. I will check it now, but do we still wan to target 1.9.4.x with this PR? Not sure if we've changed thoughts on this.

fballiano:
well, we should rebase on v20, but if we do:
Screenshot 2023-01-14 alle 15 32 31

justinbeaty:
I will rebase

justinbeaty:
Rebased & updated base branch. `fix-defer-before.js` was already removed in #1073 so I dropped that commit. Testing now.

justinbeaty:
I had to add in the `data/Connections.js` file, else I discovered an error on the Manage Categories page (expanding a tree item tried to load the subcategories via AJAX, which failed.)

Otherwise, I think it is fine. I get no JS errors on any of the three pages, and I tried moving things around, saving, making sure they were saved properly, etc.

I look forward to the day this 2007 era library can be replaced.

sreichel (approved):

sreichel:
> changed the base branch from

Can you revert this :sunglasses:

justinbeaty:
> > changed the base branch from
> 
> Can you revert this? 😎

Okay, we’ve updated jQuery, jscolor, and tinymce in 1.9.4.x already, so I guess this is okay in 1.9.4.x too. However I will have to update the branch tomorrow morning (12-ish hours from now).

I still think 1.9.4.x should be security fix only, but no such RFC enforcing that has passed and I know you’ve been trying to minimize differences between the branches.

justinbeaty:
> > changed the base branch from
> 
> Can you revert this? 😎

Switched back to 1.9.4.x and tested.

luigifab:
Is this PR ready for crash test?

fballiano:
I think that, for how things are going, we should re-retarget this to v20

fballiano:
@kyrena are you refering to the pixel difference in the space between the icon and the checkbox?

fballiano:
@justinbeaty how did you generate the ytheme-magento file? because it lacks a few lines (about the tree checkbox it lacks `input.l-tcb`), we could decide not to care about the minor differences but...

justinbeaty:
@fballiano I don't recall off the top of my head, but I'll take a look a bit later today or tomorrow morning at the latest. I'll also rebase to v20 while I'm at it.

justinbeaty:
@fballiano I believe I had just manually removed all things not related to ext-tree from the CSS file. The reason I missed that class is that it's not from `extjs`, but from `ext-tree-checkbox.js`, which is an plugin for extjs that came from *somewhere* (there's no copyright or license info in that file)...

I've re-added that class & rebased to 20.0

fballiano:
> that came from _somewhere_ (there's no copyright or license info in that file)...

😂😂

fballiano:
@justinbeaty super, seems to be perfect now. I see a few missing classes (`#loading` for example) but who knows if they're really necessary...

justinbeaty:
@fballiano yeah, I noticed those `#loading` styles too, and I don’t think they were used either. I’ll check it out momentarily, but if they were used, I can probably refactor it to use something more specific than just `#loading`.

fballiano:
agree, let's merge and see if something comes out we'll fix it later :-)

fballiano (approved):

justinbeaty:
Yes, I don't see anything creating `<div id="loading">` in extjs. Instead it creates:
```
indicatorText : '<div class="loading-indicator">Loading...</div>'
```

So I think we are all good.

ADDISON74:
With all the recent changes we should take a look into the browser console and do an audit. There is a small chance that there are still issues related to ZF1-Future and what was removed that we aren't aware.

justinbeaty:
@ADDISON74 I checked all three pages mentioned in the initial comment with the browser console opened. No errors.

elidrissidev (approved):

PR #199, opened 2017-03-02
don't save index events for indexers in manual mode

bob2021:
This is part of a group of PRs containing the changes discussed in issue #152

If an indexer is in manual mode, there is no need to save an index event as it will never be processed - manual mode indexes can only be updated using a full reindex.

Index event data can grow very large (megabytes) for mass_action events.  It is not cleaned properly after processing and new data is often appended without deduplication.

This PR prevents index events from registering with an indexer if it is in manual mode.  It also ensures that index data is cleaned during a partial 'full' reindex.

PR #2474, opened 2022-08-23
Added cron to remove old index events for manual indexers

tobihille:
### Description (*)

As discussed in https://github.com/OpenMage/magento-lts/pull/199 I created a new PR which tackles the problem from another direction: Instead of completely omitting to write the index event to the database I created a cron that will delete these events if they are not processed.
This way we maintain BC and avoid filling the DB with garbage that might not be needed.

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/199

1. Set index mode to manual in admin
2. Save a product
3. Execute select count(*) from index_event
4. Wait 24 hours
5. Run php cron.php (I executed the job from Aoe_Scheduler but the result should be the same)
6. Execute select count(*) from index_event

### Questions or comments
I would gladly discuss the introduction of `Mage_Index_Model_Observer::OLD_INDEX_EVENT_THRESHOLD_SECONDS`, in my opinion this is a quick-and-dirty way to start this PR but I do not expect it to be in the final merge.
I would prefer a config but Mage_Index does not have a system.xml - so where would be good place for it?
If we add a config: what should be the default value?
Is one day a sufficient threshold for e.g. the Hackathon_AsyncIndex (on my workplace we index every day even if we have over 100.000 products and thousand categories)?

### DIFF
--- a/app/code/core/Mage/Index/Model/Indexer.php
+++ b/app/code/core/Mage/Index/Model/Indexer.php
@@ -282,10 +282,10 @@ class Mage_Index_Model_Indexer
      * @param   Varien_Object $entity
      * @param   string $entityType
      * @param   string $eventType
-     * @throws Exception
      * @return  Mage_Index_Model_Indexer
+     * @throws Exception|Throwable
      */
-    public function processEntityAction(Varien_Object $entity, $entityType, $eventType)
+    public function processEntityAction(Varien_Object $entity, $entityType, $eventType): Mage_Index_Model_Indexer
     {
         $event = $this->logEvent($entity, $entityType, $eventType, false);
         /**
--- a/app/code/core/Mage/Index/Model/Observer.php
+++ b/app/code/core/Mage/Index/Model/Observer.php
@@ -19,6 +19,9 @@
  */
 class Mage_Index_Model_Observer
 {
+    public const OLD_INDEX_EVENT_THRESHOLD_SECONDS = 24 * 60 * 60;
+    public const OLD_INDEX_EVENT_DELETE_COUNT = 1000;
+
     /**
      * Indexer model
      *
@@ -35,6 +38,7 @@ class Mage_Index_Model_Observer
      * Store after commit observer. Process store related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processStoreSave(Varien_Event_Observer $observer)
     {
@@ -50,6 +54,7 @@ class Mage_Index_Model_Observer
      * Store group after commit observer. Process store group related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processStoreGroupSave(Varien_Event_Observer $observer)
     {
@@ -65,6 +70,7 @@ class Mage_Index_Model_Observer
      * Website save after commit observer. Process website related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processWebsiteSave(Varien_Event_Observer $observer)
     {
@@ -80,6 +86,7 @@ class Mage_Index_Model_Observer
      * Store after commit observer. Process store related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processStoreDelete(Varien_Event_Observer $observer)
     {
@@ -95,6 +102,7 @@ class Mage_Index_Model_Observer
      * Store group after commit observer. Process store group related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processStoreGroupDelete(Varien_Event_Observer $observer)
     {
@@ -110,6 +118,7 @@ class Mage_Index_Model_Observer
      * Website save after commit observer. Process website related indexes
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processWebsiteDelete(Varien_Event_Observer $observer)
     {
@@ -125,6 +134,7 @@ class Mage_Index_Model_Observer
      * Config data after commit observer.
      *
      * @param Varien_Event_Observer $observer
+     * @throws Throwable
      */
     public function processConfigDataSave(Varien_Event_Observer $observer)
     {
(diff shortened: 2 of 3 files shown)


app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -136,4 +139,35 @@ public function processConfigDataSave(Varien_Event_Observer $observer)
...
+
+        /** @var Mage_Index_Model_Process $process */
+        foreach ($manualIndexProcessCollection as $process) {
+            $unprocessedEventsCollection = $process
+                ->getUnprocessedEventsCollection()
+                ->addFieldToFilter('created_at', ['lt' => $oldEventsThreshold])
+                ->load();
+
+            foreach ($unprocessedEventsCollection as $unprocessedEvent) {
+                $unprocessedEvent->delete();
```
fballiano, line 169: I was thinking, in case there are maaaaany lines to delete could this be a bit of a problem? wouldn't be better to just do a "DELETE FROM ... WHERE ..." and maybe also set a LIMIT (I don't know, maybe 10000 records) just to be super safe

Sdfendor:
Is the introduction of a cronjob a reason to bump the version number of the module 🤔?

fballiano:
> Is the introduction of a cronjob a reason to bump the version number of the module 🤔?

I'd bump only if there's a new install script to modify the database

Sdfendor:
Fair point after all its one of the main reason to have versions of OpenMage modules to begin with.

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -136,4 +139,35 @@ public function processConfigDataSave(Varien_Event_Observer $observer)
...
+        $manualIndexProcessCollection = Mage::getSingleton('index/indexer')
+            ->getProcessesCollection()
+            ->addFieldToFilter('mode', Mage_Index_Model_Process::MODE_MANUAL);
+
+        $now = new DateTime();
+        /** @noinspection PhpUnhandledExceptionInspection */
+        $dateInterval = new DateInterval('PT' . self::OLD_INDEX_EVENT_THRESHOLD_SECONDS . 'S');
+        $oldEventsThreshold = $now
+            ->sub($dateInterval)
+            ->format('Y-m-d H:i:s');
```
Sdfendor, line 159:
```suggestion
            ->format(Varien_Db_Adapter_Pdo_Mysql::TIMESTAMP_FORMAT);
```
I didn't use `Varien_Date::DATETIME_PHP_FORMAT` which also contains this format, because you are using it clearly in a DB context here. I would like to suggest to avoid string literals where constants do exist. I found 18 usages of this literal already (can be addressed in a future PR), let's avoid adding another one 🙂.

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -20,6 +20,9 @@
 
 class Mage_Index_Model_Observer
 {
+
```
Sdfendor, line 23: You could remove the empty line after class start.

Sdfendor (changes requested):

app/code/core/Mage/Index/Model/Observer.php
tobihille: I didn't want to 'berserk' through the whole file to make the PR more concise. But it would be a pleasure to update all comments and so on, I will take this as a hint that this is indeed wanted and start right now ;-)

Sdfendor: You introduced the empty line though in this special case. Or at least Github displays it as one of your additions 🤔.

app/code/core/Mage/Index/Model/Indexer.php
```diff
@@ -284,10 +284,10 @@ public function logEvent(Varien_Object $entity, $entityType, $eventType, $doSave
      * @param   Varien_Object $entity
      * @param   string $entityType
      * @param   string $eventType
-     * @throws Exception
      * @return  Mage_Index_Model_Indexer
+     *@throws Exception|Throwable
```
Sdfendor, line 288: Whitespace missing here and personally I'm no fan of | in the throws statements in phpdocblock. But it's a possibility after all.

app/code/core/Mage/Index/Model/Observer.php
tobihille: Updated.

app/code/core/Mage/Index/Model/Indexer.php
tobihille: Fixed whitespace but left open "Exception|Throwable" because I'm unsure if this is consensus.

Sdfendor: Just an expression of personal opinion. Multiple `throws` annotations give the possibility for a description for each of them after the exception type. If there is no describing text or the same for in this case both `Exception` and `Throwable` then it wouldn't make a difference whether multiple `throws` annotations or | is used. As a bonus I've seen the variant with multiple annotations in multiple places in the codebase ;-).

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -136,4 +144,53 @@ public function processConfigDataSave(Varien_Event_Observer $observer)
...
+        foreach ($manualIndexProcessCollection as $process) {
+            $unprocessedEventsCollection = $process
+                ->getUnprocessedEventsCollection()
+                ->addFieldToFilter('created_at', ['lt' => $oldEventsThreshold])
+                ->load();
+
+            $eventList = [];
+            /** @var Mage_Index_Model_Event $unprocessedEvent */
+            foreach ($unprocessedEventsCollection as $unprocessedEvent) {
+                $eventList[] = $unprocessedEvent->getId();
```
fballiano, line 182: can we limit it to something like 1000 events? otherwise the "delete where in(...)" could become too long and throw an error

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -19,13 +19,11 @@
...
 
-/**
- * @category   Mage
- * @package    Mage_Index
- * @author     Magento Core Team <core@magentocommerce.com>
- */
 class Mage_Index_Model_Observer
 {
+    const OLD_INDEX_EVENT_THRESHOLD_SECONDS = 24 * 60 * 60;
+    const OLD_INDEX_EVENT_DELETE_COUNT = 1000;
```
fballiano, line 25:
```suggestion
    public const OLD_INDEX_EVENT_THRESHOLD_SECONDS = 24 * 60 * 60;
    public const OLD_INDEX_EVENT_DELETE_COUNT = 1000;
```

app/code/core/Mage/Index/etc/config.xml
```diff
@@ -26,6 +26,18 @@
             <version>1.6.0.0</version>
         </Mage_Index>
     </modules>
+    <crontab>
+        <jobs>
+            <index_clean_events>
+                <schedule>
+                    <cron_expr>30 * * * *</cron_expr>
```
fballiano, line 33:
```suggestion
                    <cron_expr>30 */4 * * *</cron_expr>
```
I'd make something like this if that's ok with you, to run every 4 hours, cause every 30 minutes it seems a bit too much no?

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -142,4 +152,58 @@ public function processConfigDataSave(Varien_Event_Observer $observer)
...
+
+        $now = new DateTime();
+        /** @noinspection PhpUnhandledExceptionInspection */
+        $dateInterval = new DateInterval('PT' . self::OLD_INDEX_EVENT_THRESHOLD_SECONDS . 'S');
+        $oldEventsThreshold = $now
+            ->sub($dateInterval)
+            ->format(Varien_Db_Adapter_Pdo_Mysql::TIMESTAMP_FORMAT);
+
+        $coreResource = Mage::getSingleton('core/resource');
+
```
fballiano, line 176:
```suggestion
        $coreResource = Mage::getSingleton('core/resource');
```

app/code/core/Mage/Index/Model/Observer.php
```diff
@@ -142,4 +152,58 @@ public function processConfigDataSave(Varien_Event_Observer $observer)
...
+        $writeConnection = $coreResource->getConnection('core_write');
```
fballiano, line 177:
```suggestion
        $coreResource = Mage::getSingleton('core/resource');
        $writeConnection = $coreResource->getConnection('core_write');
```

fballiano (approved):

empiricompany (approved):

Sdfendor (approved):

fballiano:
we could merge because it's 1 green + 2 gray but I notice now that it's the wrong branch, can't merge on 1.9.4.x, we've to rebase on main

fballiano:
I'm merging because it was already approved (1green + 2gray checks) but I had to rebase (checked and the diff is exactly the same)

PR #2480, opened 2022-08-23
Fixed default sort order for admin catalog search terms, closes #2477

sreichel:
### Description (*)

Fixed default sort order ("name" is not a valid column).

1. Closes OpenMage/magento-lts#2477

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Search/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Search/Grid.php
@@ -36,7 +36,7 @@ class Mage_Adminhtml_Block_Catalog_Search_Grid extends Mage_Adminhtml_Block_Widg
         parent::__construct();
         $this->setId('catalog_search_grid');
         $this->setDefaultSort('query_id');
-        $this->setDefaultDir('desc');
+        $this->setDefaultDir('asc');
         $this->setSaveParametersInSession(true);
     }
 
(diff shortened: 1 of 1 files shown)


ADDISON74:
I haven't tested it yet. Go to the last page select 2 rows then sort by any column. Can you get to the 2 selected rows again? Can you get the initial listing from lowest to highest ID?

sreichel:
In any case you can get seleceted columns with setting first row (checkboxes) to yes. Changing/fixing deault sort order should solve you problem ... if i got u right.

ADDISON74:
At first glance, your implementation does not solve the issue. 

I remain of the opinion that the ID column should be visible in the grid. Here is my argument. You propose to change the default sorting in the grid from name to search_query. This change no longer displays the searches in their order, but alphabetically. I need to keep the current behavior to know which searches have been added since the last check and this requires sorting by ID. The column exists in code but has been commented.

sreichel:
> I need to keep the current behavior to know which searches have been added since the last check

This is an argument ;)

ADDISON74:
Many thanks. I will test and let you know.

ADDISON74:
I tested the last change. It's almost perfect. 

1. This is how the grid looks when I load the page. Sorted by descending ID.

ss1

2. I selected two rows.

ss2

3. I sorted by the Results column.

ss3

4. I clicked twice on the ID column and returned to the selection.

ss4

Why did I say it's almost perfect? My opinion is to initially sort the ID column ascending and not descending. My argument is related to the two clicks I had to make on the ID column to return to the selection. Please make the following test, change in the Grid.php the default sort order to asc then go to the last page (exactly what we do now in OpenMage to see the last searches). Select a few rows, sort by another column and click on ID. You just returned to the section with one click, not two.

ADDISON74 (approved):

fballiano (approved):

Issue #685, opened 2019-05-31
Image label per store view does not work as expected


azambon:
Steps to reproduce:

Create at least one store view other than the default one
Create a simple product
From the default store view add an image to the product with the label 'AAA' and set it as the base image, small image and thumbnail
Switch to the non-default store view and make sure that the "base image", "small image" and "thumbnail" radios are set to "use default value". Save the product.
Go back to the default store view and change the image label to 'BBB'. Save the product.

	Expected result: in the frontend of the non-default store view the image should have the BBB label
	Actual result: in the frontend of the non-default store view the image has the AAA label

Back in the admin, go to the non-default store view. Set the image label to 'CCC' and save the product.

	Expected result: in the frontend of the non-default store view the image should have the CCC label
	Actual result: in the frontend of the non-default store view the image has the AAA label

In the admin add another image to the product with the label DDD in the default store view. Mark it as the default base image, small image and thumbnail and set its label to DDD

	Expected result: in the frontend of the non-default store view the image should be the new one with the DDD label
	Actual result: in the frontend of the non-default store view the image is the new one, but with the AAA label

Notice: in the admin panel product image tab all the data seem to be displayed correctly. That's because in the catalog_product_entity_media_gallery and catalog_product_entity_media_gallery_value tables the values are correct, and the admin panel uses those tables. It's the image_label, small_image_label and thumbnail_label attributes that are not epdated correctly.

From our investigations, it seems like the issue lies in the Mage_Catalog_Model_Product_Attribute_Backend_Media::beforeSave() method. When saving a product in a non-default store view where the images are set to "use default", the line

$attrData = $object->getData($mediaAttrCode);

returns false. The following in_array() checks fail and so the values for the XXX_label attributes that were loaded get duplicated in the non-default store view.

fballiano:
@azambon could you please check PR https://github.com/OpenMage/magento-lts/pull/2481 ? we should approve it and get it merged

azambon:
> @azambon could you please check PR #2481 ? we should approve it and get it merged

Hello,
thank you @fballiano :+1: 

It will be a few days before I can check this, but I will do so as soon as possible.

sreichel:
@all-contributors add @azambon bug

PR #2481, opened 2022-08-23
Fixed faulty media gallery label/description management for mutistore projects

fballiano:
We're talking about the "images tab" of the product management page:
Schermata 2022-08-23 alle 22 14 42

The complete problem explanation and testing scenario is explained in issue https://github.com/OpenMage/magento-lts/issues/685

It wasn't easy to fix, I worked on it quite some hours. Hope I didn't miss anything.

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/685


ADDISON74:
There are a lot of check failures in the "Files changed' section (at the end). Could you please investigate if they are real warning?

fballiano:
I'll check it out asap!

fballiano:
actually those notice are in the "Unchanged files with check annotations", those files weren't changed in this PR so I wouldn't do anything about those notices.

fballiano:
@azambon I'd love your comment on this

azambon:
Sory for the massive delay, I completely forgot about this!

I did try to reproduce the same scenario that I described in the issue, but I don't think that this PR actually solves it.
As far as I could see, this PR adds the possibility of marking the whole "label" and "position" columns as "use default" in non-default store views.
Even ignoring the fact that someone may want to specify "use default" only for the labels and positions of some images, the bug that I reported is about both a situation where the labels must implement the "use default" behavior, but also about a case when the labels must be different (when I said to set the label "CCC" in the non-default store view). And in both cases I still see the wrong values in the frontend (yes, I disabled all caching, just to be sure).

So, unless I'm missing something, this does not seem to fix issue #685 . At least not yet.

As I see it, there are a couple of things left to solve:
* fix the values that are saved for the `image_label`, `small_image_label` and `thumbnail_label` attributes for non-default store views. That should fix the frontend and all other places that use those attributes;
* (optionally) move the "use default" checkboxes to be per-image for the "label" and "position" columns, instead of being global as they were implemented in this PR.

fballiano:
Hi @azambon and thanks for the test,
- I'm pretty sure the values are saved correctly on the DB, if you switch storeview in the backend you should see the values you entered per every store view
- I think I didn't check the frontend at all, so that could need another fix
- I've to say I wouldn't put the checkbox "per image", that would be much harder, maybe it could be done in a second PR but to me at least it should work with "all images"

fballiano:
@azambon I tested the frotend and the image labels seem to be correct to me, both in default and the single store view, the labels changes and are the same I see in the backend. This is tested on RWD theme.

azambon:
I performed the scenario that I described in the issue once more, taking some screenshots.

As you can see, when I uncheck the new "use default" checkbox that you added in order to set the label CCC for the non-default store view, the data is saved correctly in the `catalog_product_entity_media_gallery_value` table, but the attributes in `catalog_product_entity_varchar` are not updated correctly.
And, as you can see, in the frontend of the non-default store view I still see the wrong value.

This is with a fresh installation of OpenMage with the default theme, no customizations at all and all caches disabled.
If you use a frontend theme that does not use the `image_label`, `small_image_label` and `thumbnail_label` product attributes and uses the media gallery directly, then yes, the frontend would work fine too. But those attributes would still contain a wrong value, and anything that used them would get the wrong data.

Schermata da 2022-10-10 12-18-43
Schermata da 2022-10-10 12-18-49
Schermata da 2022-10-10 12-21-12

azambon:
All caches are disabled and I did reindex (even though reindexing should not affect the values in the catalog_product_entity_* tables, only in the catalog_product_flat_* tables, if flat tables are enabled).

fballiano:
I've flat disabled, just in case

fballiano:
I'll post a few screenshots to check that i'm not drunk :-D

backend for default:
Schermata 2022-10-10 alle 12 05 38

backend for non default
Schermata 2022-10-10 alle 12 05 48

frontend for default
Schermata 2022-10-10 alle 12 06 17

and frontend for the non default storeview
Schermata 2022-10-10 alle 12 06 23

azambon:
Hmm, wait a second. In your screenshot that you labeled "backend for non default" I don't see the new checkboxes that this PR should have added. I do have those checkboxes:
Schermata da 2022-10-17 11-19-24

Question: are we checking the same code revision? I mean, I'm testing at commit 495e44f92f3c08c9da00d78c079bebeffb80de39
Schermata da 2022-10-17 11-23-47

Can you confirm that you tested at the same revision? Perhaps you have some code that was not committed?

fballiano:
I took the screens from the 1.9.4.x branch 😂 but the data was there because it was created when I was on the right branch, so it shouldn't really be different. I posted the backend screenshot to show that there was some data and it reflects in the frontend. I'll recheck asap anyways

FredericMartinez (dismissed):

empiricompany (dismissed):

app/design/adminhtml/default/default/template/catalog/product/helper/gallery.phtml
```diff
@@ -13,101 +13,136 @@
...
-<input type="hidden" id="<?php echo $_block->getHtmlId() ?>_save_image" name="<?php echo $_block->getElement()->getName() ?>[values]" value="<?php echo $_block->escapeHtml($_block->getImagesValuesJson()) ?>" />
+<input type="hidden" id="<?php echo $this->getHtmlId() ?>_save"
+       name="<?php echo $this->getElement()->getName() ?>[images]"
+       value="<?php echo $this->escapeHtml($this->getImagesJson()) ?>"/>
+<input type="hidden" id="<?php echo $this->getHtmlId() ?>_save_image"
+       name="<?php echo $this->getElement()->getName() ?>[values]"
+       value="<?php echo $this->escapeHtml($this->getImagesValuesJson()) ?>"/>
 <script type="text/javascript">
 //<![CDATA[
 var <?php echo $_block->getJsObjectName(); ?> = new Product.Gallery('<?php echo $_block->getHtmlId() ?>', <?php echo $_block->getImageTypesJson() ?>);
```
empiricompany, line 148: i can't see gallery cause in this template $_block is undefined
Warning: Undefined variable $_block in app/design/adminhtml/default/default/template/catalog/product/helper/gallery.phtml on line 148

empiricompany:
i've tested this PR with this value from backend:
default
english
frech

`eav_attribute`
eav_attribute

`catalog_product_entity_varchar` (not updated)
catalog_product_entity_varchar

`catalog_product_entity_media_gallery` (value_id reference for `catalog_product_entity_media_gallery_value` )
catalog_product_entity_media_gallery

`catalog_product_entity_media_gallery_value`  (correctly updated)
`catalog_product_entity_media_gallery_value`

`catalog_product_flat_1` (not updated no index requested -default store)
`catalog_product_flat_1`

`catalog_product_flat_2`  (not updated no index requested - store english)
`catalog_product_flat_2`

`catalog_product_flat_3`  (not updated no index requested - store french)
`catalog_product_flat_3`

result in frontend title is always from admin value default, for exampla french view:
frontend_french

app/design/adminhtml/default/default/template/catalog/product/helper/gallery.phtml
empiricompany: need to replace $_block with $this

empiricompany:
i've fixed the reported issues
please try this PR: https://github.com/fballiano/openmage/pull/21

app/design/adminhtml/default/default/template/catalog/product/helper/gallery.phtml
empiricompany: fixed here https://github.com/fballiano/openmage/pull/21/commits/a1553fe4cddb0afed75d153848784633a963f81a

fballiano:
I've integrated @empiricompany latest patch to this PR and retested it, it seems to be working correctly to me.

empiricompany (approved):

kyrena (approved):
Tested, seems working.

From #685, this does not work: _"In the admin add another image to the product with the label DDD in the default store view. Mark it as the default base image, small image and thumbnail and set its label to DDD"_ I must save again the product at store view level to have the right label.

Another thing, perhaps already the case, not sure, when I upload a new image, if I select it as default images, when I save the product, I must select again it as default images and save again, else the selection is lost.

fballiano:
true, this needs to be addressed:

- when you upload a new image, the first time you save the product, the fact that you selected the image as base/small/thumb isn't not saved

fballiano:
I've fixed [the last remaining point](https://github.com/OpenMage/magento-lts/pull/2481#issuecomment-2066321711), it was only 1 line that was removed (and now reintroduced).

Since there were already 2 gray check I'll merge this, it's an important bugfix too long overdue (and a huge amount of work to develop it).

PR #2482, opened 2022-08-23
Cosmetic changes to backend footer

fballiano:
@ADDISON74 suggested some changes to the footer of the backend, which I liked a lot coded in this PR.

Screenshot of before and after here in the original discussion https://github.com/OpenMage/magento-lts/discussions/2428

This is taken now from my local dev environment:
Schermata 2022-08-23 alle 23 00 22

Notes:
- I cleaned a minor detail in the javascripts (there was a js changing the target of the links... I removed that)
- the strings changed weren't in any CSV file, so at the moment I didn't put them in any

PR #2486, opened 2022-08-24
Better wording for adminhtml footer

fballiano:
In https://github.com/OpenMage/magento-lts/pull/2482 we changed adminhtml's footer a little bit, but at the end we didn't like the "Report a Bug" wording, which is fine as a meaning, but not very beautiful to see (because of the lowercase "a"), so with this PR we're trying to address this word.

This is my idea:
Schermata 2022-08-24 alle 11 20 15

Alternatives:
- Report Bug (I don't know why but the singular "bug" sounds weird in my head without the "a" in front)
- Report Bugs (my first choice for this PR)
- Bug Report (now that I see I probably prefer this among all of them)

let's see :-)


sreichel:
> Alternatives:

Uppercase "bug/bugs"? :rofl: 

> Bug Report (now that I see I probably prefer this among all of them)

I'd prefer this to.

fballiano:
Switched to "Bug Report"

ADDISON74:
The meaning is slightly different "Report Bug" refers to an action you can take, while "Bug Report" refers to a bug report, something existing, listed. This is where I think of my old friend named Bugtracker.

The form Bug is preferable to Bugs.

As a great programmer well said, often the little things kill us.

kiatng (approved):

fballiano:
then I don't know what to do :-D

sreichel:
> while "Bug Report"

@ADDISON74  :+1:  For my reading this link tells me DO a bug report ...

ADDISON74:
If we approve it now we will certainly come up with another change. We can still advise each other in the meantime, honestly speaking I noticed that the initial wording "Report a Bug" doesn't look good at all in the link, So I got here, but I don't have a final solution. it is true that this wording "Report a Bug" was the most appropriate, clear and concise. 

However, if I look at the two links, one refers to a community and the other to a report, it would also go to Bug Report. If I were to go further, we are not particularly interested in a bug, as that implies an error in the code, but rather to report any issue that has arisen, another wording would be "**Report Issues**".

fballiano:
To me "Bug Report", "Report Bugs" or "Report Issues" are same so any of those is totally fine with me :-)

ADDISON74:
I haven't checked the link which page loads, but of course it should direct here ```https://github.com/OpenMage/magento-lts/issues```.

We have the word issues in the link, in the name of the GitHub section, in which case I would go with the option "Report Issues". Issues is a word that covers many aspects.

fballiano:
yes the link is to https://github.com/OpenMage/magento-lts/issues is "Report Issues" makes total sense, I switched the code already

ADDISON74:
Related to the first link. I noticed that it takes me to the https://www.openmage.org page, then it could be renamed as "OpenMage Project". If it is about community then I think it would be more appropriate to take it to the GitHub Discussions section.

fballiano:
actually, since the website is really less maintained... I'd prefer a "OpenMage Project" that directly links to https://github.com/OpenMage/magento-lts/

ADDISON74:
@fballiano - Please add the changes for the first link (OpenMage Project). What I am seeing in this PR are only two files changed related to the second link.

ADDISON74:
In file ```/app/design/adminhtml/default/default/template/page/footer.phtml``` I would change in **line 25**: 

```php
<a href="<?php echo $this->getConnectWithOpenMageUrl() ?>" target="_blank"><?php echo $this->__('OpenMage Project') ?></a> |
```

In file ```/app/design/adminhtml/default/default/layout/main.xml``` I would change in **line 130**: 

```xml
<action method="setConnectWithOpenMageUrl"><url>https://github.com/OpenMage/magento-lts</url></action>
```

A summary of the changes:

OpenMage Community => OpenMage Project

getConnectwithMagentoUrl() => getConnectwithOpenMageUrl()

https://www.openmage.org/ => https://github.com/OpenMage/magento-lts

fballiano:
done, let's see what people think about this :-)

ADDISON74:
This method name should be changed from setConnectWithMagentoUrl to setConnectWith
OpenMageUrl. Please check my previous post.

fballiano:
changing method names always get people upset, don't know

sreichel:
- add new method
- mark old as deprecated and forward to new one

(Sorry for short answer... writing from phone)

fballiano:
> * add new method
> * mark old as deprecated and forward to new one

super nice.

in this case not needed (sorry, a little tired now), since it's not a real method but a getter/setter call, so I've changed both in the xml and the phtml, shoudln't be a problem.

kiatng (dismissed):
I love it.

sreichel:
> > * add new method
> > * mark old as deprecated and forward to new one
> 
> super nice.
> 
> in this case not needed (sorry, a little tired now), since it's not a real method but a getter/setter call, so I've changed both in the xml and the phtml, shoudln't be a problem.



> > * add new method
> > * mark old as deprecated and forward to new one
> 
> super nice.
> 
> in this case not needed (sorry, a little tired now), since it's not a real method but a getter/setter call, so I've changed both in the xml and the phtml, shoudln't be a problem.

What about other admin themes that have used `getConnectWithMagentoUrl()` in footer?

sreichel:
> done, let's see what people think about this :-)

What i think ... someone spent a lot of time for bringing up openmage.org, but it seems unmaintained :(

Any ideas how to show at least release notes there?

kiatng (dismissed):
Nice.

ADDISON74:
I agree that changing a method is a wrong practice related to Magento, but in this case it is questionable only in terms of whether it is used somewhere else. There are very, very small chances to be used by someone in any theme or extension. However, I appreciate that the initial form has been preserved and new methods have been added. I think this PR is backed. Any other suggestions?

fballiano:
Thanks @sreichel for the changes, you're on a whole other level 😊

I'd like to pause just a sec about changing the link to the openmage.org to the github, I don't want to downplay the official website, it's just that there's obviously less effort put into it compared to the github activity and it's probably nice to click on the link and see that the project is super alive. But I don't want to force this change or upset somebody else over this so let's do it only if we think it's a benefit for the people that are going to click on the link and for openmage itself :-)

sreichel:
> but in this case it is questionable only in terms of whether it is used somewhere else. There are very, very small chances to be used by someone in any theme or extension.

My objection did not come from nowhere. Someone may have a custom footer, with internal links added. 

> I don't want to downplay the official website

I deleted my comment b/c it was not related to your PR. It was just a general observation. I completly okay with linking to github.

Sdfendor:
I for one would have been affected by this change without having the deprecated old methods. We have an overriden `footer.phtml`. Thanks @sreichel.  I like the current solution.

ADDISON74:
If I look at the last version of this PR, changes have already been made and new methods have been created for the links. Those using custom versions of footer.phtml should not be affected. I did a test and I didn't identify any issue between the initial version that I used as custom and the version in this PR.

ADDISON74:
The only aspect I would like to discuss before approval would be the link to the OpenMage Project. 80% of my decision are for https://www.openmage.org. Practically on this page there is a presentation of the project, at the same time there is also a link to other resources such as GitHub. I would like to hear @flyingmana's opinion.

fballiano:
I've reverted the link back to openmage.org, changing that link target should probably be discussed elsewhere :-)
I think this PR is safe now and could be merged

ADDISON74 (approved):

sreichel (approved):

PR #2494, opened 2022-08-25
phpstan: Fixed calls to Mage_Core_Model_Resource_Helper_Abstract

sreichel:
### Description (*)

...

### DIFF
--- a/app/code/core/Mage/Backup/Model/Resource/Db.php
+++ b/app/code/core/Mage/Backup/Model/Resource/Db.php
@@ -87,7 +87,9 @@ class Mage_Backup_Model_Resource_Db
      */
     public function getTableDropSql($tableName)
     {
-        return Mage::getResourceHelper('backup')->getTableDropSql($tableName);
+        /** @var Mage_Backup_Model_Resource_Helper_Mysql4 $helper */
+        $helper = Mage::getResourceHelper('backup');
+        return $helper->getTableDropSql($tableName);
     }
 
     /**
@@ -99,7 +101,9 @@ class Mage_Backup_Model_Resource_Db
      */
     public function getTableCreateSql($tableName, $withForeignKeys = false)
     {
-        return Mage::getResourceHelper('backup')->getTableCreateSql($tableName, $withForeignKeys = false);
+        /** @var Mage_Backup_Model_Resource_Helper_Mysql4 $helper */
+        $helper = Mage::getResourceHelper('backup');
+        return $helper->getTableCreateSql($tableName, $withForeignKeys = false);
     }
 
     /**
@@ -113,8 +117,10 @@ class Mage_Backup_Model_Resource_Db
         $fkScript = '';
         if (!$tableName) {
             $tables = $this->getTables();
+            /** @var Mage_Backup_Model_Resource_Helper_Mysql4 $helper */
+            $helper = Mage::getResourceHelper('backup');
             foreach($tables as $table) {
-                $tableFkScript = Mage::getResourceHelper('backup')->getTableForeignKeysSql($table);
+                $tableFkScript = $helper->getTableForeignKeysSql($table);
                 if (!empty($tableFkScript)) {
                     $fkScript .= "\n" . $tableFkScript;
                 }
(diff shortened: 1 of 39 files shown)


app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
```diff
@@ -675,7 +676,7 @@ public function _getFlatColumnsDdlDefinition()
                 }
                 $prop = $describe[$this->getAttributeCode()];
                 $type = $prop['DATA_TYPE'];
-                $size = ($prop['LENGTH'] ? $prop['LENGTH'] : null);
+                $size = ($prop['LENGTH'] ?: null);
```
kiatng, line 679:
```suggestion
                $size = $prop['LENGTH'] ?: null;
```

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -1328,7 +1332,7 @@ protected function _addAttributeJoin($attributeCode, $joinType = 'inner')
         /**
          * process join type
          */
-        $joinMethod = ($joinType == 'left') ? 'joinLeft' : 'join';
+        $joinMethod = ($joinType === 'left') ? 'joinLeft' : 'join';
```
kiatng, line 1335:
```suggestion
        $joinMethod = $joinType === 'left' ? 'joinLeft' : 'join';
```

app/code/core/Mage/Catalog/Model/Resource/Category/Flat.php
```diff
@@ -705,7 +706,7 @@ protected function _getStaticColumns()
                 'type' => [$ddlType, $options],
                 'unsigned' => $_is_unsigned,
                 'nullable' => $column['NULLABLE'],
-                'default' => ($column['DEFAULT'] === null ? false : $column['DEFAULT']),
+                'default' => ($column['DEFAULT'] ?? false),
```
kiatng, line 709:
```suggestion
                'default' => $column['DEFAULT'] ?? false,
```

fballiano (approved):

app/code/core/Mage/Backup/Model/Resource/Db.php
```diff
@@ -87,7 +87,9 @@ public function getTables()
      */
     public function getTableDropSql($tableName)
     {
-        return Mage::getResourceHelper('backup')->getTableDropSql($tableName);
+        /** @var Mage_Backup_Model_Resource_Helper_Mysql4 $helper */
```
tmotyl, line 90: 1) we should not use mysql4 classes, they are deprecated, arent they?
2) this can be fixed in the phpstan plugin, so phpstan knows what types getResourceHelper factory returns without typehints (similar to how other factories are handled)

sreichel: 1. resource helpers are not deprecated (e.g. `class Mage_Core_Model_Resource_Helper_Mysql4 extends Mage_Core_Model_Resource_Helper_Abstract`)
2. change to draft and check your phpstan plugin :+1:

kiatng (approved):

app/code/core/Mage/Backup/Model/Resource/Db.php
sreichel: @tmotyl checked your plug-in, but it seems more difficult in this case b/c there some more functions called that have to be added.

I'll work on this later (still useful for extensions), but this PR would solve the issues for native phpstan ...

tmotyl: ok.
about mysql4 -> I thought that using classes with mysql4 in name is discouraged, so we should not use them in typehints

sreichel: The mysql4 classes have been deprecated for models/resource-models. The resource-helper is another story ...

Class description 
```
 * Resource helper class for MySql Varien DB Adapter
```

If i have not did not wrong, the mysql4-name is related to ...

https://github.com/OpenMage/magento-lts/blob/76bae0874d1a113684584720e097000c841b4d02/app/etc/config.xml#L34

PR #2496, opened 2022-08-25
Update Mage.php

warleyelias:
Add Statics function to improve debug for developers

### Description (*)
Add Statics function to improve debug for developers

1. In any code is possible to use just 'Mage::depura()' and then the developer could see some value, array or other information to debug their code.
2. In any code is possible to use just 'Mage::printDebugBacktrace()' and then the developer could see the trace until that point and then have some information to debug their code.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -1060,4 +1060,29 @@ final class Mage
     {
 
     }
+    
+    public static function printDebugBacktrace() {
+        $output     = "";
+        $stacks     = debug_backtrace();
+
+        foreach($stacks as $_stack) {
+            if (!isset($_stack['file'])) $_stack['file'] = '[PHP Kernel]';
+            if (!isset($_stack['line'])) $_stack['line'] = '';
+
+            $output .=  $_stack["file"]." ; ".$_stack["line"]." ; ".$_stack["function"]."\n";
+        }
+        return $output;
+    }
+    
+    /*
+    * Print out the value
+    */
+    public static function depura($mixed_var) {
+        $html = '<pre>';
+        $html .= '<div style="display: list-item;font-size: 14px; font-weight: bold; color: #FF0000; background-color: #FFFFCC; border: 1px dotted #000000;" align="left">';
+        $html .= print_r($mixed_var, true);
+        $html .= '</div>';
+        $html .= '</pre>';
+        echo $html;
+    }
 }


ADDISON74:
Thank you for your contribution. I will analyze this proposal. 

Out of curiosity, why did you choose the name of the method "depura"? It has nothing to do with its purpose. Depura means debug in Spanish.

warleyelias:
> Thank you for your contribution. I am not a fan of debugging code by such methods. I think a developer should do it locally in a test environment where an IDE like Phpstorm is used to be configured for debugging. That doesn't mean I won't analyze your proposal, maybe it's useful for debugging, quick immediate, Although many still use Zend::debug, Mage::log, var_dump / die combination, print_r, etc.
> 
> Out of curiosity, why did you choose the name of the method "depura"? Depura means debug in Spanish, but we should keep the code in English.

Hi, "depura" means debug in portuguese, of course we can change for english.

Sdfendor:
I have to agree with @ADDISON74 here. This PR introduces dev code into the production environment without even considering debug mode. 

If one activates developer mode (via environment variable which in turn sets `Mage::$_isDeveloperMode`) the system is already more verbose and prints certain information directly to screen e.g. exceptions via `printException` method (part of `Mage`). Additionally logging via `Zend_Log::DEBUG` level is always possible.  I can't say I see the benefit of those two methods in light of the already existing debugging capabilites of OpenMage itself and the fact that a configured xdebug environment is more flexible in quickly gathering debug information. 

Is it necessary to print the stack trace arbitrarily without an exception thrown (which includes the stack trace in itself)?
The debug/depurra method decorates a debug statement with html (+ inline css) which could look strange depending on the user's custom design. So why decorate the debug output of `print_r` at the first place?

tobihille (changes requested):
First: Thank you for your contribution!

But in my opinion the code is unneccesary, why not use XDebug for development with the added advantage of using breakpoints?

Also printing backtrace is a security concern when done in live environments, it could leak access data like db password or user logins.
To be fair: the function does not "print" anything but it is called "printBacktrace", so I see where this is going.

To give this a positive spin: I *can* imagine that it *might* be useful when this change is rewritten so that these data are *only* written to a log file so that in a case of an error it can be called.
On the other hand: in a case of an error we can use $exception->getTraceAsString() and throw it into Mage::log 🤔

kiatng:
Ref discussion #2466, I bet different developers have different debugging techniques. I do not use xdebug. I do have similar methods as proposed in this PR. But I agree with the others that it's not suitable to add debugging tools to `Mage`. 

Off topic: I use a test controller for debugging in the development server. I use it constantly to test functions and algorithms separately. It's like the browser's js console.

fballiano:
Hi and thanks you very much for your first contrib!
I'm very sorry to say that I wouldn't add it to the Mage class, I don't see a big "value added" compared to a `print_r($var)` or a `print_r(debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS))` so I wouldn't like to have 2 new methods (one with a non-english name) just for doing almost the same thing we can already do with a function call, but with hardcoded css.

:-(

sreichel:
Thanks for your PR, but i'm going to close this.

PR #2501, opened 2022-08-27
Removed double span elements in buttons

fballiano:
The OM base template has a very old HTML structure, with a lot of (now) useless wrappers that (at the creation time) were necessary because CSS was not so powerful.

Now the situation is very different, CSS is very powerfull, all those divs and spans are no longer necessary but not only, they're actually a problem because they make the DOM very big and google hates a big DOM, so it's our duty to provide an HTML that's basic (as in the "base" theme) and not overbloated.

This PR removed the double span elements that are present in every button, they were used to center the text in the button but they're now totally unnecessary.

## How to test

- check out this PR
- be sure you have the rwd theme selected
- navigate the openmage installation and see everything works

## Questions

Why are we checking rwd if we modified base? because base doesn't come with CSS and it's important that we don't brake rwd theme with this PR. This doesn't seem to happen.

Why don't we remove the double span from the rwd theme too? I guess it would break a lot more custom themes and it's not totally necessary, rwd it's our default theme and it could stay a little "opinionated" but those elements do not belong to the base theme.

Impacts on custom themes? This is hard to know... all themes that inherits from rwd should be fine but if they inherit directly from base... they will probably have some problems. What do we do with that?


FredericMartinez (dismissed):

fballiano:
I'll have to rewrite this on branch 20 so I'll close it ATM

PR #2503, opened 2022-08-27
Removed unused fastclick.js

fballiano:
RWD theme includes (only as a filesystem level) a javascript file called fastclick.js, but this file is never added to the html, in fast, once you remove it, there's no occurrence of anything regarding it:

Schermata 2022-08-28 alle 00 23 01

This PR removes it


tmewes (approved):

kiatng (approved):
Ref https://github.com/ftlabs/fastclick:
> Note: As of late 2015 most mobile browsers - notably Chrome and Safari - no longer have a 300ms touch delay, so fastclick offers no benefit on newer browsers, and risks introducing [bugs](https://github.com/ftlabs/fastclick/issues) into your application. Consider carefully whether you really need to use it.

ADDISON74 (approved):

PR #2444, opened 2022-08-17
Fixed detect changes in custom options section (product edit)

fballiano:
This PR should fix https://github.com/OpenMage/magento-lts/issues/1693, when editing custom options for a product (in the backend) changes (the floppy disk icon appearing in the left sidebar) are not always detected correctly.

In `app/design/adminhtml/default/default/template/catalog/product/edit/options/option.phtml` I moved this line:
```
productOption.bindRemoveButtons();
```
a little later in the code, so that it gets executed when the custom option's options are already rendered.

plus I added some more "check if element is changed" to the `app/design/adminhtml/default/default/template/catalog/product/edit/options/type/select.phtml` file

PR #2504, opened 2022-08-27
Removed Internet Explorer related libraries

fballiano:
This PR is for v20 only, where we're removing support for IE.

This PR removed selectivizr http://selectivizr.com and matchMedia https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia#browser_compatibility, two things only necessary for IE6-8.


tmewes (approved):

Sdfendor (approved):

ADDISON74 (approved):

fballiano:
@Flyingmana as per https://github.com/OpenMage/rfcs/blob/main/accepted/0001-change-pr-merging-rules.md this should be merged

luigifab (approved):
Not tested, but we don't use these files with our rwd derived theme.

ADDISON74:
Those who want to test and approve PRs should bookmark this page:

https://fabrizioballiano.com/om/pr_status.html

We have many PRs that still need an approval and some of them are obvious, for example, this one https://github.com/OpenMage/magento-lts/pull/2444.

PR #2507, opened 2022-08-28
Removed border-top from .block-title class (rwd theme)

fballiano:
In the rwd theme there's a border-top added to the block-title class, which shows like this:

Schermata 2022-08-28 alle 17 30 28

I see why it was added years ago, but now I think it looks nicer and cleaner without that border:

Schermata 2022-08-28 alle 17 30 00

It is also better on mobile in my opinion.

### DIFF
--- a/skin/frontend/rwd/default/css/styles-ie8.css
+++ b/skin/frontend/rwd/default/css/styles-ie8.css
@@ -656,7 +656,6 @@ h6, .h6 {
   position: relative;
   padding: 10px 0 0;
   margin-bottom: 5px;
-  border-top: 1px solid #cccccc;
 }
 .block-title h2,
 .block-title h3,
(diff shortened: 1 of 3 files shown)


ADDISON74 (approved):

tmewes (approved):

kiatng (approved):

sreichel (approved):

ADDISON74:
I didn't test this PR but the idea of removing those lines was a natural one. I have one question. what happens when the viewport is reduced? Were those lines the delimiters between the sections?

fballiano:
I tested the mobile version and it's better (IMHO) now, here's a couple of screenshots

Schermata 2022-08-30 alle 10 03 47
Schermata 2022-08-30 alle 10 03 41

PR #2533, opened 2022-09-01
[preview] get rid of "Adminhtml"

sreichel:
### Description (*)

I'd like to get rid off `Adminhtml`. It is just a collection of classes that belong to a specific module.

My idea is to move all code to correct place and keep empty/deprecated adminhtml-classes for BC (like mysql4-classes).


justinbeaty:
I would definitely support this, but question:

if we are changing the block class names, are we updating these in all the layout xml files and in any PHP where we createBlock manually? Or do we continue to call adminhtml/blockname ?

sreichel:
This is why it is on draft.

I dont want to change anything for layouts/xml ... so class names have to be resolved in another way.

kiatng:
This PR will make it easier to uninstall non-required modules, ref PR #2455. With an uninstall script, the files in `app/code/core/Mage/`,  `app/design`, `app/etc/modules`, and other folders, can be removed to achieve a clean uninstall. 

On that perspective, we can keep the code for the _required modules_ in `Adminhtml`.

ADDISON74:
Why did you close this draft? Any solid reason? I liked the idea.

sreichel:
I'll reopen later.

PR #2554, opened 2022-09-05
phpstan: fixed "Access to an undefined property"

sreichel:
### Description (*)

Set undefined properties. I hope its correct to make "new" properties public. 


fballiano:
why not protected?

elidrissidev:
They should probably be protected by default unless there is a usage outside the class.

sreichel:
> They should probably be protected by default unless there is a usage outside the class.

Yes, BUT ... they were public all the time and someone could have used them outside the class.

I can change it, because they should have not been used outside. Let me know.

fballiano:
right, but php 8.2 deprecates public properties so...
https://php.watch/versions/8.2/dynamic-properties-deprecated
what's the right choice here?

sreichel:
Public is still okay, but not dynamically set.

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Container.php
```diff
@@ -31,6 +31,14 @@ class Mage_Adminhtml_Block_Widget_Grid_Container extends Mage_Adminhtml_Block_Wi
     protected $_backButtonLabel;
     protected $_blockGroup = 'adminhtml';
 
+    /**
+     * @var string
+     */
+    public $_block;
```
fballiano, line 37: could this be protected?

app/code/core/Mage/Catalog/Model/Resource/Category/Tree.php
```diff
@@ -67,9 +67,13 @@ class Mage_Catalog_Model_Resource_Category_Tree extends Varien_Data_Tree_Dbp
      */
     protected $_storeId                          = null;
 
+    /**
+     * @var array
+     */
+    public $_inactiveItems;
```
fballiano, line 73: and this?

sreichel: Missed both. Thanks.

fballiano (approved):

ADDISON74 (approved):

PR #2555, opened 2022-09-05
Remove redundant polyfill code for error reporting constants

elidrissidev:
### Description (*)
This PR removes redundant polyfill code that defines some error reporting constants which should be available in all of our supported PHP versions anyways. Correct if I'm wrong as I couldn't find specific information about when these constants were introduced.

1. Everything should remain the same.

### DIFF
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -121,15 +121,6 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
     if ($errno == 0) {
         return false;
     }
-    if (!defined('E_STRICT')) {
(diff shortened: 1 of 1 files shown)


fballiano:
this is what I found

E_STRICT -> https://www.examveda.com/which-version-of-php-introduced-estrict-error-level-35395/
E_RECOVERABLE_ERROR  -> https://www.php.net/releases/5_2_0.php
E_DEPRECATED -> https://electrictoolbox.com/php-prevent-e-deprecated-error-messages/

so you should be right assuming we don't need that code anymore

Sdfendor (approved):

fballiano (approved):

ADDISON74 (approved):

PR #2564, opened 2022-09-06
Removed unused controller Mage_Shipping_ShippingController

fballiano:
### Description (*)

`Mage_Shipping_ShippingController` was deprecated in M1.3 and seems not to be used anywhere, if you search for `shipping/view` there's nothing, and in the `myaccount` area there's only a link to the `track shipment` url, but not to view the shipping (what this code did).

If you check the controller's code, it wouldn't do anything anyway.

I've also checked its layout file `app/design/frontend/base/default/layout/shipping.xml` and there's nothing about this controller.

If I didn't miss anything it should be safe to remove this file.


kiatng (approved):
URL domain/shipping/shipping/view returns empty page.

sreichel (approved):

PR #2569, opened 2022-09-07
BackUp function not works

Danieleeffe1:
I have updated the openmage with the latest version available

I have enabled the backup module

the backup function blocks after a few seconds and gives me the message:

"An error occurred while creating the backup."

php7.4
apache2

original magento system 1.9.4.5

updated to openmage 20.0.13

update method used:
pasted openmage file in the magento root

last update performed from openmage source 20.0.13 to 20.0.16

update method used:
copied and pasted openmage file into openmage root


fballiano:
I think this was opened as a PR by mistake, could you please open it in the issue section? :-)

PR #2562, opened 2022-09-06
Fixed whitespace for docblocks

fballiano:
Continuing to fix whitespace and formatting for docblocks, also switched http to https for license links.

This is the final one for app/code/core, this task drove me mad :-)

PR #2572, opened 2022-09-07
Fixed PHPStan error with Mage_Catalog_Model_Resource_Product_Option_Collection

fballiano:
Discovered in https://github.com/OpenMage/magento-lts/pull/2562 by @elidrissidev, it shouldn't be caused with that PR but anyway this PR should fix it.

Let's see if all lights are gree.

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Option/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Option/Collection.php
@@ -150,8 +150,8 @@ class Mage_Catalog_Model_Resource_Product_Option_Collection extends Mage_Core_Mo
         }
         if (!empty($optionIds)) {
             /** @var Mage_Catalog_Model_Resource_Product_Option_Value_Collection $values */
-            $values = Mage::getModel('catalog/product_option_value')
-                ->getCollection()
+            $values = Mage::getModel('catalog/product_option_value')->getCollection();
+            $values
                 ->addTitleToResult($storeId)
                 ->addPriceToResult($storeId)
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Catalog/Model/Resource/Product/Option/Collection.php
```diff
@@ -149,8 +149,9 @@ public function addValuesToResult($storeId = null)
             $optionIds[] = $option->getId();
         }
         if (!empty($optionIds)) {
-            $values = Mage::getModel('catalog/product_option_value')
-                ->getCollection()
+            /** @var Mage_Catalog_Model_Resource_Product_Option_Value_Collection $values */
+            $values = Mage::getModel('catalog/product_option_value')->getCollection();
+            $values
```
kiatng, line 154: No need to instantiate model but may be not BC. 
```suggestion   
            $collection = Mage::getResourceModel('catalog/product_option_value_collection');
            $collection
```

fballiano: I don't know, probably nobody did an override on that class and the getcollection method but this type of code is all over the code, I wouldn't change it in this specific PR

elidrissidev (approved):

fballiano:
I found where it originated and fixed it reverting previous commits

elidrissidev (approved):

ADDISON74:
All Green.

fballiano:
yeah, i was waiting for the checks to end :-)

PR #2582, opened 2022-09-09
Refactoring: Replace dirname(__FILE__) With Corresponding Constant

Sdfendor:
### Description (*)

`dirname(__FILE__)` returns the directory of the current file. Since PHP 5.3.0 the constant `__DIR__` exists and is equivalent to the function call (see [Magic constants](https://www.php.net/manual/en/language.constants.magic.php)). At the same time it's easier to grasp at first glance. Our current minimum supported PHP version is 7.3.0. 
The PR replaces all `dirname(__FILE__)` calls with __DIR__ except those in *lib/Zend* and *lib/phpseclib* which are in the process of being removed from the codebase to become externally maintained dependencies. 

### Questions or comments

In two instances `dirname(dirname(__FILE__))` is replaced with `dirname(__DIR__)`. Another variant here would be `dirname(__FILE__),2)` (directory name above the current one). I've chosen the first option but have no disposition which one better to use. Maybe some of you have an opinion regarding e.g. ease of understanding? 

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -20,7 +20,7 @@
 
 define('DS', DIRECTORY_SEPARATOR);
 define('PS', PATH_SEPARATOR);
-define('BP', dirname(dirname(__FILE__)));
+define('BP', dirname(__DIR__));
 
 Mage::register('original_include_path', get_include_path());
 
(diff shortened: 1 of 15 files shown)


ADDISON74 (approved):

sreichel:
> Maybe some of you have an opinion regarding e.g. ease of understanding?

I think `dirname(__FILE__), 2)` is more clear.

fballiano:
I like dirname(__DIR__) because:
- didn't even know dirname had a second parameter
- this way we always use __DIR__ instead of mixing __DIR__ and __FILE__

Sdfendor:
> didn't even know dirname had a second parameter

The php documentation states only since 7.0 so it's wasn't always there 😅.

sreichel:
Agree with both, but for me this is more clear ... b/c i see we look one level above ...

```
dirname(dirname(__FILE__))
dirname(__FILE__, 2)
```

PR #2591, opened 2022-09-12
Fix Phpdoc Typos in Various Files

Sdfendor:
### Description (*)

Fixed a bunch of typos all over the core but in PHPDoc comment blocks only. During that I noticed some methods with those typos as well but that is a topic for a follow-up PR.


fballiano (approved):

sreichel:
Haha ... a few minutes ago I thought about running some spellchecker 🙃

ADDISON74:
Who know how many such mistakes are there in the PHP code. These found are minor ones, but good that they have been solved. 

I don't approve anymore because there are already two greens.

PR #2592, opened 2022-09-12
Replace Date Format String Literals With Existing Constants

Sdfendor:
### Description (*)

#2474 inspired me to search for usages of string literals containing the datetime format `'Y-m-d H:i:s'` that is defined already as constant multiple times. 

`Varien_Db_Adapter_Pdo_Mysql::TIMESTAMP_FORMAT` 
`Varien_Date::DATETIME_PHP_FORMAT`

Imo to have multiple constants with the same content is fine because the constants have different contexts. This PR replaces string literals in each case with one of those two constants. I selected by context foreach case. I ensured that the corresponding columns in database have the timestamp type. 


app/code/core/Mage/AdminNotification/Model/Inbox.php
```diff
@@ -128,7 +128,7 @@ public function add($severity, $title, $description, $url = '', $isInternal = tr
         if (is_array($description)) {
             $description = '<ul><li>' . implode('</li><li>', $description) . '</li></ul>';
         }
-        $date = date('Y-m-d H:i:s');
+        $date = date(Varien_Db_Adapter_Pdo_Mysql::TIMESTAMP_FORMAT);
```
fballiano, line 131: I'd use Varien_Db_Adapter_Pdo_Mysql ::DATETIME_FORMAT since the name is more similar to the meaning, and the constant is the same

Sdfendor: I thought about those two. You are completely right, there is the third constant that also defines a format and it is the same as  `TIMESTAMP_FORMAT`. I picked `TIMESTAMP_FORMAT` for one main reason: the datrabase columns I checked all have the type **timestamp** not the type **datetime** which both are present in MySQL, are similar but have distinguishing characteristics. I totally think that the format of both will never be different though.

sreichel (approved):

app/code/core/Mage/AdminNotification/Model/Inbox.php
fballiano: wow, they've a timestamp type but the php formatting is YMDIms?? auch

fballiano:
fixed conflict

fballiano (approved):
I'll approve although this timestamp thing still makes me wonder

sreichel (approved):

PR #2598, opened 2022-09-14
[test] Move translations to module scope

sreichel:
### Description (*)

Just an idea to add translation files to the corresponding module ...

### Questions or comments

Any ideas for a simple script that moves files (autmatically) from `app/locale` to `app/code/<scope>/<vendor>/<module>/locale`?

ToDo: email templates? Leave them there?


ADDISON74:
What are the benefits of this change? I have a lot for keeping the current one.

fballiano:
it would be interesting for decoupling but at this stage I wouldn't do it.

sreichel:
> What are the benefits of this change? I have a lot for keeping the current one.

Only real benefit is to have all (csv) files that belong to a module in one place. I'm just testing to split core/Mage into composer installable modules and have no files outside module dir :)

ADDISON74:
We have as a legacy from Magento 1 a separation of the files in several places. Thus, in /app/design/ we find the template/layout/locale files, in /app/etc/modules we find the modules information, and in /app/locale we find the translation files for Frontend/Backend/Email. In Magento 2, the developers implemented what you propose for translation files.

What you consider a real benefit comes with many inconveniences on the Magento 1 file structure. At least that's what I get after a primary analysis of the proposed change. I will present them to you these days to evaluate them together.

PR #2586, opened 2022-09-10
[PHP 8.1] Fix passing null to non-nullable internal function params

elidrissidev:
### Description (*)
This PR fixes lots of "Passing null to non-nullable internal function parameters" deprecation notices introduced in PHP 8.1, which I came across when navigating or performing different actions on the frontend and admin.

The commits are grouped by function for easier review, and the fixes favor casting when appropriate, and PHPdocs to determine where to do the change. For example: `Mage_Core_Helper_Data::escapeHtml` PHPdoc says it accepts a non-nullable string, so instead of casting the parameter to a string inside the method, I cast the arguments in the usages instead.

Deprecation errors in `lib/Zend` were skipped in the hope we'll switch to zf1-future soon (?).

### Related Pull Requests

- #1812

### Manual testing scenarios (*)

1. Make sure the `error_reporting` directive in your `php.ini` includes deprecation errors, then comment out the lines below:
https://github.com/OpenMage/magento-lts/blob/97e0b70e0c2f762a03e1124c04a6fabb2be9e98f/app/code/core/Mage/Core/functions.php#L125-L128

### Questions or comments

Please make sure to review thoroughly before approving.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2615, opened 2022-09-22
[test-only] php8.1 w/o suppressed errors

sreichel:
### Description (*)

TEST. Just for reviews.

### Related Pull Requests

1. See OpenMage/magento-lts#2586

1. php8.1
2. developer mode = ON
3. comment lines as suggested in #2586

### DIFF
--- a/lib/Varien/Data/Form/Element/Abstract.php
+++ b/lib/Varien/Data/Form/Element/Abstract.php
@@ -208,7 +208,7 @@ abstract class Varien_Data_Form_Element_Abstract extends Varien_Data_Form_Abstra
      */
     protected function _escape($string)
     {
-        return htmlspecialchars($string, ENT_COMPAT);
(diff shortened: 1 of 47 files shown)


lib/Varien/Data/Form/Element/Abstract.php
```diff
@@ -208,7 +208,7 @@ public function removeClass($class)
      */
     protected function _escape($string)
     {
-        return htmlspecialchars($string, ENT_COMPAT);
+        return is_string($string) ? htmlspecialchars($string, ENT_COMPAT) : $string;
```
tmotyl, line 211: why not casting to string before passing to htmlspecialchars? This way we're sure the method always returns a string. With your change the function will not always return a string

sreichel: You're right. htmlspecialchars on null returned string before. I#ll change.

PR #2617, opened 2022-09-24
bugfix + add module names to helper

sreichel:
### Description (*)

Read: https://magento.stackexchange.com/questions/46255/ecomdev-phpunit-translation-not-working-with-mocked-helper/46256#46256

- 1st commit: added module name to all helpers
- 2nd: fix for _Imagine what bugs a module name starting with "Helper" would cause..._

### DIFF
--- a/app/code/core/Mage/Admin/Helper/Data.php
+++ b/app/code/core/Mage/Admin/Helper/Data.php
@@ -27,6 +27,8 @@
  */
 class Mage_Admin_Helper_Data extends Mage_Core_Helper_Abstract
 {
+    protected $_moduleName = 'Mage_Admin';
+
     /**
      * Configuration path to expiration period of reset password link
      */
--- a/app/code/core/Mage/AdminNotification/Helper/Data.php
+++ b/app/code/core/Mage/AdminNotification/Helper/Data.php
@@ -27,6 +27,8 @@
  */
 class Mage_AdminNotification_Helper_Data extends Mage_Core_Helper_Abstract
 {
+    protected $_moduleName = 'Mage_AdminNotification';
+
     /**
      * Last Notice object
      *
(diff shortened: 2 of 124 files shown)


kiatng (approved):

elidrissidev:
I understand the bug fix, but is it really necessary to set the module name explicitly in every helper?

fballiano:
> I understand the bug fix, but is it really necessary to set the module name explicitly in every helper?

exactly what I was thinking :-(

sreichel:
I think it is bit faster to get module name from property, instead of retrieve it from classname.

sreichel:
This method is called on nearly every page for at least ten times (or more often) ...

```
protected function _getModuleName()
{
    if (!$this->_moduleName) {
        $class = get_class($this);
        $this->_moduleName = substr($class, 0, strpos($class, '_Helper'));
    }
    return $this->_moduleName;
}
```

Why should we not set `$_moduleName` to skip that useless calls?

kiatng:
Why I agree with @sreichel: https://github.com/OpenMage/magento-lts/blob/fb12d0f2ca70ececdf99186115c25123d1c69f55/app/code/core/Mage/Core/Helper/Abstract.php#L35-L35

3 reasons:

1. It's available for the child helpers
2. It's harmless for the children to make use of it
3. It saves many computing cycle times

Flyingmana:
> Why to not set $_moduleName to skip that useless calls?

because most helper classes should not need to know their module name.
It could be seen as a violation of separation of concerns.
Especially as there was a best practice to only use the translate method from the data helper I think?

It also goes further away on how the Block classes do it, which do it fully dynamically.

anyway, this optimization saves around 0.02ms per request according to my abstracted benchmark, if all of these classes are loaded and used.

Flyingmana (approved):

elidrissidev (approved):

PR #2619, opened 2022-09-26
conditionaly insert/remove translator inline on frontend

empiricompany:
add condition to inline translator scripts
include only if translation config is enabled and and dev ip is allowed

remove Translator data on frontend like "var Translator = new Translate..."
and the inclusion of the script mage/translate.js


empiricompany:
i'm sorry i'm wrong damn!
we can't remove translattion data cause is used by javascript function "Translator.translate("

PR #2622, opened 2022-09-27
No need to check for both isset() and empty() at the same time

kiatng:
### Description (*)
These conditions

```php
isset($someVar[0]) && !empty($someVar[0];
isset($someVar[0]) && $someVar[0];
```
are the same as 
```php
!empty($someVar[0]);
```

More info [here](https://phpdelusions.net/articles/empty).

### Questions or comments
To me, this improves readability. Do you like it?


elidrissidev:
I agree with this change like the PHP documentation for `empty` says:
> That means empty() is essentially the concise equivalent to !isset($var) || $var == false.

Will review once ready.

app/code/core/Mage/Adminhtml/controllers/Catalog/Product/AttributeController.php
```diff
@@ -179,8 +179,8 @@ protected function _filterPostData($data)
             }
 
             if (!empty($data['option']) && !empty($data['option']['value']) && is_array($data['option']['value'])) {
-                $allowableTags = isset($data['is_html_allowed_on_front']) && $data['is_html_allowed_on_front']
-                    ? sprintf('<%s>', implode('><', $this->_getAllowedTags())) : null;
+                $allowableTags = empty($data['is_html_allowed_on_front'])
```
Sdfendor, line 182: A small question here: Wouldn't it be enough to use just isset here, so remove the second part? Because the defintion of isset is: 
`Determine if a variable is declared and is different than null`

In this case a flip of the ternary operator wouldn't be necessary.

kiatng: For `isset($data['is_html_allowed_on_front']) && $data['is_html_allowed_on_front']`, `isset()` checks if the key exists in the array and  `$data['is_html_allowed_on_front']` is not `null`. The second part checks for [a truthy value](https://developer.mozilla.org/en-US/docs/Glossary/Truthy).  This is the same as `isset($someVar[0]) && $someVar[0];` in my description, which is the same as `!empty($someVar[0])`. This [article](https://phpdelusions.net/articles/empty) gives a good explanation. 

When the `!` is removed in `!empty($someVar[0])`, we need to flip the value.

```php 
if (!empty($someVar[0])) {
    $var = 'not empty';
} else {
    $var = 'empty';
}
$var = !empty($someVar[0]) ? 'not empty' : 'empty';

//it's the same as:
if (empty($someVar[0])) {
    $var = 'empty';
} else {
    $var = 'not empty';
}
$var = empty($someVar[0]) ? 'empty' : 'not empty';
```

Sdfendor: Yeah I overlooked that the value of `$data['is_html_allowed_on_front']` has to be taken into account, not only that there is something that is not null but that there is something truthy. Therefore to check `empty()` is the right choice. Thanks for your answer.

luigifab:
I like it, it's an automated tool or manual edit?

kiatng:
> I like it, it's an automated tool or manual edit?

Manual. Still draft because I am not 100% sure if this is worthwhile.

matteotestoni (approved):

kiatng:
Closing because it's unnecessary, and also @sreichel posted a link about how `empty()` is bad and should not be used.

Issue #2614, opened 2022-09-21
Table 'dbName.core_file_storage' doesn't exist


daboss84:
I am on OpenMage 20.0.16 and PHP8.0

In the Exception.log, I see the following entries:

> PDOException: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'dbName.core_file_storage' doesn't exist in /var/www/vhosts/....../lib/Zend/Db/Statement/Pdo.php:228
> Stack trace:
> #0 /var/www/vhosts/....../lib/Zend/Db/Statement/Pdo.php(228): PDOStatement->execute()
> #1 /var/www/vhosts/....../lib/Varien/Db/Statement/Pdo/Mysql.php(103): Zend_Db_Statement_Pdo->_execute()
> #2 /var/www/vhosts/....../app/code/core/Zend/Db/Statement.php(290): Varien_Db_Statement_Pdo_Mysql->_execute()
> #3 /var/www/vhosts/....../lib/Zend/Db/Adapter/Abstract.php(480): Zend_Db_Statement->execute()
> #4 /var/www/vhosts/....../lib/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query()
> #5 /var/www/vhosts/....../lib/Varien/Db/Adapter/Pdo/Mysql.php(501): Zend_Db_Adapter_Pdo_Abstract->query()
> #6 /var/www/vhosts/....../lib/Zend/Db/Adapter/Abstract.php(756): Varien_Db_Adapter_Pdo_Mysql->query()
> #7 /var/www/vhosts/....../app/code/core/Mage/Core/Model/Resource/File/Storage/Database.php(144): Zend_Db_Adapter_Abstract->fetchRow()
> #8 /var/www/vhosts/....../app/code/core/Mage/Core/Model/File/Storage/Database.php(118): Mage_Core_Model_Resource_File_Storage_Database->loadByFilename()
> #9 /var/www/vhosts/....../get.php(148): Mage_Core_Model_File_Storage_Database->loadByFilename()
> #10 {main}
> 
[...]

PR #2627, opened 2022-09-30
Check if remote storage is enabled before saving local file

elidrissidev:
### Description (*)
This PR will add a simple check if remote DB storage for media is enabled before trying to write the file locally (using `get.php`). This will prevent unnecessary exceptions from being logged that say `core_file_storage` doesn't exist, since it's often created at runtime.

1. Fixes OpenMage/magento-lts#2614

1. Have logging enabled and try to access a non-existent media file. e.g: `http://<magento_host>/media/catalog/notfound.jpg`. You will see an exception logged before this patch, but not after.

### DIFF
--- a/app/code/core/Mage/Core/Model/File/Storage.php
+++ b/app/code/core/Mage/Core/Model/File/Storage.php
@@ -105,9 +105,6 @@ class Mage_Core_Model_File_Storage extends Mage_Core_Model_Abstract
             case self::STORAGE_MEDIA_DATABASE:
                 $connection = $params['connection'] ?? null;
                 $model = Mage::getModel('core/file_storage_database', ['connection' => $connection]);
-                // Always initialize database storage to ensure the table exists
-                // see: https://github.com/OpenMage/magento-lts/pull/2627
-                $params['init'] = true;
                 break;
             default:
                 return false;
--- a/app/code/core/Mage/Core/Model/File/Storage.php
+++ b/app/code/core/Mage/Core/Model/File/Storage.php
@@ -105,6 +105,9 @@ class Mage_Core_Model_File_Storage extends Mage_Core_Model_Abstract
             case self::STORAGE_MEDIA_DATABASE:
                 $connection = $params['connection'] ?? null;
                 $model = Mage::getModel('core/file_storage_database', ['connection' => $connection]);
+                // Always initialize database storage to ensure the table exists
+                // see: https://github.com/OpenMage/magento-lts/pull/2627
+                $params['init'] = true;
                 break;
             default:
                 return false;
--- a/get.php
+++ b/get.php
@@ -141,6 +141,11 @@ if (substr_count($relativeFilename, '/') > 10) {
     sendNotFoundPage();
 }
 
+// Nothing to do if DB storage is disabled
+if (!Mage::helper('core/file_storage_database')->checkDbUsage()) {
+    sendNotFoundPage();
+}
+
 $localStorage = Mage::getModel('core/file_storage_file');
 $remoteStorage = Mage::getModel('core/file_storage_database');
 try {


sreichel:
Doesnt fix it for me.

Config is set to DB w/o existing table ...

- there is an error for product image upload
- errors on frontend

Sample data: https://openmage.ddev.site/men/new-arrivals/linen-blazer.html

```
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'db.core_file_storage' doesn't exist, query was: SELECT `e`.* FROM `core_file_storage` AS `e` WHERE (filename = 'msj012t_2.jpg') AND (directory = 'catalog/product/m/s')

#0 /var/www/html/lib/Varien/Db/Statement/Pdo/Mysql.php(103): Zend_Db_Statement_Pdo->_execute(Array)
#1 /var/www/html/app/code/core/Zend/Db/Statement.php(289): Varien_Db_Statement_Pdo_Mysql->_execute(Array)
#2 /var/www/html/lib/Zend/Db/Adapter/Abstract.php(480): Zend_Db_Statement->execute(Array)
#3 /var/www/html/lib/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('SELECT `e`.* FR...', Array)
#4 /var/www/html/lib/Varien/Db/Adapter/Pdo/Mysql.php(501): Zend_Db_Adapter_Pdo_Abstract->query('SELECT `e`.* FR...', Array)
#5 /var/www/html/lib/Zend/Db/Adapter/Abstract.php(756): Varien_Db_Adapter_Pdo_Mysql->query('SELECT `e`.* FR...', Array)
#6 /var/www/html/app/code/core/Mage/Core/Model/Resource/File/Storage/Database.php(144): Zend_Db_Adapter_Abstract->fetchRow(Object(Varien_Db_Select))
#7 /var/www/html/app/code/core/Mage/Core/Model/File/Storage/Database.php(118): Mage_Core_Model_Resource_File_Storage_Database->loadByFilename(Object(Mage_Core_Model_File_Storage_Database), 'msj012t_2.jpg', 'catalog/product...')
#8 /var/www/html/app/code/core/Mage/Core/Helper/File/Storage/Database.php(202): Mage_Core_Model_File_Storage_Database->loadByFilename('catalog/product...')
#9 /var/www/html/app/code/core/Mage/Catalog/Model/Product/Image.php(787): Mage_Core_Helper_File_Storage_Database->saveFileToFilesystem('/var/www/html/m...')
#10 /var/www/html/app/code/core/Mage/Catalog/Model/Product/Image.php(346): Mage_Catalog_Model_Product_Image->_fileExists('/var/www/html/m...')
#11 /var/www/html/app/code/core/Mage/Catalog/Helper/Image.php(165): Mage_Catalog_Model_Product_Image->setBaseFile('/m/s/msj012t_2....')
#12 /var/www/html/app/design/frontend/rwd/default/template/catalog/product/view/media.phtml(37): Mage_Catalog_Helper_Image->init(Object(Mage_Catalog_Model_Product), 'image')
```

sreichel:
I even cant switch to "file" from admin config ...

- try to change config to "file" and save ... validation error "Synchronization is required."
- try synchronization ... "Base table or view not found"

....

Btw ... "synchronization" to DB means images get deleted ...

When you test this PR/issue, make sure to have a backup of your media dir ... :(

elidrissidev:
This PR was only meant to fix the error when the config is set to file, but that might be a good thing to include as well.

sreichel:
I've changed the value in core_config_data, but this could also happen with tools that import/export config.

elidrissidev:
I'm not a fan of the second commit as the storage backend isn't intended to be updated manually from the database, but here it is. It will initialize the database storage every time it's loaded which will create the table only if it doesn't exist. The second commit needs to be analyzed to ensure that it won't affect performance though.

fballiano:
at this point why don't we add an update script that creates the table?

elidrissidev:
There is already one but it's conditioned by `checkDbUsage`.

colinmollenhour:
Setting config values directly in the database, while it works in most cases, bypasses the before/after save methods of the backend models so I think should be viewed as an "unsupported" action.. So in this case the fix is to change the storage method via the admin UI so the hooks can run as intended. Import/export tools should ideally use the models to update the config rather than manipulating the db directly.

elidrissidev:
I reverted the second change to keep this PR in scope of the linked issue, which is what it was supposed to solve in the first place. Anything else can be addressed in a following PR.

kiatng:
I do not know how to test `get.php`. Please explain if you know. Reading its code, I am not entirely sure, but I think it's main sequence is 

1. Lines 66-90, it is `storage_file` logic: get `$mediaDirectory` from `resource_config.json`,  if the requested file exists, it sends it and `exit`. 
2. Lines 91-end:  it is `storage_database` logic: get the filename from db and sends it.

I think for `storage_file` config, `$mediaDirectory` is defined (guaranteed?) and table `core_file_storage` does not exist. If this is true then
https://github.com/OpenMage/magento-lts/blob/52f2c7543f0790c50565422d1f4cb31dab31a5de/get.php#L82-L90

we can just insert `sendNotFoundPage();` at line 90, after `sendFile($filePath);`. This should save some computing cycle.

sreichel (approved):

kiatng (approved):

Issue #1953, opened 2022-01-03
weird hardcoded credit card message on checkout page


f1-outsourcing:
design/frontend/base/default/template/checkout/onepage/review/totals.phtml

```
     36     <?php if ($this->needDisplayBaseGrandtotal()):?>
     37     <tr>
     38         <td class="a-right" colspan="<?php echo $_colspan; ?>">
     39             <small><?php echo $this->helper('sales')->__('Your credit card will be charged for') ?></small>
     40         </td>
     41         <td class="a-right">
     42             <small><?php echo $this->displayBaseGrandtotal() ?></small>
     43         </td>
     44     </tr>
```

![english](https://user-images.githubusercontent.com/13450219/147979224-ac23cd0e-0941-4f4e-8f2e-ece3a70ce8d6.png)
![german](https://user-images.githubusercontent.com/13450219/147979227-ae84ae65-12c4-4ed1-bad7-b2d4089f597d.png)

trabulium:
The check needDisplayBaseGrandtotal is when the Quote Currency is different to the Base Currency so the message appears. The real issue is why your Grand Total is $0.00,  I assume Gesamtsumme (great sum) is German for Grand Total. So the reason why "Your credit card will be charged for" is $0.00 is displayed when the card gets charge an Euros but you are display Pounds to the end user. If you want to add a German translation, you need to add 

"Your credit card will be charged for","Ihre Kreditkarte wird belastet mit"
to this file:
app/locale/de_DE/Mage_Sales.csv

f1-outsourcing:
But this is not a credit card check out.

fballiano:
I think it's just a wrongly written message, when base and current currency differ, there's this message appearing. The sentence should be different tho.

PR #2628, opened 2022-09-30
Update hardcoded credit card message in checkout

elidrissidev:
### Description (*)

This PR will update the hardcoded message in checkout when the quote currency is different from the base one. The old message always says "Your credit card will be charged for" regardless of the payment method. I updated it with the same message from the order view in frontend that says "Grand Total to be Charged", which should be translated in language packs.

1. Fixes OpenMage/magento-lts#1953

1. Have at least 2 allowed currencies in a store.
2. Set your currency to something different from the base currency (`currency/options/base`).
3. Go through the checkout and you should see the new message in "Review" step.


fballiano:
should we add the message to some CSV?

elidrissidev:
It already exists, I reused it from the frontend order view.

fballiano (approved):

fballiano:
> It already exists, I reused it from the frontend order view.

great

kiatng (approved):

PR #2629, opened 2022-09-30
Fixes issue #475, reduce reprocessed jpeg image file size by defaulting image quality to 85%

kiatng:
### Description (*)
Config `general/reprocess_images/active` is removed and replaced with `admin/security/reprocess_image_quality`, where it is default to 85. ~For BC, checking of `general/reprocess_images/active` takes precedent.~

image

1. Fixes OpenMage/magento-lts#475


app/code/core/Mage/Core/Model/File/Validator/Image.php
```diff
@@ -85,8 +85,13 @@ public function validate($filePath)
         list($imageWidth, $imageHeight, $fileType) = getimagesize($filePath);
         if ($fileType) {
             if ($this->isImageType($fileType)) {
-                /** if 'general/reprocess_images/active' false then skip image reprocessing. */
-                if (!Mage::getStoreConfigFlag('general/reprocess_images/active')) {
+                /** Check deprecated 'general/reprocess_images/active' for BC. If false then skip image reprocessing. */
+                if (Mage::getStoreConfig('general/reprocess_images/active') !== null) {
+                    $imageQuality = Mage::getStoreConfigFlag('general/reprocess_images/active') ? 100 : 0;
```
elidrissidev, line 90: Why not migrate the config to the new location here? And maybe also delete the old one?

fballiano: I think because there was no backend option for that, it was just an entry in the config xmls, which I don't think it's possible to migrate

elidrissidev: Right, I didn't notice.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1280,6 +1280,16 @@
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
                         </crate_admin_user_notification>
+                        <reprocess_image_quality translate="label comment">
+                            <label>Image Reprocess Quality</label>
+                            <comment>Optimum value is 85, higher value will increase file size. Set to 0 to turn off images reprocessing, which may cause security risks.</comment>
+                            <frontend_type>text</frontend_type>
```
elidrissidev, line 1286: `<frontend_type>text</frontend_type>` not required as it's the default.

ADDISON74:
This change must be accompanied by an explanation in the README. First of all, why was it introduced in Magento 1.9.3.3 and was it hardcoded until now. Strictly on security aspects. As has been discussed, it is not just a problem of not compressing an already compressed file, but a more complex, malicious one. If no suitable explanation can be found, the PRs/Issues where they were discussed can be indicated for the beginning.

kiatng:
@ADDISON74  See the original comment here 
https://github.com/OpenMage/magento-lts/blob/fb12d0f2ca70ececdf99186115c25123d1c69f55/app/code/core/Mage/Core/etc/config.xml#L500-L503

```xml
<!-- NOTE: If you turn off images reprocessing, then your upload images process may cause security risks. -->
```
This PR doesn't change what was introduced, if someone uses it in some custom `config.xml`, it will still work for them: either it is turn off or the image quality is set to 100. This takes precedent over the new feature.

The new feature is the ability to set the image quality in backend, which should satisfy everyone . Ref the discussion thread in issue #475, some want to turn it off, some wants to have another value for the image quality. With this PR, you can turn it off by setting the image quality to zero, or set any image quality you want.

kiatng:
Here is the original commit: https://github.com/OpenMage/magento-lts/commit/1172d58d812d5320b7146d4dfb583b6f3c5ba5aa

elidrissidev:
> I think I found the reason: https://web.archive.org/web/20180406222439/https://magento.com/security/patches/supee-9767
> 
> image

Good find!

kiatng:
This makes sense now: https://github.com/OpenMage/magento-lts/commit/6f7586f02cd94b0ae95780c845aca4c74d24f53e
Line 94:
```php
//replace tmp image with re-sampled copy to exclude images with malicious data
```

app/code/core/Mage/Core/etc/system.xml
kiatng: Removed.

elidrissidev (approved):
Looks solid. Suggestion to precise the commit message upon merge.

ADDISON74 (approved):

ADDISON74:
Let's wait a few more opinions before merging.

app/code/core/Mage/Core/Model/File/Validator/Image.php
colinmollenhour, line 90:
```suggestion
                    $imageQuality = Mage::getStoreConfigFlag('general/reprocess_images/active') ? 85 : 0;
```

colinmollenhour:
Just FYI, OpenMage is much less susceptible to attacks using media uploads than Magento because we addressed the "allow_symlinks" option by removing it altogether and adding extra checks to prevent use of "../" in block template paths in https://github.com/OpenMage/magento-lts/pull/442 . So even if a malicious file was uploaded one would not be able to execute it with a template hack as far as we know. Still I think it is a good idea to reprocess these images to be safe but the CVEs referenced are completely irrelevant to OpenMage since it no longer even has an "AllowSymlinks" option and is probably more secure than even when it was on.

colinmollenhour:
So for existing installations:
- XML config present to set active=1 - continue using 100 (I added suggestion to change to 85 just because 100 is so inefficient I can't see anyone really wanting it to be 100)
- XML config present to set active=0 - reprocessing is still disabled
- No XML config present - use new default of 85, user can change via Admin > Security config

Only problem is users that changed their XML will not be able to modify via the UI without removing their XML update so this could be confusing/unexpected..

I propose changing the logic so the default value is empty and if there is a non-empty value for the new config always use it, then check for the legacy config and use it and otherwise default to 85.

sreichel:
How about a RSS notification to reset this value in admin config?

sreichel:
Something I've done for my extensions when config paths changed was to add an update script that also removed orphan entry from core config table.

kiatng:
@colinmollenhour actually described a bug (users not able to modify image quality in backend if deprecated config exists), and I like his fix, which is implemented:

The new config is no longer a required entry,  image quality is now `null`, which is default to 85. Most users, including me, would not have known what value to set, so the hard-coded default value serves us well. For other users, they can set it how they want and the value will be used.

In edge cases where there is a custom XML with the deprecated config, it means it wants to skip the image reprocessing. Otherwise, there is no reason to override the config. This PR will comply with the directive. Now, the same users come across the config in backend, they can set it to zero, to continue skipping image reprocessing, or they can set a value, which will take effect.

Thanks @colinmollenhour for the neat solution.

sreichel:
Mhhh ... i'm not so happy to check for an orphan config entry just for BC.

How about an update script that move value from old to new config path?

- if someone has set it manually in config.xml and saved config, it is stored to DB ...
- optional ... if old value is above 85 (in case of not knowing the bad effects) it could/should reset to 85
- still add RSS notification

kiatng:
> How about an update script that move value from old to new config path?

By "update script", do you mean add a file in `app/code/core/Mage/Core/data/core_setup/data-upgrade-1.6.0.10-1.6.0.11.php`? I'll give it a try.

> still add RSS notification

That I do not know how to do.

colinmollenhour (dismissed):
Looks good to me.

@sreichel The code is only executed when files are uploaded so the overhead of checking a deprecated config value is nil and it's just a few lines of code. This seems like a sufficient solution to me and a database migration is extra coding time but for what added value? Probably 99% of users don't even know about this config option. The ones that disabled it will still get the same behavior but now have the option to set it in the GUI so win-win IMHO.

sreichel:
@colinmollenhour it's not about "overhead". Why to deal with a non-existing config node?

Simple update script ....

```php
<?php
$installer = $this;
$installer->startSetup();

# change config values
$installer->run("
    UPDATE {$this->getTable('core_config_data')}
    SET path = 'admin/security/reprocess_image_quality'
    WHERE path = 'general/reprocess_images/active';
");

$installer->endSetup();
```

Comment/note in config sections seems to be clear ... no need for special rss update ...

colinmollenhour:
> it's not about "overhead"

To me it's mostly a matter of time being well spent including/especially discussion time.. we have a working PR *now*.. and although it isn't complex, it is not quite that simple. The setup script you put in the comment isn't a commit so it can't be approved yet, it doesn't account for the value being set in the XML (XML values don't get saved to the database unless they are also in system.xml, I am sure of this) and the old value is boolean while the new one is 0-100.

sreichel (dismissed):

kiatng:
This statement

> if someone has set it manually in config.xml and saved config, it is stored to DB ...

can be true or false, but in our case, it is false for `general/reprocess_images/active`.

Whereas this one

> XML values don't get saved to the database unless they are also in system.xml

is true, I tested it.

It means the upgrade script needs to do a scan for `general/reprocess_images/active` in all the `config.xml` files, examine the value and do something about it.  Let's not go this route.

ADDISON74:
This is the second PR in which I make comments about the wording of some phrases. They are not discussed but simply ignored. More attention should be paid in this regard because it is what the OM user read.

kiatng:
@ADDISON74 Please comment again the ignored "wording of some phrases". If you can help to review again, I will look into it and incorporate them in this PR. 

Regarding "an explanation in the README", can you help to help to make a PR after this is merged? Basically, the original code make jpeg image files too large due to saving it at 100% quality. This PR reduces it to 85%. The backend config is really cater for edge cases where the users can play with the quality setting. I do not really see a need to explain it in README, I think a note in release commit is sufficient.

ADDISON74:
@kiatng - If you opened this page in a browser you can see my comment above. It is also in "File changed" section.

elidrissidev:
@ADDISON74 I can't any review of yours here. Perhaps you forgot to submit it?

ADDISON74:
In **Files changed** tab, [Review changes] button has a badge. Here is a screenshot:

screenshot-1

On the same page bellow you may find that review:

screenshot-2

In **Conversation** tab, my comment can be found too. See bellow

screenshot-3

My comment is 3 days old. 

In other two merged PR's the same situation. I am starting to wonder if it is still necessary to make comments since they are being ignored.

fballiano:
I learned that when it says “pending” you’re the only one that can see it. Tou have to “submit the review” or “add single comment” 😉

elidrissidev:
Your comments are not being ignored, no one can see them yet because they are still pending, you have to actually submit the review ("Review changes" -> "Submit review") for it to be public.

This is a common confusion with the GitHub, it happened with Fabrizio before too.

ADDISON74 (changes requested):

ADDISON74:
@fballiano - GitHub has an incomprehensible behavior, if I leave a comment it doesn't appear for the others (I logged out to check before). It is only displayed if I request changes. It should have been displayed as long as GitHub explicitly mentions that choosing Comment means _"Submit general feedback without explicit approval"_.  Is it a general feedback but it is displayed only for me? Anyway, I appreciate for pointing me out, I will pay more attention from now on to what is displayed when I am not logged in.

elidrissidev:
It's for batching multiple comments in a single review. But you can always just choose "Add single comment".

ADDISON74:
I used "Add single comment" button but you may see the result above, it was hidden from everyone else. I had to start a review or request for a change. Of the two options, the change request has priority and cancels all previous approvals.

ADDISON74 (approved):

elidrissidev (approved):

PR #2630, opened 2022-09-30
Small improvements to RWD's "toolbar"

fballiano:
At some resolutions the toolbar is shown with the pager on a new row:

Schermata 2022-09-30 alle 14 53 42

So I decided to do a bit of a cleanup of that part of the CSS, starting to use flexbox and removing everything that I could and that was not necessary anymore, the result is this (on a modern browser)

Schermata 2022-09-30 alle 14 09 44

and I still checked that IE8 works:

Schermata 2022-09-30 alle 14 09 32

To compile the CSS I used "compass compile" which, as far as I know, should be the right way to do it. This is the reason why there are quite some differences in the generated CSS.

### DIFF
--- a/skin/frontend/rwd/default/css/madisonisland-ie8.css
+++ b/skin/frontend/rwd/default/css/madisonisland-ie8.css
@@ -1,21 +1,36 @@
-/**
- * OpenMage
- *
- * NOTICE OF LICENSE
- *
- * This source file is subject to the Academic Free License (AFL 3.0)
- * that is bundled with this package in the file LICENSE_AFL.txt.
- * It is also available through the world-wide-web at this URL:
- * http://opensource.org/licenses/afl-3.0.php
- * If you did not receive a copy of the license and are unable to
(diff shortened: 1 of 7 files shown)


ADDISON74:
If it will be merged and taken to v20 we must not lose sight of the fact that in this version the support for IE has been removed. If there is a strict commit for IE modification then it won't be a problem. I have not checked, but such changes should be considered in the future to be easily identified.

Flyingmana:
Is this a big enough issue for v19 to need a fix?

fballiano:
no, but @sreichel wouldn't like it in v20 only :-D

fballiano:
if it was for me, you know, i'll do everything v20

elidrissidev:
I applied your change butunfortunately it seems that it conflicts with "List Mode" controls. You can test it by enabling it in: Configuration -> Catalog -> Frontend -> List Mode.


image

fballiano:
interesting, I'll check it out asap!

Issue #2511, opened 2022-08-29
Sort by position does not work


sreichel:
### Preconditions (*)

1. 1.9.4.x-dev
2. flat product tables enabed

### Steps to reproduce (*)

```
        $collection = Mage::getResourceModel('catalog/product_collection')
            ->addCategoryFilter(Mage::getModel('catalog/category')->load(3))
            ->addAttributeToSort('position')
            ->load();
        var_dump($collection->getFirstItem()->getData());
```

```
Column not found: 1054 Unknown column 'cat_index.position' in 'order clause' ...
```

This works only for flat tables turned off.

There is a PR that fix this, but it does not work for me. See #1815

elidrissidev:
From my testing, this error happens only when flat tables are ON and category filter is either not applied or category not found. Can you try commenting out the 3 lines below and see if your issue is resolved? Then I can create a PR.

https://github.com/OpenMage/magento-lts/blob/32d1591a91b0940661c130d3e2413c11867ff145/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php#L1535-L1537

sreichel:
When i comment these line the error disappears ... not tested anything more.

elidrissidev:
Why do you close this? The fix is not merged yet

ADDISON74:
It is most likely the way in which Sven expresses his dissatisfaction with how this project is managed. Unfortunately, I cannot hide the fact that such actions which are repeated periodically affect me emotionally, especially since I still believe in OpenMage. I think that others are in the same situation.

sreichel:
@elidrissidev i closed it b/c a dont we to see it anymore when viewing https://github.com/issues. Sorry.

PR #2637, opened 2022-10-01
Fix position sort with flat tables when no category filter is applied

elidrissidev:
### Description (*)

This PR will fix the issue when you sort the Product collection by position with flat tables ON, and you either didn't set the Category filter, or the Category doesn't exist (`getId()` returns `null`). The issue was what seems to be a redundant `ORDER BY` clause being added with `cat_index_position` (alias of `catalog_category_product_index.position`), which is only `JOIN`ed if the Category exists. There is already an `ORDER BY` clause being added below the removed code, _with_ the condition of Category filter being present, so this shouldn't cause any change of behavior.

1. Fixes OpenMage/magento-lts#2511

```php
// Make sure you're not on Admin store, as flat tables is only used for frontend
Mage::app()->setCurrentStore(1);

Mage::getResourceModel('catalog/product_collection')
  // Either give it a non-existent category or omit the filter
  ->addCategoryFilter(Mage::getModel('catalog/category')->load(1234))
  ->addAttributeToSort('position')
  ->load();
```

This snipped will throw the below error before this patch:
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'cat_index_position' in 'order clause', query was: SELECT `e`.`entity_id`, `e`.`type_id`, `e`.`attribute_set_id` FROM `catalog_product_flat_1` AS `e` WHERE (e.status = 1) ORDER BY `cat_index_position` ASC, `e`.`entity_id` ASC"
```

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
@@ -1524,9 +1524,6 @@ class Mage_Catalog_Model_Resource_Product_Collection extends Mage_Catalog_Model_
                 $this->getSelect()->order($this->_getAttributeFieldName($attribute) . ' ' . $dir);
                 return $this;
             }
-            if ($this->isEnabledFlat()) {
-                $this->getSelect()->order("cat_index_position {$dir}");
-            }
             // optimize if using cat index
             $filters = $this->_productLimitationFilters;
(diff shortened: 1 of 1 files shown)


luigifab (approved):
For me it's working with a local env, but we don't use sort by position on our store.

fballiano (approved):

fballiano:
merged and v20ed

PR #2641, opened 2022-10-03
Only run workflows when relevant files change

elidrissidev:
### Description (*)

Yesterday I noticed that apart from LGTM, workflows run all the time regardless of changed files in the event. So this PR is going to change that to make them trigger only when relevant files have changed.

1. Only the relevant workflows should run for this PR as an example.

### Questions or comments

I left out Sonar workflow because I'm not sure about that one.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -4,11 +4,11 @@ on:
   # Run on all pushes and on all pull requests.
   # Prevent the build from running when there are only irrelevant changes.
   push:
-    paths-ignore:
-      - '**.md'
+    paths:
+      - '**.php'
   pull_request:
-    paths-ignore:
-      - '**.md'
+    paths:
+      - '**.php'
   # Allow manually triggering the workflow.
   workflow_dispatch:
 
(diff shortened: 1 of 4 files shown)


.github/workflows/phpstan.yml
```diff
@@ -4,11 +4,11 @@ on:
   # Run on all pushes and on all pull requests.
   # Prevent the build from running when there are only irrelevant changes.
   push:
-    paths-ignore:
-      - '**.md'
+    paths:
+      - '**.php'
```
fballiano, line 8: should we run phpstan also when we change phtml files?

elidrissidev: From its configuration files I can only see app/code/* directories included. Maybe someone else can confirm that.

fballiano (approved):

.github/workflows/phpstan.yml
sreichel: > should we run phpstan also when we change phtml files?

No. W/o proper docblocks phpstan can't even resolve $this for block classes.

sreichel (approved):

Issue #2642, opened 2022-10-03
Replace LGTM.com as it's going to shutdown in Dec. 2022


elidrissidev:
From their website:

> LGTM.com will be shut down in December 2022 — we recommend that you use GitHub code scanning instead. Read more in [our blog post on the GitHub blog](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/).

PR #2644, opened 2022-10-03
Create codeql-analysis.yml

Flyingmana:

1. Fixes OpenMage/magento-lts#2642

1. ...
2. ...


ADDISON74:
Can we move forward with this issue? LGTM.com is already shut down.

fballiano:
it seems this PR is already complete

sreichel:
> it seems this PR is already complete

But ... 92 new alerts including 87 high severity security vulnerabilities. Should we fix that first?

fballiano:
where do you see those errors?

sreichel:
https://github.com/OpenMage/magento-lts/runs/8675295733

fballiano:
uff, prototype and tinymce = 99% impossible to solve

Flyingmana:
we dont need to solve this "new errors", maybe put them into an ignore or mark it as allowed to fail

sreichel:
There seems to be a baseline option ... needs some test?!?

Flyingmana:
will look into it

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,84 @@
...
+  push:
+    branches: [ "1.9.4.x", 20.0 ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "1.9.4.x", 20.0 ]
+    paths-ignore:
+      - '**/*.md'
+      - '**/*.txt'
+  schedule:
+    - cron: '33 4 * * 4'
```
sreichel, line 24: Do we need this? Set path to `js` only?

Flyingmana: removed the cron trigger.
We dont have any js files outside the js directory? then makes sense to add it, too

sreichel:
Nice, LGTM.

I'd add it to current workflow later.

.github/workflows/codeql-analysis.yml
fballiano: we've a lot of .js in /skin/

can it also check the inline js in templates?

sreichel: I'd use include `**.*.js` instead of ... to run it only when js files changed.


```
    paths-ignore:
      - '**/*.md'
      - '**/*.txt'
```

ADDISON74:
@Flyingmana - Nice work and super fast.

Flyingmana:
I now got all errors covered with paths-ignore.
Baseline is a bit more complicated I think, as the state of errors is stored in github, where we can mark the alerts as ignored then one by one.

sreichel (approved):

fballiano:
Could we test the js in templates?

Flyingmana:
> Could we test the js in templates?

iam not sure, first it should replace the LGTM what we had before. I dont think it was able to parse the templates.

But it might be able with codeql, although its a bit more work, why I would delay it for a later time

sreichel:
> But it might be able with codeql, although its a bit more work

If we have a working js code coverage  ... :+1:

fballiano (approved):

Issue #2658, opened 2022-10-16
Product flat tables broken, ref #2603


sreichel:
### Preconditions (*)

1. 19.4.x-latest
2. Sample Data installed
3. Flat product tables ON

### Steps to reproduce (*)

1. ge to https://openmage.ddev.site/men/new-arrivals.htm

### Actual result (*)

```
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '1=1 AND ( e.website_id = '1' ) AND ( e.customer_group_id = 0) AND (e.min_pric...' at line 2, query was: SELECT FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 AS `range`, COUNT(*) AS `count` FROM `catalog_product_index_price` AS `e`
 INNER JOIN `catalog_category_product_index` AS `cat_index` ON cat_index.product_id=e.entity_id AND cat_index.store_id=3 AND cat_index.visibility IN(2, 4) AND cat_index.category_id = '14' WHERE 1=1 1=1 AND ( e.website_id = '1' ) AND ( e.customer_group_id = 0) AND (e.min_price IS NOT NULL) GROUP BY FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 ORDER BY FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 ASC

#0 /var/www/html/lib/Varien/Db/Statement/Pdo/Mysql.php(104): Zend_Db_Statement_Pdo->_execute(Array)
#1 /var/www/html/app/code/core/Zend/Db/Statement.php(289): Varien_Db_Statement_Pdo_Mysql->_execute(Array)
#2 /var/www/html/lib/Zend/Db/Adapter/Abstract.php(480): Zend_Db_Statement->execute(Array)
#3 /var/www/html/lib/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('SELECT FLOOR((R...', Array)
#4 /var/www/html/lib/Varien/Db/Adapter/Pdo/Mysql.php(502): Zend_Db_Adapter_Pdo_Abstract->query('SELECT FLOOR((R...', Array)
#5 /var/www/html/lib/Zend/Db/Adapter/Abstract.php(811): Varien_Db_Adapter_Pdo_Mysql->query('SELECT FLOOR((R...', Array)
[...]

PR #2659, opened 2022-10-17
Avoid filtering on product status if flat catalog is enabled

fballiano:
Full explanation is here: https://github.com/OpenMage/magento-lts/issues/2658

The way it was coded was simply wrong, when forcing the `setPart()` the "type of where" gets lost and the system doesn't know if it is a AND or an OR.
This PR rewrites it and avoids the usage of "1=1" which wasn't the best anyway.

1. Fixes OpenMage/magento-lts#2658

Enable flat catalog, reindex, navigate to a category page, no error should be displayed.


elidrissidev:
The description and title are a bit misleading, as the change isn't a part of #2603.

fballiano:
true, changed

PR #2660, opened 2022-10-17
Rewrote the $select->setPart() usage because it looses the "where type"

fballiano:
In this part of code https://github.com/OpenMage/magento-lts/blob/15f83afae2a96a55d7500697fc2edb6a235d4c83/app/code/core/Mage/Catalog/Model/Resource/Layer/Filter/Price.php#L118-L128

There's a weird hack (that 1=1 replacement) that doesn't look very nice but most of all the usage of the `$select->setPart(Zend_Db_Select::WHERE, $wherePart);` is wrong, I'll try to explain.

When doing `$select->where("some condition")` you're actually calling https://github.com/OpenMage/magento-lts/blob/15f83afae2a96a55d7500697fc2edb6a235d4c83/app/code/core/Zend/Db/Select.php#L1031-L1052 where the $bool adds the "AND" where type.

With the current implementation this whole thing gets bypassed and the `$select->setPart(Zend_Db_Select::WHERE, $wherePart);` looses the where type, creating the problem described in https://github.com/OpenMage/magento-lts/issues/2658

This PR rewrites that small part of code removing the 1=1 hack which becomes useless, the code is more correct and looks better IMHO.

Also, you can see in the same file modified by this PR, just a few lines below https://github.com/OpenMage/magento-lts/blob/15f83afae2a96a55d7500697fc2edb6a235d4c83/app/code/core/Mage/Catalog/Model/Resource/Layer/Filter/Price.php#L129-L136 it uses the same writing style that I'm using in this PR.


fballiano:
I think this shouldn’t be closed, it’s bad code what’s in use now

sreichel:
Was closed by github/my PR... just reopen.

sreichel:
> I think this shouldn’t be closed, it’s bad code what’s in use now

What's "bad"?

fballiano:
It should be explain the in pr description

leissbua:
I like it, make me a reviewer and i approve it.

fballiano:
I think you can review it anyway, it won't be a green check but it's still very valuable, 1 green mark and 2 gray it's 100 approved

elidrissidev:
I tested this now but it doesn't resolve the problem completely, I get a stray `AND` in the WHERE clause:
```sql
SELECT FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 AS `range`, COUNT(*) AS `count`
FROM `catalog_product_index_price` AS `e`
INNER JOIN `catalog_category_product_index` AS `cat_index` ON cat_index.product_id=e.entity_id AND cat_index.store_id='1' AND cat_index.visibility IN(3, 4) AND cat_index.category_id = '2'
WHERE (AND (e.entity_id IN(554, 442, 445, 441, 446, 398, 408, 409, 410))) AND ( e.website_id = '1' ) AND ( e.customer_group_id = 0) AND (e.min_price IS NOT NULL) GROUP BY FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 ORDER BY FLOOR((ROUND((e.min_price) * 1, 2)) / 100) + 1 ASC
```

You need to have "Use In Search Results Layered Navigation" set to "Yes" for the `price` attribute to generate this issue.

fballiano:
isn't it possible that that problem comes from another site and it's not related to the modified method? cause, checking what i modified in this PR, it seems weird

leissbua:
I guess i have to agree the ORM is to stupid so the dev choose to add the 1=1 before, so the AND does not become a problem. Need to work on this, I am going to check your solution in a customer project and report back.

leissbua:
But if this happens due to ORM stupidity, and the reported error is continuous, why not just prepend 1=1 to the where part array in general. I check whats going on in the next few days.

elidrissidev:
> isn't it possible that that problem comes from another site and it's not related to the modified method? cause, checking what i modified in this PR, it seems weird

It actually makes sense, since `getPart(Zend_Db_Select::WHERE)` returns an array of conditions including their delimiter (`AND`, `OR`...), and `where()` method doesn't strip them off, so the fix is gonna be more complicated.

fballiano:
I'll try to check it out ASAP, if you've ideas they're obviously more then welcome :-)

boesbo (approved):

PR #2674, opened 2022-11-04
Fix error when calling shell script from browser

Caprico85:
### Description (*)
Running a shell script from browser may cause the PHP warning `Undefined array key "argv"`. This happens because shell/abstract.php tries to read `$_SERVER['argv']` without checking if it exists. `$_SERVER['argv']` is not set when a script is run from browser.

I know running a shell script from browser doesn't make much sense and trying it will result in the message `This script cannot be run from Browser. This is the shell script.`. But nonetheless, the script should exit gracefully and not trigger a PHP warning.

1. Open http://example.com/shell/indexer.php or any other shell script in your browser. Depending on your server settings you will see the warning.


luigifab (dismissed):

sreichel:
Why not

```
if (!isset($_SERVER['argv'])) {
    return $this;
}
```

luigifab (approved):

justinbeaty:
I am not sure empty is correct. What about a custom shell script that extends the abstract class, but has no arguments?

fballiano:
@justinbeaty I created a script and the run() method works correctly also with this patch, so I think the "empty" is not a problem in this context

fballiano (approved):

fballiano:
but actually... htaccess forbids to call shell/*.php from browser... so do we need this?

justinbeaty (approved):

justinbeaty:
I read it more carefully and see that it wouldn't cause any problems. 

Regarding htaccess, I suppose it still doesn't hurt in case someone has a misconfigured server.

fballiano:

PR #2678, opened 2022-11-06
Add buttons to switch between users and roles

luigifab:
### Description

This add two buttons to easily switch between users and roles.

In System / Permissions / Users (before/after):

b1a
b2b

In System / Permissions / Roles (before/after):

b1
b2

OpenMage 20.0.16 / PHP 8.0.25

### Contribution checklist


sreichel:
Like this, but would change the labels.

Users > Mange Users or User settings
Roles > Manage Roles or Role settings

:question:

luigifab:
Yes "Manage Users" and "Manage Roles" to not add new translations.
I already use #1359 (so "Users" and "Roles") and yesterday I forgot that is not merged.

fballiano:
@luigifab what did you decide about the "users/manage users" label? can you rebase to "main" what the label decision is sorted out?

ADDISON74:
I tested this PR and it works as described. No other changes are required. However, **I am against its integration in OpenMage**. Here are my arguments

1 - These buttons do not solve anything. There are just links to Backend pages, which can be accessed as before from the main menu.

2 - The "Roles" and "Users" pages are not accessed often, nor can they even be compared with "Orders", "Manage Products", "Manage Customers". I don't think I have fingers on one hand as many times as I have accessed them in a store over time. Putting those buttons means that it is an effort for the user to go to the main menu, but that means changing his behavior. The idea will be reached that if it is possible here, then it must be changed in other places as well.

3 - This is probably the most solid. Each page in the Backend has buttons that can be used to perform actions strictly related to its content. For example, in the "Manage Products" page I only have the [Add Product] button and nothing else. If I were to allow the presence of these buttons on the two pages, I would violate this rule and it would lead to the situation where all kinds of good or bad changes in the Backend could be proposed that we could face. A Backend page must remain as before strictly with actions related to its content.

kiatng:
Also, ACL. Some users can manage admin users but not roles. So, before rendering the button, need to check ACL?

PR #2687, opened 2022-11-06
Do not calculate tax notifications every time

luigifab:
### Description

On a multi website/store/currencies, a notification can appear about tax:

image

Today, theses notifications are calculated on each page load.
I found with Blackfire this is about 120 ms (this depend of number of stores):

notif

This PR add 3 flags when Tax configuration is saved.
So the block is calculated only if necessary and only if not ignored.

image

### Manual testing scenario

- Create 10+ store views (to see a difference).
- Go to System > Configuration > Tax > Calculation Settings.
- Set the following values (default config): _Total, Shipping Address, Including Tax, Including Tax, Including Tax, After discount, Including Tax, Custom price, Yes_.
- Save the configuration, there is no notice.
- Update _Apply Customer_ Tax to _Before Discount_.
- Save the configuration, there is a notice.
- Ignore/hide the notice.
- Save the configuration without changes, there is a notice.

With the PR or without the PR, the time to generate _Mage_Tax_Block_Adminhtml_Notifications_ does not change.
**But**, with the PR, when the block is ignored or not necessary, time to generate the block is negligible.

### Contribution checklist


sreichel:
Can you please add some information how to reproduce?

luigifab:
I updated the PR description.

sreichel:
Some changes to docblocks should be reverted ... see https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_phpDocumentor.howto.pkg.html#basics.docblocktemplate

luigifab:
Oh, I'm sorry, I was thinking that's a bug!

luigifab:
Sorry, if someone have time to resolve conflicts...
Else we can close it.

fballiano:
I fixed the conflicts, I hope I didn't mess anything up

luigifab:
Yes but this does not work like before. Not sure why. My current solution is to revert 3532 before apply 2687.

PR #2691, opened 2022-11-07
Run github workflows when their configuration files are changed

fballiano:
With https://github.com/OpenMage/magento-lts/pull/2641 we optimized the way workflow run, but we've to run PHPstan workflows when configuration files are changed, this PR should fix it.


sreichel:
Thanks. We should change this for all workflows.

fballiano:
> Thanks. We should change this for all workflows.

right, I'll check it out!

sreichel (approved):

PR #2702, opened 2022-11-10
Added getReviewSummary() to Mage_Catalog_Model_Product

luigifab:
### Description

This PR move the load of `review/review_summary` to the product model.
I don't remember exactly, but I think for us it reduce number of SQL queries in `review_entity_summary`.

OpenMage 20.0.16 / PHP 8.0.25

### Contribution checklist

### DIFF
--- a/.github/workflows/check-files.yml
+++ b/.github/workflows/check-files.yml
@@ -6,6 +6,9 @@ on:
       composer:
         description: "Count changed Composer files"
         value: ${{ jobs.check.outputs.composer }}
+      js:
+        description: "Count changed JS files"
+        value: ${{ jobs.check.outputs.js }}
       php:
         description: "Count changed PHP files"
         value: ${{ jobs.check.outputs.php }}
@@ -42,6 +45,7 @@ jobs:
     runs-on: [ubuntu-latest]
     outputs:
       composer: ${{ steps.changes-composer.outputs.composer }}
+      js: ${{ steps.changes-js.outputs.js }}
       php: ${{ steps.changes-php.outputs.php }}
       xml: ${{ steps.changes-xml.outputs.xml }}
       workflow: ${{ steps.changes-workflow.outputs.workflow }}
@@ -78,6 +82,7 @@ jobs:
             *.php
             **/*.php
             **/*.xml
+            **/*.js
             .github/workflows/**
             **phpcs**
             **php-cs-fixer**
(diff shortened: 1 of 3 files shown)


app/code/core/Mage/Catalog/Model/Product.php
```diff
@@ -2312,4 +2317,18 @@ public function getEvent()
...
+     */
+    public function getReviewSummary($storeId = null)
+    {
+        if (empty($this->_reviewSummary)) {
+            $this->_reviewSummary = Mage::getModel('review/review_summary')
+                ->setStoreId($storeId ?? Mage::app()->getStore()->getId())
+                ->load($this->getId());
+        }
+        return $this->_reviewSummary;
+    }
```
kiatng, line 2333: It makes more sense for me to not touch `getEntitySummary` in `app/code/core/Mage/Review/Model/Review.php`, and if we set `$_reviewSummary = []`, then we can do this:
```php
    public function getReviewSummary($storeId = 0)
    {
       if (!array_key_exists($storeId, $this->_reviewSummary)) {
            Mage::getModel('review/review')->getEntitySummary($this, $storeId);
            $this->_reviewSummary[$storeId] = $this->getRatingSummary();
       }
    }
```

I didn't test the above code.

app/code/core/Mage/Catalog/Model/Product.php
```diff
@@ -340,6 +340,11 @@ class Mage_Catalog_Model_Product extends Mage_Catalog_Model_Abstract
      */
     protected $_calculatePrice = true;
 
+    /**
+     * @var Mage_Review_Model_Review_Summary
+     */
+    protected $_reviewSummary;
```
kiatng, line 346: For multi-stores:
```suggestion
     * @var Mage_Review_Model_Review_Summary[]
     */
    protected $_reviewSummary = [];
```

sreichel: :+1:

app/code/core/Mage/Catalog/Model/Product.php
sreichel: :+1:

fballiano:
@luigifab could you check @kiatng's code?

luigifab:
Yes I will do that.

luigifab:
I remember why I moved the logic, this is because in our product page we have multiple templates that use `getReviewSummary()` data. Because `$this->getProduct()` or `Mage::registry('current_product')` is available everywhere, I did these changes.

You are right for multi store, I adding.

fballiano:
This PR needs to be rebased on "main" and then we should quickly decide to accept it or not.

fballiano:
rebased to main, dunno how it gives these many conflicts but anyway... please check the unresolved conversation so that we can move on with this.

luigifab:
So the merge is ok.

The idea here is to allow to `$product->getReviewSummary()` from multiple template files, by loading the summary only one time, and because we can `Mage::registry('current_product')`, it works. Perhaps I'm tired, but not sure how I can do it in another way.

fballiano (approved):
it seems all requests were taken care of thus I'm approving

kiatng (approved):

PR #2681, opened 2022-11-06
Do not autoload captcha class when disabled

luigifab:
### Description

With Blackfire, I found when captcha is disabled in configuration, a class is autoloaded.
I dislike.

captcha

OpenMage 20.0.16 / PHP 8.0.25

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #2709, opened 2022-11-15
Revert "Do not autoload captcha class when disabled"

fballiano:
Sorry everybody, I made a mistake while merging OpenMage/magento-lts#2681 and I didn't squash the commits, so now the branch doesn't look clean, would it be ok for you all to revert and then re-merge it right after?

Sorry for the inconvenience


elidrissidev:
I say don't bother. Even if you revert it, the previous history will remain the same ;)

ADDISON74 (approved):

tmotyl:
Dont revert. Having a revert and another merge will only add more confusion.

tmotyl (changes requested):
Dont revert if the only concern is the git history

PR #2712, opened 2022-11-16
Removed support for eAccelerator Cache Backend

fballiano:
http://eaccelerator.net doesn't see a release since 2012 and never got support past PHP 5.4, I think it could be safely removed.


sreichel:
> DEPRECATED: This project is deprecated and does not work with anything newer than PHP 5.3. Please use a recent version of PHP with OPcache instead!

sreichel (approved):

elidrissidev (approved):

PR #1379, opened 2021-01-08
Enhancement for categories management and products grids

luigifab:
### Description

It allow to keep current selected tab on store switch, and to keep current category and current store on refresh in categories management. It prevents to add a new category without a parent category.

It add columns to products grid in categories (status, visibility, action).

image

It open product edit in a popin (without header and footer) with the current selected store in categories and in associated products of configurable products. You can also open the edit page in a new tab or in a new window.

image

It allow to check or uncheck products only when clicking on the checkbox or in the td that contain the checkbox for all grids (products in categories ; related, upsell, crosssell, associated (for grouped and configurable products)).

OpenMage 20.0.13 / PHP 7.4.11 - 8.0.11

### Manual testing scenarios

**TEST 1**
Go to Catalog / Categories.

- Try to create a new subcategory without selecting a parent. This is not possible.
- You can check or uncheck products only when clicking on td input.
- New columns "Status", "Visibility" and "Action".
- The edit link opens product edit page in popup (with the current store).
- You can open the [Edit] link in a new browser tab (middle click or right click "Open in a new tab/window").
- Use Name, Price, Status and Visibility values from selected store.

**TEST 2**
Go to Catalog > Categories the select any category.
Go to Category Products tab.

- Select another store with store switcher.  The current tab is not lost.
- Press F5 or refresh the page. The current category and the current store are not lost.
- Uncheck all products then reset filter or filter or sort or page change. After AJAX all products are unchecked.

**TEST 3**
Go to Catalog > Products and choose any configurable product. For Associated Products grid.

- You can check or uncheck products only when clicking on td input.
- The [Edit] link opens product edit page in popup (with the current store)
- You can open the [Edit] link on a new tab (middle click or right click open in a new tab/window)

**TEST 4**
Go to Catalog > Products and choose any product.
For Related, Up-sells and Cross-sells, Associated (for grouped products) grids.

- Uncheck all products then reset filter or filter or sort or page change. After AJAX all products are unchecked.
- Check some products then reset filter or filter or sort or page change. After AJAX selected products are checked.
- You can check check or uncheck products only when clicking on td input.

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2713, opened 2022-11-16
Added new useful columns in products tab in category view

dbachmann:
### Description (*)
Sometimes when you work in the category view in backend it is necessary to edit a product. It was not possible to edit a product from the products tab in category. To improve usability in backend I added two more columns: One with the current product status and the other provides an edit link which opens the desired product in a popup. 

category_view_be

1. Go to categories
2. Open a category 
3. Open products tab
4. Click on edit


matteotestoni (dismissed):

ADDISON74:
@luigifab proposed something similar here with more futures #1379.

PS - today the number of opened PRs decreased bellow 100. Around 25 of them will close opened issues.

fballiano:
fixed conflict

fballiano:
rebased and reformatted arrays to short syntax

fballiano (dismissed):
I've removed the popup window but I think this PR is useful and should be merged

dbachmann:
@fballiano So, when someone wants to edit a product from the category view, it will open in the same window?

fballiano:
you can press "shift" when you click and it opens in a new tab/windw. modern accessibility (as far as I know) says the GUI shouldn't decide the behavior instead of the user.
Also I don't think there are other places where the backend opens new windows, I may be wrong on that but I've never seen it in my daily work.

dbachmann:
@fballiano When you edit a simple product in a configurable product a pop up opens as well ;)

fballiano:
I've removed it while I was rebasing it because that's what I believe, I wouldn't want to have anything open a new window actually

fballiano:
@ADDISON74 what do you think about this one?

ADDISON74:
@fballiano - I have to make a comparison with #1379. It's been a while since I last saw them analyzed. My opinion that both changes are useful. Many times I have added products using the category grid and I have not had the opportunity to see if the product is Enabled or Disabled. In addition, if I wanted to edit it to change the status, I had to open a new window and take several steps. @luigifab's version also brought an mandatory change, the issue with the position column and deselecting the row.

fballiano:
I see, but it seems to me that this specific one could be merged, asking you since you checked a looot of PRs about grids and columns

ADDISON74:
@fballiano - We will limit this PR strictly to the columns and their position in the table. I have identified a few problems that I will present in detail these days.

ADDISON74:
**My comments related to this PR are the following**

1 - The positioning of the columns in this product grid must be the same as the ones in the **Related**, **Up-sells**, **Cross-sells** grids, which can be found in the product editing page. A first requested change, the **Status** column must be placed to be next after the **Name** column.

status

2 - Not only the **Status** must be added as a column in this grid, but also other columns that are absolutely essential when adding products directly to the category page and that make work more efficient. Similar to the grids mentioned in point 1, **Type**, **Attribute Set Name**, **Visibility** are important and must be added. A product should not be searched only by name.

3 - This is an editable grid. If the **[Save]** button is not clicked before the **Edit** link is used, all the changes made in that page, not only in the grid (product addition/removal, position) are lost. In addition, there is no warning icon that you must save your work. As it is now, the voluntary or involuntary click of the **Edit** link loads the product editing page in the same window, at which point all the changes made are lost. I read your comments, the decision to remove the initial code was in a 1 to 1 situation. My opinion is that we should not force the users to hold down the Shift key to open a new window. Firstly, that they don't know about it, secondly, that they don't do it out of error, fatigue. The situation is similar to other cases of editing in a page where changes behind must be kept, such as Associated Products and where the Magento team has taken this into account. Adding this action must follow the existing behavior in OpenMage and this means to open the editing in another window.

4 - In the window that opens, it would be elegant to keep the behavior of the buttons in the case of editing the Associated Products. There are only three buttons in the window, see bellow. I haven't studied how it could be done, but I think it can be solved in a new Settings.php file that is added to the same directory.

buttons

5 - Regarding the **Status** column, the line ```'id' =>``` I haven't found it in other grids. It can be removed.

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
```diff
@@ -83,6 +83,7 @@ protected function _prepareCollection()
                 'category_id=' . (int) $this->getRequest()->getParam('id', 0),
                 'left'
             );
+        $collection->joinAttribute('status', 'catalog_product/status', 'entity_id', null, 'inner');
```
ADDISON74, line 86:
```suggestion
        $collection->joinAttribute('status', 'catalog_product/status', 'entity_id', null, 'inner');
        $collection->joinAttribute('visibility', 'catalog_product/visibility', 'entity_id', null, 'inner');
```

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
```diff
@@ -142,6 +143,34 @@ protected function _prepareColumns()
...
+                    'url'       => [
+                        'base'      => 'adminhtml/catalog_product/edit',
+                        'params'    => ['store' => $this->getRequest()->getParam('store')],
+                    ]
+                ],
+            ],
+            'filter' => false,
+            'sortable' => false,
+            'index' => 'stores',
+        ]);
```
ADDISON74, line 174: I didn't want to edit the file and create a commit. This change makes the grid the same with the ones in **Related**, **Up-sells** and **Cross-sells**. 

For the **Price** column the width was set to 1 ```'width'     => '1',```. It should be removed as I did bellow.

```php
protected function _prepareColumns()
    {
        if (!$this->getCategory()->getProductsReadonly()) {
            $this->addColumn('in_category', [
                'header_css_class' => 'a-center',
                'type'      => 'checkbox',
                'name'      => 'in_category',
                'values'    => $this->_getSelectedProducts(),
                'align'     => 'center',
                'index'     => 'entity_id'
            ]);
        }
        $this->addColumn('entity_id', [
            'header'    => Mage::helper('catalog')->__('ID'),
            'sortable'  => true,
            'width'     => '60',
            'index'     => 'entity_id'
        ]);
        $this->addColumn('name', [
            'header'    => Mage::helper('catalog')->__('Name'),
            'index'     => 'name'
        ]);

        $this->addColumn('type', [
            'header'    => Mage::helper('catalog')->__('Type'),
            'width'     => 100,
            'index'     => 'type_id',
            'type'      => 'options',
            'options'   => Mage::getSingleton('catalog/product_type')->getOptionArray(),
        ]);

        $sets = Mage::getResourceModel('eav/entity_attribute_set_collection')
            ->setEntityTypeFilter(Mage::getModel('catalog/product')->getResource()->getTypeId())
            ->load()
            ->toOptionHash();

        $this->addColumn('set_name', [
            'header'    => Mage::helper('catalog')->__('Attrib. Set Name'),
            'width'     => 130,
            'index'     => 'attribute_set_id',
            'type'      => 'options',
            'options'   => $sets,
        ]);

        $this->addColumn('status', [
            'header'    => Mage::helper('catalog')->__('Status'),
            'width'     => 90,
            'index'     => 'status',
            'type'      => 'options',
            'options'   => Mage::getSingleton('catalog/product_status')->getOptionArray(),
        ]);

        $this->addColumn('visibility', [
            'header'    => Mage::helper('catalog')->__('Visibility'),
            'width'     => 90,
            'index'     => 'visibility',
            'type'      => 'options',
            'options'   => Mage::getSingleton('catalog/product_visibility')->getOptionArray(),
        ]);


        $this->addColumn('sku', [
            'header'    => Mage::helper('catalog')->__('SKU'),
            'width'     => '80',
            'index'     => 'sku'
        ]);
        $this->addColumn('price', [
            'header'    => Mage::helper('catalog')->__('Price'),
            'type'  => 'currency',
            'currency_code' => (string) Mage::getStoreConfig(Mage_Directory_Model_Currency::XML_PATH_CURRENCY_BASE),
            'index'     => 'price'
        ]);
        $this->addColumn('position', [
            'header'    => Mage::helper('catalog')->__('Position'),
            'width'     => '1',
            'type'      => 'number',
            'index'     => 'position',
            'editable'  => !$this->getCategory()->getProductsReadonly()
            //'renderer'  => 'adminhtml/widget_grid_column_renderer_input'
        ]);
        $this->addColumn('action', [
            'header'    => Mage::helper('catalog')->__('Action'),
            'width'     => '50px',
            'type'      => 'action',
            'getter'    => 'getId',
            'actions'   => [
                [
                    'caption'   => Mage::helper('catalog')->__('Edit'),
                    'id'        => 'editlink',
                    'field'     => 'id',
                    'onclick'  => 'event.stopPropagation()',
                    'url'       => [
                        'base'      => 'adminhtml/catalog_product/edit',
                        'params'    => ['store' => $this->getRequest()->getParam('store')],
                    ]
                ],
            ],
            'filter' => false,
            'sortable' => false,
            'index' => 'stores',
        ]);

        return parent::_prepareColumns();
    }
```

The approach for the **Edit** action will be discussed later.

ADDISON74 (changes requested):

ADDISON74:
This should be the **Action** column. ```popup => 1``` enables only three buttons in the window [Close Window], [Reset] and [Save], like for the Associated Products.

```php
        $this->addColumn('action', [
            'header'    => Mage::helper('catalog')->__('Action'),
            'width'     => '50px',
            'type'      => 'action',
            'getter'    => 'getId',
            'actions'   => [
                [
                    'caption'   => Mage::helper('catalog')->__('Edit'),
                    'url'       => [
                        'base'      => 'adminhtml/catalog_product/edit',
                        'params'    => [
                            'store' => $this->getRequest()->getParam('store'),
                            'popup' => 1
                        ],
                    ],
                    'field'     => 'id',
                    'onclick' => 'popWin(this.href,\'_blank\',\'width=1024,height=750,resizable=1,scrollbars=1\');return false;',
                ],
            ],
            'filter' => false,
            'sortable' => false,
            'index' => 'stores',
        ]);
```

There is an issue after saving the product. The bellow window doesn't have the same behavior like the one for Associated Products. The window is not closing automatically and the [Close Window] button it is not working. The error in console is 

```console
Uncaught TypeError: Cannot read properties of undefined (reading 'updateProduct')
    at addProduct (1ae108f468e971e091309d629583f5aa/:133:44)
    at 1ae108f468e971e091309d629583f5aa/:151:1
```

window

ADDISON74:
I have updated this PR. Now the grid has the same format as those in **Related Products / Up-sells / Cross-sells**.

The unsolved problem is still the editing action:

1 - Editing a product must be done in a new window, opened by clicking on the link and having the same behavior as when the associated products are edited.

2 - There must be 3 buttons in the window. After pressing the [Save] button, the window either closes automatically or by pressing the [Close Window] button, exactly the same behavior as when the associated products are edited.

3 - Editing a product should update the grid. For example, when the status of the product is changed by saving, the change does not occur in the grid. A click on other category then coming back to the initial category will make visible the changes.

fballiano:
I don't like the opening of new windows because accessibility mandates that it has to be the user to decide if they want the same window, a new tab or a new window.

but since it's already used in configurable products, I've reverted my change and implemented the popup parameter as requested by @ADDISON74 so have the same behavior (and same window name) of the configurable products management page.

since everything seems to be working fine and there were already quite some tests, if all workflows are green I'll merge it.

fballiano (approved):

fballiano:
actually... no... adding the popup parameter has this behavior now

Screenshot 2024-04-25 alle 19 18 56

fballiano:
this is because the popup parameter calls some specific javascript code for configurables:

Screenshot 2024-04-25 alle 19 20 14

I' ve now patched the js in order to not have the js error when trying to close the windows (tested both here and configurable and both still work) but I wouldn't develop all this new javascript just to refresh the grid, it seems too much for this small functionality

fballiano:
I made the reload work yeah! so all of the points listed by @ADDISON74 should be solved

fballiano (approved):

ADDISON74:
Let me test again to provide you a feedback before approving. Many thanks for your patience in solving that js error.

This PR will be a great new feature.

ADDISON74 (approved):
I checked this PR and all my previous reported issues were resolved. Editing a product opens a new window, there are 3 buttons as in the case of associated products, if the status of the product is changed and the [Save] button is pressed, the status is immediately changed in the grid on the category page. The editing window closes automatically if the button is not pressed. I found no errors in the console. There have been similar paid or free versions in the past, but none offered what was implemented here.

@fballiano - thank you for completing this PR.

fballiano:
one more on the road to zero PRs, let's go!

PR #2725, opened 2022-11-17
Allow to load order, invoice, shipment and creditmemo by increment id

luigifab:
### Description

This PR allow to load order, invoice, shipment and creditmemo by increment id from URL.

Go to _Sales / Orders_, or _Sales / Invoices_, or _Sales / Shipments_, or _Sales / Credit memos_.
Open any item.
Your URL looks like: `.../index.php/admin/admin/sales_order/view/order_id/2/...`
Now replace the order_id by the increment_id, for example: `.../index.php/admin/admin/sales_order/view/order_id/100000002/...`

And, it works!

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Creditmemo.php
+++ b/app/code/core/Mage/Sales/Model/Order/Creditmemo.php
@@ -225,6 +225,25 @@ class Mage_Sales_Model_Order_Creditmemo extends Mage_Sales_Model_Abstract
     }
 
     /**
+     * Load creditmemo by increment id
+     *
+     * @param string $incrementId
+     * @return $this
+     */
+    public function loadByIncrementId($incrementId)
+    {
+        $ids = $this->getCollection()
+            ->addAttributeToFilter('increment_id', $incrementId)
+            ->getAllIds();
+
+        if (!empty($ids)) {
+            reset($ids);
+            $this->load(current($ids));
+        }
+        return $this;
+    }
+
+    /**
      * Retrieve Creditmemo configuration model
      *
      * @return Mage_Sales_Model_Order_Creditmemo_Config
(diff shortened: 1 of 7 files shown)


app/code/core/Mage/Sales/Model/Order/Creditmemo.php
```diff
@@ -224,6 +224,25 @@ protected function _initOldFieldsMap()
...
+    public function loadByIncrementId($incrementId)
+    {
+        $ids = $this->getCollection()
+            ->addAttributeToFilter('increment_id', $incrementId)
+            ->getAllIds();
+
+        if (!empty($ids)) {
+            reset($ids);
+            $this->load(current($ids));
+        }
```
kiatng, line 242:
```suggestion
        return $this->load($incrementId, 'increment_id');
```

luigifab: I copied/pasted _Mage_Sales_Model_Order_Invoice_, but why not!

sreichel:
> Now replace the order_id by the increment_id, for example: .../index.php/admin/admin/sales_order/view/order_id/100000002/...

Also change the parameter to `increment_id`?

luigifab:
For #2726 I can't.

sreichel:
When the url paramenter is called order_id and dont expect to pass different values to it.

luigifab:
Yes, I did that for customer support.

When customers contact us, they use **increment_id**, so our customer agents go to _Sales / Order_, then fill the _increment_id_, then press enter or filter button, then **wait**, then click on the order to open it. This is time consuming.

With this change, they can keep one tab with any order, and update the `order_id` parameter with any `increment_id`. Some agents prefer a search bar, I also added it.

Because this is a simple change, I think/hope that it will don't break anything, so I share it.

Yes it's better for dev people to use `order_id|invoice_id|shipment_id|creditmemo_id` parameter (`entity_id` in `sales_flat_[order|invoice|shipment|creditmemo]_grid`) and another `increment_id` parameter (`increment_id` in `sales_flat_[order|invoice|shipment|creditmemo]_grid`); but for users, it's easier to only update the parameter value.

kiatng (approved):
It's a useful hack.

fballiano:
I don't know about this, I see the customer service point but I kinda not like the idea of loading something by some attribute when clearly the parameter name shows something else...

PR #2729, opened 2022-11-17
Fixed "Credit Memo not created when refund issued by merchant"

luigifab:
### Description

Sometimes, the total of the order and the total of the refund is a little different, +/- 0.001, so when an admin create a refund from PayPal website (or another payment method), we have:

```
IPN "Refunded". Refund issued by merchant. Registered notification about refunded amount of -61.79 €. 
Transaction ID: "xxxx". Credit Memo has not been created. Please create offline Credit Memo.
```

For this example the order total is 61.79:
image

<del>This PR require **php-bcmath**. perhaps we can replace `bccomp` with another php code to not require on `bcmath`.</del>

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Payment.php
+++ b/app/code/core/Mage/Sales/Model/Order/Payment.php
@@ -806,7 +806,7 @@ class Mage_Sales_Model_Order_Payment extends Mage_Payment_Model_Info
             $amount = $amountRefundLeft;
         }
 
-        if ($amount != $baseGrandTotal) {
+        if (bccomp($amount, $baseGrandTotal, 2) != 0) {
             $transaction = new Varien_Object(['txn_id' => $this->getTransactionId()]);
             Mage::dispatchEvent('sales_html_txn_id', ['transaction' => $transaction, 'payment' => $this]);
             $transactionId = $transaction->getHtmlTxnId() ? $transaction->getHtmlTxnId() : $transaction->getTxnId();
--- a/app/code/core/Mage/Sales/Model/Order/Payment.php
+++ b/app/code/core/Mage/Sales/Model/Order/Payment.php
@@ -806,7 +806,7 @@ class Mage_Sales_Model_Order_Payment extends Mage_Payment_Model_Info
             $amount = $amountRefundLeft;
         }
 
-        if (bccomp($amount, $baseGrandTotal, 2) != 0) {
+        if (Mage::helper('core')->getExactDivision($amount, $baseGrandTotal) != 0) {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Sales/Model/Order/Payment.php
```diff
@@ -806,7 +806,7 @@ public function registerRefundNotification($amount)
             $amount = $amountRefundLeft;
         }
 
-        if ($amount != $baseGrandTotal) {
+        if (bccomp($amount, $baseGrandTotal, 2) != 0) {
```
elidrissidev, line 809: I think using `Mage::helper('core')->getExactDivision($amount, $baseGrandTotal) === 0` will work here, as it also has a precision of 2 decimal digits.

My question is, the values in the DB have a precision of 4 decimal digits, so is it safe to assume that no one cares about the last two?

kyrena: Moreover some currencies use [more than 2 digits](https://docs.adyen.com/development-resources/currency-codes/), for example _Jordanian Dinar / JOD_.
Not sure if Magento round every currencies to 2 digits or not.

`getExactDivision` give the same results:
```php
echo "getExactDivision\n";
var_dump(Mage::helper('core')->getExactDivision(10.01, 10.01 + 0.01) === 0);
var_dump(Mage::helper('core')->getExactDivision(10.01, 10.01 + 0.001) === 0);
var_dump(Mage::helper('core')->getExactDivision(10.01, 10.01 + 0.0001) === 0);
var_dump(Mage::helper('core')->getExactDivision(10.01, 10.01 + 0.00001) === 0);

echo "bccomp\n";
var_dump(bccomp(10.01, 10.01 + 0.01, 2) == 0);
var_dump(bccomp(10.01, 10.01 + 0.001, 2) == 0);
var_dump(bccomp(10.01, 10.01 + 0.0001, 2) == 0);
var_dump(bccomp(10.01, 10.01 + 0.00001, 2) == 0);
```

kiatng (approved):

app/code/core/Mage/Sales/Model/Order/Payment.php
kiatng: @kyrena Beware on strict type comparison: https://onlinephp.io/c/4af76
```php
$var = [0 === 0.0, 0 == 0.0];
var_dump($var);
```
Result:
```
array(2) {
  [0]=>
  bool(false)
  [1]=>
  bool(true)
}
```

fballiano (approved):

PR #2730, opened 2022-11-17
Remove mysql4

luigifab:
### Description

This PR will hurt our feelings 💣 because it remove 99% of *Mysql4* PHP class and the *deprecatedNode* XML node.
This will break 💥 many "outdated" modules.

There is a migration script: `php shell/remove-mysql4.php`
It will update all occurrences of mysql4 classes.

This is #1470 (I force pushed the branch but it was not updated).
Attached a non 

### Contribution checklist


fballiano:
we've to rebase this on the `next` branch since it's a breaking change, then I'd like to try to merge it :-)

shell/rename-mysql4-class-to-resource.php
```diff
@@ -0,0 +1,573 @@
+<?php
+
+// DO NOT RUN DIRECTLY IN YOUR PRODUCTION ENVIRONMENT!
+// This script is distributed in the hope that it will be useful, but without any warranty.
+// See https://github.com/OpenMage/magento-lts/pull/2730 for more information
```
fballiano, line 5: I'd have the normal openmage license block, but actually I wouldn't have this file directly on the repo, because it's a one-use only. I'd maybe create a repo to be added to the dev-meta package or maybe even just a gist linked in the documentation?

ADDISON74: When this PR will be approved, there must be a repository (it would be best in OpenMage) where those interested can download and use it with their extensions. A README with warnings can also be created in that repository.

luigifab: Ok for the copyright notice, but this is not a one use script, you can reuse it each time you install a new module.

fballiano: ahh ok ok sorry, I thought it was for us and not really for the end users, my bad.

kiatng (approved):

fballiano:
let's go, but we'll need for sure to add something in the README

fballiano (approved):

ADDISON74:
It is mandatory to mention this change in the README file.

If I remember correctly, there was a script that looks for files that contained certain strings of characters and modifies them. A quick solution is needed for everything that exists outside of OpenMage. I have a lot of extensions that need modification.

PR #2732, opened 2022-11-17
Do not load root category for isCategoryActive

luigifab:
### Description

With Blackfire, I found that when we call `isCategoryActive`, the root category is loaded.

Not sure why, but I have disabled that two months ago.
Perhaps it's not a good idea, but for me it's working.

### Contribution checklist


kiatng:
I also think it's not a good idea, this PR will mess up `getCurrentCategory()`, and possibly make some developers unhappy.

fballiano:
mmm it seems to me the original code is correct, at least the getCurrentCategory() method.

I mean, how are you calling isCategoryActive() and the category object is not loaded?

luigifab:
For example on home page, on cms page, on login page, on cart page...
Add `echo 'load root';` on line 159 to see this text on your frontend.

fballiano:
closing because of https://github.com/OpenMage/magento-lts/pull/2732#issuecomment-1319495873

PR #2736, opened 2022-11-17
Allow soap v2 without rewrites

luigifab:
### Description

This PR add the possibility to access API SOAP v2 without server rewrites, because:

- this URL work: _http://.../openmage/public/api-secret.php/soap/?wsdl_
- this URL do not work: _http://.../openmage/public/api-secret.php/v2_soap/?wsdl_

### Contribution checklist

### DIFF
--- a/api.php
+++ b/api.php
@@ -52,6 +52,12 @@ Mage::app()->loadAreaPart(Mage_Core_Model_App_Area::AREA_ADMINHTML, Mage_Core_Mo
 // query parameter "type" is set by .htaccess rewrite rule
 $apiAlias = Mage::app()->getRequest()->getParam('type');
 
+// allow soap v2 without rewrites
+if (empty($apiAlias) && (stripos(getenv('REQUEST_URI'), '/v2_soap/') !== false)) {
+    $apiAlias = 'v2_soap';
+    $_GET['type'] = $apiAlias;
+}
+
 // check request could be processed by API2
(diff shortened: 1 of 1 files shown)


fballiano (approved):
I see why this could be useful

Flyingmana:
why is it working for `soap` but not `v2_soap`? might be better to handle it on a different place. Also I wonder if we should check if any APIs get exposed which were not previously.

elidrissidev:
There is some confusion here. First of all, the second URL in the description (`/api-secret.php/v2_soap/?wsdl`) does not work because `api.php` accepts the type from the query parameter `?type=<adapter>`. The reason why you get an XML response from the first URL (`/api-secret.php/soap/?wsdl`) is because the `soap` adapter is the default if no type is provided.

Here is a list of the available `type`s (`v2_soap` is only an alias to `soap_v2`):

https://github.com/OpenMage/magento-lts/blob/ebabcf8499dccef5d2e9241a3df27e00535e6e8a/app/code/core/Mage/Api/etc/api.xml#L25-L70

I try connecting with the URL `http://om19.localhost/api.php/?type=soap_v2&wsdl=1` and it works fine, therefore I think this change might not be necessary.

fballiano:
Wow great find @elidrissidev

sreichel:
ping @luigifab

luigifab:
So as you want. @elidrissidev is right.

PR #2737, opened 2022-11-17
Removed onmouseover/onmouseout from adminhtml menu

luigifab:
### Description 

This simply remove onmouseover/onmouseout attributes from adminhtml main menu.

You must replicate the changes to your **custom** admin theme.

**PLEASE NOTE: This will break custom admin themes!**

### Contribution checklist


fballiano:
I like these PRs but until we get rid of v19 I'm a bit scared of merging something like this in v20 because of merging conflicts for later developments :-\

skin/adminhtml/default/openmage/menu.css
```diff
@@ -47,29 +47,29 @@
   text-align: left;
   float: left;
 }
-#nav li.over {
+#nav li:hover {
   z-index: 999;
 }
-#nav li.over a {
+#nav li:hover a {
```
fballiano, line 53: skin/adminhtml/default/openmage/scss/menu.css should also be updated :-)
(or removed?)

luigifab: Yes, true, I forget...
I'm not sure how scss works, and how to rebuild css files from scss.

ADDISON74:
This PR does not solve any existing issue.

luigifab:
Yes, it will create new bugs :)

fballiano:
I've rebased it for "next" but we need to fix the scss before merging it.

Also, we could decide to remove the scss completely...

luigifab:
[troll]can I remove the openmage admin theme?[/troll]

fballiano:
hiihihihih

skin/adminhtml/default/openmage/menu.css
luigifab: How can I rebuild the CSS from SCSS?

fballiano: it was generated with a tool named "compass" (if i'm not mistaken) which is meeega old, so in my opinion, since it's v21, we should remove the SCSS files and call it a day

ADDISON74: The existence of an SCSS file is good in my opinion, it is much easier to replace a value and have it be used in many places than to look for the value and replace it. Often the results are not the expected ones. Generating a css from scss is now very easy in IDEs.

sreichel: > How can I rebuild the CSS from SCSS?

With phpstorm just follow this: https://www.jetbrains.com/help/phpstorm/transpiling-sass-less-and-scss-to-css.html#install_sass_scss

> it was generated with a tool named "compass"

No, Compass was just an scss compiler - and its outdated now. Keep thar SCSS files and just use another tool to gernate CSS.

fballiano: that's what I said 🤦🏻‍♂️

sreichel: Um, you wanted to remove that SCSS file?

fballiano: @ADDISON74 with "css custom properties" we could do the same without the compiling process, modern css are almost as advanced as the preprocessors but without the need for a preprocessor. ;-)

@sreichel yes sven, you're the best and always right. please continue this way.

luigifab: ok! css regenerated, I think, not sure why `#xyz` was replaced by `hsl()`.

fballiano (approved):
works fine :-)

skin/adminhtml/default/openmage/menu.css
ADDISON74: "css custom properties" represents only a small part taken by CSS from SASS, which in my opinion will remain extremely powerful for a long time to come. We can take a look to CSS frameworks like Bootstrap which use SASS. I have not analyzed the scss file that is in OpenMage, if only root properties are needed then it is a solution, but SASS goes further with functions, mixins and many others.

fballiano:
I think this PR should be merged, we have to get rid of all this ancient and unneeded  javascript a step at a time and v21 is the perfect branch for this.

fballiano:
If I'm not mistaken PHPStan fails for unrelated reasons (we didn't fix all of the problems on the "next" branch) but I think this PR should be merged. nowadays we do not need javascript to have an hover state on a menu :-)

ADDISON74:
I analyzed this PR and I think it should be introduced in the main branch. It is unprofessional to see nowadays classes like "over" when they can be solved with CSS pseudo-classes. I don't think that the impact in the main branch will create any trouble.

fballiano:
I would be of the same opinion but I also think we should stick with the formal definition and since this is a breaking change I'd keep it in next. Our RFC gives us the possibility to release 2 major per year, we coould release v19, v20 and v21 almost at the same time, this would make v21 become the standard and we could move on happily :-)

ADDISON74 (approved):

PR #2758, opened 2022-11-24
SQL error calling class Mage_Bundle_Model_Resource_Price_Index method getCustomOptions

matteorossi-thespacesm:
### Description (*)

Modify $bind variable to logic to address difference in price scope 

call class Mage_Bundle_Model_Resource_Price_Index method getCustomOptions with price global scope

the following sql error appears:
```
SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens, query was: SELECT `option_table`.`option_id`, `option_table`.`is_require`, `option_table`.`type`, `price_table`.`option_price_id` AS `value_id`, `price_table`.`price`, `price_table`.`price_type` FROM `catalog_product_option` AS `option_table`
           INNER JOIN `catalog_product_option_price` AS `price_table` ON option_table.option_id = price_table.option_id AND price_table.store_id = 0 WHERE (option_table.product_id=:product_id)
```

### DIFF
--- a/app/code/core/Mage/Bundle/Model/Resource/Price/Index.php
+++ b/app/code/core/Mage/Bundle/Model/Resource/Price/Index.php
@@ -544,11 +544,10 @@ class Mage_Bundle_Model_Resource_Price_Index extends Mage_Core_Model_Resource_Db
         $price   = $this->_getAttribute('price');
         $adapter = $this->_getReadAdapter();
 
-        $bind = [
-            ':product_id' => $productId,
-            ':store_id'   => $store->getId(),
-
-        ];
+        $bind = [':product_id' => $productId];
+        if (!$price->isScopeGlobal()) {
+            $bind[':store_id'] =  $store->getId();
+        }
         $select = $adapter->select()
             ->from(
                 ['option_table' => $this->getTable('catalog/product_option')],


sreichel:
Can you please provide some test code to reproduce?

kiatng:
I tested it, and I found additional bug. First, to replicate this PR reported bug

> SQLSTATE[HY093]: Invalid parameter number: number of bound variables does not match number of tokens

In system configuration: Catalog > Price > set Catalog Price Scope to _Global_. Then simply call this `Mage::getResourceSingleton('bundle/price_index')->getCustomOptions(1, Mage::app()->getWebsite())`

For the second bug:

> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'price_store_table.option_price_id' in 'field list', query was: SELECT `option_table`.`option_id`, `option_table`.`is_require`, `option_table`.`type`, IF(price_store_table.option_price_id IS NOT NULL, price_store_table.option_price_id, price_global_table.option_price_id) AS `value_id`, IF(price_store_table.price IS NOT NULL, price_store_table.price, price_global_table.price) AS `price`, IF(price_store_table.price_type IS NOT NULL, price_store_table.price_type, price_global_table.price_type) AS `price_type` FROM `catalog_product_option` AS `option_table`
 INNER JOIN `catalog_product_option_type_value` AS `type_table` ON option_table.option_id=type_table.option_id
 INNER JOIN `catalog_product_option_type_price` AS `price_global_table` ON type_table.option_type_id=price_global_table.option_type_id AND price_global_table.store_id=0
 LEFT JOIN `catalog_product_option_type_price` AS `price_store_table` ON type_table.option_type_id=price_store_table.option_type_id AND price_store_table.store_id=:store_id WHERE (option_table.product_id=:product_id)

In system configuration: Catalog > Price > set Catalog Price Scope to _Website_. Then simply call this `Mage::getResourceSingleton('bundle/price_index')->getCustomOptions(1, Mage::app()->getWebsite())`

app/code/core/Mage/Bundle/Model/Resource/Price/Index.php
```diff
@@ -551,11 +551,10 @@ public function getCustomOptions($productId, Mage_Core_Model_Website $website)
...
 
-        $bind = [
-            ':product_id' => $productId,
-            ':store_id'   => $store->getId(),
-
-        ];
+        $bind = [':product_id' => $productId];
+        if (!$price->isScopeGlobal()) {
+            $bind[':store_id'] =  $store->getId();
+        }
```
kiatng, line 557:
```suggestion
       $bind = [':product_id' => $productId];
```
Then at line 573:
```php
        } else { //line 573
            $bind[':store_id'] =  $store->getId(); // insert here
            $valueIdCond = $adapter->getCheckSql(
```

kiatng:
To fix the second bug:

At line 645: https://github.com/OpenMage/magento-lts/blob/a559a5572b609f49d2154aa2c0b9ecede257fda8/app/code/core/Mage/Bundle/Model/Resource/Price/Index.php#L644-L648

insert the following:
```php
            $valueIdCond = $adapter->getCheckSql(
                'price_store_table.option_type_price_id IS NOT NULL',
                'price_store_table.option_type_price_id',
                'price_global_table.option_type_price_id'
            );
```

sreichel:
Mhhh, i can reproduce both errors, but ...

... my price attribute has value `2` for `is_global` in `catalog_eav_attribute` - like some others. So i always see 2nd error until i set `is_global` to `1`.

(btw: in manage attributes pages `price` is always, regardless of config?)

kiatng:
I'm not sure but there are 2 odd things in `getCustomOptions()`: 

1. it always throws an SQL exception, probably since M1.6 (I tested by replacing the file from M1.6)
2. the price scope `is_global` is taken from the attribute but how about this https://github.com/OpenMage/magento-lts/blob/879caab7a16eb335497b335725498b67da6c0a58/app/code/core/Mage/Catalog/Helper/Data.php#L228-L241

What do we do?

fballiano:
setting this to draft because of the unanswered questions/comments

fballiano:
I've been quickly checkin this PR but it seems to me that the Mage_Bundle_Model_Resource_Price_Index class isn't even involved in any reindexing (and I've a bundle item in my test store) because Mage_Bundle_Model_Resource_Indexer_Price is used instead.

@kiatng what do you think? if this is the case should we mark something as deprecated and move on?

kiatng:
+1 to mark `Mage_Bundle_Model_Resource_Price_Index` as deprecated.

kiatng:
Tracing the code to see where the method is called led to this:

https://github.com/OpenMage/magento-lts/blob/e56ab9012310ac7a18c3422d44b846bc4ec00377/app/code/core/Mage/Bundle/Model/Observer.php#L265-L300

There are 2 deprecated methods: `catalogProductLoadAfte()` and `catalogIndexPlainReindexAfter()`. The latter is still being called:

https://github.com/OpenMage/magento-lts/blob/e56ab9012310ac7a18c3422d44b846bc4ec00377/app/code/core/Mage/Bundle/etc/config.xml#L149-L154

Unless this observer is removed, it doesn't seem the method including `getCustomOptions()` and other methods involved can be deprecated.

fballiano:
weird, I've live stores with thousands of bundle products and this error never triggered.

this catalogIndexPlainReindexAfter() seems to be called by event catalogindex_plain_reindex_after, which is dispatched by Mage_CatalogIndex_Model_Indexer::_afterPlainReindex(), called by Mage_CatalogIndex_Model_Indexer::plainReindex().

this plainReindex() is referenced many times, i started to check some of them, in various observers and seems that none is actually referenced in the xmls or it never gets called........................

fballiano:
I still think that all of this code never gets called and there's probably some stuff to delete.

but I don't want to leave PRs hanging foverver, if we can't find our how to trigger the error (apart from directly calling the method) then let's deprecate all the code or close the PR.

fballiano:
This also would need to be rebased to main anyway

sreichel:
I was able to reproduce this. Should not be closed.

fballiano:
how?

fballiano:
@matteorossi-thespacesm @sreichel can you provide info on how to replicate (without directly calling the method)?

sreichel:
I can reproduced it only with calling method directly.

fballiano:
that's what i'm saying, it seems old code which should be deprecated since I can't find any way that this is called by the core

fballiano:
closing per https://github.com/OpenMage/magento-lts/discussions/3405#discussion-5437233

PR #2774, opened 2022-12-02
Allow to copy ACL and config path

luigifab:
### Description

A new feature!

Bellow click on the link, and paste the config path where you want!
You can disable it from System > Configuration > Admin > Admin theme _(admin/design/copy_path)_.

Tested with Firefox 108, Chrome 108, Opera 94.
`navigator.clipboard.writeText` require Firefox 63+, Chrome 66+, Edge 79+, Opera 53+, Safari 13.1+ ([mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText)).

**1/ Go to _System / Configuration_:**

before/after
image
image

**2/ Go to _System / Web Services / SOAP Roles_, add or edit any role:**

before/after
image
image

**3/ Go to _System / Permissions / Roles_, add or edit any role:**

before/after
image
image

### Warning

Not complete for  _System / Web Services / REST Roles_, I'm not sure how them works.

### Contribution checklist


sreichel:
I like it, but do we need this in production mode?

Change

```
Mage::getIsDeveloperMode() || Mage::helper('admin/variable')->isPathAllowed($element->getPath())
```

to

```
Mage::getIsDeveloperMode() && Mage::helper('admin/variable')->isPathAllowed($element->getPath())
```

What do think?

sreichel:
Do we have to adjust openmage-theme (SCSS, CSS) too?

luigifab:
I don't tested with openmage theme.

For production mode, normal admin can use only `isPathAllowed` in cms blocks, cms pages, wysiwyg products and categories attributes, transactional emails. In dev mode, developers can use all path in PHP files.

kyrena:
There was a module for Magento 1.4-9? that add a red icon image (for default configuration) when configuration is updated by store view. Is someone know where I can find it?

sreichel:
> There was a module for Magento 1.4-9? that add a red icon image [ image ](https://raw.githubusercontent.com/OpenMage/magento-lts/1.9.4.x/skin/adminhtml/default/default/images/error_msg_icon.gif) [ ](https://raw.githubusercontent.com/OpenMage/magento-lts/1.9.4.x/skin/adminhtml/default/default/images/error_msg_icon.gif) (for default configuration) when configuration is modified by store view. Is someone know where I can find it?

I know that extension, but cant find it (was it a hackathon project?) There is a similar one: https://github.com/ericthehacker/magento-configscopehints

Flyingmana:
good idea.
Can we put it behind a config value? Iam worried not everyone might like the potentially long extra text.
also please change it to the branch for a later version (its a new feature)
And I wonder if we should make it more flexible, but thats something for a later iteration

sreichel:
good idea.

- we could also use an icin instead of text
- duplicated js code `javascript:navigator.clipboard.writeText` could move to helper
- replace inline style `style="font-style:italic` with css

luigifab:
I think that I have added all comments to the PR (except the red icon). It's not necessary to update openmage theme.
Do I need to change the branch to 20?

sreichel:
Think v19 is correct. Will test it later.

luigifab:
Yes? No? To know if I try to add the red mark or not.

kiatng:
So I installed this PR:
image

But I don't see effects:
image

Any idea what I miss.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1055,9 +1054,17 @@
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
                         </loading_timeout>
+                        <copy_path translate="label">
+                            <label>Allow to copy/paste roles and configuration path</label>
+                            <frontend_type>select</frontend_type>
+                            <source_model>adminhtml/system_config_source_yesno</source_model>
```
kiatng, line 1060: Instead of yes/no:
1. allow in developer mode  only (enabled for developer mode)
2. always allow (enabled)
3. always do not allow (disabled)

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1055,9 +1054,17 @@
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
                         </loading_timeout>
+                        <copy_path translate="label">
+                            <label>Allow to copy/paste roles and configuration path</label>
```
kiatng, line 1058: <label>Enable resource path hints</label>

fballiano:
why is this feature behind "developer mode"? I don't see why normal admin shouldn't be able to copy ACLs from a role to another no?

luigifab:
When you are a normal admin user, you can use only allowed config path in email, cms and pages, so this PR display copy links only for allowed config paths.

But it's true that if you are a normal admin user with access to all admin features, you can add any config paths to allowed list.

This PR display all copy links only in dev mode because I supposed that there are useless when you are using the backend as a normal user, not as a dev user.

empiricompany:
I've checked out the PR, but the new feature doesn't show up in every section of system/config and in permissions. 
Also, when clicking, the path is not copied (Firefox)

demo
if i click on config nothing is copied

Nothing appears here:
demo2
roles

 I've disabled all caches and also re-logged in to the backend

app/code/core/Mage/Adminhtml/Block/Api/Tab/Rolesedit.php
```diff
@@ -77,6 +77,10 @@ protected function _getNodeJson($node, $level = 0)
             $item['sort_order'] = isset($node->sort_order) ? (string)$node->sort_order : 0;
             $item['id']  = (string)$node->attributes()->aclpath;
 
+            if (Mage::getIsDeveloperMode() && Mage::getStoreConfigFlag('admin/design/copy_path')) {
+                $item['text'] .= '</a> <a href="javascript:copyText(\'' . $item['id'] . '\');" class="copycnf">(' . $item['id'] . ')';
```
kiatng, line 81: Function name conflict, see PR #3960

js/mage/adminhtml/tools.js
```diff
@@ -11,6 +11,11 @@
  * @copyright   Copyright (c) 2019-2023 The OpenMage Contributors (https://www.openmage.org)
  * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
+
+function copyText(txt) {
```
kiatng, line 15: Function name conflict, see PR #3960

kiatng:
I agree with @fballiano, this PR needs to incorporate what was done in PR #3960.

luigifab:
Yes, but I can't do it for now.

ADDISON74:
@luigifab - any reasons for closing this PR?

PR #2777, opened 2022-12-03
Replaced Mysql4 classes in comments

sreichel:
### Description (*)

- Replaced Mysql4 classes in comments
- moved `Mage_Core_Model_Mysql4_Design_Theme_Collection` (only one mysql4 class w/o resource class?)


sreichel:
Mhhh. dont know why this error shows up only here. These files are ignored per default and cant reproduce it locally.

kiatng:
> Mhhh. dont know why this error shows up only here. These files are ignored per default and cant reproduce it locally.

Why not skip this [lib/Varien/Directory/Collection.php](https://github.com/OpenMage/magento-lts/pull/2777/files#diff-111436c54e601cbe20094e4dc8d4c039ef00c48c3370beecefff5d610ce7d6c7) and see if the error goes away. That file contains a few more issues that need fixing.

sreichel:
I already ignored for this error ... https://github.com/OpenMage/magento-lts/blob/1.9.4.x/phpstan.dist.neon#L30-L32

But it does not work (only for this PR!)

kiatng:
How about remove the file from this PR and fix it later? Would that work?

sreichel:
It could work, but then i prefer to fix it and wait with this PR.

kiatng (dismissed):

fballiano:
auch, conflict

ADDISON74:
I am not a specialist in using PHPStan. If we take a look in browser to any PR in "Files changed" section we will see errors generated by PHPStan used by GitHub. If anyone will check them with better knowledge it will be great to fix them.

sreichel:
@ADDISON74 please ignore this as long it is marked as draft.

Short explain. Some Varien classes have/had had incompatible interfaces so PHPstan was not able to check them. For this reason i've ignored them PHPstan config, but this does not work for this PR for some unknown reason. I'd tried to fix the incompatible interfaces and no the files are processed. The errors will be fixed in next commits.

PR #2782, opened 2022-12-04
Enable module Mage_Backup

eneiasramos:

1. ...
2. ...

### DIFF
--- a/app/etc/modules/Mage_Backup.xml
+++ b/app/etc/modules/Mage_Backup.xml
@@ -23,7 +23,7 @@
 <config>
     <modules>
         <Mage_Backup>
-            <active>false</active>
+            <active>true</active>
             <codePool>core</codePool>
             <depends>
                 <Mage_Core/>


sreichel:
Why? Mage_Backup has been disabled for some reasons and is only kept for users, that really want to use it. (See OpenMage/magento-lts#2259)

eneiasramos:
@sreichel A lot of people still use this functionality.

fballiano:
Totally against and already discussed

ADDISON74:
We discussed about this module here https://github.com/OpenMage/magento-lts/discussions/2024 and I presented the reasons why the Magento team disabled it in the past.

Here a PR was created to remove the module which was closed after discussions https://github.com/OpenMage/magento-lts/pull/2259.

As long as it is deactivated it could not create trouble for anyone. Whoever wants to use the module can activate it knowing that they may expose themselves to certain risks, especially those related to data protection. However, the Magento team's message was lacking in details, just a statement and that's it. Rather, the issues in this module, which are still unfixed there, would worry me more, as at first glance it does its job according to the user manual.

Although I never used it (but I did intensive tests to find out issues), I was for its preservation considering an average user who prefers this solution instead of some exports from the database (from a UI, from the command line). I also considered that some functional extensions could use classes from this module to implement various tasks.

In conclusion, I agree with the closure of this PR for the stated reasons.

PR #2783, opened 2022-12-04
Checkout address street ID same as customer

eneiasramos:
The idea is to standardize the checkout street IDs equal to the customer for address autocompletion routines.

1. ...
2. ...


elidrissidev:
It's worth noting that some third-party extensions are using these IDs in their code/themes, so this might be breaking.

sreichel:
Agree with @elidrissidev. I like standardized code, but this could break 3rd party-code for no benefit.

Issue #2781, opened 2022-12-04
Credis symlinks wrong?


sreichel:
### Preconditions (*)

1. latest 19.4.x (with composer modules added)

### Steps to reproduce (*)

1. added this to `composer.json`
```
    "magento-deploystrategy": "copy",
    "magento-deploystrategy-dev": "symlink",
    "magento-force": true,
```
2. run `composer install` ... symlinks for CM_ are created
3. delete `vendor`
4. run `composer install --no-dev`

### Expected result (*)

1. symlinks should be replaced by files

### Actual result (*)

1. Credis copy fails

![credis](https://user-images.githubusercontent.com/5022236/205515035-e1c7061f-0cf0-4cc1-96c6-63dac532e3e7.PNG)
![credis2](https://user-images.githubusercontent.com/5022236/205515259-b67630fd-d3f4-4cdd-b1e3-596f0408cd5e.PNG)

fballiano:
when I do the `composer install --no-dev` I get:

```
In Copy.php line 133:
                                                                                                                                                                                                                                                           
[...]

PR #2766, opened 2022-12-01
Merge v19 -> v20

fballiano:
Cherry picking from https://github.com/OpenMage/magento-lts/compare/v19.4.20...1.9.4.x

PR #2799, opened 2022-12-10
Re-move CM from composer

sreichel:
### Description (*)

Removed Cm_Chache from composer until symlik/submodules error is fixed.

(Added `cweagans/composer-patches` & `symplify/vendor-patches` for ZF1-future)

### Related Pull Requests

1. See OpenMage/magento-lts#2766

1. See OpenMage/magento-lts#2781


fballiano:
I don't think this is necessary, I've done some tests for https://github.com/OpenMage/magento-lts/pull/2165 and `credis` is already installed as some dependency in `vendor` so there's no need for it to be in `lib/Credis` (it's not the best that the folder is created empty but anyway...).

I've also checked that redis connection works and that the `vendor/colinmollenhour/credis/Client.php` is actually used, and it is.

So I don't think this is necessary.

sreichel:
For some reason the class was not found, but it works now.

But ... should we really add it with that symlink error?

fballiano:
In my opinion redis modules should be in our composer suggest instead of bundled in but... I don't know.

Anyway either we keep all of them or remove all of them, the symlink shouldn't be a "problem", although it's not looking great the software should work.

sreichel:
> In my opinion redis modules should be in our composer suggest

Sounds good to me, but `composer suggest` does not work for the "own" project. It will not be shown when you clone this repo and run `composer install`. It would work only for `composer require OpenMage/magento-lts` ...

However ... remove CM and add some info to README?

fballiano:
> It would work only for `composer require OpenMage/magento-lts` ...

well that should be the right way for people anyway right?

> However ... remove CM and add some info to README?

mmm thinking about it, it's been many years that the CM redis modules are bundled and people kinda expect the funcionality, if we remove those some people (the ones not using composer) will have problems updating so... I'd kinda leave things as they are now

sreichel:
> well that should be the right way for people anyway right?

yep, was just an info that suggestions are not shown for the own project/repo.

... lets keep it. :)

PR #1812, opened 2021-09-07
Some changes for compatibility with PHP 8.1

luigifab:
### Description

Not sure if all changes are the best or not.
PHP 8.0.11 and 8.1.0-rc2 (from [deb.sury.org](https://deb.sury.org/)).

### Manual testing scenario

First, disable _Suppress deprecation warnings_ in app/code/core/Mage/Core/functions.php:
```
diff --git a/app/code/core/Mage/Core/functions.php b/app/code/core/Mage/Core/functions.php
index 29e308d1b6..88b17f9c83 100644
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -136,9 +136,9 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
     }

     // Suppress deprecation warnings on PHP 7.x
-    if ($errno == E_DEPRECATED && version_compare(PHP_VERSION, '7.0.0', '>=')) {
-        return true;
-    }
+    //if ($errno == E_DEPRECATED && version_compare(PHP_VERSION, '7.0.0', '>=')) {
+    //    return true;
+    //}

     // PEAR specific message handling
     if (stripos($errfile.$errstr, 'pear') !== false) {
```

Go to your backend login page, login, go to dashboard, then enable PHP 8.1, clear your cache, and press F5.

**Commit 1 fix:**
```
Deprecated functionality: trim(): Passing null to parameter #1 ($string) of type string is deprecated
  in app/code/core/Mage/Core/Model/Config.php on line 1292
```

Apply commit 1, then press F5.

**Commit 2 fix:**
```
Deprecated functionality: PDOStatement::fetch(): Passing null to parameter #2 ($cursorOrientation)
  of type int is deprecated in lib/Zend/Db/Statement/Pdo.php on line 254
```

Solution found here: https://github.com/matomo-org/matomo/pull/17689
Apply commit 2, then press F5.

**Commit 3 fix:**
```
Deprecated functionality: str_replace(): Passing null to parameter #3 ($subject) of type array|string
  is deprecated in app/code/core/Mage/Core/Model/Resource/Config.php on line 104
```

Apply commit 3, then press F5.

Dashboard is displayed.
Clear your cache and your log directory, then press F5.

**Commit 4 fix (exception.log):**
```
Exception: Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string
  is deprecated  in app/code/core/Mage/Core/Helper/Abstract.php on line 218
```

[...]

PR #2806, opened 2022-12-12
[php8.1] deprecated passing null to str_replace() , ref #1812

sreichel:
### Related Pull Requests

1. OpenMage/magento-lts#1812
2. OpenMage/magento-lts#2586

`Deprecated functionality: str_replace(): Passing null to parameter #3 ($subject) of type array|string
  is deprecated in app/code/core/Mage/Core/Model/Resource/Config.php on line 104`

### Questions or comments

Fixed bit different. Set no default value when value from DB is NULL.


fballiano (approved):

app/code/core/Mage/Core/Model/Resource/Config.php
```diff
@@ -133,7 +133,7 @@ public function loadToXml(Mage_Core_Model_Config $xmlConfig, $condition = null)
         $deleteStores = [];
         // set stores config values from database
         foreach ($rowset as $r) {
-            if ($r['scope'] !== 'stores') {
+            if (is_null($r['value']) || $r['scope'] !== 'stores') {
```
kiatng, line 136: Didn't test, but if (is_null($r['value']) then $value = empty string in PHP7, and it would continue to line 40. Shouldn't PHP8 do the same?

sreichel: From my tests the array could look like

```
$r = [
    ...
    'value' => null
];
```

We can leave this open until the errors that popup before are fixed. Could be easier to test.

sreichel: Can you please test it?

Your right, this changes some behavoir ...

### Test
- set some config to null in DB (i've used Title Prefix form Admin -> Design config)
- load/save config
- set debug breakpoint at `Varien_Simplexml_Config::setNode()`

kiatng: I'll look into it after my holiday.

sreichel: > and it would continue to line

Right, but there we just create an empty XML-node for a `null` DB value. 

With this change the node is not created. What could happen? Access a not-existing node (`getConfigNode()` or `getNode()`)? Did not found any errors yet.

I could be an improvement to not create emtpy xml-nodes, but needs some testing ...

I'll open a new PR with casting to string to move forward. This can be investigated later. :)

sreichel:
This is the first error i see, when testing php8.1 and dont supress error.

Review needed. See comments above.

app/code/core/Mage/Core/Model/Resource/Config.php
sreichel: @elidrissidev tested this and everything looks fine. (Method description says _Load configuration values into xml config object_ and we just avoid loading `null` to config.)

elidrissidev: > Default valuew from config.xml are only used when no config path exists in `core_config_data`. It is not used when a `null` entry exists in DB.

@sreichel where did you that it's not used if value in db is null?

sreichel: Tested it.

elidrissidev: I have tested this by setting `contacts/email/recipient_email` to `NULL` in db (it has default value of "hello@ example.com" in xml), and then running:
```php
Mage::app()->getConfig()->getNode('default/contacts/email/recipient_email')
```
I could see that before adding the `is_null($r['value'])` check, it was returning an empty string value, whereas after that it's returning the value from xml. So might not be totally safe.

sreichel:
Better use https://github.com/OpenMage/magento-lts/pull/2901

app/code/core/Mage/Core/Model/Resource/Config.php
sreichel: Thanks. Tested `getConfigNode()` at the beginning - w/o default config value.

PR #2810, opened 2022-12-15
Sync v19 v20

sreichel:
### Description (*)

Some minor changes to reduce differences ... (marked same methods as deprecated).

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Image.php
+++ b/app/code/core/Mage/Catalog/Helper/Image.php
@@ -642,7 +642,7 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
         }
 
         if ($imageInfo[0] > $maxDimension || $imageInfo[1] > $maxDimension) {
-            Mage::throwException($this->__('Disalollowed file format.'));
+            Mage::throwException($this->__('Disallowed file format.'));
         }
 
         $_processor = new Varien_Image($filePath);
--- a/app/code/core/Mage/CatalogInventory/Model/Stock/Item.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Stock/Item.php
@@ -168,6 +168,7 @@ class Mage_CatalogInventory_Model_Stock_Item extends Mage_Core_Model_Abstract
      */
     protected function _initOldFieldsMap()
     {
+        // pre 1.6 fields names, old => new
         $this->_oldFieldsMap = [
             'stock_status_changed_automatically' => 'stock_status_changed_auto',
             'use_config_enable_qty_increments'   => 'use_config_enable_qty_inc'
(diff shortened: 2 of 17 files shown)


fballiano (approved):

fballiano:
every some time I've to say this, when will we get rid of v19? the differences aren't even that big to justify all the time put into fixing conflicts.

fballiano:
why do everybody hate v20? who needs ie8 and those iphone themes anymore?

sreichel:
For me its not about IE support or themes. Just want to have as less differences as possible.

fballiano:
if we have zero differences is way better

sreichel:
Then we dont need v20 at all :P

fballiano:
if we backport everything v19 will be v20

sreichel:
Actually we cant backport everything (in one step).

Think there are some changes that need reviews.

fballiano:
git merge 20.0...

I think some of the other maintainers won't be happy

app/code/core/Mage/Catalog/Helper/Image.php
```diff
@@ -642,7 +642,7 @@ public function validateUploadFile($filePath)
         }
 
         if ($imageInfo[0] > $maxDimension || $imageInfo[1] > $maxDimension) {
-            Mage::throwException($this->__('Disalollowed file format.'));
```
Flyingmana, line 645: this probably breaks several translations and should have only get changed in the translation file

Flyingmana:
@sreichel @fballiano if you want a change in supported versions: here is the place to contribute them https://github.com/OpenMage/rfcs/pull/7

app/code/core/Mage/Catalog/Helper/Image.php
fballiano: I've been arguing for the same thing for long time and I was alone saying that we should only update the translation files. At that time literally everybody was against my (and now your) idea. So please be more vocal next time. The people decided that they want to keep in sync strings in php files and in CSV files so now it is what it is.

ADDISON74: In that line it is about an exception that will be written in a log file. I think that these errors can be corrected in the PHP files and if they exist (although I doubt it) in the CSV files too. A programmer should know English, as well as a system administrator who has access to these files. On the other hand, although I do not agree to modify only the CSV files, I accepted PRs where such modifications were made, because it is for user interface and we should take care of translation without requiring efforts from those who periodically update the source code.. 

There are discussions on this topic in which the administrator takes note of the changes from one version to another and he makes the translations. From the programmer's perspective, honestly speaking, I don't like to see inconsistencies between sentences, especially when I look into the source code. I see something in the interface in English then in the code it's completely different or worse spelling mistakes like the one above, extremely annoying.

fballiano: The important thing is that we do not keep discussing the same things over and over again. I'm fine with the decision taken although it was not the choice I'd love, sometime it's a personal taste and it is fine :-)

sreichel: 1. i always said change only translation string in csv
2. @Flyingmana this PR was just to get rid of differences between branches. If we change it back, we should also revert in v20.

PR #2813, opened 2022-12-16
Removed Mage_PageCache

sreichel:
### Description (*)

`Mage_PageCache` was removed for v20 only. Created composer module to keep functionality if needed.

### Related Pull Requests

1. See OpenMage/magento-lts#2258

Can be reinstalled with `composer require openmage/module-mage-pagecache:dev-master`

### Questions or comments

@fballiano should we add these modules to releasebuilder?


fballiano:
if we backport everything we should just merge v20 in v19 and call it a day ;-)

sreichel:
That would be to easy :laughing:

fballiano:
that is what i mean when i say to drop v19 in favour of v20

fballiano:
I've created https://github.com/OpenMage/module-mage-pagecache for you ;-)

sreichel:
Pushed.

fballiano (dismissed):

fballiano:
about adding it to the release builder, I wouldn't do that (my personal opinion) for these reason:
- I wouldn't want it in v20
- it is easy to install it
- if people using the "built" zip file "overwrite" the core with the new versions (without deleting folders firts) they wouldn't even recognise that's gone (I know, that would be the same with CM modules but those are much more important in my opinion)

fballiano (approved):

ADDISON74 (approved):

PR #2825, opened 2022-12-18
Added getter/setter for getCouponCode()

sreichel:
### Description (*)

@elidrissidev https://github.com/OpenMage/magento-lts/pull/2586#issuecomment-1256919234

> escapeHtml phpdoc doesn't say it accepts nullable data, after all it doesn't make sense to escape null.
> 
> Respecting the PHPdocs will make it easier in the future to add strict type hinting to methods instead of making everything nullable.
> 
> What do you say?

> Yes, it does make no sense to escape null, but currently it is done.
> 
> I'd prefer the method, that takes most effort. Adding getter/setter methods ...

Added first getter/setter for test ... should make some changes  to templates obsolete.

### Related Pull Requests

1. See OpenMage/magento-lts#2586 ... search for `$this->getCouponCode()`

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -59,8 +59,6 @@
  * @method $this setCheckoutMethod(string $value)
  * @method string getConvertedAt()
  * @method $this setConvertedAt(string $value)
- * @method string getCouponCode()
- * @method $this setCouponCode(string $value)
  * @method string getCreatedAt()
  * @method $this setCreatedAt(string $value)
  * @method string getCustomerDob()
@@ -2088,5 +2086,22 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
             return $this;
         }
         return parent::save();
+    }
+
+    /**
+     * @return string
+     */
+    public function getCouponCode(): string
+    {
+        return (string)$this->_getData('coupon_code');
+    }
+
+    /**
+     * @param string $couponCode
+     * @return $this
+     */
+    public function setCouponCode(string $couponCode) #static with php74
+    {
+        return $this->setData('coupon_code', $couponCode);
     }
 }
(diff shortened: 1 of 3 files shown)


sreichel:
This is ready to test ... 

If it okay to fix in this way i'll add some commit.

ADDISON74 (approved):
LGTM

app/code/core/Mage/Sales/Model/Quote.php
```diff
@@ -2089,4 +2087,21 @@ public function save()
...
+    public function getCouponCode(): string
+    {
+        return (string)$this->_getData('coupon_code');
+    }
+
+    /**
+     * @param string $couponCode
+     * @return $this
+     */
+    public function setCouponCode(string $couponCode) #static with php74
```
fballiano, line 2103: what does "static with php74" comment mean?

sreichel: When adding new methods i tried to add type hints and return types if possible ... , was not sure about adding `self`.  Its a php8 change not 7.4 (https://php.watch/versions/8.0/static-return-type) ... comment can be removed.

However ... this should not be merged yet. If it's okay to add some getter/setter to avoid changes in templates, i'd also fix the other issues found by @elidrissidev.

elidrissidev:
I haven't checked this yet, usages of getCouponCode need to be checked carefully since this change will make it always a return a string even if not set.

sreichel:
Yes, but an empty string is no valid coupon code,

elidrissidev:
I know, I meant just to make there's no code that expects null or something. I remember similar issue happened before due to introducing getters that change the type.

sreichel:
> just to make sure there's no code that expects null or something

Think this would be found by phpstan ... https://phpstan.org/r/6b275cc5-439a-4af3-b201-b6c932a07b16

elidrissidev:
Just checked all the `getCouponCode` usages, the change looks safe to me.

sreichel:
Thanks for review. :+1: 

I'd add some more to fix the issues you've found during your test.

sreichel:
On hold ... #2884

sreichel:
Just fix coupon code related code here.

Flyingmana (approved):
good approach

PR #2827, opened 2022-12-20
Replace lib/Zend with shardj/zf1-future :rocket: 

sreichel:
### Description (*)

This PR replaces `lib/Zend` with [shardj/zf1-future](https://github.com/Shardj/zf1-future).

All changes made to `lib/Zend` or `app/code/core/Zend` have been replaced with patches. There are only two files left in `app/code/core/Zend` that needs reviews.

- `/app/code/core/Zend/Db/Select.php`
- `/app/code/core/Zend/Db/Statement.php`

Applied patches: https://github.com/OpenMage/magento-lts/pull/2787#issuecomment-1342255147 & https://github.com/OpenMage/magento-lts/pull/2787#issuecomment-1343425871

### Related Pull Requests

1. Closes OpenMage/magento-lts#2787

### Comments

- [ ] update README ... how add patches, special help for mac aliens :alien: 

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

Issue #2793, opened 2022-12-08
Call to undefined function rray_key_exists() in lib/Magento/Db/Adapter/Pdo/Mysql.php


hansgv:
### Preconditions (*)
OpenMage 19.4.20

### Steps to reproduce (*)
Attempt to save a product

### Expected result (*)
Saving works

### Actual result (*)
Error PHP Fatal error:  Uncaught Error: Call to undefined function rray_key_exists() in /var/www/vhosts/idlounge.fi/lib/Magento/Db/Adapter/Pdo/Mysql.php:135

I presume it's because it should be array_key_exists(). Fixing the typo seems to fix the issue.

elidrissidev:
I believe this was already fixed in #2780

PR #2835, opened 2022-12-22
Version bump

fballiano:
Since we just merged pretty important stuff, but at the same time we need more testing-time, what about a next version like 20.1.0-rc1 and 19.5.0-rc1?

Related PRs
- https://github.com/OpenMage/magento-lts/pull/2411
- https://github.com/OpenMage/magento-lts/pull/2827

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -213,18 +213,18 @@ final class Mage
         if ($majorVersion === 20) {
             return [
                 'major'     => '20',
-                'minor'     => '0',
-                'patch'     => '19',
-                'stability' => '', // beta,alpha,rc
+                'minor'     => '1',
+                'patch'     => '0',
+                'stability' => 'rc1', // beta,alpha,rc
                 'number'    => '', // 1,2,3,0.3.7,x.7.z.92 @see https://semver.org/#spec-item-9
             ];
         }
 
         return [
             'major'     => '19',
-            'minor'     => '4',
-            'patch'     => '21',
-            'stability' => '', // beta,alpha,rc
+            'minor'     => '5',
+            'patch'     => '0',
+            'stability' => 'rc1', // beta,alpha,rc
             'number'    => '', // 1,2,3,0.3.7,x.7.z.92 @see https://semver.org/#spec-item-9
         ];
     }


sreichel (approved):

sreichel:
Does upper/lower-case matter?

sreichel:
... but ... we have an error that sould be fixed. If its okay i'd remove latest release from packagist.org.

fballiano:
> Does upper/lower-case matter?

I don't think so, I used lowercase since it was already in the sample (comment next to the "alpha/beta/rc" like)

fballiano:
> ... but ... we have an error that sould be fixed. If its okay i'd remove latest release from packagist.org.

but how can we fix that anyway now?
they only way would be to create a branch from the currently released tag, fix the typo manually and release that, without PRs etc..

sreichel:
Yes. Its not so much to do, but we fix it without enforce to update to 1.9.5/...

sreichel:
You are doing the releases. Is it much work to make that hotfix release?

We could fix this bug w/o enforce to update to Zend-changes.

fballiano:
@sreichel in this case it's pretty easy, I've already created branch https://github.com/OpenMage/magento-lts/tree/19.4.21 with just 2 commits. they're pretty basic so everything should be ok. can you check it out a second and then I'll release?

also, it's a v19 hotfix only because v20 doesn't have the bug with https://github.com/OpenMage/magento-lts/issues/2793

sreichel:
Perfect.

Flyingmana:
> ... but ... we have an error that sould be fixed. If its okay i'd remove latest release from packagist.org.

be very very very careful with removing releases, this is going to break anyone who uses this in their composer.json or composer.lock. 
Replacing it does not work anyway, as it would have a different checksum and cause fails.

fballiano:
We’ll not remove anything don’t worry

sreichel:
> > ... but ... we have an error that sould be fixed. If its okay i'd remove latest release from packagist.org.
> 
> be very very very careful with removing releases, this is going to break anyone who uses this in their composer.json or composer.lock. Replacing it does not work anyway, as it would have a different checksum and cause fails.

Thanks. This is why i asked for.

Is there another propper way to avoid installing "bugged" releases?

Something like a required extension that cant we installed with certain releases?

fballiano:
Tomorrow morning I’ll release the new v19 and there will be no problem anymore 😉

fballiano:
19.4.21 is released, I was afraid of the process but at the end of the day it was pretty easy and I should have done it ealier.. it's done anyway now.

Flyingmana:
> Since we just merged pretty important stuff, but at the same time we need more testing-time, what about a next version like 20.1.0-rc1 and 19.5.0-rc1?

regarding the original question, can you link this very important stuff?
Generally I would say, releasing alpha,beta,rc releases is always okay.

But I wonder what it is, that requires the minor version bump, especially for the 19. release

fballiano:
it's the drop of CM modules and ZF1 moving them to composer

fballiano:
linkedin the PRs in the description. I think the bump should be minor, not your patch version

Flyingmana:
> it's the drop of CM modules and ZF1 moving them to composer

I see, that surely is a good reason for a minor version bump. And also a good reason for RC Tagging.

Flyingmana (approved):

PR #2837, opened 2022-12-23
phpstan: step back to level 4

sreichel:
### Description (*)

Think we made good progess. If we achieve L2/3/4 without ignoring/baseline errors, it would be great. Setting checks to L5 was to optimistic ... I already spent "hours", but without any help it is just to much.

### DIFF
--- a/phpstan.dist.baseline.neon
+++ b/phpstan.dist.baseline.neon
@@ -6,7 +6,12 @@ parameters:
 			path: app/Mage.php
 
 		-
-			message: "#^Strict comparison using \\=\\=\\= between 19 and 20 will always evaluate to false\\.$#"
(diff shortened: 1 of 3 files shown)


fballiano (approved):

ADDISON74 (approved):

tmotyl:
@sreichel In my opinion level 5 was ok to keep, (even with big baseline) because it prevents new code from introducing new errors.

sreichel:
> because it prevents new code from introducing new errors.

For this reason i've reverted that ignored errors. Level5 adds ~1000 errors to baseline.

Better l4 w/o ignoring something and fix existing ones. We still have +30 errors at level 0, that should be investigated.

PR #2850, opened 2022-12-25
Backported #1525

fballiano:
Since it's not the first time I'm dealing with a merge conflict regarding these files, I thought we could maybe backport #1525 to avoid having that as difference between v19 and v20.

What do you think?


ADDISON74 (dismissed):

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/View.php
```diff
@@ -118,11 +113,9 @@ public function getStoreCreateDateTimezone()
...
     public function getLastLoginDate()
     {
-        $date = $this->getCustomerLog()->getLoginAtTimestamp();
-        if ($date) {
-            return Mage::helper('core')->formatDate($date, Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true);
-        }
-        return Mage::helper('customer')->__('Never');
+        return ($date = $this->getCustomerLog()->getLoginAtTimestamp())
+            ? $this->formatDate($date, Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true, false)
+            : Mage::helper('customer')->__('Never');
```
sreichel, line 118:
```suggestion
        $date = $this->getCustomerLog()->getLoginAtTimestamp();
        if ($date) {
            return Mage::helper('core')->formatDate($date, Mage_Core_Model_Locale::FORMAT_TYPE_MEDIUM, true);
        }
        return Mage::helper('customer')->__('Never');
```

This is only a style change that i'd like to revert.

Btw ... set variable in condition could trigger code quality tool ... 

 `return ($date = $this->getCustomerLog()->getLoginAtTimestamp())`

sreichel:
@fballiano pls lt me try something ... if it goes wrong we need a ne PR :sunglasses:

Flyingmana:
I would say no, because its a BC break.
Is there a list with changes, which did cause this conflicts? and were they bugfixes or other changes?

sreichel:
> I would say no, because its a BC break.

I changed the code to avoid BC breaks. Please review.

(can make a new PR if wanted)

sreichel:
> Is there a list with changes, which did cause this conflicts?

### v19 only

- fixed in v20 OpenMage/magento-lts#1257
  ```
  message: "#^Return type \\(bool\\) of method Mage_Api_Model_Resource_User\\:\\:delete\\(\\) should be compatible with return type \\(\\$this\\(Mage_Core_Model_Resource_Db_Abstract\\)\\) of method Mage_Core_Model_Resource_Db_Abstract\\:\\:delete\\(\\)$#"
  path: app/code/core/Mage/Api/Model/Resource/User.php
  ```

WIP ... will update later ..

fballiano:
closing, v19 will freeze sometimes soon anyway

PR #2851, opened 2022-12-25
Added notes about composer library/modules to README (for 19.5.x and 20.1.x)

fballiano:
In #2411 and #2827 we radically changed the structure of OM, which was in place since the beginning of the project. So it's important that we update our documentation explaining the changes and the upgrade process.

Suggestions are welcome!


sreichel (approved):

ADDISON74:
I would analyze the behavior of two types of users who do not use Composer for various reasons. Those who download OM for the first time to evaluate it and those who want to upgrade from Magento or a previous versions of OM.

For the first type of users, those who will only download the archive of the repository will have to manually add all those libraries that we removed this year. They must find their list in the documentation, I would even give them the proper links. Basically I am addressing novice users and not advanced ones

For the second type of users, first of all, they must bring their version in correlation with what is now in the repository and then manually add the libraries. Here the situation should be better evaluated in case it is necessary to delete some files and directories, if it does not do an upgrade. For example, you rightly mentioned that you have to delete /lib/Zend and then add ZF1-Future.

In conclusion, do we cover with these specifications all the situations that may arise?

README.md
```diff
@@ -24,19 +16,15 @@ level of backwards compatibility to the official releases.
...
 ## Requirements
 
-- PHP 7.3+ (PHP 8.0 is supported)<br/>
-`mcrypt` extension needs to be disabled to fallback on `mcryptcompat` and `phpseclib`. 
+- PHP 7.3+ (PHP 8.0 is supported)
 - MySQL 5.6+ (8.0+ recommended) or MariaDB
 
+
+- PHP extension `intl` <small>since 1.9.4.19 & 20.0.17</small>
+- Command `patch` (v2.7+) <small>since 1.9.5.0 & 20.1.0</small>
```
fballiano, line 26: should we write about gpatch for homebrew?

fballiano:
@ADDISON74 there's no need (for any of the 2 types of users) to manually have to download the libraries, the release builder adds them already ;-) So I think we should already be covering the 2 different use cases.

README.md
sreichel: Was not sure about it.

Put it to requirements or add section for mac users?

sreichel (dismissed):

sreichel (approved):

ADDISON74 (approved):

PR #2739, opened 2022-11-17
Add label, date and datetime picker to system.xml

luigifab:
### Description

Some new fields for system.xml, example:

image

XML of the screen shot:
```xml
<head>
	<label>Crash test</label>
	<frontend_model>adminhtml/system_config_form_field_heading</frontend_model>
	<sort_order>1001</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</head>
<label_1>
	<label>label</label>
	<value>Pouet</value>
	<frontend_model>adminhtml/system_config_form_field_label</frontend_model>
	<sort_order>1002</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</label_1>
<labelbold_2>
	<label>label bold</label>
	<value>Pouet Pouet</value>
	<bold>1</bold>
	<frontend_type>label</frontend_type><!-- optionnal and useless -->
	<frontend_model>adminhtml/system_config_form_field_label</frontend_model>
	<sort_order>1003</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</labelbold_2>
<datetime_1>
	<label>datetime (original)</label>
	<frontend_model>adminhtml/system_config_form_field_datetime</frontend_model>
	<sort_order>1004</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</datetime_1>
<datetime_2>
	<label>datetime with format</label>
	<frontend_type>datetime</frontend_type><!-- optionnal and useless -->
	<frontend_model>adminhtml/system_config_form_field_datetime</frontend_model>
	<format>FORMAT_TYPE_SHORT</format>
	<sort_order>1005</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</datetime_2>
<date_3>
	<label>date</label>
	<frontend_model>adminhtml/system_config_form_field_date</frontend_model>
	<sort_order>1006</sort_order>
	<show_in_default>1</show_in_default>
	<show_in_website>1</show_in_website>
	<show_in_store>1</show_in_store>
</date_3>
<date_4>
	<label>date with format</label>
	<frontend_type>date</frontend_type><!-- optionnal and useless -->
[...]

PR #2854, opened 2022-12-26
Fixed incorrect datetime in block, ref #1525

sreichel:
### Description (*)

#1525 was added to v20 only, because it was a BC breaking change. This PR solves that issue w/o changing existing methods signature. (code from 1525)

Not sure about this change ... changing ...

```
if (!($date instanceof Zend_Date) && $date && !strtotime($date)) {
    return '';
}
if (is_null($date)) {
    $date = Mage::app()->getLocale()->date(Mage::getSingleton('core/date')->gmtTimestamp(), null, null);
```

### Related Pull Requests

1. See OpenMage/magento-lts#1525

1. Closes OpenMage/magento-lts#2850

1. see related PR
2. ...

### DIFF
--- a/app/code/core/Mage/Core/Helper/Data.php
+++ b/app/code/core/Mage/Core/Helper/Data.php
@@ -153,23 +153,40 @@ class Mage_Core_Helper_Data extends Mage_Core_Helper_Abstract
     /**
      * Format date using current locale options and time zone.
      *
-     * @param   string|Zend_Date|null $date
+     * @param   string|Zend_Date|null $date   If empty, return current datetime.
      * @param   string              $format   See Mage_Core_Model_Locale::FORMAT_TYPE_* constants
      * @param   bool                $showTime Whether to include time
      * @return  string
      */
     public function formatDate($date = null, $format = Mage_Core_Model_Locale::FORMAT_TYPE_SHORT, $showTime = false)
     {
+        return $this->formatTimezoneDate($date, $format, $showTime);
+    }
+
+    /**
+     * Format date using current locale options and time zone.
+     *
+     * @param   string|Zend_Date|null $date   If empty, return current datetime.
+     * @param   string              $format   See Mage_Core_Model_Locale::FORMAT_TYPE_* constants
+     * @param   bool                $showTime Whether to include time
+     * @param   bool                $useTimezone Convert to local datetime?
+     * @return  string
+     */
+    public function formatTimezoneDate($date = null, $format = Mage_Core_Model_Locale::FORMAT_TYPE_SHORT, $showTime = false, $useTimezone = true)
+    {
         if (!in_array($format, $this->_allowedFormats, true)) {
             return $date;
         }
-        if (!($date instanceof Zend_Date) && $date && !strtotime($date)) {
-            return '';
-        }
-        if (is_null($date)) {
-            $date = Mage::app()->getLocale()->date(Mage::getSingleton('core/date')->gmtTimestamp(), null, null);
+        if (empty($date)) {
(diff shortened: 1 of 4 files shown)


sreichel:
possibly related to #2739

fballiano:
@kiatng what do you think about this?

kiatng:
This is a better fix of the original issue. 

Is this going to be in v20? If it is and if there are custom code in the wild that uses `$useTimezone` in `formatDate()`, it'll produce unexpected result. But BC in v20 is tolerable. Is it worthwhile to add a note in README for BC in v20?

kiatng:
Test result comparing original v19 `formatDate($date)` and `formatTimezoneDate($date)`:

1. `$date = '2022-12-28 13:00:00';`, both results are "28/12/2022 21:00", correctly converted to local time.
2. `$date = null;`, both results are "28/12/2022 03:04" (GMT time)
3. `$date = '2022-12-28 13:00:00'; $date = new Zend_Date($date);`, both results are "28/12/2022 13:00", correct because timezone in `Zend_Date` was not set.
4. `$date = '2022-12-28 13:00:00'; $date = strtotime($date);`, `formatDate()` -> "", `formatTimezoneDate()` -> "28/12/2022 21:00". I think result of `formatTimezoneDate()` is preferable.

app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -153,31 +153,50 @@ public function formatPrice($price, $includeContainer = true)
     /**
      * Format date using current locale options and time zone.
      *
-     * @param   string|Zend_Date|null $date
+     * @param   string|Zend_Date|null $date   If empty, return current datetime.
```
kiatng, line 156:
```suggestion
     * @param   string|Zend_Date|int|null $date If empty, return current local datetime.
```

sreichel:
@kiatng thanks. Would be nice to had a unit-test here.

app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -153,31 +153,50 @@ public function formatPrice($price, $includeContainer = true)
...
      *
-     * @param   string|Zend_Date|null $date
+     * @param   string|Zend_Date|int|null $date If empty, return current local datetime.
      * @param   string              $format   See Mage_Core_Model_Locale::FORMAT_TYPE_* constants
      * @param   bool                $showTime Whether to include time
      * @return  string
      */
     public function formatDate($date = null, $format = Mage_Core_Model_Locale::FORMAT_TYPE_SHORT, $showTime = false)
     {
+        return $this->formatTimezoneDate($date, $format, $showTime);
```
kiatng, line 163: For v20 compatibility, maybe something like this:
```suggestion
        return func_num_args() === 4
             ? $this->formatTimezoneDate($date, $format, $showTime, func_get_args()[3]);
             : $this->formatTimezoneDate($date, $format, $showTime);
```

sreichel: Would say no. v20 change was added four month ago in 08/2022 .... if someone wrote code that uses 4th param, he/she should be able to change that line. _imho_

kiatng (approved):
Should not have serious effect in v20.

fballiano:
so the idea is to port it to v20 right? otherwise it would be actually worse to maintain

sreichel:
Add @kiatng fix to v19 and port it to v20 was my plan ... :grey_question:

ADDISON74:
Closing this PR was wrong decission. It was a better fix for that issue.

sreichel:
@ADDISON74 there is a lollow-up PR. (#2940)

Issue #2857, opened 2022-12-27
Rebuild Tag aggregation data not working


sreichel:
### Preconditions (*)

1. 1.9.4.x
2. ...

### Steps to reproduce (*)

1. backend index management
2. tags ...

### Actual result (*)

```
2022-12-27T22:20:03+00:00 ERR (3): 
PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'COUNT(tr.customer_id) + MIN(IF(tp.base_popularity IS NOT NULL, tp.base_popularity, 0))' in 'field list' in /var/www/html/vendor/shardj/zf1-future/library/Zend/Db/Statement/Pdo.php:228
Stack trace:
#0 /var/www/html/vendor/shardj/zf1-future/library/Zend/Db/Statement/Pdo.php(228): PDOStatement->execute(Array)
#1 /var/www/html/lib/Varien/Db/Statement/Pdo/Mysql.php(104): Zend_Db_Statement_Pdo->_execute(Array)
#2 /var/www/html/vendor/shardj/zf1-future/library/Zend/Db/Statement.php(303): Varien_Db_Statement_Pdo_Mysql->_execute(Array)
#3 /var/www/html/vendor/shardj/zf1-future/library/Zend/Db/Adapter/Abstract.php(480): Zend_Db_Statement->execute(Array)
#4 /var/www/html/vendor/shardj/zf1-future/library/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('INSERT INTO `ta...', Array)
#5 /var/www/html/lib/Varien/Db/Adapter/Pdo/Mysql.php(492): Zend_Db_Adapter_Pdo_Abstract->query('INSERT INTO `ta...', Array)
#6 /var/www/html/app/code/core/Mage/Tag/Model/Resource/Indexer/Summary.php(215): Varien_Db_Adapter_Pdo_Mysql->query('INSERT INTO `ta...')
#7 /var/www/html/app/code/core/Mage/Tag/Model/Resource/Indexer/Summary.php(125): Mage_Tag_Model_Resource_Indexer_Summary->aggregate()
#8 /var/www/html/app/code/core/Mage/Index/Model/Indexer/Abstract.php(145): Mage_Tag_Model_Resource_Indexer_Summary->reindexAll()
```

Related code:

`Mage_Tag_Model_Resource_Indexer_Summary::aggregate()`

```
                ->from(
[...]

PR #2858, opened 2022-12-27
Fixed tag aggregation indexer query

fballiano:
Problem is explained here:

- closes OpenMage/magento-lts#2857

The fix is just to have a correct quoting in the query.


sreichel:
Nice. Started to debug some minutes age :+1:

sreichel (approved):

ADDISON74 (approved):

PR #2297, opened 2022-07-08
Remove DISCLAIMER and change Magento -> OpenMage in header

justinbeaty:
### Description (*)

From discussion #2276

Updated file headers to change Magento -> OpenMage and remove the DISCLAIMER about modifying core files. This removed about 1.9MB in disk space.

### Manual testing scenarios (*)

Check out this branch and run:

```console
git diff -U0 HEAD~1 | grep '^[+-]' | grep -Ev '^(--- a/|\+\+\+ b/)' | awk '!x[$0]++'
```

That will print all unique changes made in this PR. You should only see comment changes. I will post the output in a follow up comment.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2866, opened 2022-12-29
Updated copyright blocks

sreichel:
### Description (*)

I took me some hours to go though all files that changes and updated copyright blocks .... not.

Just used https://github.com/OpenMage/magento-lts/pull/2295 :+1: 

### DIFF
--- a/app/code/core/Mage/Admin/Helper/Data.php
+++ b/app/code/core/Mage/Admin/Helper/Data.php
@@ -15,6 +15,7 @@
  * @category   Mage
  * @package    Mage_Admin
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2022 The OpenMage Contributors (https://www.openmage.org)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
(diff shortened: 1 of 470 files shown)


ADDISON74 (approved):

fballiano (approved):
love it

Flyingmana:
just, why?
why do we need to update the year in the copyright headers? what is the benefit for anyone?

ADDISON74:
If I remember correctly the script created by @justinbeaty makes changes only in the files that have been changed in the meantime. There is a long discussion why we need to change the copyright header in a few places (#2276, #2297) and here for legal aspects (#2295).

sreichel:
@Flyingmana we started this with #2297. Now we can/could/should keep it up to date? Are there any downsides?

fballiano:
in the old times they were updating every single file at every release. we're just updating the files that were actually modified, which makes 78% more sense ;-)
anyway it was a very long discussion and it was decided to accept that PR.

sreichel:
> which makes 78% more sense

82% ... at least.

ADDISON74:
Changing the header can also be seen as useful information. For example, a file had its last changes in the year 2022. Not that it would help much, but there are files that we have never touched inherited from Magento.

PR #2884, opened 2023-01-03
Fixes passing null to strlen() inside escapeHtml()

sreichel:
### Description (*)

> I don't disagree with fixing existing uses of escapeHtml but just for practicality sake this is going to be a big headache for those with lots of extensions so I propose changing the signature of escapeHtml to allow a nullable string (if null is passed it should return an empty string).

We just need to check if we pass as string to `strlen()` inside `escapeHtml()`. Should be enough?

### Related Pull Requests
- https://github.com/OpenMage/magento-lts/pull/2586#issuecomment-1256840473
- https://github.com/OpenMage/magento-lts/pull/2586#issuecomment-1369311108

### DIFF
--- a/app/code/core/Mage/Core/Helper/Abstract.php
+++ b/app/code/core/Mage/Core/Helper/Abstract.php
@@ -212,7 +212,7 @@ abstract class Mage_Core_Helper_Abstract
             }
         } else {
             // process single item
-            if (strlen($data)) {
+            if (is_string($data) && strlen($data)) {
                 if (is_array($allowedTags) && !empty($allowedTags)) {
                     $allowed = implode('|', $allowedTags);
                     $result = preg_replace('/<([\/\s\r\n]*)(' . $allowed . ')([\/\s\r\n]*)>/si', '##$1$2$3##', $data);


app/code/core/Mage/Core/Helper/Abstract.php
```diff
@@ -212,7 +212,7 @@ public function escapeHtml($data, $allowedTags = null)
             }
         } else {
             // process single item
-            if (strlen($data)) {
+            if (is_string($data) && strlen($data)) {
```
elidrissidev, line 215: If `$data` is not a string, the method will end up returning `null`.

aweb3r (approved):

app/code/core/Mage/Core/Helper/Abstract.php
sreichel: I did not want to change the return type, just avoid passing null to strlen.

elidrissidev (approved):

Flyingmana (approved):

sreichel:
Okay ... to late to change it to `$data !== ''`.

sreichel:
@Flyingmana @fballiano where is the project tab? I've used that to track some issues. Recently added "PHP8.1"

fballiano:
My fault, was trying to hide the double “projects/projects classic” and they both got hidden, but now it’s ok and there’s only one projects tab

PR #2889, opened 2023-01-03
Ensure null values are not passed to trim()

mattdavenport:
As Magento allows eav attributes without a backing table (`backend_table` in the DB), we should ensure that those `null` values are passable into the trim function. I noticed @sreichel used a string typecast in the larger #2586 PR , so I'm going with that same convention. Alternatively a null coalesce could be used here if that is preferred. This fixes the deprecation notice in PHP 8.1.


sreichel:
> I noticed @sreichel used a string typecast in the larger https://github.com/OpenMage/magento-lts/pull/2586 PR

Its not my PR, but LGTM ;)

mattdavenport:
Wrong tag @sreichel, my apologies! I was just looking at another 8.1 issue you were fixing 😛

sreichel (approved):

fballiano (approved):

PR #2811, opened 2022-12-16
Removed Mage_Backup

sreichel:
### Description (*)

Removed `Mage_Backup` code.

Can be reinstalled with `composer require openmage/module-mage-backup`

### Related Pull Requests

- See OpenMage/magento-lts#2259
- See OpenMage/magento-lts#2782

### Questions or comments

Repo should be moved to OM space.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2890, opened 2023-01-03
Require Mage_Backup and Mage_PageCache on v19's release builder

fballiano:
Since https://github.com/OpenMage/magento-lts/pull/2811 Mage_Backup should be in the release builder, only for v19


sreichel:
Nice.

Mhh, Mage_PageCache too?

fballiano:
done

sreichel (approved):

ADDISON74 (approved):

PR #2893, opened 2023-01-04
Added default return string to getParam()

sreichel:
### Description (*)

@elidrissidev i hope is okay to split your PR to have some progress here.

### Related Pull Requests

- OpenMage/magento-lts#2586

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit.php
@@ -301,6 +301,6 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit extends Mage_Adminhtml_Block_Wid
      */
     public function getSelectedTabId()
     {
-        return addslashes(htmlspecialchars($this->getRequest()->getParam('tab')));
+        return addslashes(htmlspecialchars($this->getRequest()->getParam('tab', '')));
     }
 }
--- a/app/code/core/Mage/Adminhtml/Controller/Report/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Controller/Report/Abstract.php
@@ -72,7 +72,7 @@ abstract class Mage_Adminhtml_Controller_Report_Abstract extends Mage_Adminhtml_
             $blocks = [$blocks];
         }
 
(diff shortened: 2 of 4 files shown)


elidrissidev (approved):

app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit.php
```diff
@@ -301,6 +301,6 @@ public function getIsConfigured()
      */
     public function getSelectedTabId()
     {
-        return addslashes(htmlspecialchars($this->getRequest()->getParam('tab')));
+        return addslashes(htmlspecialchars($this->getRequest()->getParam('tab', '')));
```
fballiano, line 304: it seems the getParam() function declaration is wrong here no? I think it would be more correct to fix it in ZF1F or to have it patched, because otherwise we'd have to patch all of the calls to getRequest()->getParams()

elidrissidev: getParam() has a default value of null, I think changing that will cause more problems.

fballiano: true, but a default value for a parameter is meant because this way you can avoid to pass it every time, which is what we (and everybody calling the getParam()) are forced to do here.

Sdfendor (approved):

app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit.php
sreichel: > because otherwise we'd have to patch all
...
> true, but ...

@fballiano i did not get it. We dont need to change `getParam()`, we just need a string returned in some cases.

kiatng (approved):
`getParam()` returns `null` by default, to change the default behavior to empty string is problematic because some code may check for `null`.

app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit.php
elidrissidev: Is this resolved?

fballiano:
cherry-picked to v20 too since there were no conflicts

PR #1313, opened 2020-11-18
Better Upload exceptions in Uploader.php

sicet7:
### Description (*)
This pull request simply adds more exceptions messages to the uploader.
The exception messages is copied from: https://www.php.net/manual/en/features.file-upload.errors.php#115746

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #2902, opened 2023-01-05
Better exceptions for file upload

fballiano:
Since https://github.com/OpenMage/magento-lts/pull/1313 was closed because the author couldn't fix it I've rewrote it and posted it.

This PR adds some exception messages in case of uploading error.

Messages and constant name/values are taken from https://www.php.net/manual/en/features.file-upload.errors.php

### DIFF
--- a/lib/Varien/File/Uploader.php
+++ b/lib/Varien/File/Uploader.php
@@ -32,6 +32,16 @@
 
 class Varien_File_Uploader
 {
+    const UPLOAD_ERRORS = array(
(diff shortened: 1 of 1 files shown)


Sdfendor (approved):

ADDISON74 (approved):
I agree with this PR for better error management in the case of uploading images. The explanatory phrases were not created but taken from here, which is fine

https://www.php.net/manual/en/features.file-upload.errors.php

elidrissidev (approved):

fballiano:
cherry-picked to branch 20.0 since there were no conflicts

Issue #2870, opened 2022-12-30
Separate admin and front-end database users for monitoring and security


loekvangool:
### Description
While monitoring OpenMage with logging/APM/slowlog tooling, I'm finding it necessary to distinguish between admin and frontend database access. For example, I'm much more interested in front-end slow queries than admin slow queries. However it's not possible to separate these two traffic flows from the database server logs. Changing the database user uses seems to be an elegant solution.

### Expected behavior
Maybe we should extend `core_read` and `core_write` with `admin_read` and `admin_write`? I'm not sure how feasible it would be, but I know that `core_read` is optional and therefore there already is a fallback mechanism in OpenMage's PDO implementation.

On top of this, it allows  tighter database authorizations for the front-end database user, e.g. the front-end probably does not need `CREATE`, `DROP` or `TRUNCATE` authorizations, limiting the impact of a security breach.

elidrissidev:
Hey @loekvangool, I've opened a PR to add this feature. Can you test it and let me know if everything is good?

loekvangool:
@elidrissidev Thanks for your quick work on this. It did seem like a good solution, but @Flyingmana 's is right about the trouble with emulation. The proposed solution with webserver variables is OK for me too, although not as elegant.

Issue #643, opened 2019-04-01
Is the community open to a backward compatible PR that would have `getConfig()` reference env variables that could override values specified in XML or the database?


tylers-username:
Sorry if this has been discussed before, I did not dig through Issue history.

I'm considering opening a PR that lets you override XML and DB config values with env variables. It should be non-breaking.

For example, you may have something like:

```xml
<default>
    <catalog>
        <some_nested_node>                
            <enabled>myname</enabled>
            <header>This is my header</header>
        </some_nested_node>
    </catalog>
</default>
```

This can be referenced today with: 

```php
echo Mage::getConfig()->getNode('catalog/some_nested_node')->header; 

# Returns: "This is my header"
```

With my proposed solution, you'd be able to implement an inveronment variable such as `MAGE_CONFIG/CATALOG/SOME_NESTED_NODE/HEADER="My new header"`. Magento would then convert any `MAGE_CONFIG/` variables to an XML node in order to maintain compatibility and load it last to ensure that it overrides any previously set values.

tylerssn:
Ah, I'd seen tools such as dotenv honor slashes. I didn't realize that environments don't natively support this. I had only tested this concept with: 

```
env MAGE_CONFIG\some\config\path=OKAY magerun dev:console

Mage::getStoreConfig('some\config\path'); # Returns "OKAY"
```

> My main concern when adding this in would be the security implications. 

I think there would have to be an exception thrown if `putenv()` was enabled. I'm not a security expert, so I don't know if or what can be done with `$_ENV`. I don't know about the "trustability" of environment variables in general, just that a number of frameworks (Laravel in PHP, Gatsby in NodeJS) implement tools such as dotenv to provide connection details. 

My main use-case for this is for establishing  connections. For example, currently on Platform.sh (the service behind Magento Cloud) I have to keep my `/app/etc/` directory writable as a mount because I cannot generate my local.xml file during deployments until my service relationships are available (db, redis, etc) - they're not available until the read-only system is locked down. I'd rather not have the config directory writable.

I think the alternative route to this is implementing something like: https://packagist.org/packages/gracious/magento-env-config

tmotyl:
Hi
Here is a good read on challenges of the env variables:
https://mattallan.me/posts/how-php-environment-variables-actually-work/

As there is a potential security risk I would say this feature should not be bundled in mage lts by default.

Flyingmana:
My general rule of thumb for this would be: "how is Magento2 handling this?"
[...]

PR #2903, opened 2023-01-05
Allow using a separate db connection for Admin

elidrissidev:
### Description (*)

This PR adds a new feature that allows optionally using separate database connections (`admin_read` and `admin_write`) for the Admin store. This can be desired for many reasons such as limiting the privileges of the frontend connection's db user. By default, these connections inherit the default ones, so they should work out of the box without any configuration.

1. Fixes OpenMage/magento-lts#2870

1. Go to your `local.xml` and add the following configuration inside `config/global/resources` element:
```xml
<admin_setup>
    <connection>
        <host><![CDATA[localhost]]></host>
        <username><![CDATA[root]]></username>
        <password><![CDATA[root]]></password>
        <dbname><![CDATA[openmage]]></dbname>
        <initStatements><![CDATA[SET NAMES utf8]]></initStatements>
        <model><![CDATA[mysql4]]></model>
        <type><![CDATA[pdo_mysql]]></type>
        <pdoType><![CDATA[]]></pdoType>
        <active>1</active>
    </connection>
</admin_setup>
<admin_write>
    <connection>
        <use>admin_setup</use>
    </connection>
</admin_write>
<admin_read>
    <connection>
        <use>admin_setup</use>
    </connection>
</admin_read>
<!-- 
It's important to override core_setup to use the new admin connection 
for setup scripts to run without problems, especially if you limit the privileges
of default connection.
-->
<core_setup>
    <connection>
        <use>admin_setup</use>
    </connection>
</core_setup>
```

Test to make sure the Admin area is now using the new connection. In my case, I made it use the database with sample data installed, whereas the frontend was using a clean database.

### DIFF
--- a/app/code/core/Mage/Core/Model/Resource.php
+++ b/app/code/core/Mage/Core/Model/Resource.php
@@ -33,9 +33,11 @@ class Mage_Core_Model_Resource
     public const AUTO_UPDATE_NEVER      = -1;
     public const AUTO_UPDATE_ALWAYS     = 1;
 
-    public const DEFAULT_READ_RESOURCE  = 'core_read';
-    public const DEFAULT_WRITE_RESOURCE = 'core_write';
-    public const DEFAULT_SETUP_RESOURCE = 'core_setup';
+    public const DEFAULT_ADMIN_READ_RESOURCE  = 'admin_read';
+    public const DEFAULT_ADMIN_WRITE_RESOURCE = 'admin_write';
+    public const DEFAULT_READ_RESOURCE        = 'core_read';
+    public const DEFAULT_WRITE_RESOURCE       = 'core_write';
+    public const DEFAULT_SETUP_RESOURCE       = 'core_setup';
 
     /**
      * Instances of classes for connection types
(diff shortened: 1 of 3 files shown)


sreichel:
Nice,

We should ...

- add some info to README
- update `local.xml.template`?
- add PR for [DDEV](https://github.com/drud/ddev/blob/ca6d01ddbc28c2a2b9439e43cfba33c66135ee7d/pkg/ddevapp/magento/local.xml)

sreichel:
Seems not to be a BC break. Change branch?

app/code/core/Mage/Core/Model/Resource.php
```diff
@@ -201,10 +203,14 @@ protected function _newConnection($type, $config)
      */
     protected function _getDefaultConnection($requiredConnectionName)
     {
+        $isAdmin = Mage::app()->getStore()->isAdmin();
+        $readResource = $isAdmin ? self::DEFAULT_ADMIN_READ_RESOURCE : self::DEFAULT_READ_RESOURCE;
+        $writeResource = $isAdmin ? self::DEFAULT_ADMIN_WRITE_RESOURCE : self::DEFAULT_WRITE_RESOURCE;
```
Flyingmana, line 208: This way the same function can for the same input return a different connection. If you send an email, the store is changed via emulation, probably not beeing Admin anymore? which means that existing transactions are ignored (or changed data from them not available) and, simply said a lot of hard to debug problems

Flyingmana:
If you need a separate connection for admin, than it makes more sense to use a different Config for the whole request.
There was a Feature planned, to provide local.xml configs via Environment Variables, which means you could configure on the webserver level to use a different connection for Admin, which would achieve the same, with less potential for hard to debug Bugs.

elidrissidev:
I will check the env variables solution later. For now this is going back to draft.

sreichel:
> There was a Feature planned, to provide local.xml configs via Environment Variables,

Do you have link for that?

elidrissidev:
> Do you have link for that?

I think this: [#643](https://github.com/OpenMage/magento-lts/issues/643)

fballiano:
I like this solution more compared to https://github.com/OpenMage/magento-lts/issues/643, could it be possible to fix the emulation?

elidrissidev:
> I like this solution more compared to #643, could it be possible to fix the emulation?

I don't know yet, haven't much time to explore this further.

Flyingmana:
When following the way with having a separate connection during the request, you probably need to introduce a state, (which increases the complexity of the class).
Maybe it also needs to be checked, how it reacts for cronjobs, Iam not sure if they use the admin or default store in the beginning. But I have seen some which change the current store.
When doing so, keep also the cleanup in mind, as some setups depend on `\Mage::reset()` to work properly, as they process multiple requests without bootstrapping a new request/process.

Also it might make sense to make this configurable and default Off, as its going to double the parallel connections to the database, which might break a lot of closely optimized Setups.

PR #2904, opened 2023-01-06
Fix Use Base Class Instead of Child Class when Referencing Constants

Sdfendor:
### Description (*)

There are two class constants being referenced by using a child class instead of using the base class, in which the constants are defined. Those constants aren't overriden. Therefore I don't think this usage of the child class is deliberate.
In itself this small mistake wouldn't bother me, but I encountered a scenario where those references cause problems. The constants aren't used withing `Mage_Newsletter` but e.g. in a block that is part of the email template grid in the admin area. In case one deactivates/removes `Mage_Newsletter` the code referencing `Mage_Newsletter_Model_Template` will fail. 

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Newsletter/Template/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Newsletter/Template/Grid.php
@@ -101,8 +101,8 @@ class Mage_Adminhtml_Block_Newsletter_Template_Grid extends Mage_Adminhtml_Block
                 'index' => 'template_type',
                 'type' => 'options',
                 'options' => [
-                    Mage_Newsletter_Model_Template::TYPE_HTML   => 'html',
-                    Mage_Newsletter_Model_Template::TYPE_TEXT   => 'text'
+                    Mage_Core_Model_Template::TYPE_HTML => 'html',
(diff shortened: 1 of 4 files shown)


elidrissidev (approved):

sreichel:
I've deactivated Mage_Newsletter but dont see any error. Where does the error appear?

fballiano:
it you deactivate Mage_Newsletter it still continues to work, but it you really remove it from the file system than you'll get the error.

And since Mage_Adminhtml doesn't depend on Mage_Newsletter it's probably correct to merge this PR.

Sdfendor:
@sreichel @fballiano Yes I was wrong about the deactivation of the module. It solely happens when deleting the module. In our current OpenMage installation we strife to keep the footprint of the installation minimal so we delete modules we don't use entirely. A place where the error happens with removed module `Mage_Newsletter` would be the transactional email grid: `System > Transactional Emails`. 
That it doesn't happen for a deactivated module makes sense too, because the constant is referenced directly not indirectly by use of the appropriate factory method. Therefore the module deactivation has no effect, the autoloader doesn't consider whether a file is "deactivated" or not and the class is loaded regardless.

fballiano (approved):

fballiano:
merged and cherry-picked to 20.0

PR #2930, opened 2023-01-13
Documenting all shell/indexer.php arguments with dashes

loekvangool:
Actual usage is unaffected, all of these are equal:

- `indexer.php --info`
- `indexer.php -info`
- `indexer.php info`


ADDISON74 (approved):

fballiano:

PR #2944, opened 2023-01-15
Don't run OpenMage as root

luigifab:
### Description

Not sure all people will be happy, but this PR prevent usage of OpenMage with **root** user.

### Contribution checklist

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -12,12 +12,18 @@
  * obtain it through the world-wide-web, please send an email
  * to license@magento.com so we can send you a copy immediately.
  *
- * @category    Mage
- * @package     Mage
+ * @category   Mage
+ * @package    Mage
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
  * @copyright  Copyright (c) 2017-2022 The OpenMage Contributors (https://www.openmage.org)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
+
+if (posix_geteuid() == 0) {
+    @http_response_code(500);
+    echo "Don't run OpenMage as root!\n";
+    exit(-1);
+}
 
 define('DS', DIRECTORY_SEPARATOR);
 define('PS', PATH_SEPARATOR);


fballiano:
is it "light enough" to be added to every request?

luigifab:
Yes I think:
```
php > $a = microtime(true); $b = posix_geteuid(); echo microtime(true) - $a,"\n",$b;
6.9141387939453E-6
1000
```

sreichel:
Why?

luigifab:
Do you know `eval` is _evil_?
I think to run PHP code with root user this is same.

sreichel:
> Not sure all people will be happy ...

Please dont get me wrong, but your description lacks of some information. 

What about users that run OM as root? Should we give some advice to change their setup?

tmotyl:
what about docker deployments? is root unsafe there too?

ADDISON74:
I could call it a potential security issue, but referring to a configuration of the machine  by someone who today installs programs in Linux and does it according to steps found through tutorials on the Internet.

I would agree to a **warning message** under certain conditions:

1. There should be an option in the Backend to determine whether this check is done or not. By default to be set to Yes.

2. A message should be displayed at the top of the website, similar to when it is in Demo mode, by which the administrator will be warned **only in the Backend** that he is running with root privileges. In this message, tell him how he can deactivate the warning.

3. Let it be just a warning, not to prevent the shopping cart from running.

The Docker issue is very well put on paper. It is also up to the administrator to decide if he runs with root privileges or not. Honestly specking, in order to run OM in production as root, you have to do a lot of steps to achieve that. I have not encountered such a situation on servers where I have not installed LAMP and Magento.

tmotyl:
I'm against putting it in index.php.
I'm in favor of making an admin panel "security report" or "environment check report" where checks like this could be performed.

luigifab:
Why not for a note in backend. So this is the end of this idea?

sreichel:
> So this is the end of this idea?

As @tmotyl said ... we should add these checks to somewhere else.

elidrissidev:
> I'm against putting it in index.php. I'm in favor of making an admin panel "security report" or "environment check report" where checks like this could be performed.

Agree, similar to what's done with the `app/etc/local.xml` file being readable.

sreichel:
@elidrissidev

OT: at discord you wrote something about to "ignore" changes made to Mage.php when running `composer install` ... can you please at it to docs (somewhere)?

elidrissidev:
> @elidrissidev
> 
> OT: at discord you wrote something about to "ignore" changes made to Mage.php when running `composer install` ... can you please add it to docs (somewhere)?

I'm considering that a temporary approach because magento-composer-installer keeps adding it back whenever you run `composer install`.

sreichel:
> I'm considering that a temporary approach

I think i worth to mention it somewhere :+1:

luigifab:
So, I will continue to use it as is, and I closing because I will not work more on it, even if a "security report" page can be useful. I don't really know docker or other deployment scripts, and yes if they run some openmage scripts with _root_ user, this is a breaking change.

ADDISON74:
Thank you for your work @luigifab. The idea is not bad at all but it needs more consistency.

I am marking this PR with **"Don't forget this PR"** label. I did the same thing for other closed PRs. Such a mark helps us to return to a PR where there are good ideas, but it was closed for various reasons.

PR #2945, opened 2023-01-15
Added frontend_type color

fballiano:
In https://github.com/OpenMage/magento-lts/pull/2937 it came the idea of having support for input type color for adminhtml forms, this PR adds it.

To test it, in a system.xml file, add some rows like these ones:

```xml
  <colortest>
      <frontend_type>color</frontend_type>
      <with_hash>1</with_hash><!-- 1 is default, saves the data with "#" as first character, if is set to 0 the "#" is not saved -->
      <label>Color</label>
  </colortest>
```

then check the "system -> configuration" so see that it works, eg:

Screenshot 2023-01-15 alle 18 10 45

### Related Pull Requests
- https://github.com/OpenMage/magento-lts/pull/2937


justinbeaty:
Oh nice one 👍. Tested and it works just fine.

Only thing to note. In @luigifab's PR, he has the option of `color` or `color {hash:true}`. The latter includes the `#` sign. However, the HTML5 element always uses the `#` sign. I don't see a way to enable or disable that with the HTML5 element.

Here's the results in the database:
```
+-----------+---------+----------+--------------------------+---------+------------+
| config_id | scope   | scope_id | path                     | value   | updated_at |
+-----------+---------+----------+--------------------------+---------+------------+
|       107 | default |        0 | system/csrf/colortest    | #a34343 | NULL       |
|       137 | default |        0 | system/csrf/color1       | EF5FFF  | NULL       |
|       138 | default |        0 | system/csrf/color2       | #B23EFF | NULL       |
+-----------+---------+----------+--------------------------+---------+------------+
```

* `colortest` is this PR
* `color1` is jscolor without `{hash:true}`
* `color2` is jscolor with `{hash:true}`

But overall, I think this is a good addition, so I approve.

justinbeaty (approved):

fballiano:
I didn’t find anything in the specification about having the hash or not 😞

justinbeaty:
I ran this SQL, to update the value from `#ff0000` to `ff0000`
```sql
update core_config_data set value='ff0000' where path LIKE "%colortest";
```
Then when I reloaded the config page, the color picker was black instead of red.

So maybe, we can add this method to Color.php:

```php
/**
 * @param string|null $index
 * @return string
 */
public function getEscapedValue($index = null)
{
    $value = parent::getValue($index);
    return '#' . ltrim($value, '#');
}
```

This would allow someone to use the color picker element on a value that doesn't have the hash in the database. Of course the new value would end up with the hash, so I don't know if that's going to mess up their code then. But, this is a new feature so it's not breaking anybody's code automatically.

ADDISON74:
In HTML5 you just choose the color and the magic happens in the background. In jscolor you can both choose the color and enter it in a box. this makes the difference between the two. Adding a hash before the string makes it a hexadecimal value. More details here https://jscolor.com/docs/#api--opts--hash.

This PR is a great feature and I support it.

justinbeaty:
> In jscolor you can both choose the color and enter it in a box. this makes the difference between the two.

You can do that with HTML5 too (at least on Chrome), but it's a bit hidden:

image

ADDISON74:
In this case your proposal to check if a hexadecimal string is preceded by a hash appears to be logical. Maybe it should be updated in the database as well, but that would complicate things to be achieved.

ADDISON74:
In jscolor hash option is set to false by default

```js
color : function(target, prop) {


		this.required = true; // refuse empty values?
		this.adjust = true; // adjust value to uniform notation?
		this.hash = false; // prefix color with # symbol?
```

fballiano:
> I ran this SQL, to update the value from `#ff0000` to `ff0000`
> 
> ```sql
> update core_config_data set value='ff0000' where path LIKE "%colortest";
> ```
> 
> Then when I reloaded the config page, the color picker was black instead of red.
> 
> So maybe, we can add this method to Color.php:
> 
> ```
> /**
>  * @param string|null $index
>  * @return string
>  */
> public function getEscapedValue($index = null)
> {
>     $value = parent::getEscapedValue($index);
>     return '#' . ltrim($value, '#');
> }
> ```
> 
> This would allow someone to use the color picker element on a value that doesn't have the hash in the database. Of course the new value would end up with the hash, so I don't know if that's going to mess up their code then. But, this is a new feature so it's not breaking anybody's code automatically.
> 
> **Edit:** Fixed the above code to call `parent::getEscapedValue` instead of `parent::getValue`

great idea but has a problem, if my value is saved (initially) without the hash, I probably expect it to keep being saved without the hash...

I'll try to work something out

justinbeaty:
> great idea but has a problem, if my value is saved (initially) without the hash, I probably expect it to keep being saved without the hash...
> 
> I'll try to work something out

Yes, agreed it would be better to allow hash=true or hash=false. I suppose there's a way to do this with two inputs (one color and one text). I will see what you come up with.

fballiano:
I've committed a version that manages the `with_hash` configuration, like this:

```
  <colortest>
      <frontend_type>color</frontend_type>
      <with_hash>1</with_hash><!-- 1 is default, saves the data with "#" as first character, if is set to 0 the "#" is not saved -->
      <label>Color</label>
  </colortest>
```

it's done with an hidden input and a small bit of javascript. maybe not the super best way but at least I didn't have to add a backend model or anything more complicated.

what do you think?

justinbeaty:
That's pretty neat! One problem I encountered. Set `with_hash` to `1`, then select a color, save the config page, and save the page again without selecting a new color. Hash will be missing in the database.

justinbeaty:
This fixes it:
```php
        $html .= '<input id="' . $id . '" type="text" name="' . $this->getName()                                                                                                                       
              . '" value="' . ($with_hash ? '#' : '') . $valueWithoutHash . '" ' . '/>' . "\n";                                                                                                        
```

fballiano:
@justinbeaty great!!!

justinbeaty (dismissed):

ADDISON74:
I added the XML code in a system.xml file and got the color field in the Backend. Initially the color selector is set to RGB mode, but the user can switch to HEX. The hex value is predicated by #. 

I went through several scenarios

1. **with_hash=0 => the user deletes the hash and saves.** In the database the value has the hash in front of it.

2. **with_hash=1 => the user deletes the hash and saves.** In the database the value has the hash in front of it.

3. **delete the <with_hash> line from the systeml.xml file => the user deletes the hash and saves.** In the database the value has the hash in front of it.

Seeing this behavior I deleted the # in front of the value and I did not save. When I used again the color picker, the value had # in front of it, which was not added by this PR. This came from browser. This behavior is found in all Chromium-based browsers (Chrome, Edge, ...). In the case of Firefox, the color selector is totally different, it does not have the HEX option.

**Chromium-based**

chrome

**Firefox**

firefox

justinbeaty:
> In the database the value has the hash in front of it.

Do you confirm this phpMyAdmin or mysql cli client? In Chrome, the hex value will always have the hash in the color picker.

justinbeaty:
Also I have an idea for possible improvement, I'll test now.

justinbeaty:
Based on feedback by Addison, and how the jscolor picker works, I tried some improvements. This lets you also manually type in the hex code in addition to using the color picker, as well as adds some validation and styles.

image

sreichel:
> it's done with an hidden input and a small bit of javascript. maybe not the super best way but at least I didn't have to add a backend model or anything more complicated.
>
> what do you think?

I think we should add that models. :stuck_out_tongue: 

Would `Mage/Adminhtml/Model/System/Config/...` be a better place?

sreichel:
> it does not have the HEX option.

RGB values should be allowed too. (?)

fballiano:
The point was supporting the html5 color input type, not recreate another jscolor from scratch

sreichel:
Maybe i'v missunderstood @ADDISON74 comment regarding chromium-based browser. Do they set a RGB-value to imput field?

justinbeaty:
> The point was supporting the html5 color input type, not recreate another jscolor from scratch

I’m not sure if this comment is regarding the PR I made to your repo. If so, I didn’t intend to recreate jscolor, just to add the ability to type a hex code (especially on firefox), and then validation to ensure the user enters a valid hex code. If you don’t like it, it’s no problem and I have no hurt feelings.

sreichel:
> and then validation to ensure the user enters a valid hex code

... and i would add backend classes to validate it php-side. (with `_beforeSave`/`_afterLoad` we can ensure to save it with/without hash)

fballiano:
I've added @justinbeaty contributions and added translations.

@ADDISON74 I think with this new version there shouldn't be any of the problems you saw.

@sreichel after @justinbeaty modification do you think feel we need backend model?

ADDISON74:
I will test it today and provide a feedback. I will check what is stored into the database as @justinbeaty asked me before.

sreichel:
I've not tested it yet, but i prefer php-side validation.

(I can add a PR later, but then we've to decide what is prefered - w/ or /wo hash)

fballiano:
Default i’d keep with hash, since it’s what the W3 specification decided

fballiano:
I was added a backend model but I've no idea how can I know if the user wanted the hash or not, which is configured in the system.xml, but the backend model doesn't receive that configuration

```
class Mage_Adminhtml_Model_System_Config_Backend_Color extends Mage_Core_Model_Config_Data
{
    /**
     * @throws Mage_Core_Exception
     * @return $this
     */
    {
        if (!(bool)preg_match(Varien_Data_Form_Element_Color::VALIDATION_REGEX_WITH_HASH, (string)$this->getValue())) {
            Mage::throwException(Mage::helper('adminhtml')
                ->__('Color must be in hexadecimal format with hash'));
        }
        return $this;
    }
}

```

justinbeaty:
@fballiano This seems to work:

```php
    {                                                                                                                                                                                                  
        $config = $this->getFieldConfig();                                                                                                                                                             
        $with_hash = $config['with_hash']; 
        // ...
```

fballiano:
@justinbeaty thanks!!

fballiano:
Just added the backend model with the validation.

Question, if the user doesn't specify `<backend_model>adminhtml/system_config_backend_color</backend_model>`, is there a way to have it as a default for frontend_type color?

justinbeaty:
I see the option for catalog product edit screen, but nothing for core config:

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Helper/Product.php#L272-L285

justinbeaty:
We could hardcode it or add something like the catalog/product helper here:

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/Model/Config/Data.php#L146-L152

ADDISON74:
It is preferable that this value is not hardcoded.

justinbeaty:
It seems to work: https://github.com/justinbeaty/magento-lts/commit/30ec1e1998fe5792c54468a406b0db677de2a430

fballiano:
thans @justinbeaty! great code! at the moment I commited something much simpler since it would be the first time we add a default backend model, not to overcomplicate things and leave "todos" ;-)

fballiano:
> It is preferable that this value is not hardcoded.

@ADDISON74 it would be only the default, users can change it with the xml if they want :-)

ADDISON74:
I have to admit that it is interesting where it ended up starting from some discussions in another PR.

justinbeaty:
> thans @justinbeaty! great code! at the moment I commited something much simpler since it would be the first time we add a default backend model, not to overcomplicate things and leave "todos" ;-)

The todos came from the code I [copy & pasted](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Helper/Product.php#L242-L270) from. I don't know *when* the Magento team planned to do this. Clearly never. 😅 

One more thing though, we should have similar logic [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php#L301-L312), which might make it better to keep in the helper.

app/code/core/Mage/Adminhtml/Model/Config/Data.php
```diff
@@ -143,12 +143,16 @@ public function save()
...
+                // Get field backend model
+                $backendClass = 'core/config_data';
+                if (isset($fieldConfig->backend_model)) {
+                    $backendClass = (string)$fieldConfig->backend_model;
+                } elseif (isset($fieldConfig->frontend_type)) {
+                    switch ((string)$fieldConfig->frontend_type) {
+                        case 'color':
+                            $backendClass = 'adminhtml/system_config_backend_color';
+                            break;
+                    }
```
justinbeaty, line 155: Careful that this seems to change behavior. What if `<backend_model>` exists, but is an empty string? Before, we would have `core/config_data` but now an exception

justinbeaty: Actually, I'm wrong on it changing behavior, because `$fieldConfig->backend_model` will be a `SimpleXMLElement`, which is never falsy.

fballiano: ye.. I thought about that but said "who the heck has a `<backend_model></backend_model>` tag... and I liked the way this code looks more... should I revert this few lines?

sreichel: > should I revert this few lines?

@fballiano wait a few minutes ... PR/Suggestion ready in a few minutes

justinbeaty: @fballiano Did you see my previous comment about the other place we may need this logic at? https://github.com/justinbeaty/magento-lts/commit/684debc1649ae1a2fa4999d4c5382c85fc81b6d3

fballiano:
> The todos came from the code I [copy & pasted](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Helper/Product.php#L242-L270) from. I don't know _when_ the Magento team planned to do this. Clearly never. 😅

clearly never 😂😂😂

> One more thing though, we should have similar logic [here](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php#L301-L312), which might make it better to keep in the helper.

checking it out now since @sreichel added some very interesting things to this PR

sreichel:
Commited CS-Fix ...

Its similiar to @justinbeaty idea. :+1:

justinbeaty:
> checking it out now since @sreichel added some very interesting things to this PR

@sreichel's changes are pretty similar to the commit I linked to. I do like that he created a new helper `adminhtml/config`. But, I think the changes in my commit are better for a few reasons:

1. It is a bit more flexible, in case we wish to implement a `getAttributeSourceModelByInputType` function. [ref](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Catalog/Helper/Product.php#L287-L300)
2. It matches existing code in OM
3. The helper function can be reused in `app/code/core/Mage/Adminhtml/Block/System/Config/Form.php` without changing the behavior there.

justinbeaty:
> Its similiar to @justinbeaty idea. 👍

Ultimately both are good :)

app/code/core/Mage/Adminhtml/Model/Config/Data.php
sreichel: @justinbeaty unfortunately i saw it too late. Your're right.

fballiano:
@justinbeaty it seemed a bit too complicated, I didn't know we already had a similar code in the repo.

fballiano:
I don't know guys, you decide.

sreichel:
:1st_place_medal: @justinbeaty

justinbeaty:
I will make the PR to your repo, few minutes.

justinbeaty:
I made https://github.com/fballiano/openmage/pull/8

I merged @sreichel's new helper class with the functions from my old helper. You can view the file [here](https://github.com/fballiano/openmage/blob/821400ac6833c8a43024c703518fba6f4d87817c/app/code/core/Mage/Adminhtml/Helper/Config.php). @sreichel -- please just double check I used the strict typing correctly.

It should be all okay, I tested it, but I should give it fresh eyes tomorrow morning. It's close to end of day here now.

sreichel:
Tested. :+1:

sreichel:
> I have to admit that it is interesting where it ended up starting from some discussions in another PR.

:rofl: 

And at the and we have a good solution ... :)

- [x] input validation (can we set defaut `<validate>` nodes too ...?)
- [x] js-validation is ready
- [ ] documentation

sreichel (dismissed):

sreichel (approved):

sreichel:
Not required for this PR, but phpstorm meta files have to be update. (new class added)

justinbeaty:
A few more changes here: https://github.com/fballiano/openmage/pull/9

1. Check to see if we have `<validate>required-entry</validate>`. Before, the backend model threw an error that the value is not in hexadecimal format. But if it's not required, we should allow an empty value.
2. Fix `$with_bool` variable. Before, `$with_hash = (bool)$config->with_hash;` would always return `true`, because `Mage_Core_Model_Config_Element` is always truthy. So, if you set `<with_hash>0</with_hash>` you could not save a value.

I hope that's it.

justinbeaty:
I did not know this would end up so complicated, but it is a good template for adding frontend types.

justinbeaty (approved):

justinbeaty:
What if we go just a little bit further. A few more lines and we can have the color attribute for products, too: https://github.com/justinbeaty/magento-lts/commit/e795f1db9cc5f2896710b0ede1d3c6b7a3fde87b

image

image

Note: Since there is no way to allow the user to set `with_hash` to true or false, hash is always enabled here.

fballiano:
can we do that in a separate PR, when PRs become too big they never get approved and we waster everybody's time.

app/code/core/Mage/Adminhtml/Helper/Config.php
```diff
@@ -0,0 +1,88 @@
...
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Default config helper
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @author     Magento Core Team <core@magentocommerce.com>
```
kiatng, line 29: Is the author correct? Also, that email no longer work.

fballiano: we have the same on every file :-\ I don't remember if we did decide something about new files :-\

justinbeaty (approved):

sreichel (approved):

app/code/core/Mage/Adminhtml/Helper/Config.php
sreichel: @justinbeaty after writing command for updating docblocks ... :heart: 

... is it possible to extend your script to replace headers for all added files since last release to change the "author"?

sreichel: > I don't remember if we did decide something about new files :-\

We already had a discussion where someone wanted to add his company name ... cant find it atm. 

I'd prefer to have something like _OpenMage Developers_ ...

sreichel (dismissed):
Some nit-picking ...

- would `color_picker` or `color_selector` be better? (not needed)
- as said in a comment ... the new class `Varien_Data_Form_Element_Color` should be in `Mage_XYZ` extending `Varien_Data_Form_Element_Text`

sreichel:
I'd also add https://github.com/justinbeaty/magento-lts/commit/e795f1db9cc5f2896710b0ede1d3c6b7a3fde87b b/c this lets you idea shine :) :sparkles:

fballiano:
> * would `color_picker` or `color_selector` be better? (not needed)

the picker is only a way to input a color, I think "color" is correct, it's also the name used by html5

> * as said in a comment ... the new class `Varien_Data_Form_Element_Color` should be in `Mage_XYZ` extending `Varien_Data_Form_Element_Text`

why tho?

app/code/core/Mage/Adminhtml/Helper/Config.php
fballiano: this should be discussed in a proper discussion and/or rfc, I wouldn't block this PR which was supposed to be simple.

fballiano: also, we should ask adobe to speak about this, we shouldn't change the copyright but it doesn't make sense to keep these one...

fballiano:
I've committed @justinbeaty's code for managing color also in products' attributes

justinbeaty:
> > * would `color_picker` or `color_selector` be better? (not needed)
> 
> the picker is only a way to input a color, I think "color" is correct, it's also the name used by html5

I agree `color` is correct, because we have `date` not `date_selector`, as well as `file` not `file_picker`, etc.

> 
> > * as said in a comment ... the new class `Varien_Data_Form_Element_Color` should be in `Mage_XYZ` extending `Varien_Data_Form_Element_Text`
> 
> why tho?

I also think `lib/varien` is good. At this point, `lib/varien` *is* Magento. I assume the company Varien never used this library for anything else and we would likely never split `lib/varien` from this repo. But if someone else ever did use `lib/varien`, then they could get a nice color picker. Plus, there are no references to `Mage::` in the new class; if there were, then I'd agree it would be a `Mage_Xyz` class instead.


> > @justinbeaty after writing command for updating docblocks ... ❤️
> >
> > ... is it possible to extend your script to replace headers for all added files since last release to change the "author"?
>
> this should be discussed in a proper discussion and/or rfc, I wouldn't block this PR which was supposed to be simple.

The script can be extended, but let's make a discussion about it.

> I've committed @justinbeaty's code for managing color also in products' attributes

Thanks, although I was okay with making a separate PR. I should do a bit more testing to see how this attribute shows up in collections, etc. I don't foresee anything negative.

justinbeaty:
@fballiano Let's revert `Suggestions by justing, adding color also for products attributes 1fd0e08` and I will add it to a new PR after this is merged. There are improvements to be made there, plus I also see the commit is missing this file: `app/code/core/Mage/Eav/Model/Entity/Attribute/Backend/Color.php`.

fballiano:
done

justinbeaty (approved):

sreichel (approved):

fballiano:
what are we doing with this PR? trash it?

elidrissidev (approved):

fballiano:
yeah!

PR #2950, opened 2023-01-16
Added counter for input fields with maximum-length attribute

sreichel:
### Description (*)

Adds js-snippet to ...

- admin product edit page
- admin product attributes edit page
- admin config section

... to show how many chars are left for valid input.

__Note:__ requires `validate-length` and `maximum-length-XXX` attribute

counter

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2952, opened 2023-01-16
Fix less than typo

justinbeaty:
### Description (*)
Fixed "less then" -> "less than" typo. Maybe it should actually be "must be less than or equal to %s symbols" though.

### Related Pull Requests
#2950


fballiano (approved):

justinbeaty:
The two places I see this used are in attribute code and customer group name. Both allow the exact maximum length characters, so "less than or equal to" would be correct, although more verbose.

sreichel (approved):

fballiano:

PR #2447, opened 2022-08-18
Added two indexes on sales_flat_order and sales_flat_order_item tables

fballiano:
This solution was already discussed and agreed in issue https://github.com/OpenMage/magento-lts/issues/384 I just created the PR into the Mage_Sales module since in 2022 I don't see the need for a new dedicated module for this specific task (as discussed in the previously linked issue).

PR #1110, opened 2020-07-13
Improving query performance for coupon code check

lemundo-team:
Adding coupon code to where clause in order to improve query performance

### Description (*)
This was one of the major bottle necks over the last few Black Fridays! Thousands of customers applying coupon codes in checkout and the checkout is causing causing more and more load. 

It might be that this query is only leading to bad performance when using certain MySQL versions like https://github.com/OpenMage/magento-lts/issues/295 does... 

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/303

### Manual testing scenarios (*)
1. Let 1000 customers apply coupon codes every minute
2. Checkout the slow query log

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #2959, opened 2023-01-17
Add index on salesrule_coupon

luigifab:
### Description

We have 220,000 coupons in `salesrule_coupon`.
This PR add an index on `code` column, like #2447.

Here is a before/after with Blackfire:

coupons-before-after

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/SalesRule/etc/config.xml
+++ b/app/code/core/Mage/SalesRule/etc/config.xml
@@ -23,7 +23,7 @@
 <config>
     <modules>
         <Mage_SalesRule>
-            <version>1.6.0.3</version>
+            <version>1.6.0.4</version>
         </Mage_SalesRule>
     </modules>
     <global>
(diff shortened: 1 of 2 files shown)


tmotyl:
what query is being run?

luigifab:
Yes, I forgot to copy/paste: `select salesrule_coupon.* from salesrule_coupon wehere (salesrule_coupon.code = ?)`

kiatng:
The column `code` is `UNIQUE KEY `UNQ_SALESRULE_COUPON_CODE` (`code`),`, according to this [stackoveflow](https://stackoverflow.com/questions/9393234/mysql-unique-field-needs-to-be-an-index#:~:text=No%2C%20you%20dont%20have%20to,will%20be%20little%20performance%20difference.), it's unnecessary to index it again. I am not sure how to interpret Blackfire output, is stackoverflow wrong?

luigifab:
Damned, you are right. Now I remember that the store use a module that removed the unique key to allow multiple coupon with same code.

sreichel:
> but I still think we should merge https://github.com/OpenMage/magento-lts/pull/2447

We already merged it.

fballiano:
I was meaning https://github.com/OpenMage/magento-lts/pull/1110 :-D

PR #2982, opened 2023-01-19
Simplify PHP version error message in index.php, install.php, api.php…

justinbeaty:
### Description (*)
The hardcoded HTML in some root files has always bothered me. First, it contains broken html: (two closing `</a>`)

```
<a href="https://www.openmage.org/magento-lts/install.html" target="">Find out</a> how to install</a>
```

But also, we really don't need those inline styles. It looks fine considering nobody should see it.

image

I also created the `$minVersion` variable so we don't have to change it twice in each file.

1. Change `$minVersion` to `9.0.0` or similar to see the error message.

### Questions or comments
Can we move `$minVersion` to some file that gets included so we don't have to duplicate it?

### DIFF
--- a/install.php
+++ b/install.php
@@ -113,9 +113,13 @@
  *
  */
 
-if (version_compare(phpversion(), '7.3.0', '<') === true) {
-    die('ERROR: Whoops, it looks like you have an invalid PHP version. OpenMage supports PHP 7.3.0 or newer.');
+$minVersion = '7.3.0';
+if (version_compare(phpversion(), $minVersion, '<') === true) {
+    echo "<h3>Whoops, it looks like you have an invalid PHP version.</h3>";
+    echo "<p>OpenMage supports PHP $minVersion or newer. <a href=\"https://www.openmage.org/magento-lts/install.html\">Find out how to install</a> OpenMage using PHP-CGI as a work-around.</p>";
+    exit;
 }
+
 set_include_path(__DIR__ . PATH_SEPARATOR . get_include_path());
 require 'app/bootstrap.php';
 require 'app/Mage.php';
--- a/api.php
+++ b/api.php
@@ -19,8 +19,9 @@
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-if (version_compare(phpversion(), '7.3.0', '<') === true) {
-    echo 'It looks like you have an invalid PHP version. OpenMage supports PHP 7.3.0 or newer';
+$minVersion = '7.3.0';
+if (version_compare(phpversion(), $minVersion, '<') === true) {
+    echo "It looks like you have an invalid PHP version. OpenMage supports PHP $minVersion or newer";
     exit;
 }
 
--- a/get.php
+++ b/get.php
@@ -18,15 +18,14 @@
  * @copyright  Copyright (c) 2016-2022 The OpenMage Contributors (https://www.openmage.org)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
-if (version_compare(phpversion(), '7.3.0', '<') === true) {
-    echo  '<div style="font:12px/1.35em arial, helvetica, sans-serif;"><div style="margin:0 0 25px 0; '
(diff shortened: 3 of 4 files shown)


sreichel:
> Can we move $minVersion to some file that gets included so we don't have to duplicate it?

I guess until `Mage:app()` isnt called, we cannot.

justinbeaty:
> > Can we move $minVersion to some file that gets included so we don't have to duplicate it?
> 
> I guess until `Mage:app()` isnt called, we cannot.

I was thinking like:

`config.php`:
```php
<?php
$minVersion = '7.3.0';
```

and in `[index|install|api|get].php`:
```php
require_once __DIR__ . '/config.php';
if (version_compare(phpversion(), $minVersion, '<') === true) {
```

But, it's kind of pointless to have a `config.php` with just one variable and it's easy enough to update in each of the four files whenever the min PHP version changes. I was just kind of throwing out ideas.

sreichel:
I can live w/o, but can we add styling to some css-files? :sunglasses:

fballiano:
In my opinion I would remove that message, we don’t check php extensions or any other thing, only php version.. I don’t see its utility this much as to have it in every call.

justinbeaty:
I wouldn't bother to add any CSS to the message, literally nobody should ever see it.

I also don't care if we remove it, I already do that in my local fork. But perhaps more people should put in their opinion on removing it.

justinbeaty:
Converting to draft since I think (if we choose to keep it) that we can put it into bootstrap.php. Maybe also check for PHP modules, which could be useful now that IIRC we require intl.

fballiano:
As we were saying in another pr we could have a system check page instead of checking at every request

justinbeaty:
> As we were saying in another pr we could have a system check page instead of checking at every request

Yes, but if the error is an incompatible PHP version or a missing module, it’s possible the user may not be able to reach that page.

Anyway, I will be off until tomorrow morning. Will modify this PR then, and then it’s up to the community on what direction to go.

fballiano:
It could be memory setting or many other parameters. Some projects have a system check script to call on cli

sreichel:
> As we were saying in another pr we could have a system check page instead of checking at every request

N98 magerun provides `sys:check` ... head over here https://github.com/netz98/n98-magerun/issues/1275 and lets finish this.

luigifab:
I suggest to move it to _Mage.php_ (like my root check even you are unhappy), or to drop it, :).
When OpenMage is installed with composer, composer check PHP version at every request.

install.php
```diff
@@ -113,9 +113,13 @@
  *
  */
 
-if (version_compare(phpversion(), '7.3.0', '<') === true) {
-    die('ERROR: Whoops, it looks like you have an invalid PHP version. OpenMage supports PHP 7.3.0 or newer.');
```
elidrissidev, line 117: `install.php` is meant to be invoked from the cli, I think that's why there wasn't HTML before.

justinbeaty:
> When OpenMage is installed with composer, composer check PHP version at every request.

@luigifab is right. As long as you don't install with `--ignore-platform-reqs` (although `--ignore-platform-req=ext-*` is okay) then the file `vendor/composer/platform_check.php` checks PHP versions in the autoloader.

Since the release builder has the composer autoloader, then all new releases will check PHP version already. So I would vote to close this PR and open another to simply remove these lines from the root files.

Although one small note: `platform_check.php` wants `>= 7.4.0`. This is because: OpenMage requires [magento-composer-installer](https://github.com/Cotya/magento-composer-installer/blob/main/composer.json#L63), which requires [symfony/console](https://github.com/symfony/console/blob/5.4/composer.json#L24), which requires [symfony/service-contracts](https://github.com/symfony/service-contracts/blob/3.2/composer.json#L20), which requires [psr/container](https://github.com/php-fig/container/blob/2.0.2/composer.json#L15), which requires `"php": ">=7.4.0"`.

So we should bump the min required version, or remove `magento-composer-installer` in the built files which is something I tried to do here: https://github.com/fballiano/openmage/pull/3

fballiano:
let's remove these lines and bump to 7.4, i'm fine with those

PR #2987, opened 2023-01-20
Fix esi parsing with turpentine

luigifab:
### Description

With Nexcess Magento Turpentine for Varnish, sometimes there is a strange bug with ESI processing (ESI is not processed).

Example: `window.location='https://.../fr/customer/address/delete/form_key/<esi:include src='http://.../fr/turpentine/esi/getFormKey/ttl/86400/method/esi/scope/global/access/private/' />/id/'+addressId;`

In #2682, the solution is to invert ' and ".

But in fact, this is right and wrong (this must be done from Turpentine side).
The real solution is to remove `CDATA` tags (https://stackoverflow.com/a/4504180/2980105).
I think I have tested without and added them for the PR.

### Contribution checklist


fballiano (approved):

kiatng (dismissed):

colinmollenhour:
If the CDATA markers are removed doesn't everything need to be htmlescaped? E.g. `"` should either be inside a CDATA or escaped as `&quot;`?

luigifab:
My html validator is okay without the CDATA:

image

elidrissidev:
AFAIK, CDATA is required only for XML parsers since they consider any appearance of `</` as a closing tag, even if it's inside quotes in JS.

Also see: https://stackoverflow.com/a/66865

colinmollenhour:
> Also see: https://stackoverflow.com/a/66865
There are a lot of answers there and I'm still confused as to what's "best" and it probably depends on your opening tag as well..

For the purpose of keeping your PR to one issue and not conflating another I think it would be best to separate the CDATA discussion from this pull request by reverting the CDATA change so it doesn't hold up your PR. Then if CDATA should be removed we could remove them everywhere, or just leave them and not fret over it.

app/design/frontend/base/default/template/customer/address/book.phtml
```diff
@@ -89,12 +89,10 @@
     <p class="back-link"><a href="<?php echo $this->escapeUrl($this->getBackUrl()) ?>"><small>&laquo; </small><?php echo $this->__('Back') ?></a></p>
 </div>
 <script type="text/javascript">
-//<![CDATA[
     function deleteAddress(addressId) {
```
fballiano, line 92:
```suggestion
//<![CDATA[
    function deleteAddress(addressId) {
```

fballiano (approved):
I've re-added CDATA as per the discussion so we can approve it and merge it

ADDISON74 (approved):

luigifab:
Oh my bad. The CDATA was not removed in _app/design/frontend/base/default/template/customer/address/book.phtml_!

Issue #2204, opened 2022-06-07
OM is not reloading the cart page if deleting a product from cart side block


ADDISON74:
### Steps to reproduce

1. Add a product to cart
2. If you are not redirected go to cart page
3. Delete the product from cart side block by clicking "Remove Item" link

### Expected result

1.  Cart page should be reloaded automatically showing no product

### Actual result

1. Cart page still shows the product in cart, even the cart side block is empty
2. If you reload the page there is no product listed, as it should be

**This is before removing the product**

![before](https://user-images.githubusercontent.com/8360474/172480708-cbdbe979-bd3a-4ddf-8e56-1e778b0096a7.jpg)

**This is after removing the product**

![after](https://user-images.githubusercontent.com/8360474/172480735-acbbd358-ca7e-4704-b07d-4e761e83318d.jpg)

**Comments**

1. The Magento 2 team fixed this issue last year. Here is the merged PR https://github.com/magento/magento2/pull/22478.

2. When I did the reporting in Magento 2 group I had an exchange of ideas with a gentleman without manners. He claimed that it was a bad UX practice to use that side block, that no one was using it anymore, but he did not offer any solution. Fortunately, the Magento team had take in consideration the issue.

**Personal Opinions**

1. I do not agree with the removal of the side block dedicated to the cart because it is helpful for tablets and desktops.

2. If you do not want a page reload implementation then the link "Remove Item" should be deleted to avoid the issue. Thus the removal of the products from the cart will be done from grid not from the side block.

3. I would go for the version proposed by the Magento team keeping the the link "Remove Item" but if it is too complicated to implement the code I suggest its removing.

Flyingmana:
2 more options:

4. remove the link only, if on a cart page
5. do the reload only on the cart page


I prefer 5. and it shouldnt be hard. Either we check for the url, or for the existence of cart specific `class=` elements to decide regarding the reload.

fballiano:
The patch used in M2 does point number 5:

<img width="734" alt="Screenshot 2023-01-21 alle 20 17 49" src="https://user-images.githubusercontent.com/909743/213885640-a9611b84-1c40-4239-89a3-9f6a4b0da209.png">

it shouldn't be that hard to do, let's see if I can get it to work

PR #2990, opened 2023-01-21
Added a forced refresh if the user modifies the minicart while being on the cart page

fballiano:
The problem is described in issue #2204.

I've checked the body class for the presence of "checkout-cart-index", it should be quite ok.

I've used `document.body.classList.contains("checkout-cart-index")` functionality that is compatible with more modern browsers (ie10), is it ok with everybody? otherwise we've to use regex like `document.body.className.match(/(\s|^)checkout-cart-index(\s|$)/)`

This PR fixes the problem when removing the product but also when changing its quantity.

### DIFF
--- a/skin/frontend/rwd/default/js/minicart.js
+++ b/skin/frontend/rwd/default/js/minicart.js
@@ -91,6 +91,7 @@ Minicart.prototype = {
             }).done(function(result) {
                 cart.hideOverlay();
                 if (result.success) {
+                    cart.refreshIfOnCartPage();
                     cart.updateCartQty(result.qty);
                     cart.updateContentOnRemove(result, el.closest('li'));
                 } else {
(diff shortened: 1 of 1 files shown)


justinbeaty:
> I've used `document.body.classList.contains("checkout-cart-index")` functionality that is compatible with more modern browsers (ie10), is it ok with everybody?

Definitely okay IMO. IE9 is EOL for 3 years now.

ADDISON74:
LGTM but I will test it tomorrow.

I've been looking through the webserver logs for the last year and I haven't found any monthly records related to IE 9 and 10. Only IE 11 around 20 visits per month, most likely a Windows 7 or 10 user who didn't want to upgrade to a modern browser.

elidrissidev (approved):
Works fine for me.

ADDISON74 (approved):
I tested this PR and the issue is solved. I found no errors in the browser console.

fballiano:
thanks guys, cherrypicked to v20

Issue #2044, opened 2022-03-30
PayPal Express Checkout Token does not exist - allowing multiple orders through


trabulium:
### Preconditions (*)
OpenMage LTS 19.4.x

1. Samesite = LAX - I can't give much further details how to replicate.

### Steps to reproduce (*)

1. We've been unable to reproduce. This is on a store with over 1M orders and never seen this issue until now
2. It seems to be related to some sort of race condition. View Nginx log here:

[30/Mar/2022:13:57:12 +1100] "GET /paypal/express/review/ HTTP/1.1" "AU" 200 36096
[30/Mar/2022:13:57:23 +1100] "POST /paypal/express/placeOrder/ HTTP/1.1" "AU" 302 5 
[30/Mar/2022:13:57:23 +1100] "POST /paypal/express/saveShippingMethod/ HTTP/1.1" "AU" 302 5 
[30/Mar/2022:13:57:23 +1100] "GET /paypal/express/review/ HTTP/1.1" "AU" 200 36095 
.....

where the saveShippingMethod hits after placeOrder but before the 302 redirect occurs (user was on Firefox). 

Mage2 users seem to be reporting the same issue and a user states "SameSite = None" resolves the issue for him. We are running SameSite LAX from this merge: https://github.com/OpenMage/magento-lts/pull/1246/files

 I'm not convinced this would fix the issue for us but since it's a 1 in a Million bug for us that we can't replicate, it's hard to determine.
https://github.com/magento/magento2/issues/28916

### Expected result (*)

1. Error is shown to the user OR
2. Order goes through without any error

### Actual result (*)
1. We received 4 orders, each said they were paid  (one for each /placeOrder attempt) with the same reference number from Paypal

I'm not expecting any bugfix soon on this, rather if someone encounters the issue, it can help them with pointers to debug easier.

fballiano:
web/cookie/cookie_samesite allows you to configure the samesite, did you test with none?

PR #2993, opened 2023-01-24
[WIP] EAV Config Cache

davidhiendl:
DO NOT MERGE YET, TESTING INSUFFICIENT

### Description (*)
Cache EAV entity and attributes as well as complementary data like store labels and attribute set information. This will eliminate several database queries per request and replace them with a single cache request.

### Related Pull Requests
#532 similar PR (likely abandoned) with some similarities 

TBD, probably lots in addition to tests

### DIFF
--- a/app/code/core/Mage/Api2/Model/Resource.php
+++ b/app/code/core/Mage/Api2/Model/Resource.php
@@ -1030,7 +1030,7 @@ abstract class Mage_Api2_Model_Resource
         $model = $this->getConfig()->getResourceWorkingModel($this->getResourceType());
 
         /** @var Mage_Eav_Model_Entity_Type $entityType */
-        $entityType = Mage::getModel('eav/entity_type')->load($model, 'entity_model');
+        $entityType = Mage::getSingleton('eav/config')->getEntityType($model, 'entity_model');
 
         /** @var Mage_Eav_Model_Entity_Attribute $attribute */
         foreach ($entityType->getAttributeCollection() as $attribute) {
--- a/app/code/core/Mage/Catalog/Model/Layer.php
+++ b/app/code/core/Mage/Catalog/Model/Layer.php
@@ -202,7 +202,7 @@ class Mage_Catalog_Model_Layer extends Varien_Object
     /**
      * Get collection of all filterable attributes for layer products set
      *
-     * @return Mage_Catalog_Model_Resource_Product_Attribute_Collection|array
+     * @return array
      */
     public function getFilterableAttributes()
     {
@@ -210,17 +210,26 @@ class Mage_Catalog_Model_Layer extends Varien_Object
         if (!$setIds) {
             return [];
         }
-        /** @var Mage_Catalog_Model_Resource_Product_Attribute_Collection $collection */
-        $collection = Mage::getResourceModel('catalog/product_attribute_collection');
-        $collection
-            ->setItemObjectClass('catalog/resource_eav_attribute')
-            ->setAttributeSetFilter($setIds)
-            ->addStoreLabel(Mage::app()->getStore()->getId())
-            ->setOrder('position', 'ASC');
-        $collection = $this->_prepareAttributeCollection($collection);
-        $collection->load();
 
-        return $collection;
+        $eavConfig=Mage::getSingleton('eav/config');
+        /** @var Mage_Catalog_Model_Resource_Eav_Attribute[] $attributes */
+        $attributes = [];
+        foreach($setIds as $setId) {
+            $setAttributeIds = $eavConfig->getAttributeSetAttributeIds($setId);
+            foreach($setAttributeIds as $attributeId) {
+                if(!isset($attributes[$attributeId])) {
+                    $attribute = $eavConfig->getAttribute(Mage_Catalog_Model_Product::ENTITY, $attributeId);
+                    if($attribute->getIsFilterable()) {
+                        $attributes[$attributeId] = $attribute;
+                    }
+                }
+            }
+        }
+        usort($attributes, function ($a,$b) {
+           return $a->getPosition() - $b->getPosition();
+        });
+
+        return $attributes;
     }
 
     /**
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Flat/Indexer.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Flat/Indexer.php
@@ -264,8 +264,8 @@ class Mage_Catalog_Model_Resource_Product_Flat_Indexer extends Mage_Index_Model_
     {
         $attributes = $this->getAttributes();
         if (!isset($attributes[$attributeCode])) {
-            $attribute = Mage::getModel('catalog/resource_eav_attribute')
-                ->loadByCode($this->getEntityTypeId(), $attributeCode);
+            $attribute = Mage::getSingleton('eav/config')
+                ->getAttribute($this->getEntityTypeId(), $attributeCode);
             if (!$attribute->getId()) {
                 Mage::throwException(Mage::helper('catalog')->__('Invalid attribute %s', $attributeCode));
             }
--- a/app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
+++ b/app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
@@ -70,8 +70,8 @@ class Mage_Downloadable_Block_Adminhtml_Catalog_Product_Edit_Tab_Downloadable_Li
         if (is_null($this->_purchasedSeparatelyAttribute)) {
             $_attributeCode = 'links_purchased_separately';
 
-            $this->_purchasedSeparatelyAttribute = Mage::getModel('eav/entity_attribute')
-                ->loadByCode(Mage_Catalog_Model_Product::ENTITY, $_attributeCode);
+            $this->_purchasedSeparatelyAttribute = Mage::getSingleton('eav/config')
+                ->getAttribute(Mage_Catalog_Model_Product::ENTITY, $_attributeCode);
         }
 
         return $this->_purchasedSeparatelyAttribute;
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -28,191 +28,255 @@
  */
 class Mage_Eav_Model_Config
 {
-    public const ENTITIES_CACHE_ID     = 'EAV_ENTITY_TYPES';
-    public const ATTRIBUTES_CACHE_ID   = 'EAV_ENTITY_ATTRIBUTES';
+    /**
+     * Cache key for storing:
+     * - EAV entities
+     * - EAV attributes
+     * - EAV attribute extension table data
+     */
+    public const EAV_DATA_CACHE = 'EAV_DATA_CACHE';
+
+    public const ENTITIES_CACHE_ID = 'EAV_ENTITY_TYPES';
+
+    protected $_storeInitialized = [];
 
     /**
-     * Entity types data
-     *
-     * @var array|null
+     * @var Mage_Eav_Model_Entity_Type[]|null
      */
-    protected $_entityData;
+    protected $_entityTypes;
 
     /**
-     * Attributes data
-     *
-     * @var array|null
+     * @var Mage_Eav_Model_Entity_Attribute_Abstract[][][]|null
      */
-    protected $_attributeData;
+    protected $_entityTypeAttributes;
 
     /**
-     * Information about preloaded attributes
-     *
-     * @var array
+     * @var Mage_Eav_Model_Entity_Attribute_Abstract[][][]|null
      */
-    protected $_preloadedAttributes              = [];
+    protected $_entityTypeAttributeIdByCode;
 
     /**
-     * Information about entity types with initialized attributes
-     *
-     * @var array
+     * Attribute set relation information. Structure:
+     * <br/>
+     * [
+     *  int attribute_id => [
+     *      int set_id => [
+     *          int group_id,
+     *          int group_sort,
+     *          int sort,
+     *      ]
+     * ]
+     * @var [][][][]mixed
      */
-    protected $_initializedAttributes            = [];
+    protected $_attributeSetInfo;
 
     /**
-     * Attribute codes cache array
-     *
-     * @var array
+     * Special local cache for default attributes to avoid re-hydrating them
+     * @var Mage_Eav_Model_Entity_Attribute_Abstract[][][]
      */
-    protected $_attributeCodes                   = [];
-
-    /**
-     * Initialized objects
-     *
-     * array ($objectId => $object)
-     *
-     * @var array|null
-     */
-    protected $_objects;
-
-    /**
-     * References between codes and identifiers
-     *
-     * array (
-     *      'attributes'=> array ($attributeId => $attributeCode),
-     *      'entities'  => array ($entityId => $entityCode)
-     * )
-     *
-     * @var array|null
-     */
-    protected $_references;
+    protected $_defaultAttributes = [];
 
     /**
      * Cache flag
      *
      * @var bool
      */
-    protected $_isCacheEnabled                    = null;
-
-    /**
-     * Array of attributes objects used in collections
-     *
-     * @var array
-     */
-    protected $_collectionAttributes              = [];
+    protected $_isCacheEnabled = null;
 
     /**
      * Reset object state
      *
-     * @deprecated
      * @return $this
+     * @deprecated
      */
     public function clear()
     {
-        $this->_entityData            = null;
-        $this->_attributeData         = null;
-        $this->_objects               = null;
-        $this->_references            = null;
-        $this->_preloadedAttributes   = [];
-        $this->_initializedAttributes = [];
+        $this->_entityTypes = null;
+        $this->_entityTypeAttributes = null;
+        $this->_entityTypeAttributeIdByCode = null;
+        $this->_storeInitialized = [];
+
+        Mage::app()->cleanCache([self::ENTITIES_CACHE_ID]);
+        // Mage::app()->removeCache(self::ENTITIES_CACHE_ID . "_" . Mage::app()->getStore()->getId());
+
         return $this;
     }
 
-    /**
-     * Get object by identifier
-     *
-     * @param   mixed $id
-     * @return  mixed
-     */
-    protected function _load($id)
+    public function __construct()
     {
-        return $this->_objects[$id] ?? null;
+    }
+
+    protected function _initializeStore($storeId = null)
+    {
+
+        if ($storeId === null) {
+            $storeId = Mage::app()->getStore()->getId();
+        } else {
+            // ensure store id is consistent
+            $storeId = (int)$storeId;
+        }
+        if (isset($this->_storeInitialized[$storeId]) && $this->_storeInitialized[$storeId]) {
+            return;
+        }
+
+        Varien_Profiler::start('EAV: ' . __METHOD__);
+
+        if ($this->_isCacheEnabled()) {
+            $this->_loadFromCache($storeId);
+        }
+
+        if (empty($this->_entityTypes)) {
+            $this->_loadEntityTypes();
+            $this->_loadAttributeSetInfo();
+        }
+
+        // load each entity attributes for given storeId
+        if (empty($this->_entityTypeAttributes[$storeId])) {
+            $this->_entityTypeAttributes[$storeId] = [];
+            foreach ($this->_entityTypes as $entityType) {
+                $this->_loadEntityAttributes($entityType, $storeId);
+                // $this->_loadEntityDefaultAttributes($entityType, $storeId);
+            }
+        }
+
+        if ($this->_isCacheEnabled()) {
+            $this->_saveToCache($storeId);
+        }
+
+        $this->_storeInitialized[$storeId] = true;
+
+        Varien_Profiler::stop('EAV: ' . __METHOD__);
     }
 
     /**
-     * Associate object with identifier
-     *
-     * @param   mixed $obj
-     * @param   mixed $id
-     * @return  Mage_Eav_Model_Config
+     * @param $storeId
+     * @return void
+     * @throws Exception
      */
-    protected function _save($obj, $id)
+    protected function _loadEntityTypes()
     {
-        $this->_objects[$id] = $obj;
-        return $this;
+        Varien_Profiler::start('EAV: ' . __METHOD__);
+
+        // load entity types
+        $this->_entityTypes = [];
+        $entityTypeCollection = Mage::getResourceModel('eav/entity_type_collection');
+        /** @var $entityType Mage_Eav_Model_Entity_Type */
+        foreach ($entityTypeCollection as $entityType) {
+            // if (empty($entityType->getAttributeModel())) {
+            //     $entityType->setAttributeModel('eav/entity_attribute');
+            // }
+            $this->_entityTypes[$entityType->getId()] = $entityType;
+        }
+
+        Varien_Profiler::stop('EAV: ' . __METHOD__);
     }
 
     /**
-     * Specify reference for entity type id
-     *
-     * @param   int $id
-     * @param   string $code
-     * @return  Mage_Eav_Model_Config
+     * @param Mage_Eav_Model_Entity_Type $entityType
+     * @param int $storeId
+     * @return void
+     * @throws Exception
      */
-    protected function _addEntityTypeReference($id, $code)
+    protected function _loadEntityAttributes($entityType, $storeId)
     {
-        $this->_references['entity'][$id] = $code;
-        return $this;
+        // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
+        $entityAttributes = $entityType->getAttributeCollection()
+            ->addStoreLabel($storeId)
+            ->getData();
+
+        $this->_entityTypeAttributes[$storeId][$entityType->getId()] = [];
+        $attributeCodes = [];
+
+        /** @var Mage_Eav_Model_Entity_Attribute_Abstract $entityAttributeData */
+        foreach ($entityAttributes as $entityAttributeData) {
+            $attributeId = $entityAttributeData['attribute_id'];
+            $attributeCode = $entityAttributeData['attribute_code'];
+            $this->_entityTypeAttributes[$storeId][$entityType->getId()][$attributeId] = $entityAttributeData;
+            $this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()][$attributeCode] = $attributeId;
+            $attributeCodes[] = $attributeCode;
+        }
+
+        $entityType->setAttributeCodes($attributeCodes);
     }
 
     /**
-     * Get entity type code by id
-     *
-     * @param   int $id
-     * @return  string
+     * @param $storeId
+     * @return void
+     * @throws Exception
      */
-    protected function _getEntityTypeReference($id)
+    protected function _loadAttributeSetInfo()
     {
-        return $this->_references['entity'][$id] ?? null;
+        Varien_Profiler::start('EAV: ' . __METHOD__);
+
+        $this->_attributeSetInfo = Mage::getResourceModel('eav/entity_attribute_set')->getSetInfo();
+
+        Varien_Profiler::stop('EAV: ' . __METHOD__);
     }
 
-    /**
-     * Specify reference between entity attribute id and attribute code
-     *
-     * @param   int $id
-     * @param   string $code
-     * @param   string $entityTypeCode
-     * @return  Mage_Eav_Model_Config
-     */
-    protected function _addAttributeReference($id, $code, $entityTypeCode)
+    protected function _loadFromCache($storeId)
     {
-        $this->_references['attribute'][$entityTypeCode][$id] = $code;
-        return $this;
+        Varien_Profiler::start('EAV: ' . __METHOD__);
+
+        $cacheData = Mage::app()->loadCache(self::ENTITIES_CACHE_ID . "_" . $storeId);
+        if ($cacheData === false) {
+            return;
+        }
+        $cacheData = unserialize($cacheData);
+
+        $this->_entityTypes = [];
+        /** @var $entityType Mage_Eav_Model_Entity_Type */
+        foreach ($cacheData['_entityTypes'] as $entityTypeData) {
+            $entityType = Mage::getModel('eav/entity_type')
+                ->setData($entityTypeData);
+            $this->_entityTypes[$entityType->getId()] = $entityType;
+        }
+
+        $this->_entityTypeAttributes[$storeId] = [];
+        /** @var $entityType Mage_Eav_Model_Entity_Type */
+        foreach ($cacheData['_entityTypeAttributes'] as $entityTypeId => $entityTypeAttributes) {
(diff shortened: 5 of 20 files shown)


davidhiendl:
@luigifab It's mostly done, there is maybe one or two small improvements but they may break stuff so I will hold off on them for now.

app/code/core/Mage/Catalog/Model/Layer.php
```diff
@@ -202,25 +202,34 @@ public function getCurrentStore()
     /**
      * Get collection of all filterable attributes for layer products set
      *
-     * @return Mage_Catalog_Model_Resource_Product_Attribute_Collection|array
+     * @return array
```
tmotyl, line 205: This is potentially a breaking change, isnt it?

luigifab:
Thanks! I applied, and for now, I have, (I haven't investigated yet, I will do):
- `Invalid attribute requested: url` on product page
- `Warning: Undefined array key 2`  in app/code/core/Mage/Eav/Model/Config.php on line 488 on home page

app/code/core/Mage/Catalog/Model/Layer.php
davidhiendl: Yes, but when you look at where/how it is used I except it to not cause issues. But I could also create a collection and fill the data by hand if needed to enhance compatibility, but I doubt its necessary.

davidhiendl:
> Thanks! I applied, and for now, I have, (I haven't investigated yet, I will do):
> 
> * `Invalid attribute requested: url` on product page
> * `Warning: Undefined array key 2`  in app/code/core/Mage/Eav/Model/Config.php on line 488 on home page

@luigifab 
- The 1st error I cannot reproduce. Is that with a default openmage installation (no custom modules/theme)? Because I didn't have those when testing.
- The 2nd error is an oversight for the fallback resolve of entity attributes when there is no store_id/attribute_set_id specified.

davidhiendl:
After some minors adjustments I think this is ready for testing by someone else than me. I've already started testing it in a real shop (staging environment only so far) without any issues.

davidhiendl:
I found some more issues when testing, mostly related to a small "side-effect" change in behavior with getAttribute(...) no longer returning an "empty" model when the attribute does not exist. I did a brief search of the codebase for any instances of getAttribute(...) calls without checks if the result is actually valid and found only the one.

davidhiendl:
Those Blackfire results seem slow, even with cache disabled I usually only see ~130ms per request, and ~40ms with cache enabled. I've deployed this into a smaller production shop earlier today and I've already seen a reduction of ~ -20ms and a 20% drop in total MySQL queries.

What kinda of setup are you testing this on? Need to reproduce it to perhaps improve it further / find out the exact reason.

luigifab:
I test on a test website, with all Magento cache enabled, Turpentine cache bypassed for Blackfire, one server for Nginx, one server for Varnish, one server for MariaDB, one server for Redis (all on the same machine, virtualized).

Probably there are some bad things in our theme or installed modules (Amasty, Mirasvit...).
I not yet read all your changes, but perhaps it's possible to cache getStore to avoid 4000 calls.

davidhiendl:
@luigifab I did some testing just now with caching the store_id, the impact is negligible and there is a problem with caching it: When using environment emulation (for example in cron jobs to generate exports or in adminhtml) it will result in the wrong store labels being returned. The only way around this I can see is to modify Mage_Core_Model_App::setCurrentStore to update/clear the cached store id but that would introduce a hard dependency to Mage_Eav_*

I did however reduce some internal calls to it and improved a few checks to improve performance, but in general the impact is negligible. The real performance gain here is the reduction of several DB queries (and sometimes repeated queries) to a single cache key load.

davidhiendl:
@luigifab I found a workaround for caching the store_id by observing the controller dispatch and caching it only for frontend controllers and made some small improvements to lookup logic for entity types. The gain was very small but it reduced the number of getStore() calls significantly.

davidhiendl:
@luigifab After some additional profiling and the changes in the last commits, this is the total impact:

image

app/code/core/Mage/Eav/Model/Observer.php
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Class Mage_Eav_Model_Observer
+ *
+ * @category   Mage
+ * @package    Mage_Eav
+ * @author     David Hiendl
+ */
+class Mage_Eav_Model_Observer
+{
+
```
fballiano, line 31:
```suggestion
```

app/code/core/Mage/Eav/Model/Observer.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /** @var Mage_Core_Controller_Varien_Action $controllerAction */
+        $controllerAction = $event->getData('controller_action');
+
+        // initialize cached store_id for frontend controllers only to avoid issues with cron jobs and admin controllers which sometimes change store view
+        if ($controllerAction instanceof Mage_Core_Controller_Front_Action) {
+            Mage::getSingleton('eav/config')->setCurrentStoreId(Mage::app()->getStore()->getId());
+        }
+    }
+
```
fballiano, line 47:
```suggestion
```

fballiano:
phpstan is complaining about some things, could you check them out?

davidhiendl:
@fballiano Fixed most of it, but some of the ones marked as ignored by phpstan are actually valid to my point of view and Phpstorms inference.

davidhiendl:
@fballiano I cleaned up the phpstan ignored patterns that are relevant to this PR as those no longer apply.

app/code/core/Mage/Api2/Model/Resource.php
```diff
@@ -1030,7 +1030,7 @@ public function getEavAttributes($onlyVisible = false, $excludeSystem = false)
         $model = $this->getConfig()->getResourceWorkingModel($this->getResourceType());
 
         /** @var Mage_Eav_Model_Entity_Type $entityType */
-        $entityType = Mage::getModel('eav/entity_type')->load($model, 'entity_model');
+        $entityType = Mage::getSingleton('eav/config')->getEntityType($model, 'entity_model');
```
kiatng, line 1033: `$model` is one of 
image
When I try
```php
$entityType = Mage::getSingleton('eav/config')->getEntityType('customer/customer', 'entity_model');
```
I get
> Invalid entity_type specified: customer/customer

I suggest reverting this change. The original load takes about 0.7ms anyway in my development server.

davidhiendl: I've just tested this and it works perfectly fine. Reverting is not an option, the whole idea is to reduce the repeated and unnecessary loading of EAV-type entities. I need more information to reproduce this.
Are you using the latest version of the PR (there was a change half-way that might affect this)?
Which context did you test this in? Is it a unmodified Openmage install?

kiatng: I am testing this function specifically to see if it can load by `entity_model`: https://github.com/OpenMage/magento-lts/blob/3d3e9163461d9fa2736c7340a68f35efbd9d10a6/app/code/core/Mage/Eav/Model/Config.php#L295-L295

The simple test says it cannot.  I test it by simply running this single statement:
```php
$entityType = Mage::getSingleton('eav/config')->getEntityType('customer/customer', 'entity_model');
```
I replaced `$model` with string `'customer/customer'`.

Note that the original load
```php
$entityType = Mage::getModel('eav/entity_type')->load($model, 'entity_model');
```
is load by `entity_model`. I think you need to change `$model` to `entity_type_code` in order to use `Mage::getSingleton('eav/config')->getEntityType()`.

davidhiendl: Kiatng: The code reference you linked is old, I extended the `Mage_Eav_Model_Config::getEntityType($code)` with a second parameter `$field` that mimics the abstract models `load(...)` second parameter.

kiatng: Can you provide the extended code for review? I cannot find it here.

davidhiendl: It's in the PR, that code snipped was linked directly from the PR changes.

kiatng: Thanks, I found it here: https://github.com/OpenMage/magento-lts/blob/fa31e6067d41ec80a725650d8f2fcad2b732eacf/app/code/core/Mage/Eav/Model/Config.php

app/code/core/Mage/Catalog/Model/Resource/Eav/Attribute.php
```diff
@@ -288,6 +287,15 @@ public function getFrontendLabel()
         return $this->_getData('frontend_label');
     }
 
+    /**
+     * Retrieve is_filterable value
+     * @return int
+     */
+    public function getIsFilterable()
```
colinmollenhour, line 294: What is the point of making this a real method, it appears to be functionally equivalent?

app/code/core/Mage/Catalog/Model/Resource/Product/Flat/Indexer.php
```diff
@@ -274,6 +272,10 @@ public function getAttribute($attributeCode)
                 ->getEntity();
             $attribute->setEntity($entity);
 
+            if (!($attribute instanceof Mage_Eav_Model_Entity_Attribute)) {
+                Mage::throwException('Product attribute(code=". $attributeCode . ") is expected to be of type Mage_Eav_Model_Entity_Attribute');
```
colinmollenhour, line 276: Should just be this since it is not translated or meant for end users to see, also not properly quoted for variable expansion:
```suggestion
                throw new Exception('Product attribute(code=' . $attributeCode . ') is expected to be of type Mage_Eav_Model_Entity_Attribute');
```

app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
```diff
@@ -67,11 +67,15 @@ public function getProduct()
...
-        if (is_null($this->_purchasedSeparatelyAttribute)) {
+        if ($this->_purchasedSeparatelyAttribute === null) {
             $_attributeCode = 'links_purchased_separately';
 
-            $this->_purchasedSeparatelyAttribute = Mage::getModel('eav/entity_attribute')
-                ->loadByCode(Mage_Catalog_Model_Product::ENTITY, $_attributeCode);
+            $attribute = Mage::getSingleton('eav/config')
+                ->getAttribute(Mage_Catalog_Model_Product::ENTITY, $_attributeCode);
+            if (!($attribute instanceof Mage_Catalog_Model_Resource_Eav_Attribute)) {
+                Mage::throwException("Attribute links_purchased_separately must be of type Mage_Catalog_Model_Resource_Eav_Attribute");
```
colinmollenhour, line 76:
```suggestion
                throw new Exception('Attribute links_purchased_separately must be of type Mage_Catalog_Model_Resource_Eav_Attribute');
```

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -28,191 +28,283 @@
...
+        $this->_storeInitialized = [];
+        $this->_entityTypes = null;
+        $this->_entityTypeByCode = null;
+        $this->_entityTypeAttributes = null;
+        $this->_entityTypeAttributeIdByCode = null;
+        $this->_attributeSetInfo = null;
+        $this->_defaultAttributes = [];
+
+        Mage::app()->cleanCache([self::ENTITIES_CACHE_ID]);
+        // Mage::app()->removeCache(self::ENTITIES_CACHE_ID . '_' . Mage::app()->getStore()->getId());
```
colinmollenhour, line 114:
```suggestion
```

app/code/core/Mage/Eav/Model/Observer.php
```diff
@@ -0,0 +1,46 @@
...
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/osl-3.0.php
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@magento.com so we can send you a copy immediately.
+ *
+ * @category   Mage
+ * @package    Mage_Eav
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
```
colinmollenhour, line 17: I don't see why to add this on new files..
```suggestion
```

app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
```diff
@@ -133,18 +133,23 @@ protected function _construct()
...
-            $entityType = Mage::getModel('eav/entity_type')->loadByCode($entityType);
-        }
-        if ($entityType instanceof Mage_Eav_Model_Entity_Type) {
-            $entityTypeId = $entityType->getId();
-        }
-        if (empty($entityTypeId)) {
-            throw Mage::exception('Mage_Eav', Mage::helper('eav')->__('Invalid entity supplied.'));
+        $model = Mage::getSingleton('eav/config')->getAttribute($entityType, $code);
+        if ($model) {
+            $this->setData($model->getData());
```
colinmollenhour, line 138: I think a "load" method should still do a clean load rather than copying the cached value to a fresh model.
```suggestion
            $this->load($model->getId());
```

colinmollenhour (commented):
Looks great overall! I just skimmed it, didn't do any testing.

app/code/core/Mage/Catalog/Model/Resource/Eav/Attribute.php
davidhiendl: In a word: performance. According to xdebug profiling it is often called several thousands of times on catalog pages when you have a decent amount of filter options. __call is fairly expensive in comparison and it saved ~1.5-2ms on our production store.

app/code/core/Mage/Catalog/Model/Resource/Product/Flat/Indexer.php
davidhiendl: Will change, oversight. There is a second location where the same logic applies.

app/code/core/Mage/Eav/Model/Config.php
davidhiendl: That call was there in a commented state before, didn't want to remove it without cause but will remove it.

app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
davidhiendl: I thought same but the problem is the load method is called in same places and the only way an attribute changes is via API/Admin and cache is cleared after that. I'd rather keep it like this, remember the admin area doesn't use cache (only local per-process).

davidhiendl: Also: even if attribute changes during a request, the _afterSave() methods will call eav/config->clear which resets the cache.

app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
davidhiendl: Fixed.

app/code/core/Mage/Catalog/Model/Resource/Eav/Attribute.php
colinmollenhour: Not sure where we draw the line on that but maybe just add a note to the vardoc then like "Optimization to avoid __call".

An interesting idea would be a script that replaces all @method annotations with functions at build time.. A thousand micro-optimizations makes at least a milli-optimization, right?

app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
colinmollenhour: I think this would be broken, right?
```php
$attribute = ...->loadByCode('x');
$attribute->setData('foo', 'bar')->save();
$attribute->loadByCode('x');
assertEquals($attribute->getData('foo'), 'bar');
```

app/code/core/Mage/Catalog/Model/Resource/Eav/Attribute.php
davidhiendl: In general, yes the performance gain is marginal for most cases but this one accounted for more than half the __call invocations on pages with product lists according to xdebug profiler.

Generating all phpdoc methods actually is an interesting idea, but won't be enough. There is a ton of fields around the code that are not documented. Maybe I'll look into it.

app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
davidhiendl: It shouldn't be, loadByCode calls resource->loadByCode which does a setData($row) which in turn does a $this->_data=$row so I think that assertion should always be false since "re-loading" the model should clear all state.

davidhiendl: No, save() calls _afterSave() which calls eav/config->clear() 
I did not add that _afterSave call btw, it was present already.

Example stack when saving an attribute:
image

davidhiendl: It's also important that the cache gets cleared *entirely* otherwise the changes wont be reflected and that might cause issues, which is why I added a `Mage::app()->cleanCache([self::ENTITIES_CACHE_ID]);` to the clear method to flush the entire tag, clearing the cache for all store variants.

colinmollenhour: Ahh ok, thanks. Nice work!

davidhiendl: Happy to explain it, this PR has substantial internal changes. It has some potential to cause problems which is also why I think it should only go to 20.x and not to 19.x.

luigifab (commented):
I tested it again this afternoon, it works for our store.
I will leave it applied on our tests env.

You have changed many things, and by reading I don't understand all. But yes I agree, it reduce number of SQL requests and total time (here Nginx and Redis are on the same server):

category:
image


cart:
image

davidhiendl:
@luigifab Reducing total request time & queries was the point. I've deployed it to one production store for several days now without any issues.

luigifab:
I forgot to test the backend, and I have a problem with Catalog > Categories, the tree of categories does not appear, and I have the following error:

image
image

Same for Categories tab of a product in Catalog > Product.
I'm searching why, perhaps is us, or another PR, or me.

luigifab:
This is:

https://github.com/OpenMage/magento-lts/blob/0baa7fab80eb09ae38db5ac998a219fdacf844e2/app/code/core/Mage/Catalog/Model/Resource/Category.php#L483-L490

For me:

```php
var_dump($id = Mage::getSingleton('eav/config')
               ->getAttribute(Mage_Catalog_Model_Category::ENTITY, 'is_active')
              ->getId());
var_dump(is_int($id));
```
The result is (PHP 8.0.27):
```
string(2) "42"
bool(false)
```

davidhiendl:
@luigifab No, its not you. I can reproduce this but strangely enough only on the openmage installation. The reason is that the attribute_id is string sometimes. I'll look into that.

davidhiendl:
@luigifab It turns out that Collection->getData() returns all column values as string. I guess I should have known better in Magento and never have added that explicit is_int(...) type check. The last commit however ensures that some of the critical attribute data has the correct data-type no matter what.

app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php
```diff
@@ -133,18 +133,23 @@ protected function _construct()
...
+            } elseif (is_string($entityType)) {
+                $entityType = Mage::getSingleton('eav/config')->getEntityType($entityType);
+            }
+            if ($entityType instanceof Mage_Eav_Model_Entity_Type) {
+                $entityTypeId = $entityType->getId();
+            }
+            if (empty($entityTypeId)) {
+                throw Mage::exception('Mage_Eav', Mage::helper('eav')->__('Invalid entity supplied.'));
+            }
+            $this->_getResource()->loadByCode($this, $entityTypeId, $code);
```
kiatng, line 151: To add
```php
 * @method Mage_Eav_Model_Resource_Entity_Attribute _getResource()
```

luigifab:
Sometimes we have: `Zend_Db_Select_Exception: You cannot define a correlation name 'al' more than once`
Here is the stack trace from a product edit page in backend:
```
#0 web/lib/Varien/Db/Select.php(277): Zend_Db_Select->_join()
#1 web/app/code/core/Zend/Db/Select.php(367): Varien_Db_Select->_join()
#2 web/app/code/core/Mage/Eav/Model/Resource/Entity/Attribute/Collection.php(427): Zend_Db_Select->joinLeft()
#3 web/app/code/core/Mage/Eav/Model/Config.php(217): Mage_Eav_Model_Resource_Entity_Attribute_Collection->addStoreLabel()
#4 web/app/code/core/Mage/Eav/Model/Config.php(172): Mage_Eav_Model_Config->_loadEntityAttributes()
#5 web/app/code/core/Mage/Eav/Model/Config.php(536): Mage_Eav_Model_Config->_initializeStore()
#6 web/app/code/core/Mage/Eav/Model/Entity/Abstract.php(503): Mage_Eav_Model_Config->getEntityAttributeCodes()
#7 web/app/code/core/Mage/Catalog/Model/Product/Type/Abstract.php(208): Mage_Eav_Model_Entity_Abstract->loadAllAttributes()
#8 web/app/code/core/Mage/Catalog/Model/Product/Type/Abstract.php(244): Mage_Catalog_Model_Product_Type_Abstract->getSetAttributes()
#9 web/app/code/core/Mage/Catalog/Model/Product.php(675): Mage_Catalog_Model_Product_Type_Abstract->getEditableAttributes()
#10 web/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tabs.php(56): Mage_Catalog_Model_Product->getAttributes()
#11 web/app/code/core/Mage/Core/Block/Abstract.php(306): Mage_Adminhtml_Block_Catalog_Product_Edit_Tabs->_prepareLayout()
#12 web/app/code/core/Mage/Core/Model/Layout.php(495): Mage_Core_Block_Abstract->setLayout()
#13 web/app/code/core/Mage/Core/Model/Layout.php(511): Mage_Core_Model_Layout->createBlock()
#14 web/app/code/core/Mage/Core/Model/Layout.php(255): Mage_Core_Model_Layout->addBlock()
#15 web/app/code/core/Mage/Core/Model/Layout.php(221): Mage_Core_Model_Layout->_generateBlock()
#16 web/app/code/core/Mage/Core/Model/Layout.php(226): Mage_Core_Model_Layout->generateBlocks()
#17 web/app/code/core/Mage/Core/Controller/Varien/Action.php(352): Mage_Core_Model_Layout->generateBlocks()
#18 web/app/code/core/Mage/Core/Controller/Varien/Action.php(263): Mage_Core_Controller_Varien_Action->generateLayoutBlocks()
#19 web/app/code/core/Mage/Adminhtml/Controller/Action.php(298): Mage_Core_Controller_Varien_Action->loadLayout()
#20 web/app/code/core/Mage/Adminhtml/controllers/Catalog/ProductController.php(272): Mage_Adminhtml_Controller_Action->loadLayout()
#21 web/app/code/core/Mage/Core/Controller/Varien/Action.php(428): Mage_Adminhtml_Catalog_ProductController->editAction()
#22 web/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(262): Mage_Core_Controller_Varien_Action->dispatch()
#23 web/app/code/core/Mage/Core/Controller/Varien/Front.php(188): Mage_Core_Controller_Varien_Router_Standard->match()
#24 web/app/code/core/Mage/Core/Model/App.php(371): Mage_Core_Controller_Varien_Front->dispatch()
#25 web/app/Mage.php(751): Mage_Core_Model_App->run()
#26 public/index.php(93): Mage::run()
```
But, I don't know why, and for now I haven't more information.
Sadly, the SQL request was not logged in the report...

davidhiendl:
@luigifab It appears that Mage_Eav_Model_Entity_Type::getAttributeCollection returns a cached collection, which means if for whatever reason the attributes are generated twice for the same store/set the addStoreLabel method runs into a conflict because it does not check.
I can as of yet not reproduce this without calling it explicitly twice, but I will modify the `addStoreLabel` method to check if it was already called.

davidhiendl:
I've been testing this in production on 3 stores for some time now. No more new issues. Would love to see this merged so I can start working on some other improvements from my modules I'd like to bring to OpenMage.

kiatng:
How to test for the performance improvement? Loading of an eav model? Loading of an eav collection? Loading of a page? Which specific areas that can be tested to show the time differences?

davidhiendl:
Whenever at least one EAV Entity type and/or a multiple attributes are loaded. For me it was most noticeable on product and category pages with product lists (store that started this has >150 attributes across various product types). With cache enabled it reduces all related queries to a single cache key lookup, but even with cache disabled it does reduce the total amount of queries because the data is fetched at once instead of in multiple bits.

Easiest way to profile is probably looking at the raw pageload speed (keep in mind the improvement is much more noticeable the longer your cache/db latencies are), or if you want it percise: xdebug in profiling mode and piping that output to a frontend (like KCacheGrind) and looking for the total impact of Mage_Eav_Model_Config, but also the impact of direct model loads from resource models that was eliminated.

colinmollenhour (commented):
I noticed it is loading/saving all attribute types at once. Maybe it is negligible but wouldn't it be better to load/save each attribute type independently? That is, better to not load customer address attributes when viewing a product page where no customer data is rendered. At least that's how I read it.

fballiano:
we should also add a line in the README, in the "Between OpenMage 19.x and 20.x" section

davidhiendl:
> I noticed it is loading/saving all attribute types at once. Maybe it is negligible but wouldn't it be better to load/save each attribute type independently? That is, better to not load customer address attributes when viewing a product page where no customer data is rendered. At least that's how I read it.

Yes, it loads all at once. But only deserializes them as needed, if you load each you need multiple cache requests which defeats the purpose (at least partially).

davidhiendl:
> we should also add a line in the README, in the "Between OpenMage 19.x and 20.x" section

Will do.

davidhiendl:
The impact of only loading the cached data without hydrating it (eg. keep it in array from after deserialize) is next to nothing. If you need a 2nd cache request (you usually do for category data when you have a product or customer data if logged in or cart is present) you loose far more time then just loading it at once.

colinmollenhour (approved):
I've not performed any testing but have read through the code and agree with the concept and don't see any problems.

fballiano (approved):
since we're releasing a "rc1" version, if we don't do this now...

luigifab:
I forgot some comments ago: before this PR, you can `->addAttributeToSelect('test')` when `test` product attribute does not exist, now you can't:
```
Invalid attribute requested: test

#0 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(443): Mage::exception()
#1 app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php(504): Mage_Eav_Model_Entity_Collection_Abstract->addAttributeToSelect()
#2 app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php(421): Mage_Catalog_Model_Resource_Product_Collection->addAttributeToSelect()
#3 app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php(504): Mage_Eav_Model_Entity_Collection_Abstract->addAttributeToSelect()
#4 app/code/community/Luigifab/Urlnosql/Controller/Router.php(201): Mage_Catalog_Model_Resource_Product_Collection->addAttributeToSelect()
#5 app/code/community/Luigifab/Urlnosql/Controller/Router.php(82): Luigifab_Urlnosql_Controller_Router->getProduct()
#6 app/code/core/Mage/Core/Controller/Varien/Front.php(188): Luigifab_Urlnosql_Controller_Router->match()
#7 app/code/core/Mage/Core/Model/App.php(371): Mage_Core_Controller_Varien_Front->dispatch()
#8 app/Mage.php(750): Mage_Core_Model_App->run()
#9 index.php(41): Mage::run()
#10 {main}
  thrown in app/Mage.php on line 648
  catched by Mage::printException()
```

luigifab:
I found a new bug (perhaps it's only me, perhaps this is other PR that I also use... I don't searched yet).

- Go to customer edit page in backend (I have multiple store views, one website, and `customer/account_share/scope` is _per website_).
- Clear all cache.
- Refresh the page.

I have (with OpenMage 20.1.0-rc1 and PHP 8.2):
```
Cannot call addStoreLabel for different store views on the same collection

#0 app/code/core/Mage/Eav/Model/Config.php(213): Mage_Eav_Model_Resource_Entity_Attribute_Collection->addStoreLabel()
#1 app/code/core/Mage/Eav/Model/Config.php(172): Mage_Eav_Model_Config->_loadEntityAttributes()
#2 app/code/core/Mage/Eav/Model/Config.php(532): Mage_Eav_Model_Config->_initializeStore()
#3 app/code/core/Mage/Eav/Model/Entity/Abstract.php(503): Mage_Eav_Model_Config->getEntityAttributeCodes()
#4 app/code/core/Mage/Eav/Model/Entity/Abstract.php(938): Mage_Eav_Model_Entity_Abstract->loadAllAttributes()
#5 app/code/core/Mage/Core/Model/Abstract.php(292): Mage_Eav_Model_Entity_Abstract->load()
#6 app/code/core/Mage/Adminhtml/controllers/CustomerController.php(61): Mage_Core_Model_Abstract->load()
#7 app/code/core/Mage/Adminhtml/controllers/CustomerController.php(112): Mage_Adminhtml_CustomerController->_initCustomer()
#8 app/code/core/Mage/Core/Controller/Varien/Action.php(428): Mage_Adminhtml_CustomerController->editAction()
#9 app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(262): Mage_Core_Controller_Varien_Action->dispatch()
#10 app/code/core/Mage/Core/Controller/Varien/Front.php(188): Mage_Core_Controller_Varien_Router_Standard->match()
#11 app/code/core/Mage/Core/Model/App.php(371): Mage_Core_Controller_Varien_Front->dispatch()
#12 app/Mage.php(750): Mage_Core_Model_App->run()
#13 index.php(41): Mage::run()
#14 {main}
  thrown in app/code/core/Mage/Eav/Model/Resource/Entity/Attribute/Collection.php on line 441
  catched by Mage::printException()
```

davidhiendl:
I will create a follow-up PR for the non-existent attribute, that's definitely a bug and will cause problems. I will take a look at the other problem after that, but I can't reproduce it just yet.

darinda:
i can confirm the Exception(`Cannot call addStoreLabel for different store views on the same collection`) appearing in the backend when viewing a placed order (`[..]sales_order/view/order_id/%orderID%)` and when the caches are purged.

It happens with OpenMage 20.1.0-rc1 and PHP 8.0 as well as PHP 8.1. Didn't test 8.2 yet.

davidhiendl:
Okay, I figured out the reason. It's because `Mage_Eav_Model_Entity_Type::getAttributeCollection` re-uses the same collection. I'm testing a rewrite of this method right now and will submit another followup PR shortly.

luigifab:
With #2044 and #3057, I approve.
For us it works so good with PHP 8.0 and 8.2.

ADDISON74:
This PR requires revision because it introduced two issues that we reported today. IT is possible that find more.

PR #3000, opened 2023-01-27
Clear hasDataChanges flag after loading collection items

elidrissidev:
### Description (*)

PR created as a replacement for https://github.com/OpenMage/magento-lts/pull/2042#issuecomment-1406779856

When any collection is loaded, all its items have the `hasDataChanges` flag set to `true` because of the call to `setData()`. This causes unnecessary saves for unchanged items.

I also added a getter method to the status history model to prevent the setter from marking the model as dirty again, similar to what was done in #2066.

### Related Pull Requests

- See #2042
- See #523

1. Load any collection and call `hasDataChanges()` on the items, it should be `false`.
2. Load status history collection from Order and call `hasDataChanges()` on the items, it should return `false`.
```php
$order = Mage::getModel('sales/order')->load(123);
foreach ($order->getStatusHistoryCollection() as $history) {
    var_dump($history->hasDataChanges()); // bool(false)
}
```

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Status/History.php
+++ b/app/code/core/Mage/Sales/Model/Order/Status/History.php
@@ -38,7 +38,6 @@
  * @method $this setParentId(int $value)
  * @method string getStatus()
  * @method $this setStatus(string $value)
- * @method $this setStoreId(int $value)
  * @method int getIsVisibleOnFront()
  * @method $this setIsVisibleOnFront(int $value)
  */
@@ -69,7 +68,7 @@ class Mage_Sales_Model_Order_Status_History extends Mage_Sales_Model_Abstract
     }
 
     /**
-     * Set order object and grab some metadata from it
+     * Set order object
      *
      * @param   Mage_Sales_Model_Order $order
      * @return  $this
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Sales/Model/Order/Status/History.php
```diff
@@ -69,18 +68,27 @@ protected function _construct()
...
     }
 
+    /**
+     * Get store id
+     *
+     * @throws Mage_Core_Model_Store_Exception
+     */
+    public function getStoreId(): int
+    {
+        return $this->getStore()->getStoreId();
```
fballiano, line 89: is it $this->getStore()->getStoreId() or $this->getOrder()->getStoreId()

elidrissidev: `getStore()`, since it checks if the Order is set first, otherwise defaults to current store.

fballiano (approved):

app/code/core/Mage/Sales/Model/Order/Status/History.php
fballiano: true! stupid question :-)

kiatng (approved):
Tested.

fballiano:
cherrypicked to v20

PR #2957, opened 2023-01-17
Reduce getId calls

luigifab:
### Description

This reduce number of `getId()` calls (found with Blackfire).
Same as #2677 and #2699.

getid-before

Yes this will not change many things, but this + many others...

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3007, opened 2023-02-02
Reduce getId calls

luigifab:
### Description

This reduce number of getItemId() calls (found with Blackfire).
Same as #2957 #2677 #2699 #1571.

before (with a big cart with 51 downloadable, simple and bundle products):
before-getitemid

after:
after-getitemid

To check the PR, I added `if ($quoteItemId != $quoteItem->getItemId()) exit('boom');`.
Yes this will not change many things, but this + many others...ac

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/CatalogInventory/Model/Observer.php
+++ b/app/code/core/Mage/CatalogInventory/Model/Observer.php
@@ -273,8 +273,8 @@ class Mage_CatalogInventory_Model_Observer
         $canRemoveErrorFromQuote = true;
 
         /** @var Mage_Sales_Model_Quote_Item $quoteItem */
-        foreach ($quoteItems as $quoteItem) {
-            if ($quoteItem->getItemId() == $item->getItemId()) {
+        foreach ($quoteItems as $quoteItemId => $quoteItem) {
+            if ($quoteItemId == $item->getItemId()) {
                 continue;
             }
 


app/code/core/Mage/CatalogInventory/Model/Observer.php
```diff
@@ -273,8 +273,8 @@ protected function _removeErrorsFromQuoteAndItem($item, $code)
         $canRemoveErrorFromQuote = true;
 
         /** @var Mage_Sales_Model_Quote_Item $quoteItem */
-        foreach ($quoteItems as $quoteItem) {
-            if ($quoteItem->getItemId() == $item->getItemId()) {
+        foreach ($quoteItems as $quoteItemId => $quoteItem) {
```
Flyingmana, line 276: where is the ID coming from in the Collection? As far as I see its build in most of the cases without the itemId as key.

Flyingmana (changes requested):
see question in code

app/code/core/Mage/CatalogInventory/Model/Observer.php
Flyingmana: would an alternate be to actually code getItemId() instead of relying on the __call() method?

luigifab: It's seems that majority of collection have the "id" as key?

From [lib/Varien/Data/Collection.php](https://github.com/OpenMage/magento-lts/blob/1.9.4.x/lib/Varien/Data/Collection.php), there are some references for id used as key. But perhaps I'm wrong.

From the method `load` of a collection, there is something about the id:
https://github.com/OpenMage/magento-lts/blob/fc98af9ed61e6fc225ac0e864a5fdf564aac1a36/lib/Varien/Data/Collection/Db.php#L583-L592

https://github.com/OpenMage/magento-lts/blob/fc98af9ed61e6fc225ac0e864a5fdf564aac1a36/lib/Varien/Data/Collection.php#L359-L365

elidrissidev: The item id is indeed the key for all collections.

Flyingmana (approved):

elidrissidev (approved):

fballiano:

PR #3008, opened 2023-02-02
Fix phpseclib class namespace in Varien_Io_Sftp

elidrissidev:
### Description (*)

A small fix related to `phpseclib/phpseclib` composer dependency which uses a different namespace than the one that was installed in `lib/`. I also removed the unneeded `require_once` from the top of the file, and corrected an undefined variable usage.

### Related Pull Requests

- See #2411

### DIFF
--- a/lib/Varien/Io/Sftp.php
+++ b/lib/Varien/Io/Sftp.php
@@ -19,8 +19,6 @@
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
-require_once('phpseclib/Net/SFTP.php');
-
 /**
  * Sftp client interface
  *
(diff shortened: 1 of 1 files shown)


fballiano (approved):

lib/Varien/Io/Sftp.php
```diff
@@ -117,7 +115,7 @@ public function rmdir($dir, $recursive = false)
             $list = $this->_connection->nlist();
             if (!count($list)) {
                 // Go back
-                $this->_connection->chdir($pwd);
+                $this->_connection->chdir($cwd);
```
kiatng, line 118: Is it better to change line 111 instead:

from `$cwd = $this->_connection->pwd();`
to `$pwd = $this->_connection->pwd();`

elidrissidev: I think `pwd` means "print working directory", and `cwd` means "current working directory". So `cwd` is the value resulting from calling `pwd()`.

Flyingmana (approved):

fballiano:
merged and ported to 20.0

PR #3011, opened 2023-02-03
set phpstan to bleedingEdge config

Flyingmana:
there is a performance improvement for phpstan possible when switching to the currend bleedingEdge config.

https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types#lower-memory-consumption

Heard about it while researching for a different project ( https://github.com/phpstan/phpstan/discussions/8813 )

### Questions or comments

There are some incompatibilities mentioned, and while I checked I also saw that we get some errors.
Not sure yet about what needs to change, but for the future we will need to fix it anyway anywhen


elidrissidev:
Nice, I noticed an improvement from 1:35 to only 40s when using the new configuration!

elidrissidev:
Fixed fix for some phpstan issues, but some tricky ones popped up.

index.php
```diff
@@ -28,6 +28,7 @@
 require MAGENTO_ROOT . '/app/bootstrap.php';
 require_once $mageFilename;
 
+
```
fballiano, line 31:
```suggestion
```

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Abstract.php
```diff
@@ -40,8 +40,9 @@ public function getCategory()
 
     public function getCategoryId()
     {
-        if ($this->getCategory()) {
-            return $this->getCategory()->getId();
+        $category = $this->getCategory();
+        if (is_object($category) && $category->getId()) {
```
fballiano, line 44:
```suggestion
        if (is_object($category)) {
```

@elidrissidev what about this?

fballiano:
fixed conflict with baseline by using 1.9.4.x baseline

fballiano:
guys, this is so incredibly much faster that I suggest we regenerate the baseline based on the code that's now in the PR and merge it. what do you think?

fballiano:
I've regenerated the baseline (from a linux machine, to avoid problems with macos), we could merge it probably

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Abstract.php
elidrissidev: Actually now that I think about it, what was reason of changing the `if ($this->getCategory())` condition? The only PHPStan error in this file was about the `Right side of && is always false` which you already fixed below..

elidrissidev:
Resolved some issues that were newly added to baseline since they were easy to fix.

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Abstract.php
```diff
@@ -53,7 +54,8 @@ public function getCategoryName()
 
     public function getCategoryPath()
     {
-        if ($this->getCategory()) {
+        $category = $this->getCategory();
+        if (is_object($category) && $category->getId()) {
```
fballiano, line 58:
```suggestion
        if ($this->getCategory()) {
```

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Abstract.php
fballiano, line 44:
```suggestion
        if ($this->getCategory()) {
```

fballiano (approved):

app/code/core/Mage/Adminhtml/Block/Catalog/Category/Abstract.php
```diff
@@ -93,8 +93,9 @@ public function getRoot($parentNodeCategory = null, $recursionLevel = 3)
             $tree = Mage::getResourceSingleton('catalog/category_tree')
                 ->load(null, $recursionLevel);
 
-            if ($this->getCategory()) {
-                $tree->loadEnsuredNodes($this->getCategory(), $tree->getNodeById($rootId));
+            $category = $this->getCategory();
+            if (is_object($category) and $category->getId()) {
```
justinbeaty, line 97: Not a big deal, but using `&&` would be more consistent with the codebase.

```php
if (is_object($category) && $category->getId()) {
```

fballiano, line 97:
```suggestion
            if (is_object($category) && $category->getId()) {
```

fballiano: true, done

elidrissidev: what is this change for?

fballiano:
@elidrissidev  I lost your last comment, I guess that change was an older one (since this PR went thru many revisions), I've removed it, let's see if it works

elidrissidev (approved):
Looks good

fballiano (approved):

fballiano:
merged and v20ed

elidrissidev:
btw when you clear the description of the final squashed commit, be sure to keep the any special lines in it such as `Signed-off-by`,`Co-authored-by`, etc. Else that information will not appear on the merged commit

fballiano:
sorry, sometimes there's so much stuff in there that doesn't seem useful at all :-(

elidrissidev:
No problem :)

PR #3014, opened 2023-02-07
Fix bug in admin interface where Bundle Option Is Required setting is not loaded.

rfeese:
### Description (*)

This fixes a bug in the admin area when editing Bundle Items Options on a Bundle type product. When "Is Required" on an Option is set to No, the select element on the screen is not initialized with the selected value from the database because of an oversight in the check for existence of a variable with value zero (in the case of Is Required = "No").

1. In Admin interface, Create or Edit a Bundle type product.
2. On the "Bundle Items" tab, Create / Edit an existing Option.
3. Change the Option "Is Required" setting to "No".
4. Save and Continue Edit or Save and return to Edit the product.
5. Notice that the "Is Required" setting reverts back to "Yes" (the first option).

This happens because the javascript that initializes the "Is Required" select element only runs if the value is "Yes" (1) due to the code checking the value of the variable -- "No" (0) evaluates to false.

Note that the Is Required setting *is* saved correctly based on what *currently* shows in the Admin screen (Is Required will be set to Yes unless you change it before saving), and appears to be handled correctly in the customer view.

### Questions or comments

I think that the intention of the code was to check for the existence of the variable, so I modified the check to look for whether the variable was defined using typeof.

I suspect this type of problem could show up in other places, but have not attempted to find other instances of this yet.

### DIFF
--- a/app/design/adminhtml/default/default/template/bundle/product/edit/bundle/option.phtml
+++ b/app/design/adminhtml/default/default/template/bundle/product/edit/bundle/option.phtml
@@ -114,7 +114,7 @@ Bundle.Option.prototype = {
         }
 
         //set selected is_require
-        if (data.required) {
+        if (typeof data.required !== 'undefined') {
             $A($(this.idLabel + '_'+data.index+'_required').options).each(function(option){
                 if (option.value==data.required) option.selected = true;
             });


ADDISON74:
LGTM

elidrissidev (approved):
Good find. Works as intended.

fballiano (approved):

fballiano:
merged and cherrypicked to 20.0

PR #3020, opened 2023-02-10
Fix bug in admin interface where Bundle Option Select User Defined Qty setting is not loaded.

rfeese:
### Description (*)

After #3014, I found another similar issue on the Bundle Options admin screen. 
This fixes a bug in the admin area when editing Bundle Items Options on a Bundle type product. When bundle option selection price type is "Fixed" or zero (0), "User Defined Qty" is not initialized with the selected value from the database because of an oversight in the check for existence of a variable with value zero (in the case of selection_price_type = "Fixed").

### Related Pull Requests

#3014 

1. In Admin interface, Create or Edit a Bundle type product.
2. On the "Bundle Items" tab, Create / Edit an existing Option.
3. Add a Selection if one does not already exist.
4. Change the "User Defined Qty" on a Selection to "No".
5. Save and Continue or Save and return to Edit the product.
6. Notice that "User Defined Qty" is not loaded with the value "No".

This happens because the javascript does not update the select element due to the "selection_price_type" evaluating to zero. Note that selection_price_type field may be hidden as it appears to be related to the overall product pricing settings.

### Questions or comments

In doing more research, I found a similar problem fixed in a slightly different way: #2441 . I am not sure if you have any preference on the syntax.

### DIFF
--- a/app/design/adminhtml/default/default/template/bundle/product/edit/bundle/option/selection.phtml
+++ b/app/design/adminhtml/default/default/template/bundle/product/edit/bundle/option/selection.phtml
@@ -154,13 +154,13 @@ Bundle.Selection.prototype = {
 
         Element.insert(tbody[0], {'bottom':this.template.evaluate(data)});
 
-        if (data.selection_price_type) {
+        if (typeof data.selection_price_type !== 'undefined') {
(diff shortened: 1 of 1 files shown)


fballiano (approved):

ADDISON74 (approved):

fballiano:

PR #1377, opened 2021-01-08
Add weight for API salesOrderShipmentAddTrack

luigifab:
### Description
This PR add the possibility to specify a weight when adding a tracking number to a shipment.

[API doc1](https://r-martins.github.io/m1docs/guides/m1x/api/soap/sales/salesOrderShipment/sales_order_shipment.addTrack.html) [API doc2](https://devdocs-openmage.org/guides/m1x/api/soap/sales/salesOrderShipment/sales_order_shipment.addTrack.html)
```
Type | Name | Description
string | sessionId | Session ID
string | shipmentIncrementId | Shipment increment ID
string | carrier | Carrier code (ups, usps, dhl, fedex, or dhlint)
string | title | Tracking title
string | trackNumber | Tracking number
double | weight | Weight of the shipment (optional)
```

OpenMage 20.0.10 / PHP 7.4.11 and 8.0.5

### Manual testing scenarios
Update and run:
```php
$proxy = new SoapClient('https://.../api/soap/?wsdl');
$sessionId = $proxy->login('login', 'pass');
var_dump($proxy->call($sessionId, 'sales_order_shipment.addTrack',
    ['shipment_id', 'custom', 'test', 'trackingNumber', 5]));
```

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #3022, opened 2023-02-15
Remove minOccurs attr from WSDL message part

elidrissidev:
### Description (*)

As mentioned in https://github.com/OpenMage/magento-lts/pull/1377#issuecomment-1431312619, the WSDL message `<part>` may not contain a `minOccurs` and it throws an error when calling the endpoint.

### Related Pull Requests

- See https://github.com/OpenMage/magento-lts/pull/1377

### DIFF
--- a/app/code/core/Mage/Sales/etc/wsdl.xml
+++ b/app/code/core/Mage/Sales/etc/wsdl.xml
@@ -844,7 +844,7 @@
         <part name="carrier" type="xsd:string" />
         <part name="title" type="xsd:string" />
         <part name="trackNumber" type="xsd:string" />
-        <part name="weight" type="xsd:double" minOccurs="0" />
+        <part name="weight" type="xsd:double" />
     </message>
     <message name="salesOrderShipmentAddTrackResponse">
         <part name="result" type="xsd:int" />


leissbua:
Thank you. Hopefully not offending anyone, but the quality of the releases must change and has gotten noticeably worse. Any developer changing the WSDL has to flush WSDL cache, to not repeat killing WSDL again and again.

elidrissidev:
No offense taken. I do agree that a lot of bugs slipped through last year, especially after the project got more maintainers and code merging and release velocity increased.

I will do the best from my side to ensure this does not happen again in the future, and I'm sure others will too.

Thanks for reporting!

kiatng (approved):

fballiano:
are we breaking it yet again now? is this extremely necessary? https://github.com/OpenMage/magento-lts/pull/1377 was merged in august of last year, it doesn't make sense to me to break the WSDL again to fix something that's in our code since a year.

also, the quality of the releases is as high as we can, as I said many times here, some of us are dedicating MANY hours to this project completely for free, we have ZERO sponsorship and very less help. whoever wants more should participate more. period.

leissbua:
Thanks for your feedback, i appreciate it. It is a wonder, that we did not hear of the problem. you will not create trouble by fixing a broken WSDL. It needs to be fixed, and no trouble will be created by fixing it. If it is not fixed you will fix it later when people complain. Compalints will start when people flush their WSDL cache, thats why i recommend to flush WSDL whenever working on the WSDL.

"whoever wants more should participate more. period." I understand what you are trying to say, but quality should not be related to money, because quick hacking never succeeds. But i fully agree, there should be sponsorship as MageOne makes profit and could not exist without OpenMage. Create a donation pool and we surely will donate, and feedback to good or bad things is always given.

fballiano:
quality is related to what we can do, we all give what we can, doing the best that we can. so I'm not accepting that "the quality is low" when our platform is probably the easiest to upgrade. many platforms have more problems than ours (while upgrading) and upgrading with zero effort is some kind of utopia.

anyway I'm checking the WDSL and API systems for my customers (that use this version) and the systems are working fine so I'm not sure I want to break it again. is it creating a practical problem or it's just about the wdsl validation?

leissbua:
It is creating a practical problem, every WSDL validator denies the WSDL, as its broken. Use SOAP-UI if you want to test. Setup MS AX or SAP if you want to see what problems they have.

fballiano:
Sincerely I don't understand everything from your message.

anyway, as you can see, we're already checking your report. my customers have ERPs using SOAP_V2 and they're obviously working, I'm test the WSDL with boomerang and I'm seeing no problem. I don't have SOAP-UI, if you have it you can send a screenshot or something of the error.

since you complain about quality, I'm just making sure that we don't break it again for the people that upgraded and flushed their cache already.

my boomerang client working fine with latest openmage v20:
Screenshot 2023-02-15 alle 17 46 54

elidrissidev:
> Sincerely I don't understand everything from your message.
> 
> anyway, as you can see, we're already checking your report. my customers have ERPs using SOAP_V2 and they're obviously working, I'm test the WSDL with boomerang and I'm seeing no problem. I don't have SOAP-UI, if you have it you can send a screenshot or something of the error.
> 
> since you complain about quality, I'm just making sure that we don't break it again for the people that upgraded and flushed their cache already.
> 
> my boomerang client working fine with latest openmage v20:

There's a typo with `wsdl` in the URL. Could you try executing the request in which the weight was added?

As far as I can see, removing the attribute won't create any issue since it doesn't belong there in the first place. Whether an error is thrown or not might be client-dependent, so it should be removed.

fballiano:
the 2 systems I've tested already have the "weight" parameter and I don't see any error.

what do you mean "there's a typo in the WDSL url"?

elidrissidev:
> what do you mean "there's a typo in the WDSL url"?

In the screenshot you shared, there's a typo in the URL (`?wdsl`)

fballiano:
probably "boomerang" autoadded the correct wsdl=1 parameter by itself. i didn't test the single endpoint but setting up the test call didn't generate any error and my customer that have the soap system in production are working fine.

if the problem is generated when calling the single endpoint that's another story but not what's been reported.

fballiano:
also, I'd like to hear from @luigifab since he's the author of the original PR.

leissbua:
The discussion is fully pointless, as the XML is simply not valid, I have no care why time is wasted with discussions.

image

I did not fix this for fun, it was a client issue with Microsoft Axapta, that could be reproduced with SOAP-UI on our side:

**Error:**
image

**Reason:**
```
minOccurs="0"
```

must not be in a message, as it invalidates the WSDL.

**Explanation:**
In case you did not know, WSDL is not flushed via Magento-Cache flush, but resides in /tmp - so if you did not delete manually you wont see any error, as i stated a few times.

luigifab (approved):
I'm sorry for that, but since I using this change, I haven't had any problems.
I updated yesterday and flush all wsdl cache (I think, I hope), all still working.

I'm not sure how to validate openmage wsdl.

leissbua:
You can validate it at any time, simply via this website:

https://www.wsdl-analyzer.com/

PR #3023, opened 2023-02-16
Support for Google Analytics 4

fballiano:
I've added configuration and template code for a basic Google Analytics 4 support.

Screenshot 2023-02-16 alle 20 05 08

I've also deprecated all methods regarding Universal Analytics and Analytics 3 since they will be phased our by google in July 2023. At that point we should also remove the [IP anonymization since it's the default in GA4](https://support.google.com/analytics/answer/9019185#IP&zippy=%2Cin-this-article), quote:

> In Google Analytics 4, IP masking is not necessary since IP addresses are not logged or stored.


ADDISON74:
Good job. Can someone test this PR in production?

app/code/core/Mage/GoogleAnalytics/Model/System/Config/Source/Type.php
```diff
@@ -37,12 +37,16 @@ public function toOptionArray()
     {
         return [
             [
-                'value' => Mage_GoogleAnalytics_Helper_Data::TYPE_UNIVERSAL,
-                'label' => Mage::helper('googleanalytics')->__('Universal Analytics')
+                'value' => Mage_GoogleAnalytics_Helper_Data::TYPE_ANALYTICS4,
+                'label' => Mage::helper('googleanalytics')->__('Google Analytics 4')
```
elidrissidev, line 41: Translation missing for this string

elidrissidev:
Unfortunately, I don't use GA4 so I can't provide much help here, though it LGTM.

app/code/core/Mage/GoogleAnalytics/Model/System/Config/Source/Type.php
fballiano: added! thanks!

fballiano:
the javascript code is copied from google's pages

elidrissidev (approved):
Untested

ADDISON74 (approved):

fballiano:
merged and v20ed

empiricompany:
Is there any reason why Mage_GoogleAnalytics_Block_Ga::_getPageTrackingCode() wasn't used but the code was inserted directly into ga.phtml? I missed this detail in the review

app/design/frontend/base/default/template/googleanalytics/ga.phtml
```diff
@@ -24,7 +24,17 @@ $_helper = $this->helper('googleanalytics');
 $_accountId = $_helper->getAccountId();
 ?>
 <?php if (!$this->helper('core/cookie')->isUserNotAllowSaveCookie()): ?>
-    <?php if($_helper->isUseUniversalAnalytics()): ?>
+    <?php if ($_helper->isUseAnalytics4()): ?>
+        <!-- BEGIN GOOGLE ANALYTICS 4 CODE -->
+        <script async src="https://www.googletagmanager.com/gtag/js?id=<?= $_accountId ?>"></script>
+        <script>
+            window.dataLayer = window.dataLayer || [];
```
empiricompany, line 31: Is there any reason why Mage_GoogleAnalytics_Block_Ga::_getPageTrackingCode() wasn't used here but the code was inserted directly into ga.phtml? I missed this detail in the review

fballiano: I think I've used the same coding style that was already there from the previous implementations, If I recall correctly

empiricompany: It's a small detail, but I mean to wrap this part inside _getPageTrackingCode() like for universal. 
```
gtag('js', new Date());
gtag('config', '<?= $_accountId ?>');
```
Then there's also getPageName that seems deprecated. I will open a PR to discuss it.

PR #3024, opened 2023-02-17
Add more info to admin auth exceptions

loekvangool:


app/code/core/Mage/Api2/Model/Auth/User/Admin.php
```diff
@@ -63,7 +63,7 @@ public function getRole()
             /** @var Mage_Api2_Model_Acl_Global_Role $role */
             $role = $collection->getFirstItem();
             if (!$role->getId()) {
-                throw new Exception('Admin role not found');
+                throw new Exception('Admin role for user ID '.$this->getUserId().' not found');
```
fballiano, line 66:
```suggestion
                throw new Exception('Admin role for user ID ' . $this->getUserId() . ' not found');
```

app/code/core/Mage/Api2/Model/Auth/User/Admin.php
```diff
@@ -92,7 +92,7 @@ public function getType()
     public function setRole($role)
     {
         if ($this->_role) {
-            throw new Exception('Admin role has been already set');
+            throw new Exception('Admin role has been already set to '.$this->_role.' for user ID '.$this->getUserId());
```
fballiano, line 95:
```suggestion
            throw new Exception('Admin role has been already set to ' . $this->_role . ' for user ID ' . $this->getUserId());
```

app/code/core/Mage/Api2/Model/Auth/User/Admin.php
```diff
@@ -63,7 +63,7 @@ public function getRole()
             /** @var Mage_Api2_Model_Acl_Global_Role $role */
             $role = $collection->getFirstItem();
             if (!$role->getId()) {
-                throw new Exception('Admin role not found');
+                throw new Exception('Admin role for user ID ' . $this->getUserId() . ' not found');
```
fballiano, line 66: I'd add also the role id no?

elidrissidev:
Do you run into these exceptions often that you need to include more debugging to the message? From a quick glance, I'm only able to find one reference of `Mage_Api2_Model_Auth_User_Admin::setRole()` and it's in the same file inside the condition block `if (!$this->_role) {`, so it shouldn't really be thrown.

loekvangool:
I have some kind of misconfiguration ATM from a 3rd party tool that uses our mage API. While tracking it down I got these messages on `getRole()`, not on `setRole()` but I spotted it and though why not add some TLC there as well.

app/code/core/Mage/Api2/Model/Auth/User/Admin.php
loekvangool: There is no role ID if I get the code correctly. This exception fires when there's no role found for the user ID.

fballiano (approved):

ADDISON74 (approved):

fballiano:
merged and v20ed

PR #3029, opened 2023-02-17
Added check for coupon expiration date

fballiano:
Mage_SalesRule_Model_Validator::_canProcessRule() checks the validity of a coupon code that is in the process of being added to the cart, but it doesn't check for the single coupon expiration date. This PR adds this check.

1. in the backend go to "promotions -> Shopping Cart Price Rules"
2. create a new rule with "specific coupon" (but also autogenerated coupons have the same issue) and type a coupon code
3. be sure that the rule also has some "action", ex: 10% discount
4. save the rule and check that it's working on the frontend

Now, what we're about to check isn't really supported by the core but I think it's still worth fixing.

5. open your database console and check for the `salesrule_coupon` table, you'll find a record with the coupon code you just created above
6. modify the `expiration_date` column for that specific coupon and set it in the past
7. go to the frontend, apply the coupon and notice that it gets applied anyway

It is true that the core wouldn't manage the expiration_date column but, since it exists, I know extensions (actually I was developing one today when I found this bug) that set the expiration_data for a single coupon (let's say you want a coupon to be usable only for 7 days after its generation), and it doesn't make sense to me that the core doesn't check for the expiration_date of a coupon.

After applying this PR you'll notice that the expiration_date gets correctly checked, if present.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1373,15 +1373,6 @@
       "contributions": [
         "code"
       ]
-    },
-    {
-      "login": "davidhiendl",
-      "name": "David Hiendl",
-      "avatar_url": "https://avatars.githubusercontent.com/u/11006964?v=4",
-      "profile": "https://github.com/davidhiendl",
-      "contributions": [
-        "code"
-      ]
     }
   ],
   "contributorsPerLine": 7
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ "main", "v20", "v19" ]
+    branches: [ "1.9.4.x", 20.0 ]
   pull_request:
     # The branches below must be a subset of the branches above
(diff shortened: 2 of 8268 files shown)


elidrissidev:
From my testing, updating `salesrule_coupon.expiration_date` only was not enough, but after updating `salesrule.from_date` everything works as expected and I get "Coupon code %s is invalid" on the frontend

fballiano:
> From my testing, updating `salesrule_coupon.expiration_date` only was not enough, but after updating `salesrule.from_date` everything works as expected and I get "Coupon code %s is invalid" on the frontend

the from_date/to_date is checked somewhere else in the code but that wasn't a problem for me with the customer i did this change for.

FredericMartinez (dismissed):

empiricompany (dismissed):

luigifab (dismissed):
Seems good.

fballiano:
@colinmollenhour I should be able to merge this (bypassing checks) but the button doesn't show anymore and I can't find any setting about that in the branch protection rules :-\

empiricompany (dismissed):

PR #3031, opened 2023-02-17
Some PHPStan fixes

fballiano:


fballiano:
I was checking some more but they're super hard to test/fix...

ADDISON74 (dismissed):

app/Mage.php
```diff
@@ -646,7 +646,8 @@ public static function exception($module = 'Mage_Core', $message = '', $code = 0
      */
     public static function throwException($message, $messageStorage = null)
     {
-        if ($messageStorage && ($storage = self::getSingleton($messageStorage))) {
+        if ($messageStorage) {
+            $storage = self::getSingleton($messageStorage);
```
tmotyl, line 650: this is not an equivalent of the previous code.
the phpstan error sounds like a false positive. 

I believe that getSingleton can return false also (the phpdock on get singleton should be adapted.

fballiano: if we also have false positives I've no idea how to handle this stuff

tmotyl: In this case its not phpstan fault as the phpdoc is wrong. 
but anyway Phpstan is just a tool, which also has bugs and sometimes reports false positives.
:)
Can you fix the getSingleton signature, so were good to go with this patch?

fballiano: true, I really missed that, it should be done

lib/Varien/Io/File.php
```diff
@@ -70,9 +70,9 @@ class Varien_Io_File extends Varien_Io_Abstract
     /**
      * Stream open file pointer
      *
-     * @var resource
+     * @var resource|null
      */
-    protected $_streamHandler;
+    protected $_streamHandler = null;
```
tmotyl, line 75: why do we need to declare null here?

lib/Varien/Object/Cache.php
```diff
@@ -31,9 +31,9 @@ class Varien_Object_Cache
     /**
      * Singleton instance
      *
-     * @var Varien_Object_Cache
+     * @var Varien_Object_Cache|null
      */
-    protected static $_instance;
+    protected static $_instance = null;
```
tmotyl, line 36: same here

lib/Varien/Io/File.php
fballiano, line 75:
```suggestion
    protected $_streamHandler;
```

lib/Varien/Object/Cache.php
fballiano, line 36:
```suggestion
    protected static $_instance;
```

app/Mage.php
```diff
@@ -526,7 +526,7 @@ public static function getModel($modelClass = '', $arguments = [])
      *
      * @param   string $modelClass
      * @param   array $arguments
-     * @return  Mage_Core_Model_Abstract
+     * @return  Mage_Core_Model_Abstract|false
```
fballiano, line 529: @tmotyl adding the "|false" triggered quite some other things, should we add them to the baseline or try to fix them?

fballiano:
IDK what happens, on my local machine I get zero errors, here a milion, i've no freaking idea.

elidrissidev:
@fballiano do you get some error about "parallel jobs didn't finish"? Cause it happens to me on macOS too and it won't show phpstan errors

app/Mage.php
```diff
@@ -48,6 +48,14 @@
 
 Varien_Autoload::register();
 
+/** AUTOLOADER PATCH **/
```
elidrissidev, line 51: Is this addition intentional?

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
```diff
@@ -423,6 +424,9 @@ public function initFields($fieldset, $group, $section, $fieldPrefix = '', $labe
                     }
 
                     $sourceModel = Mage::getSingleton($factoryName);
+                    if (!$sourceModel) {
+                        Mage::throwException("Entity type model '{$factoryName}' is not found");
```
elidrissidev, line 428: Could we force phpstan to "trust" the usage by adding a `/** @var */` line above the declaration? It's a bit too much to add this checks everywhere

phpstan.dist.baseline.neon
```diff
@@ -500,21 +485,6 @@ parameters:
 			count: 1
 			path: app/code/core/Mage/Adminhtml/controllers/Report/StatisticsController.php
 
-		-
-			message: "#^Method Mage_Adminhtml_Model_Sales_Order_Create\\:\\:moveQuoteItem\\(\\) invoked with 2 parameters, 3 required\\.$#"
```
elidrissidev, line 504: Was it intentional to remove all these errors from the baseline? Cause this one for example wasn't resolved and shows up on the "Files" tab

app/Mage.php
fballiano: auch :-\

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
fballiano: I think the same, it is too much but...

I'll try adding a `/** @var */` and see it works

phpstan.dist.baseline.neon
fballiano: the thing is that phpstan kept telling me "these errors in the baseline are ignored and not necessary" so I kept removing, but there's for sure something wrong with my phpstan (although I've no idea what)

fballiano:
> @fballiano do you get some error about "parallel jobs didn't finish"? Cause it happens to me on macOS too and it won't show phpstan errors

@elidrissidev no I don't get any errors (I'm on a quite powerful machine), phpstan was outputting "ok no errors" at the end of the run. but there's for sure something wrong with phpstan's cache, so now I'm running this command instead

```/usr/local/opt/php@7.3/bin/php vendor/bin/phpstan clear-result-cache;
/usr/local/opt/php@7.3/bin/php vendor/bin/phpstan analyze```

which maaaybe should work, I'm starting now to re-check this PR

let's see

File: app/code/core/Mage/Adminhtml/Block/System/Config/Form.php (hunk 6 above)
Comment (fballiano) [reply app/code/core/Mage/Adminhtml/Block/System/Config/Form.php:428]: I did it other cases (in this PR) but in this specific one I'd leave it like this, what do you think?

**Comment by fballiano** [comment]
green light, finally!

**Comment by fballiano** [comment]
let's see what happens with bleedingedge

**Comment by fballiano** [comment]
still green light!

**Comment by elidrissidev** [comment]
No problem :)

File: app/code/core/Mage/Adminhtml/Block/System/Config/Form.php (hunk 6 above)
Comment (elidrissidev) [reply app/code/core/Mage/Adminhtml/Block/System/Config/Form.php:428]: Sorry for the delay. Throwing an exception could help with debugging a typo, but the message can be clarified a bit (Source model instead of entity type).

Comment (fballiano) [inline app/code/core/Mage/Adminhtml/Block/System/Config/Form.php:428]:
```suggestion
                        Mage::throwException("Source model '{$factoryName}' is not found");
```
@elidrissidev right, I kinda tried to use the same message everywhere for translation porpuses  but these messages aren't translated anyway.. fixed :-)

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
```diff
@@ -437,7 +441,7 @@ public function initFields($fieldset, $group, $section, $fieldPrefix = '', $labe
                                 $optionArray[] = ['label' => $label, 'value' => $value];
                             }
                         }
-                    } else {
+                    } elseif (method_exists($sourceModel, 'toOptionArray')) {
                         $optionArray = $sourceModel->toOptionArray($fieldType == 'multiselect');
                     }
                     $field->setValues($optionArray);
```
kiatng, line 447: `$optionArray` can be undefined. Move line 435?

image

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
```diff
@@ -637,9 +641,9 @@ public function getScope()
      */
     public function getScopeLabel($element)
     {
-        if ($element->show_in_store == 1) {
+        if ((string)$element->show_in_store == '1') {
```
kiatng, line 644: How about
```suggestion
        if ((int)$element->show_in_store === 1) {
```
since we expect `int 1|0`.

app/code/core/Mage/Paypal/Controller/Express/Abstract.php
```diff
@@ -470,7 +474,7 @@ protected function _initCheckout()
      * Combined getter/setter
      *
      * @param string $setToken
-     * @return Mage_Paypal_ExpressController|string
+     * @return Mage_Paypal_Controller_Express_Abstract|string
```
kiatng, line 477: I think it's
```suggestion
     * @return $this|string
```

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
fballiano: moved the declaration, is it enought?

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
```diff
@@ -637,9 +641,9 @@ public function getScope()
      */
     public function getScopeLabel($element)
     {
-        if ($element->show_in_store == 1) {
+        if ((int)$element->show_in_store === 1) {
             return $this->_scopeLabels[self::SCOPE_STORES];
-        } elseif ($element->show_in_website == 1) {
+        } elseif ((string)$element->show_in_website == '1') {
```
kiatng, line 646: One more:

```suggestion
        } elseif ((int)$element->show_in_website === 1) {
```

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
fballiano: done :-)

app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
```diff
@@ -425,21 +426,27 @@ public function initFields($fieldset, $group, $section, $fieldPrefix = '', $labe
...
                             foreach ($sourceModel->$method() as $value => $label) {
                                 $optionArray[] = ['label' => $label, 'value' => $value];
                             }
                         }
-                    } else {
+                    } elseif (method_exists($sourceModel, 'toOptionArray')) {
                         $optionArray = $sourceModel->toOptionArray($fieldType == 'multiselect');
                     }
+
+                    $optionArray ?? Mage::throwException("Missing method 'toOptionArray()' in source model '{$factoryName}'");
```
kiatng, line 449: I asked ChatGPT, it doesn't like it.
image

kiatng: But, its opinion can be corrected:
image

kiatng: Last:
image

elidrissidev: IMO it's confusing because it's an expression not a statement, which may give the impression that an assignment to a variable was missed. It's also not used anywhere in the code, so I prefer the longer version.

fballiano: I rewrote it in a slightly different way cause I didn't like the flow, I think it's more readable now.

kiatng (approved):

fballiano:
sorry @kiatng, I'd to fix 2 phpcs errors :-\

kiatng (approved):

elidrissidev (approved):

fballiano:
merged and v20ed

PR #3038, opened 2023-02-22
EAV Config Cache followup: select non-existent attribute exception

davidhiendl:
This PR fixes an exception that is trigger when a collection selects a non-existent attribute because `getAttribute()` no longer returns an empty model for non-existent attributes.

### Related Pull Requests
#2993

```php
$productCollections = Mage::getResourceModel("catalog/product_collection");
$productCollections->addAttributeToSelect("test_does_not_exist");
```

### DIFF
--- a/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
@@ -438,6 +438,10 @@ abstract class Mage_Eav_Model_Entity_Collection_Abstract extends Varien_Data_Col
             } else {
                 $attrInstance = Mage::getSingleton('eav/config')
                     ->getAttribute($this->getEntity()->getType(), $attribute);
+                // failure to resolve an attribute from the eav config implies it does not exist
+                if (empty($attrInstance)) {
+                    return $this;
+                }
             }
             if (empty($attrInstance)) {
                 throw Mage::exception(


davidhiendl:
@luigifab Followup fix for your issue in #2993

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -438,6 +438,10 @@ public function addAttributeToSelect($attribute, $joinType = false)
             } else {
                 $attrInstance = Mage::getSingleton('eav/config')
                     ->getAttribute($this->getEntity()->getType(), $attribute);
+                // failure to resolve an attribute from the eav config implies it does not exist
+                if (empty($attrInstance)) {
+                    return $this;
```
tmotyl, line 443: Shouldnt we log situations like this? Typo in the attribute name is a  bug, a deweloper would like to know about

davidhiendl: It's not necessarily a bug, it could just be a deleted attribute that was referenced somewhere. But perhaps throw an exception in developer mode and log it in production mode. Let me know what you prefer and I defer to you and make the change.

luigifab: For me, nothing, or only a log when dev mode is enabled.

luigifab (approved):

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
fballiano: I also don't care, if an attribute disappears there will be an errors in some template or some other code without the need to do anything here, although maybe it would be a bit cleaner

fballiano (approved):

PR #3044, opened 2023-02-24
Eav config cache followup for multi scope load

davidhiendl:
Due to the way `Mage_Eav_Model_Entity_Type::getAttributeCollection` reuses the collection and the fact that `Attribute_Collection::addStoreLabel` does not support adding multiple stores at once it's impossible to load multiple stores attributes during the same context (page request, cli call, cron job, ...).

This is a followup that adds a non-breaking bypass to this behaviour, allowing retrieval of a non-cache collection.

It also stops the API from sometimes loading more data then actually requested, due to the collection being modified (but this did not appear to break anything).

### Related Pull Requests
#2993

1. Fixes https://github.com/OpenMage/magento-lts/pull/2993#issuecomment-1435667814
2. Fixes https://github.com/OpenMage/magento-lts/pull/2993#issuecomment-1443027162

1. Disable cache
2. Create multiple store views (2+)
3. Create order in demo shop
4. Visit order page in admin

### DIFF
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -209,7 +209,7 @@ class Mage_Eav_Model_Config
     protected function _loadEntityAttributes($entityType, $storeId)
     {
         // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
-        $entityAttributes = $entityType->newAttributeCollection(null, false)
+        $entityAttributes = $entityType->newAttributeCollection()
             ->addStoreLabel($storeId)
             ->getData();
 
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -209,7 +209,7 @@ class Mage_Eav_Model_Config
     protected function _loadEntityAttributes($entityType, $storeId)
     {
         // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
-        $entityAttributes = $entityType->getAttributeCollection(null, false)
+        $entityAttributes = $entityType->newAttributeCollection(null, false)
             ->addStoreLabel($storeId)
             ->getData();
 
@@ -221,7 +221,7 @@ class Mage_Eav_Model_Config
             $attributeId = $entityAttributeData['attribute_id'];
             $attributeCode = $entityAttributeData['attribute_code'];
 
-            // workaround for getAttributeCollection()->getData() returning all columns as string
+            // workaround for newAttributeCollection()->getData() returning all columns as string
             foreach (self::NUMERIC_ATTRIBUTE_COLUMNS as $key) {
                 if (!isset($entityAttributeData[$key])) {
                     continue;
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -209,7 +209,7 @@ class Mage_Eav_Model_Config
     protected function _loadEntityAttributes($entityType, $storeId)
     {
         // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
-        $entityAttributes = $entityType->getAttributeCollection()
+        $entityAttributes = $entityType->getAttributeCollection(null, false)
             ->addStoreLabel($storeId)
             ->getData();
 
--- a/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
+++ b/app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
@@ -438,6 +438,10 @@ abstract class Mage_Eav_Model_Entity_Collection_Abstract extends Varien_Data_Col
             } else {
                 $attrInstance = Mage::getSingleton('eav/config')
                     ->getAttribute($this->getEntity()->getType(), $attribute);
+                // failure to resolve an attribute from the eav config implies it does not exist
+                if (empty($attrInstance)) {
+                    return $this;
+                }
             }
             if (empty($attrInstance)) {
                 throw Mage::exception(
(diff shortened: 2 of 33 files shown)


davidhiendl:
@luigifab @darinda Thanks for reporting. Here is the PR if you need this right away.

app/code/core/Mage/Eav/Model/Entity/Type.php
```diff
@@ -99,24 +99,33 @@ public function loadByCode($code)
...
      * Retrieve entity type attributes collection
      *
-     * @param   int $setId
-     * @return  Mage_Eav_Model_Resource_Entity_Attribute_Collection
+     * @param int|null $setId
+     * @param bool $useCache reuse local cache for collection
+     * @return Mage_Eav_Model_Resource_Entity_Attribute_Collection
      */
-    public function getAttributeCollection($setId = null)
+    public function getAttributeCollection(int $setId = null, bool $useCache = true)
```
fballiano, line 106: changing the signature breaks overridden classes, should we do something about that or we consider it ok since it's v20?

davidhiendl: Oh yeah my bad, it was just reflex adding the types to parameters. I will remove them.

luigifab (dismissed):
Tested with PHP 8.0 and 8.2.

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
```diff
@@ -438,6 +438,10 @@ public function addAttributeToSelect($attribute, $joinType = false)
             } else {
                 $attrInstance = Mage::getSingleton('eav/config')
                     ->getAttribute($this->getEntity()->getType(), $attribute);
+                // failure to resolve an attribute from the eav config implies it does not exist
+                if (empty($attrInstance)) {
```
kiatng, line 442: Depending on issue #3057, may be need to check for `!$attrInstance->getId()` too.

davidhiendl: Yes, just waiting for 2nd comment in #3057 but I also agree that we probably should just return an empty model. Once we have a 3rd opinion I will create another PR.

app/code/core/Mage/Eav/Model/Entity/Type.php
```diff
@@ -99,24 +99,33 @@ public function loadByCode($code)
...
      * Retrieve entity type attributes collection
      *
-     * @param   int $setId
-     * @return  Mage_Eav_Model_Resource_Entity_Attribute_Collection
+     * @param int|null $setId
+     * @param bool $useCache reuse local cache for collection
+     * @return Mage_Eav_Model_Resource_Entity_Attribute_Collection
      */
-    public function getAttributeCollection($setId = null)
+    public function getAttributeCollection($setId = null, $useCache = true)
```
fballiano, line 106: do we have to change the signature of this public method?

davidhiendl: If I want to re-use the code, mostly yes. But since it's an optional parameter it should not create any problems.
Alternatively I can refactor the code into a second method `newAttributeCollection` that always returns a fresh one, which is then used by `getAttributeCollection` as needed. Which do you prefer?

fballiano: the thing is (but I wanted to have other people join the conversation) that modifying the signature of a public method should be considered a major change, since all classes extending that method (which, ok, most probably are zero in the world) will break

justinbeaty: I think changing the signature is okay in this case, especially since we're still on RC for v20.1.0. The error should appear very quickly upon testing if anyone has override the method, and it's a quick fix.

Generally though, if it makes more sense to introduce a new method, that would be preferable in most cases. But, we also don't want to litter the code with slight variations of similar method names to avoid this problem.

fballiano: do we need the useCache param tho? it seems that's just used in the initialization, isn't it possible to find a way to detect that we are in the initialization phase without the param? maybe I'm just blabbering

fballiano: nothing magic, if some property doesn't exists it means cache was not loaded etc..

fballiano: something like

Screenshot 2023-03-09 alle 14 38 09

davidhiendl: I know, but it's a delicate problem - you cannot predict what someone else does to a "cached" and therefor shared collection downstream, which caused the problem with for example adding store labels to it that triggered the backend exception when multiple scopes were initialized in a single context.

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -209,7 +209,7 @@ protected function _loadEntityTypes()
     protected function _loadEntityAttributes($entityType, $storeId)
     {
         // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
-        $entityAttributes = $entityType->getAttributeCollection()
+        $entityAttributes = $entityType->newAttributeCollection(null, false)
```
fballiano, line 212:
```suggestion
        $entityAttributes = $entityType->newAttributeCollection()
```

first param should be null by default, second one doesn't exist if i'm not mistaken

davidhiendl:
Is it possible to get this merged, would make it easier to get the changes for #3057 committed and into a PR?

fballiano (approved):

fballiano:
@luigifab could you check it again?

fballiano:
> Is it possible to get this merged, would make it easier to get the changes for #3057 committed and into a PR?

when we get another green light I'll merge it right away. I keep commenting stuff cause I hate when PRs go sleepy.

app/code/core/Mage/Eav/Model/Entity/Collection/Abstract.php
fballiano: @davidhiendl this has beed addressed and we can consider it closed?

davidhiendl: @fballiano It will be addressed in the separate PR that will fix #3057 as it does not really belong here but in that PR instead.

ADDISON74 (approved):

Issue #3046, opened 2023-02-27
Undefined array key warning in Mage_Catalog_Model_Resource_Category_Flat when having Category Attributes removed.


pisc-software:
### Preconditions (*)

1. OpenMage 20.0.0.20 and older
2. Magento 1.9

### Steps to reproduce (*)

1. Have Catalog Category as flat table
2. Remove Category Attribute that was used on categories and data is present in the EAV Tables
3. Run a Reindex of the Category Flat Data
4. system.log is showing "Undefined array key XXX" where XXX is the attribute_id of the removed attribute. On large category structures this will flood the system.log unnecessary.

### Expected result (*)

1. No log warning as array access is validated

### Actual result (*)

1. Log warning as above.

### Bug Fix
Fix required in 'Mage_Catalog_Model_Resource_Category_Flat':

```
class Mage_Catalog_Model_Resource_Category_Flat extends Mage_Index_Model_Resource_Abstract
{
 // Line 841 - Add array index check:
    /**
     * Return attribute values for given entities and store
     *
     * @param array $entityIds
     * @param int $store_id
     * @return array
     */
    protected function _getAttributeValues($entityIds, $store_id)
    {
        if (!is_array($entityIds)) {
            $entityIds = [$entityIds];
        }
        $values = [];

        foreach ($entityIds as $entityId) {
            $values[$entityId] = [];
        }
        $attributes = $this->_getAttributes();
        $attributesType = [
            'varchar',
            'int',
            'decimal',
            'text',
            'datetime'
        ];

        /**
         * Fix: Verify that attribute_id is in attribute list
         */
        foreach ($attributesType as $type) {
            foreach ($this->_getAttributeTypeValues($type, $entityIds, $store_id) as $row) {
                if (isset($attributes[$row['attribute_id']]))
                    $values[$row['entity_id']][$attributes[$row['attribute_id']]['attribute_code']] = $row['value'];
            }
        }
        return $values;
    }

}

```

fballiano:
thank you for reporting it, can you check https://github.com/OpenMage/magento-lts/pull/3047?

PR #3047, opened 2023-02-27
Fixed undefined array key warning in Mage_Catalog_Model_Resource_Category_Flat

fballiano:
Everything is explained in https://github.com/OpenMage/magento-lts/issues/3046

1. Fixes https://github.com/OpenMage/magento-lts/issues/3046


pisc-software:
Yes, that is correct as the fix we applied also to a customer store. We debugged on that store, and also checked the system.log today. It fixes the issue.
Many thanks for the commit.

pisc-software (approved):
Yes, that is the exact change we applied to a production store. We checked also the store today and it fixes the warning message also in system.log as expected.

ADDISON74 (approved):

elidrissidev (approved):

fballiano:
merged and v20ed

PR #3049, opened 2023-02-28
Make PHPStan run independently from PHPCS

fballiano:
In our workflows PHPStan runs only when PHP-CS/PHP-CS-Fixer are finished, but I think it shouldn't be that way, I'd prefer to have PHPStan run independently from PHP-CS so that we may find problems in more than a single "point" at the sametime.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -73,7 +73,7 @@ jobs:
 
   phpstan:
     name: PHPStan
-    needs: [check, composer, phpcs, php-cs-fixer]
+    needs: [check, composer]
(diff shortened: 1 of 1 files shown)


sreichel:
The idea was to fail fast on code style errors ...

fballiano:
Obviously, but it’s also making us wait minutes for phpstan to run and many times they check different stuff that can run at the same time

fballiano:
plus phpstan now it's really fast to run

kiatng (approved):

colinmollenhour (approved):

PR #3052, opened 2023-03-01
Enabled PHP 8.2 in composer.json, updated README with software compatibility info

fballiano:
I'm testing OM with PHP8.2 and it works for me as good as 8.1, we should at least let people install OM via composer with 8.2.

This PR enables that and updates the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,18 +47,13 @@ level of backwards compatibility to the official releases.
 
 ## Requirements
 
-- PHP 7.3+ (PHP 8.0 is supported, PHP 8.1 is work in progress)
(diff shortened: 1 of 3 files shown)


elidrissidev (approved):

kiatng (approved):

fballiano:
merged and v20ed

ADDISON74:
I think that warning about the possible malfunction of some extensions should have been kept. We assume no responsibility if someone migrates to a PHP 8.x version and has trouble with 3rd party source code.

fballiano:
I really think it’s something people should already know. Its like installing windows 11 without knowing that you may have issues with an old version of photoshop or something. 
Hope i explained it properly but that’s why i felt it wasn’t needed

fballiano:
Also, this readme is clearly written for developers, not end users, so they have to know

PR #3053, opened 2023-03-02
Make Prototype Validator optional

loekvangool:
### Description (*)
I'm removing `prototype/validation.js` from some of my pages, but some shared JavaScript libs assume it's still there and start to generate errors.

The validator is not necessary on every page, hence it should be possible to remove it without running into errors.

In the slightly longer term, I'd like to see the whole thing removed and instead switch to the new HTML5 validation options, which are plenty: https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation.

### DIFF
--- a/js/varien/form.js
+++ b/js/varien/form.js
@@ -28,7 +28,9 @@ VarienForm.prototype = {
         this.cache      = $A();
         this.currLoader = false;
         this.currDataIndex = false;
-        this.validator  = new Validation(this.form);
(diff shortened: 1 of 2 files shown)


loekvangool:
Related to #3054

fballiano:
at a first glance it should be ok

fballiano (approved):
why not

fballiano:
merged and 20ed

PR #3056, opened 2023-03-03
bugfix: return first cart item when adding bundle product to cart

eneiasramos:
### Description (*)

This fixes the cart item returned by the method Mage_Sales_Model_Quote::addProductAdvanced()

1. Try to add some bundle product with children product into cart
2. Compare the cart item instance returned by the method Mage_Sales_Model_Quote::addProductAdvanced()

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -1110,7 +1110,7 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
 
         Mage::dispatchEvent('sales_quote_product_add_after', ['items' => $items]);
 
-        return reset($items);
+        return $parentItem;
     }
 
     /**
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -1110,8 +1110,7 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
 
         Mage::dispatchEvent('sales_quote_product_add_after', ['items' => $items]);
 
-        reset($items);
-        return current($items);
+        return reset($items);
     }
 
     /**
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -1110,7 +1110,9 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
 
         Mage::dispatchEvent('sales_quote_product_add_after', ['items' => $items]);
 
-        return $item;
+        reset($items);
+
+        return current($items);
     }
 
     /**
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -1111,7 +1111,6 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
         Mage::dispatchEvent('sales_quote_product_add_after', ['items' => $items]);
 
         reset($items);
-
         return current($items);
     }
 


app/code/core/Mage/Sales/Model/Quote.php
```diff
@@ -1110,7 +1110,9 @@ public function addProductAdvanced(Mage_Catalog_Model_Product $product, $request
 
         Mage::dispatchEvent('sales_quote_product_add_after', ['items' => $items]);
 
-        return $item;
+        reset($items);
+
```
fballiano, line 1114:
```suggestion
```

fballiano:
what problem is this difference creating?

eneiasramos:
@fballiano I updated the code to use only one method :)

fballiano:
nice!

I still didn't really understand the problem though 😅

eneiasramos:
> nice!
> 
> I still didn't really understand the problem though sweat_smile

The problem happens when we add some composite product (like bundle) to the cart via API.

If we try to manipulate the cart item after its inclusion into cart (for example setting a custom price to composite product by item), we got the last simple product from the bundle (not the bundle item itself).

So the fixes returns the first item of array (the cart item of bundle product itself)

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Checkout/Model/Cart/Product/Api.php#L73

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Sales/Model/Quote.php#L1074

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Sales/Model/Quote.php#L1078

https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Sales/Model/Quote.php#L1113

fballiano (approved):

kiatng (dismissed):

elidrissidev:
I saw there were a couple of places in the code that operate on the returned item, are you sure that won't cause any issues?

fballiano:
I was quickly checking but... mmmmmmm... this code:

Screenshot 2023-03-05 alle 12 48 33

$item can never be a string no?

elidrissidev:
> $item can never be a string no?

IIRC a string is being returned somewhere if there was an error when adding the product.

eneiasramos:
> I was quickly checking but... mmmmmmm... this code:
> 
> Screenshot 2023-03-05 alle 12 48 33
> 
> $item can never be a string no?

Where is this code exactly?

eneiasramos:
> > $item can never be a string no?
> 
> IIRC a string is being returned somewhere if there was an error when adding the product.

Did you test with the fix?

kiatng:
Can `$item` be a string?

https://github.com/OpenMage/magento-lts/blob/74beed6ab7a4aab46a48c55f0d8af0b8614a729a/app/code/core/Mage/Adminhtml/Model/Sales/Order/Create.php#L790-L795

Ref Line 795 ` if (is_string($item)) { `, `$item` can be indeed a string:

https://github.com/OpenMage/magento-lts/blob/74beed6ab7a4aab46a48c55f0d8af0b8614a729a/app/code/core/Mage/Sales/Model/Quote.php#L1054-L1062

kiatng:
@eneiasramos Instead of `return reset($items);`, is it the same as `return $parentItem;` ? This can make it more readable.

eneiasramos:
> I saw there were a couple of places in the code that operate on the returned item, are you sure that won't cause any issues?

Yes! I'm sure!

eneiasramos:
> @eneiasramos Instead of `return reset($items);`, is it the same as `return $parentItem;` ? This can make it more readable.

Perfect! I updated the code. Thanks!

https://github.com/OpenMage/magento-lts/pull/3056/commits/19c9cebfa19670a4a18ff993ebde0651f1030a06#diff-9f9291d61e7064ad1f3c1ce6914102e92c722114b8f0abee79d9dfb3df3d20b1R1113

fballiano (approved):

ADDISON74 (approved):

fballiano:
merged and v20ed

eneiasramos:
Thank you my dears!

PR #2662, opened 2022-10-19
Fixes #2658

sreichel:
### Description (*)

#2603 was for filter disabled products in admin backend / create orders. Some changes have inpact on frontend and should be reverted.

### Related Pull Requests

1. Fixes OpenMage/magento-lts#2603

### Fixed Issues (if relevant)

1. Fixes OpenMage/magento-lts#2658
2. Closes OpenMage/magento-lts#2659
3. Closes OpenMage/magento-lts#2660

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3060, opened 2023-03-03
Remove status filter from Product Collection in CatalogSearch Layer

elidrissidev:
### Description (*)

A followup to #2662 which removed the `status` filter from Product Collection in Catalog Layer to fix the issue discussed in https://github.com/OpenMage/magento-lts/pull/2603#issuecomment-1451514520.

This issue could be fixed by #2660, but this change is just to keep the consistency between Catalog and CatalogSearch Layers.

### Related Pull Requests

- See #2662
- See #2603

### DIFF
--- a/app/code/core/Mage/CatalogSearch/Model/Layer.php
+++ b/app/code/core/Mage/CatalogSearch/Model/Layer.php
@@ -60,10 +60,7 @@ class Mage_CatalogSearch_Model_Layer extends Mage_Catalog_Model_Layer
             ->addPriceData()
             ->addTaxPercents()
             ->addStoreFilter()
-            ->addUrlRewrite()
-            ->addAttributeToFilter('status', [
-                'in' => Mage::getSingleton('catalog/product_status')->getVisibleStatusIds()
-            ]);
+            ->addUrlRewrite();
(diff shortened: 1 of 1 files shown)


fballiano:
question, are we loosing the status filter completely or it was duplicate so this way we're avoiding the duplicate?

elidrissidev:
> question, are we loosing the status filter completely or it was duplicate so this way we're avoiding the duplicate?

We're avoiding the duplicate only when flat tables are enabled, else the filter is never added to the collection.

elidrissidev:
An important step to reproduce this issue is to make sure `price` attribute has "Use In Search Results Layered Navigation" value set to "Yes", otherwise this issue will not occur.

fballiano:
but we should have the status filter if flat is not enabled no?

elidrissidev:
It wasn't there before, and it was also removed from the Catalog view, so Idk...

fballiano:
I thought I added the filter on status, cause disabled product should never be visible on frotend, and there was the method for that but the method was empty... I don't know...

fballiano (approved):

kiatng (approved):

PR #3070, opened 2023-03-08
Moved dependabot to weekly again

fballiano:
PHPStan releases a new release every few days, we do not need to keep approving dependabot upgrades all the time, if dependabot would run once per week it would be more than enough, so that we don not get flooded with phpstan updates and we can focus on other activities too :-)


elidrissidev (approved):
Was just about to do it 😄

fballiano:
merged and v20ed

Issue #3071, opened 2023-03-08
Import Currency Rates with Fixer API / Apilayer.com


TangLiang:
As the Fixer API changed,it's not work now.
Need update:
Mage_Directory_Model_Currency_Import_Fixerio.php

Changed code:
```php
protected $_url = 'https://api.apilayer.com/currency_data/live?source={{CURRENCY_FROM}}&currencies={{CURRENCY_TO}}';
```
```php
    protected function _convertBatch(array $data, string $currencyFrom, array $currenciesTo): array
    {
        $accessKey = Mage::getStoreConfig(self::XML_PATH_FIXERIO_API_KEY);
        if (empty($accessKey)) {
            $this->_messages[] = Mage::helper('directory')
                ->__('No API Key was specified or an invalid API Key was specified.');
            $data[$currencyFrom] = $this->_makeEmptyResponse($currenciesTo);
            return $data;
        }
        $currenciesImploded = implode(',', $currenciesTo);
        $url = str_replace(
            ['{{CURRENCY_FROM}}', '{{CURRENCY_TO}}'],
            [$currencyFrom, $currenciesImploded],
            $this->_url
        );
        $timeLimitCalculated = 2 * (int) Mage::getStoreConfig(self::XML_PATH_FIXERIO_TIMEOUT)
            + (int) ini_get('max_execution_time');
        @set_time_limit($timeLimitCalculated);
        try {
            $response = $this->_getServiceResponse($url,$accessKey);
        } catch (Exception $e) {
            ini_restore('max_execution_time');
        }
        if (!$this->_validateResponse($response, $currencyFrom)) {
            $data[$currencyFrom] = $this->_makeEmptyResponse($currenciesTo);
            return $data;
        }
        foreach ($currenciesTo as $currencyTo) {
            if ($currencyFrom == $currencyTo) {
                $data[$currencyFrom][$currencyTo] = $this->_numberFormat(1);
            } else {
                if (empty($response['quotes'])) {
                    $this->_messages[] = Mage::helper('directory')
                        ->__('We can\'t retrieve a rate from %s for %s.', $url, $currencyTo);
                    $data[$currencyFrom][$currencyTo] = null;
                } else {
                    $data[$currencyFrom][$currencyTo] = $this->_numberFormat((float) $response['quotes'][$currencyFrom.$currencyTo]);
                }
            }
        }
        return $data;
    }
```
```php
    protected function _getServiceResponse(string $url, $accessKey, int $retry = 0): array
[...]

PR #3072, opened 2023-03-08
Fixed import currency rates with Fixer API / Apilayer.com, ref #3071

fballiano:
I've just put the code provided in https://github.com/OpenMage/magento-lts/issues/3071 in a PR, didn't really check it.

### DIFF
--- a/app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
+++ b/app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
@@ -43,7 +43,7 @@ class Mage_Directory_Model_Currency_Import_Fixerio extends Mage_Directory_Model_
      *
      * @var string
      */
-    protected $_url = 'https://api.apilayer.com/fixer/latest?apikey={{ACCESS_KEY}}&base={{CURRENCY_FROM}}&symbols={{CURRENCY_TO}}';
+    protected $_url = 'https://api.apilayer.com/currency_data/live?source={{CURRENCY_FROM}}&currencies={{CURRENCY_TO}}';
 
     /**
      * Information messages stack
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
```diff
@@ -162,16 +156,16 @@ protected function _convertBatch(array $data, $currencyFrom, array $currenciesTo
      * @param int $retry
      * @return array
      */
-    protected function _getServiceResponse($url, $retry = 0)
+    protected function _getServiceResponse(string $url, $accessKey, int $retry = 0): array
```
fballiano, line 159: I'm wondering about this one...

kiatng:
I investigated this. Here's what I found:

1. https://apilayer.com/ is a marketplace for APIs, anyone can publish his/her APIs on the platform.
2. There are currently 3 currency exchange APIs listed: [FIXER API](https://apilayer.com/marketplace/fixer-api) (in which the original OM is based on), [CURRENCY DATA API](https://apilayer.com/marketplace/currency_data-api) (this PR is trying to change), and [EXCHANGE RATES DATA API](https://apilayer.com/marketplace/exchangerates_data-api)
3. I subscribed to the Fixer API to obtain the apikey. Then, I tested the URL: `https://api.apilayer.com/fixer/latest?source=USD&currencies=MYR&apikey={my_api_key}` in the browser, it returns some result without error.

Based on my preliminary investigation, this PR is wrong. It is subscribing the Currency Data API whereas the original OM code is for the Fixer API. For Currency Data API, it should be in a different class: `Mage_Directory_Model_Currency_Import_CurrencyData.php`

Did I miss something?

fballiano:
so maybe we could have multiple adapter for Fixer, CurrencyData and EchangeRatesData, it would be nice

Issue #3057, opened 2023-03-03
EAV Config Cache Bug: getAttribute() returns false at creating new attribute


kiatng:
### Preconditions (*)

Related to issue #2993. When investigating issue #3055, I was able to create a new attribute in ver 20.0.16 with this:

```php
        $store = Mage::app()->getStore(Mage_Core_Model_App::ADMIN_STORE_ID);
        $eavConfig = Mage::getSingleton('eav/config');
        $attributes = array(
            'select_invoice_type' => array(
                'frontend_label'        => 'Identification for Electronic Invoicing',
                'label'                 => 'Ident Type Billing',
                'backend_type'          => 'text',
                'frontend_input'        => 'select',
                'source'                => 'eav/entity_attribute_source_table',
                'is_user_defined'       => 1,
                'is_system'             => 0,
                'is_visible'            => 1,
                'is_required'           => 0,
            )
        );

        foreach ($attributes as $attributeCode => $data) {
            $attribute = $eavConfig->getAttribute('customer_address', $attributeCode);
            $attribute->setWebsite($store->getWebsite());
            $attribute->addData($data);
            $usedInForms = array(
                'adminhtml_customer_address',
                'customer_address_edit',
                'customer_register_address',
                'customer_account_create'
            );

            $attribute->setData('used_in_forms', $usedInForms);
            $attribute->save();
        }
```

However, with ver 20.1.0-rc2 I get 
> Fatal error:  Uncaught Error: Call to a member function setWebsite() on bool in .../Celera/Crib/controllers/PxController.php:55

The line is `$attribute->setWebsite($store->getWebsite());`.

### Steps to reproduce (*)

1. As described.
2. Or just examine the output of `Mage::getSingleton('eav/config')->getAttribute('customer_address', 'select_invoice_type')`.

### Expected result (*)

1. To be able to create new attributes as older version without error.
2. `Mage::getSingleton('eav/config')->getAttribute('customer_address', 'select_invoice_type')` returns `Mage_Customer_Model_Attribute`

### Actual result (*)

1. Fatal error in latest ver.
2. `Mage::getSingleton('eav/config')->getAttribute('customer_address', 'select_invoice_type')` returns `false`

davidhiendl:
This happens because getAttribute no longer returns an empty model when the attribute does not exist. I think it's not a good idea to use this behavior to create an attribute in the first place.

The way core installers add attributes is, for example:
```php
$installer->addAttribute('customer', 'confirmation', [
[...]

Issue #3055, opened 2023-03-02
Error Invalid method Mage_Customer_Model_Attribute::_getDefaultSourceModel


gbgssoftware:
### Preconditions (*)

1.OpenMage 19.4.18
2.

### Steps to reproduce (*)

I use a form that adds fields to customer_address
this is the source:

```php
$setup = $this;
$setup->startSetup();

$addressHelper = Mage::helper('customer/address');
$store         = Mage::app()->getStore(Mage_Core_Model_App::ADMIN_STORE_ID);
$eavConfig = Mage::getSingleton('eav/config');

$attributes = array(
    'field1' => array(
        'frontend_label'        => 'field1',
        'label'                 => 'field1',
        'backend_type'          => 'varchar',
        'frontend_input'        => 'text',
        'is_user_defined'       => 1,
        'is_system'             => 0,
        'is_visible'            => 1,
        'is_required'           => 0,
    ),
    'field2' => array(
        'frontend_label'        => 'field2',
        'label'                 => 'field2',
        'backend_type'          => 'varchar',
        'frontend_input'        => 'text',
        'is_user_defined'       => 1,
        'is_system'             => 0,
        'is_visible'            => 1,
        'is_required'           => 0,
    ),
    'select_invoice_type' => array(
        'frontend_label'        => 'Identification for Electronic Invoicing',
        'label'                 => 'Ident Type Billing',
        'backend_type'          => 'text',
        'frontend_input'        => 'select',
        'source'                => 'eav/entity_attribute_source_table',
        'is_user_defined'       => 1,
        'is_system'             => 0,
        'is_visible'            => 1,
        'is_required'           => 0,
  )
);




foreach ($attributes as $attributeCode => $data) {
  $attribute = $eavConfig->getAttribute('customer_address', $attributeCode);
  $attribute->setWebsite($store->getWebsite());
  $attribute->addData($data);
  $usedInForms = array(
      'adminhtml_customer_address',
      'customer_address_edit',
      'customer_register_address',
      'customer_account_create'
  );

  $attribute->setData('used_in_forms', $usedInForms);
  $attribute->save();
}
```

When logging into magento, I get the following message:

```
[...]

PR #3074, opened 2023-03-09
EAV Config Cache revert return empty attribute model

davidhiendl:
Changes made in #2993 changed the behaviour of `getAttribute` to return `false` instead of an empty model. While this might be a more desirable return type it breaks some behavior that relies on it. This PR reverts that change and returns a fresh empty model when `getAttribute("non_existent_code")` is called.

### Related Pull Requests
#2993

1. #3057
2. #3055 (underlying cause listed in #3057)

### DIFF
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -209,7 +209,7 @@ class Mage_Eav_Model_Config
     protected function _loadEntityAttributes($entityType, $storeId)
     {
         // preload attributes in array form to avoid instantiating models for every attribute even if it is never accessed
-        $entityAttributes = $entityType->newAttributeCollection(null, false)
+        $entityAttributes = $entityType->newAttributeCollection()
             ->addStoreLabel($storeId)
             ->getData();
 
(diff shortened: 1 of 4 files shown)


app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -484,6 +484,11 @@ public function getAttribute($entityType, $code, $storeId = null)
             $attribute = $this->_getDefaultAttributeIfExists($entityType, $code, $storeId);
         }
 
+        // maintain old behavior and return an empty model to avoid breaking compatibility
```
fballiano, line 487:
```suggestion
```

davidhiendl: I think it should stay, as there is no clear explanation why an empty model is returned instead of false otherwise.

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -484,6 +484,10 @@ public function getAttribute($entityType, $code, $storeId = null)
             $attribute = $this->_getDefaultAttributeIfExists($entityType, $code, $storeId);
         }
 
+        if (!$attribute) {
```
fballiano, line 487:
```suggestion
        // return an empty model to avoid breaking compatibility
        if (!$attribute) {
```

app/code/core/Mage/Eav/Model/Config.php
fballiano: restored it but slightly changed

fballiano (approved):

PR #3078, opened 2023-03-10
Fixed "if newsletter module disabled, customer backend would blank without notice"

sidealice:
removeable for the wishlist/giftmessage/sendtofriend/newsletter modules

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
@@ -73,13 +73,15 @@ class Mage_Adminhtml_Block_Customer_Edit_Tabs extends Mage_Adminhtml_Block_Widge
                 'url'       => $this->getUrl('*/*/wishlist', ['_current' => true]),
             ]);
 
-            /** @var Mage_Adminhtml_Block_Customer_Edit_Tab_Newsletter $block */
-            $block = $this->getLayout()->createBlock('adminhtml/customer_edit_tab_newsletter');
-            if (Mage::getSingleton('admin/session')->isAllowed('newsletter/subscriber')) {
-                $this->addTab('newsletter', [
-                    'label'     => Mage::helper('customer')->__('Newsletter'),
-                    'content'   => $block->initForm()->toHtml()
-                ]);
+            if (Mage::helper('core')->isModuleOutputEnabled('Mage_Newsletter')) {
+                /** @var Mage_Adminhtml_Block_Customer_Edit_Tab_Newsletter $block */
+                $block = $this->getLayout()->createBlock('adminhtml/customer_edit_tab_newsletter');
+                if (Mage::getSingleton('admin/session')->isAllowed('newsletter/subscriber')) {
+                    $this->addTab('newsletter', [
+                        'label'     => Mage::helper('customer')->__('Newsletter'),
+                        'content'   => $block->initForm()->toHtml()
+                    ]);
+                }
             }
 
             if (Mage::getSingleton('admin/session')->isAllowed('catalog/reviews_ratings')) {
--- a/app/code/core/Mage/Adminhtml/Block/Sales/Order/View/Giftmessage.php
+++ b/app/code/core/Mage/Adminhtml/Block/Sales/Order/View/Giftmessage.php
@@ -50,10 +50,14 @@ class Mage_Adminhtml_Block_Sales_Order_View_Giftmessage extends Mage_Adminhtml_B
      */
     protected function _beforeToHtml()
     {
-        if ($this->getParentBlock() && ($order = $this->getOrder())) {
-            $this->setEntity($order);
+        if (Mage::helper('core')->isModuleOutputEnabled('Mage_Giftmessage')) {
+            if ($this->getParentBlock() && ($order = $this->getOrder())) {
+                $this->setEntity($order);
+            }
+            return parent::_beforeToHtml();
+        } else {
+            return parent::_beforeToHtml();
         }
-        return parent::_beforeToHtml();
     }
 
     /**
@@ -63,15 +67,16 @@ class Mage_Adminhtml_Block_Sales_Order_View_Giftmessage extends Mage_Adminhtml_B
      */
     protected function _prepareLayout()
     {
-        $this->setChild(
-            'save_button',
-            $this->getLayout()->createBlock('adminhtml/widget_button')
-                ->setData([
-                    'label'   => Mage::helper('giftmessage')->__('Save Gift Message'),
-                    'class'   => 'save'
-                ])
-        );
-
+        if (Mage::helper('core')->isModuleOutputEnabled('Mage_Giftmessage')) {
+            $this->setChild(
+                'save_button',
+                $this->getLayout()->createBlock('adminhtml/widget_button')
+                    ->setData([
+                        'label'   => Mage::helper('giftmessage')->__('Save Gift Message'),
+                        'class'   => 'save'
+                    ])
+            );
+        }
         return $this;
     }
 
--- a/app/etc/modules/Mage_Captcha.xml
+++ b/app/etc/modules/Mage_Captcha.xml
@@ -21,9 +21,7 @@
             <codePool>core</codePool>
             <depends>
                 <Mage_Adminhtml/>
-                <Mage_Customer/>
-                <Mage_Sendfriend/>
-                <Mage_Wishlist/>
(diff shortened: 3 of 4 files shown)


ADDISON74:
This PR does not have a proper description. It would be natural to understand the issue, how to reproduce it and how to fix it. Looking in the source code and understanding the changes is not a thing at all a professional approach.

kiatng:
I cannot replicate this issue:
1. I disable `Mage_Newsletter` in the _Advanced_ tab in backend System Configuration.
2. I disable the cache for Configurations, Layouts, and Blocks.
3. I navigate to a backend customer page.
4. I can see the tab Newsletter, on the page it's a grid. I do not see a blank page.

Question: did you remove the Newsletter code from the core?

kiatng:
A suggestion on PR: limit the file changes to one issue to make it easier to review. You have changes to files that do not seem to relate to Newsletter. Can you please remove those changes and commit again?

sidealice:
> I cannot replicate this issue:
> 
> 1. I disable `Mage_Newsletter` in the _Advanced_ tab in backend System Configuration.
> 2. I disable the cache for Configurations, Layouts, and Blocks.
> 3. I navigate to a backend customer page.
> 4. I can see the tab Newsletter, on the page it's a grid. I do not see a blank page.
> 
> Question: did you remove the Newsletter code from the core?

Yes, if disabled the module or remove the code you can see the problem.
And this would help to sepreate the modules out of core. In the future OM can maintain necessary core, the other modules e.g. newsletter/wishlist/giftmessage/sendtofriend could be a module better.

sidealice:
> This PR does not have a proper description. It would be natural to understand the issue, how to reproduce it and how to fix it. Looking in the source code and understanding the changes is not a thing at all a professional approach.

Yes, I may update the description to ' support remove the wishlist/giftmessage/sendtofriend/newsletter module without issues'

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
elidrissidev, line 2: No need for new lines after opening tag

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
```diff
@@ -80,13 +81,15 @@ protected function _beforeToHtml()
...
-                ]);
+            if (Mage::helper('core')->isModuleOutputEnabled('Mage_Newsletter')) {
+                /** @var Mage_Adminhtml_Block_Customer_Edit_Tab_Newsletter $block */
+                $block = $this->getLayout()->createBlock('adminhtml/customer_edit_tab_newsletter');
+                if (Mage::getSingleton('admin/session')->isAllowed('newsletter/subscriber')) {
+                    $this->addTab('newsletter', [
+                        'label'     => Mage::helper('customer')->__('Newsletter'),
+                        'content'   => $block->initForm()->toHtml()
+                    ]);
+                }
```
elidrissidev, line 92: I wonder if we can move this code out into the newsletter module and instead add the tab through layout xml 🤔. Maybe someone with more experience in this area can clarify if that's possible.

kiatng: +1 totally possible.

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
fballiano, line 2:
```suggestion
```

app/etc/modules/Mage_Captcha.xml
```diff
@@ -27,9 +27,7 @@
             <codePool>core</codePool>
             <depends>
                 <Mage_Adminhtml/>
-                <Mage_Customer/>
-                <Mage_Sendfriend/>
-                <Mage_Wishlist/>
+                <Mage_Customer/>
```
fballiano, line 30: is the change to this file necessary for this PR?

lib/Varien/Data/Form/Abstract.php
```diff
@@ -139,7 +140,13 @@ public function addField($elementId, $type, $config, $after = false)
...
             $className = 'Varien_Data_Form_Element_' . ucfirst(strtolower($type));
         }
-        $element = new $className($config);
+
+        if (class_exists($className)) {
+            $element = new $className($config);
+        } else {
+            $className = 'Varien_Data_Form_Element_Note';
+            $element = new $className($config);
+        }
```
fballiano, line 149: is this also neecssary for this PR? cause I normally would prefer the error instead

kiatng:
As suggested by @elidrissidev, we need to do a proper separation of concerns. `Mage_Customer` depends on the following modules:

https://github.com/OpenMage/magento-lts/blob/5318b8617612f07a2a10882c3c4eed262726eb3f/app/etc/modules/Mage_Customer.xml#L28-L32

All other modules here should be separated:

1. `Mage_Sales`
2. `Mage_Catalog`
3. `Mage_Newsletter`
4. `Mage_Wishlist`

Besides the tabs, which can be easily added with the event `adminhtml_block_widget_tabs_html_before` (see PR #1358), the rendering of the pages need to be separated as well:

image

Related to PR #2455. It means also PR #2533. And down the rabbit hole. What should we do?

app/etc/modules/Mage_Captcha.xml
sidealice: if we want the module of sendfriend/wishlist independent to other modules we have to modify this file. otherwise we can't remove wishlist/sendfriend module.

sidealice: and I checked the captcha using event and not deep bind to other modules. thus we can remove sendfriend and wishlist.

lib/Varien/Data/Form/Abstract.php
sidealice: if not add these lines, we would meet errors in the customer/view and order/view page at the backend.
because sometimes newsletter or giftmessage added attribute to the entity. if disabled the module later, the attribute related entity view page would get error. 

This error is not obviously before, and if we need error, we can use another way to notify backend operator. but not stop them using the system with blank page.

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tabs.php
sidealice: > As suggested by @elidrissidev, we need to do a proper separation of concerns. `Mage_Customer` depends on the following modules:
> 
> https://github.com/OpenMage/magento-lts/blob/5318b8617612f07a2a10882c3c4eed262726eb3f/app/etc/modules/Mage_Customer.xml#L28-L32
> 
> All other modules here should be separated:
> 
> 1. `Mage_Sales`
> 2. `Mage_Catalog`
> 3. `Mage_Newsletter`
> 4. `Mage_Wishlist`
> 
> Besides the tabs, which can be easily added with the event `adminhtml_block_widget_tabs_html_before` (see PR #1358), the rendering of the pages need to be separated as well:
> 
> image
> 
> Related to PR #2455. It means also PR #2533. And down the rabbit hole. What should we do?

suggest we can first let the module sepreated into the repository.  later we can refactor one by one, I just afraid if we want a perfect solution and that may took too much time to completed.

app/etc/modules/Mage_Captcha.xml
elidrissidev: Like mentioned before, I suggest tackling these changes one by one for ease of review.

elidrissidev:
After looking more into this, IMO it's better to keep the refactoring until another major version, since it means moving classes around which breaks BC. For now, handling with a condition is good enough.

app/code/core/Mage/Adminhtml/Block/Sales/Order/View/Giftmessage.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
fballiano, line 2:
```suggestion
```

lib/Varien/Data/Form/Abstract.php
```diff
@@ -1,4 +1,5 @@
 <?php
+
```
fballiano, line 2:
```suggestion
```

fballiano:
this time it worked correctly, so I'm ok with this PR

empiricompany (approved):

ADDISON74 (approved):

Issue #3030, opened 2023-02-17
session_start() errors should be logged and/or handled


loekvangool:
The session functions `session_start(), `session_set_save_handler()`, `session_set_cookie_params()` are not checking for `headers_sent()` or other issues, and issues are not handled gracefully.

If you get errors like:

```
ERR (3): Warning: session_set_save_handler(): Cannot change save handler when headers already sent  in app/code/local/Cm/RedisSession/Model/Session.php on line 85
ERR (3): Warning: session_set_cookie_params(): Cannot change session cookie parameters when headers already sent  in app/code/core/Mage/Core/Model/Session/Abstract/Varien.php on line 124
ERR (3): Warning: session_start(): Cannot start session when headers already sent  in app/code/core/Mage/Core/Model/Session/Abstract/Varien.php on line 150
```

..it probably means output is starting somewhere where it should not (an error, a stray whitespace).

I'd argue that these errors should be logged, along with some useful metadata, like the URL. Also, we might consider letting the request through. An anonymous user/bot may not care about having a session? This is treachy terriroty of course...what if he adds something to his cart?

In any case, `session_start()` does not respond well to `try { } catch {}` so it'll have to be something else: https://stackoverflow.com/questions/16568262/php-session-start-general-error-handling

PR #3082, opened 2023-03-13
Log PHP session errors and prevent PHP errors

loekvangool:
Fixes https://github.com/OpenMage/magento-lts/issues/3030


fballiano:
I would rewrite the methods in a different way

```
if (!check) {
  error
  return
}

if (!check) {
  error
  return
}

do what you need to do
```

it seems much more readable to me.

fballiano:
@loekvangool can you rebase this PR on the branch "main", I can't do it on your repo

fballiano:
the app/code/core/Mage/Core/Model/Session/Abstract/Varien.php is now very different from the version you worked on so I think the whole thing has to be rethough

ADDISON74:
This PR is attached to 1.9.4.x branch and there are still about 25 in this situation.

We should do a cleaning and stay with 3 branches in OpenMage repository: v19, main and next. PRs that cannot be edited should be deleted and added again with the edit option. From my estimations around 20 PRs will close issues too, so that we reduce the number of reports below 100. Anyway we are doing well compared to Magento 2, where the number of PRs is around 1K and the reported issues is over 1.5K.

PS - We should rename v19 branch to **lts** because that is the goal, who wants to stay with Magento without BC's this is the place for a while.

fballiano:
@loekvangool could you rebase it on the "main" branch? cause it's different from the one you worked on and it doesn't really match anymore.

fballiano:
I'll close this at the moment cause I can't rebase it since the main branch is now a bit different from the point to where it was when this PR was created.

feel absolutely free to reopen when you're ready to pick it up :-)

PR #3086, opened 2023-03-14
Fix exception logged when no Category image is uploaded

elidrissidev:
### Description (*)
As @luigifab pointed out in https://github.com/OpenMage/magento-lts/discussions/3025#discussioncomment-5196210, updating a Category without an image keeps logging an exception "No file was uploaded" in `exception.log`.

This error is normally being ignored by checking `$e->getCode()` because the Category image is optional, but the new error handling introduced in #2902 isn't passing the error code with the `Exception` object making the exception be logged.

### Related Pull Requests

- See #2902

1. Make sure logs are enabled.
2. Go to Catalog -> Categories and try to save any category, then check `exception.log` file.

### DIFF
--- a/lib/Varien/File/Uploader.php
+++ b/lib/Varien/File/Uploader.php
@@ -146,7 +146,11 @@ class Varien_File_Uploader
 
     public const SINGLE_STYLE = 0;
     public const MULTIPLE_STYLE = 1;
-    public const TMP_NAME_EMPTY = 666;
+
+    /**
+     * @deprecated Use UPLOAD_ERR_NO_FILE instead
+     */
+    public const TMP_NAME_EMPTY = UPLOAD_ERR_NO_FILE;
 
     /**
      * Resulting of uploaded file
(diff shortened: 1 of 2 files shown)


lib/Varien/File/Uploader.php
```diff
@@ -161,10 +161,10 @@ public function __construct($fileId)
         $this->_setUploadFileId($fileId);
         if (empty($this->_file['tmp_name']) || !file_exists($this->_file['tmp_name'])) {
             $errorCode = $this->_file['error'] ?? 0;
+            $code = empty($this->_file['tmp_name']) ? self::TMP_NAME_EMPTY : 0;
             if ($errorCode && isset(self::UPLOAD_ERRORS[$errorCode])) {
-                throw new Exception(self::UPLOAD_ERRORS[$errorCode]);
+                throw new Exception(self::UPLOAD_ERRORS[$errorCode], $code);
             }
-            $code = empty($this->_file['tmp_name']) ? self::TMP_NAME_EMPTY : 0;
             throw new Exception('File was not uploaded.', $code);
```
elidrissidev, line 168: To me, the message "File was not uploaded" is similar to `UPLOAD_ERR_NO_FILE => 'No file was uploaded',`, I wonder if it'd be better to just merge them together instead of throwing different Exceptions in two different cases 🤔, but not sure if any 3rd party code could be relying on that...

elidrissidev:
Got rid of the devil number ;), now `TMP_NAME_EMPTY` is deprecated and replaced with `UPLOAD_ERR_NO_FILE` so not breaking BC as long as extensions are using the constant.

fballiano (approved):

ADDISON74 (approved):

fballiano:
merged and 20ed

PR #3087, opened 2023-03-14
Added possibility to set backend locale per every admin user

fballiano:
This PR adds the possibility to set a `locale` for the `backend` for every `backend user`, adding a column to the `admin_user` table and the UI components to use this new column.

This is the new field added to the "System -> My Account" form:

Screenshot 2023-03-14 alle 22 39 50

1. login to the backend
2. click "system -> my account"
3. set the backend locale using the new field
4. save
5. logout
6. login to the backend again
7. the locale of the backend should be the one selected


fballiano:
PHPStan fails for unrelated issues with the 20.0 branch

luigifab:
No tested, but is there the possibility to leave the field empty?

fballiano:
@luigifab there is

app/code/core/Mage/Adminhtml/controllers/System/AccountController.php
```diff
@@ -67,6 +65,10 @@ public function saveAction()
             $user->setPasswordConfirmation($this->getRequest()->getParam('password_confirmation', false));
         }
 
+        $backend_locale = $this->getRequest()->getParam('backend_locale', false);
+        $backend_locale = $backend_locale == 0 ? null : $backend_locale;
+        $user->setBackendLocale($backend_locale);
```
luigifab, line 70: What is this horrible variable name? `$backend_locale` => `$backendLocale` no?

fballiano: ehmm in other PRs I had to switch to camelcase (which I never liked) and now the otherway around?

example:
Screenshot 2023-03-16 alle 20 47 00

fballiano: or:
Screenshot 2023-03-16 alle 20 47 43

fballiano: I'm doing too many things at the same time, apologies @luigifab 

fixing it

fballiano, line 70:
```suggestion
        $backendLocale = $this->getRequest()->getParam('backend_locale', false);
        $backendLocale = $backendLocale == 0 ? null : $backendLocale;
        $user->setBackendLocale($backendLocale);
```

luigifab: ... I vote for `$backendLocale` like `$orderItem`.

fballiano: it has to be camelcase since it's the standard, although it's not my cup of tea.

it was already used in this PR, I just had the wrong one in one of the files

elidrissidev:
Have you thought about saving the locale in the `extra` field instead of a new column?

fballiano:
@elidrissidev ah! noup, didn't even notice it existed. It's interesting. I guess it makes sense.

Is "extra" supposed to be used by the "system" or by normal developers? I ask this because I'm wondering if who manipulates the data in that field knows that some more stuff can be added by the core and that they should act accordingly.

kiatng:
The `extra` field is to save the state in the System Configuration: 
https://github.com/OpenMage/magento-lts/blob/24176a8b5098451fa10e30b18c38b33aad2e5310/app/code/core/Mage/Adminhtml/controllers/System/ConfigController.php#L305-L329

So, if you use it, it will be overwritten on lines 322-325. We need a new column.

fballiano:
@kiatng thank you!

elidrissidev:
I know it's used for user config state, but it's put inside array key `configState` to allow storing additional data in there. Anyways, adding a new column doesn't hurt :)

FredericMartinez (approved):

m-overlund (approved):
Tested without issues

luigifab (approved):

matteotestoni (approved):

fballiano:
I'm merging cause we have 1 green check and 1 gray check but 2 more gray checks were dismissed only cause of the branch change, without code change (and the branch is still v20 at the end of the day)

PR #3088, opened 2023-03-14
Integrated autoload optimization in the composer.json file

fballiano:
In https://github.com/OpenMage/magento-lts/pull/3081 @elidrissidev introduced an optimization of the autoloader, which is great, but it only affects the zip file that we release.

With this PR, enabling the optimization in the composer.json file, we should be able to enable it for everybody and all the workflows (probably speeding some of them a little bit).

What do you think about this idea?

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/3081

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,9 +41,6 @@ jobs:
     - name: Composer install
       run: composer install --prefer-dist --no-progress --ignore-platform-req=ext-* --no-dev
 
-    - name: Optimize autoloader
-      run: composer dump-autoload --optimize
-
     - name: Create ZIP file
       run: zip -rq openmage-${{ github.event.release.tag_name }}.zip . -x '.git/*'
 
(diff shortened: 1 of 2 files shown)


elidrissidev:
The reason why I didn't include it in `composer.json` is because that will make it always optimize the autoloader, and the Composer documentation says:

> Note: You should not enable any of these optimizations in development as they all will cause various problems when adding/removing classes. The performance gains are not worth the trouble in a development setting.

fballiano:
but people that are extracting the zip on top of their projects are also developing and probably adding/removing classes.

elidrissidev:
I don't know honestly, if you think we should just recommend it in README but remove it from the release workflow then go ahead.

fballiano:
I don't know but it seems to me the use cases are the same so if we can't enable it by default... I'm not sure it should be in the released zipball.

PR #3091, opened 2023-03-15
PHP 7.4 is now the minimum required version

fballiano:
Let's see if we get enough vote to move to PHP 7.4, as it's [being discussed here](https://github.com/OpenMage/magento-lts/discussions/3089).

I think that, since very long time we now about to release a minor, it would be the right time to move to php7.4.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.3']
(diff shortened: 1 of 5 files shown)


fballiano:
I'm not really sure about the workflow (it seems more changes should be required) but let's see.

colinmollenhour (approved):

tmewes (approved):

elidrissidev (approved):
Bump that number up!

fballiano:
yeah, thanks!

Issue #3090, opened 2023-03-15
Google Analytics 4 / GA4 does not track conversions (orders)


loekvangool:
https://github.com/OpenMage/magento-lts/pull/3023 implements the basic GA4 snippet from Google, but it's probably not ready for production.

I haven't run it, but by looking at the code, this PR does not implement order (conversion) tracking (`_getOrdersTrackingCode()`). Without conversion tracking, I would say that GA4 is not "ready". I think this is the right event to track: https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#purchase

Also, I cannot help notice that this PR was approved without proper testing, or at least if there was testing, it was not documented in the Issue (see discussion in https://github.com/OpenMage/magento-lts/pull/3023).

In order to implement conversion tracking, something like this is needed: https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/GoogleAnalytics/Block/Ga.php#L206

Note: It seems from GA4 docs that tracking pageviews is not necessary, hence in theory `_getPageTrackingCode()` does not need to be implemented for GA4.

fballiano:
it never tracked conversions if I remember correctly, that's something for modules like Enhanced Commerce etc.

The PR tracks page visits, convertions and events depends too much on the business logic of the website and anyway it's not to be considered a "bug" of the current implementation.

loekvangool:
It's a serious degradation of functionality. Without conversions, Google smart campaigns fail. At the very least it should be documented.

ADDISON74:
I confirm what @fballiano said. It was just an injection into the HTML source code of the GA script. What is reported here is a feature request. There are extensions which implemented track conversions. PR's are welcome.

loekvangool:
> the previous implementation (which is anyway still there) didn't track conversions either AFAIK so I don't understand the degradation...

I don't think so. I can confirm that orders are tracked with the pre-GA4 implementation. The code was already mentioned in my original post but it's here: https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/GoogleAnalytics/Block/Ga.php#L206

fballiano:
ok, it should be quite easy to add

loekvangool:
Aprreciate it, I will find time for testing

PR #3092, opened 2023-03-15
Added order tracking for Google Analytics 4

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3090 it was pointed out that the previous UA/GA implementation supported the order data tracking, while the GA4 implementation isn't.

This PR aims to add the order tracking to GA4.

1. Fixes https://github.com/OpenMage/magento-lts/issues/3090

1. Enable Analytics 4 in the backend configuration
2. Place an order
3. Check that there are no javascript errors and that the order data is sent to analytics


sreichel:
Possible to add tax and category information?

fballiano:
I've added the tax amount but I wouldn't add the categories, the trees can be too different from store to store (very deep, with categories with the same names across branches etc) so IMHO we'd some configuration options and multiple implementations for that.

I'll copy paste some code I wrote as a test:
```
                $categories = [];
                /** @var Mage_Catalog_Model_Category $category */
                foreach ($product->getCategoryCollection() as $category) {
                    $categories[] = $category->getId();
                    foreach ($category->getParentCategories() as $parentCategory) {
                        $categories[] = $category->getId();
                    }
                }
                $categories = array_unique($categories);
                $categoryCollection = Mage::getResourceModel('catalog/category_collection')
                    ->addAttributeToSelect('name')
                    ->addAttributeToFilter('is_active', 1)
                    ->addAttributeToFilter('entity_id', array('in' => $categories));
                $categories = [];
                foreach ($categoryCollection as $category) {
                    $categories[] = $category->getName();
                }
                $categories = array_slice(array_values(array_unique($categories)), 0, 5);
                foreach ($categories as $i => $category) {
                    $i += 1;
                    $i = $i == 1 ? '' : $i;
                    $orderData['item_category' . $i] = $category;
                }
```
but I'm not committing it, I think we shouldn't have the category for the time being.

fballiano:
@loekvangool could you test this?

fballiano:
@loekvangool?

app/code/core/Mage/GoogleAnalytics/Block/Ga.php
```diff
@@ -196,6 +198,46 @@ protected function _getOrdersTrackingCodeUniversal()
         return implode("\n", $result);
     }
 
+    /**
+     * @return string|void
```
shyim, line 202:
```suggestion
     * @return string
```

as you do an echo in the template I would always return a string.

app/code/core/Mage/GoogleAnalytics/Block/Ga.php
```diff
@@ -196,6 +198,46 @@ protected function _getOrdersTrackingCodeUniversal()
...
+            foreach ($order->getAllVisibleItems() as $item) {
+                $orderData['items'][] = [
+                    'item_id' => $item->getSku(),
+                    'item_name' => $item->getName(),
+                    'quantity' => $item->getQtyOrdered(),
+                    'price' => $item->getBasePrice(),
+                    'discount' => $item->getBaseDiscountAmount()
+                ];
+            }
+            $result[] = "gtag('event', 'purchase', " . json_encode($orderData) . ");";
```
shyim, line 236:
```suggestion
            $result[] = "gtag('event', 'purchase', " . json_encode($orderData, JSON_THROW_ON_ERROR) . ");";
```

empiricompany (approved):

fballiano:
This PR is the reason why I won't even take a look at https://github.com/OpenMage/magento-lts/issues/3121, you develop something that somebody asked, you update it based on someone else's feedback and then nobody cares taking a look at it. neither the person who asked for it nor the other one that asked for modifications...

What a great waste of time.

ADDISON74:
It's not a waste of time, unfortunately we lack people to allocate a little of their daily time for this project. I wouldn't push things, because there are still a few of us who are moving something around here. I also felt a frustration last year when for a while I didn't see any approved PR.

sreichel:
> nor the other one that asked for modifications...

Me? I gave thumbs up ... ?!?

app/code/core/Mage/GoogleAnalytics/Block/Ga.php
fballiano: true! thank you!

empiricompany (dismissed):

matteotestoni:
GA3 stop july. Is possibile approve this before july?

fballiano:
@matteotestoni if you review it (clickin on the "files changed" and then "review" button) you can make the difference, with 3 reviews with non-admins we can still merge!

matteotestoni (dismissed):

matteotestoni:
> @matteotestoni if you review it (clicking on the "files changed" and then "review" button) you can make the difference, with 3 reviews with non-admins we can still merge!

done

fballiano:
I've to rebase it on the main branch 🤞

matteotestoni (approved):

empiricompany (approved):

ADDISON74:
ga

Based on the image above I would like to discuss a few questions

- When there will be only the "Google Analytics 4" option, do we still need the other options in the **Type** list?

- Has anyone used the "IP anonymization" facility so far?

Testing this PR, the script is found in the page source code. Unfortunately I can't make the changes in a production server that I already manage to check what is displayed in the GA interface.

```html
            <!-- BEGIN GOOGLE ANALYTICS 4 CODE -->
        <script async src="https://www.googletagmanager.com/gtag/js?id=ADDISON74"></script>
        <script>
            window.dataLayer = window.dataLayer || [];
            function gtag(){dataLayer.push(arguments);}
            gtag('js', new Date());
            gtag('config', 'ADDISON74');
                    </script>
        <!-- END GOOGLE ANALYTICS 4 CODE -->
```

fballiano:
in the future i'd still leave the select, for an eventual GA5 or whatever.

the IP anonymization I've used it in the past. it's not necessary for GA4 since it's the default and only way the data is sent.

ADDISON74 (approved):
I approved it because this change must be made ASAP. I appreciate that Google Analytics 4 is the first option in the list. If there will be any issue, I have no doubt that they will not be reported and solved.

Issue #3066, opened 2023-03-06
Associated product in admin not visible


matteotestoni:
### Preconditions (*)

1. OpenMage 19.5.0 rc1
2.

### Steps to reproduce (*)

1. open a configurable product with associated product in backend
2.

### Expected result (*)

1. ajax action and load associated product
2.

### Actual result (*)

1. ajax loading but admin empty
2.

ADDISON74:
What do you mean by "admin empty"? Most likely you wanted to say that there are no associated products in the grid. Basically, although the configurable product has associated products they are not displayed.

I tried to reproduce this issue with the latest OpenMage 19 version cloned from this repository and Magento Sample Pack. All configurable products have associated products. Then I created a configurable product and attached simple products, again the editing of the product displays the associated products in the grid.

Are you using OpenMage V19.5 installed from scratch or a custom version with extensions?

matteotestoni:
![image](https://user-images.githubusercontent.com/34275189/223162115-bdd061d8-f836-4d62-bf8d-9a1cbf65a134.png)

matteotestoni:
On inserting product all ok. But when i edit there is an error. I am debugging but by now i cannot understand.
I migrated to openmage 20.1 rc1

matteotestoni:
as soon i enter in editing i have this log
Mage_Eav_Exception: Invalid attribute requested: simple_product_color in /var/www/vhosts/artigrafichemacaluso.it/httpdocs/app/Mage.php:645
but in eav_attribute simple_product_color is not present

fballiano:
I think it's already solved in https://github.com/OpenMage/magento-lts/pull/3038

matteotestoni:
don't work. redis flushed, cache disabled.
Maybe, where to search for  Invalid attribute requested: simple_product_color? in database? in files?

ADDISON74:
I confirm this bug with an earlier version of OpenMage v19. I didn't spend time to debug and I installed the last version the other days. The issues is gone.

fballiano:
The problem is here https://github.com/OpenMage/magento-lts/commit/bad624456add751005ebe63a295e659df16c9b3e#r104605921 

Let's wait for @luigifab to see if he has an idea on how to fix it

luigifab:
For me I haven't any problem with _Associated products_ tab of configurable products:
[...]

PR #2715, opened 2022-11-16
Keep current selection for Related, Up-sells, Cross-sells, Associated grids

luigifab:
### Description

Go to Catalog > Products and choose any product.
For Related, Up-sells, Cross-sells, Associated (for grouped products) grids.

- Uncheck all products, then reset filter or filter or sort or change the page, after AJAX all products are unchecked.
- Check some products, then reset filter or filter or sort or change the page, after AJAX selected products are checked.

Try Sales / Order grid, this is the same thing.
This is an extract of #1379.

OpenMage 20.0.17 / PHP 8.0.25

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #3093, opened 2023-03-15
Fixed problem with associated_products not visible in backend

fballiano:
If you create a grouped product, when you associate some products to the grouped and save, they won't appear in the "associated products" tab.

This is because https://github.com/OpenMage/magento-lts/pull/2715 changed the lines that are reverted in this PR, switching the default value from `null` to `[]` and causing a [series of other checks to fail](https://github.com/OpenMage/magento-lts/issues/3066#issuecomment-1470956037).

With this PR the funcionality is working again.

### Related Pull Requests
1. https://github.com/OpenMage/magento-lts/pull/2715

1. Fixes https://github.com/OpenMage/magento-lts/issues/3066

1. create a grouped product
2. associate some other products
3. save
4. associated products are not shown (before the PR)

### DIFF
--- a/app/code/core/Mage/Adminhtml/controllers/Catalog/ProductController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/Catalog/ProductController.php
@@ -405,7 +405,7 @@ class Mage_Adminhtml_Catalog_ProductController extends Mage_Adminhtml_Controller
         $this->_initProduct();
         $this->loadLayout();
         $this->getLayout()->getBlock('catalog.product.edit.tab.related')
-            ->setProductsRelated($this->getRequest()->getPost('products_related', []));
+            ->setProductsRelated($this->getRequest()->getPost('products_related', null));
         $this->renderLayout();
     }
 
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

kiatng:
Doesn't this "unfix" the bug in issue #2715? How about only change this `->setProductsGrouped($this->getRequest()->getPost('products_grouped', null));` for grouped products?

fballiano:
I didn't notice that this PR it's actually a complete revert of https://github.com/OpenMage/magento-lts/pull/2715 

uff

the thing is we can't have the grouped products not working at all.

fballiano:
@kiatng I've reverted everything except the groupedProducts line, which make grouped products work for me again

luigifab:
There is a strange thing, without this, if you press reset filter, then all is working fine.

fballiano:
> There is a strange thing, without this, if you press reset filter, then all is working fine.

@luigifab the "reset filter" button calls the "superGroupGridOnly" method which wasn't affected by https://github.com/OpenMage/magento-lts/pull/2715, that's why it still works :-)

kiatng (approved):

kiatng (dismissed):

fballiano:
without this one we can't close 19.5/20.1, grouped products are literally broken

fballiano:
merged and 20ed

PR #3098, opened 2023-03-19
Removed module Mage_Poll

fballiano:
From my experience the Mage_Poll module was almost never used by anybody, and in 2023 I think it really is unnecessary in our core, this PR removes it.

If you agree with this PR I'll create a separate repository for the module (as we did for Mage_Backup), so that users that still want can install it.

**NOTE**: PHPStan errors are not related to this PR, they appear only because this PR targets branch 20.0 and PHPStan's baseline is broken on branch 20.0.


ADDISON74:
I often encountered the situation when stores wanted to find out the opinion of visitors. Here comes this module to offer a solution. For example, in a store, the delivery company chosen to ship the orders created big problems and the complaints multiplied. The shop owner made this aspect public through a message accompanied by a poll and asked for the vote. Following the result, he had discussions with the forwarder and gave up his services. Today, very few stores keep in touch with customers in a real way about problems. Most have the impression that if they have better prices, discount coupons, the business is milk and honey.

I have nothing against removing the source code from the distribution as long as the respective modules are kept in a repository.

fballiano:
let's say nowadays there are other tools like chat or surveys connected to the newsletters

justinbeaty:
I definitely don't care what my customer's favorite color is, let's remove it!

In general I'd support the removal of any module that isn't essentially core functionality, moving that code to a separate repo, and optionally including it in the release builder.

justinbeaty (approved):

fballiano:
@justinbeaty I've targeted only v20, since I hope we'll approve the new release schedule/branches I don't think it's worth doing it in v19

justinbeaty:
@fballiano agreed. I think the last v19.5.0 release is good for v19 for getting Zend etc out of the repo, but I wouldn't want to spend much time on v19 either.

elidrissidev (approved):

tmewes (approved):

fballiano:
Merging as per https://github.com/OpenMage/rfcs/blob/main/accepted/0001-change-pr-merging-rules.md

ADDISON74:
I have nothing against removing this module. I find that the opinions are the majority in this regard. Please just put the module files in a repository for those who want to use it in the future. Basically it is a module that works and has not generated any issues according to reports so far.

fballiano:
creating the repo for Mage_Poll now,  actually I'll have to update the README here too.

fballiano:
Published as https://packagist.org/packages/openmage/module-mage-poll

kiatng:
I am not against this PR. However, I think for a PR that remove more than a few files should wait for at least 3 days before merging. So that community members living in other time zones or not-working can have a chance to voice an opinion.

sreichel:
+1 for removal. -1 for v20-only :(

PR #3102, opened 2023-03-19
Deprecation errors are not suppressed anymore

fballiano:
Since we're moving forward with PHP versions and people will need to upgrade their stack sooner or later I think it's time to avoid suppressing deprecation errors.

Without the special treatment that you see in app/code/core/Mage/Core/functions.php all deprecation errors will be logged, and if DEVELOPER_MODE is enabled then they will trigger an error.

Why I did this? Cause it's better that people start to see this logs and fix their deprecation errors because they will become errors with new versions of PHP.

Also this PR removes the need for the specific DEV_PHP_STRICT that way introduced lately.

1. test with php 8.1
2. edit a header.phtml and add a `trim();` (which triggers a deprecation error)
3. refresh, now there's an error logged in system.log
4. enable developer mode
5. refresh, now the deprecation error is a full error shown on page


kiatng (approved):

ADDISON74 (approved):

fballiano:
merged and 20ed

PR #3108, opened 2023-03-20
Disabled and disallowed SWF file extension

fballiano:
In the wysiwyg editor there's still some SWF related code, which has to go ;-)

### DIFF
--- a/app/code/core/Mage/Cms/Model/Wysiwyg/Config.php
+++ b/app/code/core/Mage/Cms/Model/Wysiwyg/Config.php
@@ -92,7 +92,7 @@ class Mage_Cms_Model_Wysiwyg_Config extends Varien_Object
                 Mage::getBaseUrl('js') . 'mage/adminhtml/wysiwyg/tiny_mce/themes/advanced/skins/default/content.css',
             'width'                         => '100%',
             'plugins'                       => [],
-            'media_disable_flash'           => Mage::helper('cms')->isSwfDisabled()
+            'media_disable_flash'           => true
         ]);
 
         $config->setData('directives_url_quoted', preg_quote($config->getData('directives_url')));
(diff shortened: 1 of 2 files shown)


shyim (approved):

FredericMartinez (approved):

m-overlund (approved):

app/code/core/Mage/Cms/Model/Wysiwyg/Config.php
```diff
@@ -92,7 +92,7 @@ public function getConfig($data = [])
                 Mage::getBaseUrl('js') . 'mage/adminhtml/wysiwyg/tiny_mce/themes/advanced/skins/default/content.css',
             'width'                         => '100%',
             'plugins'                       => [],
-            'media_disable_flash'           => Mage::helper('cms')->isSwfDisabled()
+            'media_disable_flash'           => true
```
ADDISON74, line 95: Wouldn't it be more appropriate to keep the initial form since the function returns only the true value?

fballiano: that method will be removed in the future, that's why I removed the call to it, it will be easier to identify it as unused ;-)

sreichel: @fballiano maybe i deleted it b/c i changed my mind or dont want to start a discussion about it. Please ignore it.

fballiano: @sreichel it is not the fist time you do it and you keep adding emojis to comments "against" my contributes. as I said, after the amount of words I spent defending you this behaviour is extremely disappointing. and childish.

sreichel: @fballiano there is nothing "against" you! thumbs down means i disagree with it, nothing else. When i deleted a commet, just not to start a discussion - move on. Pls dont take it personal.

Issue #3112, opened 2023-03-22
cron does run twice becuase of wrong $isShellDisabled check


gfemorris:
### Preconditions (*)
OpenMage Version 19.4.23

### Steps to reproduce (*)
1. set up cron via cron.sh like this
```
/bin/sh cron.sh cron.php -mdefault
/bin/sh cron.sh cron.php -mdefault
```
2. $isShellDisabled is true because `!str_contains(strtolower(PHP_OS), 'win')` in cron.php in line 46 is always true. It needs to be `str_contains(strtolower(PHP_OS), 'win')`

3. 
```
 if ($isShellDisabled) {
        Mage::dispatchEvent('always');
        Mage::dispatchEvent('default');
    }
```
is executed and therefor always and default are executed twice.

### Actual result (*)
On our setup this lead to all schedules beeing duplicated in the cron_schedule table and therfor for example emails were sent twice.

elidrissidev:
Confirmed this bug, originating from #2380. This may have been caused by confusion over the fact that `PHP_OS` is set to `Darwin` in macOS, so with the negation operator it's actually resulting is `false` which is what you'd expect in a non-Win*** system.

sreichel:
@all-contributors add @gfemorris bug

PR #3113, opened 2023-03-22
Fix PHP_OS check in cron.php

elidrissidev:
### Description (*)

Some improvements were done previously to `cron.php` in #2380, including a refactor to `$isShellDisabled` which decides whether or not to use the shell generate the cron schedules and run them.

The problem comes from the condition `!str_contains(strtolower(PHP_OS), 'win')`, which is the opposite of what it should be and it'll cause the cron events to be dispatched twice from the lines below:
https://github.com/OpenMage/magento-lts/blob/e2d5fed660af577477936eda18ab207789856cfb/cron.php#L74-L79

This was likely caused by confusion over the fact that PHP_OS is set to Darwin in macOS (@fballiano's OS), so with the negation operator it's actually resulting in false which is what you'd expect in a non-win system.

### Related Pull Requests

- See https://github.com/OpenMage/magento-lts/pull/2380

1. Fixes OpenMage/magento-lts#3112

1. Add a `var_dump($isShellDisabled)` to `cron.php` before the `try catch` block, and observe the value before and after.

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -42,14 +42,15 @@ try {
 umask(0);
 
 $disabledFuncs = array_map('trim', preg_split("/,|\s+/", strtolower(ini_get('disable_functions'))));
+$isWinOS = !str_contains(strtolower(PHP_OS), 'darwin') && str_contains(strtolower(PHP_OS), 'win');
 $isShellDisabled = in_array('shell_exec', $disabledFuncs)
-    || !str_contains(strtolower(PHP_OS), 'win')
+    || $isWinOS
     || !shell_exec('which expr 2>/dev/null')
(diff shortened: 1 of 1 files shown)


fballiano (approved):
darwindows 😅 what the heck

the problem was only on windows anyway right?

elidrissidev:
> darwindows 😅 what the heck
> 
> the problem was only on windows anyway right?

ikr, it was windows all along!

ADDISON74 (approved):

fballiano:
merged and 20ed.

PS: I'm trying to change the titles because they're used to generate the release notes...

PR #3123, opened 2023-03-29
[BC] Removed double span element

fballiano:
The OM base template has a very old HTML structure, with a lot of (now) useless wrappers that (at the creation time) were necessary because CSS was not so powerful.

Now the situation is very different, CSS is very powerful, all those divs and spans are no longer necessary but not only, they're actually a problem because they make the DOM very big and google hates a big DOM, so it's our duty to provide an HTML that's basic (as in the "base" theme) and not bloader.

This PR removed the double span elements that are present in every button, they were used to center the text in the button but they're now totally unnecessary.

## How to test

- check out this PR
- be sure you have the rwd theme selected
- navigate the openmage installation and see everything works

## Related PRs
- https://github.com/OpenMage/magento-lts/pull/2501

## Questions

Why are we checking rwd if we modified base? because base doesn't come with CSS and it's important that we don't brake rwd theme with this PR. This doesn't seem to happen.

Why don't we remove the double span from the rwd theme too? I guess it would break a lot more custom themes and it's not totally necessary, rwd it's our default theme and it could stay a little "opinionated" but those elements do not belong to the base theme.

Impacts on custom themes? This is hard to know... all themes that inherits from rwd should be fine but if they inherit directly from base... they will probably have some problems. What do we do with that?

It should be considered a breaking change and reason for a major release.


splendidinternet (dismissed):
Good idea, looks good, did not test it on a test instance though.

matteotestoni (dismissed):

fballiano:
rebased on `next`

ADDISON74 (approved):

matteotestoni (approved):

ADDISON74:
This PR should also be included in OpenMage 20.

fballiano:
mmm we can't do it, it will break something for custom themes, it is a breaking change and, at the moment, has to stay on next

matteotestoni (approved):

fballiano:
@empiricompany sure, that's why it's in the "next" branch for v21.0, the thing is google punishes for having too many dom elements, we have to simplify the DOM yes or yes and having 3 dom elements for every button, uselessly, should be avoided.

empiricompany (approved):

fballiano:
merging because of 1green check + 2 gray checks (actually would be 3 gray)

ADDISON74:
Considering that several PRs are only for the next branch, I created a label named "for previous versions" so that those who want to integrate in v19 or v20 can find them immediately and create their own patches.

As far as I am concerned, if this PR is not merged into v20 I will use it on my own.

fballiano:
I'm using "next" for changes that could have an impact, just to avoid problems you know :-)

PR #3124, opened 2023-03-30
Ensure $price and $rate are both numeric types

m-overlund:
Ensures price and rate are of type floatval (and not string).

Fixes this TypeError
"TypeError: Unsupported operand types: string * int"

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Directory/Model/Currency.php
+++ b/app/code/core/Mage/Directory/Model/Currency.php
@@ -183,10 +183,10 @@ class Mage_Directory_Model_Currency extends Mage_Core_Model_Abstract
         } else {
             $rate = $this->getRate($toCurrency);
             if ($rate) {
-                 // Ensure $price and $rate are both numeric types
-                 $price = floatval($price);
-                 $rate = floatval($rate);
-                 return $price * $rate;
+                // Ensure $price and $rate are both numeric types
+                $price = floatval($price);
+                $rate = floatval($rate);
+                return $price * $rate;
             }
         }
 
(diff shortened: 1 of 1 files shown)


fballiano:
how do you reproduce the error?

```php
    public function convert($price, $toCurrency = null)
    {
        $price = (float) $price;
        if (is_null($toCurrency)) {
            return $price;
        } else {
            $rate = (float) $this->getRate($toCurrency);
            if ($rate) {
                return $price * $rate;
            }
        }
```

but it still lacks a default return statement... the original method doesn't seem to be super well written...

m-overlund:
I'll revert the change and reproduce.
I think it's a custom code issue - due hiding the price on products out of stock / products we do not sell anymore.

Despite custom code, I don't like the missing "validation" of input.

fballiano:
yep sure, that's why I added the casting (changing from the floatval function because casting is more widely used in our code)

fballiano:
but that method expect floats and it's stated in the docbloc, I don't think it's possible to check every input of every method

joshua-bn:
If you're getting an error you're sending it the wrong data. PHP is already trying to cast it to a number to make the arithmetic work but it fails because of some other character(s) in there. You want it to break in that case.

elidrissidev:
Agree with @joshua-bn, better to get the full stack-trace and find out where the string is coming from.

sreichel:
> The best solution would be a type on the args.

Y, but does v20 "allow" changes method sigantures (adding types)?

fballiano:
@joshua-bn @elidrissidev are we closing this one? let's just not leave it hanging here forever.

m-overlund:
Maybe not 100% related but #3135 is also a case of typecasting / string/int issues.
decisions need to be made :)

m-overlund:
non-issue.
Casted string to Mage::helper('core')->currency()

Closing.

PR #3139, opened 2023-04-04
Updated README.md to reflect new RFC 0002 - Release Schedule

colinmollenhour:
Implements "Codify the strategy into the main project README.md file using the descriptions above" from [RFC 0002](https://github.com/OpenMage/rfcs/blob/main/accepted/0002-release-schedule.md#implementation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ In a nutshell:
 - MySQL 5.6+ (8.0+ recommended) or MariaDB
 - optional: Redis 5.x, 6.x and 7.0.x are supported
 
+
 - PHP extension `intl` <small>since 1.9.4.19 & 20.0.17</small>
 - Command `patch` 2.7+ (or `gpatch` on MacOS/HomeBrew) <small>since 1.9.5.0 & 20.1.0</small>
 
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ level of backwards compatibility to the official releases.
 
 ## Table of contents
 
+- [Releases and Versioning](#releases-and-versioning)
+  - [Currently Maintained Versions](#currently-maintained-versions)
 - [Requirements](#requirements)
 - [Installation](#installation)
   - [Manual Install](#manual-install)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,29 +30,48 @@ level of backwards compatibility to the official releases.
...
+- 20.x is the latest MAJOR version and will receive PATCH updates until 2 years after the date that 21.x is released.
+- 19.4.x will receive PATCH updates until April 4, 2025.
+
 ## Requirements
 
 - PHP 7.4+ (PHP 8.0 is supported, PHP 8.1 supported but some warnings may be shown/logged, PHP 8.2 is usable but still being tested)
 - MySQL 5.6+ (8.0+ recommended) or MariaDB
 - optional: Redis 5.x, 6.x and 7.0.x are supported
 
-
```
fballiano, line 53: this line was necessary to have a line break between the "real" software stack and other things, so maybe probably we could remove both of the line breaks :-)

README.md
```diff
@@ -20,6 +20,8 @@ level of backwards compatibility to the official releases.
 
 ## Table of contents
 
+- [Releases and Versioning](#releases-and-versioning)
```
fballiano, line 23: I wouldn't have probably put this point as the first one of the readme, other that that it's ok

README.md
```diff
@@ -20,6 +20,8 @@ level of backwards compatibility to the official releases.
...
+  - [Currently Maintained Versions](#currently-maintained-versions)
```
fballiano, line 24: If "releases and versioning" wasn't the first point, then I'd like this one to be the first one, since it's pretty important for everybody but also easy to read.

README.md
colinmollenhour: Yeah I struggled with organization.. Feel free to modify if you have a better organization but it didn't seem to fit anywhere in the middle and it is too important to be at the end..

README.md
colinmollenhour: Yes, it just lacks context without the Releases and Versioning paragraph, but open to modifications/suggestions. :)

README.md
fballiano: I'll think about it :-)

README.md
colinmollenhour: Thanks, I restored the line.

README.md
fballiano: at the end of the day I think the way you put it it's right

fballiano (approved):

ADDISON74 (approved):
I approved it because these clarifications must be made as soon as possible. Most likely there will be changes along the way, but we will solve them with other PRs.

fballiano:
merged and cherrypicked to v19 too

Issue #3127, opened 2023-03-31
Extreme memory usage when applying catalog rules


ioweb-gr:
### Preconditions (*)
1. OpenMage 20.0.10

### Steps to reproduce (*)
1. Create 11 catalog rules which apply discounts to products where category is one of 1,2,....N  and at least 10 clauses where manufacturer is 1..N to apply a discount percentage 
2. Use a medium sized catalog of about 40K products
3.  Try to apply all rules 

e.g. 

![image](https://user-images.githubusercontent.com/20220341/229061819-2100b106-f121-40f9-817f-86b298ca148c.png)

### Expected result (*)

1. Rules are applied

### Actual result (*)

1. Memory exhausted. In my case it required 16GB of RAM to actually execute the `\Mage_Adminhtml_Promo_CatalogController::applyRulesAction`

e.g. to simulate in a CLI script

```php
<?php
require_once('app/Mage.php');
umask(0);
Mage::app('admin');

ini_set('memory_limit', '1G');
Mage::getModel('catalogrule/rule')->applyAll();
``` 

Will yield 

`PHP Fatal error:  Allowed memory size of 1073741824 bytes exhausted (tried to allocate 75497472 bytes) in /var/www/vhosts/example.com/httpdocs/app/code/community/Varien/Profiler.php on line 169`

And it will crash until I put 16G in the limit.

ioweb-gr:
One additional problem in these conditions, if I added more than 15 manufacturers in the same clause ( the image has 5), the rule is not processed entirely.

ADDISON74:
Could you check this issue with different OpenMage versions? I would start first with Magento 1.9.4.5.

ioweb-gr:
I can tell you we've been carrying this problem since Magento 1.9.4.1 but it was manageable because it required approximately 2G-3G of memory. However as our needs have increased the past years we've reached the point where we reach 16G of memory and I felt it's at a point we need to debug this.

We also had some workarounds of external scripts running the rules one by one and this suited us until now as well. 

I could try later versions probably  if that would help though

elidrissidev:
@ioweb-gr Could you test #3140 and see if it fixes your issue?

rvelhote:
@ioweb-gr We don't have this problem and we have a catalog about the same size as yours, lots of promotions going on all the time and more complex rules than the ones you showed.

[...]

PR #3140, opened 2023-04-04
Avoided unnecessary calls to string translation methods for catalog/sales rules labels

luigifab:
### Description

This PR stop translations of rules labels from frontend.
Perhaps the best way is to remove all translations from model, and translate only from blocks? I don't know.

Here a before/after with Blackfire with a cart with ~50 products (simple, configurable, bundles):

original (before):
big-cart-before

after without the static property:
big-cart-after-effect

after:
static
big-cart-after

Can fix partially #3127 or not.
Tested with PHP 8.0 and 8.2 with OM rc2.

If you have others modules or custom dev, that do the same, you need to update your files to have better performance on your website. Example with [HiPay module's](https://github.com/hipay/hipay-fullservice-sdk-magento1/blob/master/src/app/code/community/Allopass/Hipay/Model/Rule/Condition/Product/Subselect.php).

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/CatalogRule/Model/Rule/Action/Product.php
+++ b/app/code/core/Mage/CatalogRule/Model/Rule/Action/Product.php
@@ -42,10 +42,10 @@ class Mage_CatalogRule_Model_Rule_Action_Product extends Mage_Rule_Model_Action_
     public function loadOperatorOptions()
     {
         $this->setOperatorOption([
-            'to_fixed' => Mage::helper('cataloginventory')->__('To Fixed Value'),
-            'to_percent' => Mage::helper('cataloginventory')->__('To Percentage'),
-            'by_fixed' => Mage::helper('cataloginventory')->__('By Fixed value'),
-            'by_percent' => Mage::helper('cataloginventory')->__('By Percentage'),
+            'to_fixed'   => static::$translate ? Mage::helper('cataloginventory')->__('To Fixed Value') : 'To Fixed Value',
+            'to_percent' => static::$translate ? Mage::helper('cataloginventory')->__('To Percentage') : 'To Percentage',
+            'by_fixed'   => static::$translate ? Mage::helper('cataloginventory')->__('By Fixed value') : 'By Fixed value',
+            'by_percent' => static::$translate ? Mage::helper('cataloginventory')->__('By Percentage') : 'By Percentage',
         ]);
         return $this;
     }
--- a/app/code/core/Mage/Rule/Model/Action/Abstract.php
+++ b/app/code/core/Mage/Rule/Model/Action/Abstract.php
@@ -36,8 +36,20 @@
  */
 abstract class Mage_Rule_Model_Action_Abstract extends Varien_Object implements Mage_Rule_Model_Action_Interface
 {
+    /**
+     * Flag to enable translation for loadOperatorOptions/loadValueOptions/loadAggregatorOptions/getDefaultOperatorOptions
+     * It's useless to translate these data on frontend
+     *
+     * @var bool
+     */
+    protected static $translate;
+
     public function __construct()
     {
+        if (!is_bool(static::$translate)) {
+            static::$translate = Mage::app()->getStore()->isAdmin();
+        }
+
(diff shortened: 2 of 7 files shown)


fballiano:
wow, I wasn't expecting the `__` to generate all those calls, I thought it would immediately hit a cache :\

fballiano (approved):

joshua-bn:
Does it make any difference for time?

fballiano:
if I read the screenshots correctly it goes from 5secs to 3.6secs (with @luigifab's DB data obviously)

app/code/core/Mage/CatalogRule/Model/Rule/Action/Product.php
```diff
@@ -42,10 +42,10 @@ public function loadAttributeOptions()
     public function loadOperatorOptions()
     {
         $this->setOperatorOption([
-            'to_fixed' => Mage::helper('cataloginventory')->__('To Fixed Value'),
-            'to_percent' => Mage::helper('cataloginventory')->__('To Percentage'),
-            'by_fixed' => Mage::helper('cataloginventory')->__('By Fixed value'),
-            'by_percent' => Mage::helper('cataloginventory')->__('By Percentage'),
+            'to_fixed'   => static::$translate ? Mage::helper('cataloginventory')->__('To Fixed Value') : 'To Fixed Value',
```
kiatng, line 45: What is `static::$translate`? I cannot find its reference, also:

> Undefined property '$translate'.intelephense(1014)

app/code/core/Mage/CatalogRule/Model/Rule/Action/Product.php
```diff
@@ -42,10 +42,10 @@ public function loadAttributeOptions()
...
+            'to_percent' => static::$translate ? Mage::helper('cataloginventory')->__('To Percentage') : 'To Percentage',
+            'by_fixed'   => static::$translate ? Mage::helper('cataloginventory')->__('By Fixed value') : 'By Fixed value',
+            'by_percent' => static::$translate ? Mage::helper('cataloginventory')->__('By Percentage') : 'By Percentage',
         ]);
```
kiatng, line 49: Not sure if this works:
```suggestion
        if (static::$isAdmin) {
            $this->setOperatorOption([
                'to_fixed' => Mage::helper('cataloginventory')->__('To Fixed Value'),
                'to_percent' => Mage::helper('cataloginventory')->__('To Percentage'),
                'by_fixed' => Mage::helper('cataloginventory')->__('By Fixed value'),
                'by_percent' => Mage::helper('cataloginventory')->__('By Percentage'),
            ]);
        } else {
            $this->setOperatorOption([]);
       }
```
`$isAdmin` is more meaningful as `$translate` is applicable in frontend.  It reduces the number of comparisons to just one.

app/code/core/Mage/Rule/Model/Action/Abstract.php
```diff
@@ -36,8 +36,20 @@
  */
 abstract class Mage_Rule_Model_Action_Abstract extends Varien_Object implements Mage_Rule_Model_Action_Interface
 {
+    /**
+     * Flag to enable translation for loadOperatorOptions/loadValueOptions/loadAggregatorOptions/getDefaultOperatorOptions
+     * It's useless to translate these data on frontend
+     *
+     * @var bool
+     */
+    protected static $translate;
```
kiatng, line 45:
```suggestion
    protected static $isAdmin;
```

kiatng:
I think it's better to stop the translation in here:
https://github.com/OpenMage/magento-lts/blob/6851e17f5f74f82241762b8f1c0779e6f0d54b51/app/code/core/Mage/Rule/Model/Action/Abstract.php#L39-L52

To something like:
```php
    public function __construct()
    {
        parent::__construct();
        if (!Mage::app()->getStore()->isAdmin()) {
            return;
        }
        $this->loadAttributeOptions()->loadOperatorOptions()->loadValueOptions();

        // ...
    }
```

ADDISON74 (approved):

PR #3141, opened 2023-04-04
Updated branches for codeql analysis workflow

fballiano:
This PR is very simple and the code should be self explanatory :-)


luigifab (commented):
I suppose this will work!

luigifab (approved):

ADDISON74 (approved):

PR #3146, opened 2023-04-05
[BC] Added form key validation to Contacts form

fballiano:
All credits for this PR go to @elidrissidev which created it originally in https://github.com/OpenMage/magento-lts/pull/2347. I think that PR should be merged but I made a slight modification, using the default "enable CSFR" instead of a new one.

This is BC because it requires all users to update their custom theme (if they customised their contact form template)

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/2347

### Fixed Issues (if relevant)
1. Fixes #1911

### Manual testing scenarios (*)

Explained in https://github.com/OpenMage/magento-lts/pull/2347

PR #3147, opened 2023-04-05
[BC ]Unified CSRF configuration

fballiano:
Following https://github.com/OpenMage/magento-lts/pull/3146 I'd like to unify all the CSRF configuration in a single one, the actual situation is a bit confusing with three different flags for the same thing (general one, newsletter, checkout).

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/3146

Fixes the problem that some user may enabled CSRF only for some of the modules and not for others.

### Testing
- enable `system/csrf/use_form_key`
- check that both the newsletter forms and checkout have the form_key added to their `<form` element (actually the checkout does everything with ajax request so you'll have to check that the form_key is added to those requests in the developer console) as shown in this screenshot:
Screenshot 2023-04-05 alle 15 34 42


app/code/core/Mage/Adminhtml/Block/Checkout/Formkey.php
```diff
@@ -30,7 +30,7 @@ class Mage_Adminhtml_Block_Checkout_Formkey extends Mage_Adminhtml_Block_Templat
      */
     public function canShow()
     {
-        return !Mage::getStoreConfigFlag('admin/security/validate_formkey_checkout');
+        return !Mage_Core_Controller_Front_Action::_isFormKeyEnabled();
```
elidrissidev, line 33: I would leave `isFormKeyEnabled` as-is and replace the config key with the constant `Mage_Core_Controller_Front_Action::XML_CSRF_USE_FLAG_CONFIG_PATH`.

fballiano: maybe we could have a middleground moving that logic to a helper method instead of Mage_Core_Controller_Front_Action?

fballiano: I believe we should have a public method to encourage (make life easier) for people implementing new forms to use the CSRF

elidrissidev: Sure, a core helper would be better.

fballiano: I've redone this part, what do you think about it?

matteotestoni (approved):

elidrissidev:
We can also remove `newsletter/security/enable_form_key` from "New Config Options" in README.

fballiano:
great catches, done, thank you!

fballiano:
heck! I searched and missed that, it's done now!

elidrissidev (dismissed):

app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -1001,4 +1001,12 @@ public function unEscapeCSVData($data)
         }
         return $data;
     }
+
+    /**
+     * @return bool
+     */
+    public function isFormKeyEnabled()
```
sreichel, line 1008:
```suggestion
    public function isFormKeyEnabled(): bool
```

Would use return/paremeter types for every new method.

app/code/core/Mage/Checkout/sql/checkout_setup/install-1.6.0.0.php
sreichel, line 1: Would it be better (for existing installs) to remove this entry with an update script?

ADDISON74: This is a good observation. What settings previously existed must be deleted. Obviously their existence does not affect functionality, but they are orphan records that should no longer be found in the database.

fballiano: It would be better but if for some reason somebody needs to rollback the software, the db configuration is gone…
So… pros and cons…

ADDISON74: If someone does a rollback then it is justified to have that recording, otherwise not. What would be the percentage of such a situation happening? If it's small, I am in favor of cleaning up behind, let's go ahead with the changes. Any other opinion is welcome.

sreichel: In case of a rollback `core_resource` should be reseted too. (a rollback would add the config value again)

fballiano: how could they regenerated the config value stored in the database?

i mean, not rollback as in db transaction, but somebody that wants to revert to the previous version.

app/code/core/Mage/Core/Helper/Data.php
fballiano: done, right, i've removed the docblock since it wouldn't have anything, not sure if that's the right way but I think so.

sreichel: I like doc blocks. Even if its redundant, its _easier_ to read ...

fballiano: mah, i'm not sure, anyway i've readded it

elidrissidev (dismissed):

matteotestoni:
how can i download openmage next?

fballiano:
@matteotestoni checkout the `next` branch then do a `composer install`, note that at the moment it's not in sync with all that was committed to `main`

ADDISON74:
We had a check for CSRD inherited from Magento in the process of placing the order. Then we came up with the following additions #1866 and #3146. I haven't analyzed this PR but I understand that everything I listed above is put in one place from now on, am I right?

What happens to the orphaned settings in core_config_data table? Maybe we should list them and everyone interested can delete them if they want.

fballiano:
> everything I listed above is put in one place from now on, am I right?

yep, if you enable "use CSRF" then it's used in every form, instead of having to enable it in 3 different places

> What happens to the orphaned settings in core_config_data table? Maybe we should list them and everyone interested can delete them if they want.

they stay were they are, but for config data we're talking about a few records, the gain in storage (if we remove them) is negligible but it totally breaks the possibility to roll back to an older version of openmage. so IMHO i'd not remove the unused configuration data.

ADDISON74:
The advantage of a unification is that if there are other forms that need it, it is no longer necessary to create a setting in the Backend configuration, but it is applied directly.

Can you make a list of values from the table that remain orphans? I'm going to delete them myself and I don't think I'll ever need them back.

elidrissidev:
> Can you make a list of values from the table that remain orphans? I'm going to delete them myself and I don't think I'll ever need them back.

- `newsletter/security/enable_form_key`
- `admin/security/validate_formkey_checkout`

ADDISON74:
I will test this PR and provide you a feedback ASAP.

app/code/core/Mage/Checkout/sql/checkout_setup/install-1.6.0.0.php
fballiano: New installs needs it. and old install dont need this change

sreichel: Why to keep an orphan db entry?

ADDISON74:
I am testing this PR. A **first comment**, I don't like that this value becomes hidden. Removing it from the Backend has both good and bad parts.

```XML_CSRF_USE_FLAG_CONFIG_PATH   = 'system/csrf/use_form_key'```

The **second comment** is that this unification should also be done in previous versions of next, at least in v20.

**Comment by ADDISON74** [comment]
Here is what appears in the **Security** section in the **next** branch



And here is what appears in the **Security** section in the branch of this PR. It is obvious that the "Enable Form Key Validation On Checkout" option has been removed

**Comment by ADDISON74** [comment]
**The third comment** is related to the purpose of this unification. It would be natural to see the modified option with the name "Enable From Key Validation" and under the label "Important! Enabling this option means that your custom templates used in checkout process, newsletter and contact contain form_key output. Otherwise the forms may not work".

**The fourth comment**, in the README.md file it is mandatory to mention which templates need to be modified.

**A fifth comment**, the unification proposed here is only for checkout and newsletter. Are we missing the contact form or maybe I am wrong?

**Comment by elidrissidev** [comment]
> **A fifth comment**, the unification proposed here is only for checkout and newsletter. Are we missing the contact form or maybe I am wrong? I am referring here to controlling all 3 forms with a single Enable/Disable option.

It was already done in #3146

**Comment by fballiano** [comment]
> And here is what appears in the **Security** section in the branch of this PR. It is obvious that the "Enable Form Key Validation On Checkout" option has been removed

sure, because it's been unified with the general one, so both the newletter and checkout configurations are removed, it is the purpose of the PR :-)

**Comment by fballiano** [comment]
> **The third comment** is related to the purpose of this unification. It would be natural to see the modified option with the name "Enable From Key Validation" and under the label "Important! Enabling this option means that your custom templates used in checkout process, newsletter and contact contain form_key output. Otherwise the forms may not work".

mmmm the point is that everybody has to do that, all of the plaform has CSRF on everything nowadays, we can't be the only one that treat this fundamental security thing as a "sorry, maybe you want to enable it but it's causing you trouble" thing, in my opinion at least

> **The fourth comment**, in the README.md file it is mandatory to mention which templates need to be modified.

fair enough, although for custom checkouts it's impossible to know

**Comment by ADDISON74** [comment]
@elidrissidev - What I commented on point 5 is not resolved in #3146. I explained in the message, the 3 forms should be controlled by an option in the Security section. It will be set by default to the value No (this means to revert PR #871) until the developer checks his custom templates, then to make the warning message disappear it will set the value to Yes. Only Newsletter and Checkout offered this control facility, not Contact.

@fballiano - even if this unification is in the next version, who will use this branch I doubt it will have an adapted theme. Therefore, he must adapt his theme. He will only know about this if the warning that was up until PR #871 is still displaye and to have a control of the 3 forms after making the modification, that is to confirm the form_keys feature works. I didn't find the dropdown in the Backend, that's why I left a comment. 

From the engineer's point of view, I have to limit myself to the worst user, if I create and offer a solution only for the smart ones, who pretend to be at my level of understanding or above, I will fail to sell my product on a large scale.

**Comment by elidrissidev** [comment]
> the 3 forms should be controlled by an option in the Security section.

They are already being controlled by the default config: `system/csrf/use_form_key`.

**Comment by fballiano** [comment]
> Here is what appears in the **Security** section in the **next** branch
> And here is what appears in the **Security** section in the branch of this PR. It is obvious that the "Enable Form Key Validation On Checkout" option has been removed

the setting is now only one and it's the original, which was always in:
Screenshot 2023-04-29 alle 17 58 20

and it's still available (tested this PR right now just to check)

**Comment by ADDISON74** [comment]
If the value from that option is used then we have to create a label bellow "Important! Enabling this option means that your custom templates used in checkout process, newsletter and contact contain form_key output. Otherwise the forms may not work".

Also if it is set to No we should have that warning message under the main menu.

**Comment by ADDISON74** [comment]
My comments based on @elidrissidev and @fballiano feedback

1. First observation, the word **protection** in "CSRF protection" must be written with a capital P. Thus we follow a rule regarding Backend texts where each word begins with a capital letter, except small texts 1 - 3 letters.

2. The important label must be inserted under the option "**Important! Enabling this option means that your custom templates used in checkout process, contact and newsletter contain form_key output. Otherwise the forms may not work**".

3. I disabled the option and the text beginning with Important is displayed at the top under the menu. It must be modified as text and **link** because it no longer concerns the Checkout form only, "**Important: Formkey validation on checkout process, contact and newsletter are disabled. This may expose security risks. We strongly recommend to enable CSRF Protection in [System / System].**"

formkey_message

If the option in the configuration **Advanced > System > CSRF protection** controls the 3 forms Contact, Newsletter, Checkout then **let's disable it** and do some tests.

4. **Contact Form** doesn't have the form_key, regardless of whether the option is enabled or not. I also checked the main branch and identified the same problem.

contact_form

5. **Newsletter Form** has the form_key even if the protection is disabled.

newsletter_form

6. **Checkout Form** has the form_key even if the protection is disabled.

checkout_form

**Comment by ADDISON74** [comment]
My comments based on @elidrissidev and @fballiano feedback

1. First observation, the word **protection** in "CSRF protection" must be written with a capital P. Thus we follow a rule regarding Backend texts where each word begins with a capital letter, except small texts 1 - 3 letters.

2. The important label must be inserted under the option "**Important! Enabling this option means that your custom templates used in checkout process, contact and newsletter contain form_key output. Otherwise the forms may not work**".

3. I disabled the option and the paragraph beginning with Important is displayed at the top under the menu. It must be modified as text and **link** because it no longer concerns the Checkout form only, "**Important: Formkey validation on checkout process, contact and newsletter are disabled. This may expose security risks. We strongly recommend to enable it in [System / CSRF Protection].**"

formkey_message

**Comment by fballiano** [comment]
but that label is default to yes since 2015 :-) I think that mentioning in the release notes and README is enough and it shouldn't be seen as a "problem" because it's an important security feature, when upgrading you upgrade your templates (cause the info is in the release notes) and that's it

**Comment by fballiano** [comment]
1. I've changed the translation of that label, although that specific label wasn't touched at all by this PR.
2. I think that mentioning in the release notes and README is enough and it shouldn't be seen as a "problem" because it's an important security feature, when upgrading you upgrade your templates (cause the info is in the release notes) and that's it. If we add that disclaimer, all of the correctly upgraded websites will have that warning forever and I don't think it's the right choice.
3. I've written it in a slightly different way "Important: Formkey validation is disabled. This may expose security risks. We strongly recommend to enable in Advanced / System, to protect your checkout process, newsletter subscription, contact form and more.", is that ok? maybe we can add something "please check with your developer that your custom templates are compatible with CSRF" in this same warning?

**Comment by ADDISON74** [comment]
**2** - Here I leave it to your discretion how many read the README file. I noticed that a lot of people don't read it and only those of us who are aware of the changes know about them. In another PR, the information about CSRF unification must be added in a section of the README.

**3.1** - The link which appears in the warning is **Admin / Security**

link1

Based on this the new link must be **System / CSRF Protection**

link2

**3.2** - For the warning message, the paragraph I propose is the following (keeping the idea from Magento and yours)

**Important: Formkey validation is disabled. This may expose you to security risks. We strongly recommend to enable it in System / CSRF Protection to protect your checkout, contact and newsletter forms.**

I chose to use "forms" because I am addressing the administrator who will immediately understand what it is about. I sorted them in the alphabetical order to be easier to identify. When we will add other protected forms we will insert them in the message.

**Comment by fballiano** [comment]
2. it is true that people don't read the README, I don't even blame them, they should read the release notes but we weren't the best at writing them precisely, I would actually add a docs/release-notes/v21 and have a text file for every release, I think it's easier for us to document the changes this way instead of keeping it too short in the README which is anyway already too long and a bit messy
3.1 I had already changed the link before
3.2 so much better. I've changed it (but I left the "Advanced" word as per the section of the admin configuration)

File: app/design/adminhtml/default/default/template/notification/formkey.phtml [hunk 3]
```diff
@@ -20,9 +20,9 @@
...
         <strong style="color:red">Important: </strong>
-        <span>Formkey validation on checkout disabled. This may expose security risks.
-        We strongly recommend to Enable Form Key Validation On Checkout in
-        <a href="<?php echo $this->getSecurityAdminUrl(); ?>">Admin / Security</a>,
-        for protect your own checkout process. </span>
+        <span>Formkey validation is disabled. This may expose you to security risks.
+        We strongly recommend to enable in
+        <a href="<?php echo $this->getEnableCSRFUrl(); ?>">Advanced / System / CSRF Protection</a>,
+        to protect your checkout, contact and newsletter forms.</span>
     </div>
```
ADDISON74, line 27: Do you think that this warning should be translated into the language of the chosen interface in the Backend? I think so, it should exist in a local CSV file.

fballiano: in theory it should, but:
- it was never translated before (it exists since years)
- it is very long and with dynamic content which makes it a bit harder to put in the CSV
I'd leave it like that...

ADDISON74 (approved):

elidrissidev (approved):

elidrissidev:
PHPStan is failing though

fballiano:
@elidrissidev it's failing on the next branch because we've to do https://github.com/OpenMage/magento-lts/pull/3163 (now that there's 8.1 in the composer.json) and probably fix something, but it's not related to this PR ;-)

PR #3148, opened 2023-04-05
[BC] Converted RWD's default logos to SVG (and removed some Magento names)

fballiano:
This PR removes the reference to the `logo.gif` files (and _email.gif and _print.gif) which do not look good on high resolution displays, replacing all of them with the SVG version, which looks perfect at all resolutions.

I've also replaced a few occurrences of the "Magento" name that were out of place in those contexts. 

Since we're removing some files it should be considered a breaking change, although I'm sure nobody is using those file in any production environment.

BEFORE:
Screenshot 2023-04-06 alle 16 05 09

AFTER:
Screenshot 2023-04-06 alle 16 04 58


matteotestoni (approved):

app/code/core/Mage/SalesRule/Model/Validator.php
```diff
@@ -179,6 +179,11 @@ protected function _canProcessRule($rule, $address)
                         $rule->setIsValidForAddress($address, false);
                         return false;
                     }
+                    // check coupon expiration
```
elidrissidev, line 182: This doesn't belong here I think?

sreichel:
Better add a new update script to change email templates?

fballiano:
email temples should read the logo variable with something like the `{{var something}}` syntax so nothing to update there. 

Anyway I don't think I'd want to force force an update on user created content like a template (which they probably would be heavily customised).

sreichel:
You are right. "Reset" the templates is no good idea.

Mayby add so conditions to that update script ...

```
if (emailTemplateDb === emailTemplateDefault) {
    // replace logos here
}
```

fballiano:
It can't be done, templates in the repo do not reference the logo files:

Screenshot 2023-04-07 alle 23 21 08

they're not even part of the RWD theme, they're part of the "app/locale" translation packages

app/code/core/Mage/Sales/sql/sales_setup/mysql4-upgrade-0.6.2-0.7.0.php
```diff
@@ -19,8 +19,8 @@
...
...             </thead>\r\n                                    <tbody>\r\n                                 <tr>\r\n                                        <td valign=\"top\" style=\"padding:7px 9px 9px 9px; border:1px solid #bebcb7; border-top:0; background:#f8f7f5;\">{{var order.shipping_address.getFormated(\'html\')}}</td>\r\n                                     <td>&nbsp;</td>\r\n                                     <td valign=\"top\" style=\"padding:7px 9px 9px 9px; border:1px solid #bebcb7; border-top:0; background:#f8f7f5;\">{{var order.shipping_description}}</td>\r\n                                   </tr>\r\n                                   </tbody>\r\n                                </table><br/>\r\n\r\n{{var items_html}}<br/>\r\n      {{var order.getEmailCustomerNote()}}                          \r\n                                <p>Thank you again,<br/><strong>Magento Demo Store</strong></p>\r\n\r\n\r\n                             </td>\r\n                           </tr>\r\n                       </table>\r\n                    \r\n                    </td>\r\n               </tr>\r\n           </table>\r\n            </div>\r\n' WHERE template_code='New order (HTML)';
```
ADDISON74, line 22: ```Thank you again,<br/><strong>Magento Demo Store.```

We should replace Magento with OpenMage.

fballiano: done!

app/code/core/Mage/Core/Model/Email/Template/Abstract.php
```diff
@@ -117,7 +117,7 @@ protected function _getLogoUrl($store)
                 return Mage::getBaseUrl('media') . $uploadDir . '/' . $fileName;
             }
         }
-        return Mage::getDesign()->getSkinUrl('images/logo_email.gif');
+        return Mage::getDesign()->getSkinUrl('images/logo.svg');
```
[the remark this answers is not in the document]
fballiano: mmm as always you have a great point. would it be possible if the user has a custom theme with a custom logo? I wouldn't want to be stuck with a gif logo forever but it's still can't force the users to migrate to svg.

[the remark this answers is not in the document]
kiatng: Something like the `??` operator? If SVG does not exist, fallback to GIF.

kiatng, line 120: Would this work?

```suggestion
        $fullFileName = Mage::getDesign()->getFilename('images/logo_email.gif', ['_type' => 'skin']);
        return file_exists($fullFileName)
             ? Mage::getDesign()->getSkinUrl('images/logo_email.gif');
             : Mage::getDesign()->getSkinUrl('images/logo.svg');
```

fballiano: I tested your code and it's great as usual (there's a `;` on the second line which shouldn't be there but it's no big deal).

The thing I'm wondering is... also using composer installation of a OM project, no file ever gets deleted from the file system also if it's deleted on our repo. which means that logo_email.gif will always be there...

This is a pickle.

fballiano: if we could remove the `base` folder from the fallback...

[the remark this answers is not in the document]
justinbeaty: I believe that some clients, GMail in particular, don't support SVG at all.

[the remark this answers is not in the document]
fballiano: mmmm https://designmodo.com/html-css-emails I guess I'll have to revert the change regarding the email part and leave it as it was before...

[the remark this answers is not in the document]
justinbeaty: It's truly a shame we can't use SVG in email. One option though, is that we could modify kiatng's code to check for the existence of logo_email.svg, and fallback to logo_email.gif. That way, if we ever can use SVG in email, we have the code in place. Or if someone doesn't care about GMail support... It shouldn't affect anything negatively, since I would guess nobody already has a file named logo_email.svg in their source.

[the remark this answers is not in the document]
fballiano: mmm I think it's easier to reintroduce it in the future if there will be the support, it's literally a couple of lines

app/code/core/Mage/Sales/sql/sales_setup/mysql4-upgrade-0.6.2-0.7.0.php
```diff
@@ -19,8 +19,8 @@
...
```
justinbeaty, line 22: We should revert these upgrade scripts too, right?

app/code/core/Mage/SalesRule/Model/Validator.php
```diff
@@ -178,11 +178,6 @@ protected function _canProcessRule($rule, $address)
                         $rule->setIsValidForAddress($address, false);
                         return false;
                     }
-                    // check coupon expiration
-                    if ($coupon->hasExpirationDate() && ($coupon->getExpirationDate() < Mage::getModel('core/date')->date())) {
```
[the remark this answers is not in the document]
fballiano: I don't understand why this keeps happening to me lately, I "git reset --hard" all the time and still these things keep happening...
thanks for pointing it out

app/code/core/Mage/Sales/sql/sales_setup/mysql4-upgrade-0.6.2-0.7.0.php
fballiano: I've reverted the logo.svg parts but there are still some difference because there were soe "Magento" names that I've converted to "OpenMage"

justinbeaty:
What should we do about logo_src_small ? Since SVG is scalable, we have no need for that anymore. 

https://github.com/OpenMage/magento-lts/blob/1342703287ce96400ec654567b747b227de6e43b/app/design/frontend/rwd/default/template/page/html/header.phtml#L35

We just end up with two  tags:

image

We could remove it, because it seems an outdated practice. Or, we could check to see if the two are the same path, and output only one  tag?

(I'll be back in an hour or so).

fballiano:
@justinbeaty I've left only one logo in the header, the sizing in my opinion is not super right but I'd check that in another PR

justinbeaty:
https://github.com/OpenMage/magento-lts/blob/6b57a6cf189918da88327a974f54145080c3e467/skin/frontend/rwd/default/scss/layout/_header.scss#L134-L139

https://github.com/OpenMage/magento-lts/blob/6b57a6cf189918da88327a974f54145080c3e467/skin/frontend/rwd/default/css/styles.css#L3243-L3248

These lines can also go away too.

For the sizing -- I was going to mention it, but ultimately I think it's fine as it is now. The old logo.gif just had vertical padding in the image vs the svg which is cropped properly.

One last thing -- do we deprecate this method?

fballiano:
@justinbeaty those CSS were already removed.

I'll think about the getLogoSrcSmall() method for a moment.

justinbeaty:
> @justinbeaty those CSS were already removed.

There's a second set a few lines after what you already removed in both style.css and _header.scss:

image

-----

> I'll think about the getLogoSrcSmall() method for a moment.

Yeah, I'm not sure the best thing to do here. I guess one day when we have rwd2 then we definitely don't need it, but for now I'm not sure.

fballiano:
@justinbeaty right, pushed a new commit.

about the getLogoSrcSmall, I've marked it as deprecated

justinbeaty (approved):

ADDISON74:
I looked over the code and the changes concern only the frontend and the printed page. The rest for the email remained as it was initially. Am I right?

The problem of viewing in different email clients was well discussed. SVG is not well digested by some of them. However, as a small improvement, maybe we should leave behind the GIF format and replace the file named logo.gif with an image in PNG format.

fballiano:
yes the emails remain untouched

ADDISON74 (approved):

empiricompany (approved):

fballiano:
Merged per 1 green and 2 gray checks

sreichel:
Not so nice. I disagree with changing existing update scripts :/ Better add a new one.

fballiano:
Not everybody always have your same ideas.

sreichel:
Why not add a new update script - as it was done before?

ADDISON74:
If someone does an installation from scratch then the db script will be executed. If he does not then the script is not run and the old data remains in the database. I tend to agree with Sven and with the rule that when changes are made to the db installation scripts, new ones should be created, precisely so that the change is operated in all cases.

fballiano:
To ask a change after these many weeks… 
Anyway I don’t remember everything of this pr since it’s been there for looong time anyway to change user data is not always a good idea just because it seems cleaner.

so old data stays ok while a new install gets the new data.

fballiano:
And anyway anybody can always write a new pr with a new script if he feels thats necessary

sreichel:
@ADDISON74 thanks. There are different situation. Rolling back DB, rolling back files, resetting `core_resource`, plain install ...

sreichel:
> precisely so that the change is operated in all cases.

:+1:

sreichel:
> To ask a change after these many weeks…

Cmon ... https://github.com/OpenMage/magento-lts/pull/3148#issuecomment-1500661500

ADDISON74:
There are few of us who still contribute to this project. For this solid reason the current PR has not been completely brushed. but I am glad that there are still pairs of eyes that see issues and that they need to be discussed. I am sorry to say it publicly, but a maintainer should not ask others to create a PR or a new script, we are still addressing people who have expectations from us and above all a different level of knowledge. It's like going to a car service and the mechanic asks us to do something related to his knowledge.

fballiano:
Again, not everybody has to do what you want sven, your opinion is not the right one just because you have it.

You know how to open a pr if you want, don’t just complain when somebody else tries to do something

fballiano:
@ADDISON74 i do not agree. Maintainers should make their hands dirty instead pf just complaining when somebody does something.

sreichel:
@fballiano i highly appreciate what you do for OM, but please keep quality over quantity.

fballiano:
Sorry, I do not think you aprecciate my work here since your are always criticising everything I do.
Also, sorry for not meeting your quality standards, there's a million things to do on this project and I prefer doing stuff to the best of my skills instead of trying to be perfect and never finish.
There are many extremely more skilled developers than me here, I'm the first one knowing that. My first objecting is to have them create great stuff, merge it and continue to move on.

You could help me housekeeping the 85PRs that need to be taken care of, we've to examine the 1.9.4.x ones and understand if they're still useful (and rebase them) or close them. I believe that it's absolutely necessary to respect people's time and work and everybody that submitted a PR that's been abandoned should receive an update, sometimes an apology, sometimes we should finish their job. That would give users more will to participate, and I think everybody knows we would need more workforce.

sreichel:
> Sorry, I do not think you aprecciate my work here since your are always criticising everything I do.

__I really do!__ I just do not agree with every change.

ADDISON74:
I believe that what we do in this repository, we do it because we like to be here, to improve our knowledge, to offer our experiences to others. At the end there should be that absolutely natural satisfaction of the programmer who leaves something behind. That's why I appreciate any opinion for and against, provided it helps us move forward.

In the case of this PR, I agree with Sven, the changes should also cover the current installations. If Sven can make a new PR it would be wonderful, I approve it immediately.

sreichel:
> You could help me housekeeping

Na. I have done it for years and have no mood for it.

(PS: when you close PRs ... maybe add `fixed`, `invalid`, `wont be fixed` tags ...)

PR #3155, opened 2023-04-09
[PHP 8.2] Fixed some deprecation warnings

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3152 @ADDISON74 found out some PHP 8.2 deprecation warnings, this PR should fix them.

Note: at the moment it still doesn't fix the problems with Mage_Paypal_Model_Config that apparently were introduced with https://github.com/OpenMage/magento-lts/pull/2759

### Related Pull Requests
1. https://github.com/OpenMage/magento-lts/pull/2759

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/3152

### DIFF
--- a/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
+++ b/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
@@ -75,7 +75,9 @@ class Mage_ConfigurableSwatches_Helper_Data extends Mage_Core_Helper_Abstract
      */
     public static function normalizeKey($key)
     {
-        if ($key === null || $key === '') return '';
+        if ($key === null || $key === '') {
+             return '';
+        }
         if (function_exists('mb_strtolower')) {
             return trim(mb_strtolower($key, 'UTF-8'));
         }
--- a/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
+++ b/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
@@ -75,6 +75,7 @@ class Mage_ConfigurableSwatches_Helper_Data extends Mage_Core_Helper_Abstract
      */
     public static function normalizeKey($key)
     {
+        if ($key === null || $key === '') return '';
         if (function_exists('mb_strtolower')) {
             return trim(mb_strtolower($key, 'UTF-8'));
         }
--- a/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
+++ b/app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
@@ -76,7 +76,7 @@ class Mage_ConfigurableSwatches_Helper_Data extends Mage_Core_Helper_Abstract
     public static function normalizeKey($key)
     {
         if ($key === null || $key === '') {
-             return '';
+            return '';
         }
         if (function_exists('mb_strtolower')) {
             return trim(mb_strtolower($key, 'UTF-8'));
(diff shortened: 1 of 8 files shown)


app/code/core/Mage/Rule/Model/Condition/Abstract.php
```diff
@@ -812,6 +812,7 @@ public function validateAttribute($validatedValue)
      */
     protected function _compareValues($validatedValue, $value, $strict = true)
     {
+        $validatedValue = $validatedValue ?? '';
         if ($strict && is_numeric($validatedValue) && is_numeric($value)) {
             return $validatedValue == $value;
         } else {
```
kiatng, line 818:
```suggestion
        if ($strict && is_numeric($validatedValue) && is_numeric($value)) {
            return $validatedValue == $value;
        } else {
             $validatedValue = $validatedValue ?? '';
```

app/code/core/Mage/Eav/Model/Entity/Attribute/Source/Table.php
```diff
@@ -75,7 +75,7 @@ public function getAllOptions($withEmpty = true, $defaultValues = false)
     public function getOptionText($value)
     {
         $isMultiple = false;
-        if (strpos($value, ',')) {
+        if (strpos($value ?? '', ',')) {
```
kiatng, line 78:
```suggestion
        if ($value && strpos($value, ',')) {
```

app/code/core/Mage/Rule/Model/Condition/Abstract.php
fballiano: it should be ok now :-)

app/code/core/Mage/SalesRule/Model/Validator.php
```diff
@@ -962,7 +962,7 @@ protected function _addDiscountDescription($address, $rule)
         $label = '';
         if ($ruleLabel) {
             $label = $ruleLabel;
-        } elseif (strlen($address->getCouponCode())) {
+        } elseif (strlen($address->getCouponCode() ?? '')) {
```
elidrissidev, line 965: `coupon_code` getter in Quote model was overridden in #2825 to always return empty string, maybe a similar approach can be done with address.

fballiano: @elidrissidev I think I did it the right way now

app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -75,6 +75,7 @@ public function getHyphenatedString($str)
      */
     public static function normalizeKey($key)
     {
+        if ($key === null || $key === '') return '';
```
kiatng, line 78:
```suggestion
        if ($key === null || $key === '') {
             return '';
        }
```

app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -75,6 +75,9 @@ public function getHyphenatedString($str)
      */
     public static function normalizeKey($key)
     {
+        if ($key === null || $key === '') {
+             return '';
```
fballiano, line 79:
```suggestion
            return '';
```

kiatng (approved):

ADDISON74 (approved):
Let's merge it and I will make a list with the warnings left behind we still get in the log files.

PR #3160, opened 2023-04-11
PHP8.1 is now the minimum required version for OM v21

fballiano:
I think for OM v21 we should require at least PHP 8.1 (I'd actually do 8.2 if you agree...), which will also help us reduce PHPStan baseline (will be done in a later PR).

- phpstan and workflows updates will come with separate PRs


kiatng (dismissed):

luigifab:
This will break many installations.

fballiano:
It is v21 @luigifab

luigifab:
Yes, but I still dislike, for me OpenMage still works with PHP 7.2+, (with composer php polyfill), even if I use it with PHP 8.0, and test it with PHP 8.2.

fballiano:
Because we shouldn’t encourage people to use outdated stacks which are not maintained upstream.

kiatng (approved):

ADDISON74 (dismissed):

ADDISON74:
I was looking over a PHP usage report in 2023. Over 50% are still on 7.4. Debian 11 has its own implementation with LTS until 2026. This does not mean that version 8 cannot be installed. For this reason I approved the PR and I believe that OpenMage must keep up with the new versions and not remain anchored in older versions. Remember how a company (Inchoo) provided an extension for a long time for M1, because the Magento team was not able to have its own implementation. Then many stayed with PHP 5.6 and filled the forums of the time with problems related to 7.0. When the hosting service provider forces you to switch to new versions, then you really have a problem.

fballiano:
I think we should make things easy for us, we focus on the main thing, like we did for the release-versioning RFC, people will probably make OM work on 7.4 with no problems, but we run our tests, workflow and everything on the current version.

I want to be in the present, not preserver the work of somebody who never want to update. if you never want to update, stay on php5.6 with magento 1.9.3 and you're perfectly fine. until you get a GDPR audit. then you hoped you would have updated your systems when you had time.

fballiano:
I'm merging since it was already approved but I had to solve a conflict with composer.lock (only the hash changed) which invalidated the reviews.

positive side effects: phpstan is now complainig about php 8.1+ problems, which are the ones that we're struggling to find!

PR #3163, opened 2023-04-11
Github workflows now run on PHP 8.1

fballiano:
If https://github.com/OpenMage/magento-lts/pull/3160 gets approved then this PR will change all github workflows to run on PHP 8.1 minimum.


sreichel:
As long as 7.4 is supported, id not change this.

fballiano:
as I wrote in the PR description, if https://github.com/OpenMage/magento-lts/pull/3160 gets approved.

sreichel:
So we have differentt workflows for "main" and "next" (and "v19")?

fballiano:
the push/pr workflows run on their own branch configuration

fballiano:
this is why this PR's "checks" have only php 8.1/8.2 while the others stay normal with php 7.4

kiatng (approved):

PR #3164, opened 2023-04-11
Replaced setData by setEvent

luigifab:
### Description

This small PR replace `setData` by `setEvent`.
I found that with Blackfire some months ago, but I lost the before/after.

https://github.com/OpenMage/magento-lts/blob/cbdac32a562b8622340c1d4fc77b964ee09dcd58/lib/Varien/Event/Observer.php#L121-L127

Yes this will not change many things, but this + many others...

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Core/Model/App.php
+++ b/app/code/core/Mage/Core/Model/App.php
@@ -1376,7 +1376,7 @@ class Mage_Core_Model_App
             }
 
             foreach ($events[$eventName]['observers'] as $obsName => $obs) {
-                $observer->setData(['event' => $event]);
+                $observer->setEvent($event);
                 Varien_Profiler::start('OBSERVER: ' . $obsName);
                 switch ($obs['type']) {
                     case 'disabled':


Sdfendor (approved):

fballiano:
mmm this specific PR seems to me to make things worse because the setEvent calls the setData, so we're adding a function call for every element of the foreach

it will change nothing but actually I don't really see why it would be better to implement the PR :-)

kiatng:
The replacement is not the same as the original:
```php
                $observer->setData(['event' => $event]); // this remove all previous `_data` in $observer
                $observer->setEvent($event); // this add the `$event` data to `_data`.
```
See:
https://github.com/OpenMage/magento-lts/blob/1342703287ce96400ec654567b747b227de6e43b/lib/Varien/Object.php#L243-L249

When you pass an array, the entire `_data` is set to the array.

So, unless there is a strong case to change the original behavior, I object to this PR.

luigifab:
You're so right, I forgot that `$observer = new Varien_Event_Observer();` is not in the `foreach`...

fballiano:
I'll closed this one at the moment, after https://github.com/OpenMage/magento-lts/pull/3164#issuecomment-1515634071

luigifab:
I pushed another version on same branch... but github doesn't update the PR.
Here is a test:

```php
// php web/scripts/nul.php
// 0.042194128036499
// 0.027787923812866
$observer = new Varien_Event_Observer();
$event = new Varien_Event(['a' => 1]);
$event->setName('pouet');

$start = microtime(true);
for ($i = 0; $i < 10000; $i++) {
	for ($j = 0; $j < 50; $j++) {
		$observer->setData(['event' => $event]);
	}
}
echo microtime(true) - $start,"\n";

$start = microtime(true);
for ($i = 0; $i < 10000; $i++) {
	for ($j = 0; $j < 50; $j++) {
		$observer->setEvent($event);
	}
}
echo microtime(true) - $start,"\n";
```

fballiano:
it doesn't allow me to reopen the PR, it usually happens if the original branch or repo was deleted :-\

PR #3171, opened 2023-04-13
Removed all @author tags from docblocks

fballiano:
The `@author ` tag is used to document the author of Structural Elements as stated in https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/author.html#author

I think it shouldn't be considered as a copyright/license tag (that info is elsewhere) and since we (as in openmage contributors) do not add the author of every patch in the source code (it would be extremely overkill and `git` has the info anyway in the `git blame`) I think we can safely remove all of those tags.

Also, the email written in the removed tags is probably dead since years.

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -49,8 +49,6 @@ foreach (glob(BP . DS . 'app' . DS . 'etc' . DS . 'includes' . DS . '*.php') as 
 
 /**
  * Main Mage hub class
- *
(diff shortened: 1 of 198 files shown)


luigifab (approved):
The script for copyright date will be happy.

fballiano:
Thanks for reminding me to check the squash comment (if it will be merged) not to screw with the copyright script!!

ADDISON74 (approved):

PR #3177, opened 2023-04-15
[PHP 8.2] Fixed deprecation error for creation of dynamic property

fballiano:
After https://github.com/OpenMage/magento-lts/pull/2759, when visiting a product page or the cart page (with developer mode) we get:

`Deprecated functionality: Creation of dynamic property Mage_Paypal_Model_Config::$visible_on_product is deprecated  in app/code/core/Mage/Paypal/Model/Config.php on line 781"`

This PR partially reverts #2759 but only for 2 specific properties

How to test:
- enable developer mode
- visit product page or cart

### DIFF
--- a/app/code/core/Mage/Paypal/Model/Config.php
+++ b/app/code/core/Mage/Paypal/Model/Config.php
@@ -19,32 +19,6 @@
  *
  * @category   Mage
  * @package    Mage_Paypal
- *
- * @property mixed $allow_ba_signup;
- * @property mixed $api_cert;
- * @property mixed $api_password;
- * @property mixed $api_signature;
- * @property mixed $api_username;
- * @property mixed $apiAuthentication;
- * @property mixed $apiPassword;
- * @property mixed $apiSignature;
- * @property mixed $apiUsername;
- * @property mixed $business_account;
- * @property mixed $businessAccount;
- * @property mixed $buttonFlavor;
- * @property mixed $buttonType;
- * @property mixed $cctypes;
- * @property mixed $debug;
- * @property mixed $lineItemsEnabled;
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

elidrissidev:
The thing is, `<visible_on_cart>` and `<visible_on_product>` are in Paypal's `config.xml`, and if they exist as a property, `__get` will not be called and a `null` value will be retrieved every time. But since the `__get` method is setting them dynamically, I think there's no way around it other than to refactor the method.

elidrissidev:
An alternative approach could be to use array instead (untested):

```patch
     public function __get($key)
     {
+        if (array_key_exists($this->_config, $key)) {
+            return $this->_config[$key];
+        }
+
         $underscored = strtolower(preg_replace('/(.)([A-Z])/', "$1_$2", $key));
         $value = Mage::getStoreConfig($this->_getSpecificConfigPath($underscored), $this->_storeId);
         $value = $this->_prepareValue($underscored, $value);
-        $this->$key = $value;
-        $this->$underscored = $value;
+        $this->_config[$key] = $value;
+        $this->_config[$underscored] = $value;
         return $value;
     }
```

fballiano:
@elidrissidev your solution is much better, you want to commit it or should I do it?

elidrissidev:
I won't have the chance now, you can do it if you're free.

fballiano:
I can never make a sandbox paypal account work but I've reworked this PR, I thought to check _config also for the underscored key... what do you think?

elidrissidev (approved):
Tested. Accessing properties works fine and no more deprecation errors.

ADDISON74 (approved):

PR #2939, opened 2023-01-14
PHP8.1 Fixes passing null to trim() - for non-existing payment methods

sreichel:
### Fixed Issues (if relevant)

```
Deprecated functionality: trim(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Core/Model/Config.php on line 1422

#0 [internal function]: mageCoreErrorHandler(8192, 'trim(): Passing...', '/var/www/html/a...', 1422)
#1 /var/www/html/app/code/core/Mage/Core/Model/Config.php(1422): trim(NULL)
#2 /var/www/html/app/code/core/Mage/Core/Model/Config.php(1444): Mage_Core_Model_Config->getModelClassName(NULL)
#3 /var/www/html/app/Mage.php(532): Mage_Core_Model_Config->getModelInstance(NULL, Array)
#4 /var/www/html/app/code/core/Mage/Payment/Helper/Data.php(46): Mage::getModel(NULL)
#5 /var/www/html/app/code/core/Mage/Payment/Helper/Data.php(162): Mage_Payment_Helper_Data->getMethodInstance('pbridge')
```

### Manual testing scenarios (*)

1. use sample data
2. go to admin -> sales -> Recurring Profiles
3. ...

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3181, opened 2023-04-15
Fixed logging too many exceptions for payment methods

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3179 @ADDISON74 pointed out that, using the backend, and opening an order detail page, some exceptions were logged about missing payment methods.

This is because the list of payment methods tries to load the model (for the payment method) and logs an exception if it's not found.

This is ok in some cases (when the order was paid with a method that has been removed later) but not in all the other cases.

So I thought to kinda rewrite the getPaymentMethodList method of Mage_Payment_Helper_Data.

If I'm not mistaken the before/after output of the getPaymentMethodList method stays the same:
before:
Screenshot 2023-04-15 alle 16 33 49

after:
Screenshot 2023-04-15 alle 16 33 21

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/2939

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/3179

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -393,7 +393,7 @@ final class Mage
      *
      * @param string $path
      * @param null|string|bool|int|Mage_Core_Model_Store $store
-     * @return string|null
+     * @return mixed
      */
     public static function getStoreConfig($path, $store = null)
     {
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -393,7 +393,7 @@ final class Mage
      *
      * @param string $path
      * @param null|string|bool|int|Mage_Core_Model_Store $store
-     * @return mixed
+     * @return string|null
      */
     public static function getStoreConfig($path, $store = null)
     {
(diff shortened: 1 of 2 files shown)


ADDISON74:
This PR fixes **Sales > Orders > View** any order. No more lines in exceptions.log.

We still have the same issue in **Customers > Manage Customers > Edit any customer**. Will you create a new PR for Customers?

(PHP 8.1)

fballiano:
I thought it was the same but I’ll check it out asap!

fballiano:
@ADDISON74 I've added the fix for the customer detail page in this PR because it was sharing part of the code ;-)

ADDISON74 (dismissed):
Tested. No more warnings in the exceptions.log file related to the missing payment methods. For both sections in Backend (View Order and Edit Customer).

ADDISON74 (approved):

elidrissidev (approved):

PR #3190, opened 2023-04-18
Added ids to stores list

luigifab:
### Description

This small PR adds ids to System > Stores list.
This allow to CTRL+F any ID, even it's not very use full.

image

### Contribution checklist


fballiano (approved):
The IDs are shown when you "hover" on the links to website/store/storeview but I like this PR anyway!

fballiano:
@ADDISON74 maybe I'm saying something stupid, but is there a possibility you don't like these changes because of the font @luigifab is using? I also don't super love that font but it's not part of this PR.
This PR is maybe not extremely necessary but it happened to me many time to be hovering the store name just to see the ID, without the possibility to copy it, that's why I still think this is a good addition :-)

ADDISON74:
@fballiano - The problem is not related to the font face. I don't like that the information listed implicitly as in the image shown by me was completely replaced with something else. I proposed a middle option if the implementation of this PR is desired.

```(Code: , ID: )```

for all three columns. In fact, I kept the code word and added the ID maintaining the same format with round brackets.

elidrissidev:
Personally I dislike it because it bloats the table, but I can see it being useful to some people. My only suggestion is to keep the old format as mentioned by @ADDISON74.

fballiano:
mmm I think that the ID should always be the first column, it is the first column of every mask in our backend. Having the code as the first column seems a bit unnatural to me.

But I guess that since there's no consensus here and the PR is as it is, we have no other way as to close it.

fballiano:
ok guys it seems we don't have consensus here so sadly I'll close the PR.

IMHO this table is used every once in a while and it would be more useful to have some more data, who care of a bit more clutter if you open this mask once per year.

but I respect the common opinion that doesn't like this change.

sorry @luigifab

Issue #3174, opened 2023-04-14
Upgrading from Magento 1.9 to OpenMage 19.4.23 with PHP 8.1 corrupts HTML templates


djixas:
### Preconditions (*)

1. Magento 1.9.4.5 
2. OpenMage 19.4.23
3. PHP 8.1

### Steps to reproduce (*)

1. Take any Magento 1.9 install and upgrade to OpenMage, then enable PHP 8.1, this might also work on default OpenMage, I am not sure
2. Now try to place an order, click admin forgot pass or anything else, so you get email from TEMPLATES

### Expected result (*)

1. It should display emails correctly
2.

Sample "Test" email displays this content with PHP 7.4:

> Name: 
> Email: 
> Telephone: 
> 
> Comment: This is a test template email - your email settings are correct!
> 

### Actual result (*)

1. [Screenshots, logs or description]
![mail](https://user-images.githubusercontent.com/12409425/232037148-4b3015dc-fcbd-46da-babf-3a9dc1d11bbf.png)

2.

Sample "Test" email displays this content with PHP 8.1:

=0A=0AName: =0AEmail: =0ATelephone: =0A=0AComment: This is a test template email - your email settings are correct!

Tested also with https://github.com/icecubenz-open-mage/Icecube_CustomEmailServer just to be safe, same issue, but the issue disappears if I downgrade to PHP 7.4, then email templates work properly. And it's for all emails, forgot admin pass, etc. Already tried reloading all email templates, same issue. 

PHP ini settings / modules are identical.

djixas:
P.S. Tested with two different Magento upgrades, one done via old school upgrade script of yours, other is by replacing all old files with OpenMage. It also keeps generating mailformed address errors:

> A message that you sent contained one or more recipient addresses that were incorrectly constructed:
> 
>   =?utf-8?B??= <>: missing or malformed local part
> 
> This address has been ignored. There were no other addresses in your message, and so no attempt at delivery was possible.
> 
> ------ This is a copy of your message, including all the headers. ------
> 
> To: =?utf-8?B??= <>
> Subject: =?utf-8?B?Q29udGFjdCBGb3Jt?=
> X-PHP-Originating-Script: 1047:Sendmail.php
> From: xxx <xxxxx@xxxx.xxx>
>  Date: Sun, 16 Apr 2023 14:05:12 +0000
>  Content-Type: text/plain; charset=utf-8
>  Content-Transfer-Encoding: quoted-printable
[...]

PR #3191, opened 2023-04-18
Replaced Zend_Mail_Transport_Sendmail with Laminas\Mail\Transport\Sendmail

fballiano:
After https://github.com/OpenMage/magento-lts/issues/3174 () I thought that maybe Zend_mail would be a good candidate for a replacement with a more modern implementation so I made a test.

- this PR only replaces Zend_Mail_Transport_Sendmail with its Laminas counterpart
- this PR does not replace the entire Zend_Mail (at the moment, let's see what the reception of this code and decide what to do next)
- I've tested the email sending and it works correctly
- I've also tested installing aschroder/smtp_pro to see if this change was breaking it but it seems not to break it
- I've targeted this PR to `main` since it doesn't seem at the moment to be breaking anything, maybe it should be considered BC and targeted to `next`?

**How to test?** be sure to be using sendmail on your server, install the PR, composer install, try to send a "forgot password" email

What do you think about this idea?

### DIFF
--- a/app/code/core/Mage/Core/Model/Email/Queue.php
+++ b/app/code/core/Mage/Core/Model/Email/Queue.php
@@ -193,7 +193,7 @@ class Mage_Core_Model_Email_Queue extends Mage_Core_Model_Abstract
             if ($message->getId()) {
                 $parameters = new Varien_Object($message->getMessageParameters());
                 if ($parameters->getReturnPathEmail() !== null) {
-                    $mailTransport = new Zend_Mail_Transport_Sendmail("-f" . $parameters->getReturnPathEmail());
+                    $mailTransport = new \Laminas\Mail\Transport\Sendmail("-f" . $parameters->getReturnPathEmail());
                     Zend_Mail::setDefaultTransport($mailTransport);
                 }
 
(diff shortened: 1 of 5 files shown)


fballiano:
exactly, from what I've seen zf1future is not super quick in fixing problems (it's probably also a good thing to be careful, for sure) which could be a bit of a problem. it is also used by a limited amount of people compared to laminas or symfony.

and thanks to composer we're facilitated in adding different vendors, if needed.

laminas is mostly compatible with zend framework, there are still many differences, but in this case the syntax was the same.

I think that, without the zf1future step (which we will require for long time anyway), we wouldn't be able to do something like this specific PR.

So let's see :-)

fballiano:
I've done some more work trying to rewrite the whole Zend_Mail dependency but... I don't know... the whole paradigm has changed and you've to call a `$transport->send()` instead of `$mail->send($transport)` and to send a text/html email you've to do https://docs.laminas.dev/laminas-mail/message/attachments/

So, I don't know, is it worth the risk? Or better stick with the zf1f implementation and maybe add a new "zf1f patch" to our composer for the moment?

ADDISON74:
Maybe we should "force" the change in ZF1-Future. If they don't want to accept it, a patch can be implemented as @sreichel did with other changes. We take the parts we need from Laminas-Mail and implement them to solve the issue. 

I haven't checked, but what we were using before ZF1-Future would have created trobule with PHP 8.x?

fballiano:
> I haven't checked, but what we were using before ZF1-Future would have created trobule with PHP 8.x?

No idea actually, i can imagine it would have

fballiano:
it was an interesting chat to have anyway :-)

ADDISON74:
Indeed we discovered Laminas.

PR #3197, opened 2023-04-19
Avoid JS errors if the zipcode field has no label

azambon:
### Description
This PR solves a JS error that can happen if the theme used does not have a label for the zipcode field. The possibly buggy behavior was introduced by https://github.com/OpenMage/magento-lts/commit/34c0acf5b974d043fdeefab5bbdee9adf961affa

### Steps to reproduce
* start with a clean installation of magento LTS;
* customize the frontend theme so that the zipcode address field has no label;
* go to the checkout and choose to add a new address.

### Expected result
No JS errors should be thrown;

### Actual result
There is a JS error on line 403 or 413 of the js/varien/form.js file since the "label" variable is undefined.

### DIFF
--- a/js/varien/form.js
+++ b/js/varien/form.js
@@ -410,7 +410,7 @@ ZipUpdater.prototype = {
                 wildCard.hide();
             }
         } else {
-            if (label != undefined && !label.hasClassName('required')) {
+            if (label !== undefined && !label.hasClassName('required')) {
                 label.addClassName('required');
             }
             this.zipElement.addClassName('required-entry');
--- a/js/varien/form.js
+++ b/js/varien/form.js
@@ -384,13 +384,13 @@ ZipUpdater.prototype = {
 
     _setPostcodeOptional: function(){
         this.zipElement = $(this.zipElement);
-        if (this.zipElement == undefined) {
+        if (this.zipElement === undefined) {
             return false;
         }
 
         // find label
         var label = $$('label[for="' + this.zipElement.id + '"]')[0];
-        if (label != undefined) {
+        if (label !== undefined) {
             var wildCard = label.down('em') || label.down('span.required');
             if (!wildCard) {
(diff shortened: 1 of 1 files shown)


luigifab (dismissed):

js/varien/form.js
```diff
@@ -400,7 +400,7 @@ ZipUpdater.prototype = {
 
         // Make Zip and its label required/optional
         if (optionalZipCountries.indexOf(this.country) != -1) {
-            if (label.hasClassName('required')) {
+            if (label != undefined && label.hasClassName('required')) {
```
kiatng, line 403:
```suggestion
            if (label !== undefined && label.hasClassName('required')) {
```

js/varien/form.js
```diff
@@ -410,7 +410,7 @@ ZipUpdater.prototype = {
                 wildCard.hide();
             }
         } else {
-            if (!label.hasClassName('required')) {
+            if (label != undefined && !label.hasClassName('required')) {
```
kiatng, line 413:
```suggestion
            if (label !== undefined && !label.hasClassName('required')) {
```

kiatng (commented):
Why not strict check.

js/varien/form.js
kiatng: The check is inconsistent in the file, it has `typeof` and one of which is strict check. Probably different authors in the past leaving different signatures. The trend is going strict type.

azambon: Indeed, I wrote the checks in the same way as it was already done a few lines earlier. I tried to maintain consistency at least within this single method.

azambon:
> Could you please specify the steps to follow to reproduce this issue?

@ADDISON74 I edited the initial message

fballiano (approved):
i've converted the checks in that method to strict, let's not get stuck here :-)

ADDISON74 (approved):

PR #3199, opened 2023-04-20
Replace Zend_Validate with laminas-validate

sreichel:
### Description (*)

#3191 was an good idea imho. Followed this and started to replaced `Zend_Validate` with laminas components.

Some tests ... https://gist.github.com/sreichel/067428567d49c9ceab907fd10a36e2dc

### Questions or comments

Draft: only need some tests to replace all other Zend_Validate code ... (not commited yet)

Btw ... we dont need to stick with Zf1-Furture, There is also https://github.com/zf1s/zf1 that would allow to remove not needed Zend modules ...


ADDISON74:
You made good observations. If we could replace Zend 1 with Laminas, it would be a huge step. I can't figure out how this will affect existing extensions and what the level of work would be to modify the OpenMage source code. It would probably be simpler than removing Prototype library, on which many extensions relied when implementations on the JavaScript side were needed.

At this moment, I think that ZF1-Future will hold the umbrella for a good while.

sreichel:
> I can't figure out how this will affect existing extensions and what the level of work would be to modify the OpenMage source code.

@ADDISON74 its not about to replace ZF1 at all (for now), but some parts could aleady be changed to [laminas](https://github.com/OpenMage/magento-lts/discussions/2932#discussioncomment-4692582)...

It _could_ break 3rd-party code that extends `Zend_Validate` classes (seems very unlikly to me). There is only one file left in OM that needs adjustments .

> At this moment, I think that ZF1-Future will hold the umbrella for a good while.

Just wanted to mention [zf1s/zf1](https://github.com/zf1s/zf1) because of its structre (we only need a few parts from Zend1)

ADDISON74:
The only problem I see is from the developer's point of view. It would be preferable to have a single framework, not pieces of several (Laminas, ZF1-Future, ...).

Regarding the alternative to ZF1-Future, if we end up using certain parts of that repository, then we have to make sure that we have the best version between the two alternatives.

As I saw that the repositories move slowly, I think it was not a very good idea to remove the framework from OpenMage and link it with Composer. I think it would have been more appropriate to have a separate repository in OpenMage for this framework. So we no longer waited after the PR to be approved and we had a control of the source code.

PR #3201, opened 2023-04-21
Fixed deprecation warning on Mage_Catalog_Helper_Image

fballiano:


app/code/core/Mage/Catalog/Helper/Image.php
```diff
@@ -564,13 +564,18 @@ protected function getImageFile()
      */
     protected function parseSize($string)
     {
+        if ($string === null) {
+            return false;
+        }
+
         $size = explode('x', strtolower($string));
-        if (count($size) === 2) {
+        if (is_array($size) && count($size) === 2) {
```
elidrissidev, line 572: `explode` will always return an array, so no need for `is_array` check.

fballiano: done, thank you!

ADDISON74 (approved):

PR #3204, opened 2023-04-21
RWD theme: updated jQuery to 3.7.0

fballiano:
I realize now that we're still shipping jquery 1.12, this has to be changed.


ADDISON74:
This is a "sensitive" PR that deserves to be treated with all attention. 

1. The browser console must be followed because errors and warnings will certainly appear. 
2. The migration script must be added because no theme works fully with version jQuery v3.
3. jQuery v1 must not be removed from the OpenMage distribution, even if the branch is next.

Personally, I tried to use jQuery v3 with Magento/OpenMage, but the migration effort was too big to be worth the time invested. In addition, the extensions needed changes and I didn't always find solutions.

fballiano:
mmm weird I migrated a customer with not much pain, but anyway jquery1 has vulnerability issues and it should be removed.

justinbeaty:
Link to previous jQuery version upgrade, although that was just a minor release upgrade in the 1.x branch: #2442

Ultimately, we should figure out what is the best way to manage our frontend deps. Composer, npm, bower, etc?

fballiano:
I'd like to avoid things like bower/npm etc... but I've no idea if composer can't manage js libraries

justinbeaty:
We should use [jquery-migrate](https://github.com/jquery/jquery-migrate/tree/main). That should provide backwards compat for 1.12.x:

```
<script src="https://code.jquery.com/jquery-3.6.4.js"></script>
<script src="https://code.jquery.com/jquery-migrate-3.4.1.js"></script>
```

I think that would be enough to launch this onto the `next` branch. Definitely if there's known vulns in jQuery 1.x we should stop including it by default.

fballiano:
using the migrate in the next branch? then it could be in the main...

justinbeaty:
> but I've no idea if composer can't manage js libraries

It could somewhat, I apparently tested it here: https://github.com/OpenMage/magento-lts/pull/2442#issuecomment-1219678600

But it didn't play nice with `magento-core-composer-installer` (I think just because it doesn't know about `magento-root-dir`)

fballiano:
or we can add a line in the README to say to use the migrate script in case of problems

justinbeaty:
> using the migrate in the next branch? then it could be in the main...

I haven't used jQuery for some time and I don't recall if it provides 100% compatibility. 

Or maybe migrate script in main, and no migrate script in next?

fballiano:
if no-migrate in next, then it's already this PR ;-)

fballiano:
and if the theme gets so broken, it's probably time to rethink the whole theme because for sure that one has a lighthouse score of 23 and cannot (and should not) survive 2023.

ADDISON74:
From what I can see, the JQuery project is still maintained. Maybe a solution should be found to add it as a dependency and not update it periodically in this repository.

justinbeaty:
> From what I can see, the JQuery project is still maintained. Maybe a solution should be found to add it as a dependency and not update it periodically in this repository.

That's what I tried to do here: https://github.com/justinbeaty/magento-lts/commit/ab05de9369abc9e640986c5410aa72b40ff0fbf5

It works great if you `git clone` and `composer install` this repo, but it's not working if you use `aydin-hassan/magento-core-composer-installer`.

I think that we may need to build this functionality directly into `magento-hackathon/magento-composer-installer`.

sreichel:
> From what I can see, the JQuery project is still maintained. Maybe a solution should be found to add it as a dependency and not update it periodically in this repository.

https://github.com/components/jquery ?

justinbeaty:
> > From what I can see, the JQuery project is still maintained. Maybe a solution should be found to add it as a dependency and not update it periodically in this repository.
> 
> https://github.com/components/jquery ?

See my previous comment where I attempted to get this to work. Maybe you'd have better luck figuring it out though....

fballiano:
Ye but that repo doesnt seem to be updated with the 3.6.4

sreichel:
You wrote you'd not like to add `npm` .... but thats up2date :P

fballiano:
My personal preference is not to add more tools, jqury gets updated once every million years, I dont think we need npm just for that

sreichel:
> just for that

What about default scss compiler? Just looked at the `js` dir ... tinymce, jscolor, _some_calendar have packages. In some way its the same as composer - just for js/css libs.

justinbeaty:
> What about default scss compiler? Just looked at the `js` dir ... tinymce, jscolor, _some_calendar have packages. In some way its the same as composer - just for js/css libs.

It does get messy though with two separate package managers. What about if a magento module would like to include some JS? It can't interface with npm unless you make the user both `composer install` and `npm install` it.

Ultimately, I think the best way is to add support to `magento-hackathon/magento-composer-installer` for installing JS libs to the proper place.

fballiano:
yes ok but in the meanwhile I wouldn't keep jquery 1.12 here, also because we do not control the code of the  composer-installer

justinbeaty:
> yes ok but in the meanwhile I wouldn't keep jquery 1.12 here, also because we do not control the code of the composer-installer

For now, the manual update is just fine -- we've being doing it that way for a while now.

I'm for the change, but haven't tested the PR yet.

fballiano:
I've retested the RWD frontend and I see zero problems or warning after the update.

empiricompany (dismissed):

kiatng (dismissed):

fballiano:
updated to 3.7.0, did some tests and seem to work perfect on RWD

kiatng (approved):

ADDISON74 (approved):
I will approve it because it needs to be integrated in OpenMage:next, that's the only way we will be able to find out if errors occur. Keep the console open!

PR #3205, opened 2023-04-21
Fixed PHPStan error in Mage_HTTP_Client_Socket and Mage_HTTP_Client_Curl

fballiano:
I rewrote the `list($key, $val) = explode("=", $values[0]);`

But it seems to me that the whole Mage_HTTP* is never used, should we at least deprecate it? or remove it in `next`?


justinbeaty:
I guess PHPStan is a bit wrong here, because `$val` very well could have been null:

```php
list($key, $val) = explode('=', 'foo');

var_dump($key); // string(3) "foo"'
var_dump($val); // NULL
```

I'd be curious if it throws the error if you used `if ($val === null)`. Alternatively, you could use `if (!array_key_exists(1, $keyValueArray))` (since index 0 will always be set). Or even just `if (count($keyValueArray) >= 2)`.

---

It seems like some people might be using this library, see: #913. However, I support the removal of these classes (move it to an external repo, I suppose.) We should encourage people to use the power of Composer here, and just install something like [Guzzle](https://packagist.org/packages/guzzlehttp/guzzle) or one of the many other helper libraries. These built-in libraries like this are so old-school.

fballiano:
@justinbeaty I've removed the first unnecessary check. I wouldn't care too much about phpstan sincerely because it also is not perfect... anyway since nowadays if you don't do phpstand you're not cool... it bothers me to have a million errors.

what if we deprecate all these classes in lib/Mage, copy them to a separate repo and remove them from `next`?

justinbeaty:
> what if we deprecate all these classes in lib/Mage, copy them to a separate repo and remove them from `next`?

I think it's okay to deprecate / remove. It's an easy fix for someone to add back if they used it.

There's probably a lot more in that lib/Mage folder we can remove...

fballiano:
I've removed the whole lib/Mage and see no error browsing around...

justinbeaty:
I think `Mage_Exception` might be the only used class from there.

fballiano:
you're considering lib/Mage too I think, cause I can see only these occurrences:
Screenshot 2023-04-21 alle 21 08 24

which should be converted to Mage_Core_Exception I think

fballiano:
ps: same think could be for lib/Magento... more or less..

justinbeaty:
> you're considering lib/Mage too I think, cause I can see only these occurrences:
I meant that Mage_Exception was the only class defined in lib/Mage that was used outside of lib/Mage. We're saying the same thing.

To figure this out, I grepped for each of those classes in `app/` and found nothing. I also grepped for `mage/archive` etc, but I don't think they're loadable via getModel anyway.

I think we could rewrite Mage_Exception -> Mage_Core_Exceptoin too. I don't know why there's two, probably just legacy crap. If needed, we can even do this somewhere:

```php
if (!class_exists('Mage_Exception ')) {
  class Mage_Exception extends Exception
  {
  }
}
```

fballiano:
legacy crap 😂

fballiano:
I'd do all of these things in separate PRs anyway

justinbeaty (approved):

justinbeaty:
> I'd do all of these things in separate PRs anyway

Yeah, so then this PR is fine to me. approved.

sreichel:
> I guess PHPStan is a bit wrong here, because `$val` very well could have been null:
> 
> ```
> list($key, $val) = explode('=', 'foo');
> 
> var_dump($key); // string(3) "foo"'
> var_dump($val); // NULL
> ```

I guess phpstan "_thinks_" correct here.

The given example throws a notice on php7 (_Notice: Undefined offset: 1 in_ `list(....) = explode(...)`) and a warning for php8 (_Warning: Undefined array key 1_)

Suggestion:

```
list($key, $val) = array_pad(explode("=", $values[0]), 2, null);
```

https://phpstan.org/r/6cef421c-5e0e-411e-9782-3a1f17b8aa6d

sreichel:
LGTM, but do we need `!strlen($key)`?

fballiano:
mmm I thought it would be better to check the key too and not only the value, in case after the trim it becomes empty, dunno.

sreichel (approved):

PR #3206, opened 2023-04-21
Replaced Mage_Exception usages with Mage_Core_Exception

fballiano:
In https://github.com/OpenMage/magento-lts/pull/3205 we had a talk about the files in lib/Mage (which are mostly unused), the only one that's actually used in the core is the Mage_Exception.

This class is anyway used only in really few places, so it's probably worth replacing it with Mage_Core_Exception, which is the really used one.


justinbeaty (approved):
Looks good to me.

elidrissidev (approved):
LGTM

ADDISON74 (approved):

PR #3207, opened 2023-04-21
RWD: removed useless code

fballiano:
If I'm not mistaken this code doesn't precisely nothing:

```
    enquire.register('screen and (min-width: ' + (bp.medium + 1) + 'px)', {
        match: function () {
            $j('.menu-active').removeClass('menu-active');
            $j('.sub-menu-active').removeClass('sub-menu-active');
            $j('.skip-active').removeClass('skip-active');
        },
        unmatch: function () {
            $j('.menu-active').removeClass('menu-active');
            $j('.sub-menu-active').removeClass('sub-menu-active');
            $j('.skip-active').removeClass('skip-active');
        }
    });
```

since the match and unmatch code is exactly the same.

This PR removed this part of code.

### DIFF
--- a/skin/frontend/rwd/default/js/app.js
+++ b/skin/frontend/rwd/default/js/app.js
@@ -777,23 +777,6 @@ $j(document).ready(function () {
     });
 
     // ==============================================
-    // Enquire JS
-    // ==============================================
-
-    enquire.register('screen and (min-width: ' + (bp.medium + 1) + 'px)', {
-        match: function () {
-            $j('.menu-active').removeClass('menu-active');
-            $j('.sub-menu-active').removeClass('sub-menu-active');
-            $j('.skip-active').removeClass('skip-active');
-        },
-        unmatch: function () {
-            $j('.menu-active').removeClass('menu-active');
-            $j('.sub-menu-active').removeClass('sub-menu-active');
-            $j('.skip-active').removeClass('skip-active');
-        }
-    });
-
-    // ==============================================
     // UI Pattern - Media Switcher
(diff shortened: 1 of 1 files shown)


justinbeaty:
Away from my computer now. But I think it is used to close any active menus when switching screen sizes. IIRC, Magento duplicated some menu structure for mobile and regular sizes, so it could have been used to fix a glitch where the menu is stuck open.

Perhaps try opening the mini cart or user menu and switch screen size to see if you encounter a bug.

justinbeaty:
So it's not as bad as I remember, but there are bugs caused by the removal of that code.

To reproduce, go to mobile viewport and open the menu (category navigation). Then resize the viewport to non-mobile:

image

On non-mobile, `header-nav` and `header-search` should never act as skip-links like the user menu and minicart. But, on mobile they do, so when resizing without the enquire code, the `skip-active` class isn't removed and you'll see the grey background on the navigation element.

It's not super buggy, but I recall having done some customizations on an old site where not resetting the menus on viewport change caused more problems. So, I'd probably leave this code in, but in #3208 it could be made more clear, such as:

```js
resetMenuState = function (mq) {
  $j('.menu-active').removeClass('menu-active');
  $j('.sub-menu-active').removeClass('sub-menu-active');
  $j('.skip-active').removeClass('skip-active');
}
maxWidthMediumMediaQuery.addEventListener('change', resetMenuState );
```

fballiano:
thanks a lot @justinbeaty, this PR has to be closed, I'll fix the other one tho ;-)

PR #3208, opened 2023-04-21
RWD: removed enquire.js and converted to window.matchMedia

fballiano:
Enquire.js is an old js library, last updated 6 years ago:
https://github.com/WickyNilliams/enquire.js

It was used to run js code based on the size of the viewport.

This is not necessary anymore since window.matchMedia is now widely available:
https://caniuse.com/?search=window.matchMedia

This PR removed enquire.js and replace all the calls with matchMedia API calls. It also includes https://github.com/OpenMage/magento-lts/pull/3207

From my tests it works fine.


ADDISON74:
My only concern is related to existing extensions. If any of them use the libraries that we are removing, issues will arise. Here I can give the example of jscolor, which was removed at some point and brought back because the MageWorx Advanced Product Option is based on it.

fballiano:
Yes but some developer can readd the js or fix it in a different way 😊
La latest version of enquire is 6 years old and google hates too much javascript and old javascript libraries, we’ve to clean it up sometimes 😊

elidrissidev (commented):
Missing declaration keyword (`var`/`let`) for all the newly added functions.

fballiano:
@elidrissidev done!

fballiano:
@justinbeaty @empiricompany @Sdfendor would you mind checking this one out? thank you!

Sdfendor:
JS is not my greatest forte. While this looks good and harmless enough, I would give the others mentioned give the advantage here.

empiricompany (approved):

ADDISON74:
I have analyzed this PR and I agree with it, but it must be rebased to the main branch. I don't consider it will create problems, in the worst case I would keep the js file in the distribution for a while.

fballiano:
I don't know, because strictly speaking this is a breaking change although it is true that almost nobody will have problems with this. If moving the PR to `main` will make it go forward then let's change it.

fballiano:
what are your doubts about this PR?

Piralein (approved):

tmotyl:
looks good to me. but should this also be documented as breaking change to watch out when updating?

fballiano:
> looks good to me. but should this also be documented as breaking change to watch out when updating?

@tmotyl absolutely, I'll put it here: https://github.com/OpenMage/magento-lts/pull/3388 as soon as it's merged ;-) although I think we can't continue to add things to README, we've to find another way IMHO

fballiano:
@tmotyl I've added info to the README in this PR already.

tmotyl (approved):

fballiano:
thank you! merging per 1green+2gray!

PR #3209, opened 2023-04-22
Updated docker-compose to mysql8 and redis7

fballiano:
Since mysql8.0 and redis7 are supported, I think they could be enabled in the docker-compose


colinmollenhour:
I just tested and worked for me!

colinmollenhour (approved):

PR #3210, opened 2023-04-22
Removed PHPStan stubs

fballiano:
A stub is a configuration file for PHPStan that overrides a specific docblock which resides in the vendor folder (which we don't have control on).

We've 2 stubs for PHPStan but I thought it could be a good idea to try to remove them, so I adapted our code to respect the (in this case) ZF1 docblock.

It seems to be working correctly, PHPStan is happy and we removed 2 "overrides".


ADDISON74 (approved):

app/code/core/Mage/Eav/Model/Adminhtml/System/Config/Source/Inputtype/Validator.php
```diff
@@ -22,7 +22,7 @@
 class Mage_Eav_Model_Adminhtml_System_Config_Source_Inputtype_Validator extends Zend_Validate_InArray
 {
     /**
-     * @var array<string, string>|null
+     * @inheritdoc
```
sreichel, line 25: Remove property?

fballiano: mmm it would rewrite some more rewrite because the base class initialise the array with something in it, the openmage does does it completely differently and uses a different message.

luigifab (approved):
Not tested but looks good.

PR #3214, opened 2023-04-24
Converted msrp_rwd.js to vanilla javascript

fballiano:
This is a small conversion from prototypejs to vanilla javascript, it affects a minor file: msrp_rwd.js

### Manual testing scenarios
1. enable RWD theme
2. enable MSRP in configuration -> sales -> sales -> Minimum Advertised Price
3. also enable "Apply MAP (Default Value)"
4. flush caches
5. go to frontend, browse do a product page
6. price is hidden by default, if you click on these 2 links:
Screenshot 2023-04-24 alle 17 27 56
an overlay popup is shown with some info

### DIFF
--- a/skin/frontend/rwd/default/js/msrp_rwd.js
+++ b/skin/frontend/rwd/default/js/msrp_rwd.js
@@ -12,8 +12,8 @@
  */
 
 Catalog.Map.showHelp = Catalog.Map.showHelp.wrap(function (parent, event) {
-    var helpBox = $('map-popup');
-    var bodyNode = $$('body')[0];
+    var helpBox = document.getElementById('map-popup');
+    var bodyNode = document.getElementsByTagName('body')[0];
 
     // Resolve calculation bug in parent so we can actually use these classes...
     if (helpBox && this != Catalog.Map && Catalog.Map.active != this.link) {


ADDISON74 (approved):

skin/frontend/rwd/default/js/msrp_rwd.js
```diff
@@ -12,8 +12,8 @@
  */
 
 Catalog.Map.showHelp = Catalog.Map.showHelp.wrap(function (parent, event) {
-    var helpBox = $('map-popup');
-    var bodyNode = $$('body')[0];
+    var helpBox = document.getElementById('map-popup');
+    var bodyNode = document.getElementsByTagName('body')[0];
```
luigifab, line 16: `document.querySelector('body')`?

justinbeaty: Or just `document.body` should work too?

luigifab: There is warning with PhpStorm about `document.body`. I don't remember exactly which.

ADDISON74: I approved the PR because he replaced the $ and $$ exactly with the vanilla JavaScript equivalents, which are easy to understand in this case. Obviously other variants can be used, but I like ID and TagName more, especially because they are together. Obviously a personal opinion...

justinbeaty (approved):

skin/frontend/rwd/default/js/msrp_rwd.js
justinbeaty: Since it’s such a small change I approve any of the options. The straight replacement with JS equivalent would be just fine if we applied it broadly across the codebase, instead of trying to optimize with querySelector — which could be handled on a case by case basis.

fballiano: https://stackoverflow.com/questions/18247289/what-is-the-difference-between-queryselectorall-and-getelementsbytagname

in this case it seems to me there's absolutely no difference

luigifab (approved):

elidrissidev:
Since `helpBox` and `bodyNode` are now an `HTMLElement`, the calls below to add and remove a class should be changed to `element.classList.add()` and `element.classList.remove()` respectively.

https://github.com/OpenMage/magento-lts/blob/d3dcc7631786a9cca651eb4931dbc638f1b1a4b6/skin/frontend/rwd/default/js/msrp_rwd.js#L22-L31

fballiano:
I tested the helpbox when I did the PR and it was working, I'm checkin it anyway

fballiano:
right, removeClassName and addClassName are from prototypejs, they still work but I'll change them in a few secs

elidrissidev:
Ahh ok, so prototype is adding those methods to native `HTMLElement`s too.

PR #3222, opened 2023-04-27
Load all child blocks in Adminhtml Head block

elidrissidev:
### Description (*)

First roadblock I encountered when creating the extension mentioned in #3219, I can't add a custom child block to admin head 😢, and since the content is dynamically created through calling PHP code I can't use `addJs` either.

I don't know why this change wasn't done before, maybe it caused some issue? but it seems to work fine for me.

1. Go to layout xml of any module and try to add a child block to adminhtml head, in my case:
```xml
...
<default>
    <reference name="head">
        <block type="elidrissidev_debugbar/head" name="debugbar_head" />
    </reference>
</default>
...
```

Should be loaded after this change.

### DIFF
--- a/app/design/adminhtml/default/default/template/page/head.phtml
+++ b/app/design/adminhtml/default/default/template/page/head.phtml
@@ -50,5 +50,4 @@
 </script>
 
 <?php echo $this->helper('core/js')->getTranslatorScript() ?>
-<?php echo $this->getChildHtml('calendar'); ?>
-<?php echo $this->getChildHtml('optional_zip_countries'); ?>
+<?php echo $this->getChildHtml(); ?>


fballiano (approved):
since the getchildhtml outputs all of the children if no name is passed, it seems good to me.

did a brief test and worked fine

fballiano:
I wonder if there may be other blocks that were not printed before and that now get printed and that could cause problems.

elidrissidev:
The only block I see is `directory/js/optional_zip_countries.phtml` which is being added to the head in multiple places.

elidrissidev:
Do you think we should backport this to v19 for the next release?

fballiano:
it wouldn't cause any harm, absolutely, but I think we shouldn't anyway :-)

fballiano:
sure, but I think this PR still makes a lot of sense :-)

kyrena (approved):

ADDISON74 (approved):

luigifab:
Is there a general issue now?

do:
```php
<?php echo $this->getChildHtml('debugbar_head') ?>
<?php echo $this->getChildHtml() ?>
```
and the debugbar is displayed 2x

do:
```php
<?php echo $this->getChildHtml('debugbar_head') ?>
<?php echo $this->getChildHtml() ?>
<?php echo $this->getChildHtml() ?>
```
and the debugbar is displayed 3x

Or it's me? and it's already the time to travel again!

elidrissidev:
@luigifab I think that's the intended behaviour, `$this->getChildHtml()` will return HTML for all children, whereas `$this->getChildHtml('debugbar_head')` will return HTML of `debugbar_head` block only.

PR #3224, opened 2023-04-28
Add recent orders in order view

luigifab:
### Description

Add a new (ajax) tab in _Sales > Order > Any Orders of registered customers_.

_For customer view:_
customer

_For order view:_
orders

### Questions

Two things:
- is `escape` for grid columns useless?
- is `addFieldToSelect` for flat collections useless?

### Contribution checklist


fballiano:
I checked this PR but... I've to say I'm not sure about this, to access the order history for registered users you can click the `user name` and then you have the `orders` tab so I wouldn't add a tab `order history` in another order :-\

ADDISON74:
@fballiano - I agree with your comment. If @empiricompany will add in the future other grids in the Backend to be customized, then in the Orders grid we can add the Group column and thus the orders will be able to be filtered by registered or not registered buyers.

You can mark it with the [Don't forget this PR] label.

Issue #3228, opened 2023-05-03
GA4: purchase event wrong


pquerner:
PR #3092 introduced the usage of the purchase event trigger for GA4, however I believe its not fully functional.

### Preconditions (*)

1. 20.1.0-rc3 or 19.5.0-rc3 required (implementation of GA4)
2. GA4 must be configured in the backend and ready to use

### Steps to reproduce (*)

1. Configure GA4 and make a purchase
3. The purchase event should not be triggered
4. Reason is, that according to https://developers.google.com/analytics/devguides/collection/ga4/reference/events?client_type=gtag#purchase the `shipping`, `tax` and `value` and for the `item`, the `price` and `quantity` should be of number format.
5. It is however a string value (ie: according to `number_format`)

### Expected result (*)

1. The event purchase should be triggered

### Actual result (*)

1. The event is not triggered (due to inaccurate purchase event structure)

Fix would be to cast the values of `value`, `shipping` and `tax` via `floatval` to a float value.

https://github.com/OpenMage/magento-lts/blob/4e9520c39047f32155b0fa0b307ae69b8bae3cbd/app/code/core/Mage/GoogleAnalytics/Block/Ga.php#L212

and here

https://github.com/OpenMage/magento-lts/blob/4e9520c39047f32155b0fa0b307ae69b8bae3cbd/app/code/core/Mage/GoogleAnalytics/Block/Ga.php#L224-L226

(`int` cast for `quantity` and `floatval` for `price` and `discount`)

fballiano:
getBasePrice and getBaseDiscountAmount are already floats, while quantity could also be decimal so we can't cast it to int

pquerner:
In my testing the items `price` is a `string` value. Might be some wierd thing on our end then..

fballiano:
can you try https://github.com/OpenMage/magento-lts/pull/3231?

although I have the same kind of code on a live store and it works fine

fballiano:
mmmm it is true that $item->getBasePrice() does return a string instead of a float (although it's documented otherwise)...

pquerner:
You said you have this running on a client, can you confirm (if you have access to the GA account) that the conversions are empty? Or maybe Google is graceful (enough) that it does the typecasting for us?

Sadly I cannot test the PR on our site, since I am in the process of implementing it myself and all my ways of adding a "event purchase" so far have failed (atleast are not visible). Not sure whats going on, hence my first idea the typejuggling thingy here.
[...]

PR #3231, opened 2023-05-03
Fixed type casting for Analytics 4

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3228 it was pointed out that maybe some of the types in the Google Analytics 4 implementation of the "purchase" event may not be correct.

Mainly because the `number_format` function (required to return only a specific number of decimal) returns a string.

With this PR those values are casted back to float.

GA4 documentation: https://developers.google.com/analytics/devguides/collection/ga4/reference/events


fballiano:
at the moment I'll close and see how the original issue go.

pquerner (commented):
LGTM, although the items types are still wrong. Is this something you want to have in this PR also?

fballiano:
I don't trust casting 100%, since it makes 5.00 become 5 in the json output. I think google will like 5 less than "5.00", wait 24-48h and the data will be visible for sure.

pquerner:
Got a valid point there.

PR #3176, opened 2023-04-14
Revert "Deprecation errors are not suppressed anymore (#3102)"

sreichel:
### Description (*)

This reverts commit 400c85a55ccc4927b79b379ec19e93b8fea994c9.

`DEVELOPER_MODE` is turned on to see exceptions, but i really dont want to see "Deprecated functionality:" errors for common actions.

### Related Pull Requests

1. See OpenMage/magento-lts#3102

### Manual testing scenarios (*)

1. php 8.1
2. admin -> customer ->select a customer

```
Deprecated functionality: preg_match(): Passing null to parameter #2 ($subject) of type string is deprecated  in /var/www/html/app/design/adminhtml/default/default/template/widget/tabs.phtml on line 29

#0 [internal function]: mageCoreErrorHandler(8192, 'preg_match(): P...', '/var/www/html/a...', 29)
#1 /var/www/html/app/design/adminhtml/default/default/template/widget/tabs.phtml(29): preg_match('/\\s?ajax\\s?/', NULL)
#2 /var/www/html/app/code/core/Mage/Core/Block/Template.php(251): include('/var/www/html/a...')
#3 /var/www/html/app/code/core/Mage/Core/Block/Template.php(288): Mage_Core_Block_Template->fetchView('adminhtml/defau...')
```

### Questions or comments

Should be removed in `next` when php81+ is battle tested?

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

Issue #3175, opened 2023-04-14
Can't retrieve entity config (after delete extension)


sreichel:
### Preconditions (*)

1. `main` branch, works on `v19`

Bug comes from #2993

### Steps to reproduce (*)

1. install an extension that has custom eav attributes
2. uninstall, but leave orphan db entries
3. open admin panel

### Expected result (*)

1. works

### Actual result (*)

1.
```
Can't retrieve entity config: ave_sizechart/eav_attribute

#0 /var/www/html/app/code/core/Mage/Core/Model/Resource.php(280): Mage::throwException('Can't retrieve ...')
#1 /var/www/html/app/code/core/Mage/Core/Model/Resource/Db/Abstract.php(269): Mage_Core_Model_Resource->getTableName('ave_sizechart/e...')
#2 /var/www/html/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(480): Mage_Core_Model_Resource_Db_Abstract->getTable('ave_sizechart/e...')
#3 /var/www/html/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(565): Mage_Core_Model_Resource_Db_Collection_Abstract->getTable('ave_sizechart/e...')
#4 /var/www/html/app/code/core/Mage/Eav/Model/Resource/Entity/Attribute/Collection.php(112): Mage_Core_Model_Resource_Db_Collection_Abstract->join('ave_sizechart/e...', 'additional_tabl...')
#5 /var/www/html/app/code/core/Mage/Eav/Model/Entity/Type.php(110): Mage_Eav_Model_Resource_Entity_Attribute_Collection->setEntityTypeFilter(Object(Mage_Eav_Model_Entity_Type))
#6 /var/www/html/app/code/core/Mage/Eav/Model/Config.php(212): Mage_Eav_Model_Entity_Type->getAttributeCollection()
#7 /var/www/html/app/code/core/Mage/Eav/Model/Config.php(172): Mage_Eav_Model_Config->_loadEntityAttributes(Object(Mage_Eav_Model_Entity_Type), 0)
#8 /var/www/html/app/code/core/Mage/Eav/Model/Config.php(386): Mage_Eav_Model_Config->_initializeStore()
#9 /var/www/html/app/code/core/Mage/Eav/Model/Entity/Abstract.php(267): Mage_Eav_Model_Config->getEntityType('customer')
[...]

PR #2882, opened 2023-01-02
Added DEV_PHP_STRICT env variable

sreichel:
### Description (*)

When testing against PHP8.1 you have to comment some lines, mentioned in related PRs.

PR adds support for `DEV_PHP_STRICT` env variable to not change functions.php every time.

DDEV example:

```
web_environment:
- MAGE_IS_DEVELOPER_MODE=1
- DEV_PHP_STRICT=1
```

### Related Pull Requests

1. OpenMage/magento-lts#1812
2. OpenMage/magento-lts#2586

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3234, opened 2023-05-04
Revert "Deprecation errors are not suppressed anymore" #3102

sreichel:
### Description (*)

Please revert #3102!

With php 81. i _cant_ neither use `main` because of #3175 nor `v19` with enabled dev-mode because of deprecation errors.

### Related Pull Requests

1. See OpenMage/magento-lts#3176

1. use php8.1
2. enable developer mode
3. run into errors
4. ....

### DIFF
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -118,8 +118,8 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
     }
 
     // Suppress deprecation warnings on PHP 7.x
-    // set environment variable DEV_PHP_STRICT to 1 will show E_DEPRECATED errors
-    if ((!isset($_ENV['DEV_PHP_STRICT']) || $_ENV['DEV_PHP_STRICT'] != '1')
+    // set environment variable DEV_PHP_STRICT to 0 will hide E_DEPRECATED errors
+    if ((isset($_ENV['DEV_PHP_STRICT']) && $_ENV['DEV_PHP_STRICT'] == '0')
         && $errno == E_DEPRECATED
         && version_compare(PHP_VERSION, '7.0.0', '>=')
     ) {
--- a/app/code/core/Mage/Core/functions.php
+++ b/app/code/core/Mage/Core/functions.php
@@ -117,6 +117,15 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
         return false;
     }
 
+    // Suppress deprecation warnings on PHP 7.x
+    // set environment variable DEV_PHP_STRICT to 1 will show E_DEPRECATED errors
+    if ((!isset($_ENV['DEV_PHP_STRICT']) || $_ENV['DEV_PHP_STRICT'] != '1')
+        && $errno == E_DEPRECATED
+        && version_compare(PHP_VERSION, '7.0.0', '>=')
+    ) {
+        return true;
+    }
+
     // PEAR specific message handling
     if (stripos($errfile . $errstr, 'pear') !== false) {
         // ignore strict and deprecated notices
--- a/README.md
+++ b/README.md
@@ -330,7 +330,13 @@ grep -rn 'urn:Magento' --include \*.xml
 - Open your site in browser
   ```bash
   ddev launch
-  ``` 
+  ```
+
+## Development with PHP 8.1+
+
+Deprecation errors are supressed by default.
+
+If you want to work on PHP 8.1+ support, set environment variable `DEV_PHP_STRICT` to `1`, to show all errors.
 
 ## PhpStorm Factory Helper
 


ADDISON74:
I faced this problem. the application will not run, the modification must be done manually.

However, #3102 brought advantages that we cannot deny. That's how I discovered a lot of errors/warnings in 8.2 that we didn't know about and which are mostly fixed. Unfortunately, I am the only one who approved those PRs and we've been in this situation for a long time without incorporating them into the OpenMage code faster.

elidrissidev:
FYI there are 2 ways to prevent errors (deprecations or otherwise) from being thrown and blocking the page:

1. You just disable Developer Mode and all errors will be logged instead of thrown.
2. Update `error_reporting` directive in your `php.ini` (globally or via `.user.ini`) to only include the errors you're interested in.

Do these options not cover your case?

fballiano:
cleaning an orphan eav attribute should be a matter of seconds.

sreichel:
@elidrissidev disable developer mode is not an option. What was wrong with setting env `DEV_PHP_STRICT=1` to optionally show deprecation warnings? (#2882)

fballiano:
What was wrong was already explain in the other pr.

sreichel:
This?

> Cause it's better that people start to see this logs and fix their deprecation errors

Please let the users decide to show/hide deprecaction errors. With php8.1 there are still a lot, but i dont want to deal with them every time.

fballiano:
Exactly, nobody wants to deal with these prs every time

fballiano:
While we are trying to fix the few remainings of the 8.1 warnings

sreichel:
@fballiano the env var was added for those who activly work on php81+ support - without changing behavoir for others. 

I cant work on other things while running 8.1 .... :/

ADDISON74:
The PR description is not exactly the right one. This is not about deleting an orphaned attribute after removing an extension. That error is confirmed and we are still waiting for a solution from the one who made the recent changes in EAV.

The question that arises is why do we need another environment variable DEV_PHP_STRICT when we already have MAGE_IS_DEVELOPER_MODE? From the explanation given by @sreichel it appears that it is useful especially for the PHP8.1 version. The first variable D_P_S covers parts that the second variable M_I_D_M does not, practically they can work in parallel.

sreichel:
> The question that arises is why do we need another environment variable DEV_PHP_STRICT when we already have MAGE_IS_DEVELOPER_MODE?

DEV-mode controlls different thing, `DEV_PHP_STRICT ` was just to show deprecation errors optionally. (escpacially for 8.1+)

Updated PR to optionally hide them.

fballiano:
You could download this pr and apply it as a composer patch to your project

ADDISON74:
In such situations, a democratic vote would be the best solution. As far as I am concerned, I have nothing against this PR to be included in OpenMage. There were other similar cases where developers wanted to solve their particular situations and proposed PRs in this regard and we approved them. However, Sven is an important contributor to this project and his proposals over time have found their rightful place, beyond some personal disputes. Let's not easily forget the integration of ZF1-Future, PHPStan and many others.

sreichel:
> You could download this pr and apply it as a composer patch to your project

You can only patch files in vendor, not in project files ... no.

ADDISON74:
In the case of ZF1-Future, we outsourced the code by linking it to the developer's repository and it ended up in the vendor directory after which we applied some patches, Composer making the work possible.

This page that was approved recently https://www.openmage.org/2023/05/01/customize-your-openmage.html, it provides a solution for those who do not see their ideas accepted in OpenMage. Now it comes the question, in order to patch the OpenMage code, do we need in vendor an openmage/magento-lts directory? I don't think this aspect has been analyzed yet.

fballiano:
you can do it with zero problems:
https://github.com/fballiano/mcommerce/blob/main/composer.json

app/code/core/Mage/Core/functions.php
```diff
@@ -117,6 +117,15 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
         return false;
     }
 
+    // Suppress deprecation warnings on PHP 7.x
+    // set environment variable DEV_PHP_STRICT to 0 will hide E_DEPRECATED errors
+    if ((isset($_ENV['DEV_PHP_STRICT']) && $_ENV['DEV_PHP_STRICT'] == '0')
+        && $errno == E_DEPRECATED
+        && version_compare(PHP_VERSION, '7.0.0', '>=')
```
fballiano, line 124: at this point it would be better to mage developer mode = error reporting

fballiano:
> I hope we don't end up proposing patches to everyone who wants basic features.

was this ever done? never. so, in this case, it was done for a reason.

sreichel:
> you can do it with zero problems:

Locally i work with `git pull` and have no `openmage/magento-lts` dependency.

sreichel:
Can we just restore a working behavoir? (At least for `v19`)

- deprecation warnings were hidden all the time ... (b/c they do not harm)
- a ENV var was added to show/hide them
- everything got removed and everyone faces deprecation warnings for php81

I REALLY dont understand what we are discussing bout. 

Actually the best way to work arround it, is not to use php81 at all.

Flyingmana:
is there a problem with the mentioned "error_reporting" approach`? Do we overwrite the ini setting in OpenMage? Because then we should make this configurable for Dev Mode, as people might want different levels, sometimes without error reporting, sometimes even without notices.
Personally I did always hardcode changed error reporting somewhere 🤔

sreichel:
> as people might want different levels, sometimes without error reporting, sometimes even without notices.

Thats it, This change killed the possibilty to hide/show deprecation warnings ...

Flyingmana:
So I tested it out, and I could confirm the error_reporting is not overwritten from the dev mode, so setting in the php.ini `error_reporting = 24575` (which is E_ALL & ~E_DEPRECATED) prevents your reported issue.
Therefor I vote for closing this PR

sreichel:
PR was w/o adjusting php.ini ...

sreichel:
Thank for wasting my time.

PR #3238, opened 2023-05-05
Improved handling of ignore_user_agents

empiricompany:
Modified the code to skip request logging for all user agents that contain any substring of an ignored agent in the ignore_user_agents config. The previous code only matched the exact string of user_agent with the ignore_user_agents config using in_array function, which resulted in some requests not being skipped from logging. Now, the code uses stristr function to match the string partially and case-insensitive, which improves the functionality of ignore_user_agents.

### DIFF
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -105,11 +105,12 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
             $userAgent = $this->_httpHelper->getHttpUserAgent();
-            array_walk($ignoreAgents, function ($item) use ($userAgent) {
-                if (stristr($userAgent, $item)) {
+            foreach ($ignoreAgents as $ignoreAgent) {
+                if (stripos($userAgent, $ignoreAgent) !== false) {
                     $this->_skipRequestLogging = true;
+                    break;
                 }
-            });
+            })
         }
         if ($this->_logCondition->isLogDisabled()) {
             $this->_skipRequestLogging = true;
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -105,7 +105,7 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
         $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
-            if (in_array($userAgent, $ignoreAgents)) {
+            if (in_array($userAgent, $ignoreAgents) || stristr($userAgent, $ignoreAgents)) {
                 $this->_skipRequestLogging = true;
             }
         }
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -101,13 +101,15 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
     protected function _construct()
     {
         $this->_init('log/visitor');
-        $userAgent = $this->_httpHelper->getHttpUserAgent();
         $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
-            if (in_array($userAgent, $ignoreAgents) || stristr($userAgent, $ignoreAgents)) {
-                $this->_skipRequestLogging = true;
-            }
+            $userAgent = $this->_httpHelper->getHttpUserAgent();
+            array_walk($ignoreAgents, function ($item) use ($userAgent) {
+                if (stristr($userAgent, $item)) {
+                    $this->_skipRequestLogging = true;
+                }
+            });
         }
         if ($this->_logCondition->isLogDisabled()) {
             $this->_skipRequestLogging = true;
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -110,11 +110,17 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
             $userAgent = $this->_httpHelper->getHttpUserAgent();
-            foreach ($ignoreAgents as $ignoreAgent) {
-                if (stripos($userAgent, $ignoreAgent) !== false) {
+            try {
+                $ignoreAgents = array_map(function ($userAgent) {
+                    return preg_quote($userAgent, '/');
+                }, $ignoreAgents);
+                $ignoreAgentRegex = implode('|', $ignoreAgents);
+                
+                if (preg_match("/{$ignoreAgentRegex}/i", $userAgent)) {
                     $this->_skipRequestLogging = true;
-                    break;
                 }
+            } catch (Exception $e) {
+                Mage::logException($e);
             }
         }
     }
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -110,17 +110,11 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
             $userAgent = $this->_httpHelper->getHttpUserAgent();
-            try {
-                $ignoreAgents = array_map(function ($userAgent) {
-                    return preg_quote($userAgent, '/');
-                }, $ignoreAgents);
-                $ignoreAgentRegex = implode('|', $ignoreAgents);
-                
-                if (preg_match("/{$ignoreAgentRegex}/i", $userAgent)) {
+            foreach ($ignoreAgents as $ignoreAgent) {
+                if (stripos($userAgent, $ignoreAgent) !== false) {
                     $this->_skipRequestLogging = true;
+                    break;
                 }
-            } catch (Exception $e) {
-                Mage::logException($e);
             }
         }
     }
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -101,6 +101,11 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
     protected function _construct()
     {
         $this->_init('log/visitor');
+        if ($this->_logCondition->isLogDisabled()) {
+            $this->_skipRequestLogging = true;
+            return;
+        }
+
         $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
@@ -110,10 +115,7 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
                     $this->_skipRequestLogging = true;
                     break;
                 }
-            })
-        }
-        if ($this->_logCondition->isLogDisabled()) {
-            $this->_skipRequestLogging = true;
+            }
         }
     }
 
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -46,7 +46,7 @@
             <semrushbot>SemrushBot</semrushbot>
             <alphabot>AlphaBot</alphabot>
             <facebookexternalhit>facebookexternalhit</facebookexternalhit>
-            <seotester2>SEOTesterBot/seotester2</seotester2>
+            <seotester2>SEOTesterBot</seotester2>
             <ZoominfoBot>ZoominfoBot</ZoominfoBot>
             <amazonbot>Amazonbot</amazonbot>
         </ignore_user_agents>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -40,7 +40,6 @@
             <mod_pagespeed>mod_pagespeed</mod_pagespeed>
             <mj12bot>MJ12bot</mj12bot>
             <yandexbot>YandexMobileBot</yandexbot>
-            <yandexmobile>YandexMobileBot</yandexmobile>
             <yandeximage>YandexImages</yandeximage>
             <opensiteexplorer>spbot</opensiteexplorer>
             <semrushbot>SemrushBot</semrushbot>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -44,7 +44,7 @@
             <semrushbot>SemrushBot</semrushbot>
             <alphabot>AlphaBot</alphabot>
             <facebookexternalhit>facebookexternalhit</facebookexternalhit>
-            <seotester2>SEOTesterBot</seotester2>
+            <seotester>SEOTesterBot</seotester>
             <ZoominfoBot>ZoominfoBot</ZoominfoBot>
             <amazonbot>Amazonbot</amazonbot>
         </ignore_user_agents>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -29,9 +29,26 @@
             </entities>
         </ignoredModules>
         <ignore_user_agents>
-            <google1>Googlebot/1.0 (googlebot@googlebot.com http://googlebot.com/)</google1>
-            <google2>Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)</google2>
-            <google3>Googlebot/2.1 (+http://www.googlebot.com/bot.html)</google3>
+            <google1>Googlebot</google1>
+            <googleimage>Googlebot-Image</googleimage>
+            <mauibot>MauiBot</mauibot>
+            <blexbot>BLEXBot</blexbot>
+            <hrefsbot>AhrefsBot</hrefsbot>
+            <bingbot>bingbot</bingbot>
+            <dotbot>DotBot</dotbot>
+            <istellabot>istellabot</istellabot>
+            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
+            <mj12bot>MJ12bot</mj12bot>
+            <yandexbot>YandexMobileBot</yandexbot>
+            <yandexmobile>YandexMobileBot</yandexmobile>
+            <yandeximage>YandexImages</yandeximage>
+            <opensiteexplorer>spbot</opensiteexplorer>
+            <semrushbot>SemrushBot</semrushbot>
+            <alphabot>AlphaBot</alphabot>
+            <facebookexternalhit>facebookexternalhit</facebookexternalhit>
+            <seotester2>SEOTesterBot/seotester2</seotester2>
+            <ZoominfoBot>ZoominfoBot</ZoominfoBot>
+            <amazonbot>Amazonbot</amazonbot>
         </ignore_user_agents>
         <helpers>
             <log>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -30,7 +30,6 @@
         </ignoredModules>
         <ignore_user_agents>
             <google1>Googlebot</google1>
-            <googleimage>Googlebot-Image</googleimage>
             <mauibot>MauiBot</mauibot>
             <blexbot>BLEXBot</blexbot>
             <hrefsbot>AhrefsBot</hrefsbot>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -29,7 +29,7 @@
             </entities>
         </ignoredModules>
         <ignore_user_agents>
-            <google1>Googlebot</google1>
+            <google>Googlebot</google>
             <mauibot>MauiBot</mauibot>
             <blexbot>BLEXBot</blexbot>
             <hrefsbot>AhrefsBot</hrefsbot>
--- a/app/code/core/Mage/Log/etc/config.xml
+++ b/app/code/core/Mage/Log/etc/config.xml
@@ -29,24 +29,28 @@
             </entities>
         </ignoredModules>
         <ignore_user_agents>
-            <google>Googlebot</google>
-            <mauibot>MauiBot</mauibot>
-            <blexbot>BLEXBot</blexbot>
-            <hrefsbot>AhrefsBot</hrefsbot>
-            <bingbot>bingbot</bingbot>
-            <dotbot>DotBot</dotbot>
-            <istellabot>istellabot</istellabot>
-            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
-            <mj12bot>MJ12bot</mj12bot>
-            <yandexbot>YandexMobileBot</yandexbot>
-            <yandeximage>YandexImages</yandeximage>
-            <opensiteexplorer>spbot</opensiteexplorer>
-            <semrushbot>SemrushBot</semrushbot>
+            <adsbot>AdsBot</adsbot>
             <alphabot>AlphaBot</alphabot>
-            <facebookexternalhit>facebookexternalhit</facebookexternalhit>
-            <seotester>SEOTesterBot</seotester>
-            <ZoominfoBot>ZoominfoBot</ZoominfoBot>
             <amazonbot>Amazonbot</amazonbot>
+            <bingbot>bingbot</bingbot>
+            <blexbot>BLEXBot</blexbot>
+            <dotbot>DotBot</dotbot>
+            <facebookexternalhit>facebookexternalhit</facebookexternalhit>
+            <google>Googlebot</google>
+            <google-site-verification>Google-Site-Verification</google-site-verification>
+            <hrefsbot>AhrefsBot</hrefsbot>
+            <istellabot>istellabot</istellabot>
+            <mauibot>MauiBot</mauibot>
+            <mj12bot>MJ12bot</mj12bot>
+            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
+            <opensiteexplorer>spbot</opensiteexplorer>
+            <pinterestbot>Pinterestbot</pinterestbot>
+            <semrushbot>SemrushBot</semrushbot>
+            <seotester>SEOTesterBot</seotester>
+            <yandexbot>YandexBot</yandexbot>
+            <yandeximage>YandexImages</yandeximage>
+            <yandexmobilebot>YandexMobileBot</yandexmobilebot>
+            <zoominfobot>ZoominfoBot</zoominfobot>
         </ignore_user_agents>
         <helpers>
             <log>


empiricompany:
We could also improve the config by including a new bot and optimizing the strings to ignore the bot versions and other information.

app/code/core/Mage/Log/etc/config.xml
`<config>

    <global>

        <ignore_user_agents>
            <google1>Googlebot</google1>
            <amazon>Amazonbot</amazon>
        </ignore_user_agents>

`

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -105,7 +105,7 @@ protected function _construct()
         $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
-            if (in_array($userAgent, $ignoreAgents)) {
+            if (in_array($userAgent, $ignoreAgents) || stristr($userAgent, $ignoreAgents)) {
```
fballiano, line 108: does this work for you? because $ignoreAgents is an array but the second parameter of `stristr` has to be a string. and shouldn't the first argument be the haystack?

empiricompany: Yes, I apologize for rushing to propose a solution without testing it. I have updated the code by removing the exact match search in the array and replacing it with a string search.

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,26 @@
             </entities>
         </ignoredModules>
         <ignore_user_agents>
-            <google1>Googlebot/1.0 (googlebot@googlebot.com http://googlebot.com/)</google1>
-            <google2>Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)</google2>
-            <google3>Googlebot/2.1 (+http://www.googlebot.com/bot.html)</google3>
+            <google1>Googlebot</google1>
+            <googleimage>Googlebot-Image</googleimage>
```
fballiano, line 33:
```suggestion
```
this is not necessary since googlebot will be matched

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,26 @@
...
+            <mauibot>MauiBot</mauibot>
+            <blexbot>BLEXBot</blexbot>
+            <hrefsbot>AhrefsBot</hrefsbot>
+            <bingbot>bingbot</bingbot>
+            <dotbot>DotBot</dotbot>
+            <istellabot>istellabot</istellabot>
+            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
+            <mj12bot>MJ12bot</mj12bot>
+            <yandexbot>YandexMobileBot</yandexbot>
+            <yandexmobile>YandexMobileBot</yandexmobile>
```
fballiano, line 43:
```suggestion
            <yandexbot>YandexMobileBot</yandexbot>
```

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,26 @@
             </entities>
         </ignoredModules>
         <ignore_user_agents>
-            <google1>Googlebot/1.0 (googlebot@googlebot.com http://googlebot.com/)</google1>
-            <google2>Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)</google2>
-            <google3>Googlebot/2.1 (+http://www.googlebot.com/bot.html)</google3>
+            <google1>Googlebot</google1>
```
fballiano, line 32:
```suggestion
            <google>Googlebot</google>
```

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,26 @@
...
+            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
+            <mj12bot>MJ12bot</mj12bot>
+            <yandexbot>YandexMobileBot</yandexbot>
+            <yandexmobile>YandexMobileBot</yandexmobile>
+            <yandeximage>YandexImages</yandeximage>
+            <opensiteexplorer>spbot</opensiteexplorer>
+            <semrushbot>SemrushBot</semrushbot>
+            <alphabot>AlphaBot</alphabot>
+            <facebookexternalhit>facebookexternalhit</facebookexternalhit>
+            <seotester2>SEOTesterBot</seotester2>
```
fballiano, line 49:
```suggestion
            <seotester>SEOTesterBot</seotester>
```

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -101,13 +101,15 @@ public function __construct(array $data = [])
...
             $ignoreAgents = $ignoreAgents->asArray();
-            if (in_array($userAgent, $ignoreAgents)) {
-                $this->_skipRequestLogging = true;
-            }
+            $userAgent = $this->_httpHelper->getHttpUserAgent();
+            array_walk($ignoreAgents, function ($item) use ($userAgent) {
+                if (stristr($userAgent, $item)) {
+                    $this->_skipRequestLogging = true;
+                }
+            });
```
fballiano, line 112:
```suggestion
            foreach ($ignoreAgents as $ignoreAgent) {
                if (stristr($userAgent, $ignoreAgent)) {
                    $this->_skipRequestLogging = true;
                    break;
                }
            })
```

sreichel: Or `stripos($userAgent, $ignoreAgent) !== false`?

sreichel: Just an idea ... when log is disabled, there is no need to check user-agent ...

```php
    /**
     * Object initialization
     */
    protected function _construct()
    {
        $this->_init('log/visitor');
        if ($this->_logCondition->isLogDisabled() || $this->isIgnoredUserAgent()) {
            $this->_skipRequestLogging = true;
        }
    }

    /**
     * @return bool
     */
    protected function isIgnoredUserAgent(): bool
    {
        $ignoreAgents = $this->_config->getNode('global/ignore_user_agents');
        if ($ignoreAgents) {
            $userAgent = $this->_httpHelper->getHttpUserAgent();
            foreach ($ignoreAgents->asArray() as $ignoreAgent) {
                if (stripos($userAgent, $ignoreAgent) !== false) {
                    return true;
                }
            }
        }

        return false;
    }
```

fballiano (dismissed):

app/code/core/Mage/Log/Model/Visitor.php
sreichel: @fballiano are you concerned about the new method?

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,24 @@
...
+            <mj12bot>MJ12bot</mj12bot>
+            <yandexbot>YandexMobileBot</yandexbot>
+            <yandeximage>YandexImages</yandeximage>
+            <opensiteexplorer>spbot</opensiteexplorer>
+            <semrushbot>SemrushBot</semrushbot>
+            <alphabot>AlphaBot</alphabot>
+            <facebookexternalhit>facebookexternalhit</facebookexternalhit>
+            <seotester>SEOTesterBot</seotester>
+            <ZoominfoBot>ZoominfoBot</ZoominfoBot>
+            <amazonbot>Amazonbot</amazonbot>
```
kiatng, line 49: Where do you get the list? Is it better to order it alphabetically?

sreichel: > Where do you get the list?

Also searched for something like this. Maybe adopt https://github.com/fabiomb/is_bot?

kiatng: Also, is it a good idea to ignore **every** crawlers/bots, or **only** the popular ones? I do not know how many bots are out there, could be millions. I tend to think it's better to ignore popular bots, and log the rest. The log data could show anomaly patterns that the site owner can take certain actions. Does it make sense?

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -101,17 +101,22 @@ public function __construct(array $data = [])
...
         if ($ignoreAgents) {
             $ignoreAgents = $ignoreAgents->asArray();
-            if (in_array($userAgent, $ignoreAgents)) {
-                $this->_skipRequestLogging = true;
+            $userAgent = $this->_httpHelper->getHttpUserAgent();
+            foreach ($ignoreAgents as $ignoreAgent) {
+                if (stripos($userAgent, $ignoreAgent) !== false) {
+                    $this->_skipRequestLogging = true;
+                    break;
+                }
```
kiatng, line 117: The longer the list, the slower it gets. I am weary.

app/code/core/Mage/Log/etc/config.xml
empiricompany: > Where do you get the list? Is it better to order it alphabetically?

It's a list that I have built over time, the list of bots can be endless and never definitive. It could also be added to the configuration.

app/code/core/Mage/Log/Model/Visitor.php
empiricompany: Can a regexp help to improve performance?

empiricompany:
Do you think it would make sense to add a 4° option 'Logged In Only' to activate logs only for logged-in customers?

app/code/core/Mage/Log/Model/Visitor.php
kiatng: regex is probably worst. I don't have any idea how to do this faster.

app/code/core/Mage/Log/etc/config.xml
kiatng: I think Sven's finding on `is_bot` is a good suggestion. Community maintained list is better.

kiatng: See also https://github.com/JayBizzle/Crawler-Detect

fballiano: the is_bot does it in the way of this PR, but has a big list and before in the comments we were saying "if the list gets big..."

is_bot is GPL2 which is not compatible with OSL/AFL

Crawler-Detect works with regex and the list is very big

maybe we could have a long string instead of an array, to make things simple?

app/code/core/Mage/Log/Model/Visitor.php
sreichel: > regex is probably worst. I don't have any idea how to do this faster.

Also expected regex to be slower, but it doesn't seem so.

```php
$ignoreAgents = range(0, 500000);
$ignoreAgents = array_map('strval', $ignoreAgents);
$ignoreAgents[] = 'google';
$userAgent = 'Googl'; #test: google, Google, Googl

function isIgnoredUnserAgent(string $userAgent, array $ignoreAgents): bool
{
    foreach ($ignoreAgents as $ignoreAgent) {
        if (stripos($userAgent, $ignoreAgent) !== false) {
            return true;
        }
    }
    return false;
}

function isIgnoredUnserAgentRegex(string $userAgent, array $ignoreAgents): bool
{
    $ignoreAgent = implode(' ', $ignoreAgents);
    return preg_match("/\b{$userAgent}\b/i", $ignoreAgent) === 1;
}

var_dump(isIgnoredUnserAgent($userAgent, $ignoreAgents));
var_dump(isIgnoredUnserAgentRegex($userAgent, $ignoreAgents));
```

## isIgnoredUnserAgent
<table rules="rows" width="90%" cellspacing="1" cellpadding="2" bordercolor="#bdc7d8" border="1" align="center">
<tbody><tr bgcolor="#bdc7d8" align="right"><th align="left"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=fn">Function Name</a></nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=ct">Calls</a></nobr></th><th class="vwbar"><nobr>Calls%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=wt">Incl. Wall Time<br>(microsec)</a></nobr></th><th class="vwbar"><nobr>IWall%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=cpu">Incl. CPU<br>(microsecs)</a></nobr></th><th class="vwbar"><nobr>ICpu%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=mu">Incl.<br>MemUse<br>(bytes)</a></nobr></th><th class="vwbar"><nobr>IMemUse%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgent&amp;source=ddev&amp;sort=pmu">Incl.<br> PeakMemUse<br>(bytes)</a></nobr></th><th class="vwbar"><nobr>IPeakMemUse%</nobr></th></tr><tr bgcolor="#e0e0ff"><td><b><i><center>Current Function</center></i></b></td></tr><tr><td><a href="">isIgnoredUnserAgent</a></td><td class="vbar">1</td>
<td class="vbar">0.0%</td>
<td class="vbbar">452,966</td>
<td class="vbbar">28.0%</td>
<td class="vbar">450,998</td>
<td class="vbar">28.1%</td>
<td class="vbar">1,120</td>
<td class="vbar">0.0%</td>
<td class="vbar">0</td>
<td class="vbar">0.0%</td>
</tr><tr bgcolor="#ffffff"><td style="text-align:right;color:blue">Exclusive Metrics  for Current Function</td><td class="vbar"></td><td class="vbar"></td><td type="Child" metric="wt" class="vbbar">281,408</td>
<td type="Child" metric="wt" class="vbbar">62.1%</td>
<td type="Child" metric="cpu" class="vbar">214,517</td>
<td type="Child" metric="cpu" class="vbar">47.6%</td>
<td type="Child" metric="mu" class="vbar">592</td>
<td type="Child" metric="mu" class="vbar">52.9%</td>
<td type="Child" metric="pmu" class="vbar">0</td>
<td type="Child" metric="pmu" class="vbar">N/A%</td>
</tr><tr bgcolor="#e0e0ff"><td><b><i><center>Parent function</center></i></b></td></tr><tr><td><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=main%28%29&amp;source=ddev">main()</a></td><td type="Parent" metric="ct" class="vbar">1</td>
<td type="Parent" metric="ct" class="vbar">100.0%</td>
<td type="Parent" metric="wt" class="vbbar">452,966</td>
<td type="Parent" metric="wt" class="vbbar">100.0%</td>
<td type="Parent" metric="cpu" class="vbar">450,998</td>
<td type="Parent" metric="cpu" class="vbar">100.0%</td>
<td type="Parent" metric="mu" class="vbar">1,120</td>
<td type="Parent" metric="mu" class="vbar">100.0%</td>
<td type="Parent" metric="pmu" class="vbar">0</td>
<td type="Parent" metric="pmu" class="vbar">N/A%</td>
</tr><tr bgcolor="#e0e0ff"><td><b><i><center>Child function</center></i></b></td></tr><tr><td><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=stripos&amp;source=ddev">stripos</a></td><td type="Child" metric="ct" class="vbar">500,002</td>
<td type="Child" metric="ct" class="vbar">100.0%</td>
<td type="Child" metric="wt" class="vbbar">171,558</td>
<td type="Child" metric="wt" class="vbbar">37.9%</td>
<td type="Child" metric="cpu" class="vbar">236,481</td>
<td type="Child" metric="cpu" class="vbar">52.4%</td>
<td type="Child" metric="mu" class="vbar">528</td>
<td type="Child" metric="mu" class="vbar">47.1%</td>
<td type="Child" metric="pmu" class="vbar">0</td>
<td type="Child" metric="pmu" class="vbar">N/A%</td>
</tr></tbody></table>

## isIgnoredUnserAgentRegex
<table rules="rows" width="90%" cellspacing="1" cellpadding="2" bordercolor="#bdc7d8" border="1" align="center">
<tbody><tr bgcolor="#bdc7d8" align="right"><th align="left"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=fn">Function Name</a></nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=ct">Calls</a></nobr></th><th class="vwbar"><nobr>Calls%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=wt">Incl. Wall Time<br>(microsec)</a></nobr></th><th class="vwbar"><nobr>IWall%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=cpu">Incl. CPU<br>(microsecs)</a></nobr></th><th class="vwbar"><nobr>ICpu%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=mu">Incl.<br>MemUse<br>(bytes)</a></nobr></th><th class="vwbar"><nobr>IMemUse%</nobr></th><th class="vwbar"><nobr><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=isIgnoredUnserAgentRegex&amp;source=ddev&amp;sort=pmu">Incl.<br> PeakMemUse<br>(bytes)</a></nobr></th><th class="vwbar"><nobr>IPeakMemUse%</nobr></th></tr><tr bgcolor="#e0e0ff"><td><b><i><center>Current Function</center></i></b></td></tr><tr><td><a href="">isIgnoredUnserAgentRegex</a></td><td class="vbar">1</td>
<td class="vbar">3.2%</td>
<td class="vbbar">9,859</td>
<td class="vbbar">0.6%</td>
<td class="vbar">8,006</td>
<td class="vbar">0.5%</td>
<td class="vbar">3,393,248</td>
<td class="vbar">10.1%</td>
<td class="vbar">0</td>
<td class="vbar">0.0%</td>
</tr><tr bgcolor="#ffffff"><td style="text-align:right;color:blue">Exclusive Metrics  for Current Function</td><td class="vbar"></td><td class="vbar"></td><td type="Child" metric="wt" class="vbbar">24</td>
<td type="Child" metric="wt" class="vbbar">0.2%</td>
<td type="Child" metric="cpu" class="vbar">0</td>
<td type="Child" metric="cpu" class="vbar">0.0%</td>
<td type="Child" metric="mu" class="vbar">664</td>
<td type="Child" metric="mu" class="vbar">0.0%</td>
<td type="Child" metric="pmu" class="vbar">0</td>
<td type="Child" metric="pmu" class="vbar">N/A%</td>
</tr><tr bgcolor="#e0e0ff"><td><b><i><center>Parent function</center></i></b></td></tr><tr><td><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=main%28%29&amp;source=ddev">main()</a></td><td type="Parent" metric="ct" class="vbar">1</td>
<td type="Parent" metric="ct" class="vbar">100.0%</td>
<td type="Parent" metric="wt" class="vbbar">9,859</td>
<td type="Parent" metric="wt" class="vbbar">100.0%</td>
<td type="Parent" metric="cpu" class="vbar">8,006</td>
<td type="Parent" metric="cpu" class="vbar">100.0%</td>
<td type="Parent" metric="mu" class="vbar">3,393,248</td>
<td type="Parent" metric="mu" class="vbar">100.0%</td>
<td type="Parent" metric="pmu" class="vbar">0</td>
<td type="Parent" metric="pmu" class="vbar">N/A%</td>
</tr><tr bgcolor="#e0e0ff"><td><b><i><center>Child functions</center></i></b></td></tr><tr><td><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=implode&amp;source=ddev">implode</a></td><td type="Child" metric="ct" class="vbar">1</td>
<td type="Child" metric="ct" class="vbar">50.0%</td>
<td type="Child" metric="wt" class="vbbar">9,510</td>
<td type="Child" metric="wt" class="vbbar">96.5%</td>
<td type="Child" metric="cpu" class="vbar">8,006</td>
<td type="Child" metric="cpu" class="vbar">100.0%</td>
<td type="Child" metric="mu" class="vbar">3,392,048</td>
<td type="Child" metric="mu" class="vbar">100.0%</td>
<td type="Child" metric="pmu" class="vbar">0</td>
<td type="Child" metric="pmu" class="vbar">N/A%</td>
</tr><tr bgcolor="#e5e5e5"><td><a href="/xhprof/index.php?run=645a42885d6a7&amp;symbol=preg_match&amp;source=ddev">preg_match</a></td><td type="Child" metric="ct" class="vbar">1</td>
<td type="Child" metric="ct" class="vbar">50.0%</td>
<td type="Child" metric="wt" class="vbbar">325</td>
<td type="Child" metric="wt" class="vbbar">3.3%</td>
<td type="Child" metric="cpu" class="vbar">0</td>
<td type="Child" metric="cpu" class="vbar">0.0%</td>
<td type="Child" metric="mu" class="vbar">536</td>
<td type="Child" metric="mu" class="vbar">0.0%</td>
<td type="Child" metric="pmu" class="vbar">0</td>
<td type="Child" metric="pmu" class="vbar">N/A%</td>
</tr></tbody></table>

app/code/core/Mage/Log/etc/config.xml
empiricompany: we can obtain a string for a regexp like "/google|bing|amazon/" imploding ignore agents array like:
```
$ignoreAgents = $ignoreAgents->asArray();
$ignoreAgents = implode("|", $ignoreAgents);
$isBot = (bool) preg_match("/{$ignoreAgents}/i", $userAgent);
```
we are sure regexp is slower then a foreach over stripos? maybe because exit on the first match?

Anyway, just searching for the main bots string would already make a big difference, and we have the same problem with sessions, which is more serious because sessions initiated by unidentified bots grow rapidly. In fact, I wanted to open a separate PR for it.

sreichel: Made some tests, but using `preg_match` changes behavoir. :/

empiricompany: What do you mean by changes behavior? Are you referring to performance? Here's an example regex that I could build
https://regex101.com/r/kIOxuL/1

fballiano: it will work since there will be no "google" in the string but googlebot, at least the list that we have in this PR would work for sure

empiricompany: I can't understand works exactly like stripos for me, regarding performance I don't know which method is fastest

sreichel: User-agent can be anything ... just name your user-agent like an ignored bot and and an suffix to name.

E.g. "googlebot" is ignored. You user-agent is called "testgooglebot".  `stripos` would ignore it, `preg_match` does not.

empiricompany: I'm sorry, you'll have to forgive me, but I still don't understand 😕


Here's a test I did to be 100% sure: https://reqbin.com/odyy5npi.
```
<?php
$userAgent = 'testgooglebot';
$ignoreAgents[] = 'googlebot';
$ignoreAgents[] = 'bing';

foreach($ignoreAgents as $ignoreAgent) {
	if (stripos($userAgent, $ignoreAgent) !== false) {
		echo 'matched by stripos';
		break;
	}
}
echo PHP_EOL;
$ignoreAgentRegex = implode('|', $ignoreAgents);
if(preg_match("/{$ignoreAgentRegex}/i", $userAgent) !== false){
  echo 'matched by regexp';
}
```

As you can see, if the user agent is called testgooglebot both match it, also stripos not ignore it.
There will never be a real user agent with the string googlebot in it and if preg_match is more efficient, I would keep that

empiricompany:
I got a bit confused, sorry. I did some tests and found that stripos is slightly more performant than preg_match. Now, it seems OK

fballiano (approved):

fballiano:
Let’s go, thanks @empiricompany

sreichel:
> I got a bit confused, sorry. I did some tests and found that stripos is slightly more performant than preg_match. Now, it seems OK

Can you please share test results? With my tests `preg_match` was faster on large arrays.

I thinks its okay (but i still would move that to a new method to reduce complexity in `_construct()`)?

ADDISON74:
See an updated list here https://github.com/mitchellkrogza/nginx-ultimate-bad-bot-blocker. In addition, the number is much higher.

Besides an issue of decreased performance, the advantages of this PR must be carefully analyzed. Everyone should keep their own list and keep it as short as possible. I use .htaccess not to ignore bots but to block them.

app/code/core/Mage/Log/etc/config.xml
```diff
@@ -29,9 +29,28 @@
...
+            <mj12bot>MJ12bot</mj12bot>
+            <mod_pagespeed>mod_pagespeed</mod_pagespeed>
+            <opensiteexplorer>spbot</opensiteexplorer>
+            <pinterestbot>Pinterestbot</pinterestbot>
+            <semrushbot>SemrushBot</semrushbot>
+            <seotester>SEOTesterBot</seotester>
+            <yandexbot>YandexBot</yandexbot>
+            <yandeximage>YandexImages</yandeximage>
+            <yandexmobilebot>YandexMobileBot</yandexmobilebot>
+            <zoominfobot>ZoominfoBot</zoominfobot>
```
ADDISON74, line 53: My opinion is that the list should be as short as possible, limited to the most important bots. Some have nothing to look for there. Everyone can configure their own list.

empiricompany:
> > I got a bit confused, sorry. I did some tests and found that stripos is slightly more performant than preg_match. Now, it seems OK
> 
> Can you please share test results? With my tests `preg_match` was faster on large arrays.
> 
> I thinks its okay (but i still would move that to a new method to reduce complexity in `_construct()`)?



> > I got a bit confused, sorry. I did some tests and found that stripos is slightly more performant than preg_match. Now, it seems OK
> 
> Can you please share test results? With my tests `preg_match` was faster on large arrays.
> 
> I thinks its okay (but i still would move that to a new method to reduce complexity in `_construct()`)?

Besides having read on various forums that it's more performant, I did some simple tests with microtime. I didn't have time to do more complex tests on resource usage.

The result of the tests is a difference of just a few microseconds on a large array taken from here 
https://github.com/fabiomb/is_bot/blob/master/is_bot.php:

2023-05-13T09:07:38+00:00 DEBUG (7): Benchmark preg_match 0.0002
2023-05-13T09:07:38+00:00 DEBUG (7): Benchmark foreach>stripos 0.0001

The code with foreach also seems less complex and clearer

empiricompany:
> See an updated list here https://github.com/mitchellkrogza/nginx-ultimate-bad-bot-blocker. In addition, the number is much higher.
> 
> Besides an issue of decreased performance, the advantages of this PR must be carefully analyzed. Everyone should keep their own list and keep it as short as possible. I use .htaccess not to ignore bots but to block them.

I also use Nginx to block bad bots at the source, but I don't want to block Googlebot.

I don't think there are performance differences simply by changing the comparison method. This was supposed to be a very simple PR to prevent exclusion if the version changed in the user_agent string for googlebot, like Googlebot/2.1 change to Googlebot/2.0.

I wanted to include the "compatible;" keyword in the exclusion list or by default in the method, which I think is only used by bots?. However, I left this decision to the discretion of the developer.

If we want to develop a more complex mechanism, I'm open to other ideas (for example, we could save the check in session for subsequent requests).

app/code/core/Mage/Log/etc/config.xml
fballiano: agree with @ADDISON74, although I think this list is ok, in the sense that to have about 20 lines shouldn't make any difference in computation times. but if we were to import like 1000 then it would probably not be a good idea

empiricompany: Yes, I leave the decision up to you. For me, the most important thing is the exclusion method.

fballiano (approved):
I still feel think PR could be merged as it is, I like it.

fballiano:
probably it could be improved in the future (like everything) but I don't see any point that would prevent us from implementing it :-)

sreichel (approved):

PR #3237, opened 2023-05-04
Minor phpstan-fixes for php8.1

sreichel:
### Description (*)

Some fixes for php8.1 (bc)

### Related Pull Requests

1. See OpenMage/magento-lts#3163

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3240, opened 2023-05-09
Removed unmatched errors from PHPStan baseline after PHP8.1 upgrade

fballiano:
After https://github.com/OpenMage/magento-lts/pull/3163 PHPStan doesn't match some of the errors that were matched before, thus we can remove them from the baseline


ADDISON74 (approved):

kiatng (approved):

sreichel:
:+1:  Most others are fixed in #3237 .

PR #3245, opened 2023-05-11
Updated `Varien_Object::getData()` and added `getDataByKey()` & `getDataByPath()`

sreichel:
### Description (*)

https://magento.stackexchange.com/questions/248180/magento-2-good-practice-to-use-avoid-magic-getters

Idea was to add  `getDataByKey()` & `getDataByPath()`, but then did some xhprof tests for `getData()`.

https://gist.github.com/sreichel/85730194e5e6ad52876700b4b90aafc3

Code comes from M2 ...

### ToDo

Replace magic getters (?)


kiatng (approved):
Tested.

sreichel:
Any updates here? (Still no time to work on it)

lib/Varien/Object.php
```diff
@@ -318,56 +320,61 @@ public function getData($key = '', $index = null)
...
+        return $data;
+    }
+
+    /**
+     * Get object data by particular key
+     *
+     * @param string $key
+     * @return mixed
+     */
+    public function getDataByKey($key)
```
pquerner, line 375: Only accept `string` ?

lib/Varien/Object.php
```diff
@@ -318,56 +320,61 @@ public function getData($key = '', $index = null)
...
+
+    /**
+     * Get object data by path
+     *
+     * Method consider the path as chain of keys: a/b/c => ['a']['b']['c']
+     *
+     * @param string $path
+     * @return mixed
+     */
+    public function getDataByPath($path)
```
pquerner, line 352: Only accept `string` ?

kiatng: Yes. `$path` must be of type string.

lib/Varien/Object.php
kiatng: Yes, `$key` is expected to be of type string.

fballiano:
I wouldn't know where to start testing it, it's for sure ok but the getdata/setdata is such a core thing that a minor mistake will break everything and there are many use cases to test.
is it absolutely safe to merge it in main?
also, I've no idea how to read the performance data here https://gist.github.com/sreichel/85730194e5e6ad52876700b4b90aafc3

sreichel:
> also, I've no idea how to read the performance data

Indeed, its really hard to read. (also removed some lines) :(

I had testscript, that called `getData` in a nested loop for 200.002 times with different input ... 

- simple Data .... `simplestring`
- nested Data ... `my/nested/value` ... (these calls use `getDataByPass`)
- ...

First chart is for `getData()`. First two lines are the old and the new method including all sub-calls. The two below is the method i did change.

I dont have the script anymore ... but you can see for 200.002 calls, the old method is called 200.009 time. Depending on input it has called itself and/or `strpos` ... so old `Varien_Object::getData` is slower.

Second chart compared ...

- magic methods (slowest) 
- call to `getData($key)`
- call to `getDataByKey($key)` (core code should be checked to use this one)

Code comes from M2 ... im pretty sure, its save to merge.

lib/Varien/Object.php
```diff
@@ -300,13 +300,15 @@ public function unsetOldData($key = null)
...
      *
-     * If $key is empty will return all the data as an array
-     * Otherwise it will return value of the attribute specified by $key
+     * If $key is not defined will return all the data as an array.
+     * Otherwise it will return value of the element specified by $key.
+     * It is possible to use keys like a/b/c for access nested array data
      *
      * If $index is specified it will assume that attribute data is an array
-     * and retrieve corresponding member.
+     * and retrieve corresponding member. If data is the string - it will be explode
```
ADDISON74, line 310: If data is **a** string

ADDISON74 (changes requested):

lib/Varien/Object.php
sreichel: Mhh, im no native speaker, but "the" seem correct here.

Hhowever ... can you please use github suggestion function? (ctrl-g)

ADDISON74: The wording in that context is not exactly appropriate. It is not an error to turn this PR upside down, it can reformulated. As I can live with some of the OM translations, I can live with this wording.

sreichel:
Thanks.

Issue #1226, opened 2020-09-23
Remove "MAGE" cache tag


joshua-bn:
### Summary (*)

There is a cache tag named "MAGE" which contains every cache item except config. There's no mechanism to actually do anything with this tag.

I have over 2 million cache items on one of my clusters. I tried loading `zc:ti:MAGE` (or whatever the key is) and it took so long that I was afraid I'd crash my cluster so I stopped.

### Examples (*)

```
        /**
         * Add global magento cache tag to all cached data exclude config cache
         */
        if (!in_array(Mage_Core_Model_Config::CACHE_TAG, $tags)) {
            $tags[] = Mage_Core_Model_App::CACHE_TAG;
        }
```

### Proposed solution

Remove the code above and any other references to `Mage_Core_Model_App::CACHE_TAG`. Leave the constant there for backwards compatibility.

joshua-bn:
Discussion from Discord: https://discord.com/channels/408199701196963842/408199701196963844/758049538241069127

colinmollenhour:
This tag is actually used with the "Flush Magento Cache" button vs the "Flush Cache Storage" button which flushes everything. However, I agree that this is an ill-conceived "feature" because it puts strain on the backend and goes against the whole idea of tagging. I would be for removing this tag and removing the corresponding "Flush Magento Cache" button or changing it's behavior to instead do something like loop through every cache type removing them one at a time. This would also be a good opportunity to implement a safe config refresh inside the UI so that user's don't crash their stores with these buttons.

joshua-bn:
@colinmollenhour I believe that was the intention, but that's not how it's implemented...

Get the URL
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/Block/Cache.php#L40

The URL (flushSystem)
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/Block/Cache.php#L68 

Call cleanCache (no tags)
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Adminhtml/controllers/CacheController.php#L67

Call the clean method of the cache (no tags)
https://github.com/OpenMage/magento-lts/blob/1.9.4.x/app/code/core/Mage/Core/Model/App.php#L1227

Jumping through some interfaces and stuff to get to your Redis class:
[...]

PR #3246, opened 2023-05-11
Removed "MAGE" cache tag

fballiano:
This PR implements the suggestions discussed in https://github.com/OpenMage/magento-lts/issues/1226 and fixes it.

The MAGE cache tag is only used in order to be able to flush all cache without actually flushing the whole cache storage and since sharing the cache storage with another project/software/whatever would be a very bad practice, we can think about safely remove it.

Removing it will allow for less cache storage space usage and less data to process.

### Questions that need to be answered 1

https://github.com/OpenMage/magento-lts/blob/e464b270a1deca5e9dfd8307eeb178d5a5921ce1/app/code/core/Mage/Admin/Model/Resource/Variable.php#L31-L43

This method only uses Mage_Core_Model_App::CACHE_TAG as cache tag, what are we doing with that method?

### Questions that need to be answered 2

https://github.com/OpenMage/magento-lts/blob/e464b270a1deca5e9dfd8307eeb178d5a5921ce1/app/code/core/Mage/Admin/Model/Resource/Block.php#L59-L75

same thing as before

### Questions that need to be answered 3

https://github.com/OpenMage/magento-lts/blob/e464b270a1deca5e9dfd8307eeb178d5a5921ce1/app/code/core/Mage/Core/Model/Cache.php#L421-L434

do we just replace this method with the flush() method?

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/1226

### DIFF
--- a/app/code/core/Mage/Core/Model/App.php
+++ b/app/code/core/Mage/Core/Model/App.php
@@ -37,12 +37,6 @@ class Mage_Core_Model_App
     public const DISTRO_LOCALE_CODE = 'en_US';
 
     /**
-     * Cache tag for all cache data exclude config cache
-     *
-     */
-    public const CACHE_TAG = 'MAGE';
-
-    /**
      * Default store Id (for install)
      */
     public const DISTRO_STORE_ID       = 1;
--- a/app/code/core/Mage/Admin/Model/Resource/Block.php
+++ b/app/code/core/Mage/Admin/Model/Resource/Block.php
@@ -70,7 +70,7 @@ class Mage_Admin_Model_Resource_Block extends Mage_Core_Model_Resource_Db_Abstra
         Mage::app()->saveCache(
             Mage::helper('core')->jsonEncode($data),
             self::CACHE_ID,
-            [Mage_Core_Model_App::CACHE_TAG]
+            [Mage_Core_Model_Resource_Db_Collection_Abstract::CACHE_TAG]
         );
     }
 
--- a/app/code/core/Mage/Admin/Model/Resource/Variable.php
+++ b/app/code/core/Mage/Admin/Model/Resource/Variable.php
@@ -38,7 +38,7 @@ class Mage_Admin_Model_Resource_Variable extends Mage_Core_Model_Resource_Db_Abs
         Mage::app()->saveCache(
             Mage::helper('core')->jsonEncode($data),
             self::CACHE_ID,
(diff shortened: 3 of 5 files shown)


app/code/core/Mage/Core/Model/App.php
```diff
@@ -36,12 +36,6 @@ class Mage_Core_Model_App
 
     public const DISTRO_LOCALE_CODE = 'en_US';
 
-    /**
-     * Cache tag for all cache data exclude config cache
-     *
-     */
-    public const CACHE_TAG = 'MAGE';
-
```
ADDISON74, line 44: I guess PHPStan is right by removing this line we get in many places "Access to undefined constant Mage_Core_Model_App::CACHE_TAG.". I am seeing this error when opening the PR in browser.

fballiano: yes, the 3 methods mentioned in the PR description that still have to be fixed (first we have to decide how to) ;-)

colinmollenhour:
> This method only uses Mage_Core_Model_App::CACHE_TAG as cache tag, what are we doing with that method?

Ideally they are associated with a cache type, I'd say either config (Mage_Core_Model_Config::CACHE_TAG) or collections (Mage_Core_Model_Resource_Db_Collection_Abstract::CACHE_TAG) or add a new tag or cache type.

> do we just replace this method with the flush() method?

Personally I think `flush()` should be used since this is fast and won't leak any untagged entries, but if some are opposed to that I think looping over the types would also work:

```php
foreach ($this->getTypes() as $type => $flag) {
    $this->cleanType($type);
}
```

fballiano:
all done, thanks @colinmollenhour as usual!

I think this is ready for review now

colinmollenhour (approved):

empiricompany (approved):

ADDISON74 (approved):

sreichel:
Merge latest into #3248  and see ...

```
FAILED
ERROR: Warning: Undefined variable $res  in /var/www/html/app/code/core/Mage/Core/Model/Cache.php on line 426
```

fballiano:
can't reproduce this problem with "main" and developer_mode on, how do you find it?

sreichel:
Not at home atm ... run 'clean()' method from script or test my ddev command.

'$res' is not set ...

fballiano:
true, PR coming in 3 seconds

PR #3247, opened 2023-05-12
Resized panels height in attribute set backend view

fballiano:
This PR replaces https://github.com/OpenMage/magento-lts/pull/1311 and it is coauthored by @loekvangool.

This PR is slightly different cause it doesn't force any height/overflow, reducing the forced CSS lines and letting the browser fill all space

BEFORE:
Screenshot 2023-05-12 alle 18 22 44

AFTER:
Screenshot 2023-05-12 alle 18 22 18

### DIFF
--- a/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
@@ -60,7 +60,7 @@
                     </tbody>
                 </table>
             </div>
-            <div id="tree-div2" style="height:600px;margin-top:5px;overflow:auto"></div>
+            <div id="tree-div2" style="height:500px;margin-top:5px;overflow:auto"></div>
             <script type="text/javascript">
             //<![CDATA[
             var allowDragAndDrop = <?php echo ($this->getIsReadOnly() ? 'false' : 'true'); ?>;
--- a/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
@@ -60,7 +60,7 @@
                     </tbody>
                 </table>
             </div>
-            <div id="tree-div2" style="height:400px; margin-top:5px;overflow:auto"></div>
+            <div id="tree-div2" style="margin-top:5px"></div>
             <script type="text/javascript">
             //<![CDATA[
             var allowDragAndDrop = <?php echo ($this->getIsReadOnly() ? 'false' : 'true'); ?>;
--- a/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
+++ b/app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
@@ -60,7 +60,7 @@
                     </tbody>
                 </table>
             </div>
-            <div id="tree-div2" style="margin-top:5px"></div>
+            <div id="tree-div2" style="height:600px;margin-top:5px;overflow:auto"></div>
             <script type="text/javascript">
             //<![CDATA[
             var allowDragAndDrop = <?php echo ($this->getIsReadOnly() ? 'false' : 'true'); ?>;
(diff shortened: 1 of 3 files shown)


kiatng (dismissed):

ADDISON74:
This PR can be useful only when the number of groups and attributes is reduced. The moment you have 250 attributes like I have in a store, if I put them in groups I will end up scrolling again. I don't like not being able to make moves at the same level in the browser window. 

The only change I would make would be for the two columns to be equal, increasing a little the height (space 1 + 2 in the image bellow), but I would keep the scroll in both, it helps me move the attributes to the groups where I need very fast, **AT THE SAME LEVEL**. You don't have to prove me right, I recommend you take a test to reach the same conclusion..

height

loekvangool:
I don't get that @ADDISON74. Scrolling is not removed with the PR I think?

ADDISON74:
@loekvangool - please look at the image posted by the PR author, which is after modification. The scroll disappears for both columns, below you have a screenshot after I checked out on this PR. The height of the viewport is around 1.5 x 1080p.

screenshot

Our goal must be to equalize the two columns height and keep the scrolls **precisely to perform operations at the same level on the page**, from one column to another, in a normal browser windows. Imagine to take an attribute from the bottom, let's say two scrolls from the window size and move it to the first group will be damn annoying if this PR is approved.

fballiano:
mmm if the goal is being able to equalize before dragging the elements, then I think 400px (the original value) is better than 700px (which on my macbook creates a double scrolling, one in the div, one in the window).

but it is true, with height auto and the browser based scrolling it is harder to drag and scroll.

wait, i'll change the PR

fballiano:
reset the PR to fixed height, but of 600px

ADDISON74:
In order to set the values as close as possible to what they should be, I played directly in the browser. I had the following in mind:

**display**: 1920 x 1080
**browser**: Chrome + Bookmarks visible
**OpenMage**: no messages bellow the nav menu and "Display Demo Store Notice" hidden.

In my tests **tree-div1 = 500p**x and **tree-div2 = 550 px**. The columns appear as equal now and the window has no scroll.

columns

If I hide the Bookmarks bar, the value of 600px still generates scrolling in the window and leaves a lot of unused space at the bottom. So the most suitable values would be 500px / 550px. Please do your checks.

scroll

fballiano:
I've changed it to 500.

All of this effort to change from 400 to 500, dunno if it was worth it actually

fballiano:
but thanks @ADDISON74 because 1920x1080 is still the default resolution for most of the people

ADDISON74:
If we lost so much energy in such a PR, the values must be 500px (tree-div1) and 550px (tree-div2). At least the containers would be equal and the energy should have been usefully spent.

fballiano:
ahhh i didn't notice they weren't starting from the same point

if i do 500 and 550 i get this
Screenshot 2023-05-14 alle 17 52 37


but if i do 500 and 520 it seems ok
Screenshot 2023-05-14 alle 17 53 19

fballiano:
ok it depends if the title of the 3rd panel goes on 2 lines...
Screenshot 2023-05-14 alle 17 54 25

I'll set it to 550

app/design/adminhtml/default/default/template/catalog/product/attribute/set/main.phtml
```diff
@@ -60,7 +60,7 @@
                     </tbody>
                 </table>
             </div>
-            <div id="tree-div2" style="height:400px; margin-top:5px;overflow:auto"></div>
+            <div id="tree-div2" style="height:500px;margin-top:5px;overflow:auto"></div>
```
fballiano, line 63:
```suggestion
            <div id="tree-div2" style="height:550px;margin-top:5px;overflow:auto"></div>
```

ADDISON74 (approved):

sreichel:
To avoid linebreaks, how about to move the left attribute selection to top and make it a "2-column" layout?

fballiano:
it would solve the linebreak but removing a lot of vertical space, which is what the original idea of the PR is about, so I'd say no. there's also not really the need for more horizontal space.

maybe we could just make the titles smaller but it's a no-problem imho.

fballiano:
@loekvangool could you review this one?

ADDISON74:
We should merge this PR. One more vote, but please test it before.

loekvangool:
I’m not in a position to test until next week.

PR #3248, opened 2023-05-12
Added DDEV command to install OM (incl. sample data)

sreichel:
### Description (*)

Adds ddev-command `ddev install-openmage` to install OM w/ or w/o sample data.

There a three optional flags ...

`-d` use default values for admin-user, db, ... w/o getting asked for
`-s` install sample data ...
`-k` keep sample data package (to avoid to download it again and again)
`-q` drop current install and re-install with default values

Run `ddev install-openmage -d -s` to install OM /w sample data.

### Questions or comments

Install works only WITH sample data!

php7.4
```
FAILED
ERROR: Error in file: "/var/www/html/app/code/core/Mage/Customer/sql/customer_setup/mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php" - Notice: Undefined offset: 0  in /var/www/html/app/code/core/Mage/Core/Model/App.php on line 1005
```

php8.1
```
FAILED
ERROR: Error in file: "/var/www/html/app/code/core/Mage/Customer/sql/customer_setup/mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php" - Warning: Undefined array key 0  in /var/www/html/app/code/core/Mage/Core/Model/App.php on line 1005
```

Updates welcome.

### DIFF
--- a/.ddev/commands/web/install-openmage
+++ b/.ddev/commands/web/install-openmage
@@ -1,0 +1,94 @@
+#!/bin/bash
+
+## Description: Install OpenMage
+## Usage: install-openmage
+## Example: ddev install-openmage
+
+ROOT="${PWD}"
+
+LOCALXML="${ROOT}/app/etc/local.xml"
+if [ -f "${LOCALXML}" ]; then
+  read -r -p "OpenMage is already installed. Delete local.xml and drop database[y/N]:" DELETE
+  DELETE=${DELETE,,} # to lower
+  if [[ "${DELETE}" =~ ^(yes|y) ]]; then
+    mysql -u db -h db -e "DROP SCHEMA IF EXISTS db; CREATE SCHEMA db;";
+    rm "${LOCALXML}"
+  else
+    exit 1
+  fi
+fi
+
+read -r -p "Install sample data? [y/N]" INSTALL_SAMPLE_DATA
+INSTALL_SAMPLE_DATA=${INSTALL_SAMPLE_DATA,,} # to lower
+if [[ $INSTALL_SAMPLE_DATA =~ ^(yes|y) ]]; then
+  SAMPLE_DATA_URL=https://github.com/Vinai/compressed-magento-sample-data/raw/master/compressed-magento-sample-data-1.9.2.4.tgz
+  SAMPLE_DATA_DIRECTORY="${ROOT}/.sampleData"
+  SAMPLE_DATA_FILE=sample_data.tgz
+
+  if [[ ! -d "${SAMPLE_DATA_DIRECTORY}" ]]; then
+    echo "Creating backup directory"
+    mkdir -p "${SAMPLE_DATA_DIRECTORY}"
+  fi
+
+  cd "${SAMPLE_DATA_DIRECTORY}" || exit
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data"
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompress sample data"
+  tar xf "${SAMPLE_DATA_FILE}"
+
+  echo "Copy sample data"
+  cp -r magento-sample-data-1.9.2.4/* "${PWD}/"
+
+  echo "Import sample data"
+  mysql -u db -h db db < "${SAMPLE_DATA_DIRECTORY}/magento-sample-data-1.9.2.4/magento_sample_data_for_1.9.2.4.sql"
+
+  # remove sample data
+  if [[ "${SAMPLE_DATA_KEEP_FLAG}" ]]; then
+    rm -rf $(
+      find . -maxdepth 1 -type f -name "*" ! -name "${SAMPLE_DATA_FILE}")
+  else
+    cd "${ROOT}" || exit
+    rm -rf "${SAMPLE_DATA_DIRECTORY}"
+  fi
+fi
+
+cd "${ROOT}" || exit
+
+read -r -p "Admin user [admin]:" ADMIN_USER
+ADMIN_USER=${ADMIN_USER:-admin}
+read -r -p "Admin firstname [OpenMage]:" ADMIN_FIRSTNAME
+ADMIN_FIRSTNAME=${ADMIN_FIRSTNAME:-OpenMage}
+read -r -p "Admin lastname [User]:" ADMIN_LASTNAME
+ADMIN_LASTNAME=${ADMIN_LASTNAME:-User}
+read -r -p "Admin email [admin@example.com]:" ADMIN_EMAIL
+ADMIN_EMAIL=${ADMIN_EMAIL:-admin@example.com}
+read -r -p "Admin password [veryl0ngpassw0rd]:" ADMIN_PASSWORD
+ADMIN_PASSWORD=${ADMIN_PASSWORD:-veryl0ngpassw0rd}
+
+php -f install.php -- \
+  --license_agreement_accepted 'yes' \
+  --locale 'en_US' \
+  --timezone 'America/New_York' \
+  --db_host 'db' \
+  --db_name 'db' \
+  --db_user 'db' \
+  --db_pass 'db' \
+  --db_prefix '' \
+  --url "${DDEV_PRIMARY_URL}" \
+  --use_rewrites 'yes' \
+  --use_secure 'no' \
+  --secure_base_url "${DDEV_PRIMARY_URL}" \
+  --use_secure_admin 'no' \
+  --admin_username "${ADMIN_USER}" \
+  --admin_lastname "${ADMIN_LASTNAME}" \
+  --admin_firstname "${ADMIN_FIRSTNAME}" \
+  --admin_email "${ADMIN_EMAIL}" \
+  --admin_password "${ADMIN_PASSWORD}" \
+  --session_save 'files' \
+  --admin_frontname 'admin' \
+  --backend_frontname 'admin' \
+  --default_currency 'USD' \
+  --skip_url_validation 'yes'
--- a/.ddev/commands/web/install-openmage
+++ b/.ddev/commands/web/install-openmage
@@ -2,26 +2,40 @@
 
 ## Description: Install OpenMage
 ## Usage: install-openmage
-## Example: ddev install-openmage -d -s -k
+## Example: ddev install-openmage -d -s -k -q
 
 ROOT="${PWD}"
 
+QUIET_INSTALL_FLAG=''
 SAMPLE_DATA_FLAG=''
 SAMPLE_DATA_KEEP_FLAG=''
 USE_DEFAULT_FLAG=''
 
-while getopts 'dsk' flag; do
+while getopts 'dskq' flag; do
   case "${flag}" in
     d) USE_DEFAULT_FLAG='true' ;;
     s) SAMPLE_DATA_FLAG='true' ;;
     k) SAMPLE_DATA_KEEP_FLAG='true' ;;
+    q) QUIET_INSTALL_FLAG='true' ;;
   esac
 done
 
+
+#quiet install
+if [[ "${QUIET_INSTALL_FLAG}" ]]; then
+  USE_DEFAULT_FLAG='true'
+  SAMPLE_DATA_FLAG='true'
+fi
+
 LOCALXML="${ROOT}/app/etc/local.xml"
 if [ -f "${LOCALXML}" ]; then
-  read -r -p "OpenMage is already installed. Delete local.xml and drop database[y/N]:" DELETE
-  DELETE=${DELETE,,} # to lower
+  if [[ "${QUIET_INSTALL_FLAG}" ]]; then
+    DELETE='y'
+  else
+    read -r -p "OpenMage is already installed. Delete local.xml and drop database[y/N]:" DELETE
+    DELETE=${DELETE,,} # to lower
+  fi
+
   if [[ "${DELETE}" =~ ^(yes|y) ]]; then
     mysql -u db -h db -e "DROP SCHEMA IF EXISTS db; CREATE SCHEMA db;";
     rm "${LOCALXML}"
@@ -31,7 +45,7 @@ if [ -f "${LOCALXML}" ]; then
 fi
 
 # install sample data
-if [[ "${SAMPLE_DATA_KEEP_FLAG}" ]]; then
+if [[ "${SAMPLE_DATA_FLAG}" ]]; then
   INSTALL_SAMPLE_DATA='y'
 else
   read -r -p "Install sample data? [y/N]" INSTALL_SAMPLE_DATA
--- a/.ddev/commands/web/install-openmage
+++ b/.ddev/commands/web/install-openmage
@@ -2,9 +2,21 @@
 
 ## Description: Install OpenMage
 ## Usage: install-openmage
-## Example: ddev install-openmage
+## Example: ddev install-openmage -d -s -k
 
 ROOT="${PWD}"
+
+SAMPLE_DATA_FLAG=''
+SAMPLE_DATA_KEEP_FLAG=''
+USE_DEFAULT_FLAG=''
+
+while getopts 'dsk' flag; do
+  case "${flag}" in
+    d) USE_DEFAULT_FLAG='true' ;;
+    s) SAMPLE_DATA_FLAG='true' ;;
+    k) SAMPLE_DATA_KEEP_FLAG='true' ;;
+  esac
+done
 
 LOCALXML="${ROOT}/app/etc/local.xml"
 if [ -f "${LOCALXML}" ]; then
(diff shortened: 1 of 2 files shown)


ADDISON74:
I will test this PR these days, being a fan of DDEV on Windows + Docker + WSL2.

BTW, we have not yet determined what happens with the local.xml file which is automatically created by DDEV when we configure a new project and thus we cannot access to the Backend running the installation. Maybe we should also make an command to run a query for creating an admin account with a simple password. We had this conversion here #2648.

ADDISON74:
@sreichel - I see that you addressed in the script the issue of creating the administrator account for Backend.

sreichel:
@ADDISON74 in case of an existing `local.xml` the script stops. It just to (re-)install OM with default ddev values.

ADDISON74:
A few comments

1 - If the command ```ddev install-openmage``` is run and the question "OpenMage is already installed. Delete local.xml and drop database[y/N]:" is answered with no, the error 'Failed to run install-openmage : exit' is displayed status 1'. However it is not a failure of running the script, but it is coming from the install.php file. We cannot do more about it.

2 -  DDEV provides support for SSL certificates. The arguments in the install script ```use_secure``` and ```use_secure_admin``` should be set to 'yes'.

3 - You can work a little on the phrases

"OpenMage is already installed. Delete local.xml and drop the database? [y/N] :". We keep the same format like in "Install sample data? [y/N]", it is a question.

"Install sample data? [y/N] :". You missed the colon.

Otherwise the script has no other issues and it is extremely useful. I recommend everyone who tests OpenMage to use (DDEV + Docker + WSL2), especially in Windows, if you do not work directly in a Linux distribution.

ADDISON74:
We should create a DDEV command to create/update the admin account.

sreichel:
> We should create a DDEV command to create/update the admin account. 

IMHO we should better support N98-magerun for these tasks.

ADDISON74:
I think it's faster with a DDEV command (e.g. ```ddev admin-openmage```) than waiting for an update to n98. You tried to move forward this great tool here https://github.com/netz98/n98-magerun/issues/1275, but it remained at the draft stage for a long time.

PS - a similar bash script like the one in this PR based on n98 can be found here https://github.com/cmuench/n98-magerun/blob/feature/php81/.ddev/commands/web/install-openmage.

.ddev/commands/web/install-openmage
```diff
@@ -0,0 +1,137 @@
...
+  ADMIN_USER=${ADMIN_USER:-admin}
+  read -r -p "Admin firstname [OpenMage]:" ADMIN_FIRSTNAME
+  ADMIN_FIRSTNAME=${ADMIN_FIRSTNAME:-OpenMage}
+  read -r -p "Admin lastname [User]:" ADMIN_LASTNAME
+  ADMIN_LASTNAME=${ADMIN_LASTNAME:-User}
+  read -r -p "Admin email [admin@example.com]:" ADMIN_EMAIL
+  ADMIN_EMAIL=${ADMIN_EMAIL:-admin@example.com}
+  read -r -p "Admin password [veryl0ngpassw0rd]:" ADMIN_PASSWORD
+  ADMIN_PASSWORD=${ADMIN_PASSWORD:-veryl0ngpassw0rd}
+  read -r -p "Database prefix []:" DB_PREFIX
```
ADDISON74, line 110: This should be "Tables prefix" not "Database prefix". Internal it is db_prefix which is correct but the meaning is completely different. When you install OM in browser the input label is "Tables Prefix".

tables_prefix

.ddev/commands/web/install-openmage
```diff
@@ -0,0 +1,137 @@
...
+  --admin_username "${ADMIN_USER}" \
+  --admin_lastname "${ADMIN_LASTNAME}" \
+  --admin_firstname "${ADMIN_FIRSTNAME}" \
+  --admin_email "${ADMIN_EMAIL}" \
+  --admin_password "${ADMIN_PASSWORD}" \
+  --session_save 'files' \
+  --admin_frontname 'admin' \
+  --backend_frontname 'admin' \
+  --default_currency 'USD' \
+  --skip_url_validation 'yes'
```
ADDISON74, line 137: I would add the following value ```--enable-charts 'yes''```

web_access_options

.ddev/commands/web/install-openmage
```diff
@@ -0,0 +1,137 @@
...
+    rm -rf "${SAMPLE_DATA_DIRECTORY}"
+  fi
+fi
+
+cd "${ROOT}" || exit
+
+if [[ "${USE_DEFAULT_FLAG}" ]]; then
+  ADMIN_USER='admin'
+  ADMIN_FIRSTNAME='OpenMage'
+  ADMIN_LASTNAME='User'
```
ADDISON74, line 95: You can use 'Administrator' instead of 'User'.

ADDISON74:
In all ```read``` lines we need a blank space at the end.

Before: ```read -r -p "Install sample data? [y/N]"```
After: ```read -r -p "Install sample data? [y/N] "```

In this way the input from the keyboard will be more visible.

ADDISON74:
The name of the command should start with **openmage-** prefix, so that we can add other commands related to OpenMage,that are easy to remember. e.g,

**openmage-install** = what you just created in this PR
**openmage-admin** = update/create admin account (will come)
...

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+
+## Description: Install OpenMage
+## Usage: openmage-install
```
ADDISON74, line 4: Description: Install OpenMage w/o Sample Data

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+  SAMPLE_DATA_FLAG='true'
+fi
+
+LOCALXML="${ROOT}/app/etc/local.xml"
+if [ -f "${LOCALXML}" ]; then
+  if [[ "${QUIET_INSTALL_FLAG}" ]]; then
+    DELETE='y'
+  else
+    read -r -p "OpenMage is already installed. Delete local.xml and drop database [y/N]: " DELETE
+    DELETE=${DELETE,,} # to lower
```
ADDISON74, line 36: "OpenMage is already installed. Delete the local.xml file and drop the database? [y/N] "

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+  else
+    exit 1
+  fi
+fi
+
+# install sample data
+if [[ "${SAMPLE_DATA_FLAG}" ]]; then
+  INSTALL_SAMPLE_DATA='y'
+else
+  read -r -p "Install sample data? [y/N]: " INSTALL_SAMPLE_DATA
```
ADDISON74, line 51: "Install sample data? [y/N] "

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+  INSTALL_SAMPLE_DATA=${INSTALL_SAMPLE_DATA,,} # to lower
+fi
+
+if [[ $INSTALL_SAMPLE_DATA =~ ^(yes|y) ]]; then
+  SAMPLE_DATA_URL=https://github.com/Vinai/compressed-magento-sample-data/raw/master/compressed-magento-sample-data-1.9.2.4.tgz
+  SAMPLE_DATA_DIRECTORY="${ROOT}/.sampleData"
+  SAMPLE_DATA_FILE=sample_data.tgz
+
+  if [[ ! -d "${SAMPLE_DATA_DIRECTORY}" ]]; then
+    echo "Creating backup directory"
```
ADDISON74, line 61: "Create backup directory..."

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+    mkdir -p "${SAMPLE_DATA_DIRECTORY}"
+  fi
+
+  cd "${SAMPLE_DATA_DIRECTORY}" || exit
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data"
```
ADDISON74, line 67: "Download sample data..."

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompress sample data"
```
ADDISON74, line 71: "Uncompress sample data..."

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+  tar xf "${SAMPLE_DATA_FILE}"
+
+  echo "Copy sample data"
+  cp -r magento-sample-data-1.9.2.4/* "${PWD}/"
+
+  echo "Import sample data"
```
ADDISON74, line 77: "Import sample data..."

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,138 @@
...
+  cd "${SAMPLE_DATA_DIRECTORY}" || exit
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data"
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompress sample data"
+  tar xf "${SAMPLE_DATA_FILE}"
+
+  echo "Copy sample data"
```
ADDISON74, line 74: "Copy sample data..."

ADDISON74:
The functionality is exactly what it needs. I made only a few proposals to improve the relationship with the user. Thank you @sreichel for this PR that improves OpenMage's testing mode and helps by reducing the time it takes to install a new instance.

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,154 @@
...
+  cd "${SAMPLE_DATA_DIRECTORY}" || exit
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data ..."
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompress sample data ..."
+  tar xf "${SAMPLE_DATA_FILE}"
+
+  echo "Copy sample data ..."
```
ADDISON74, line 90: Copying

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,154 @@
...
+  cp -r magento-sample-data-1.9.2.4/* "${PWD}/"
+
+  echo "Import sample data ..."
```
ADDISON74, line 93: Importing

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,154 @@
...
+    mkdir -p "${SAMPLE_DATA_DIRECTORY}"
+  fi
+
+  cd "${SAMPLE_DATA_DIRECTORY}" || exit
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data ..."
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompress sample data ..."
```
ADDISON74, line 87: Uncompressing

ADDISON74 (commented):
Only 4 modifications I request :)

.ddev/commands/web/openmage-install
sreichel: You can make it bit easier .... mark code that needs changes and hit `ctrl-g` ;)

ADDISON74 (dismissed):

sreichel:
@ADDISON74 can you please test without sample data?

ADDISON74:
**PHP 7.4.33** - I was able to install both variants. There were no errors when I used the variant without installing the sample data.
**PHP 8.1.16** - The same here. SUCCESS.
**PHP 8.2.3** - The same here too. SUCCESS.

sreichel:
Strange. Does table-prefix work for you?

ADDISON74:
I was able to add a prefix to the tables, see the images below. PHP 8.2.3:

table_prefix

table_prefix-2

sreichel:
When it works for you a default prefix should be added. (`om_`).

Will add some options later ... like session storage ...

ADDISON74:
My request was off-topic to this PR and I deleted it, there were two messages. In the meantime I opened a PR that solves that situation differently from n98-magerun, where it is a lot of work to be brought to compatibility with PHP8.1. I deleted your answer because the question no longer exists.

.ddev/commands/web/openmage-install
```diff
@@ -123,8 +123,8 @@ else
   if [[ $INSTALL_SAMPLE_DATA =~ ^(yes|y) ]]; then
     TABLE_PREFIX=''
   else
-    read -r -p "Table prefix []: " TABLE_PREFIX
-    TABLE_PREFIX=${TABLE_PREFIX:-}
+    read -r -p "Table prefix [om_]: " TABLE_PREFIX
+    TABLE_PREFIX=${TABLE_PREFIX:-om_}
```
ADDISON74, line 127: What happens if I don't want a table prefix when I am not using the sample data? By default, 'om_' is added if I press Enter. The previous version was the correct one in my opinion.

sreichel: Good catch. Will update tomorrow.

imho ... table-prefix should be used by default to find errors in modules that dont care about possible set prefix.

sreichel: fixed

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,154 @@
...
+  read -r -p "Admin lastname [Administrator]: " ADMIN_LASTNAME
+  ADMIN_LASTNAME=${ADMIN_LASTNAME:-Administrator}
+  read -r -p "Admin email [admin@example.com]: " ADMIN_EMAIL
+  ADMIN_EMAIL=${ADMIN_EMAIL:-admin@example.com}
+  read -r -p "Admin password [veryl0ngpassw0rd]: " ADMIN_PASSWORD
+  ADMIN_PASSWORD=${ADMIN_PASSWORD:-veryl0ngpassw0rd}
+  if [[ $INSTALL_SAMPLE_DATA =~ ^(yes|y) ]]; then
+    TABLE_PREFIX=''
+  else
+    read -r -e -i 'om_' -p "Table prefix [om_]: " TABLE_PREFIX
```
ADDISON74, line 126: I do not use a prefix in the name of the tables. If I press Enter it will add om_ to all table and I don't want that. It is better to leave the previous version (empty) and whoever wants a prefix to enter it.

sreichel: I case of testing 3rd-party code its really useful to test with prefix set. I'd like to keep it.

ADDISON74: I would like this script not to condition me in the choices I make.

sreichel: This script is only for testing. I want to have it set by default to find error like this ... https://magento.stackexchange.com/a/167944/46249

I'll keep it.

ADDISON74: I consider this script very useful and I understand that you only want to test the version with tables having a prefix. That's why I have nothing against it. But the file system and database can be used later in production with a server configuration other than DDEV. That's what I am doing at the moment, with a shop that I have to get going in the next period. I did not use a prefix for the tables. 

If I were to test your installation script without Sample Data, I would not be able to change the prefix except by editing the file with sed. For this reason, I would prefer to be able to install without a prefix. You could mention that you recommend using a table prefix to cover this situation as well.

.ddev/commands/web/openmage-install
```diff
@@ -0,0 +1,154 @@
...
+  if [[ ! -f "${SAMPLE_DATA_DIRECTORY}/${SAMPLE_DATA_FILE}" ]]; then
+    echo "Downloading sample data ..."
+    wget "${SAMPLE_DATA_URL}" -O "${SAMPLE_DATA_FILE}"
+  fi
+
+  echo "Uncompressing sample data ..."
+  tar xf "${SAMPLE_DATA_FILE}"
+
+  echo "Copying sample data ..."
+  cp -r magento-sample-data-1.9.2.4/* "${PWD}/"
```
ADDISON74, line 88: There is an issue in this command. Its execution is done in the .sampleData directory and thus both /media and /skin are copied here and not in the root directory of OpenMage as it should be.

Here is the fix

```cp -r magento-sample-data-1.9.2.4/* "${ROOT}/"```

ADDISON74 (approved):
Please approve this PR. I have been using it for some time and there are no issues. I made some changes in the displayed phrases and corrected the directory where they are copied from the downloaded archive /media and /skin directories.

fballiano:
@ADDISON74 what about your comment about table prefixes?

ADDISON74:
The command deals with two situations:

1. installation without Sample Data and here the table prefix must be used.

2. installation with Sample Data, here you can no longer use the prefix because the tables are already created.

I agree with Sven in the first case to cover the situation of the table prefix from the very beginning. I started from the idea that what is being worked on in DDEV can be put into production and that's why I didn't want a prefix. But it can be easily solved by exporting the database and executing a sed command to delete the prefix. I haven't analyzed it, but I think that a DDEV command can be created to do this automatically.

fballiano (approved):
approving because asked, not tested

sreichel:
@fballiano you wrote you are not using ddev ... its just a shell script that should work w/o ddev too.

PR #3251, opened 2023-05-14
Do not reindex catalog_product_flat for disabled stores

empiricompany:
### Description (*)
skip reindex product flat tables for disabled store view

### Questions or comments
We have one store with 6 language stores, of which only one is enabled. The reindexing of the product flat tables takes unnecessarily long. The idea is to skip the reindexing of disabled stores. I don't think this will create any problems because these tables are not actually used on the frontend. Are there any potential issues with this approach?

We should add a message indicating that a reindex is required when the store is re-enabled, similar to what happens with the index catalog_product_attribute.


app/code/core/Mage/Catalog/Model/Resource/Product/Flat/Indexer.php
```diff
@@ -1401,6 +1401,7 @@ protected function _arrayNextKey(array $array, $key)
     public function reindexAll()
     {
         foreach (Mage::app()->getStores() as $storeId => $store) {
+            if (!$store->getIsActive()) continue;
```
fballiano, line 1404:
```suggestion
            if (!$store->getIsActive()) {
                continue;
            }
```

fballiano:
wow, this could be a great improvement, how did it slip for so many years...

sorry for reformatting the "continue", I also prefer the one line but PHPCS doesn't :-(

fballiano (approved):

sreichel:
Message after enable store needs to be added.

empiricompany:
> Message after enable store needs to be added.

I'm not an indexer expert, but now the reindex request message also appears for `catalog_product_flat`. 
The code that checks if it's a save and was previously disabled was taken from the EAV index (app/code/core/Mage/Catalog/Model/Product/Indexer/Eav.php).
I'm just wondering if anything else needs to be added in `_registerEvent`.

app/code/core/Mage/Catalog/Model/Product/Indexer/Flat.php
```diff
@@ -151,7 +151,7 @@ public function matchEvent(Mage_Index_Model_Event $event)
             } else {
                 /** @var Mage_Core_Model_Store $store */
                 $store = $event->getDataObject();
-                if ($store && $store->isObjectNew()) {
+                if ($store && ($store->isObjectNew() || $store->getOrigData('is_active') != $store->getIsActive() && $store->getIsActive())) {
```
empiricompany, line 154: maybe it would be best to wrap the new condition into a new statement?

fballiano: I'm wondering if we need the `$store->getOrigData('is_active') != $store->getIsActive()` part, if the store is active `$result` must be `true` anyway? am I missing something?

empiricompany: Yes, I thought that too, but I used the same code I found here which catches the EAV reindex alert too:



app/code/core/Mage/Catalog/Model/Product/Indexer/Eav.php::_registerEvent()@123
```
protected function _registerEvent(Mage_Index_Model_Event $event)
    {
        $entity = $event->getEntity();

        if ($entity == Mage_Catalog_Model_Product::ENTITY) {
            switch ($event->getType()) {
                case Mage_Index_Model_Event::TYPE_DELETE:
                    $this->_registerCatalogProductDeleteEvent($event);
                    break;

                case Mage_Index_Model_Event::TYPE_SAVE:
                    $this->_registerCatalogProductSaveEvent($event);
                    break;

                case Mage_Index_Model_Event::TYPE_MASS_ACTION:
                    $this->_registerCatalogProductMassActionEvent($event);
                    break;
            }
        } elseif ($entity == Mage_Catalog_Model_Resource_Eav_Attribute::ENTITY) {
            switch ($event->getType()) {
                case Mage_Index_Model_Event::TYPE_SAVE:
                    $this->_registerCatalogAttributeSaveEvent($event);
                    break;
            }
        } elseif ($entity == Mage_Catalog_Model_Convert_Adapter_Product::ENTITY) {
            $event->addNewData('catalog_product_eav_reindex_all', true);
        } elseif ($entity == Mage_Core_Model_Store::ENTITY) {
            if ($event->getType() === Mage_Index_Model_Event::TYPE_SAVE) {
                /** @var Mage_Core_Model_Store $store */
                $store = $event->getDataObject();
                if ($store->getOrigData('is_active') != $store->getIsActive() && $store->getIsActive()) {
                    $event->getProcess()->changeStatus(Mage_Index_Model_Process::STATUS_REQUIRE_REINDEX);
                }
            }
        }
    }
```


I also tried to recreate a new store view, and it seems to work correctly.

empiricompany: if think this means

if is a new store view => true
if is not a new store view and is_active is changed and is changed in an active status => true

Then, if you save with an inactive status, the alert will not be shown.

app/code/core/Mage/Catalog/Model/Product/Indexer/Flat.php
```diff
@@ -146,12 +146,13 @@ public function matchEvent(Mage_Index_Model_Event $event)
                 $result = false;
             }
         } elseif ($entity == Mage_Core_Model_Store::ENTITY) {
+            Mage::log(__METHOD__);
```
tmotyl: ?

empiricompany: fixed here: https://github.com/OpenMage/magento-lts/pull/3251/commits/29e531aed6595d457131015c2fc03584a38106a9

roby255 (approved):

fballiano:
merging per 1green and 2grays

Issue #3184, opened 2023-04-16
[PHP 8.2] Warnings in the log files during order placement


ADDISON74:
Use Magento Sample Pack. Add the following products to the cart "Alice in Wonderland" and "Falling by I am Not Lefthanded".

**1. Cart Page - system.log file**

```php
2023-04-16T21:54:25+00:00 ERR (3): Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Downloadable/Helper/Catalog/Product/Configuration.php on line 59
```

**2. Checkout - Onepage Billing Information - system.log file**

```php
2023-04-16T21:57:35+00:00 ERR (3): Deprecated functionality: Creation of dynamic property Mage_Paypal_Model_Config::$cctypes is deprecated  in /var/www/html/app/code/core/Mage/Paypal/Model/Config.php on line 781
```

**3. Checkout - After placing the order - system.log file**

```php 
2023-04-16T21:59:48+00:00 ERR (3): Deprecated functionality: trim(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/SalesRule/Model/Rule.php on line 203
```

fballiano:
Hi @ADDISON74,

1. should be done with https://github.com/OpenMage/magento-lts/pull/3187
2. can you check if you still get the error with https://github.com/OpenMage/magento-lts/pull/3181? maybe that fix also fixes this one
3. will do

TangLiang:
same issue on OpenMage 20.1.0-rc3

````````````````
2023-04-21T02:39:29+00:00 ERR (3): Deprecated functionality: Creation of dynamic property Mage_Paypal_Model_Config::$active is deprecated  in /home/wwwroot/lnmp01/domain/xxx.com/web/app/code/core/Mage/Paypal/Model/Config.php on line 782
2023-04-21T02:39:29+00:00 ERR (3): Deprecated functionality: Creation of dynamic property Mage_Paypal_Model_Config::$visible_on_cart is deprecated  in /home/wwwroot/lnmp01/domain/xxx.com/web/app/code/core/Mage/Paypal/Model/Config.php on line 782
[...]

PR #3253, opened 2023-05-14
Fixed PHP8.2 deprecated warning in Mage_Cron_Model_Observer

fballiano:
This was reported by https://github.com/OpenMage/magento-lts/issues/3184#issuecomment-1546858817

### DIFF
--- a/app/code/core/Mage/Cron/Model/Observer.php
+++ b/app/code/core/Mage/Cron/Model/Observer.php
@@ -220,7 +220,7 @@ class Mage_Cron_Model_Observer
 
         $now = time();
         foreach ($history->getIterator() as $record) {
-            if (strtotime($record->getExecutedAt()) < $now - $historyLifetimes[$record->getStatus()]) {
+            if (strtotime($record->getExecutedAt() ?? '') < $now - $historyLifetimes[$record->getStatus()]) {
                 $record->delete();
             }
         }


ADDISON74 (approved):

app/code/core/Mage/Cron/Model/Observer.php
```diff
@@ -220,7 +220,7 @@ public function cleanup()
 
         $now = time();
         foreach ($history->getIterator() as $record) {
-            if (strtotime($record->getExecutedAt()) < $now - $historyLifetimes[$record->getStatus()]) {
+            if (strtotime($record->getExecutedAt() ?? '') < $now - $historyLifetimes[$record->getStatus()]) {
```
kiatng, line 223: I am not sure if this is better, but the intent is clearer:

```suggestion
            if (empty($record->getExecutedAt())
                || (strtotime($record->getExecutedAt()) < $now - $historyLifetimes[$record->getStatus()])
            ) {
```

Note that `strtotime('')` is `false`. The comparison is odd `if (false < $someValue)`.

fballiano: correct at always, I was sure strtotime('') would give the current timestamp...

kiatng (approved):

Issue #3252, opened 2023-05-14
Missing mod_php section in .htaccess for PHP8 settings


ADDISON74:
If I remember well we discussed an improved version for the .htaccess file last year. Then a section for PHP8 settings was proposed. Today when I looked in the content of the file we only have specific settings for PHP7, not for PHP8.

Here is what we have at this moment in the file

```
############################################
## php7 settings

<IfModule mod_php7.c>

############################################
## adjust max execution time

    php_value max_execution_time 18000

############################################
## disable automatic session start
## before autoload was initialized

    php_flag session.auto_start off

############################################
## enable resulting html compression

    #php_flag zlib.output_compression on

###########################################
# disable user agent verification to not break multiple image upload

    php_flag suhosin.session.cryptua off

</IfModule>
```

fballiano:
I chcked the history of the .htaccess file but I don't see any change that removed that section, probably we never added it at the end. unless it was in a commit on some branch that got lost.

fballiano:
it was this PR https://github.com/OpenMage/magento-lts/pull/2306 anyway

ADDISON74:
The exchange of ideas in that PR showed that improvments are needed in .htaccess. We did the changes for PHP8 in other directories such as media, skin, but not in the root file. We can add a few lines with some important settings.

PR #3254, opened 2023-05-14
Added PHP8 configurations to .htaccess

fballiano:
This PR copies php7 settings to php8 in our htaccess file, it was reported in https://github.com/OpenMage/magento-lts/issues/3252 and follows https://github.com/OpenMage/magento-lts/pull/2306

there was a question, is it mod_php.c or mod_php8.c but I've tested that mod_php.c works and that's what also magento 2 uses (you could check in their repo)

### Fixed Issues
1. Fixes https://github.com/OpenMage/magento-lts/issues/3252

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -40,6 +40,34 @@
 
 </IfModule>
 
+############################################
+## php8 settings
+
+<IfModule mod_php.c>
+
+############################################
+## adjust max execution time
+
+    php_value max_execution_time 18000
+
+############################################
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

fballiano:
@ADDISON74 please feel free, I don't always do that because I don't always want to overwhelm the few active people we have :-(

elidrissidev (approved):

Issue #3264, opened 2023-05-15
Can't create customer from backend.


TangLiang:
### PHP 8.2.4,OpenMage 20.1.0-rc3

**create customer from backend without Date Of Birth**
`````````````````````
[15-May-2023 16:19:57 UTC] PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string - string in /home/wwwroot/lnmp01/domain/magentorc_05/web/vendor/shardj/zf1-future/library/Zend/Locale/Math/PhpMath.php:97
Stack trace:
#0 /home/wwwroot/lnmp01/domain/magentorc_05/web/vendor/shardj/zf1-future/library/Zend/Locale/Format.php(387): Zend_Locale_Math_PhpMath::Sub('0', '', 0)
#1 /home/wwwroot/lnmp01/domain/magentorc_05/web/vendor/shardj/zf1-future/library/Zend/Locale/Format.php(666): Zend_Locale_Format::toNumber('', Array)
#2 /home/wwwroot/lnmp01/domain/magentorc_05/web/vendor/shardj/zf1-future/library/Zend/Filter/NormalizedToLocalized.php(110): Zend_Locale_Format::toFloat('', Array)
#3 /home/wwwroot/lnmp01/domain/magentorc_05/web/lib/Varien/Data/Form/Filter/Date.php(72): Zend_Filter_NormalizedToLocalized->filter('')
#4 /home/wwwroot/lnmp01/domain/magentorc_05/web/app/code/core/Mage/Eav/Model/Attribute/Data/Abstract.php(191): Varien_Data_Form_Filter_Date->inputFilter('')
#5 /home/wwwroot/lnmp01/domain/magentorc_05/web/app/code/core/Mage/Eav/Model/Attribute/Data/Date.php(34): Mage_Eav_Model_Attribute_Data_Abstract->_applyInputFilter('')
#6 /home/wwwroot/lnmp01/domain/magentorc_05/web/app/code/core/Mage/Eav/Model/Form.php(355): Mage_Eav_Model_Attribute_Data_Date->extractValue(Object(Mage_Core_Controller_Request_Http))
#7 /home/wwwroot/lnmp01/domain/magentorc_05/web/app/code/core/Mage/Adminhtml/controllers/CustomerController.php(623): Mage_Eav_Model_Form->extractData(Object(Mage_Core_Controller_Request_Http), 'account')
[...]

PR #3268, opened 2023-05-17
Mage_Eav_Model_Attribute_Data_Date::extractValue() should not return false

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3264 a user is having trouble with Mage_Eav_Model_Attribute_Data_Date::extractValue(), which, in some cases, it could return false instead of "array|string" (the types in the method signature).

Since the "extracted value" should be array or string, and since it's always used as those types, it could be ok to avoid returning false, in case the input filtering method fails.

what do you thing?


kiatng:
Why not change `* @return array|string` to `* @return false|string`. I cannot find where it can return an array.

fballiano:
we could totally do it but I'm wondering about the people that used that docblock already and created code thinking that it would never return a boolean, which is the case of https://github.com/OpenMage/magento-lts/issues/3264 (I think)

:-\
dunno what's the best choice here

sreichel:
I'd fix the doc blocks ...

fballiano:
so, we change the docblock and don't care about the issue that started from somebody's code not expecting a boolean? usually we would go all the way not to break anybody's code.

kiatng:
Idk, but argument can also be made on somebody code that expects a false. IMHO, docblock is wrong and should be corrected.

fballiano:
@kiatng you mean that I should remove the "array" part?

sreichel:
Does phpstorm does not suggest `false|string` for you?

sreichel:
@fballiano I can't remember to have changed code to match docblocks. We always fixed incorrect docs.

fballiano:
> Does phpstorm does not suggest `false|string` for you?

in this moment it autosuggest "@return array|false|string"

fballiano:
> @fballiano I can't remember to have changed code to match docblocks. We always fixed incorrect docs.

ok but that's not what I said, in this case fixing the docblock kinda breaks 3rd party code, while fixing the code to respect the documentation would fix 3rd party code too.

anyway your 2 opinions are against the current implementation, I'll wait some more and than change the docblock ;-)

sreichel:
> in this moment it autosuggest `@return array|false|string`

Mhh. For me its just `false|string`, what seems to be correct for returned `_applyInputFilter()`.

kiatng:
This PR is trying to fix issue #3264, I copy-paste the reported trace:

```
[15-May-2023 16:19:57 UTC] PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string - string in /vendor/shardj/zf1-future/library/Zend/Locale/Math/PhpMath.php:97
Stack trace:
#0 /vendor/shardj/zf1-future/library/Zend/Locale/Format.php(387): Zend_Locale_Math_PhpMath::Sub('0', '', 0)
#1 /vendor/shardj/zf1-future/library/Zend/Locale/Format.php(666): Zend_Locale_Format::toNumber('', Array)
#2 /vendor/shardj/zf1-future/library/Zend/Filter/NormalizedToLocalized.php(110): Zend_Locale_Format::toFloat('', Array)
#3 /lib/Varien/Data/Form/Filter/Date.php(72): Zend_Filter_NormalizedToLocalized->filter('')
#4 /app/code/core/Mage/Eav/Model/Attribute/Data/Abstract.php(191): Varien_Data_Form_Filter_Date->inputFilter('')
#5 /app/code/core/Mage/Eav/Model/Attribute/Data/Date.php(34): Mage_Eav_Model_Attribute_Data_Abstract->_applyInputFilter('')
#6 /app/code/core/Mage/Eav/Model/Form.php(355): Mage_Eav_Model_Attribute_Data_Date->extractValue(Object(Mage_Core_Controller_Request_Http))
#7 /app/code/core/Mage/Adminhtml/controllers/CustomerController.php(623): Mage_Eav_Model_Form->extractData(Object(Mage_Core_Controller_Request_Http), 'account')
...
```

The error is replicated [here](https://onlinephp.io?s=s7EvyCjg5VIpSi0uzSlRsFVQN1BX0FVQV7fm5SooyswriS_SgEpqAoUA&v=8.2.6), the code is traced to `$this->_applyInputFilter($value)` where `$value` is `''`:

```
#5 /app/code/core/Mage/Eav/Model/Attribute/Data/Date.php(34): Mage_Eav_Model_Attribute_Data_Abstract->_applyInputFilter('')
```

So, this PR won't fix issue #3264. However, it is still worthwhile to fix the incorrect docblock.

app/code/core/Mage/Eav/Model/Attribute/Data/Date.php
```diff
@@ -30,7 +30,11 @@ class Mage_Eav_Model_Attribute_Data_Date extends Mage_Eav_Model_Attribute_Data_A
     public function extractValue(Zend_Controller_Request_Http $request)
     {
         $value = $this->_getRequestValue($request);
-        return $this->_applyInputFilter($value);
+        $value = $this->_applyInputFilter($value);
+        if ($value === false) {
+            return '';
+        }
+        return $value;
```
kiatng, line 37: To preserve the original return type and to fix `Fatal error: Uncaught TypeError: Unsupported operand types: string - string`
```suggestion
        return empty($value) ? false : $this->_applyInputFilter($value);
````

fballiano: mmmm but in this case if $value = 0 it returns false, it shouldn't happen since it's a date but wouldn't it be better to use another check instead?

kiatng: `$value` is a `mixed` type. But I suspect it is `string|false`. It is a user input value in a form. So, if `$value == 0`, then user input is invalid, meaning, returning `false` is correct.

fballiano: mmm ok, I can't directly commit your code because of this error:

Screenshot 2023-05-23 alle 22 27 40

I'll commit it from my IDE

kiatng (approved):
Approved although I am not sure about the docblock `array` type in `@return`. The docblock for other methods in the class is consistent about  `$value` being an array while also checking for `false`.

ADDISON74 (approved):

PR #3269, opened 2023-05-17
Fixed PHPStan warnings for branch "next" (PHP 8.1)

fballiano:
The "next" branch now runs on PHP 8.1, this generated a few new PHPStan errors, this PR aims to fix them.

The only weird change that I've done is the removal of the deprecated method `getChildren()` in `app/code/core/Mage/Api/Model/Wsdl/Config/Element.php`, it was deprecated in https://github.com/OpenMage/magento-lts/pull/1613 and, since this is the "next" branch, I think we can remove it.


sreichel:
We should think about to split phpstan-baseline ...

- a base file  ... for all branches
- include additional files for php7/8 specific errors

fballiano:
I don't know if it's possible (since I don't know much about phpstan) but I don't know if it's worth the efforst since when next becomes main then the v20 will be frozen etc etc

app/code/core/Mage/Media/Model/File/Image.php
kiatng, line 1: Shouldn't this be in main branch as well?

lib/Varien/Io/Ftp.php
kiatng, line 1: I think this should be in main branch as well.

app/code/core/Mage/Media/Model/File/Image.php
fballiano: @kiatng https://www.php.net/GdImage that class doesn't exist in php7 so I can't put it in the main branch :-(

kiatng: But it's only docblock.

fballiano: Yes but phpstan doesnt like it vecause the workflows run on php7 so phpstan reports it as a non existing class

kiatng (approved):

elidrissidev (approved):

fballiano:
sorry guys, I had to push the composer.lock because it didn't pass the "validate" check (there was a conflict merging main to next a few days ago).

if all tests pass I'll merge right away without bothering you more, sorry!

PR #3271, opened 2023-05-17
Some PHPStan fixes for the lib/Varien folder

fballiano:
What do you think about these?


lib/Varien/Db/Tree.php
```diff
@@ -65,11 +65,6 @@ public function __construct($config = [])
             // convenience variable
             $db = $config['db'];
 
-            // use an object from the registry?
-            if (is_string($db)) {
-                $db = Zend::registry($db);
-            }
-
```
fballiano, line 72: since Zend class doesn't exist.. this code was never called so it can also be removed

lib/Varien/File/Uploader/Image.php
```diff
@@ -26,7 +26,7 @@ class Varien_File_Uploader_Image extends Varien_File_Uploader
     public function __construct($file = null)
     {
         register_shutdown_function([$this, 'destruct']);
-        $this->newUploader($file);
+        $file = null;
```
fballiano, line 29: if we don't use a parameter phpstan will complain

lib/Varien/Filter/Template.php
```diff
@@ -128,9 +128,6 @@ public function filter($value)
                 foreach ($constructions as $index => $construction) {
                     $replacedValue = '';
                     $callback = [$this, $directive];
-                    if (!is_callable($callback)) {
-                        continue;
-                    }
```
fballiano, line 133: it seems this would always be true

lib/Varien/Db/Tree/NodeSet.php
sreichel, line 1: Nothing returned?

lib/Varien/Db/Tree/NodeSet.php
```diff
@@ -58,13 +58,10 @@ public function valid(): bool
...
 
-    #[\ReturnTypeWillChange]
     public function next()
     {
-        if ($this->_current > $this->_currentNode) {
-            return false;
-        } else {
-            return $this->_current++;
+        if ($this->_current <= $this->_currentNode) {
+            $this->_current++;
```
sreichel, line 64: Nothing returned?

lib/Varien/File/Uploader/Image.php
fballiano: https://www.php.net/manual/en/class.iterator.php says the next() method returns void

sreichel: The interface says it returns void ... that method returned something.

lib/Varien/Db/Tree/NodeSet.php
elidrissidev: > Nothing returned?

Per the `\Iterator` interface, `next()` should only advance the pointer forward.


Have you checked if any code relied on the return type?

sreichel: I did not found any code that directly uses this, but there may be 3rd party code that relies on this.

This changes the "returned" value to `null` for no reason. Isnt `#[\ReturnTypeWillChange]` excactly for this?

fballiano: it doesn't seem the best to implement an interface and then do whatever you want, but anyway phpstan is complaining that that method shouldn't return anything.

lib/Varien/File/Uploader/Image.php
elidrissidev: In this case since `$file` is not passed to parent class, I think it can be removed completely from the constructor.

fballiano: sure but I didn't want to change the signature because in theory it would be a breaking change. actually, for the next branch, this file can be removed because it's not used (as the treenode)

elidrissidev: I already tested it, it doesn't break if the argument is removed even if it's being passed explicitly to the constructor: https://3v4l.org/RDdcO

fballiano: sure sure, but the problem would be about somebody extending that class and getting a "method is not compatible with parent", am i right?

elidrissidev: Doesn't seem to be causing any issue either: https://3v4l.org/1PdNq, anyways I'm sleepy now so I might be missing something 😅

fballiano: @elidrissidev you're 100% right, I've changed it, thanks!

elidrissidev (dismissed):

lib/Varien/Db/Tree.php
```diff
@@ -223,7 +218,7 @@ public function clear($data = [])
 
         try {
             $this->_db->insert($this->_table, $data);
-        } catch (PDOException $e) {
+        } catch (Zend_Db_Adapter_Exception $e) {
```
kiatng, line 221: Can you explain this change? How about line 269, which also catches PDOException?

lib/Varien/Db/Tree/NodeSet.php
```diff
@@ -58,13 +58,10 @@ public function valid(): bool
         return isset($this->_nodes[$this->_current]);
     }
 
-    #[\ReturnTypeWillChange]
     public function next()
```
kiatng, line 61: Also need to be compatible with method `Iterator::next()`:
```suggestion
    public function next(): void
```

image

elidrissidev (approved):

kiatng (approved):

PR #3278, opened 2023-05-25
Run install and upgrade scripts within a db transaction

elidrissidev:
### Description (*)

While investigating the issue I found in https://github.com/OpenMage/magento-lts/pull/2210#issuecomment-1562624227, I had to constantly undo the partial changes made by data upgrade script every time it failed, things like duplicate inserts etc...

I think all install and upgrade scripts should be wrapped in a transaction to prevent the inconvenience of having to reset them.

1. Add a new install or upgrade script to any extension with multiple queries and make it throw an error intentionally at +2nd step.
2. After it throws the error, the changes made by the first step should not be persisted.

### DIFF
--- a/app/code/core/Mage/Core/Model/Resource/Setup.php
+++ b/app/code/core/Mage/Core/Model/Resource/Setup.php
@@ -603,6 +603,7 @@ class Mage_Core_Model_Resource_Setup
             $fileType = pathinfo($fileName, PATHINFO_EXTENSION);
             $this->getConnection()->disallowDdlCache();
             try {
+                $this->getConnection()->beginTransaction();
                 switch ($fileType) {
                     case 'php':
                         $conn   = $this->getConnection();
@@ -624,7 +625,9 @@ class Mage_Core_Model_Resource_Setup
                 if ($result) {
                     $this->_setResourceVersion($actionType, $file['toVersion']);
                 }
+                $this->getConnection()->commit();
             } catch (Exception $e) {
+                $this->getConnection()->rollBack();
                 throw Mage::exception('Mage_Core', Mage::helper('core')->__('Error in file: "%s" - %s', $fileName, $e->getMessage()));
             }
             $version = $file['toVersion'];


fballiano:
this would impact all installations of every module or just the main OM installation?

side note: M2 has a "clean-database" option in the install script to handle these situations

elidrissidev:
Not sure I get what you mean, if you apply this PR, each **new** data/sql install/upgrade script will run within it's own transaction, so if it generates any error changes will be rolled back instead of it remaining partially executed.

fballiano:
I tried installing OM with this patching, breaking one of the install files, but I still got some tables created

Screenshot 2023-05-25 alle 12 18 19

so this only affects the single install/upgrade sqls, which should probably be a good thing anyway.

could there be any impact on 3rd party modules that start transactions on their own?

elidrissidev:
> I tried installing OM with this patching, breaking one of the install files, but I still got some tables created

Yes, a transaction is started for each file separately.

> could there be any impact on 3rd party modules that start transactions on their own?

Good observation. I'm not sure, but from what I see from the DB Adapter, nested transactions are not doing anything besides incrementing the nesting level counter:

https://github.com/OpenMage/magento-lts/blob/e3c88f156185dac795e8dcdf6e7c4249becdf94a/lib/Varien/Db/Adapter/Pdo/Mysql.php#L219-L228

colinmollenhour:
In MySQL, DDL commands cause implicit commit, so I think the transactionLevel counter would get "off" as a result.. So I think the only simple way is for migration scripts to use transactions as needed and as appropriate, although almost nobody does..

elidrissidev:
> In MySQL, DDL commands cause implicit commit, so I think the transactionLevel counter would get "off" as a result.. So I think the only simple way is for migration scripts to use transactions as needed and as appropriate, although almost nobody does..

Ahh that probably explains why it wasn't added from the start, so I guess since it would only make sense to wrap DML queries inside a transaction, this PR is not very useful.

Issue #3281, opened 2023-05-28
Deprecated functionality: parse_url(): 


TangLiang:
> Deprecated functionality: parse_url(): Passing null to parameter #1 ($url) of type string is deprecated in 
 on line 1214

https://github.com/OpenMage/magento-lts/blob/e3c88f156185dac795e8dcdf6e7c4249becdf94a/app/code/core/Mage/Core/Model/Url.php#L1211-L1224

From the 1214 line, if the `http_referer` is empty, the error log will be logged.

TangLiang:
`````````````````````````````````````````````````````
Deprecated functionality: auto_detect_line_endings is deprecated  in /lib/Varien/Io/File.php on line 128

PR #3282, opened 2023-05-28
Removed auto_detect_line_endings since it's deprecated in 8.1

fballiano:
In https://github.com/OpenMage/magento-lts/issues/3281#issuecomment-1565841710 @TangLiang reported that auto_detect_line_endings is nowadays deprecated, more info about it here: https://php.watch/versions/8.1/auto_detect_line_endings-ini-deprecated

### DIFF
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -124,7 +124,7 @@ class Varien_Io_File extends Varien_Io_Abstract
             throw new Exception('Permission denied for write to ' . $this->getFilteredPath($this->_cwd));
         }
 
-        if (!ini_get('auto_detect_line_endings')) {
+        if (PHP_VERSION_ID < 80100 && !ini_get('auto_detect_line_endings')) {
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

kiatng (approved):

elidrissidev:
I would have just removed it altogether since `\r` is so ancient.

fballiano:
I thought of deleting it only in "next" with the other PR that I've posted, just to be safe, is it ok?

elidrissidev:
> I thought of deleting it only in "next" with the other PR that I've posted, just to be safe, is it ok?

Yes

> Note that auto_detect_line_endings directive only enables line-ending detection for \r characters, which are only used in legacy ("classic") Mac OS environments. It is not necessary to enable auto_detect_line_endings for PHP to detect standard \n and \r\n line-endings, which are common and used in Linux, Mac OS, and Windows systems.

fballiano:
oki :-)

PR #3283, opened 2023-05-28
Removed auto_detect_line_endings in Varien_Io_File

fballiano:
This PR is for the `next` branch, where minimum php version is 8.1.

auto_detect_line_endings was deprecated in php 8.1 thus we can remove this line.

Reading https://php.watch/versions/8.1/auto_detect_line_endings-ini-deprecated I think we can completely remove this lines from next `next` branch.

Related PR: https://github.com/OpenMage/magento-lts/pull/3282 for `main` branch.


ADDISON74 (approved):

fballiano:
PHPStan fails because it's still broken on the next branch, but nothing related to this PR

kiatng (approved):

PR #3287, opened 2023-05-30
Removed support for UPS CGI API which was shut down

fballiano:
Past week UPS shut down their old CGI API that openmage also uses in the Mage_Usa module
Screenshot 2023-05-27 alle 22 34 28
so now that part of that module is completely useless and should be removed.

More info in https://github.com/OpenMage/magento-lts/discussions/3280

I'll for sure add a line in the release notes of the next release about this.

Fixes https://github.com/OpenMage/magento-lts/issues/3358


luigifab (commented):
Look good, but after update, without saving the configuration, UPS will not work?

fballiano:
that API is not working since 2-3 weeks ago because UPS shut those servers down, nowaday only the XML version is working

kyrena:
Perhaps consider making `UPS` equivalent to `UPS_XML` for transparent upgrade.

fballiano:
@kyrena they use different sets of authentication credentials so it wouldn't work

luigifab (approved):
Not tested but look good.

fballiano:
the news feed is a good idea, the script I don't think it's needed, if you are using UPS-CGI your website is down since 3 weeks and you have to manually request a new access key and reconfigure everything...

ADDISON74:
We have to review and approve this PR. Check this issue #3358.

ADDISON74 (approved):

PR #3302, opened 2023-06-05
Added instructions for cash on delivery

luigifab:
### Description

Like for `banktransfer` orders, this PR add instructions for `cashondelivery` orders.

Moreover, this PR add a second configuration for instructions for orders.
So you can have two texts, one for quotes, and another one for orders.

If I'm not wrong, added behaviour for `cashondelivery` this is exactly the same as for `banktransfer`.

### Contribution checklist


empiricompany (approved):

ADDISON74:
By default both **Bank Transfer Payment** and **Cash On Delivery Payment** have **Instructions fields**. See bellow

screenshot_4

If these fields are filled in we can see them listed in checkout. 

screenshot_3

This PR comes to add one more field for both payments. What is the purpose?

empiricompany:
by default, cash on delivery instructions are printend only in the checkout and not in the order details/email. 
This PR adds a separate field "istructions_order", but perhaps it's better to keep a single field.

config
immagine

frontend order detail
immagine

backend order detail
immagine

fballiano:
I also think, at first glance, that a single field would be easier

empiricompany (changes requested):
I don't see a reason to add a second field; I would keep a single field for the instructions.

ADDISON74:
The conclusion is correct, there is no need for another field for these payment methods. If the COD information is not displayed on the order editing page and in the email messages sent to the buyer, then the change must be made here.

[OFF TOPIC]
For those who are interested, I've been using an extension for a long time that allows to have my own payment methods, in addition to the already existing ones.

https://github.com/mobweb/MobWeb_MultipleBankTransferPaymentMethods

fballiano:
@ADDISON74 what about we add this multiple banktransfer to the core? and multiple flatrate too? which I've always had to add anyway to all my customers?

fballiano:
@luigifab would it be ok with you to modify the PR and use the same "instructions" field in the order too?

ADDISON74:
@fballiano

1. That extension is very useful for custom payment methods. Implicitly it comes with its section names in the Backend. 

btp

I changed their name because they are not "Bank Transfer Payment" in my case. It can be added into the OpenMage core because it is useful. As section names I would change to "Custom Payment Method 1", "Custom Payment Method 2". I would reduce the number of them to just  two.

2. If the author of a PR does not respond to the request within a reasonable time  that PR can be modified by those who have editing rights. An RFC addendum could be useful.

empiricompany:
> @ADDISON74 what about we add this multiple banktransfer to the core? and multiple flatrate too? which I've always had to add anyway to all my customers?

Personally, I disagree with adding it to the core. I would leave it among the recommended modules. 
But what is the policy for this type of decision? Do I have any voting rights? Otherwise, I won't interfere in these matters anymore..

fballiano:
to add it to the core, there's the need of a PR, which needs votes :-) so you have the right to your opinion as everybody else :-)

ADDISON74:
BTW MobWeb has a lot of nice extensions, some of them I am using in production for many years

https://github.com/mobweb?page=1&tab=repositories

empiricompany:
> BTW MobWeb has a lot of nice extensions, some of them I am using in production for many years
> 
> https://github.com/mobweb?page=1&tab=repositories

wow useful resource!

luigifab:
We are using one text for quotes and another one for orders, so if I update the PR, I'm going to have more work.

app/code/core/Mage/Payment/Block/Info/Cashondelivery.php
```diff
@@ -0,0 +1,53 @@
...
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Payment
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 12:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Payment/Model/Method/Banktransfer.php
```diff
@@ -45,6 +45,12 @@ class Mage_Payment_Model_Method_Banktransfer extends Mage_Payment_Model_Method_A
      */
     public function getInstructions()
     {
+        $paymentInfo = $this->getInfoInstance();
+        if (($paymentInfo instanceof Mage_Sales_Model_Order_Payment) && !empty($text = $this->getConfigData('instructions_order'))) {
+            $order  = $paymentInfo->getOrder();
+            $amount = $order->formatPriceTxt($order->getData('grand_total'));
```
fballiano, line 51: is it possible for this to be null and thus trigger a null parameter warning?

app/design/frontend/base/default/template/payment/info/cashondelivery.phtml
```diff
@@ -0,0 +1,30 @@
...
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category    design
+ * @package     base_default
+ * @copyright   Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright   Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 12:
```suggestion
 * @copyright   Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

fballiano:
@luigifab I see there's a fallback between instructions_order and instructions, which makes be be ok with the PR, although I don't like 100% it doesn't mean I shouldn't approve it.

question: when do the "instructions order" get displayed? on the frontend? in emails? because the name "instructions order" is a bit unclear, maybe we could make it a bit cleared?

luigifab:
Orders instructions are displayed when orders are placed: so in email, in frontend, in backend.

And when there are no orders instructions, quotes instructions are used. There are displayed everywhere: onepage checkout, email, backend, frontend.

fballiano:
> Orders instructions are displayed when orders are placed: so in email, in frontend, in backend.
> 
> And when there are no orders instructions, quotes instructions are used. There are displayed everywhere: onepage checkout, email, backend, frontend.

ok it's now clear, I think that maybe the wording should explain it a bit more

for example the string "instructions on order", @ADDISON74 do you have a clearer wording for that or you think it's easy to understand?

also, below instructions and instructions_order we should have comments explaining what they are for and how the system uses it, @ADDISON74 a couple of easy sentences come to your mind?

ADDISON74:
@luigifab - any reasons for closing this PR?

PR #3307, opened 2023-06-06
Allow System Config of Type Serialized Array to <depend> on other field 

pquerner:
### Description (*)

If you have a serialized array config and want it to "\<depend\>" on some… other field, in the original core this wasn't possible. The js simply didn't work in this scenario and thus the field was always visible.

This branch fixes this, by allowing both: regular (normal fields) and "serialized array" type of fields.

### Related Pull Requests
None

None

1. Find a system config which is type "serialized array" (ie: design/theme/template_ua_regexp)
2. Add this field right above this config temporarly 

```xml
<xx translate="label">
    <label>xx</label>
    <sort_order>10</sort_order>
</xx>
```

3. Add a "\<depends\>" node to its declaration (app/code/core/Mage/Core/etc/system.xml around line 339)

```xml
<depends><xx><value>1</value></xx></depends>
```

### DIFF
--- a/js/mage/adminhtml/form.js
+++ b/js/mage/adminhtml/form.js
@@ -496,7 +496,12 @@ FormElementDependenceController.prototype = {
         }
         return result;
     },
-
+    hideElem : function(ele) {
+        ele.hide();
+    },
+    showElem : function(ele) {
+        ele.show();
+    },
     /**
      * Define whether target element should be toggled and show/hide its row
      *
@@ -507,8 +512,17 @@ FormElementDependenceController.prototype = {
      */
     trackChange : function(e, idTo, valuesFrom)
     {
+        let upLevels = this._config.levels_up;
+        let ele;
         if (!$(idTo)) {
-            return;
+            if ($('row_' + idTo)) {
+                idTo = 'row_' + idTo;
+                ele = $(idTo);
+            } else {
+                return;
+            }
+        } else {
+            ele = $(idTo).up(upLevels);
         }
 
         // define whether the target should show up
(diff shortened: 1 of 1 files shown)


js/mage/adminhtml/form.js
```diff
@@ -536,22 +550,22 @@ FormElementDependenceController.prototype = {
...
-            $(idTo).up(this._config.levels_up).select('input', 'select', 'td').each(function (item) {
+            ele.select('input', 'select', 'td').each(function (item) {
                 // don't touch hidden inputs (and Use Default inputs too), bc they may have custom logic
                 if ((!item.type || item.type != 'hidden') && !($(item.id+'_inherit') && $(item.id+'_inherit').checked)
                     && !(currentConfig.can_edit_price != undefined && !currentConfig.can_edit_price)) {
                     item.disabled = false;
                 }
             });
-            $(idTo).up(this._config.levels_up).show();
+            this.showElem(ele);
```
kiatng, line 560: I feel that `ele.show()` and later `ele.hide()` is cleaner. What do you think?

kiatng:
Tested and worked. I tested like this:

https://github.com/OpenMage/magento-lts/blob/8681773d0e1d6f5c4e2395bbdd9c88f5ddcc8858/app/code/core/Mage/Core/etc/system.xml#L329-L333

After line 333, add
```xml
<depends><template><value>abc</value></template></depends>
```
When template is "abc":
image
When template is not:
image

js/mage/adminhtml/form.js
```diff
@@ -507,8 +512,17 @@ FormElementDependenceController.prototype = {
...
     {
+        let upLevels = this._config.levels_up;
+        let ele;
         if (!$(idTo)) {
-            return;
+            if ($('row_' + idTo)) {
+                idTo = 'row_' + idTo;
+                ele = $(idTo);
+            } else {
+                return;
```
kiatng, line 522:
```suggestion
            idTo = 'row_' + idTo;
            ele = $(idTo);
            if (!ele) {
                return;
```

pquerner: Maybe we should put the 

```
if (!ele) {
    return;
}
```

beneath all, when we sorted which "element" we want to try to attach to.

What do you think?

kiatng: sure.

pquerner: After consideration I just accepted your change-request as-is.

kiatng (approved):

fballiano (approved):

PR #3314, opened 2023-06-09
Fix #3312 ExtJS Tree Checkbox Update

### LINKED ISSUE #3312: ownerTree.container is undefined for product categories
With OpenMage 20.1.0-rc4, when I go to _Products > Any Product > Categories_, at store view level, I get the following error: `ownerTree.container is undefined`

https://github.com/OpenMage/magento-lts/blob/1431628ec5de406b7faf5ba2ed06ea02bcbd233d/js/extjs/ext-tree-checkbox.js#L65-L69

My 10 seconds solution is to replace, line 67/68 by: `var targetNode = n.parentNode ? n.parentNode.ui.getContainer() : (n.ownerTree.container ? n.ownerTree.container.dom : n.ownerTree.innerCt.dom);`

But I'm not sure why, and if it's only me.

justinbeaty:
### Description (*)
PR #2473 upgraded ExtJS from 1.0.1 to 1.1.1, but there was a breaking change in that library. So, the `extjs-tree-checkbox.js` file needed to be updated. The checkbox tree file was not part of ExtJS, but a plugin that somebody wrote and included into Magento. Luckily, the author wrote a comment with the fix.

1. Fixes OpenMage/magento-lts#3312

1. Go to Products > Any Product > Categories, at store view level
2. Should see no errors

### Questions or comments

We have an alternate solution:

```js
var targetNode = 
    n.parentNode ? n.parentNode.ui.getContainer() : 
    (n.ownerTree.container ? n.ownerTree.container.dom : n.ownerTree.innerCt.dom);
```

But I don't think it's necessary, since ExtJS renamed container -> innerCt. The above would only be necessary if we need to support someone who is still using ExtJS 1.0.1.

### DIFF
--- a/js/extjs/ext-tree-checkbox.js
+++ b/js/extjs/ext-tree-checkbox.js
@@ -65,7 +65,7 @@ Ext.extend(Ext.tree.CheckboxNodeUI, Ext.tree.TreeNodeUI, {
     render : function(bulkRender){
         var n = this.node;
         var targetNode = n.parentNode ?
-            n.parentNode.ui.getContainer() : n.ownerTree.container.dom; /* in later svn builds this changes to n.ownerTree.innerCt.dom */
+            n.parentNode.ui.getContainer() : n.ownerTree.innerCt.dom; /* in previous svn builds this was n.ownerTree.container.dom */
         if(!this.rendered){
             this.rendered = true;
             var a = n.attributes;


fballiano (approved):
works!

elidrissidev (approved):

PR #3324, opened 2023-06-16
 Fixed passing null is deprecated for htmlspecialchars in Mage_Core_Helper_Abstract

kyrena:
### Description

This fix `passing null to parameter 1 of type string is deprecated` for `htmlspecialchars` with PHP 8.1/8.2.
Perhaps it's not the best way (I never had this warning when `$addSlashes=true`).
It seems that this appear on backend customer edit page.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Core/Helper/Abstract.php
+++ b/app/code/core/Mage/Core/Helper/Abstract.php
@@ -343,6 +343,7 @@ abstract class Mage_Core_Helper_Abstract
      */
     public function quoteEscape($data, $addSlashes = false)
     {
+        $data = (string)$data;
         if ($addSlashes === true) {
             $data = addslashes($data);
         }
--- a/app/code/core/Mage/Core/Helper/Abstract.php
+++ b/app/code/core/Mage/Core/Helper/Abstract.php
@@ -343,7 +343,9 @@ abstract class Mage_Core_Helper_Abstract
      */
     public function quoteEscape($data, $addSlashes = false)
     {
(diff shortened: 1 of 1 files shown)


kyrena:
lool found the reason: some module `return null` instead of `return ''` for `getTabTitle`...
so perhaps this is useless.

app/code/core/Mage/Core/Helper/Abstract.php
```diff
@@ -343,6 +343,7 @@ public function jsQuoteEscape($data, $quote = '\'')
      */
     public function quoteEscape($data, $addSlashes = false)
     {
+        $data = (string)$data;
```
kiatng, line 346: Is this more efficient?
```suggestion
        if (!$data) {
            return $data;
        }
```

kyrena: Yes, but this is due to a third party code not up to date, so I close this one?

kiatng: I think we still need to fix it.

fballiano:
I've commited @kiatng's version, I think it should be ok right?

fballiano (approved):

app/code/core/Mage/Core/Helper/Abstract.php
fballiano: I think we should close this one. the function is clearly marked to accept a string and there seems there are no misuse of the type inside our code

PR #3326, opened 2023-06-16
Better clearing of session messages after display

kyrena:
### Description

Rarely, I viewed flash notices two times on the same page, because multiple blocks call `<?php echo $this->getMessagesBlock()->getGroupedHtml() ?>`.

So this PR prevent that.

### Contribution checklist


app/code/core/Mage/Core/Block/Messages.php
```diff
@@ -247,6 +247,7 @@ public function getGroupedHtml()
         if ($html) {
             $html .= '</' . $this->_messagesFirstLevelTagName . '>';
         }
+        $this->_messages = null;
```
ADDISON74, line 250: PHPStan is not happy with this statement.

ADDISON74 (approved):

app/code/core/Mage/Core/Block/Messages.php
ADDISON74: In my case I push the wrong button and I approved this PR. It needs more 2 green approves.

The reported issue I encountered a few times over the years.

fballiano, line 250:
```suggestion
        $this->getMessageCollection()->clear();
```

kyrena: Yes, I don't remember if I tested or not this solution before.
Working for now, I continue testing it.

app/code/core/Mage/Core/Block/Messages.php
```diff
@@ -247,6 +247,7 @@ public function getGroupedHtml()
         if ($html) {
             $html .= '</' . $this->_messagesFirstLevelTagName . '>';
         }
+        $this->getMessageCollection()->clear();
```
fballiano, line 250:
```suggestion
        $this->_messages = $this->getMessageCollection()->clear();
```

fballiano (approved):

PR #3328, opened 2023-06-16
Added an easy way to debug objects

kyrena:
### Description

This PR is an old idea to easily debug any `Varien_Object`.

For example in a template: `<?php echo $this ?>` / `<?php echo $product ?>`
When dev mode is enabled, the webpage display:
- `warning_echo_object[Mage_Catalog_Block_Product_View]`
- `warning_echo_object[Mage_Catalog_Model_Product#750]`

Instead of: `Object of class ... could not be converted to string`.
I'm not saying it's very useful.

### Contribution checklist


lib/Varien/Object.php
```diff
@@ -854,4 +854,21 @@ public function flagDirty($field, $flag = true)
...
+     */
+    public function __toString()
+    {
+        if (Mage::getIsDeveloperMode()) {
+            if (PHP_SAPI != 'cli') {
+                return empty($this->getId()) ? '<div><b>warning_echo_object[' . get_class($this) . ']</b></div>' :
+                    '<div><b>warning_echo_object[' . get_class($this) . '#' . $this->getId() . ']</b></div>';
+            }
+            return empty($this->getId()) ? "\n" . 'warning_echo_object[' . get_class($this) . ']' . "\n" :
+                "\n" . 'warning_echo_object[' . get_class($this) . '#' . $this->getId() . ']' . "\n";
```
fballiano, line 870:
```suggestion
            if (PHP_SAPI == 'cli') {
                return empty($this->getId()) ? "\n" . 'warning_echo_object[' . get_class($this) . ']' . "\n" :
                    "\n" . 'warning_echo_object[' . get_class($this) . '#' . $this->getId() . ']' . "\n";
            }
            return empty($this->getId()) ? '<div><b>warning_echo_object[' . get_class($this) . ']</b></div>' :
                '<div><b>warning_echo_object[' . get_class($this) . '#' . $this->getId() . ']</b></div>';
```

I think == ifs are more readable than != one :-)

lib/Varien/Object.php
```diff
@@ -854,4 +854,21 @@ public function flagDirty($field, $flag = true)
         }
         return $this;
     }
+
+    /**
+     * @throw Exception
+     * return string
+     */
+    public function __toString()
```
fballiano, line 862: I like the idea a lot, I was wondering, is it possible that somebody is already casting a varien_object to string for some reason (or mistake) and we're braking something?

luigifab: I suppose that if you cast to string any object that extends `Varien_Object`, you have your custom `__toString()` method? But I saw that PHPStan is unhappy.

fballiano: very weird, doesn't seem to be related to this PR...

pquerner:
I dont see the point in this really.
However, what I can see it adding something like this https://github.com/symfony/var-dumper if you *really* need to debug like that.

fballiano:
ye, I also think that probably the symfony composer is a better idea, and it's for sure more complete and feature rich.

for this reason I'll close this PR, so sorry! 😞

PR #3329, opened 2023-06-16
Fixed passing null for iconv_substr is deprecated in Mage_Catalog_Block_Product_View

kyrena:
### Description

This fix `passing null to parameter 1 of type string is deprecated` for `iconv_substr` with PHP 8.1/8.2.
Perhaps it's not the best way. Only when description of product is empty.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Catalog/Block/Product/View.php
+++ b/app/code/core/Mage/Catalog/Block/Product/View.php
@@ -61,9 +61,12 @@ class Mage_Catalog_Block_Product_View extends Mage_Catalog_Block_Product_Abstrac
             }
             $description = $product->getMetaDescription();
             if ($description) {
-                $headBlock->setDescription(($description));
+                $headBlock->setDescription($description);
             } else {
-                $headBlock->setDescription(Mage::helper('core/string')->substr($product->getDescription(), 0, 255));
+                $description = $product->getDescription();
+                if ($description) {
+                    $headBlock->setDescription(Mage::helper('core/string')->substr($description, 0, 255));
+                }
             }
 
             /** @var Mage_Catalog_Helper_Product $helper */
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/Block/Product/View.php
```diff
@@ -61,9 +61,12 @@ protected function _prepareLayout()
...
             $description = $product->getMetaDescription();
             if ($description) {
-                $headBlock->setDescription(($description));
+                $headBlock->setDescription($description);
             } else {
-                $headBlock->setDescription(Mage::helper('core/string')->substr($product->getDescription(), 0, 255));
+                $description = $product->getDescription();
+                if ($description) {
+                    $headBlock->setDescription(Mage::helper('core/string')->substr($description, 0, 255));
+                }
```
kiatng, line 69: Need to `setDescription()` is empty:
```suggestion 
                    $headBlock->setDescription(Mage::helper('core/string')->substr($description, 0, 255));
               } else {
                    $headBlock->setDescription('');
                }
```

Just in case it's called like this later: https://github.com/OpenMage/magento-lts/blob/2dad20ca0b2345267d7c52090308f216e649e41a/app/design/frontend/base/default/template/page/html/head.phtml#L17-L17

app/code/core/Mage/Catalog/Block/Product/View.php
kiatng, line 1: Taking a second look, I think it should be fixed here `Mage::helper('core/string')->substr()` instead. In fact, all the functions in `Mage_Core_Helper_String` should be checked for the "passing null" error. What do you guys think?

fballiano: I think your last PR is perfect for that.

kyrena: Without 3341 and 3324, with the current PR and some other small futur PR, our instance is deprecated free with PHP 8.2. Crash test in production coming soon. But yes, we do not use all OpenMage features.

kiatng:
Completed in PR #3341.

PR #3336, opened 2023-06-20
Fixed passing null for trim is deprecated in Mage_Catalog_Model_Attribute_Backend_Customlayoutupdate

kyrena:
### Description

This fix `passing null to parameter #1 ($string) of type string is deprecated` for `trim` with PHP 8.1/8.2.
Perhaps it's not the best way.

I get this notice when I save a simple product associated to a configurable product at store view level.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Attribute/Backend/Customlayoutupdate.php
+++ b/app/code/core/Mage/Catalog/Model/Attribute/Backend/Customlayoutupdate.php
@@ -32,7 +32,7 @@ class Mage_Catalog_Model_Attribute_Backend_Customlayoutupdate extends Mage_Eav_M
     public function validate($object)
     {
         $attributeName = $this->getAttribute()->getName();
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/Model/Attribute/Backend/Customlayoutupdate.php
kiatng, line 1: How about fix it here: https://github.com/OpenMage/magento-lts/blob/bda0b6f3fe253abf0188fa3a02f1dcd7d7c9a4d8/app/code/core/Mage/Eav/Model/Entity/Attribute/Abstract.php#L177-L180

kyrena: The problem is not the name of the attribute, but the attribute value.

kiatng (approved):

fballiano (approved):

PR #3338, opened 2023-06-20
Fixed eav_config_entity throwing error message on removed entity model

davidhiendl:
A fix for https://github.com/OpenMage/magento-lts/issues/3175 ensuring that the eav entity type class is present before proceeding to load it to gracefully handle removed modules and/or leftover data.

Purposely throws exception in developer mode to highlight the leftover data and only logs a statement once every time data is loaded from the database.

### Related Pull Requests
- Introduced with https://github.com/OpenMage/magento-lts/pull/2993 cache/performance optimization due to eagerly loading all data 

- https://github.com/OpenMage/magento-lts/issues/3175 

1. Disable eav cache
2. Add entry to eav_entity_type table with non-existent model class

### DIFF
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -198,7 +198,7 @@ class Mage_Eav_Model_Config
                 if (Mage::getIsDeveloperMode()) {
                     throw new Exception('Failed loading of eav entity type because it does not exist: ' . $entityModelClass);
                 } else {
-                    Mage::log('skipped loading of eav entity type because it does not exist: ' . $entityModelClass);
+                    Mage::log('Skipped loading of eav entity type because it does not exist: ' . $entityModelClass);
                 }
                 continue;
             }
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -196,7 +196,7 @@ class Mage_Eav_Model_Config
             $fqEntityModelClass = Mage::getConfig()->getModelClassName($entityModelClass);
             if (!class_exists($fqEntityModelClass)) {
                 if (Mage::getIsDeveloperMode()) {
-                    throw new Exception('failed loading of eav entity type because it does not exist: ' . $entityModelClass);
+                    throw new Exception('Failed loading of eav entity type because it does not exist: ' . $entityModelClass);
                 } else {
                     Mage::log('skipped loading of eav entity type because it does not exist: ' . $entityModelClass);
                 }
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -189,7 +189,6 @@ class Mage_Eav_Model_Config
 
         /** @var Mage_Eav_Model_Entity_Type $entityType */
         foreach ($entityTypeCollection as $entityType) {
-
             // Ensure eav entity type model class is defined, otherwise skip processing it.
             // This check prevents leftover eav_entity_type entries from disabled/removed modules creating errors and
             // is necessary because the entire EAV model is now loaded eagerly for performance optimization.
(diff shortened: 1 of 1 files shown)


davidhiendl:
Please note I only fixed the eager loading of entity types causing issues, eager loading of attributes with non-existent models always caused issues in the past as well (eg. deleting a module with a class that was used as custom attribute source_model or frontend/backend type model always caused issues in the past, removing those checks would be quite extensive and I at least prefer to be warned "early" of such issues).

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -186,8 +186,24 @@ protected function _loadEntityTypes()
         $this->_entityTypes = [];
         $this->_entityTypeByCode = [];
         $entityTypeCollection = Mage::getResourceModel('eav/entity_type_collection');
+
         /** @var Mage_Eav_Model_Entity_Type $entityType */
         foreach ($entityTypeCollection as $entityType) {
+
```
fballiano, line 192:
```suggestion
```

phpcs is complaining

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -186,8 +186,23 @@ protected function _loadEntityTypes()
...
+            $fqEntityModelClass = Mage::getConfig()->getModelClassName($entityModelClass);
+            if (!class_exists($fqEntityModelClass)) {
+                if (Mage::getIsDeveloperMode()) {
+                    throw new Exception('failed loading of eav entity type because it does not exist: ' . $entityModelClass);
+                } else {
+                    Mage::log('skipped loading of eav entity type because it does not exist: ' . $entityModelClass);
+                }
+                continue;
+            }
+
```
ADDISON74, line 205: There should be a capital letter at the beginning of the exception messages to keep the rule we already have. The message that starts with a lowercase letter can appear as a phrase added to another message.

ADDISON74 (changes requested):

app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -186,8 +186,23 @@ protected function _loadEntityTypes()
...
+            // Ensure eav entity type model class is defined, otherwise skip processing it.
+            // This check prevents leftover eav_entity_type entries from disabled/removed modules creating errors and
+            // is necessary because the entire EAV model is now loaded eagerly for performance optimization.
+            $entityModelClass = $entityType['entity_model'];
+            $fqEntityModelClass = Mage::getConfig()->getModelClassName($entityModelClass);
+            if (!class_exists($fqEntityModelClass)) {
+                if (Mage::getIsDeveloperMode()) {
+                    throw new Exception('Failed loading of eav entity type because it does not exist: ' . $entityModelClass);
+                } else {
+                    Mage::log('skipped loading of eav entity type because it does not exist: ' . $entityModelClass);
```
ADDISON74, line 201: skipped => Skipped

ADDISON74 (changes requested):

app/code/core/Mage/Eav/Model/Config.php
davidhiendl: Oh, because of your comment I thought it only applies to exception message.

sreichel:
Accidentally deleted all my VBox images and had no time to recreate.

Can't test it during next days.

ADDISON74:
@sven - I will make time these days to simulate your case. DDEV in my case requests only a couple of minutes to run an OM instance from scratch. I will install an extension then delete it from the file system. I should get your error, then I checkout this PR to see what happens. Do I need to take other steps to reproduce the issue?

sreichel:
@ADDISON74 thanks. I know it takes only minutes to setup OM, but I've to do some more things. Need a new HDD, update my whole system from Win7 (do not laugh 😂), but the most important ... my beloved dog is 9y now and I just want to spend as much time as possible with her.

To reproduce ... install the extension you sent me and remove all files. No further steps.

davidhiendl:
It's much easier to simple insert a new entry into the eav_entity_type table with a non-existent model class.

ADDISON74:
I confirm this PR solves the issue reported here #3175.

ADDISON74 (approved):

fballiano (approved):

PR #3340, opened 2023-06-20
Fixed passing null for number_format is deprecated for downloadable products

kyrena:
### Description

This fix `passing null to parameter #1 ($num) of type float is deprecated` for `number_format` with PHP 8.1/8.2.
Perhaps it's not the best way.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
+++ b/app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
@@ -151,18 +151,19 @@ class Mage_Downloadable_Block_Adminhtml_Catalog_Product_Edit_Tab_Downloadable_Li
         $links = $productType->getLinks($this->getProduct());
         $priceWebsiteScope = Mage::helper('downloadable')->getIsPriceWebsiteScope();
         foreach ($links as $item) {
+            $price = $item->getPrice();
             $tmpLinkItem = [
-                'link_id' => $item->getId(),
-                'title' => $this->escapeHtml($item->getTitle()),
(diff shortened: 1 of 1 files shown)


fballiano (approved):

app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php
kiatng, line 1: Is it better to fix it here: https://github.com/OpenMage/magento-lts/blob/bda0b6f3fe253abf0188fa3a02f1dcd7d7c9a4d8/app/code/core/Mage/Downloadable/Block/Adminhtml/Catalog/Product/Edit/Tab/Downloadable/Links.php#L228-L230

fballiano: how would you do that? cause if $value is null I don't see what we should return

kiatng: Ref https://onlinephp.io/c/5e8d8, when `$value = null`, it used to return `'0.00'`. Also, see the docblock:

```php
   /**
     * Return formatted price with two digits after decimal point
     *
     * @param float $value
     * @return string
     */
    public function getPriceValue($value)
    {
       // suggestion:
        if (!$value) {
            return '0.00';
        }
        return number_format($value, 2, null, '');
    }
```

Note that when `$value = ''`: we have

> Fatal error: Uncaught TypeError: number_format(): Argument #1 ($num) must be of type float, string given in /home/user/scripts/code.php:3
Stack trace:
#0 /home/user/scripts/code.php(3): number_format('', 2, NULL, '')

PR #3345, opened 2023-06-23
Fixed implicit conversion from float to int in Mage_SalesRule_Model_Validator

kyrena:
### Description

This PR fix `Implicit conversion from float 10.1 to int loses precision` with PHP 8.1/8.2, with a coupon code of 10.1% for all items of cart.

### Contribution checklist


fballiano:
Actually I'm not sure I get the reason behind this PR, why cast to 10 if the discount is 10.1? I understand that 10.1 is an unusual number but remove the possibility for decimal discounts doesn't seem the way to go, to me.

app/code/core/Mage/SalesRule/Model/Validator.php
```diff
@@ -465,10 +465,10 @@ public function process(Mage_Sales_Model_Quote_Item_Abstract $item)
             $discountAmount = $result->getDiscountAmount();
             $baseDiscountAmount = $result->getBaseDiscountAmount();
 
-            $percentKey = $item->getDiscountPercent();
             /**
              * Process "delta" rounding
              */
+            $percentKey = (int)$item->getDiscountPercent();
```
fballiano, line 471:
```suggestion
            $percentKey = (string) $item->getDiscountPercent();
```

fballiano:
So, I checked that part of the core, I think casting to string is much better, that variable is only used as an array_key, so as a string we risk less conflict if a cart as 2 discounts of the same value (at least for the integer part).

I think the core is wrong in this part, they shouldn't have used the discount percent as a key, but the rule id, or something unique.

app/code/core/Mage/SalesRule/Model/Validator.php
kiatng: Isn't `$item->getDiscountPercent()` actually returns type string already? So, this won't do anything.

kiatng:
Is this PR fixing a bug?

app/code/core/Mage/SalesRule/Model/Validator.php
fballiano: getDiscountPercent() returns a float
Screenshot 2023-07-14 alle 08 13 36

elidrissidev: I think the `@method` annotations are misleading here, it does indeed return a `string`:

Screen Shot 2023-07-14 at 12 11 56 PM

fballiano: mmmm could be, but then I think @kyrena opened this because they had the error of their systems

fballiano: and since, as a discount percent, it would be correct to have a float, but then it's used as an array key, I think we could force cast it to string just in case :-D

elidrissidev: Yes of course, `string` is the way to go.

fballiano (approved):

elidrissidev (approved):

PR #3346, opened 2023-06-23
 Fixed foreach argument must be of type array with API

kyrena:
### Description 

Not sure if it's due to PHP 8.1/8.2 or not, but this appear with the following examples:

```php
// v1
$proxy = new SoapClient('https://example.org/api/soap/?wsdl', ['trace' => 1, 'user_agent' => 'Local/Test']);
$sessionId = $proxy->login('covid', $pass);
$start = microtime(true);
$oiid = 123456789;
var_dump($proxy->call($sessionId, 'sales_order.info', $oiid));
var_dump($proxy->call($sessionId, 'sales_order_invoice.create', $oiid)); // HERE
$siid = $proxy->call($sessionId, 'sales_order_shipment.create', $oiid); // HERE
var_dump($siid);
var_dump($proxy->call($sessionId, 'sales_order_creditmemo.create', $oiid));
var_dump($proxy->call($sessionId, 'sales_order_shipment.addTrack', [$siid, 'custom', 'Test apiv1', '123', 5]));
echo microtime(true) - $start,"\n";

// v2
$proxy = new SoapClient('https://example.org/api/v2_soap/?wsdl', ['trace' => 1, 'user_agent' => 'Local/Test']);
$sessionId = $proxy->login('covid', $pass);
$start = microtime(true);
$oiid = 123456789;
var_dump($proxy->salesOrderInfo($sessionId, $oiid));
var_dump($proxy->salesOrderInvoiceCreate($sessionId, $oiid)); // HERE
$siid = $proxy->salesOrderShipmentCreate($sessionId, $oiid); // HERE
var_dump($siid);
var_dump($proxy->salesOrderCreditmemoCreate($sessionId, $oiid));
var_dump($proxy->salesOrderShipmentAddTrack($sessionId, $siid, 'custom', 'Test apiv2', '123', 5));
echo microtime(true) - $start,"\n";

```

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Sales/Model/Order/Invoice/Api/V2.php
+++ b/app/code/core/Mage/Sales/Model/Order/Invoice/Api/V2.php
@@ -33,12 +33,17 @@ class Mage_Sales_Model_Order_Invoice_Api_V2 extends Mage_Sales_Model_Order_Invoi
      */
     public function create($invoiceIncrementId, $itemsQty = [], $comment = null, $email = false, $includeComment = false)
     {
+        if (!is_array($itemsQty)) {
+            $itemsQty = [];
+        }
+
+        /** @var Mage_Sales_Model_Order $order */
         $order = Mage::getModel('sales/order')->loadByIncrementId($invoiceIncrementId);
         $itemsQty = $this->_prepareItemQtyData($itemsQty);
-        /** @var Mage_Sales_Model_Order $order */
+
         /**
-          * Check order existing
-          */
+         * Check order existing
+         */
         if (!$order->getId()) {
             $this->_fault('order_not_exists');
         }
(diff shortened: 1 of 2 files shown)


fballiano:
I guess it's some custom code or 3rd party module that's calling these methods with wrong data. I don't know, to force the itemQty to an empty array could avoid the error but... the methods shouldn't be called with uniterable data, IMHO it would be better to throw an exception.

elidrissidev:
I would say there needn't be any check there, the argument is documented to accept an `array`, and doing type checking for every argument doesn't seem that clean IMO.

One thing I'd like to check, why do the API functions receive `null` as `$itemQty` even though it's not send from the client and it has a default value in the method signature??

fballiano:
I do agree with @elidrissidev, we cannot check everything, and if the problem comes from a 3rd party module, it should be fixed there in this case.

I'll close this one at the moment, if any other info becomes available we'll check on it for sure :-)

kyrena:
For creditmemo there is a `null => []` via _prepareCreateData.

https://github.com/OpenMage/magento-lts/blob/3e7be1102e6cd0d33913daae9788a21e726efea0/app/code/core/Mage/Sales/Model/Order/Creditmemo/Api/V2.php#L62-L68


https://github.com/OpenMage/magento-lts/blob/3e7be1102e6cd0d33913daae9788a21e726efea0/app/code/core/Mage/Sales/Model/Order/Invoice/Api/V2.php#L83-L86

PR #3347, opened 2023-06-23
Removed Google Analytics3 and Universal Analytics

fballiano:
Since Google is discontinuing Analytics3 and UniversalAnalytics (https://support.google.com/analytics/answer/11583528?hl=en) I thought that maybe this PR could be ok, removing almost everything related to those 2 services.

I thought to target `main`, we could target `next` instead but I don't think it's really necessary since these features will be useless anyway in a few days.

I've also removed the install script since it was forcing the usage of analytics3, which would be wrong anyway.


fballiano:
If this gets approved I'd like to add GoogleTagManager to OM.

fballiano:
this PR and https://github.com/OpenMage/magento-lts/pull/3347#issuecomment-1604468487 are important cause at the moment website with GTM can't have enhanced ecommerce data with analytics4

ADDISON74 (approved):

roby255 (approved):

fballiano:
@empiricompany what do you think?

empiricompany:
> @empiricompany what do you think?

I agree to discontinue the unsupported analytics platform. GTM is just a simple script that can be added in configuration design html scripts, it doesn't do anything on its own. Instead, I wanted to add support for the AdWords tag for conversion tracking.

empiricompany (approved):

fballiano:
yes, but I want to have GTM/analytics selection and not just ga4, since I've already 2 customers that have GTM and, with rc5, can't really use enhanced ecommerce data :-)

thank you anyway, I'll work on a PR ASAP and let's see!

fballiano:
merging per 1green + 2grays, thank you!

fballiano:
@empiricompany I wanted to have the GTM container instead of the ga4 script, then call the gtag funcion for the enhanced ecommerce. it should work if the gtag is injected by GTM no? am I missing something?

empiricompany:
> @empiricompany I wanted to have the GTM container instead of the ga4 script, then call the gtag funcion for the enhanced ecommerce. it should work if the gtag is injected by GTM no? am I missing something?

yes, then we have to simple add new option to inject gtm instead of gtag

PR #3355, opened 2023-06-27
Fixed getQuote() returns null in quote item and quote address objects.

kiatng:
### Description (*)
Example:
```php
$quoteItem = Mage::getModel('sales/quote_item')->load($itemId);
$quote = $quoteItem->getQuote(); // null value but it should not be!
```

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote/Address.php
+++ b/app/code/core/Mage/Sales/Model/Quote/Address.php
@@ -427,6 +427,9 @@ class Mage_Sales_Model_Quote_Address extends Mage_Customer_Model_Address_Abstrac
      */
     public function getQuote()
     {
+        if (is_null($this->_quote)) {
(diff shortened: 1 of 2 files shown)


ADDISON74 (approved):

fballiano (approved):

pquerner:
Mh, maybe sometimes you dont want the object appended automatically. Shouldnt you be able to chose this?

kiatng:
> Mh, maybe sometimes you dont want the object appended automatically. Shouldnt you be able to chose this?

There is a `setQuote()` right above it:
https://github.com/OpenMage/magento-lts/blob/c830ba5a1e686ef3faecab5d3406fc1ec261e80a/app/code/core/Mage/Sales/Model/Quote/Address.php#L414-L435

pquerner:
Thanks! I thought this would auto-append it at load-times. My bad. :)

PR #3356, opened 2023-06-27
Added support for Google Tag Manager injection, compatible with Analytics 4 and enhanced ecommerce data

fballiano:
This PR does a minor refactor of some GA4 code and adds the support for the injection of Google Tag Manager.

### A screenshot of the backend
Screenshot 2023-06-27 alle 23 53 25

### Tests I've done
- GA4 alone: working and sending enhanced ecommerce
- GTM injecting GA4: working and sending enhanced ecommerce
- GTM not injecting GA4 working and sending enhanced ecommerce

If you check the code, you may thing that GA4 scripts will be added to the HTML also when GTM is injecting them. I've done multiple tests but GTM is somehow detecting it and not injecting GA4 if it's already in the HTML. So, I think we could leave it like this.

### Notes
I've renamed the _getOrdersTrackingCodeAnalytics4() method to _getEnhancedEcommerceDataForAnalytics4() since it's really not about order tracking, it does much more than that, and it's a new method so it's not a problem to rename it.

### Questions
- Should I have done a new helper instead of using the default one?

### DIFF
--- a/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
@@ -109,15 +109,10 @@ gtag('set', 'user_id', '{$customer->getId()}');
      *
      * @return string
      * @throws Mage_Core_Model_Store_Exception
+     * @deprecated
      */
     protected function _getOrdersTrackingCode()
     {
-        /** @var Mage_GoogleAnalytics_Helper_Data $helper */
-        $helper = $this->helper('googleanalytics');
-        if ($helper->isUseAnalytics4()) {
-            return $this->_getOrdersTrackingCodeAnalytics4();
-        }
-
         return '';
     }
 
@@ -127,7 +122,7 @@ gtag('set', 'user_id', '{$customer->getId()}');
      * @return string
      * @throws JsonException
      */
-    protected function _getOrdersTrackingCodeAnalytics4()
+    protected function _getEnhancedEcommerceDataForAnalytics4()
     {
         $result = [];
         $request = $this->getRequest();
--- a/app/code/core/Mage/GoogleAnalytics/Block/Gtm.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Gtm.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_GoogleAnalytics
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2022 The OpenMage Contributors (https://www.openmage.org)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * GoogleTagManager Page Block
+ *
+ * @category   Mage
+ * @package    Mage_GoogleAnalytics
+ */
+class Mage_GoogleAnalytics_Block_Gtm extends Mage_Core_Block_Template
+{
+    /**
+     * @return bool
+     */
+    protected function _isAvailable()
+    {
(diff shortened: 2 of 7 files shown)


roby255 (approved):

fballiano:
@empiricompany what do you think about this one?

empiricompany:
> @empiricompany what do you think about this one?

good! i suggest this changes to better integrate with GTM that already include ga4 script

check if Google Analytics 4 -> Account Number, if not set prevent to inject duplicated script
separate useid tracking like enhaced ecommerce to keep it out of the condition

app\design\frontend\base\default\template\googleanalytics\ga.phtml
```
<?php
/** @var Mage_GoogleAnalytics_Block_Ga $this */
$_helper = $this->helper('googleanalytics');
?>
<?php if (!$this->helper('core/cookie')->isUserNotAllowSaveCookie()): ?>
    <?php if ($_helper->isUseAnalytics4()): ?>
        <?php $_gaAccountId = $_helper->getAccountId() ?>
        <!-- BEGIN GOOGLE ANALYTICS 4 CODE -->
        <?php if ($_gaAccountId) : ?>
        <script async src="https://www.googletagmanager.com/gtag/js?id=<?= $_gaAccountId ?>"></script>
        <?php endif; ?>
        <script>
            window.dataLayer = window.dataLayer || [];
            function gtag(){dataLayer.push(arguments);}
            <?php if ($_gaAccountId) : ?>
            <?php echo $this->_getPageTrackingCode($_gaAccountId) ?>
            <?php endif; ?>

            <?php echo $this->_getUserIdTrackingCodeForAnalytics4() ?>

            <?php echo $this->_getEnhancedEcommerceDataForAnalytics4() ?>
        </script>
        <!-- END GOOGLE ANALYTICS 4 CODE -->
    <?php endif ?>
<?php endif ?>
```

app\code\core\Mage\GoogleAnalytics\Block\Ga.php

  ```
  /**
     * Send information about User ID
     *
     * @return string
     */
    protected function _getUserIdTrackingCodeForAnalytics4()
    {
        //add user_id
        if ($this->helper('googleanalytics')->isUserIdEnabled() && Mage::getSingleton('customer/session')->isLoggedIn()) {
            $customer = Mage::getSingleton('customer/session')->getCustomer();
            $trackingCode.= "
gtag('set', 'user_id', '{$customer->getId()}');
";
        }
        return $trackingCode;
    }
```


add some comment in configuration fields like:
immagine

what do you think? 
maybe is better to add a new configuration yes/no in GTM like "GTM include ga4 script?" then check it in the code? 
or change yes/no configuration in Google Analytics->Enable like this:

- No
- Yes
- Yes, but only compile dataLayer (this hide field Account Number)
?

I believe that GTM detects if the GA4 script is already present, but it would still result in an additional third-party script loading.

fballiano:
@empiricompany 
I tested all what you write before publishing the PR :-)

because of the way the 2 scripts are conceived by google, the GA4 one is requested BEFORE the GTM one, so if we don't include the GA4 external script, the last part of js code you wrote will trigger an error.

anyway there's no 3rd script loaded, GTM doesn't send the GA4 one if it's already loaded, that's what I was meaning with my original description, here's the proof

Screenshot 2023-06-29 alle 13 18 36

empiricompany:
> @empiricompany 
> I tested all what you write before publishing the PR :-)
> 
> because of the way the 2 scripts are conceived by google, the GA4 one is requested BEFORE the GTM one, so if we don't include the GA4 external script, the last part of js code you wrote will trigger an error.
> 
> anyway there's no 3rd script loaded, GTM doesn't send the GA4 one if it's already loaded, that's what I was meaning with my original description, here's the proof
> 
> Screenshot 2023-06-29 alle 13 18 36
> 
> 

ah ok so there is no problem, but at this point gtm script can be pasted in various script in design although is more comfortable to have a specific configure, initially I have understanded that you want to insert an alternative to ga4 like old universal not a separate script
for me is ok

empiricompany (approved):

fballiano:
initially I thought we needed it "in alternative" to ga4, but then, during developing, it doesn't work. yes it could be pasted elsewhere but we may do some further development if we have it here, and also it gives the users the perception of a much more integrated feature

ADDISON74 (approved):
LGTM

fballiano:
merging per 1 green + 2 gray

PR #3363, opened 2023-07-03
Allowed customization of the checkout page identifiers for the Mage_GoogleAnalytics module

fballiano:
While implementing the new GA4 feature for a customer I found out that, with a custom checkout, it doesn't track the begin_checkout event.

So I thought to create a couple of constants that can be easily overridden to allow for this feature to work for any checkout, in fact, for my customer I've simply create a class like this:

```php
<?php
 
class Customernme_Analytics4_Block_GoogleAnalytics_Ga extends Mage_GoogleAnalytics_Block_Ga
{
    protected const CHECKOUT_MODULE_NAME = "firecheckout";
    protected const CHECKOUT_CONTROLLER_NAME = "index";
}
```

and now it tracks the checkout event.

there are probably better ways, any suggestion is welcome.

### DIFF
--- a/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
@@ -21,6 +21,9 @@
  */
 class Mage_GoogleAnalytics_Block_Ga extends Mage_Core_Block_Template
 {
+    protected const CHECKOUT_MODULE_NAME = "checkout";
+    protected const CHECKOUT_CONTROLLER_NAME = "onepage";
+
     /**
      * Render regular page tracking javascript code
      * The custom "page name" may be set from layout or somewhere else. It must start from slash.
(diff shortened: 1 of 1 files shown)


fballiano:
@empiricompany since you are one of the author and for sure you use this feature, what do you think about this?

empiricompany:
> @empiricompany since you are one of the author and for sure you use this feature, what do you think about this?

well done! thanks

empiricompany (approved):

empiricompany:
> there are probably better ways, any suggestion is welcome.

probably a better way is to introduce new events then attach to it such as initcheckout, viewcart, etc..but we can improve it in the future

ADDISON74 (approved):

roby255 (approved):

fballiano:
@empiricompany yes at least we should have an event to be able to modify the json data that is being sent to analytics, for sure we'll add more stuff to this code in the near future, but thanks a lot at the moment!

fballiano:
merging per 2gray and 1 green, thanks!

PR #3365, opened 2023-07-05
Clear config cache after maintenance check for DB Update

Hanmac:
### Description (*)

if using `maintenance.ip` to bypass `maintenance.flag`, it stops DB Updates. This should fix it.

### Related Pull Requests

Fixes DB Update problem from #1449

1. use `maintenance.flag` with `maintenance.ip`
2. have an Magento Module that still has DB Update scripts to run
3. try to clean up the Cache to make Magento run the DB Updates
4. This currently fails because the does already exist (See Description on  #1449 )

`App::init` does create the Config Cache, so when it `App::run` is called with `_initModules` and `loadModulesCache`, it sees the Cache Config Cache is already created and doesn't ask again.
The added Lines do clear up the part of this specific config does, so it calls `Mage_Core_Model_Resource_Setup::applyAllUpdates`

### DIFF
--- a/index.php
+++ b/index.php
@@ -51,7 +51,8 @@ if (file_exists($maintenanceFile)) {
         include_once __DIR__ . '/errors/503.php';
         exit;
     }
-    // need to remove Config Cache make it check for DB updates
+
+    // remove config cache to make the system check for DB updates
     $config = Mage::app()->getConfig();
     $config->getCache()->remove($config->getCacheId());
 }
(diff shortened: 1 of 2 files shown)


Hanmac:
the maintenance files should be protected in the htaccess
but i don't know if that is the case yet.

yes the User that sets up the maintenance files should take care of them so they are removed when the maintenance is over

index.php
```diff
@@ -51,6 +51,9 @@
         include_once __DIR__ . '/errors/503.php';
         exit;
     }
+    // need to remove Config Cache make it check for DB updates
```
fballiano, line 54:
```suggestion

    // remove config cache to make the system check for DB updates
```

fballiano (approved):

PR #3366, opened 2023-07-06
Added decimal/thousand separator specification when sending Analytics4 data

fballiano:
when product price is higher than a thousand EUR, it gets sent to analytics this way:

`"currency":"EUR","value":"4,245.42"`

This PR removes the thousand separator, so that it gets sent this way:

`"currency":"EUR","value":"4245.42"`

This PR also adds "quantity" to the begin_checkout data, which was missing.

### DIFF
--- a/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
@@ -320,6 +320,7 @@ gtag('set', 'user_id', '{$customer->getId()}');
                         'item_id' => $_product->getSku(),
                         'item_name' => $_product->getName(),
                         'price' => number_format($_product->getFinalPrice(), 2, '.', ''),
+                        'quantity' => (int) $productInCart->getQty(),
                     ];
                     if ($_product->getAttributeText('manufacturer')) {
                         $_item['item_brand'] = $_product->getAttributeText('manufacturer');
--- a/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
+++ b/app/code/core/Mage/GoogleAnalytics/Block/Ga.php
@@ -143,12 +143,12 @@ gtag('set', 'user_id', '{$customer->getId()}');
                 $_removedProduct = Mage::getModel('catalog/product')->load($removedProduct);
                 $eventData = [];
                 $eventData['currency'] = Mage::app()->getStore()->getCurrentCurrencyCode();
-                $eventData['value'] = number_format($_removedProduct->getFinalPrice(), 2);
+                $eventData['value'] = number_format($_removedProduct->getFinalPrice(), 2, '.', '');
                 $eventData['items'] = [];
                 $_item = [
                     'item_id' => $_removedProduct->getSku(),
                     'item_name' => $_removedProduct->getName(),
-                    'price' => number_format($_removedProduct->getFinalPrice(), 2),
+                    'price' => number_format($_removedProduct->getFinalPrice(), 2, '.', ''),
                 ];
                 if ($_removedProduct->getAttributeText('manufacturer')) {
                     $_item['item_brand'] = $_removedProduct->getAttributeText('manufacturer');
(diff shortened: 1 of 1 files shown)


fballiano:
@empiricompany

pquerner:
Makes we wonder why we still not send the 'correct' data type to google. What if we take a look at other shops GA4 implementation and see what we can see?

fballiano:
why are you saying we don't send the correct data types?

pquerner:
Do you remember #3228 ?

fballiano:
If only I had the memory for all the PRs 😅 anyway that's been solved, right?

pquerner:
Yes it was, although the issue is still present and we are working on good faith that the data gets somewhere parsed correctly. Either in js, or in the transfer to google, or at google before it gets to a "event status". I dont know how to explain.

But this purchase data for instance, we still send the `value` key with "number_format" (a string) instead of the wanted "Number" (whatever that means, thanks google, but it surely doesnt mean "String", because otherwise it would be stated like that):

image

fballiano:
at the end of the day we're sending a json so the typing is kinda loose. yes the values are within quotes but I don't think that's a problem for google. and I can't find any documentation for that "Number" data type.

pquerner:
Yes, but we also don't have examples (do we?) by Google that shows how to properly seperate prices in their thousands or more. So again, good faith that it works. Right?

fballiano:
there are some examples in the doc, but the prices shown are smaller than a thousand. I am sure that the number_format of this PR works, the "," as thousand separator I've many doubts.

empiricompany:
i'm testing the two formats version

`gtag('event', 'purchase', {"currency":"USD","transaction_id":"302000011","value":"4,245.42","coupon":"","shipping":"0.00","tax":"0.00","items":[{"item_id":"test-over1000","item_name":"test over 1000","quantity":1,"price":"4,245.42","discount":"0.00","item_category":"New Arrivals"}]});`

`gtag('event', 'purchase', {"currency":"USD","transaction_id":"302000012","value":"4245.42","coupon":"","shipping":"0.00","tax":"0.00","items":[{"item_id":"test-over1000","item_name":"test over 1000","quantity":1,"price":"4245.42","discount":"0.00","item_category":"New Arrivals"}]});`

tomorrow i can say if change something / is irrilevant / or whatever..

i don't think google check data type in json if is float or string

pquerner:
There is a [debug mode ](https://developers.google.com/analytics/devguides/collection/ga4/set-up-ecommerce#verify-data) but you will see that the data looks a-like.

empiricompany:
once again, the debug mode is not reliable. The only way to verify that the data is being interpreted correctly is to wait for 24 hours and check how the data is being read by Analytics -> Reports or with Looker Studio

anyway...
immagine
immagine


immagine
immagine

fballiano:
There's another problem here, the "add_product_to_cart" doesn't pass the added quantity, there are cases where the website has a minimum quantity, so the current implementation is wrong.

empiricompany:
> There's another problem here, the "add_product_to_cart" doesn't pass the added quantity, there are cases where the website has a minimum quantity, so the current implementation is wrong.

but sales_quote_add_item event is fired if minimium qty is not met?

track qty in add_to_cart is not supported in current implementation cause the observer save only product id that will be catched in observer
if we want to track add_to_cart qty for product we have to refactor it
from a quick view in observer we can save an array like sku => qty to track qty

fballiano:
it doesn't actually matter if it's fired of not, the observer doesn't carry the quantity data so I think google assumes it's "1", which is wrong

empiricompany:
> it doesn't actually matter if it's fired of not, the observer doesn't carry the quantity data so I think google assumes it's "1", which is wrong

ok so a possible solution is to store in session key pairs
sku => qty

fballiano:
I think it would be better to store the quote_item, it least we don't need another catalog/product->load()

also, that event is not fired correctly for the project I'm working on (ajax add to cart modules), so I'd explore checkout_cart_product_add_after and/or sales_quote_product_add_after

empiricompany:
so seems to be correct remove thousand separator

#302000011: "value":"4,245.42"
#302000012: "value":"4,245.42"

immagine

PS: sorry for screen in Italian, i can't change the language

empiricompany (approved):

ADDISON74 (approved):

pquerner (approved):

PR #3367, opened 2023-07-07
New feature to repopulate product form in frontend on errors.

kiatng:
### Description (*)
For products with custom options, when validation fails and the page is redirected to the product, all inputs would be lost. The user would need to reenter all the fields. With this PR, the product form will be repopulated thereby improving the user experience. Tested for all product custom option types except file. File upload is not applicable because it's not possible to re-attach the file in the browser.

1. [backend] Create a simple product, set _Maximum Qty Allowed in Shopping Cart_ to 10; in _Custom Options_ tab, add different input types: text, area, drop-down, checkbox, etc. except file type.
2. [frontend] On the product page, fill all the fields, set _Qty_ to 11. Click the _Add to Cart_ button. 
3. Error "The maximum quantity allowed for purchase is 10." appears. All input fields are empty.
4. Apply this PR, repeat step 2. The product form is repopulated.

### DIFF
--- a/app/code/core/Mage/Checkout/controllers/CartController.php
+++ b/app/code/core/Mage/Checkout/controllers/CartController.php
@@ -257,11 +257,13 @@ class Mage_Checkout_CartController extends Mage_Core_Controller_Front_Action
 
             $url = $this->_getSession()->getRedirectUrl(true);
             if ($url) {
+                $this->_setProductBuyRequest();
                 $this->getResponse()->setRedirect($url);
             } else {
                 $this->_redirectReferer(Mage::helper('checkout/cart')->getCartUrl());
             }
         } catch (Exception $e) {
+            $this->_setProductBuyRequest();
             $this->_getSession()->addException($e, $this->__('Cannot add the item to shopping cart.'));
             $this->_goBack();
         }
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Checkout/controllers/CartController.php
```diff
@@ -711,4 +714,15 @@ protected function _getCustomerSession()
     {
         return Mage::getSingleton('customer/session');
     }
+
+    /**
+     * Set product form data in checkout session for populating the product form
+     * in case of errors in add to cart process.
```
fballiano, line 720:
```suggestion
     * in case of errors in add to cart process.
     * @return void
```

dunno if it's a good thing to add the return in the doc

app/code/core/Mage/Checkout/controllers/CartController.php
```diff
@@ -711,4 +714,15 @@ protected function _getCustomerSession()
...
+     */
+    protected function _setProductBuyRequest()
```
fballiano, line 722:
```suggestion
    protected function _setProductBuyRequest(): void
```

or this? not sure if it's the right syntax

ADDISON74:
It is working as expected, including "Input Type" = File. See bellow

screenshot_1

After you agree on the comments made, I will approve it.

**[OFF TOPIC]**

Testing this PR I identified an issue. If [Input Type] is one of **Date / Date & Time / Time**, by pressing the [Add to cart] button an error message appears. Look how the <em></em> tag is incorrectly in the message. If you take a look in the source code, here is how it is inserted

```<span>Please specify date required option &lt;em&gt;color&lt;/em&gt;.</span>```

screenshot_2

kiatng:
Fixed issue reported by @ADDISON74 
image

ADDISON74 (approved):

fballiano (approved):

PR #3377, opened 2023-07-12
Rewrote part of GA4 to support add/remove quantity from cart and cart_updates

fballiano:
The previous implementation:
- didn't support the add_to_cart quantity parameter, this was a problem when there are products with minimum_quantity (meaning you will have to add X amount to the cart, not only 1, while ga4 was tracking only 1).
- was forcing a `Mage::getModel('catalog/product')->load()` for every product, which is now removed since it was unneeded
- wasn't working with cart_updates (when you change the quantity of a product in the cart page), now this PR calculates the difference and triggers add_to_cart or remove_to_cart events

This version should be much more complete than before, hope you like it.

PR #3378, opened 2023-07-12
Removed double save on quote in Mage_Checkout_CartController::ajaxUpdateAction()

fballiano:
While developing https://github.com/OpenMage/magento-lts/pull/3377 we found out that the popup ajax cart:
Screenshot 2023-07-12 alle 15 22 20

is actually saving the quote two times with this code:
Screenshot 2023-07-12 alle 15 22 46

the first save is not needed, so this PR removes it, making it look like this:
Screenshot 2023-07-12 alle 15 23 26

You can check that the update product quantity, in the ajax cart, still works the same.
I've also checked the data on the database and it's consistent.


ADDISON74 (approved):

kiatng (approved):
Tested.

PR #3301, opened 2023-06-02
Rewrote afterSave of product and category image attributes

luigifab:
### Description

For catalog and category image attributes, this PR allow to:
- override extensions list
- same process for product and category
- rewrite `Mage_Core_Model_File_Uploader` like `core/file_validator_image`
- don't try to upload an image when `$_FILES` is empty (see #3086)

Tested with PHP 8.0 and 8.2.

### Contribution checklist
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3384, opened 2023-07-16
Allowed to upload images in WebP format

empiricompany:
### Description (*)
Since the webp format is now supported by all modern browsers *, I have added the ability to upload images in webp format.
* https://caniuse.com/webp

This PR allow to upload an image in webp format and add support to resize and cache it correctly in frontend.

PS: I know that there are extensions/techniques to dynamically generate the WebP version on the frontend, but based on the target audience, with this PR, we can decide to upload them directly in the correct format.

Test all scenarios with an image in webp format

Tested scenarios:
- [x] Product images
- [x] Product Custom Options Input Type: File
immagine
- [x] Category Image and Thumbnail Image
- [x] CMS WYSIWYG Media Storage

Untested scenarios:
- [ ] Import/export
- [ ] API

I create this pull request as a draft since there are areas that I have not yet tested.
Tested on PHP 7.4.29 with GD 2.1.0 (support to webp should be enabled by default)


empiricompany:
@fballiano can you rebase this PR since this commit?
[first draft webp support catalog media product and cms](https://github.com/OpenMage/magento-lts/pull/3384/commits/c6ceb853e42df038766625cbe36d7c054d591fa5)

Sorry if I always make a huge mess. 😅 but what matters is the substance

app/code/core/Mage/Eav/Model/Attribute/Data/Image.php
```diff
@@ -47,6 +47,7 @@ protected function _validateByRules($value)
             1   => 'gif',
             2   => 'jpg',
             3   => 'png',
+            18  => 'webp'
```
fballiano, line 50: just out of curiosity, why is that 18?

fballiano:
the PR looks ok, I don't think there's anything to rebase :-)

fballiano:
but actually I'd like to see https://github.com/OpenMage/magento-lts/pull/3301 merged before this one, since they work on the same stuff

app/code/core/Mage/Eav/Model/Attribute/Data/Image.php
empiricompany: it's IMAGETYPE_WEBP constant

fballiano: ahhh ok ok thanks!

fballiano:
:-) actually we could merge #3301 so this PR now needs a bit of reworking, but it's a good thing

fballiano:
I think I fixed the conflicts, hope I didn't screw anything in the process

fballiano:
IMHO it would be great to have a single 

```
    /**
     * @var string[]
     */
    public const ALLOWED_IMAGES_EXTENSIONS = ['webp', 'jpg', 'jpeg', 'png', 'gif', 'bmp'];
```

instead of many arrays throughout the code, all of the code should be referencing to that constant :-)

empiricompany:
> IMHO it would be great to have a single
> 
> ```
>     /**
>      * @var string[]
>      */
>     public const ALLOWED_IMAGES_EXTENSIONS = ['webp', 'jpg', 'jpeg', 'png', 'gif', 'bmp'];
> ```
> 
> instead of many arrays throughout the code, all of the code should be referencing to that constant :-)

yes but it was keep untouched for backward compatibility reasons
maybe we can create new constant and return it keeping in the deprecated methods?

fballiano:
the one that I posted is a new one ;-)

empiricompany:
> Personally, I would not put webp as the first image format in the list that is seen in the Backend. I would list it at the end, because it is a format after all the ones that have been so far.

sorry i can't figure out where is showed

fballiano:
he says only in the array list

empiricompany:
> Personally, I would not put webp as the first image format in the list that is seen in the Backend. I would list it at the end, because it is a format after all the ones that have been so far.

done

fballiano:
I've unified all into a single constant, see if you like it :-)

ADDISON74:
So far so good. 

I would like to test "**Image type product options**" mentioned in the description, but I don't know anything about this feature. Any help?

fballiano:
not sure but probably you have to create a new product attribute of type image:

Screenshot 2023-07-18 alle 21 26 14

ADDISON74:
That type of attribute adds a new column in the Images section. I asked because he made me curious about what it could be and I don't already know.

screenshot

fballiano:
ahhhh ok then I don't know 😅

empiricompany:
@ADDISON74 you have to create a product custom options of type image that allow customers to upload an image in the frontend

ADDISON74:
Now I understand what it is about. It should have been mentioned as "File type in custom options" . A word led to misunderstanding.

ADDISON74 (approved):
I tested this PR and encountered no issues. It is a new feature in OpenMage much discussed in the forums of the time.

empiricompany:
> Now I understand what it is about. It should have been mentioned as "File type in custom options" . A word led to misunderstanding.

yes sorry i've updated PR description

fballiano:
great, let's merge it, it's a nice little new addition for the upcoming release :-)

ADDISON74:
**[Only if it is confirmed]**

I would add a warning in the README file that issues may arise with Magmi. Here is a report https://github.com/dweeves/magmi-git/issues/596.

empiricompany:
> I would add a warning in the README file that issues may arise with Magmi. Here is a report [dweeves/magmi-git#596](https://github.com/dweeves/magmi-git/issues/596).

Maybe now it works, we need to try

ADDISON74:
First we have to do a test for the Import feature we already have in OpenMage and confirm that it works (CSV file, a few images uploaded, ...). I will check these days.

Issue #3394, opened 2023-07-20
Admin theme issue (overlapping menu items)


seifer7:
### Preconditions (*)

1. OpenMage version 20.0.20

### Steps to reproduce (*)

1. Have enough top-level menu items to wrapped menu to more than 1 line
2. Open a page from a menu on the bottom row of menu items
3. Notice that the selected menu is now highlighted
4. Hover over a menu item ABOVE the highlighted menu

### Expected result (*)

1. On step 4 the hover menu displayed should appear ON TOP of the other menu items
![image](https://github.com/OpenMage/magento-lts/assets/13601073/dc9a04be-dc8c-4441-aaff-a207cbe8b991)

### Actual result (*)

1. On step 4 the hover menu is displayed BELOW the highlighted menu item (current page)
![image](https://github.com/OpenMage/magento-lts/assets/13601073/922f6651-c359-4930-9439-e0830c6a0337)

seifer7:
Fix should be as simple as removing the `z-index` from `#nav li.active`
https://github.com/OpenMage/magento-lts/blob/8572935b999f4212e6e4ebd7094be1dfc1df13e8/skin/adminhtml/default/openmage/scss/menu.scss#L63-L67

As far as I can tell, that `z-index` was only added to fix `li` separators from appearing on-top of the active menu item in the legacy admin skin. The separators are not used in the new theme.

fballiano:
@seifer7 thank you for this report, do you want to open a PR or you want me to do it?

elidrissidev:
This issue is also happening with the default admin theme, so would be better to fix in both.

seifer7:
> This issue is also happening with the default admin theme, so would be better to fix in both.

Oh, thank you. I will have a look at it soon.

ADDISON74:
Please take in consideration changing the SCSS files too.

sreichel:
@all-contributors add @seifer7 bug

PR #3396, opened 2023-07-20
Admin theme issue (overlapping menu items)

seifer7:
### Description (*)

Small CSS update for the adminhtml. Removed the `z-index` from the `active` class used on the top menu bar to prevent the active item from rendering on top of the hover menu in some instances.

1. Fixes OpenMage/magento-lts#3394

1. Have enough top-level menu items to wrapped menu to more than 1 line
2. Open a page from a menu on the bottom row of menu items
3. Notice that the selected menu is now highlighted
4. Hover over a menu item ABOVE the highlighted menu
5. The pop-up/hover menu appears on top of the highlighted active menu

### Questions or comments

This is my first time working with `sass`.
I don't understand why it has changed various colours from hex to hsl representation, and added an additional decimal point to rgb values also.
How do I correct this? Do we care about this?


fballiano:
thanks @seifer7!

I think the changes in the css generation are due to the tool that's being used, if I understood correctly the original css were generated with something called "compass" which is mega old.

to avoid problems I'd modify the sass and css by hand, avoiding unrelated changes

fballiano:
I reworked this PR manually changing the files instead of regenerating the CSS from the SASS.

Tested and working.

There's no fix needed for the "legacy" backend theme, since it doesn't support wrapping for the main menu ;-)

fballiano (dismissed):

fballiano:
but if I shrink the screen I should have the same result no? To me it doesn't happen:

Screenshot 2023-07-24 alle 16 42 26

fballiano:
and that same element doesn't have a zIndex:

```
#nav li.active {
    margin-left: -1px;
    background: url(images/nav1_active.gif) no-repeat;
    color: #fff;
    font-weight: bold;
}
```

elidrissidev:
You need to have many menu items to make it wrap. Responsive menu is only part of the openmage theme.

Screen Shot 2023-07-24 at 4 47 33 PM

fballiano:
since the nav.active doesn't have a zIndex there's proabbly need of another kind of fix

fballiano:
actually I was wrong and there was a zIndex also on the legacy theme, I've removed and it solves the problem. I doesn't seem to me to generate other new problems.

fballiano (approved):

fballiano:
shall we try to test/merge this before the next RC?

elidrissidev:
I'll test it in a few hours and report back.

elidrissidev (approved):
Tested on the legacy theme

PR #3397, opened 2023-07-20
Fixed a bug with UPS that prevented saving the shipping methods page from backend

fballiano:
https://github.com/OpenMage/magento-lts/pull/3287 created a bug that prohibited the user from saving theshipping method backend page, because now these fields are required also when the shipping method is disabled:

Screenshot 2023-07-20 alle 17 17 05

This PR checks that the method is enabled before adding the required-entry class to those fields.


ADDISON74 (approved):

fballiano:
@empiricompany @roby255 would you mind give this a check?

roby255 (approved):

empiricompany (approved):

Issue #3407, opened 2023-07-25
Shipping Method cannot calculate shippingcosts based on "tax_subtotal" of the quote


EliasKotlyar:
### Preconditions (*)
1. OpenMage 19.4.*

### Steps to reproduce (*)
1. Create some custom module
2. Custom Module should contain a Shipping Method
3. Shipping Method should calculate Shipping based on "tax_subtotal" of the quote

### Expected result (*)
It should behave like in Magento 1.9.4.x. Tax-Subtotal should be usable in the shipping calculation

### Actual result (*)
You cannot use "tax-subtotal" in shipping method, as its not initialized yet with a value.

### Dev-Info

In the Magento 1.9.4.4 there are following code in app/code/core/Mage/Tax/etc/config.xml, which is responsible for the sort order of total calculation:
```
                    <tax_subtotal>
                        <class>tax/sales_total_quote_subtotal</class>
                        <after>freeshipping</after>
                        <before>tax,discount</before>
                    </tax_subtotal>
```
(see https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Tax/etc/config.xml#L165)

In Openmage, the code was changed to:
```
                    <tax_subtotal>
                        <class>tax/sales_total_quote_subtotal</class>
                       <after>subtotal,nominal,shipping,freeshipping</after>
                        <before>tax,discount</before>
                    </tax_subtotal>
```
Its unclear why it was done, but it seems to be some "upstream" upload from 1.9.3.x:
https://github.com/OpenMage/magento-lts/commit/d92a24aa27360439a6f6ed68ba9260685b2f9e54

ADDISON74:
If I remember well we discussed about this issue before in PR https://github.com/OpenMage/magento-lts/pull/1104

EliasKotlyar:
I read the PR #1104 but for me its unclear what exactly is the outcome. Will it be left as it is, or is someone willing to merge the any of the PRs which are resolving this issue?

fballiano:
This is the commit that introduced the change: https://github.com/OpenMage/magento-lts/commit/d92a24aa27360439a6f6ed68ba9260685b2f9e54 from 2018

ADDISON74:
Please read my comment that I made here https://github.com/OpenMage/magento-lts/pull/1104#issuecomment-915588882.

[...]

PR #3408, opened 2023-07-25
Shipping Method cannot calculate shipping costs based on "tax_subtotal" of the quote

EliasKotlyar:
1. Fixes OpenMage/magento-lts#3407


fballiano:
@ADDISON74 genuine question, I fail to understand what "Shipping Method cannot calculate shipping costs based on tax_subtotal" means, like, where is the problem happening? I don't know that part of openmage and I can't understand how important this problem really is

fballiano:
also, the commit that made this change originally seems to be https://github.com/OpenMage/magento-lts/commit/d92a24aa27360439a6f6ed68ba9260685b2f9e54, which, in my opinion, is a mistake because the [latest version of the tax/config.xml is clearly different](https://github.com/OpenMage/magento-mirror/blob/magento-1.9/app/code/core/Mage/Tax/etc/config.xml).

this is really weird but it scares me a bit to approve this change

ADDISON74:
It's been a while since we discussed this topic. I encountered problems with the setting in Magento vanilla, then with the one in OpenMage. I didn't have time to check in detail what was happening, but the problem was immediately solved by the modification shown above. Please check https://github.com/OpenMage/magento-lts/issues/3407 for more information and links to other discussions have been provided. I think there are not only those 4 options, but more. It would be ideal to be able to configure them in the Backend as a multiselect list.

Issue #3043, opened 2023-02-24
Deprecate decorate*() JS functions


loekvangool:
### Description (*)
JavaScript functions like `decorateTable()` are [popular ](https://github.com/OpenMage/magento-lts/search?q=decoratetable) in the templates. Their purpose is to add CSS classes `odd`, `even`, `first`, `last` to HTML tables, lists, etc.

This seems wildly outdated and required browser repainting, leading to lower Web Vitals scores. CSS3 can easily target these entities without writing classes with JavaScript after initial page load.

Note: Even when there are no actual visual changes, Chrome still reports a repaint and deducts points. These functions should not be used.

fballiano:
I'd love to drop them :-)

justinbeaty:
Previous discussion about `decorate*()` functions was here: https://github.com/OpenMage/magento-lts/pull/1073#issuecomment-663162784

Maybe instead of removing, we can add a config setting to enable / disable them? Something like a "legacy theme compatibility" switch? That way, you can disable to improve performance if needed.

loekvangool:
Thanks for referencing #1073 I hadn't found that one yet.

This project requires PHP 7.3 which was released in december 2018. Meanwhile, the CSS discussed in this issue has been supported in Chrome since 2010 (https://caniuse.com/css-sel2). I just wanted to point that out, it seems like a huge difference between how frontend and backend code is viewed by some.

To be brutally honest I don't have these functions on my own template anymore, so I'm good if it stays.

justinbeaty:
@loekvangool I think the hesitation about removing it from the theme was that perhaps there is JS out there that finds nodes based on these classes and manipulates the DOM expecting these classes.

If nothing else, I'd be happy to make a PR to implement the config in the admin, but I don't even use the default theme either so I wouldn't be able to give it a good testing.

tmotyl:
Just kill it with fire :)
V20 doesnt have to be backward compatible

loekvangool:
>@loekvangool I think the hesitation about removing it from the theme was that perhaps there is JS out there that finds nodes based on these classes and manipulates the DOM expecting these classes.
Yes that's clear, however we require PHP 7.3 which means all users need to go through their code to fix stuff that does not work anymore. It seems fair to declare a decent interval in which OM expects users to update their code? If you mean code that's _part of OM_ that needs `decorate*()`, I agree it should be part of the removal.

[...]

PR #3410, opened 2023-07-26
Removed decorate*() js functions

fballiano:
M1 had 4 decorate*() javascript functions, that are used (mostly) to add even/odd classes to HTML tables and things like that. The removal of these functions was discussed a couple of times already:
- https://github.com/OpenMage/magento-lts/pull/1073#issuecomment-663162784
- https://github.com/OpenMage/magento-lts/issues/3043
but I think that in 2023 we should work on this.

This PR:
- removes all the calls to decorate*() funcions from the phtml templates
- deprecates the 4 javascript functions in js/varien/js.js and empties the function "body"

Since the functions are not removed by this PR it's technically not a breaking change but... it changes the final rendering of the pages so... I'd still merge this in the `next` branch.

This could also be considered a small step forwards removing prototypejs.

Fixes https://github.com/OpenMage/magento-lts/issues/3043


ADDISON74:
I agree. Do you have in plan a replacement to keep the same behavior?

fballiano:
not at the moment (didn't have time yet) but I think it could be added :-)

ADDISON74:
> This could also be considered a small step forwards removing prototypejs.

At some point we may be able to do this, but I'm worried about the extensions that use this library. I have so far five in production, either I should make a conversion to jQuery (the fastest way) or to JavaScript (a little more complicated and time-consuming).

If I don't adopt any of the solutions, then I can keep the library and add it manually for functionality. What if we removed everything that is old in OpenMage, including this library and it would be added by those who need it with Composer?

fballiano:
the functions are still there, so there will be no break, but yes you'll lose the functionality. I think that who really needs these functions can just copy them and put them in their javascripts ;-) although for sure somebody has created a library for that.

loekvangool:
> At some point we may be able to do this, but I'm worried about the extensions that use this library. I have so far five in production, either I should make a conversion to jQuery (the fastest way) or to JavaScript (a little more complicated and time-consuming).

Can this be done with CSS rules instead of JS? Better idea for Web Vitals scores.

empiricompany (approved):

sreichel:
> but yes you'll lose the functionality

Should even/odd/first/last be added to CSS-files to keep current output?

loekvangool:
> Should even/odd/first/last be added to CSS-files to keep current output?

I think they were talking about removing Prototype.js as a whole, which would lead to issues with 3rd party extension beyond CSS classes.

fballiano:
> Should even/odd/first/last be added to CSS-files to keep current output?

RWD doesn't have even/odd stripes already, as shown in the screenshots.

loekvangool:
> RWD doesn't have even/odd stripes already, as shown in the screenshots.

There is an impact, but it's mostly [marginal](https://github.com/search?q=repo%3AOpenMage%2Fmagento-lts+%28even+OR+odd%29+language%3ACSS&type=code).

fballiano:
@loekvangool the only one is in skin/adminhtml/default/default/boxes.css (the other ones as alread CSS based even/odd ;-)) I can for sure take a look at skin/adminhtml/default/default/boxes.css no problem

loekvangool:
There are a couple more hits in boxes.css, but yeah it's barely anything.

There's also a CSS class `zebra-table` that does it the CSS way, but it looks unused.

sreichel:
> > Should even/odd/first/last be added to CSS-files to keep current output?
> 
> RWD doesn't have even/odd stripes already, as shown in the screenshots.

Your're right for RWD, but PR removes it from "base" template.

I know other themes have been removed, but i know some stores that rely on modern theme ... is this a sure shot?

fballiano:
> > RWD doesn't have even/odd stripes already, as shown in the screenshots.
> 
> There is an impact, but it's mostly [marginal](https://github.com/search?q=repo%3AOpenMage%2Fmagento-lts+%28even+OR+odd%29+language%3ACSS&type=code).

@loekvangool seems grids work fine anyway with this PR:
Screenshot 2023-08-01 alle 08 39 27

loekvangool:
Indeed

sreichel:
"nth-child" ... this is why i asked for. I thought even/odd/... has to be changed.

ADDISON74:
I have not tested this PR and I will not approve it without being convinced that the change is mandatory. Those functions apply to many classes/IDs of some elements in the page. For example class/ID "product_comparison", I need to see the effect in the browser before and after the change. All these elements may need to be added to the CSS/SCSS files to maintain the decoration effect.

fballiano:
I don't think there's anything that could be added in the CSS, `base` doesn't have css, `adminhtml` and  `rwd` don't use that functionality.
also, this PR targets `next`, where I think it's correct to remove 100 javascript calls that were needed years ago but it's much more performant to do the same thing via css instead of js.

kyrena:
Can I suggest a `console.warn('decorateXyz is deprecated');` to force developers to update code?

js/varien/js.js
```diff
@@ -33,122 +33,33 @@ function setPLocation(url, setFocus){
  *
  * @param elements - array of elements to be decorated
  * [@param decorateParams] - array of classes to be set. If omitted, all available will be used
+ * @deprecated
```
kiatng, line 36: Is it better to not remove the code first, only mark it deprecated?

fballiano: I would deprecate it in main and remove it in next, otherwise it will take forever and we already have many breaking changes in next.
I could open a PR with deprecating this funcions in main but I wouldn't deprecate them in next, those old things from the past have to go as fast as possible IMHO

fballiano:
After https://github.com/OpenMage/magento-lts/pull/3526 got merged maybe we can merge this one in the `next` branch ;-)

kiatng (approved):

luigifab (approved):

PR #3411, opened 2023-07-26
Updated copyright docblocks

fballiano:
This PR is created running the `shell/update-copyright.php` script for the `main` branch.

I think it's important that we update the docblock copyright data before the next stable release.

I'll create a separate PR for v19.


kiatng (approved):

pquerner:
Is this something that really needs to happen nowadays?

Nice read: https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

ADDISON74:
Opinions are divided. In addition, I always put the legal side first, which protects me from any problems that might arise later. In OpenMage we created a script that updates all the files that have been modified and it is not a big effort to achieve such PRs.

I would read this opinion too. **Why keep the year?** https://matija.suklje.name/how-and-why-to-properly-write-copyright-statements-in-your-code.

PR #3418, opened 2023-07-29
Incorrect stock vs sales in high load and high concurrency

fballiano:
This PR takes the discussion in https://github.com/OpenMage/magento-lts/issues/364 in order not to loose the code.

Untested by me, needs testing but I think it's important to keep it,

Fixes https://github.com/OpenMage/magento-lts/issues/364


sreichel:
For some reasons codesniffer checks did not run ... so no complains about old array syntax?

fballiano:
it's because the branch is mega old.
can't push to that repo.
i'll have to recreate the pr

PR #3419, opened 2023-07-29
Incorrect stock vs sales in high load and high concurrency

### LINKED ISSUE #364: Incorrect stock versus sales

fballiano:
This PR
- takes the discussion in #364 in order not to loose the code.
- needs a better name and description
- is not tested by me, needs testing but it's here cause I thought it was valuable

Fixes #364


app/code/core/Mage/CatalogInventory/Model/Stock/Item.php
```diff
@@ -783,6 +783,40 @@ protected function _beforeSave()
...
+        try {
+            $this->_beforeSave();
+            $this->getResource()->updateRecord($this->getId(), [
+                'is_in_stock' => $this->getIsInStock(),
+                'stock_status_changed_auto' => $this->getStockStatusChangedAutomatically(),
+                'low_stock_date' => $this->getLowStockDate()
+            ]);
+            $this->_afterSave();
+            $this->_getResource()
+                ->addCommitCallback(array($this, 'afterCommitCallback'))
```
sreichel, line 809:
```suggestion
                ->addCommitCallback([$this, 'afterCommitCallback'])
```

app/code/core/Mage/CatalogInventory/Model/Resource/Stock/Item.php
```diff
@@ -109,4 +109,19 @@ protected function _prepareDataForTable(Varien_Object $object, $table)
...
+
+    /**
+     * Update specific fields of a single row.
+     *
+     * @param int $itemId Item id of the row being updated
+     * @param mixed[] $data Array structure of fields being updated
+     *
+     * @return $this
+     */
+    public function updateRecord($itemId, $data)
```
sreichel, line 121:
```suggestion
    public function updateRecord(int $itemId, array $data)
```

Untested.

PR #3425, opened 2023-08-03
Allowed admin grid row to open in new tab

justinbeaty:
### Description (*)
This always bugged me. On any admin grid, you can click the row to open whatever URL was set with the `getRowUrl` method in the Block class. But the link opens via JS `window.location.href`, so you cannot open in a new tab.

This PR improves the handling of row clicks, see manual testing scenarios.

In both a mass action grid (i.e. Sales > Orders) and a regular grid (i.e. Promotions > Catalog Price Rules), try on a row:

1. left click -> should open in existing tab (same as before)
2. ctrl+click (Windows) -> should open new tab
3. command+click (macOS) -> should open new tab
4. middle click -> should open new tab
5. right click -> nothing

### DIFF
--- a/js/mage/adminhtml/grid.js
+++ b/js/mage/adminhtml/grid.js
@@ -439,7 +439,11 @@ varienGridMassaction.prototype = {
                 return;
             }
             if (trElement.title) {
-                setLocation(trElement.title);
+                if (evt.ctrlKey) {
+                    window.open(trElement.title, '_blank');
+                } else {
+                    setLocation(trElement.title);
+                }
             }
             else{
                 var checkbox = Element.select(trElement, 'input');
--- a/js/mage/adminhtml/grid.js
+++ b/js/mage/adminhtml/grid.js
@@ -311,13 +311,19 @@ varienGrid.prototype = {
     }
 };
 
+function shouldOpenGridRowNewTab(evt){
+    return evt.ctrlKey // Windows ctrl + click
+        || evt.metaKey // macOS command + click
+        || evt.button == 1 // Middle mouse click
+}
+
 function openGridRow(grid, evt){
     var trElement = Event.findElement(evt, 'tr');
     if(['a', 'input', 'select', 'option'].indexOf(Event.element(evt).tagName.toLowerCase())!=-1) {
         return;
     }
     if(trElement.title){
-        if (evt.ctrlKey || evt.button == 1) {
+        if (shouldOpenGridRowNewTab(evt)) {
             window.open(trElement.title, '_blank');
         } else {
             setLocation(trElement.title);
@@ -440,6 +446,7 @@ varienGridMassaction.prototype = {
         this.getOldCallback('init_row')(grid, row);
     },
     onGridRowClick: function(grid, evt) {
+
         var tdElement = Event.findElement(evt, 'td');
         var trElement = Event.findElement(evt, 'tr');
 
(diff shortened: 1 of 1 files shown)


pquerner:
While we're at it, the search results from the global search are also only opened in the current tab.
Luckily you can patch the js from the outside without much fuss since its a prototype class thing, but this could also be added while this PR is open, right? :D

justinbeaty:
@pquerner I have that global search disabled since forever, but I looked into it. Maybe something [like this](https://github.com/justinbeaty/magento-lts/commit/f73f68280c4e8ff5e5c430ac0942d7f636259404), but it's not perfect.

I think it can go into a separate PR though if you want to take lead on it.

pquerner:
For reference, this was the piece of js I implemented:

```
const originalAutocompleterUpdateElementFunc = Autocompleter.Base.prototype.updateElement

// For Global Search, open in new tab
Autocompleter.Base.prototype.updateElement = function(selectedElement) {
    if (this.options.updateElement && this.options.indicator.includes('global_search_indicator')) {
        window.open(selectedElement.getAttribute('url'));
        return;
    }
    originalAutocompleterUpdateElementFunc(selectedElement);
    return;
}

you could get away with it in the phtml file, but I didnt want to overwrite the phtml file for that.
```

pquerner:
Oh, right. The `event` variable wasnt accessibly (in a good way) there. ;D now I remember

justinbeaty:
@pquerner yeah, that was the tricky part a bit. It's a bit weird though with the state of the autocompleter after the new window is opened, since it was assumed we would be navigating away.

justinbeaty:
I made a few more enhancements. Now:

* Left click on row -> default / old behavior
* Middle click on row -> new tab
* Ctrl + click on row -> new tab
* Right click on row -> nothing

I forgot that middle click doing nothing and right click opening the page were also annoying things.

kiatng (approved):
Tested. I like it.

js/mage/adminhtml/grid.js
```diff
@@ -56,6 +56,7 @@ varienGrid.prototype = {
 
                 Event.observe(this.rows[row],'mouseover',this.trOnMouseOver);
                 Event.observe(this.rows[row],'mouseout',this.trOnMouseOut);
+                Event.observe(this.rows[row],'mousedown',this.trOnClick);
```
fballiano, line 59: is Event a prototypejs thingy?

kiatng: Yes. See http://api.prototypejs.org/dom/Event/observe/index.html

justinbeaty: FYI, this is needed because middle click doesn’t fire the click event.

elidrissidev:
This PR solves a big pain, but unfortunately it does not work on macOS because they decided that CTRL + click should toggle the context menu. It should be possible to support that by checking for `event.metaKey`.

justinbeaty:
Added in support for command click, but haven't tested on macOS.

js/mage/adminhtml/grid.js
fballiano: mmm I'd much rather remove some prototype code instead of adding it :-\

fballiano:
tested: command click works on MacOS

js/mage/adminhtml/grid.js
justinbeaty: I would like to remove prototype too, but at some point it will be better to completely overhaul the JS code used in admin grids, because the whole file is based on `Class.create()`:

```js
var varienGrid = new Class.create();
varienGrid.prototype = {
  // ...
```

Converting a few lines in this file seems futile -- I like smaller wins where we can completely remove prototype from a certain phtml file, or completely rewrite a .js file.

fballiano (approved):

PR #3427, opened 2023-08-04
Better PHP errors logging for API

kyrena:
### Description

When an error occur with API, currently you can have something like this in system.log (from my memory): `argument must be of type arrayapp/code/core/Mage/Sales/Model/Order/Creditmemo/Api/V2.php`

With this PR, we have: `argument must be of type array in app/code/core/Mage/Sales/Model/Order/Creditmemo/Api/V2.php on line 86`

See #3346.

### Contribution checklist


app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
```diff
@@ -35,9 +35,9 @@ public function __construct()
      * @param string $errorFile
      * @return bool
      */
-    public function handlePhpError($errorCode, $errorMessage, $errorFile)
+    public function handlePhpError($errorCode, $errorMessage, $errorFile, $errLine)
     {
-        Mage::log($errorMessage . $errorFile);
+        Mage::log($errorMessage . '  in ' . $errorFile . ' on line ' . $errLine);
```
elidrissidev, line 40: I like it. We could also set the level to `Zend_Log::ERR` cause it appears that it's logging them as `DEBUG` messages.

kiatng: Also there is an extra space in `'  in '`.

fballiano, line 40:
```suggestion
        Mage::log($errorMessage . ' in ' . $errorFile . ' on line ' . $errLine, Zend_Log::ERR);
```

fballiano (approved):

ADDISON74 (approved):

app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
luigifab: The double spaces is not an error.
PHP add also two spaces here, see also here:

https://github.com/OpenMage/magento-lts/blob/0ac3f19a23eed02fbb27cc5586254a5eb7d30af9/app/code/core/Mage/Core/functions.php#L182

But it's not really important, and phpdoc was not updated 😸.

PR #3433, opened 2023-08-06
Fixes system config dependencies multi value

empiricompany:
### Description
I'm not sure if this is an old bug or if I am missing something in order to achieve the same result, but it seems that the ability to specify multiple values for configurations has already been implemented. However, it appears to be buggy.

In this configuration example the "wysiwyg/skin" configuration should only be displayed if the "wysiwyg/enabled" field has one of the following values: enabled, hidden. To achieve this, a "separator" field has been added to allow for the separation of multiple values.

```
<wysiwyg translate="label">
    <label>WYSIWYG Options</label>
    <sort_order>100</sort_order>
    <fields>
        <enabled translate="label">
            <label>Enable WYSIWYG Editor</label>
            <source_model>adminhtml/system_config_source_cms_wysiwyg_enabled</source_model>
            <sort_order>1</sort_order>
        </enabled>
        <skin translate="label">
            <label>Skin WYSIWYG Editor</label>
            <source_model>adminhtml/system_config_source_cms_wysiwyg_skin</source_model>
            <sort_order>1</sort_order>
            <depends>
                <enabled>
                    <value>enabled|hidden</value>
                    <separator>|</separator>
                </enabled>
            </depends>
        </skin>
    </fields>
</wysiwyg>
```

I have searched through the PRs, but I haven't found a solution to achieve the same result.

Upon investigating the code in Form.php, it appears that this issue has remained unresolved for the past 12 years. This suggests that it may be an old bug that still needs to be addressed. Additionally, it should be noted that the variable "$dependent" cannot be an array, but rather an object.

blame

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php
@@ -335,8 +335,8 @@ class Mage_Adminhtml_Block_System_Config_Form extends Mage_Adminhtml_Block_Widge
                             . $dependentFieldNameValue;
                         $shouldBeAddedDependence = true;
                         $dependentValue = (string)(isset($dependent->value) ? $dependent->value : $dependent);
-                        if (isset($dependent['separator'])) {
-                            $dependentValue = explode((string)$dependent['separator'], $dependentValue);
+                        if (isset($dependent->separator)) {
+                            $dependentValue = explode((string)$dependent->separator, $dependentValue);
                         }
                         $dependentFieldName = $fieldPrefix . $dependent->getName();
                         $dependentField     = $dependentFieldGroup->fields->$dependentFieldName;


S0FTWEX:
This PR seems to be code breaking. My modules are using this standard code notation in system.xml:

`
...
  <depends>
      <enabled separator="|">
          <value>enabled|hidden</value>
      </enabled>
  </depends>
...
`

Maybe, it would be great to support both notations.

S0FTWEX (changes requested):
It seems to be code breaking. My modules are using this standard code notation in system.xml:

```
...
  <depends>
      <enabled separator="|">
          <value>enabled|hidden</value>
      </enabled>
  </depends>
...
```

So, the original code is correct. Maybe, it would be great to support both notations ($dependent['separator']  and  $dependent->separator).

empiricompany:
> It seems to be code breaking. My modules are using this standard code notation in system.xml:
> 
> ```
> ...
>   <depends>
>       <enabled separator="|">
>           <value>enabled|hidden</value>
>       </enabled>
>   </depends>
> ...
> ```
> 
> So, the original code is correct. Maybe, it would be great to support both notations ($dependent['separator']  and  $dependent->separator).

oh ok! so it's this the correct sintax it's OK for me I update my code no problem

Issue #860, opened 2019-11-26
[Feature] Add North-Macedonia to the country list / update country list


uni-t:
As North-Macedonia has finally declared itself independent it's time to add it to the country list of Magento.

We have also noticed that other countries or regions are also missing from the country list os it might be a tedious though usefull thing to go through the full country list and add/update countries when necessary.

PR #3436, opened 2023-08-09
Updated countries list

elidrissidev:
### Description (*)

I updated the countries list using the latest data from: https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2

### Related

- See https://github.com/OpenMage/magento-lts/issues/860

1. Apply the patch, clean the cache, and check that the data was updated.


app/code/core/Mage/Directory/data/directory_setup/data-upgrade-1.6.0.3-1.6.0.4.php
```diff
@@ -0,0 +1,35 @@
...
+$installer->getConnection()->delete(
+    $installer->getTable('directory/country'),
+    ['country_id = ?' => 'AN']
+);
+
+$data = [
+    ['BQ', 'BQ', 'BES'],
+    ['CW', 'CW', 'CUW'],
+    ['SS', 'SS', 'SSD'],
+    ['SX', 'SX', 'SXM'],
```
mmenozzi, line 31: We should also add North-Macedonia as said in [this discussion](https://github.com/OpenMage/magento-lts/issues/860). I think we should also add other countries added or removed after the first version of Magento 1 (not easy to do).

elidrissidev: North Macedonia has the country code MK which already exists.

I did a comparison between the countries in the database currently and the ones from the Wikipedia link and these are all the differences I found.

mmenozzi (approved):

app/code/core/Mage/Directory/data/directory_setup/data-upgrade-1.6.0.3-1.6.0.4.php
```diff
@@ -0,0 +1,35 @@
...
+/** @var Mage_Core_Model_Resource_Setup $installer */
+$installer = $this;
+
+// The Netherlands Antilles (AN) was divided into Bonaire, Saint Eustatius and Saba (BQ, BES, 535),
+// Curaçao (CW, CUW, 531) and Sint Maarten (Dutch part) (SX, SXM, 534).
+//
+// See: https://www.iso.org/obp/ui/#iso:code:3166:AN
+$installer->getConnection()->delete(
+    $installer->getTable('directory/country'),
+    ['country_id = ?' => 'AN']
```
fballiano, line 24: are we safe to delete this record? should we do something for the people that have some settings dependent on it?

Flyingmana: I would like to avoid deleting country codes on existing instances. There might happen a lot of surprises, or shops simply continue to use them 👀

maybe its better to just remove the adding, so it only affects new instances

fballiano: I know it's probably not super-mega-clean but I tend to agree with @Flyingmana

ADDISON74: The only concern is only for the information collected in the past, not for the future. I have not evaluated what happens if it is removed, it is no longer displayed in the orders, customer accounts? If there are no issues it can be removed. In addition, I don't know more about that country, one thing is to delete the USA and another country AN. Has anyone used this country in OpenMage?

elidrissidev: I'll update it.

kiatng:
How about `Zend_Locale` for country names?

elidrissidev:
> How about `Zend_Locale` for country names?

I have no knowledge about that. Does it create an issue if they're not in `Zend_Locale`?

empiricompany (approved):

ADDISON74 (approved):

elidrissidev:
The new country names need to be added to zf1-future as well in order for them to be displayed properly. So this has to go back to draft until that's done.

elidrissidev:
ouch there's soo many xml files to edit 😂

fballiano (approved):

Issue #3435, opened 2023-08-09
Google Analytics throws Fatal Error


daboss84:
### Preconditions (*)

1. OM 20.1.0
2. Multishop
3. Product in mulitple categories in multiple shops
4. PHP Version 8.1.16 (in docker)

### Steps to reproduce (*)

1. Add such a product to cart

### Expected result (*)

1. Product is added to cart without errors

### Actual result (*)

> Fatal error: Uncaught TypeError: Mage_GoogleAnalytics_Helper_Data::getLastCategoryName(): Return value must be of type string, null returned in /app/app/code/core/Mage/GoogleAnalytics/Helper/Data.php:186
> Stack trace: 
> #0 /app/app/code/core/Mage/GoogleAnalytics/Model/Observer.php(88): Mage_GoogleAnalytics_Helper_Data->getLastCategoryName()
> 
> #1 /app/app/code/core/Mage/Core/Model/App.php(1441): Mage_GoogleAnalytics_Model_Observer->processItemsAddedOrRemovedFromCart() 
> #2 /app/app/code/core/Mage/Core/Model/App.php(1419): Mage_Core_Model_App->_callObserverMethod() 
> #3 /app/app/Mage.php(509): Mage_Core_Model_App->dispatchEvent() 
> #4 /app/app/code/core/Mage/Core/Model/Abstract.php(509): Mage::dispatchEvent() 
> #5 /app/app/code/core/Mage/Sales/Model/Quote/Item.php(788): Mage_Core_Model_Abstract->_afterSave() 
> #6 /app/app/code/core/Mage/Core/Model/Abstract.php(380): Mage_Sales_Model_Quote_Item->_afterSave() 
> #7 /app/app/code/core/Mage/Sales/Model/Quote/Item.php(771): Mage_Core_Model_Abstract->save() 
> #8 /app/app/code/core/Mage/Core/Model/Resource/Db/Collection/Abstract.php(643): Mage_Sales_Model_Quote_Item->save() 
> #9 /app/app/code/core/Mage/Sales/Model/Quote.php(380): Mage_Core_Model_Resource_Db_Collection_Abstract->save() 
> #10 /app/app/code/core/Mage/Core/Model/Abstract.php(380): Mage_Sales_Model_Quote->_afterSave() 
> #11 /app/app/code/core/Mage/Sales/Model/Quote.php(2087): Mage_Core_Model_Abstract->save() 
[...]

PR #3438, opened 2023-08-10
Fixing invalid null-return in Google Analytics 

daboss84:
### Description / Fixed Issues (*)
See #3435 :
The problem is, that the _categorieIds contains ALL its categories, of all its subshops. The poped category is not from the subshop from where it is added to the cart. Probably that's why the category is not loaded and its name is null.


empiricompany:
Hello @daboss84, thanks for reporting these issues. What about a more granular selection like this? Also, fix a case where the category is not active but is being incorrectly tracked. (we have some customers that put products in hidden categories for internal purpose)
I've tested it in a multi-website/store environment.

```
    /**
     * Returns last category name
     *
     * @param Mage_Catalog_Model_Product $product
     * @return string|false
     */
    public function getLastCategoryName($product): string|false
    {
        // force root category to current store
        $rootCategory = Mage::getModel('catalog/category')
            ->load(Mage::app()->getStore()->getRootCategoryId());

        $_lastCat = Mage::getResourceModel('catalog/category_collection')
            ->addAttributeToSelect('name')
            ->addIdFilter($product->getCategoryIds())
            ->addIsActiveFilter()
            ->addFieldToFilter('path', array('like' => $rootCategory->getPath() . '/%'))
            ->addOrder('level')
            ->getFirstItem();

        return $_lastCat->getName() ?: false;
    }
```

Reference: https://stackoverflow.com/questions/15505221/how-do-i-get-the-category-ids-that-a-product-is-in-with-respect-to-the-store-tha



Test conducted:

websites
categories

correct skip Categories of "Secondary Website" and disabled categories "New arrivals 1.2"
```
2023-08-10T08:43:33+00:00 DEBUG (7): Array
(
    [0] => gtag('event', 'add_to_cart', {"currency":"USD","value":"99.00","items":[{"item_id":"test-multi-shop","item_name":"Test multi webstire store product","price":"99.00","quantity":1,"item_brand":"","item_category":"New Arrivals"}]});
    [1] => gtag('event', 'view_item', {"currency":"USD","value":"99.00","items":[{"item_id":"test-multi-shop","item_name":"Test multi webstire store product","list_name":"Product Detail Page","item_category":"New Arrivals","price":"99.00"}]});
)
```
without excluding no active categories add_to_cart tracks it wrong:
```
2023-08-10T08:43:33+00:00 DEBUG (7): Array
(
    [0] => gtag('event', 'add_to_cart', {"currency":"USD","value":"99.00","items":[{"item_id":"test-multi-shop","item_name":"Test multi webstire store product","price":"99.00","quantity":1,"item_brand":"","item_category":"New Arrivals 1.2"}]});
    [1] => gtag('event', 'view_item', {"currency":"USD","value":"99.00","items":[{"item_id":"test-multi-shop","item_name":"Test multi webstire store product","list_name":"Product Detail Page","item_category":"New Arrivals","price":"99.00"}]});
)
```

If I re-enable "New Arrivals 1.2," it correctly picks it.

fballiano:
I think we should use a collection instead of a "load in a loop", the performance hit (in this case) wouldn't be big but a collection just looks better and safer

daboss84:
Hi,
sorry for the delay, I am on a business trip right now.
So if I get you right, we're going with @empiricompany's solution? Then I'll go forward and merge his pull request? However, I've no time for testing it, too. But his results look good.

fballiano:
@daboss84 sorry for the delay, I'd merge @empiricompany's PR https://github.com/daboss84/magento-lts/pull/1 cause a collection it's always safer than loading a in a loop, I'll try to test it our as soon as it gets published here ;-)

app/code/core/Mage/GoogleAnalytics/Helper/Data.php
```diff
@@ -172,17 +172,23 @@ public function isUserIdEnabled($store = null)
...
+
+        $_lastCat = Mage::getResourceModel('catalog/category_collection')
+            ->addAttributeToSelect('name')
+            ->addIdFilter($product->getCategoryIds())
+            ->addIsActiveFilter()
+            ->addFieldToFilter('path', array('like' => $rootCategory->getPath() . '/%'))
+            ->addOrder('level')
+            ->getFirstItem();
+
+        return $_lastCat->getName() ?: false;
```
fballiano, line 191:
```suggestion
        return $_lastCat->getName() ?: '';
```

since the method is called getLastCategoryName I'd expect it to return a string, or a null in the worst case?

daboss84: I'd prefer an empty string. I guess in the end it would be used as "empty string" later anyways?

app/code/core/Mage/GoogleAnalytics/Helper/Data.php
```diff
@@ -172,17 +172,23 @@ public function isUserIdEnabled($store = null)
      * Returns last category name
      *
      * @param Mage_Catalog_Model_Product $product
-     * @return string
+     * @return ?string
      */
-    public function getLastCategoryName($product): string
+    public function getLastCategoryName($product): ?string
```
fballiano, line 177:
```suggestion
    public function getLastCategoryName($product): string
```

app/code/core/Mage/GoogleAnalytics/Helper/Data.php
empiricompany: I returned false because in other parts it is tracking false in the datalayer, but it's the same.

fballiano:
phpstan is not happy

daboss84:
What does that mean exaclty?

empiricompany:
@daboss84 merge last review suggests from @fballiano and should be fix also phpstan

this is the correct version:

```
     /**
     * Returns last category name
     *
     * @param Mage_Catalog_Model_Product $product
     * @return string
     */
    public function getLastCategoryName($product): string
    {
        // force root category to current store
        $rootCategory = Mage::getModel('catalog/category')
            ->load(Mage::app()->getStore()->getRootCategoryId());

        $_lastCat = Mage::getResourceModel('catalog/category_collection')
            ->addAttributeToSelect('name')
            ->addIdFilter($product->getCategoryIds())
            ->addIsActiveFilter()
            ->addFieldToFilter('path', array('like' => $rootCategory->getPath() . '/%'))
            ->addOrder('level')
            ->getFirstItem();

        return $_lastCat->getName() ?: '';
    }
```

daboss84:
Ah ok, thanks. Yes I can do that in the evening (in a couple of hours)

fballiano:
I've pushed a new version to this PR, let me know if it works for you

daboss84:
Hi @fballiano I just run a quick test and it looks to me, like the category is empty now.

grafik

empiricompany:
the problem is  this commit https://github.com/OpenMage/magento-lts/pull/3438/commits/ad9d88181cb844570ba362328062bb086cd5f4fc 

`$storeRootCategory = Mage::app()->getStore()->getRootCategoryId();`
then
`->addFieldToFilter('path', ['like' => "{$storeRootCategory}/%"])`

condition results: like '2/%' where 2 is the root category of store
but this not include the root category 1, the condition must be like '1/2/%'

catalog_category_entity

in my version i'm using getPath() that return 1/2 including root category id avoiding this issues
we can simply change the condition to 
`->addFieldToFilter('path', ['like' => "1/{$storeRootCategory}/%"])`
assuming that the root category as id 1 or there is same method to get the root correct category?
or simply restore the loading of root category model then getPath()`
```
$rootCategory = Mage::getModel('catalog/category')
            ->load(Mage::app()->getStore()->getRootCategoryId());
[..]
->addFieldToFilter('path', array('like' => $rootCategory->getPath() . '/%'))
```

daboss84:
I just committed those changes. Now it looks like it's working in my test environment.

daboss84:
Do I have anything to do with this? Or can I leave it as it is? If I have to merge, I can do it in a couple of hours (european night time)
grafik

fballiano:
I think the first time I've to approve the workflow run.

sorry about the rootCategory thing, I tested it in my environment and it was working fine and I really wanted to avoid another model load...

fballiano (approved):

kiatng (approved):

PR #2353, opened 2022-08-01
GUI to Edit EAV Attributes & Sets - Category

justinbeaty:
### Description (*)
[WIP] This commit was previously part of #2317. It is broken out so that we can discuss ways to integrate the category EAV attributes into root categories. Discussions about this are in that PR.

### Related Pull Requests
#2317
#2264 

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #3451, opened 2023-08-19
Adds product image column to product grid

empiricompany:
It adds a column with the product preview image, which I believe is the most important aspect in this area:
- catalog product list
- catalog category products tab
- catalog product view in tabs upsell, related and crossell

I have added a new column renderer for product images. The images are resized to 64px by default to reduce file size, and they are generated in media/catalog/product/cache/0/64x64/ directory. 
Image previews are linked to the original url.

I think this is an important feature to add to OpenMage.

This is an initial step towards implementing advanced backend grids. The idea is to develop customizable grids where users can choose which columns to display as they prefer that display current columns by default and all additionaly is disabled to reduce possibile backward compatibility.

catalog_product
catalog_category
catalog_product_tabs

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
@@ -83,6 +83,8 @@ class Mage_Adminhtml_Block_Catalog_Category_Tab_Product extends Mage_Adminhtml_B
                 'category_id=' . (int) $this->getRequest()->getParam('id', 0),
                 'left'
             );
+        $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
+
         $this->setCollection($collection);
 
         if ($this->getCategory()->getProductsReadonly()) {
@@ -118,6 +120,14 @@ class Mage_Adminhtml_Block_Catalog_Category_Tab_Product extends Mage_Adminhtml_B
             'width'     => '60',
             'index'     => 'entity_id'
         ]);
+        $this->addColumn(
+            'image',
+            [
+                'header' => Mage::helper('catalog')->__('Image'),
+                'type'  => 'image',
+                'index' => 'image',
+            ]
+        );
         $this->addColumn('name', [
             'header'    => Mage::helper('catalog')->__('Name'),
             'index'     => 'name'
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
@@ -124,7 +124,7 @@ class Mage_Adminhtml_Block_Catalog_Category_Tab_Product extends Mage_Adminhtml_B
             'image',
             [
                 'header' => Mage::helper('catalog')->__('Image'),
-                'type'  => 'image',
+                'type'  => 'productimage',
                 'index' => 'image',
             ]
         );
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab/Product.php
@@ -83,6 +83,7 @@ class Mage_Adminhtml_Block_Catalog_Category_Tab_Product extends Mage_Adminhtml_B
                 'category_id=' . (int) $this->getRequest()->getParam('id', 0),
                 'left'
             );
+
         $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
 
         $this->setCollection($collection);
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Crosssell.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Crosssell.php
@@ -139,7 +139,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Crosssell extends Mage_Admin
             'image',
             [
                 'header' => Mage::helper('catalog')->__('Image'),
-                'type'  => 'image',
+                'type'  => 'productimage',
                 'index' => 'image',
             ]
         );
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Crosssell.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Crosssell.php
@@ -95,6 +95,8 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Crosssell extends Mage_Admin
             $collection->addFieldToFilter('entity_id', ['in' => $productIds]);
         }
 
+        $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
+
         $this->setCollection($collection);
 
         return parent::_prepareCollection();
@@ -132,6 +134,15 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Crosssell extends Mage_Admin
             'width'     => 60,
             'index'     => 'entity_id'
         ]);
+
+        $this->addColumn(
+            'image',
+            [
+                'header' => Mage::helper('catalog')->__('Image'),
+                'type'  => 'image',
+                'index' => 'image',
+            ]
+        );
 
         $this->addColumn('name', [
             'header'    => Mage::helper('catalog')->__('Name'),
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
@@ -97,7 +97,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Related extends Mage_Adminht
         }
 
         $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
-        
+
         $this->setCollection($collection);
         return parent::_prepareCollection();
     }
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
@@ -141,7 +141,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Related extends Mage_Adminht
             'image',
             [
                 'header' => Mage::helper('catalog')->__('Image'),
-                'type'  => 'image',
+                'type'  => 'productimage',
                 'index' => 'image',
             ]
         );
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Related.php
@@ -96,6 +96,8 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Related extends Mage_Adminht
             $collection->addFieldToFilter('entity_id', ['in' => $productIds]);
         }
 
+        $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
+        
         $this->setCollection($collection);
         return parent::_prepareCollection();
     }
@@ -134,6 +136,15 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Related extends Mage_Adminht
             'width'     => 60,
             'index'     => 'entity_id'
         ]);
+
+        $this->addColumn(
+            'image',
+            [
+                'header' => Mage::helper('catalog')->__('Image'),
+                'type'  => 'image',
+                'index' => 'image',
+            ]
+        );
 
         $this->addColumn('name', [
             'header'    => Mage::helper('catalog')->__('Name'),
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
@@ -106,6 +106,8 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Upsell extends Mage_Adminhtm
             $collection->addFieldToFilter('entity_id', ['in' => $productIds]);
         }
 
+        $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
+
         $this->setCollection($collection);
         return parent::_prepareCollection();
     }
@@ -134,6 +136,16 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Upsell extends Mage_Adminhtm
             'width'     => 60,
             'index'     => 'entity_id'
         ]);
+        
+        $this->addColumn(
+            'image',
+            [
+                'header' => Mage::helper('catalog')->__('Image'),
+                'type'  => 'image',
+                'index' => 'image',
+            ]
+        );
+
         $this->addColumn('name', [
             'header'    => Mage::helper('catalog')->__('Name'),
             'index'     => 'name'
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
@@ -141,7 +141,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Upsell extends Mage_Adminhtm
             'image',
             [
                 'header' => Mage::helper('catalog')->__('Image'),
-                'type'  => 'image',
+                'type'  => 'productimage',
                 'index' => 'image',
             ]
         );
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Upsell.php
@@ -136,7 +136,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Edit_Tab_Upsell extends Mage_Adminhtm
             'width'     => 60,
             'index'     => 'entity_id'
         ]);
-        
+
         $this->addColumn(
             'image',
             [
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
@@ -159,7 +159,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
             'image',
             [
                 'header' => Mage::helper('catalog')->__('Image'),
-                'type'  => 'image',
+                'type'  => 'productimage',
                 'index' => 'image',
             ]
         );
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
@@ -61,7 +61,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
                 'left'
             );
         }
-        
+
         $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
 
         if ($store->getId()) {
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
@@ -46,11 +46,11 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
     {
         $store = $this->_getStore();
         $collection = Mage::getModel('catalog/product')->getCollection()
-            ->addAttributeToSelect('sku')
-            ->addAttributeToSelect('name')
-            ->addAttributeToSelect('attribute_set_id')
-            ->addAttributeToSelect('type_id');
-
+        ->addAttributeToSelect('sku')
+        ->addAttributeToSelect('name')
+        ->addAttributeToSelect('attribute_set_id')
+        ->addAttributeToSelect('type_id');
+        
         if (Mage::helper('catalog')->isModuleEnabled('Mage_CatalogInventory')) {
             $collection->joinField(
                 'qty',
@@ -61,6 +61,9 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
                 'left'
             );
         }
+        
+        $collection->joinAttribute('image', 'catalog_product/image', 'entity_id', null, 'left');
+
         if ($store->getId()) {
             //$collection->setStoreId($store->getId());
             $adminStore = Mage_Core_Model_App::ADMIN_STORE_ID;
@@ -151,6 +154,16 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
                 'index' => 'entity_id',
             ]
         );
+
+        $this->addColumn(
+            'image',
+            [
+                'header' => Mage::helper('catalog')->__('Image'),
+                'type'  => 'image',
+                'index' => 'image',
+            ]
+        );
+
         $this->addColumn(
             'name',
             [
--- a/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Grid.php
@@ -50,7 +50,7 @@ class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Wid
         ->addAttributeToSelect('name')
         ->addAttributeToSelect('attribute_set_id')
         ->addAttributeToSelect('type_id');
-        
+
         if (Mage::helper('catalog')->isModuleEnabled('Mage_CatalogInventory')) {
             $collection->joinField(
                 'qty',
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php
@@ -288,6 +288,9 @@ class Mage_Adminhtml_Block_Widget_Grid_Column extends Mage_Adminhtml_Block_Widge
             case 'theme':
                 $rendererClass = 'adminhtml/widget_grid_column_renderer_theme';
                 break;
+            case 'image':
+                $rendererClass = 'adminhtml/widget_grid_column_renderer_image';
+                break;
             default:
                 $rendererClass = 'adminhtml/widget_grid_column_renderer_text';
                 break;
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php
@@ -288,8 +288,8 @@ class Mage_Adminhtml_Block_Widget_Grid_Column extends Mage_Adminhtml_Block_Widge
             case 'theme':
                 $rendererClass = 'adminhtml/widget_grid_column_renderer_theme';
                 break;
-            case 'image':
-                $rendererClass = 'adminhtml/widget_grid_column_renderer_image';
+            case 'productimage':
+                $rendererClass = 'adminhtml/widget_grid_column_renderer_productimage';
                 break;
             default:
                 $rendererClass = 'adminhtml/widget_grid_column_renderer_text';
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * OpenMage
  *
@@ -24,6 +25,12 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Image extends Mage_Adminh
 
     protected $_defaultWidth = 128;
 
+    /**
+     * Renders grid column
+     *
+     * @param   Varien_Object $row
+     * @return  string
+     */
     public function render(Varien_Object $row)
     {
         $image = $this->_checkImageIsSelected($row);
@@ -32,17 +39,22 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Image extends Mage_Adminh
             //return '<img src="' . Mage::getStoreConfig('catalog/placeholder/image_placeholder') . ' " />';
         }
         $imageDimensions = $this->getColumn()->getWidth() ?: $this->_defaultWidth;
-        $imageSrc = $this->_getHelperImage($image)
-        ->resize($imageDimensions, $imageDimensions);
+        $imageSrc = $this->_getHelperImage($image)->resize($imageDimensions, $imageDimensions);
         $imageUrl = $this->_getImageUrl($image);
         $result = '';
         $result .= '<a href="' . $imageUrl . '" title="' . basename($imageUrl) . '" target="_blank">';
-            $result .= '<img src="' . $imageSrc . '" alt="' . basename($imageUrl) . '" title="' . basename($imageUrl) . '" width="' . $imageDimensions . '" height="' . $imageDimensions . '"/>';
+        $result .= '<img src="' . $imageSrc . '" alt="' . basename($imageUrl) . '" title="' . basename($imageUrl) . '" width="' . $imageDimensions . '" height="' . $imageDimensions . '"/>';
         $result .= '</a>';
         return $result;
     }
 
-    public function renderExport(Varien_Object $row) : string
+    /**
+     * Render column for export
+     *
+     * @param Varien_Object $row
+     * @return string
+     */
+    public function renderExport(Varien_Object $row): string
     {
         $image = $this->_checkImageIsSelected($row);
         if (!$image) {
@@ -51,25 +63,41 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Image extends Mage_Adminh
         return $this->_getImageUrl($image);
     }
 
+    /**
+     * @return string|null
+     */
     public function renderCss()
     {
         return parent::renderCss() . ' a-center';
     }
 
-    protected function _getHelperImage($image) : Mage_Catalog_Helper_Image
+    /**
+     * @param string $image
+     * @return Mage_Catalog_Helper_Image
+     */
+    protected function _getHelperImage($image): Mage_Catalog_Helper_Image
     {
         $dummyProduct = Mage::getModel('catalog/product');
         return Mage::helper('catalog/image')->init($dummyProduct, $this->getColumn()->getAttributeCode(), $image);
     }
 
-    protected function _getImageUrl($image) : string {
+    /**
+     * @param string $image
+     * @return string
+     */
+    protected function _getImageUrl($image): string
+    {
         if (!$image) {
             return '';
         }
         return Mage::getBaseUrl('media') . 'catalog/product/' . $image;
     }
 
-    private function _checkImageIsSelected($row) : string|null
+    /**
+     * @param Varien_Object $row
+     * @return string|false
+     */
+    private function _checkImageIsSelected($row): string|false
     {
         $value = $this->_getValue($row);
         if (!$value || $value == 'no_selection') return false;
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
@@ -53,7 +53,7 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Image extends Mage_Adminh
 
     public function renderCss()
     {
-        return 'a-center';
+        return parent::renderCss() . ' a-center';
     }
 
     protected function _getHelperImage($image) : Mage_Catalog_Helper_Image
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Image.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2022-2023 The OpenMage Contributors (https://www.openmage.org)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Adminhtml grid item renderer product image
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ */
+class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Image extends Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Abstract
+{
+
+    protected $_defaultWidth = 128;
+    
+    protected function _getHelperImage($image)
+    {
+        $dummyProduct = Mage::getModel('catalog/product');
+        return Mage::helper('catalog/image')
+            ->init($dummyProduct, $this->getColumn()->getAttributeCode(), $image);
+    }
+
+    private function _checkImageIsSelected($row)
+    {
+        $value = $this->_getValue($row);
+        if (!$value || $value == 'no_selection') return false;
+        return $value;
+    }
+
+    public function render(Varien_Object $row)
+    {
+        $result = '';
+
+        $imageWidth = intval($this->getColumn()->getImageWidth()) > 0 ?: 128;
+        $imageHeight = intval($this->getColumn()->getImageHeight()) > 0 ?: 128;
+
+        $image = $this->_checkImageIsSelected($row);
+        if ($image) {
+            
+            $imageSrc = $this->_getHelperImage($image)
+                ->resize($imageWidth, $imageHeight);
+            $imageUrl = Mage::getBaseUrl('media') . 'catalog/product/' . $image;
+
+            $result .= '<a href="' . $imageUrl . '" title="' . basename($imageUrl) . '" target="_blank">';
+                $result .= '<img src="' . $imageSrc . '" alt="' . basename($imageUrl) . '" title="' . basename($imageUrl) . '" width="' . $imageWidth . '" height="' . $imageHeight . '"/>';
+            $result .= '</a>';
+
+        }/*  else {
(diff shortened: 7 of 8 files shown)


pquerner:
Imho this should be made optional. Our instance doesnt have images, instead they are coming from CDN anyways and we've overwritten the parts, but it would make it (I think) like we have "none" images for any product - and simply add a column for no reason.

Also I think some vendors doesnt want that even if they have their product images attached to magento.

ADDISON74:
We should have the implementation of Magento 2 grids in OpenMage, but more extensive changes are needed.

There are a lot of implementations all over the Internet and for a while I considered displaying an image in the grid as useful for quick visual identification of the product, but the feedback I received was negative. First of all, the grid was getting bigger, especially when 200 products were viewed on the page. The page loading time increases, there are requests for each individual image. 

I could agree with this PR only if there is also the option to disable the display of the column in product grid, which should be implicitly set in Backend to No.

ADDISON74:
... there must be the option of control in the Backend for each individual grid where images are displayed: **Product Grid, Category Grid, Related/Up-sells/Cross-sells Grids**.

empiricompany:
> We should have the implementation of Magento 2 grids in OpenMage, but more extensive changes are needed.
> 
> There are a lot of implementations all over the Internet and for a while I considered displaying an image in the grid as useful for quick visual identification of the product, but the feedback I received was negative. First of all, the grid was getting bigger, especially when 200 products were viewed on the page. The page loading time increases, there are requests for each individual image.
> 
> I could agree with this PR only if there is also the option to disable the display of the column in product grid, which should be implicitly set in Backend to No.

I agree, I will convert this PR into a draft because I want to immediately configure the choice of columns to display, which by default does nothing in order to avoid any compatibility and performance issues. I will complete the PR in the next few days while waiting for further feedback.

ADDISON74:
In addition, we must set up in the Backend, in the same section that controls each individual grid, the size of the image.

Displaying an image of 64 px height compared to 200 displayed products on the page or maybe even more if someone uses other custom values, will increase the use of the scroll, will at least triple the display.

empiricompany:
What do you think about last commit?

I've used traits to add columns to the grid and i've added configuration to enable product image column and setting the width.
Trait should be the safest way to avoid compatibility issues.

config

If this path can be ok, I can proceed with additional features and reorganize the configuration more effectively, for which I would dedicate a separate tab.

Same questions about:
- there is better location where to place the traits?
- i have to implement same sort of method to configure the order of columns, now i've just placed product image after entity_id

PS: i've to implement same solution (like an interface) cause traits cannot have constant, like that:
https://wiki.php.net/rfc/constants_in_traits

then

```
interface Mage_Adminhtml_Block_Widget_Grid_Config_Product_Columns {
    public const CONFIG_PATH_GRID_COLUMNS = 'admin/grid_catalog/columns';
    public const CONFIG_PATH_GRID_COLUMN_IMAGE_WIDTH = 'admin/grid_catalog/imagewith';
}
```

```
class Mage_Adminhtml_Block_Catalog_Product_Grid extends Mage_Adminhtml_Block_Widget_Grid implements Mage_Adminhtml_Block_Widget_Grid_Config_Product_Columns_Interface
{
    use Mage_Adminhtml_Block_Widget_Grid_Config_Product_Columns;
```

waiting for feedback

ADDISON74:
I would like to be able to control the display of the column in each of the 5 grids where it appears: **Product, Category, Up-Sells, Cross-Sells, Related**.

From the attached image it appears that it is a unitary control and not an individual one.

The idea of using multiselect is not a bad one

- **Catalog Grid Configuration** - it is OK, it requires the translation label
- **Columns** - OK too
- **Product Image** - here I would control the 5 grids for the image, adding the 5 options like this

Product - Image Thumbnail
Category - Image Thumbnail
Cross-Sells - Image Thumbnail
Related - Image Thumbnail
Up-Sells - Image Thumbnail

But if he has other custom grids then he will no longer have control.

I think it would have been better to have a discussion in the dedicated section about this idea and then you could proceed to its implementation.

empiricompany:
> I would like to be able to control the display of the column in each of the 5 grids where it appears: **Product, Category, Up-Sells, Cross-Sells, Related**.
> 
> From the attached image it appears that it is a unitary control and not an individual one.
> 
> The idea of using multiselect is not a bad one
> 
>     * **Catalog Grid Configuration** - it is OK, it requires the translation label
> 
>     * **Columns** - OK too
> 
>     * **Product Image** - here I would control the 5 grids for the image, adding the 5 options like this
> 
> 
> Product - Image Thumbnail Category - Image Thumbnail Cross-Sells - Image Thumbnail Related - Image Thumbnail Up-Sells - Image Thumbnail
> 
> But if he has other custom grids then he will no longer have control.
> 
> I think it would have been better to have a discussion in the dedicated section about this idea and then you could proceed to its implementation.

Yes, through the trait and the way I have organized the code, I would be able to implement dedicated configurations for each type of grid (catalog_product, catalog_category/product, upsell, related, crossell, and other grids like sales), where you can set additional columns besides the product image a its order.
i want feedback about current implementation with traits

ADDISON74:
**Some ideas related to the presentation**

1 - The label name **Enable Grids** creates confusion, because we do not enable and disable thosee grids. Another form must be found, e.g. **Available Grids**.

2 - I would name the options in the multiselect list based on a convention (main category + action), ordered alphabetically. We are removing the Catalog to reduce its length and we must quickly understand where the grids are

**Categories / Products
Products / List
Products / Edit / Associated
Products / Edit / Cross-Sells
Products / Edit / Related
Products / Edit / Up-Sells**

3 - When the product is configurable, a grid appears in **Associated Products** tab. I added it to the list of grids above.

empiricompany:
@ADDISON74, please provide a PR to explain your vision in detail and what you are trying to achieve

ADDISON74:
@empiricompany - I will provide you a feedback these days.

ADDISON74:
@empiricompany - first of all, I don't have time to implement this PR in the way I would like it. Secondly, I don't consider this feature important at the moment, I like the one in Magento 2, but I can live without it as before.

Because you proposed a PR, I tested your implementation and provided you a feedback. If it ends up in OpenMage, it must be functional and useful.

sreichel:
Mhh maybe better to fork https://github.com/jayelkaake/enhancedgrid or https://github.com/blmage/mage-enhanced-admin-grids?

Btw constants in traits are only for php 8.2.

sreichel:
> PS: phpstan report that traits cannot have constant, but using self it's ok
> i've to implement an interface like that?

Think "yes", interface should work.

Untested: i guess it would also work to add that constants to a block. or helper class. Constants should be accessible from inside the trait?

empiricompany:
> Mhh maybe better to fork https://github.com/jayelkaake/enhancedgrid or https://github.com/blmage/mage-enhanced-admin-grids?
> 
> Btw constants in traits are only for php 8.2.

I used https://github.com/blmage/mage-enhanced-admin-grids, but it is no longer compatible with OpenMage and PHP 8.2. The code is very messy and heavily relies on create_function and hacks. I haven't tested it yet, but this one seems more interesting: https://github.com/jayelkaake/enhancedgrid.

Are you suggesting that it's better to create a separate extension rather than including it in the core of OpenMage?

ADDISON74:
Sven's feedback is welcome. I think the best option would be using an extension. Those who need it can install it, in addition, we don't have to maintain it when problems are reported. It is what we did with a lot of core modules like Backup, Poll, both having their repositories

I remain of the opinion that such proposals should be discussed first and then proceed to their implementation or not.

fballiano:
I've a different opinion (sorry, I'm getting back on track, I was away for a few days), I think such funcionality would be a nice addition to the core and at least it would be maintained, many extensions become unmaintained with time sadly.

Backup/Poll were ancient module that IMHO have no reason to exist anymore in 2023, that's why they were removed, but grid customization would be awesome to have.

If there was a really good working module out there, than by any mean let's use it and don't reinvent the wheel, but with my customers we old old clanky modules that half-work nowadays so... I don't know... it doesn't seem there's any complete solution out there to reccommend.

empiricompany:
Thanks to everyone, I would like a lot of feedback on this. I am working on it a lot. The latest features seem very useful to me, such as displaying the creation/update date (I use this column a lot to identify products updated through external integrations) and the addition of a second horizontal scroll at the top if the columns don't all fit in the grid

ADDISON74:
@empiricompany - I will help you with testing and feedback if you want to continue with this PR. 

We should analyze if it can be implemented as an extension in OpenMage to be enabled/disabled. In case someone does not want to load it to have full control.

empiricompany:
> @empiricompany - I will help you with testing and feedback if you want to continue with this PR.
> 
> We should analyze if it can be implemented as an extension in OpenMage to be enabled/disabled. In case someone does not want to load it to have full control.

@ADDISON74 - The beauty is that the new feature is disabled by default, it is non-invasive, and it doesn't load anything. I would like to have this functionality officially maintained with OpenMage and not have to rewrite it as an external extension. The idea is to extend the possibility of customization to other grids as well.

fballiano:
So, I checked it out briefly, here's my personal feedback:
- I'm not sure about adding traits to the codebase, since it would be the first time (and I don't love them in general)
- @ADDISON74 it's right about the "Enable Grids" label, I don't understand what it really does
- why "created_at/updated_at" are not in the "columns" multiselect?
- I think we would need a separate configuration for every grid because having a single one seems kinda limiting

sorry, I'm not sure about this one as it is now, yes it allows to add images and to "turn off" some columns but at the moment it seems to come with a bit too many caveats, at least in my opinion.

empiricompany:
@fballiano @ADDISON74, please check the latest version. We have now separated the configuration for each grid area

@sreichel, I have simplified the trait and moved all constants to a helper class. The future idea is to create a kind of "decorator" class for each grid type (catalog_product, sales_, etc.) that the trait will read to remain generic.

I would also like some feedback on the names of the new methods and where it is most appropriate to place the classes (I was thinking of a ?new folder for Traits?). @fballiano I chose to create a trait to add new methods to the class without having to repeat them in all relevant blocks and avoid duplicated code. In my opinion, it is a good strategy, and I use them a lot in Laravel.

empiricompany:
i've updated PR description with new screenshots

ADDISON74:
A short feedback after looking over the updates.

1. If there are 1000 attributes in a store, that multiselect list will be very large. You could consider that in the management page of an attribute, add a new **Backend Properties** section to the Properties tab to have a **Use in Grid** with Yes/No options. In this way, everyone will be able to set their own short list of attributes for the grids and you won't have to list them all in the configuration.

2. Why is Product Image Preview a multiselect? I would only use the base image and not offer this option.

3. Please check that there is another grid to take into account. When the product type is configurable, this grid appears in the Associated Products tab.

4. TBD which is the best place in the backend where to add the configuration section.

5. We must work on naming the labels and then add the translations. Let's not abuse comments where it really isn't necessary.

6. TBD if it's fine to use traits, considering that I haven't really come across it in the OpenMage code.

ADDISON74:
At point #3 it cannot be multiselect, Only one image from the list can be used.

empiricompany:
> At point #3 it cannot be multiselect, Only one image from the list can be used.

why?

ADDISON74:
> > At point #3 it cannot be multiselect, Only one image from the list can be used.
> 
> why?

If I select 10 images which one will be used in the grid?

empiricompany:
> > > At point #3 it cannot be multiselect, Only one image from the list can be used.
> > 
> > 
> > why?
> 
> If I select 10 images which one will be used in the grid?

All selected ones will be displayed.
Please before quickly comment the PR test it.

ADDISON74:
I don't think it's a good idea to allow multiple images to be listed in the grid, but let's see other opinions. One alone is enough. Basically what is the role of the image? That of identifying the product faster when scrolling through the list. At the moment in OpenMage, finding one or more products can be done extremely easily by filtering the columns.

It was not a quick comment between two sips of coffee. The posted image helped me enough. I noticed that sometimes the authors of a PR tend to be subjective or to feel hurt by the comments. What can I say? It is an interesting experience to create PR's because you really have to convince as many people as possible.

fballiano:
it could be interesting to have more than one images (although probably not the case for everybody) in case different "image attributes" are important to show important part of the products and combined in different way through the catalog so, at the end of the day, why not.

fballiano:
@empiricompany is there a way now to decide the order of the attributes in the grid?

empiricompany:
> @empiricompany is there a way now to decide the order of the attributes in the grid?

Currently no, is a Improvements feature listed in the description

empiricompany:
> @empiricompany what about custom column that are not attributes? for example I have weird columns that are calculated at runtime and do not have any real column on the database :-\
> 

The attribute columns are listed as follows, so it shouldn't cause any issues:
./app/code/core/Mage/Adminhtml/Model/System/Config/Source/Grid/Catalog/Product/Column.php

> also, probably the order is an important part which could change important part of the code, we should think how we can do that, let see, for example, if there wasn't a multiselect but simply a textarea where people could input a "new line" separated list of columns, it would actually support custom column, created_at/updated_at, attributes and also the ordering... doesn't look as good but...

As reported to @ADDISON74 I was thinking about a custom multiselect form component where you can filter and sort the items. I could inject some javascript code to transform the multiselect into a more advanced one.
It should also take into account the default columns shown for sorting, for example, if I want the image to come after the name.

empiricompany:
@fballiano - The method `_prepareColumns()` is called first, so you should see your columns added through its override.

empiricompany:
@fballiano - I have added column sorting (compatible with any additional columns added by overwriting the class). The JavaScript code needs a bit of organizing, but it seems functional and lightweight to me.

@ADDISON74 - The sorting of the columns needs to be tested. I cannot show it through a screenshot.

empiricompany:
@ADDISON74 - I have added a GIF to the description to observe the behavior when sorting the columns.

empiricompany:
maybe we can add a button aside the search to enable/disable drag&drop functionality

app/code/core/Mage/Adminhtml/Model/System/Config/Source/Grid/Catalog/Product/Column.php
```diff
@@ -0,0 +1,76 @@
...
+        'visibility',
+        'status',
+    ];
+
+    /**
+     * Retrieve attributes as array
+     *
+     * @return array
+     */
+    public function toOptionArray()
```
fballiano, line 54:
```suggestion
    public function toOptionArray(): array
```

app/code/core/Mage/Adminhtml/Helper/Widget/Grid/Config.php
```diff
@@ -0,0 +1,157 @@
...
+    {
+        return $this->getStoreConfigGridId(self::CONFIG_PATH_GRID_ENABLED);
+    }
+
+    /**
+     * Get column created_at is enabled
+     *
+     * @return bool
+     */
+    public function isCreatedAtEnabled()
```
fballiano, line 107:
```suggestion
    public function isCreatedAtEnabled(): bool
```

fballiano:
I would have preferred the reordering to be in the "settings" part in order not to add javascripts to the main part that everybody uses :-(

empiricompany:
> I would have preferred the reordering to be in the "settings" part in order not to add javascripts to the main part that everybody uses :-(

Yes, me too, but the problem is that in the system config, I still can't know which columns have been added, even through code (as in the case you mentioned). It's only possible to know them after at least one initial load. There might be another solution, and it's not difficult to modify.

ADDISON74:
1 - An observation on the artistic impression. in the Magento theme I would change the red color of the column when moving. An orange shade from the buttons is much more suitable.

2 - I don't like how the **Advanced Grids** tab looks in the configuration page. A variant would be to remove the word Advanced so that only Grids remain and I would move it as the last position in the list, after Developer.

3 - I would change the word View with the action that is done on the product, which is Edit. correct would be Catalog Product Edit Related Products. In addition, I would eliminate "Grid Columns" because it is obvious what it refers to, so I would reduce the number of words and it would look much better.

4 - We should have a button to reset the grids to default. A [Reset Grid] button either on the grid/section page, for example next to [Reset Filter] button, or on the configuration page, in every section.

ADDISON74:
Regarding observation #2 this is how it should look. The only problem is to determine if it is named **Custom Grids**, **Product Grids** or just **Grids**. 

**A translation label must be created.**

custom_grid
product_grid

ADDISON74:
Regarding observation no. 3 this is how it should look.  It is based on the following rules **[Main Section in Catalog Menu] + [Action] + [Section]**

**Translations label must be created.**

headers

ADDISON74:
Regarding the content of a section.  I have marked the observations below

1 -  Changed to **Custom Grids**, **Product Grids** or just **Grids**.

2 - I prefer **Enable** and remove the label, it is useless because the action is obvious. It is all over Magento backend.

3 - **Show Created At Column** and remove the label

4 - **Show Updated At Column** and remove the label

5 - The label shouldn't use **uppercases**. It is a rule in Magento backend.

6 - Just **Product Image Width** and remove the label.

All these texts must have translations.

text

ADDISON74:
Regarding observation no. 1 this is what I would like to change, the **red background** and the pointer icon, like the one used in dropdown lists.

color

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Config/Product/Columns.php
```diff
@@ -0,0 +1,208 @@
...
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2018-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 13:
```suggestion
 * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
since all these files are new they should be like this ;-)

empiricompany:
@fballiano I have rewritten everything by removing the trait and implementing it similarly to the massaction component. 
I have updated the description where the new interface is explained in detail, which allows for easily adding additional configurations for other targets.
I need feedback from developers on how to better organize things. In my opinion, this approach should not create any backward compatibility issues and would allow for easy and organized extension to other components.

@ADDISON74 - Thank you for your feedback. I focused on refactoring the code and left the details for later. I will make some adjustments in the next few days.

empiricompany:
ah @ADDISON74 with the new solution, I easily added the configuration for the associated product grid as well!

ADDISON74:
@empiricompany - wonderful, after you finish this PR, maybe you spend some time to check two other important PRs that should have been integrated for a long time and that are in the draft stage

**GUI to Edit EAV Attributes & Sets - Custome**r https://github.com/OpenMage/magento-lts/pull/2352
**GUI to Edit EAV Attributes & Sets - Category** https://github.com/OpenMage/magento-lts/pull/2353

We really need fresh forces.

empiricompany:
> @empiricompany - wonderful, after you finish this PR, maybe you spend some time to check two other important PRs that should have been integrated for a long time and that are in the draft stage
> 
> **GUI to Edit EAV Attributes & Sets - Custome**r #2352 **GUI to Edit EAV Attributes & Sets - Category** #2353
> 
> We really need fresh forces.

Yes, I know, i had started some testing on this PR #2352, but I didn't receive any feedback, so I abandoned it. It's an interesting as complex PR, and I'm not sure if I will find the time.

empiricompany:
@ADDISON74 - I have updated the labels in the configuration, now it seems clear

ADDISON74:
**Here are my comments**

1 - You chose the right name for the link. I wouldn't insert it where it is now, but I move it to the end, because most of us would learn with the first four being there, one after another. Being an addition, a new feature, it should go to the end of the list.

tab_menu

2 - You have chosen the headers well, now it is clearer where the grids are located inside the Backend. There are still some fixes to be made, see the image bellow

1. The text must be "Display Created At Column". As it is now, the explanation is missing.

2. The text must be "Display Update At Column" (similar to point 1).

3. Words 2 and 3 must start with a capital letter. It's the rule in Backend, i.e. "Product Image Preview".

4. Words 2 and 3 must start with a capital letter (see point 3), meaning "Product Image Width".

tab_section

3 - This needs to be fixed in both themes (Magento legacy, OpenMage), also changing the caret icon. https://github.com/OpenMage/magento-lts/pull/3451#issuecomment-1698914309.

4 - All texts, labels must be able to be translated, so added to the local files.

5 - A custom grid should be able to be brought to the initial OpenMage form. A button solves this problem. Without this feature, we risk that someone will want to go back to the default grid and he doesn't know how to do it. Releasing it without a reset button would not be right.

ADDISON74:
6 - When the grid is arranged by default, the two columns "Created At" and "Updated At" are enabled, both should be displayed on the right, before the Action column. later the user can move them where he wants.

ADDISON74:
**Question 1** 
What happens with the custom arrangement of the columns when there are more users? Is the arrangement available for each individual?

empiricompany:
> **Question 1** What happens with the custom arrangement of the columns when there are more users? Is the arrangement available for each individual?

No, it is not possible. The order of the columns is saved in the global configuration

empiricompany:
@ADDISON74 - i've changed system config labels following your indications

empiricompany:
Github tells me that Openmage has deleted one of my comments 2 days ago. Can I know which one it was and what was wrong with it? It's a bit frustrating.

ADDISON74:
> > **Question 1** What happens with the custom arrangement of the columns when there are more users? Is the arrangement available for each individual?
> 
> No, it is not possible. The order of the columns is saved in the global configuration

This is a bit nasty because if I want a certain arrangement and someone else changes it, I will go and throw the coffee cup on his shirt :)

ADDISON74:
> Github tells me that Openmage has deleted one of my comments 2 days ago. Can I know which one it was and what was wrong with it? It's a bit frustrating.

It was the discussion between us when I pointed you those two PRs to look at. Because it is increasingly difficult for someone else to follow the discussions, I removed my post and yours. That's why I say more and more often that it's better to discuss and comment separately. Even this post I am writing right now is useless now and I have to delete it.

fballiano:
I actually prefer this way, to have a single grid for all users, I understand the need but the M2 implementation always caused me troubles because somebody was saying "I don't see this and that" and nobody understood what was the problem because everybody sees a different thing.

ADDISON74:
> @ADDISON74 - i've changed system config labels following your indications

All right I appreciate you solved point no. 2. I am waiting for your opinion related to the others.

empiricompany:
@ADDISON74 - Where would you place the "reset column order" button? Is it better in the grid or in the configuration?

empiricompany:
> 1 - You chose the right name for the link. I wouldn't insert it where it is now, but I move it to the end, because most of us would learn with the first four being there, one after another. Being an addition, a new feature, it should go to the end of the list.

in my opinion it's better where it is now, close to the admin settings, not under the developer settings.

empiricompany:
> I actually prefer this way, to have a single grid for all users, I understand the need but the M2 implementation always caused me troubles because somebody was saying "I don't see this and that" and nobody understood what was the problem because everybody sees a different thing.

I agree, and I have added in the features list to be implemented the possibility to restrict users the ability to sort columns

ADDISON74:
> @ADDISON74 - Where would you place the "reset column order" button? Is it better in the grid or in the configuration?

Let's discuss separately in the Discussions section. When we find the right place you can implement it. There are several aspects that must be taken into account.

ADDISON74:
The following labels and comments used in the configuration section must be placed in a local CSV file as well

```
Advanced Grid Configuration
Advanced Grids
Display Created At Column
Display Updated At Column
Product Image Preview
Product Image Width
Select product attributes to show as column in grid
Select product media image type
Set product image preview width

Catalog > Manage Products
Catalog > Manage Products > Edit | Related Product
Catalog > Manage Products > Edit | Up-Sells
Catalog > Manage Products > Edit | Cross-Sells
Catalog > Manage Products > Edit | Associated Products
Catalog > Manage Categories | Category Products
```

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1679,5 +1679,526 @@
...
+            <groups>
+                <productGrid>
+                    <label>Catalog > Manage Products</label>
+                    <sort_order>1</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <fields>
+                        <enabled>
+                            <label>Enable</label>
```
ADDISON74, line 1699: All **Enable** words must be replaced with **Enabled** to follow the Backend rule related to activate a feature with two options Yes/No.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1679,5 +1679,526 @@
...
+                </productGrid>
+                <related_product_grid>
+                    <label>Catalog > Manage Products > Edit | Related Products</label>
+                    <sort_order>2</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <fields>
+                        <enabled>
+                            <label>Enable</label>
```
ADDISON74, line 1784: All **Enable** words must be replaced with **Enabled** to follow the Backend rule related to activate a feature with two options Yes/No.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1679,5 +1679,526 @@
...
+                </related_product_grid>
+                <up_sell_product_grid>
+                    <label>Catalog > Manage Products > Edit | Up-Sells</label>
+                    <sort_order>3</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <fields>
+                        <enabled>
+                            <label>Enable</label>
```
ADDISON74, line 1869: All **Enable** words must be replaced with **Enabled** to follow the Backend rule related to activate a feature with two options Yes/No.

empiricompany:
Again, "OpenMage has deleted your comment 1 hours ago." Sorry, but I'm a bit confused, and I'm thinking of starting a personal project because my clients need a clean solution like this and here there is no collaboration

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1679,5 +1679,526 @@
...
+                </cross_sell_product_grid>
+                <super_product_links>
+                    <label>Catalog > Manage Products > Edit | Associated Products</label>
+                    <sort_order>5</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <fields>
+                        <enabled>
+                            <label>Enable</label>
```
ADDISON74, line 2039: All **Enable** words must be replaced with **Enabled** to follow the Backend rule related to activate a feature with two options Yes/No.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1679,5 +1679,526 @@
...
+                </super_product_links>
+                <catalog_category_products>
+                    <label>Catalog > Manage Categories | Category Products</label>
+                    <sort_order>6</sort_order>
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>0</show_in_website>
+                    <show_in_store>0</show_in_store>
+                    <fields>
+                        <enabled>
+                            <label>Enable</label>
```
ADDISON74, line 2124: All **Enable** words must be replaced with **Enabled** to follow the Backend rule related to activate a feature with two options Yes/No.

ADDISON74 (changes requested):

app/code/core/Mage/Adminhtml/controllers/GridController.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Check is allowed access to action
+     *
+     * @return true
+     */
+    protected function _isAllowed()
+    {
+        return true;
```
fballiano, line 49: I think we have to check for an ACL here

fballiano:
I was testing this PR and I committed some super-minor polishing.

I was thinking:
- do you think it would be a nice idea to add a configuration for "enabled drag&drop column reorder"? because now, after customizing which columns should appear or not, the drag&drop stays enabled forever, which may not be desirable
- do you think we could remove the "created at/updated at" configuration and create a new multiselect (different from the one containing all the attributes) that shows all of the columns in the grid that are not attributes? and which would make possible to hide/show also all of the default columns?

fballiano:
@ADDISON74 I should have added all the strings to the CSV, apart from some things that I'd like to check if we should do differently as per [my previous commit](https://github.com/OpenMage/magento-lts/pull/3451#issuecomment-1719722096) ;-)

ADDISON74:
> Again, "OpenMage has deleted your comment 1 hours ago." Sorry, but I'm a bit confused, and I'm thinking of starting a personal project because my clients need a clean solution like this and here there is no collaboration

@empiricompany - I explained to you in another message that you definitely received that I will delete the off-topic conversation between us because we already have a PR that can no longer be followed. I deleted the respective messages, it says that I was the one who did this action and you ask again the same question? This time I will leave these messages, including this one, at the risk of making a long conversation out of this PR.

ADDISON74:
> I was testing this PR and I committed some super-minor polishing.
> 
> I was thinking:
> 
> * do you think it would be a nice idea to add a configuration for "enabled drag&drop column reorder"? because now, after customizing which columns should appear or not, the drag&drop stays enabled forever, which may not be desirable
> * do you think we could remove the "created at/updated at" configuration and create a new multiselect (different from the one containing all the attributes) that shows all of the columns in the grid that are not attributes? and which would make possible to hide/show also all of the default columns?

I agree with both observations:

1. Column arrangement is a feature of a grid and it should be enabled/disabled. By default it should be set to No. The following case must be analyzed. All available grids have the option "Enabled as Yes/No". If we add a new option "Enable Column Arrangement" as Yes/No, should it or not on depend the first being set to Yes? I think that arranging/moving the columns should be conditional on the custom grid being enabled.

2. Those two values for "Created At" and "Updated At", which are now highlighted, should be in the multiselect list, in one place with all the others.

fballiano:
2. can't be done in that same multiselect because in the multiselect there are only product attributes, while created_at/updated_at are not ;-)

fballiano:
@ADDISON74 I'd implement the "enable column arrangement" this way:
Screenshot 2023-09-15 alle 00 59 13

meaning, I'd prefer to have one single "enable column arrangement" instead of having one per every grid. why is this? because you may need to enable/disable the advanced grid funcionality "per grid", but you will enable the drag&drop, reconfigure all your grids as you deside, disable drag&drop. and is seems much easier for the user to have only one.

is it ok for you both?

app/code/core/Mage/Adminhtml/controllers/GridController.php
empiricompany: Yes, it is in the list of features to be implemented

empiricompany:
new feature:
- a new button to enable/disable column sorting 
- a new button to reset column sorting to default

empiricompany:
> @ADDISON74 I'd implement the "enable column arrangement" this way: Screenshot 2023-09-15 alle 00 59 13
> 
> meaning, I'd prefer to have one single "enable column arrangement" instead of having one per every grid. why is this? because you may need to enable/disable the advanced grid funcionality "per grid", but you will enable the drag&drop, reconfigure all your grids as you deside, disable drag&drop. and is seems much easier for the user to have only one.
> 
> is it ok for you both?

@fballiano - i have implement it in a different way more user friendly, check it out!

empiricompany:
> > Again, "OpenMage has deleted your comment 1 hours ago." Sorry, but I'm a bit confused, and I'm thinking of starting a personal project because my clients need a clean solution like this and here there is no collaboration
> 
> @empiricompany - I explained to you in another message that you definitely received that I will delete the off-topic conversation between us because we already have a PR that can no longer be followed. I deleted the respective messages, it says that I was the one who did this action and you ask again the same question? This time I will leave these messages, including this one, at the risk of making a long conversation out of this PR.

Okay, but put yourself in my shoes. I lose track of the conversation, and it seems like I've written something wrong. 
I don't like doing it this way, but anyway, it's okay

fballiano:
uuuu me liky this new version!!!

empiricompany:
> uuuu me liky this new version!!!

@fballiano I need to fix some minor issues when there are multiple grids like related / crosssell products. ACL permission and  factory are missing, but we are close to a version ready for review
I hope this will be a base for other grids like order list

fballiano:
I've just committed some minor change but I think the current implementation of the "rearrange columns" buttons brings some problems:

First one is that it loads the new drag&drop js also when it's the rearranging mode is disabled (that's probably easy to fix).

Second is that we need 2 different labels for this button "rearrange columns" and "exit rearrangement" or something like that, which has to be switched in javascript

Third and more complicated that the previously the buttons layout was setup in order to work also at its lowest resolution possible:
Screenshot 2023-09-17 alle 23 16 58

but now all the buttons are breaking it:
Screenshot 2023-09-17 alle 23 17 24

I think that having the drag&drop enabled/disabled in the config section would have:
- avoided making a more complex javascript
- made it easier to load or not load the scripts
- avoided adding the buttons to the layout

So, I'm not sure how to proceed here.

empiricompany:
@fballiano - Good observations... The first and second issues are easy to fix. Regarding the third problem, it seems that the minimum resolution for the current theme is 960px


actually minimum resolution to fit all ui buttons is 1193px


Adding a _display; flex_ to `.filter-actions ` improves a little the result:
filter_action_flex
(this is at 960px)

But we should decide which is the minimum supported resolution is. Could it be 1024px?

--- 

I would like to keep these controls in the grid rather than having to go back and forth in the configuration (where a reset foreach grid would also be implemented). I'm thinking of a possible solution: I was experimenting with adding a cog icon that, when clicked, opens a popover where we can also include other options (for example, saving specific views, enabling/disabling columns, etc.).
cog_icon

Another solution could be like https://github.com/empiricompany/mage-enhanced-admin-grids/ but adds a lot of space
mage-enhanced-grids

js/mage/adminhtml/grid.js
```diff
@@ -947,6 +947,7 @@ class varienGridAdvanced {
     }
 
     enableColumnsRearrangement() {
+        document.getElementById(this.grid.containerId + '_table').setAttribute('style', 'margin-top: 10px');
```
empiricompany, line 950: why this? it adds a white space in legacy theme
why

fballiano: it is to move the "triangle" simbol above the grid, instead of below the heading. because below the headers it get hidden but the column that's being dragged so it's not visible

fballiano:
the thing is that I don't think people will rearrange columns all the time, they will set it up and forget it, same as adding new columns. that's why I think that having those buttons displayed all the times is not very useful.

also I think I found a bug, if I click the "reset order" on a grid which is NOT the main product grid, a js error is shown and the reset doesn't work.

js/mage/adminhtml/grid.js
fballiano: Screenshot 2023-09-18 alle 11 04 20

ADDISON74:
I didn't have time to check the new changes in the last week. Those two added buttons I hope appear only when arranging the columns is enabled.

ADDISON74:
This is a complex PR which should have been managed in its own repository as an OpenMage extension. I hope that in the future the starting skeleton will be done in a PR then changes will be made in as many PRs as possible so that it is as easy as possible to follow and approve.

empiricompany:
> the thing is that I don't think people will rearrange columns all the time, they will set it up and forget it, same as adding new columns. that's why I think that having those buttons displayed all the times is not very useful.
> 

It can be coupled with your idea in the configuration to enable/disable this UI for rearranging columns. We also need to implement a permission ACL

> also I think I found a bug, if I click the "reset order" on a grid which is NOT the main product grid, a js error is shown and the reset doesn't work.

@fballiano fixed here [fix multi-grid reset button handler](https://github.com/OpenMage/magento-lts/pull/3451/commits/974b792e56ae02ffd1f45e9bdd170ec0aa50d8fa)

fballiano:
I'd agree of having a PR for the main funcionality and add more later with new PRs, but if nobody approves then it makes sense to continue in the single PR, until it makes sense obviously ;-)

js/mage/adminhtml/grid.js
empiricompany: cause you have changed the arrow to be displayed on top, if you not like the bottom arrow we can add a vertical line
like this:

fballiano: as I wrote before, if you have it on the bottom, it gets hidden by the whole column being dragged and it's not visible.

maybe the vertical line is ok, it it's visible during dragging.

fballiano: done!

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Advanced/Abstract.php
```diff
@@ -0,0 +1,100 @@
...
+ */
+abstract class Mage_Adminhtml_Block_Widget_Grid_Advanced_Abstract extends Mage_Adminhtml_Block_Widget
+{
+    /**
+     * Sets Advanced Gird template
+     */
+    public function __construct()
+    {
+        parent::__construct();
+        $this->setResetConfirmText(Mage::helper('core')->jsQuoteEscape(Mage::helper('core')->__('Are you sure?')));
```
fballiano, line 33: @empiricompany I've changed this text, the previous one was slightly wrong ("resetted" is a typo actually) but most of all this way translations are already in every language pack

app/code/core/Mage/Adminhtml/controllers/GridController.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * Check is allowed access to action
+     *
+     * @return bool
+     */
+    protected function _isAllowed()
+    {
+        return Mage::getSingleton('admin/session')->isAllowed(static::ADMIN_RESOURCE);
```
fballiano, line 55: @empiricompany I've introduced this mega basic ACL check, it checks for "manage products". I've been thinking about it for a while. The problem is, do we want to introduce an ACL for every grid? It seems very overkill.

At the moment there has never been an ACL for modifying the backend interface, since it was never possible before.

fballiano: I was doing some tests today and I was thinking, what if...
- system->configuration page has a button "enable column rearrangement"
- the buttons just sets a variable in the session
- the current logged-in user can do the d&d on the grids

with this we wouldn't need ACL at all and it would remove the 2 buttons from the grid page.

what do you think?

empiricompany: I'm not a fun of storing this in session I want to keep same behavior across browser with the same account

I prefer to combine your proposed system config select to enable rearrangement of columns across all grids and set this to No by default


then add an acl per user role to permit to change this config and others

and also check it in grid controller action

in my next time slot dedicated to this project I propose you this behavior

empiricompany:
@fballiano - thanks for the changes (but I can't accept or decline your implementations because they are directly made to this source and not through a separate PR to my repo)

Regarding your question: there isn't a specific reason, but I would always load the double scroll because it is useful and it comes into play at lower resolutions. In fact, we can move it to the grid.

I will test it in the next days, thanks!

fballiano:
you can revert the changes if you feel so, I did only we we agreed on.

I asked about the doublescroll because it didn't feel related to this PR (the name) and I didn't read the content of the file. Now I think I get why is that. it does the horizontal scrolling of the grid.

empiricompany:
> you can revert the changes if you feel so, I did only we we agreed on.

Sure, should I manually revert your single commits eventually then force a push, or is there a simpler method (maybe from github ui)?

fballiano:
if you git revert a commit you don't have to force push

fballiano:
AFAIK ACLs can't be applied to system/configurations

js/mage/adminhtml/grid-double-scroll.js
```diff
@@ -0,0 +1,97 @@
...
+        scrollbarTop.appendChild(document.createElement('div'));
+        scrollbarTop.style.overflow = 'auto';
+        scrollbarTop.style.overflowY = 'hidden';
+        scrollbarTop.firstChild.style.height = '0';
+        scrollbarTop.firstChild.style.paddingTop = '1px';
+        scrollbarTop.firstChild.appendChild(document.createTextNode('\xA0'));
+        this.wrapperScrollBar.parentNode.insertBefore(scrollbarTop, this.wrapperScrollBar);
+
+        this.scrollbarTop = scrollbarTop;
+    }
```
fballiano, line 72: I've read this file in more detail, the only thing that I'd try to do, in order to avoid as much javascript as possible, is to have this div directly into the `widget/grid.phtml` it should work ;-)

empiricompany: I don't think it's so better only for save a few lines because if someone overwrites grid.phtml we lost it.

If you care about its size we can minify it as mage/adminhtml/grid-double-scroll.min.js and distribute also the original source mage/adminhtml/grid-double-scroll.js and can be done to others files like grid.js.

More javascript is not necessary bad if it's usefull, IMHO this is a usefull feature because if the columns don't fit in the window and I want to view the last column in a particular row, I have to scroll down to the end and then up to that row.
If we compare to other modern backends like Magento 2, our JS/Memory footprint is better

fballiano: I really woulnd't compare anything to M2 and I still believe, also from a performance standpoint, there's no need for element injection in js

empiricompany: Okay, let's leave this decision pending for now

fballiano: also, grid.phtml is already modified by this PR

empiricompany: have you tested if works correctly when there is more grids like in product view (related, upsell etc..) ?

empiricompany: This script, as it is, is independent from the template. The only requirement is that the table is wrapped inside a div.hor-scroll, so it would also work on third-party implementations with a custom template that not include div.hor-scroll-top

fballiano: at the moment the script is called by a modified version of the template so it wouldn't work. I understand that it is callable by any situation but I'll never agree that many lines of javascript are better than 1 line of html + 1 line of css, both for readability and performance.
people may have a customized grid.html (doubt it but possible) but they could also have customized layout xml so the script won't be loaded, both approaches have pros and cons but the script is not 100% better.

empiricompany: > at the moment the script is called by a modified version of the template so it wouldn't work

Then the script should be modified to work with this external div, with all handling of the events relative to the current grid

> people may have a customized grid.html (doubt it but possible) but they could also have customized layout xml so the script won't be loaded, both approaches have pros and cons but the script is not 100% better.

This is not necessarily true, the script should always be injected as grid.js unless explicitly removed in a layout update. 
IMHO I prefer to keep grid.phtml clean without a div that is not clear what it does, in terms of performance it doesn't change anything. 

But these are just details, the priority for me is that PR should be tested and reviewed on its implementation solution with interface contract, naming convention, etc.

sreichel:
For system config ACL read https://alanastorm.com/custom_magento_system_configuration/

Maybe it helps.

sreichel:
Suggestions:

- add strict type to all new files
- add type hints to all new methods/properties 
- try to pass phpstan level 9 for all new classes

empiricompany:
> Does it still work to add grid columns via observer?

yes

empiricompany:
> For system config ACL read https://alanastorm.com/custom_magento_system_configuration/
> 
> Maybe it helps.
> 
 
can't add an acl for system config

sreichel:
> can't add an acl for system config

If it does not work to set ACL "one level deeper" to config section i think its worth to raise an issue.

empiricompany:
> > can't add an acl for system config
> 
> If it does not work to set ACL "one level deeper" to config section i think its worth to raise an issue.

no it's a different type of ACL
I needed it to limit the display of column configurations or permission to rearrange orders for example for each user, but it is not possible to do so.

empiricompany:
> I am not sure about it, but EnhancedGrid extension has a user based grid config. (or it was a modified version of it)

yes it has it, but it works differently it doesn't save in system config but it has two tables

Issue #3459, opened 2023-08-21
Google Maps stopped working on all OpenMage instances after update to 3.54.1


Tomasz-Silpion:
### Summary (*)

It seems that last Friday Google mass updated their maps API to  version **3.54** https://developers.google.com/maps/documentation/javascript/versions which is totally, no longer compatible with _prototype.js_ Array.from = $A; override.

Previously warning was indicated that _This site overrides Array.from() with an implementation that doesn't support iterables, which could cause Google Maps JavaScript API v3 to not work correctly._ now it causes error that prevents maps to display:

```
Oops! Something went wrong.
This page didn't load Google Maps correctly. See the JavaScript console for technical details.
```
![google-maps-354-vs-prototype](https://github.com/OpenMage/magento-lts/assets/5328659/c355fb9f-312d-432c-a531-f2f71fda6dbb)
So far, I've found that similar issue was just reported on https://www.googlecloudcommunity.com/gc/Integration-Services/google-places-API-3-54-1-not-working/m-p/623266

### Examples (*)

1) Paste Google Map render example such as the V3 one specified in https://developers.google.com/maps/documentation/javascript/v2tov3 on any OpenMage page (even through CMS). I don't want to publicly expose mine API key here so please generate one on your own.
2) See the map fails to render due Array.from() override.

### Proposed solution

prototype.js should no longer override Array.from, let's consider how should we handle that deprecated but necessary library in next OpenMage releases.

PR #3460, opened 2023-08-21
Fix conflict between Prototype.js and Google API (#3459)

Tomasz-Silpion:
### Description (*)

Since version 3.54 Google Maps are no longer compatible with prototype.js Array.from override. Prototype lib should no longer override Array.from method in modern browsers.

1. Fixes OpenMage/magento-lts#3459

1. Paste Google Map render example such as the V3 one specified in https://developers.google.com/maps/documentation/javascript/v2tov3 on any OpenMage page (even through CMS). I don't want to publicly expose mine API key here so please generate one on your own.
2. Map should no longer fail to render due Array.from() override.
3. Please check/analyze if that prototype.js mod made in PR does not break native OpenMage frontend on admin areas.

### DIFF
--- a/js/prototype/prototype.js
+++ b/js/prototype/prototype.js
@@ -1111,7 +1111,7 @@ function $w(string) {
   return string ? string.split(/\s+/) : [];
 }
 
-Array.from = $A;
+Array.from = Array.from || $A;
 
 
 (function() {


js/prototype/prototype.js
```diff
@@ -1111,7 +1111,7 @@ function $w(string) {
   return string ? string.split(/\s+/) : [];
 }
 
-Array.from = $A;
+Array.from = Array.from || $A;
```
fballiano, line 1114: genuine question because it's hard to test with a browser that doesn't have Array.from nowadays. if Array.from is not defined, won't it cause an error instead of just evaluating the second part of the "or" statement?

fballiano: answering myself, tested with

```
Array.from = undefined;
alert(Array.from);
Array.from = Array.from || $A;
```

and it works without generating errors

kiatng: Popular browsers since 2016 support `Array.from()`, so it's probably cleaner to just remove the polyfill in line 1114 `Array.from = $A`?

fballiano: to me it's fine either ways, I guess this one is the safest approach in order to modify the library the less possible

kiatng (approved):

PR #3473, opened 2023-08-28
Fix deprecated functionality str_contains() in Design/Package.php

ma4nn:
### Description (*)
This small PR fixes a deprecation warning if e.g. `getCacheKey()` is called on a `Mage_Core_Block_Template` block instance without a template assigned (which may happen because the `template` attribute in that class is of type `string|null`):
```
Deprecated functionality: str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated  in /var/www/htdocs/app/code/core/Mage/Core/Model/Design/Package.php on line 459

#0 [internal function]: mageCoreErrorHandler(8192, 'str_contains():...', '/var/www/htdocs...', 459)
#1 /var/www/htdocs/app/code/core/Mage/Core/Model/Design/Package.php(459): str_contains(NULL, '..')
#2 /var/www/htdocs/app/code/core/Mage/Core/Model/Design/Package.php(497): Mage_Core_Model_Design_Package->getFilename(NULL, Array)
#3 /var/www/htdocs/app/code/core/Mage/Core/Block/Template.php(120): Mage_Core_Model_Design_Package->getTemplateFilename(NULL, Array)
#4 /var/www/htdocs/app/code/core/Mage/Core/Block/Template.php(352): Mage_Core_Block_Template->getTemplateFile()
#5 /var/www/htdocs/app/code/core/Mage/Core/Block/Abstract.php(1341): Mage_Core_Block_Template->getCacheKeyInfo()
```

In addition it corrects the return type for `getTemplate()`.

1. Output the cache key of a new block instance somewhere: 
`<?= $this->getLayout()->createBlock('core/template', 'test')->getCacheKey() ?>`

### DIFF
--- a/app/code/core/Mage/Core/Block/Template.php
+++ b/app/code/core/Mage/Core/Block/Template.php
@@ -86,7 +86,7 @@ class Mage_Core_Block_Template extends Mage_Core_Block_Abstract
     /**
      * Get relevant path to template
      *
-     * @return string
+     * @return string|null
      */
     public function getTemplate()
     {
@@ -117,7 +117,7 @@ class Mage_Core_Block_Template extends Mage_Core_Block_Abstract
         if ($area) {
             $params['_area'] = $area;
         }
-        return Mage::getDesign()->getTemplateFilename($this->getTemplate(), $params);
+        return Mage::getDesign()->getTemplateFilename((string)$this->getTemplate(), $params);
     }
 
     /**


app/code/core/Mage/Core/Block/Template.php
```diff
@@ -117,7 +117,7 @@ public function getTemplateFile()
         if ($area) {
             $params['_area'] = $area;
         }
-        return Mage::getDesign()->getTemplateFilename($this->getTemplate(), $params);
+        return Mage::getDesign()->getTemplateFilename((string)$this->getTemplate(), $params);
```
fballiano, line 120:
```suggestion
        return (string)Mage::getDesign()->getTemplateFilename($this->getTemplate(), $params);
```
what if we do this instead? pros/cons?

ma4nn: @fballiano nice thanks for your rapid feedback :-) Unfortunately this would not fix the issue because the `str_contains()` deprecation warning is in `Mage_Core_Model_Design_Package::getFilename()` which is called by `Mage_Core_Model_Design_Package::getTemplateFilename()`

Imho the optimum solution would be to change `getFilename()` and all related methods to use proper type hints but this would require rather more changes and therewith risk.

fballiano: mmm I see, but if $this->getTemplate() returns null, Mage::getDesign()->getTemplateFilename() would also not return anything right? if it is so, then maybe instead of calling Mage::getDesign()->getTemplateFilename() we should add an if/then statement?

kiatng:
Ref https://github.com/OpenMage/magento-lts/blob/b569b8822f4c9a801b88c3778dab0182b280ea8d/app/code/core/Mage/Core/Block/Template.php#L60-L65

If `setTemplate()` is not called, `$_template` is `null`, and `getTemplate()` returns `null`.  If we declare it as string:

```php
     * @var string
     */
    protected $_template = '';
```

It will now be type `string` and not `null`. I think this is a good solution. What do you think?

app/code/core/Mage/Core/Block/Template.php
ma4nn: If `getTemplate()` returns `null`, `getTemplateFilename()` returns - apart from the warning - the path without the filename, e.g. `frontend/base/default//template/`.

With my proposed solution the logic/return values stay exactly the same, only the deprecation warning is fixed.

ma4nn:
@kiatng thanks for your feedback. Yes that would be a good solution and more logical 👍 
Personally I opted for the other solution, because I wasn't sure if we should change the `getTemplate()` API (which probably is used very often) and make e.g. calls like `if ($this->getTemplate() === null)` fail then.

So my solution is probably the one with the lower impact, but yours is the more comprehensive one.

kiatng:
We can do a search on `getTemplate()`: https://github.com/search?q=repo%3AOpenMage%2Fmagento-lts+getTemplate%28%29&type=code&p=1

There is no `if ($this->getTemplate() === null)` type of condition. But there might be 3rd-party code that has it. Still, my opinion is to declare `protected $_template = '';` in ver 20.

fballiano:
picking up with this after a while, I'd also like what @kiatng suggested since all of the getTemplate() calls are expecting a string, this way we fix all of them ;-)

@ma4nn would you change this PR?

fballiano:
This PR is not editable by admins so I can't fix it. If it doesn't get updated I'll close it and start a new one.

Issue #3474, opened 2023-08-29
Composer patches not applied successfully


ma4nn:
### Preconditions (*)
1. OpenMage 20.1.0
2. PHP 8.1
3. Mac Ventura 13.4.1

### Steps to reproduce (*)
1. Setup clean OpenMage project according to [instructions](https://github.com/OpenMage/magento-lts#composer), especially enabling composer patches

### Expected result (*)
1. Composer patches should be applied successfully with e.g. `composer install`

### Actual result (*)
1. Applying composer patches for `shardj/zf1-future` fail with error:
```
  - Applying patches for shardj/zf1-future
    https://raw.githubusercontent.com/OpenMage/composer-patches/74862e7a9fa29dd62c5c02d585d931ce9e3c5815/patches/MAG-1.1.1.patch (MAG-1.1.1)
patch '-p1' --no-backup-if-mismatch -d '/var/www/vendor/shardj/zf1-future' < '/var/folders/bw/lwdvq1d17lz45m5mb4677vqm0000gn/T/64ed951f11b24.patch'
patching file 'library/Zend/Mime.php'
Patch creates file that already exists!  Assume -R? [y]
```

Imho the issue occurs because all patches use `/dev/null` as base which according to `patch` man-page try to create a new file:
> you can create a file by sending out a diff that compares a null file to the file you want to create.  If the file you want to create already exists in the target directory when the diff is applied, then patch will identify the patch as potentially reversed and offer to reverse the patch.

fballiano:
on mac you've to install gpatch because the macOS patch is differente somehow

ma4nn:
@fballiano yes you're right `brew install gpatch && brew link gpatch --force --overwrite` fixed the issue, thanks 🥳 And it is also mentioned already in the OM requirements - shame on me 🙈 

After this fix only 2 patches are still failing to apply:
1. `MAG-1.9.3.9.patch` 
2. `OM-ZF1F-344.patch`

-> both seem to be already merged into `shardj/zf1-future`

Perhaps it would make sense to use another composer patches plugin such as [vaimo/composer-patches](https://github.com/vaimo/composer-patches/blob/master/docs/USAGE_BASIC.md#version-constraints) that can handle version constraints for the patches?

ma4nn:
I see, this issue has already been raised in https://github.com/OpenMage/magento-lts/issues/3445.

[...]

PR #3475, opened 2023-08-29
Fixed ZF1F version in composer.json to avoid conflicts with our patches

fballiano:
when implementing our composer patches for ZF1F we didn't consider that our 20.1.0 tagged release can't now run a `composer install` because ZF1F is now been updated and the patches can't be applied anymore.

Explanation at https://github.com/OpenMage/magento-lts/issues/3474

Thus we have to specify the exact version number of ZF1F in the composer.json.

Fixes https://github.com/OpenMage/magento-lts/issues/3474

### DIFF
--- a/.github/workflows/composer.yml
+++ b/.github/workflows/composer.yml
@@ -26,4 +26,4 @@ jobs:
           restore-keys: ${{ runner.os }}-composer-
 
       - name: Validate composer
-        run: composer validate --strict
+        run: composer validate --strict --no-check-all
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "pelago/emogrifier": "^7.0",
         "phpseclib/mcrypt_compat": "^2.0.3",
         "phpseclib/phpseclib": "^3.0.14",
-        "shardj/zf1-future": "^1.22",
+        "shardj/zf1-future": "1.23.5",
         "symfony/polyfill-php74": "^1.27",
         "symfony/polyfill-php80": "^1.27",
         "symfony/polyfill-php81": "^1.27"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "902e142809dc8d50b1e628dad67320a2",
+    "content-hash": "d2c7d1ef752587a200077be1ef2a74de",
     "packages": [
         {
             "name": "colinmollenhour/cache-backend-redis",


ADDISON74 (dismissed):

fballiano:
composer validation is complaining but we have to do it this way... I don't see any other possibility

ma4nn:
> composer validation is complaining but we have to do it this way... I don't see any other possibility

as mentioned in https://github.com/OpenMage/magento-lts/issues/3474 another possibility would be to use composer patches plugins such as [vaimo/composer-patches](https://github.com/vaimo/composer-patches/blob/master/docs/USAGE_BASIC.md#version-constraints) that can handle version constraints for the patches

fballiano:
so, I was checking this matter more in detail, but I still think we should have a strict version dependency here.

let's make this example
- now main is compatible with latest version of zf1f
- we can't add any version constraint to the patches because we have no idea if some of them will ever be integrated in zf1f in the future
- 1.23.5 could have one of the patches, or not, so we can't constraint "now" to "use this patch only if <=1.23.5 because we may need the patch in the next version, or not

fballiano:
@kiatng @elidrissidev @colinmollenhour @ADDISON74 we've to address this as soon as possible and tag a new release, at this moment all installations via composer are broken

ma4nn:
@fballiano yes you're right, there is no optimal solution. 

My (untested) idea was to constraint e.g. the `MAG-1.9.3.9.patch` to `"shardj/zf1-future": "< 1.22.1"`.

Imho the whole idea of the composer patches is to live with the fact that there may be future updates to the patched library that will make your patch fail as you mentioned. In this case one could either add a `patches-ignore` section to their own `composer.json` or wait until the patch version constraint is updated with the next OpenMage version.

This approach has the benefit that you profit by updates of the `zf1-future` library without releasing a new OpenMage version which is very important imho.

fballiano:
> My (untested) idea was to constraint e.g. the `MAG-1.9.3.9.patch` to `"shardj/zf1-future": "< 1.22.1"`.

sure, but that patch is already removed from our composer.json for, for a next release, we don't have a need for the version contstraint, but it doesn't fix the problem anyway, sadly

> This approach has the benefit that you profit by updates of the `zf1-future` library without releasing a new OpenMage version which is very important imho.

with our fast pacing updates I don't think that we really need to allow zf1f updates, and allowing the upgrade with a failing patch could be tricky too :-\

ADDISON74:
From the conversations I followed carefully, I think the two patches should be removed (I think this has happened) and a new version must be released. When OM it is installed with Composer, it will get the current ZF1-Future version, according to the version mentioned in composer.json.

Also, the current OM version is either completely removed, or we correct it with a ZF1-Future version up to X, as you also showed.

The less pleasant part is that we have to be careful when a new version of ZF1-Future is launched and not to approve the automatic the PR created by the bot without checking the existing patches.

fballiano:
> The less pleasant part is that we have to be careful when a new version of ZF1-Future is launched and not to approve the automatic the PR created by the bot without checking the existing patches.

not really, because our composer requirement for zf1f allows the upgrade regardless we approve that automatic PR or not

fballiano:
@colinmollenhour dependabot will tell us 1 week later (in the worst case)

thing is, composer validation is complaining not to use exact version number...

colinmollenhour:
Is it ok to use `composer validate --strict --no-check-all`?

fballiano:
I was checking that parameter earlier but the documentation is not super clear to me, anyway I think it’s our only way

ADDISON74:
I am not an advanced Composer user, but maybe we should ask for an opinion in their forum. I noticed that there are quite interesting discussions on various topics.

https://github.com/composer/composer/discussions

fballiano:
ok from https://getcomposer.org/doc/03-cli.md it's clear that the no-check-all is the parameter we want in this case:

> --no-check-all: Do not emit a warning if requirements in composer.json use unbound or overly strict version constraints.

fballiano:
@colinmollenhour we will have to address this also in v19 because composer installation are broken in v19 at the same way...

ADDISON74:
Reading the documentation, we don't really have other options. What @colinmollenhour proposed would be the solution.

fballiano:
it's already implemented in this PR ;-)

ADDISON74 (approved):

empiricompany (approved):

Issue #3453, opened 2023-08-20
Uncaught Error: Class "Zend_Controller_Request_Http" not found in get.php


S0FTWEX:
### Preconditions (*)

Using OpenMage version 20.1.0 & PHP8.2

### Steps to reproduce (*)

1. Hit the URL of non-existent image, e.g.: https://www.domain.tld/media/catalog/category/blankx.jpg
2. Got error 'PHP message: PHP Fatal error:  Uncaught Error: Class "Zend_Controller_Request_Http" not found in get.php:62

### Solution (*)

I suggest to move the piece of code in get.php - from lines 78-84:
```
$mageFilename = 'app/Mage.php';

if (!file_exists($mageFilename)) {
    echo $mageFilename . ' was not found';
}

require_once $mageFilename;
```

to move above the line 62:
`$request = new Zend_Controller_Request_Http();`

fballiano:
I get a 404:

<img width="916" alt="Screenshot 2023-08-20 alle 23 46 16" src="https://github.com/OpenMage/magento-lts/assets/909743/62914a45-0cf2-46b8-9185-21b384eefafd">

do you need developer mode or something else in order to trigger the error?

kiatng:
I cannot replicate, I get a 404 as well on the OM 20.1.0 PHP8.2.

S0FTWEX:
> I get a 404:
> <img alt="Screenshot 2023-08-20 alle 23 46 16" width="916" src="https://user-images.githubusercontent.com/909743/261879862-62914a45-0cf2-46b8-9185-21b384eefafd.png">
> 
> do you need developer mode or something else in order to trigger the error?

My OpenMage was installed manually (extracting openmage-v20.1.0.zip file). Class `Zend_Controller_Request_Http` is defined only in vendor/shardj/zf1-future/library/Zend/Controller/Request/Http.php but 'vendor' is not mentioned in included paths in get.php (lines 32-35). So my question is: How can manually installed OpenMage load the Zend_Controller_Request_Http class in this file (get.php) if 'vendor' is not include path?

ferdiusa:
This issue can be replicated if the .htaccess inside the media/ directory is properly parsed. (this actually redirects all request through get.php)
Some apache configurations skip this (and use the /.htaccess instead of /media/.htaccess for files in subdirectories). Try accessing 
http://openmage.test/media/missingfile.jpg
[...]

PR #3476, opened 2023-08-29
Added composer patch to get.php, ref. #3453

fballiano:
If you've a webserver supporting rewrites (apache), and you try to access a /media/nonexisting.jpg file you don't get a 404, but you get an error 500.
The rewrite goes to /get.php, which was not patched for composer autoload, which fails to load some classes.

More explanations at https://github.com/OpenMage/magento-lts/issues/3453

Fixes https://github.com/OpenMage/magento-lts/issues/3453

### DIFF
--- a/get.php
+++ b/get.php
@@ -28,7 +28,7 @@ require $bp . '/app/bootstrap.php';
 /**
  * Set include path
  */
-
+$paths = [];
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'local';
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'community';
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'core';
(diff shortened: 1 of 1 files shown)


S0FTWEX (approved):
Yes, this composer vendor autoloader patch fixes the problem ;-)

get.php
```diff
@@ -28,20 +28,30 @@
...
 $appPath = implode($ps, $paths);
 set_include_path($appPath . $ps . get_include_path());
-
 include_once 'Mage/Core/functions.php';
 include_once 'Varien/Autoload.php';
 
 Varien_Autoload::register();
 
+/** AUTOLOADER PATCH **/
+$autoloaderPath = getenv('COMPOSER_VENDOR_PATH');
```
kiatng, line 45: When I tested this `$autoloaderPath = getenv('COMPOSER_VENDOR_PATH');` in my OM instance that returns 404, the value of `$autoloaderPath` is `false`.  I kinda expect it to be `true` to bypass the if block. Test script added at line 46:

```php
/** AUTOLOADER PATCH **/
$autoloaderPath = getenv('COMPOSER_VENDOR_PATH');
var_dump($autoloaderPath);
exit();
```
Output:
```
bool(false)
```

If I added my test script before line 52, the output is
```
string(33) "/home/celeraes/public_html/vendor"
```
which is expected because the if block is executed. But in my case, I don't need this patch because https://celeraeshop.com/media/catalog/category/blankx.jpg returns 404.

So, why not just add the vendor path in `$paths` and do away with the check. See
https://github.com/OpenMage/magento-lts/blob/10b63c1f65bd38f4d75a1c049d9dedaf9dffddde/lib/Varien/Autoload.php#L52-L59

Lines 56-58 check if file exists and do the `require` thing, same as your code.

get.php
```diff
@@ -28,20 +28,30 @@
 /**
  * Set include path
  */
-
+$paths = [];
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'local';
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'community';
 $paths[] = $bp . $ds . 'app' . $ds . 'code' . $ds . 'core';
 $paths[] = $bp . $ds . 'lib';
```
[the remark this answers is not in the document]
fballiano: in this case I just copied the code from index.php/Mage.php

get.php
fballiano: do you have your webserver usign the htaccess in the media folder? until I enabled it (I use Caddy now so it doesn't support htaccess) I couldn't see the error, but the error is there.

This PR just copies the code from index.php/Mage.php since get.php is actually an entry point for OpenMage, it should have the same code as index.php

kiatng (approved):

elidrissidev (approved):

Issue #3449, opened 2023-08-18
API session clean-up


colinmollenhour:
There needs to be an automatic mechanism in place to clean up expired API sessions. I assumed for a long time there was such a mechanism but this table just grows indefinitely without a cleanup script in place to delete expired sessions.

Also with the Basic Auth it would be a good improvement to bypass the recording of the session in the first place since the session only lives for the one request.

See https://github.com/OpenMage/magento-lts/pull/3443#issuecomment-1683569315

PR #3477, opened 2023-08-30
Fixed unnecessary entries in table `api-session` when using insta-login in API calls

kiatng:
…in in API calls.

### Description (*)
The API session has default expiry of 3600 secs from the time of the last API call. This expiry is managed in the table `api_asession`.  With the introduction of insta-login in PR #3443, the session expiry become irrelevant. This PR bypassed adding entries to the various api tables. In my test, this saves 10 to 20 mini-seconds to insta-login.

### Related Pull Requests
PR #3443 

Partially fixed issue #3449.

1. No new entries are added in the table `api_asession` with insta-login.
2. Normal entries are added in the table `api_asession` with "normal" login.

### DIFF
--- a/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
+++ b/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
@@ -116,16 +116,6 @@ abstract class Mage_Api_Model_Server_Handler_Abstract
     }
 
     /**
-     *  Check session expiration
-     *
-     *  @return  bool
-     */
-    protected function _isSessionExpired()
-    {
-        return $this->_getSession()->isSessionExpired();
-    }
-
-    /**
      * Dispatch webservice fault
      *
      * @param string $faultName
--- a/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
+++ b/app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
@@ -88,6 +88,21 @@ abstract class Mage_Api_Model_Server_Handler_Abstract
     }
 
     /**
+     * Allow insta-login via HTTP Basic Auth
+     *
+     * @param string $sessionId
+     * @return $this
+     */
+    protected function _instaLogin(&$sessionId)
+    {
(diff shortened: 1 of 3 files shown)


app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
```diff
@@ -100,16 +115,6 @@ protected function _isAllowed($resource, $privilege = null)
...
     }
 
-    /**
-     *  Check session expiration
-     *
-     *  @return  bool
-     */
-    protected function _isSessionExpired()
-    {
-        return $this->_getSession()->isSessionExpired();
```
kiatng, line 110: Note: This line has a bug, `isSessionExpired()` expects a param `$user`:
https://github.com/OpenMage/magento-lts/blob/2a2a2fb504247e8966f8ffc2e17d614be5d43128/app/code/core/Mage/Api/Model/Session.php#L186-L186

I've checked, this function is not used.

colinmollenhour (approved):
Very nice!

app/code/core/Mage/Api/Model/Session.php
```diff
@@ -105,8 +127,15 @@ public function clear()
     public function login($username, $apiKey)
     {
         $user = Mage::getModel('api/user')
-            ->setSessid($this->getSessionId())
-            ->login($username, $apiKey);
+            ->setSessid($this->getSessionId());
+        if ($this->getIsInstaLogin() && $user->authenticate($username, $apiKey)) {
+            Mage::dispatchEvent('api_user_authenticated', [
```
fballiano, line 132: do we need to dd this event to the documentation?

kiatng: No need to document, this event is carried from here:
https://github.com/OpenMage/magento-lts/blob/2a2a2fb504247e8966f8ffc2e17d614be5d43128/app/code/core/Mage/Api/Model/User.php#L263-L275

It is dispatched after authentication, added here for consistency.

fballiano:
we can remove a line from the phpstan baseline in order to fix the workflow

fballiano:
it says that we've to remove that line from the baseline, since it's not triggering an error anymore

elidrissidev (approved):

fballiano (approved):

Issue #3270, opened 2023-05-17
Product Attribute Type "Multiselect" Not Updated in EAV and Product_Flat Tables


empiricompany:
### Preconditions (*)

OpenMage 20.1.0-rc5 but also encountered in latest 1.9.4.5 magento version
php 7.4

### Steps to reproduce (*)

1. Create a new product attribute with this config:
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/0c69cdbb-1cbe-4018-8a40-e83e1c7b3b83)

2. Add same options value
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/d29f2c73-2787-4b86-bc67-0c34e45f6170)

3. Assign attribute to an Attribute Set

4. Assign all options to a product
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/72d4c0f8-a90b-4ea8-bf08-80139db21735)

5. Remove one or more options from attribute
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/ec44eec4-0e70-4f49-b3df-c66689807d4a)

### Expected result (*)

1. `catalog_product_entity_text`.value is updated without id of option deleted
2. `catalog_product_flat` is updated without id of option deleted

### Actual result (*)

1. `catalog_product_entity_text`.value is not updated 
new value should be 238,239,240 without orphan 241 id option
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/80d8d380-227c-4809-9075-d606a8217e94)

2. also in `catalog_product_flat` column is not updated
![immagine](https://github.com/OpenMage/magento-lts/assets/5071467/73f7d7ba-3b17-4a41-932c-6ca8dfea4676)

loekvangool:
I've known this bug since it was still called Magento. I agree it's an imperfection but I have to add that all functionality seems to work just fine and as such I would classify this as a cosmetic issue.

kiatng:
I can confirm that for dropdown type such as color, if an option is deleted and there is a product with that color, the value in the table `catalog_product_entity_int` would still have the deleted color.

loekvangool:
> [edit] It's a bug because if you load the product, it would have a color which no longer exist.
AFAIK there is zero impact unless you actually examine the database row. OM silently ignores the deprecated value. Calls to `getData()`, `getAttributeText()` etc. have never lead me to major issues with this.

[...]

PR #3479, opened 2023-08-30
Added a cron job to clean expired API sessions

kiatng:
### Description (*)
The api session expiry is managed in the table `api_session`. However, the old sessions are only deleted during login for that user only, line 268:
https://github.com/OpenMage/magento-lts/blob/2a2a2fb504247e8966f8ffc2e17d614be5d43128/app/code/core/Mage/Api/Model/User.php#L263-L270

If the client does not call `endSession()` and if the client doesn't login again, old sessions will accumulate in the table. In production, I have entries as far back as 2013.

This PR added a daily cron to remove all old sessions.

### Related Pull Requests
PR #3477

Together with #3477, this PR fixed issue #3449

### DIFF
--- a/app/code/core/Mage/Api/Model/Cron.php
+++ b/app/code/core/Mage/Api/Model/Cron.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Api
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2019-2023 The OpenMage Contributors (https://www.openmage.org)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+class Mage_Api_Model_Cron
+{
+    /**
+     * Clean session table
+     *
+     * @param Mage_Cron_Model_Schedule $schedule
+     * @return $this
+     */
+    public function sessionCleanUp($schedule)
+    {
+        Mage::getResourceSingleton('api/user')->cleanOldSessions(null);
+        return $this;
+    }
+}
--- a/app/code/core/Mage/Api/Model/Cron.php
+++ b/app/code/core/Mage/Api/Model/Cron.php
@@ -20,7 +20,7 @@ class Mage_Api_Model_Cron
      * @param Mage_Cron_Model_Schedule $schedule
      * @return $this
      */
-    public function sessionCleanUp($schedule)
+    public function cleanOldSessions($schedule)
     {
         Mage::getResourceSingleton('api/user')->cleanOldSessions(null);
         return $this;
--- a/app/code/core/Mage/Api/Model/Cron.php
+++ b/app/code/core/Mage/Api/Model/Cron.php
@@ -8,8 +8,7 @@
  *
  * @category   Mage
  * @package    Mage_Api
- * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
- * @copyright  Copyright (c) 2019-2023 The OpenMage Contributors (https://www.openmage.org)
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
  * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
  */
 
--- a/app/code/core/Mage/Api/etc/config.xml
+++ b/app/code/core/Mage/Api/etc/config.xml
@@ -17,7 +17,7 @@
 <config>
     <modules>
         <Mage_Api>
-            <version>1.6.0.2</version>
+            <version>1.6.0.3</version>
         </Mage_Api>
     </modules>
     <global>
--- a/app/code/core/Mage/Api/etc/config.xml
+++ b/app/code/core/Mage/Api/etc/config.xml
@@ -118,7 +118,7 @@
                     <cron_expr>0 35 * * *</cron_expr>
                 </schedule>
                 <run>
-                    <model>api/cron::sessionCleanup</model>
+                    <model>api/cron::cleanOldSessions</model>
                 </run>
             </api_session_cleanup>
         </jobs>
--- a/app/code/core/Mage/Api/etc/config.xml
+++ b/app/code/core/Mage/Api/etc/config.xml
@@ -115,7 +115,7 @@
         <jobs>
             <api_session_cleanup>
                 <schedule>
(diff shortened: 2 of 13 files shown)


app/code/core/Mage/Api/etc/config.xml
```diff
@@ -111,6 +111,18 @@
             </modules>
         </translate>
     </adminhtml>
+    <crontab>
+        <jobs>
+            <api_session_cleanup>
+                <schedule>
+                    <cron_expr>0 0 * * *</cron_expr>
```
fballiano, line 118: should this be configurable like some other cron running time?

kiatng: Can be configurable but unnecessary, it just needs to run once a day.  The first run may take longer time, subsequently it should be rather fast.

fballiano: cause at midnight many things start and it could be a tricky time for some websites

kiatng:
Failed CS-Fixer due to misaligned comment. Does it fix automatically or do I need to make another commit to fix it?

app/code/core/Mage/Api/etc/config.xml
kiatng: OK, I can make it configurable, wait for it.

fballiano:
@kiatng it doesn't fix it automatically

app/code/core/Mage/Api/etc/config.xml
colinmollenhour: I don't think it needs to be configurable, I like to just pick a "random" time that is off-peak to avoid always using midnight exactly when there is not a reason to run at a specific time (anything but a time-sensitive report).

```suggestion
                    <cron_expr>0 35 * * *</cron_expr>
```

app/code/core/Mage/Api/Model/Cron.php
```diff
@@ -0,0 +1,29 @@
...
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Api
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2019-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 12:
```suggestion
 * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Api/Model/Cron.php
```diff
@@ -0,0 +1,28 @@
...
+
+class Mage_Api_Model_Cron
+{
+    /**
+     * Clean session table
+     *
+     * @param Mage_Cron_Model_Schedule $schedule
+     * @return $this
+     */
+    public function sessionCleanUp($schedule)
```
fballiano, line 23:
```suggestion
    public function cleanOldSessions($schedule)
```
I'd use the cleanOldSessions which is the same name used in app/code/core/Mage/Api/Model/Resource/User.php

but also, is it ok for this method to have an argument? I never say any cron method to have arguments, I can't even test it directly because obviously `Mage::getModel('api/cron')->sessionCleanUp()` triggers an error

kiatng: You can test like this: `Mage::getModel('api/cron')->cleanOldSessions(null);`

fballiano: but the $schedule parameter is not used in the method itself, why is it there if it's not used?

kiatng: I prefer it because it is called with the param `$schedule`, line 298 and line 320:
https://github.com/OpenMage/magento-lts/blob/0f032f375598b50d1c9c553206a9c750796bf1d6/app/code/core/Mage/Cron/Model/Observer.php#L298-L320

I usually set a message `$schedule->setMessages($msg);` which will be saved in the table.

kiatng: See also https://github.com/OpenMage/magento-lts/blob/0f032f375598b50d1c9c553206a9c750796bf1d6/app/code/core/Mage/Reports/Model/Observer.php#L13-L21

`$schedule` is not used.

kiatng: Actually, it's customary in the core to have the param `$schedule`, see [app\code\core\Mage\Sales\Model\Observer.php](https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Sales/Model/Observer.php)

kiatng:
This PR is actually not very useful. The old sessions in the table do not cause any problem. I am considering closing it. What do you think?

fballiano:
well I think we shouldn't have leftover records forever so I think it's kinda useful :-)

I'll retest it later

kiatng:
I don't like the idea of another cron running everyday just to clean entries that may or may not be there. Especially with insta-login, the table will be empty. Off-topic, I cannot get the insta-login working in SOAP.

fballiano:
well, at the end of the day if there are no records to clean it will run for 0.0 seconds ;-)

elidrissidev:
Maybe running it everyday is an overkill. How about increasing it to 1 week or month?

elidrissidev:
> Off-topic, I cannot get the insta-login working in SOAP.

Did you try using the SOAP V1 way with the `call` method?

colinmollenhour (approved):

fballiano:
I prefer it to stay daily because it will have to process fewer records, almost zero, but if some store has many sessions then it won't happen that the cron will have to process thousands of records

kiatng:
> > Off-topic, I cannot get the insta-login working in SOAP.
> 
> Did you try using the SOAP V1 way with the `call` method?

Yes, I tried SOAP v1. insta-login uses HTTP basic auth, the error I got was related to getting the WSDL. But it's not important for me as JSON-RPC is now in production and is so much faster.

empiricompany:
From my experience with different clients that make extensive use of APIs, the number of records involved is always low, but the primary goal is to ensure a clean database, not just to save memory.

It's just an idea, but we should consider merging it into a single cron job along with other future tasks for cleaning and maintaining the database, in order to avoid occupying a dedicated cron time slot? 
I'm thinking of other actions to clean up obsolete data like this: https://github.com/OpenMage/magento-lts/issues/3270

fballiano:
I still prefer to have small cron jobs because they run quickly and they're easier to debug in case something gets stuck. also, internal cron jobs run only one at a time, so if you have a long running one, you're blocking all of the other from running and that's something I surely prefer to avoid as much as possible

fballiano (approved):
tested, working (the code is actually the same at before so ;-)

kiatng:
Great discussion. On @empiricompany idea of merging crons into one, problem is that modules can be disabled, so we can't merge different jobs in different modules.

PR #3480, opened 2023-08-30
Fixed out-of-range column `lognum` in the table `api_user`

kiatng:
### Description (*)
The column `lognum` was type `smallint`, which has a limit of 65,535.
https://github.com/OpenMage/magento-lts/blob/2a2a2fb504247e8966f8ffc2e17d614be5d43128/app/code/core/Mage/Api/sql/api_setup/install-1.6.0.0.php#L160-L160

Production:
image

### Questions or comments
I'm not sure what's the use of lognum, especially when it hits the limit, it's useless. Nonetheless,  it needs fixing.

### DIFF
--- a/app/code/core/Mage/Api/etc/config.xml
+++ b/app/code/core/Mage/Api/etc/config.xml
@@ -17,7 +17,7 @@
 <config>
     <modules>
         <Mage_Api>
-            <version>1.6.0.2</version>
+            <version>1.6.0.3</version>
         </Mage_Api>
     </modules>
     <global>
--- a/app/code/core/Mage/Api/sql/api_setup/mysql4-upgrade-1.6.0.2-1.6.0.3.php
+++ b/app/code/core/Mage/Api/sql/api_setup/mysql4-upgrade-1.6.0.2-1.6.0.3.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * OpenMage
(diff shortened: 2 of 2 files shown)


fballiano:
wow, more than 65.000 logins?

kiatng:
> wow, more than 65.000 logins?

Yes, not making this up, and probably has been out of range for ages.

ADDISON74:
1 login/hour x 24h/day x 30d/month x 12m/year = 8640. 

In 10 years, if he does this, he can exceed a smallint. I reported to one hour because that's how long a session could last in general. If he logs out after 10 minutes and comes back in, then the number can increase. The observation is good because there can be several users on a single account.

fballiano:
dunno, I think it's a bit overkill, I think that the original idea is to check if a user does too many logins in order to take actions, in that case a smallint is fine.

anyway, not a big deal in any cases.

fballiano (approved):

fballiano:
I hoped to get some other opinion but ok, let's merge

m-overlund:
our api users (for Klaviyo and WMS) has maxed out after only a year or so.

ADDISON74:
> our api users (for Klaviyo and WMS) has maxed out after only a year or so

In this case the PR will be useful for them and others in this situation.

Issue #3472, opened 2023-08-28
Uncaught TypeError: Mage_Sales_Model_Quote::setCouponCode()


S0FTWEX:
### Preconditions (*)

1. OpenMage 20.1.0
2. PHP 8.2

### Steps to reproduce (*)

1. Place an order with no coupon code.
2. Run this PHP script to create quote from the order:
```
$orderId = XXXXX;
$order = Mage::getModel('sales/order')->load($orderId);
$quote = Mage::getSingleton('sales/convert_order')->toQuote($order);
```

### Expected result (*)

1. Quote has to be created.

### Actual result (*)

```
Fatal error: Uncaught TypeError: Mage_Sales_Model_Quote::setCouponCode(): Argument #1 ($couponCode) must be of type string, null given, called in /lib/Varien/Object.php on line 394 and defined in /app/code/core/Mage/Sales/Model/Quote.php on line 2102
TypeError: Mage_Sales_Model_Quote::setCouponCode(): Argument #1 ($couponCode) must be of type string, null given, called in /lib/Varien/Object.php on line 394 in /app/code/core/Mage/Sales/Model/Quote.php on line 2102
```

fballiano:
mmm interesting, the setCoupon method is:

```php 
    public function setCouponCode(string $couponCode)
    {
        return $this->setData('coupon_code', $couponCode);
    }
```

so it requires a string, but the `toQuote()` method is copying filling the quote data using `copyFieldset()`

```php
public function copyFieldset($fieldset, $aspect, $source, $target, $root = 'global')
    {
        if (!(is_array($source) || $source instanceof Varien_Object)
            || !(is_array($target) || $target instanceof Varien_Object)
        ) {
            return false;
        }
        $fields = Mage::getConfig()->getFieldset($fieldset, $root);
        if (!$fields) {
            return false;
        }

        $sourceIsArray = is_array($source);
        $targetIsArray = is_array($target);

        $result = false;
        foreach ($fields as $code => $node) {
            if (empty($node->$aspect)) {
                continue;
            }

            if ($sourceIsArray) {
                $value = $source[$code] ?? null;
            } else {
                $value = $source->getDataUsingMethod($code);
            }

            $targetCode = (string)$node->$aspect;
            $targetCode = $targetCode == '*' ? $code : $targetCode;

            if ($targetIsArray) {
                $target[$targetCode] = $value;
            } else {
                $target->setDataUsingMethod($targetCode, $value);
            }

[...]

PR #3492, opened 2023-09-04
Fix a Type Error when converting an Order with no Coupon to a Quote

elidrissidev:
### Description (*)

A small for a Type Error that occurs when converting an Order to a Quote, the `coupon_code` setter should accept `null` since the database field is nullable.

### Related Pull Requests

- See OpenMage/magento-lts#2825

1. Fixes OpenMage/magento-lts#3472

```php
$orderId = XXXXX;
$order = Mage::getModel('sales/order')->load($orderId);
$quote = Mage::getSingleton('sales/convert_order')->toQuote($order);
```

### DIFF
--- a/app/code/core/Mage/Sales/Model/Quote.php
+++ b/app/code/core/Mage/Sales/Model/Quote.php
@@ -2096,10 +2096,10 @@ class Mage_Sales_Model_Quote extends Mage_Core_Model_Abstract
     }
 
     /**
-     * @param string $couponCode
+     * @param string|null $couponCode
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

sreichel:
> I would have removed the method, I don't see benefit in having it, but let's get this fixed however, which is far more important

Real methods are much more faster.  (see #3245) From my POV - like in M2- every DB field should have its getter/setter methods. Already started it, but thats a lot of work. Type casting like `(string) $xyz` could or should be avoided. _imho_

Issue #3496, opened 2023-09-06
Incompatible interface deprecations in Mage_Core_Model_Resource_Session


TangLiang:
``````````````````````

Deprecated functionality: Return type of Mage_Core_Model_Resource_Session::open($savePath, $sessName) should either be compatible with SessionHandlerInterface::open(string $path, string $name): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice  in /home/wwwroot/lnmp01/domain/magentorc_04/web/app/code/core/Mage/Core/Model/Resource/Session.php on line 165
Deprecated functionality: Return type of Mage_Core_Model_Resource_Session::close() should either be compatible with SessionHandlerInterface::close(): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice  in /home/wwwroot/lnmp01/domain/magentorc_04/web/app/code/core/Mage/Core/Model/Resource/Session.php on line 175
Deprecated functionality: Return type of Mage_Core_Model_Resource_Session::read($sessId) should either be compatible with SessionHandlerInterface::read(string $id): string|false, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice  in /home/wwwroot/lnmp01/domain/magentorc_04/web/app/code/core/Mage/Core/Model/Resource/Session.php on line 188
 Deprecated functionality: Return type of Mage_Core_Model_Resource_Session::write($sessId, $sessData) should either be compatible with SessionHandlerInterface::write(string $id, string $data): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice  in /home/wwwroot/lnmp01/domain/magentorc_04/web/app/code/core/Mage/Core/Model/Resource/Session.php on line 211
Deprecated functionality: Return type of Mage_Core_Model_Resource_Session::destroy($sessId) should either be compatible with SessionHandlerInterface::destroy(string $id): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice  in /home/wwwroot/lnmp01/domain/magentorc_04/web/app/code/core/Mage/Core/Model/Resource/Session.php on line 244
[...]

PR #3497, opened 2023-09-06
Fixed incompatible interface deprecations in DB Session Adapter

elidrissidev:
### Description (*)

Small fix to make the methods in DB session adapter compatible with `SessionHandlerInterface`.

1. Fixes OpenMage/magento-lts#3496

1. Switch `<session_save>` in `local.xml` to `db`.
2. Clean the cache and check `system.log`.

### DIFF
--- a/app/code/core/Mage/Core/Model/Resource/Session.php
+++ b/app/code/core/Mage/Core/Model/Resource/Session.php
@@ -162,6 +162,7 @@ class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Inter
      * @param string $sessName ignored
      * @return bool
      */
+    #[\ReturnTypeWillChange]
     public function open($savePath, $sessName)
     {
         return true;
@@ -172,6 +173,7 @@ class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Inter
      *
      * @return bool
      */
+    #[\ReturnTypeWillChange]
     public function close()
     {
         $this->gc($this->getLifeTime());
@@ -185,6 +187,7 @@ class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Inter
      * @param string $sessId
      * @return string
      */
+    #[\ReturnTypeWillChange]
     public function read($sessId)
     {
         $select = $this->_read->select()
@@ -208,6 +211,7 @@ class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Inter
      * @param string $sessData
      * @return bool
      */
+    #[\ReturnTypeWillChange]
     public function write($sessId, $sessData)
     {
         $bindValues = [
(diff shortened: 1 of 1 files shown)


elidrissidev:
I tried adding the type hints following [`SessionHandlerInterface`](https://www.php.net/manual/en/class.sessionhandlerinterface.php) signature and only adding `ReturnTypeWillChange` to the methods with union types which aren't supported pre-PHP 8, but PHPStan keeps complaining 🤷‍♂️

colinmollenhour:
I didn't see your PR and made an attempt to ignore these errors here: #3498 but it seems yours actually works. 😄 

What do you think about changing:
```
class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Interface
```
to
```
class Mage_Core_Model_Resource_Session implements SessionHandlerInterface
```
?

elidrissidev:
@colinmollenhour I've switched it. `Zend_Session_SaveHandler_Interface` wasn't referenced anywhere anyways.

fballiano:
if we change the interface of a public class, shouldn't we have this PR in the next branch?

elidrissidev:
The interface is unused in the core and the only reference I found was commented out 15 years ago.

fballiano:
😂 ok then, it seems enough

some of our code is really old

fballiano:
on my local machine phpstan runs without errors also without the `ReturnTypeWillChange` now that we're using the `SessionInterface`, should we still keep the `ReturnTypeWillChange`?

elidrissidev:
The issue this PR was meant to be fixing is PHP deprecation errors related to incompatible types with the interface. PHPStan never complained about them in `main`, but I think it should fix it in `next`.

fballiano:
sure, that's why I'm running phpstan on my local machine with php8, which is the difference that triggers the error in next ;-)

(actually i've to admin I ran it with php8.2 instead of 8.1)

elidrissidev:
In `next` proper types could be added to this class. Here it's either you remove `ReturnTypeWillChange` which causes PHP deprecation messages, or you try to introduce types and but PHPStan's expectations do not conform with what's in PHP documentation, it seems to expect the methods to have `mixed` type which is not available in PHP 7.x.

fballiano:
true, [this version](https://github.com/fballiano/openmage/blob/testphpstan/app/code/core/Mage/Core/Model/Resource/Session.php) works correctly but it would have to be committed only on the `next` branch

colinmollenhour:
Does this patch also fix the next branch? My only goal was to get the PHPStan action passing again - if this does that then I am for it. :)
What is the difference between main and next that causes next to fail in the first place?

colinmollenhour (approved):
Looks good to me. The change of which interface class is implemented should be benign.

fballiano:
I would have preferred the proper fix on the next branch :-\

elidrissidev:
@fballiano Feel free to open one for `next` to introducing proper types :)

fballiano:
done, but I think this one should be merged too, right?

elidrissidev:
Yeah. The PR for `next` should wait until this one is merged and forward-ported so you can rebase it and remove the `ReturnTypeWillChange` since it won't be needed there.

fballiano (approved):

fballiano:
then I guess it's time ;-)

fballiano:
merged, forward ported and updated https://github.com/OpenMage/magento-lts/pull/3499 ;-)

PR #3499, opened 2023-09-07
Fixed implementation of SessionHandlerInterface in Mage_Core_Model_Resource_Session

fballiano:
Alternative (but actually both can be merged) version to https://github.com/OpenMage/magento-lts/pull/3497 with correct (I hope) typing, it can be merged only in `next` because it uses mixed typing which is not available in PHP7.

Should fix PHPStan errors on the `next` branch.

### DIFF
--- a/app/code/core/Mage/Core/Model/Resource/Session.php
+++ b/app/code/core/Mage/Core/Model/Resource/Session.php
@@ -264,6 +264,6 @@ class Mage_Core_Model_Resource_Session implements SessionHandlerInterface
                 return $this->_write->delete($this->_sessionTable, $where);
             }
         }
-        return false;
+        return 0;
     }
 }
--- a/app/code/core/Mage/Core/Model/Resource/Session.php
+++ b/app/code/core/Mage/Core/Model/Resource/Session.php
@@ -19,7 +19,7 @@
  * @category   Mage
  * @package    Mage_Core
  */
-class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Interface
+class Mage_Core_Model_Resource_Session implements SessionHandlerInterface
 {
     /**
      * Session maximum cookie lifetime
@@ -162,7 +162,7 @@ class Mage_Core_Model_Resource_Session implements Zend_Session_SaveHandler_Inter
      * @param string $sessName ignored
      * @return bool
      */
-    public function open($savePath, $sessName)
+    public function open($savePath, $sessName): bool
     {
         return true;
     }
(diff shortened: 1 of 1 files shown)


elidrissidev (approved):

app/code/core/Mage/Core/Model/Resource/Session.php
```diff
@@ -185,10 +183,9 @@ public function close()
      * Fetch session data
      *
      * @param string $sessId
-     * @return string
+     * @return string|false
      */
-    #[\ReturnTypeWillChange]
-    public function read($sessId)
+    public function read($sessId): string|false
```
kiatng, line 188: line 201:
```php
         return (string)$data;
```
to reflect the return type, it should be

```php
        return $data;
```

After the change, you can test with
```php
$value = Mage::getResourceSingleton('core/session')->read('abc'); // $value is false
```

app/code/core/Mage/Core/Model/Resource/Session.php
```diff
@@ -257,19 +252,18 @@ public function destroy($sessId)
...
             if ($this->_automaticCleaningFactor == 1 ||
                 rand(1, $this->_automaticCleaningFactor) == 1
             ) {
                 $where = ['session_expires < ?' => Varien_Date::toTimestamp(true)];
-                $this->_write->delete($this->_sessionTable, $where);
+                return $this->_write->delete($this->_sessionTable, $where);
             }
         }
-        return true;
+        return false;
```
kiatng, line 267: Not sure about this. wouldn't this mean `gc()` has failed? See example in https://www.php.net/manual/en/class.sessionhandlerinterface.php
image

elidrissidev: The docs for `gc()` says:
> Returns the number of deleted sessions on success, or false on failure. Note this value is returned internally to PHP for processing.

So it should return 0 instead of `false`.

fballiano, line 267:
```suggestion
        return 0;
```

fballiano: done, thanks!

app/code/core/Mage/Core/Model/Resource/Session.php
fballiano: done, although the db->fetchOne() is marked to return a string...

fballiano:
sorry for the mess with the commits, it should be ok now.

phpcs breaks because of https://github.com/OpenMage/magento-lts/pull/3500 totally unrelated to this PR.

ADDISON74 (approved):

colinmollenhour (approved):

PR #3514, opened 2023-09-12
Show cache status info of blocks in template path hints

empiricompany:
### Description (*)
This is a small trick I usually use to debug cache blocks. I consider it useful to quickly find cache-related issues without installing other extensions.

1. Enable Template Path Hints
2. Frontend must show:
 - red for no cached block
 - orange for blocks nested in a cached block
 - green for cached blocks

demo

Inspiration from:
https://github.com/AOEpeople/Aoe_TemplateHints

### DIFF
--- a/app/code/core/Mage/Core/Block/Template.php
+++ b/app/code/core/Mage/Core/Block/Template.php
@@ -215,13 +215,13 @@ class Mage_Core_Block_Template extends Mage_Core_Block_Abstract
         if (!is_null($this->getCacheLifetime())) {
             return 'green';
         } else {
-            $_currentParentBlock = $this;
+            $currentParentBlock = $this;
             $i = 0;
-            while ($i++ < 20 && $_currentParentBlock instanceof Mage_Core_Block_Abstract) {
-                if (!is_null($_currentParentBlock->getCacheLifetime())) {
+            while ($i++ < 20 && $currentParentBlock instanceof Mage_Core_Block_Abstract) {
+                if (!is_null($currentParentBlock->getCacheLifetime())) {
                     return 'orange'; // not cached, but within cached
                 }
-                $_currentParentBlock = $_currentParentBlock->getParentBlock();
+                $currentParentBlock = $currentParentBlock->getParentBlock();
             }
         }
         return 'red';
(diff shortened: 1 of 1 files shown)


sreichel:
Maybe OM can reuse AoE code and add it to core?

That's an extension I used for every shop.

pquerner:
> Maybe OM can reuse AoE code and add it to core?
> 
> That's an extension I used for every shop.

I've opened a discussion here https://github.com/OpenMage/magento-lts/discussions/3485 about how that should/can be handled.

empiricompany:
This extension does not aim to reproduce all the functionality of aoe_templatehints because it would be too complex and invasive. It only adds basic information about the cache without exposing all the details provided by the aoe extension, such as cache key info, class methods, etc.

For me, it is enough as basic functionality without installing a separate extension because I can inspect these details in the code. Actually, I don't have time to integrate all the aoe features. 

Eventually I can quickly add this informations in the title attribute of the block name hover without loading an additional tooltip library used in aoe, dependent on prototype.

demo

fballiano:
do we have to use the "_" prefixed variables everywhere? on the inside of a function/method I really don't enjoy reading the code when it's full of those.

ADDISON74:
**OFF TOPIC:** Adding to the OM core is an important decision. If it is not something complex, it can be done as this PR is, but in the case of an extension/module, it would be preferable to be added with Composer. There are several extensions that have become "must-haves" over the years, but almost all of them are no longer maintained. If we want to be integrated them, forks can be made in this repository and we keep them separate from the OM core. The best example is CM_Redis which was separated from OM.

app/code/core/Mage/Core/Block/Template.php
```diff
@@ -205,6 +205,28 @@ public function getShowTemplateHints()
...
+     */
+    private function _getCacheHintStatus(): string
+    {
+        if (!is_null($this->getCacheLifetime())) {
+            return 'green';
+        } else {
+            $_currentParentBlock = $this;
+            $i = 0;
+            while ($i++ < 20 && $_currentParentBlock instanceof Mage_Core_Block_Abstract) {
+                if (!is_null($_currentParentBlock->getCacheLifetime())) {
```
ADDISON74, line 221: How did you set the condition to be limited to 20? Is there a possibility that this value will ever be exceeded?

empiricompany:
> do we have to use the "_" prefixed variables everywhere? on the inside of a function/method I really don't enjoy reading the code when it's full of those.

No, for me it's indifferent. To tell the truth, should only be used for private methods and variables.

app/code/core/Mage/Core/Block/Template.php
empiricompany: Should it make a maximum of 20 clicks, right?

ADDISON74: Of course, but is this threshold more or less? I was interested in how you set it to the value of 20.

empiricompany: I copied the value and function from AOE, it reaches a maximum of 20 nested blocks

fballiano (approved):

ADDISON74:
It can go to the main branch as well.

I use this extension in production when needed to quickly find out the cause of problems related to templates.
https://www.magepsycho.com/easy-template-path-hints.html

fballiano:
as usual it can't be rebased easily, @empiricompany if you can't rebase it I'll backport it to main manually when it gets approved

kiatng (approved):

fballiano:
I've also backported this commit to `main`.

sreichel:
> e have to use the "_" prefixed variables everywhere? on the inside of a function/method I really don't enjoy reading the code when it's full of those.

:+1:

PR #3518, opened 2023-09-15
Added sharpening filter to Varien_Image_Adapter_Gd2 to fix blurry thumbnails

fballiano:
With GD2, we've had the "blurry thumbnails" problems for many years now, some have created 3rd party modules based on image-magick but today I was wondering around the internet and found [this stackoverflow page](https://stackoverflow.com/questions/2799159/php-gd-image-resizing-to-smaller-new-image-blurring-problem) and thought to give it a try and it seems to work really nicely to me!

before:
Screenshot 2023-09-15 alle 11 21 21

after:
Screenshot 2023-09-15 alle 11 20 23

It seems a really good improvement so I thought to create a PR for it.

UPDATE 1: tested on a customer's website, didn't like the result, on desktop/mobile-preview looks ok but on real mobile device looked a bit jaggy

### DIFF
--- a/lib/Varien/Image/Adapter/Gd2.php
+++ b/lib/Varien/Image/Adapter/Gd2.php
@@ -409,6 +409,7 @@ class Varien_Image_Adapter_Gd2 extends Varien_Image_Adapter_Abstract
         );
 
         // sharpening filter to avoid blurry thumbnails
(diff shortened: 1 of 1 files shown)


f1-outsourcing:
https://github.com/OpenMage/magento-lts/discussions/3921#discussioncomment-12001143

Issue #3528, opened 2023-09-19
Refresh updated_at date on mass product updates actions


Tomasz-Silpion:
### Description (*)

I would like to track last changes on products even if they were updated with mass_action.

### Expected behavior (*)

Mass status/attributes update from admin area should change products updated_at dates in `catalog_product_entity` table in same manner edit tab save does.

### Benefits

OpenMage users would be able to track changes to products more easily.

### Additional information

1. Go to admin panel, enter any product, change status and save. Query `catalog_product_entity` table - product updated_at value was changed.
2. Go to admin panel, use mass status/attributes update then query `catalog_product_entity` table - product updated_at values were **not** changed.

![catalog_product_mass_action](https://github.com/OpenMage/magento-lts/assets/5328659/b2ad2376-97ab-4557-a978-b13ae2374766)

![catalog_product_entity_table](https://github.com/OpenMage/magento-lts/assets/5328659/5b3b39f1-6b76-4333-9c43-a659fbe7d77e)

PR #3529, opened 2023-09-19
Fixed updated_at doesn't update on mass actions

Tomasz-Silpion:
### Description (*)

Currently mass products action don't leave any trace in `catalog_product_entity` table making it unclear when real changes were applied. Changing product status from edit form updates `updated_at` column with save date while mass actions do not.

1. Fixes OpenMage/magento-lts#3528

1. Go to admin panel, enter any product, change status and save. Query into `catalog_product_entity table` with ```SELECT * FROM `catalog_product_entity` ORDER BY `updated_at` DESC;```  see that product `updated_at` value was refreshed.
2. Go to admin panel, use mass status/attributes update then query `catalog_product_entity table`.  Product `updated_at` values were not updated.

Compare this behavior before and after PR.

### Questions or comments

Feel free to edit. I've made my proposal in `Mage_Catalog_Model_Resource_Product_Action` class

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
@@ -85,7 +85,7 @@ class Mage_Catalog_Model_Resource_Product_Action extends Mage_Catalog_Model_Reso
     }
 
     /**
-     * Updated the "updated_at" field for all entity_ids passed
+     * Update the "updated_at" field for all entity_ids passed
      *
      * @param array $entityIds
      * @return void
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
@@ -74,6 +74,10 @@ class Mage_Catalog_Model_Resource_Product_Action extends Mage_Catalog_Model_Reso
                 $this->_processAttributeValues();
             }
             $this->_getWriteAdapter()->commit();
+
+            $this->getWriteConnection()->update($this->getTable('catalog/product'), [
+                'updated_at' => Varien_Date::now()
+            ], $this->getWriteConnection()->quoteInto('entity_id IN (?)', $entityIds));
         } catch (Exception $e) {
             $this->_getWriteAdapter()->rollBack();
             throw $e;
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
@@ -73,16 +73,34 @@ class Mage_Catalog_Model_Resource_Product_Action extends Mage_Catalog_Model_Reso
                 }
                 $this->_processAttributeValues();
             }
-            $this->_getWriteAdapter()->commit();
 
-            $this->getWriteConnection()->update($this->getTable('catalog/product'), [
-                'updated_at' => Varien_Date::now()
-            ], $this->getWriteConnection()->quoteInto('entity_id IN (?)', $entityIds));
+            $this->_updateUpdatedAt($entityIds);
+            $this->_getWriteAdapter()->commit();
         } catch (Exception $e) {
             $this->_getWriteAdapter()->rollBack();
             throw $e;
         }
 
         return $this;
+    }
+
+    /**
+     * Updated the "updated_at" field for all entity_ids passed to the method
+     * @param array $entityIds
+     * @return void
+     * @throws Zend_Db_Adapter_Exception
+     */
+    protected function _updateUpdatedAt(array $entityIds): void
+    {
+        $updatedAt = Varien_Date::now();
+        $catalogProductTable = $this->getTable('catalog/product');
+        $writeConnection = $this->getWriteConnection();
+
+        $entityIdsChunks = array_chunk($entityIds, 1000);
+        foreach ($entityIdsChunks as $entityIdsChunk) {
+            $writeConnection->update($catalogProductTable, [
+                'updated_at' => $updatedAt
+            ], $writeConnection->quoteInto('entity_id IN (?)', $entityIdsChunk));
+        }
     }
 }
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
@@ -85,7 +85,8 @@ class Mage_Catalog_Model_Resource_Product_Action extends Mage_Catalog_Model_Reso
     }
 
     /**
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
```diff
@@ -74,6 +74,10 @@ public function updateAttributes($entityIds, $attrData, $storeId)
                 $this->_processAttributeValues();
             }
             $this->_getWriteAdapter()->commit();
+
+            $this->getWriteConnection()->update($this->getTable('catalog/product'), [
+                'updated_at' => Varien_Date::now()
+            ], $this->getWriteConnection()->quoteInto('entity_id IN (?)', $entityIds));
```
fballiano, line 80: is it possible that the IN clause becomes too long (in case of many many products) and breaks?

fballiano: I think we should split the query in blocks of 1000 records like shown a few lines above :-)

fballiano:
I've rewritten the PR for:
- running before the commit of the transaction
- running in chunks instead of processing all entity_ids in a single query

fballiano:
note: this will make mass updates slower, but I agree with the sentiment of the PR.

app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
```diff
@@ -81,4 +83,24 @@ public function updateAttributes($entityIds, $attrData, $storeId)
...
+    {
+        $updatedAt = Varien_Date::now();
+        $catalogProductTable = $this->getTable('catalog/product');
+        $writeConnection = $this->getWriteConnection();
+
+        $entityIdsChunks = array_chunk($entityIds, 1000);
+        foreach ($entityIdsChunks as $entityIdsChunk) {
+            $writeConnection->update($catalogProductTable, [
+                'updated_at' => $updatedAt
+            ], $writeConnection->quoteInto('entity_id IN (?)', $entityIdsChunk));
```
kiatng, line 103: Style changes:
```suggestion
     * Update the "updated_at" field.
     *
     * @param array $entityIds
     * @return void
     * @throws Zend_Db_Adapter_Exception
     */
    protected function _updateUpdatedAt(array $entityIds): void
    {
        $updatedAt = Varien_Date::now();
        $catalogProductTable = $this->getTable('catalog/product');
        $adapter = $this->_getWriteAdapter();

        $chunks = array_chunk($entityIds, 1000);
        foreach ($chunks as $chunk) {
            $adapter->update(
                $catalogProductTable, 
                ['updated_at' => $updatedAt],
                $adapter->quoteInto('entity_id IN (?)', $chunk)
            );
```

fballiano: I kinda met you halfway 😅 what do you think about the new version?

fballiano (approved):

app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
```diff
@@ -81,4 +83,25 @@ public function updateAttributes($entityIds, $attrData, $storeId)
 
         return $this;
     }
+
+    /**
+     * Updated the "updated_at" field for all entity_ids passed
```
kiatng, line 88:
```suggestion
     * Update the "updated_at" field for all entity_ids passed
```

fballiano (approved):

app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
```diff
@@ -73,6 +73,8 @@ public function updateAttributes($entityIds, $attrData, $storeId)
                 }
                 $this->_processAttributeValues();
             }
+
+            $this->_updateUpdatedAt($entityIds);
```
pquerner, line 77: Should this happen after the `commit`?

fballiano: why? I think it should be committed together with the data, otherwise they could be out of sync

pquerner: Because the commit might have more observers - and might entirely cut the process alltogether. Why update update_at before we're really done?

fballiano: if the transaction fails, the commit fails and the updated_at doesn't get updated

pquerner: And the first part?
`Because the commit might have more observers`

fballiano: if the commit has an obsever that cuts the process, and the updatedAt is in the same transaction, then it work correctly (if i'm not missing anything). if the updatedAt is outside of the commit it will be executed also if the transaction fails, which is wrong.

pquerner: Forget about `cut the process`. `anything` is done in the `commit`. Maybe alter the data, maybe add more data, whatever. Shouldnt the `updated_at` be done `after` that is done, instead of before?
I am just proposing moving the lines, not stepping out of the try/catch.

As I read that, the code in `$this->_updateUpdatedAt` doesnt care about the `commit` or am I wrong?

fballiano: I see, but then we should check if something was committed or otherwise not update the updatedAt.

btw it seems quite wrong for the commit observer to add/remove more data than the one in process of being commited 😅

anyway let's see what other people say :-)

pquerner: Moving the lines shouldnt hurt and because the commit went through successfuly, we know for sure that we can update the updated_at field.
As per `but then we should check if something was committed or otherwise not update the updatedAt.` is that possible without knowing previous data vs. data after the commit? That sounds quite heavy.

fballiano: 1. if the commit observer removes some data -> the updatedAt would be wrong
2. if the commit observer adds some data -> the updatedAt would be wrong
3. if the commit observer removed everything -> we would still update all the updatedAt

if it is so important to move the updatedAt outside of the commit, I prefer not to have it all like it is now in production.

kiatng (approved):
Tested with a small set of data.

app/code/core/Mage/Catalog/Model/Resource/Product/Action.php
kiatng: Unless there is **actual test case** that shows the update should be after the commit, I have to agree with @fballiano

PR #3530, opened 2023-09-20
Improved stampede prevention with empty config cache under high loads

colinmollenhour:
### Description (*)

In a situation where generating the config takes a long time and the store is receiving heavy traffic - although there is already a basic stampede-prevention mechanism in place, it is not completely effective. There are three aggravating factors that this PR aims to fix:

1. When using read replicas, sending the reads to the replicas to check if the config saving is locked makes it susceptible to race conditions caused by replication lag. Reads in this instance should always go to the master.
2. The locking mechanism does a read first, then a write, then assumes it is safe to proceed, but if many processes do the read at the same time before the first one does a write, there can be multiple processes that "have" the lock. Now, a second read will ensure that only one process has the lock, the others will skip saving the config cache which reduces load on the backend and frontends (which have to serialize the XML which can be quite large).
3. There is a 5 second "hold" time that will cause processes to wait for the cache to be warmed. I've increased this to 10 seconds since if it takes significantly longer than 5 seconds to save the cache in a high load environment that can result in a large number of other processes generating the cache when it's faster to just wait for it to be saved then read the cache.

One idea I did not implement is to just respond with a 503 error when the hold time is exceeded to avoid having a large number of processes trying to generate a fresh cache all at once. But this would be a more drastic change so I didn't implement that for now.

I tested point 2 using 20 concurrent requests and some logging lines sprinkled in the code:
image

This proves that the mechanism caused only one process to save the cache where four processes would have previously.

### Questions or comments

If anyone has a production environment that experiences cache stampedes, I'd be grateful if you gave it a try and report your findings.

**NOTE**: For the fix for point 1 to work, the Redis cache library needs to be updated with this as yet unmerged update. I plan to update it soon and it is fully BC and this patch will still work without it, just without having any effect on the issue in point 1.
https://github.com/colinmollenhour/Cm_Cache_Backend_Redis/pull/177


fballiano:
what do you think about https://github.com/fballiano/magento-cache-regeneration-lock?

colinmollenhour:
> what do you think about https://github.com/fballiano/magento-cache-regeneration-lock?

@fballiano Ahh, very nice! I thought about using `GET_LOCK()` shortly after pushing this, it is a good strategy.

Pros:
- Less load on the cache backend
- Fully atomic so there are no race conditions
- Simple and fast

Cons:
- Breaks the database abstraction - we only support MySQL variants anyway so this probably shouldn't be a deal-breaker - it's also used already by `Mage_Index` so is not exactly new although there is at least the appearance of abstraction there.
- Will not work perfectly on MySQL multi-master setups unless all writes go the same node. I don't know of anyone using multiple write nodes successfully anyway - do you? If someone was using a multi-master cluster it would probably still work ok, just there would be as many processes writing config cache as there are db masters which is at least guaranteed to be a small number.

Notes:
- Your patch implements the locking in a different location so I think it needs to also be added to `Mage_Core_Model_Config::init`  to cover all methods of initialization - in a similar manner or some refactoring to prevent code duplication.
- Minor improvement: I recommend using `try { ... } finally { ... }` to release locks so that unhandled throwables cannot cause the lock release to be skipped.
- The 504 error is not exactly accurate - 500 or 503 would be more accurate - since this page could be visible to end users it might warrant using `require_once Mage::getBaseDir() . DS . 'errors' . DS . '503.php'; die();` instead - consistent with when session initialization fails.

fballiano:
@colinmollenhour at the time of developing that patch (around 2015 or a bit before, I was using it for many customers before releasing it) I tried many different ways but the get_lock seems the easiest and still most effective way :-)

cons:
- exactly, it's already used in mage_index, but also, IMHO, we'll never support anything different from mysql (and derivatives)
- noup, I don't even know somebody really using slaves DB but no multimaster

pros:
- it doesn't depend on redis which not everybody may use (probably they don't need the lock because they'll have small stores)
- since it's managed by mysql itself, it should be pretty reliable

when I developed it I tried to keep it as simple as possible to make it work across multiple versions of magento1, but if we try to implement it in the core we can make it a bit better (like with the try/catch/finally)

do you think I should make a PR out of it? it is possible that the get_lock could create problems to some systems? could it be "disabled" on some databases?

colinmollenhour:
@fballiano I updated to use GET_LOCK - effectively the same thing your patch does - and removed the old cache-based mechanism and reorganized a bit.

It uses a 3 second timeout for http and a 60 second timeout for CLI - reasoning that for http you don't want to risk overloading your http server in a high load scenario and it's better to basically act like maintenance mode. For CLI there shouldn't be hundreds or thousands of processes so it is ok to wait longer.

Here is a screenshot showing what happens when I added some `Mage::log` calls during a flood of requests with a cold cache:
image

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,43 @@ public function getCache()
...
      */
-    protected function _getCacheLockId()
+    public function getCacheSaveLock()
     {
-        return $this->getCacheId() . '.lock';
+        if (!Mage::app()->useCache('config')) {
+            return;
+        }
+        $connection = Mage::getSingleton('core/resource')->getConnection('core_write');
+        if (!$connection->fetchOne("SELECT GET_LOCK('core_config_cache_save_lock', ?)", [PHP_SAPI === 'cli' ? 60 : 3])) {
```
colinmollenhour, line 509:
```suggestion
        $lockWaitTime = getenv('MAGE_CONFIG_CACHE_LOCK_WAIT') ?: (PHP_SAPI === 'cli' ? 60 : 3);
        if (!$connection->fetchOne("SELECT GET_LOCK('core_config_cache_save_lock', ?)", [$lockWaitTime])) {
```

app/code/core/Mage/Core/Model/App.php
```diff
@@ -432,14 +432,21 @@ protected function _initCache(array $cacheInitOptions = [])
...
-            if ($this->_config->isLocalConfigLoaded() && !$this->_shouldSkipProcessModulesUpdates()) {
-                Varien_Profiler::start('mage::app::init::apply_db_schema_updates');
-                Mage_Core_Model_Resource_Setup::applyAllUpdates();
-                Varien_Profiler::stop('mage::app::init::apply_db_schema_updates');
+            try {
+                $this->_config->getCacheSaveLock();
+                if (!$this->_config->loadModulesCache()) {
+                    $this->_config->loadModules();
+                    if ($this->_config->isLocalConfigLoaded() && !$this->_shouldSkipProcessModulesUpdates()) {
+                        Varien_Profiler::start('mage::app::init::apply_db_schema_updates');
```
fballiano, line 440: do we need to have the profiler?

fballiano: oops sorry it was there before, sorry!

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: This let's one override the default cache lock wait time with an environment variable, what do you think? Could also be moved to an XML config variable or just left hard-coded.
Regarding the default values I think sites with low traffic could afford to increase the wait time and sites with higher traffic need to keep it low. E.g. if you have 1000r/s then waiting 3 seconds will pile up 3000 requests whereas if you have 10r/s then only 30 requests is not a big deal.

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,43 @@ public function getCache()
...
+            if (PHP_SAPI === 'cli') {
+                throw new Exception('Could not get lock on cache save operation.');
+            } else {
+                require_once Mage::getBaseDir() . DS . 'errors' . DS . '503.php';
```
fballiano, line 513:
```suggestion
                require Mage::getBaseDir() . DS . 'errors' . DS . '503.php';
```

it will never happen more than once but I think this shouldn't be a require_once

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,44 @@ public function getCache()
...
+    /**
+     * Release the cache saving lock after it is saved or no longer needed
+     *
+     * @return void
+     */
+    public function releaseCacheSaveLock()
+    {
+        if (!Mage::app()->useCache('config')) {
+            return;
+        }
```
fballiano, line 529:
```suggestion
```

weird suggestion, I wouldn't release the lock only if cache is enabled, just in case between the "get lock" and the "release lock" something disables the cache and suddenly the lock is never released

does it make sense?

fballiano:
apart from my last comments it looks fine to me. I ran basic tests and found no problem

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: Is there a PSR or something that recommends using require instead of require_once? I don't care either way, just curious.:)

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: There is no way for the `useCache` return value to change between the calls to `getCacheSaveLock` and `releaseCacheSaveLock` in the same process, so this is just added to avoid an unnecessary `SELECT RELEASE_LOCK(...` query rather than wrapping all of the calls to it with `if (..useCache..)` or using more complex if/else structures in the calling methods. Trying to favor readability but I guess that's subjective. The overhead of this method is extremely low after the first call, basically jsut checking local variables and an array by index.

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: The `local.xml` file is already read in at this point but not yet merged with db config so it could go there, but then it wouldn't be dynamic.

The 3 seconds was picked trying to optimize between preventing the server(s) from getting so overloaded they start failing in other ways and how long it should reasonable take to save the cache. I think in most cases 3 seconds is enough time to save the cache so no requests would get dropped, but if the timeout were say 60, then if it did take over 3 seconds to save, that could allow a *lot* of requests to pile up. Also, in ecommerce a TTFB over 3 seconds is already atrocious so not sure how long it is worth waiting after that?

Ideally the system is configured to just drop connections after it becomes overloaded before they hit PHP anyway, but with AWS auto-scaling it could cause your container count to jump way up unnecessarily since this should just be a very temporary condition. 

Those are my thoughts. Happy to concede if you have another suggestion. 🙂

colinmollenhour: Perhaps a good timeout would be the amount of time a normal config cache save *should* take on a healthy system - so if healthy then no requests will fail but if unhealthy they will start to abort to avoid a more catastrophic failure. Any idea what such a number would be? Maybe 6 seconds? If it takes more than that I would think either your systems are really slow or your config is holding way too much stuff and should be lightened up. :)

app/code/core/Mage/Core/Model/Config.php
fballiano: I was also thinking about wrapping everything in the "if usecache" check but at the end of the day your version is more readable and well written

app/code/core/Mage/Core/Model/Config.php
fballiano: it's just because if that file was already "required" before, then the content wouldn't run and you'd just jump to the `die()` so I think it's more correct to have a proper require there.

fballiano:
@colinmollenhour I can't find your comment about the timeout anymore but when you say that "if the timeout is 60 but after 3 the cache is generated and requests still keep piling up" I don't think it should be that way, if the lock gets released after 3 seconds then all the piled requests get executed right away, I think.

fballiano:
actually I think it's good to have a slightly bigger timeout because it forbids other requests (when timeout expires) from restarting the cache generation, that's what crashes servers, much more than requests piling up, at least in my experience

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,44 @@ public function getCache()
...
+        $connection = Mage::getSingleton('core/resource')->getConnection('core_write');
+        $lockWaitTime = getenv('MAGE_CONFIG_CACHE_LOCK_WAIT') ?: (PHP_SAPI === 'cli' ? 60 : 3);
+        if (!$connection->fetchOne("SELECT GET_LOCK('core_config_cache_save_lock', ?)", [$lockWaitTime])) {
+            if (PHP_SAPI === 'cli') {
+                throw new Exception('Could not get lock on cache save operation.');
+            } else {
+                require Mage::getBaseDir() . DS . 'errors' . DS . '503.php';
```
midlan, line 514: I would LOG somewhere the reason why application returned 503.

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,44 @@ public function getCache()
...
-    protected function _getCacheLockId()
+    public function getCacheSaveLock()
     {
-        return $this->getCacheId() . '.lock';
+        if (!Mage::app()->useCache('config')) {
+            return;
+        }
+        $connection = Mage::getSingleton('core/resource')->getConnection('core_write');
+        $lockWaitTime = getenv('MAGE_CONFIG_CACHE_LOCK_WAIT') ?: (PHP_SAPI === 'cli' ? 60 : 3);
+        if (!$connection->fetchOne("SELECT GET_LOCK('core_config_cache_save_lock', ?)", [$lockWaitTime])) {
```
midlan, line 510: Using `GET_LOCK` feature requires MySQL 5.7+, so we should update requirements section in `README.md` file.

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -501,13 +491,44 @@ public function getCache()
...
+     *
+     * @return void
+     */
+    public function releaseCacheSaveLock()
+    {
+        if (!Mage::app()->useCache('config')) {
+            return;
+        }
+        $connection = Mage::getSingleton('core/resource')->getConnection('core_write');
+        $connection->fetchOne("SELECT RELEASE_LOCK('core_config_cache_save_lock')");
```
midlan, line 531: I would put the name of lock `core_config_cache_save_lock` to constant, to avoid magic-string in unit tests.

fballiano:
@colinmollenhour there's now a conflict

colinmollenhour:
@fballiano This one should have been merged before #3533 because that one extends this one.. so most if not all of this is already merged now.. I will take a closer look.

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: Ahh, I thought we already required 5.7, but agreed the README needs to be updated.. Also `GET_LOCK()` is already used in `app/code/core/Mage/Index/Model/Resource/Helper/Mysql4.php`

colinmollenhour: Updated README.md

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: Added log statement with NOTICE level

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: What unit tests? 😉

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -502,17 +502,18 @@ public function getCache()
      */
     public function getCacheSaveLock($waitTime = null, $ignoreFailure = false)
     {
-        if (! Mage::app()->useCache('config')) {
+        if (!Mage::app()->useCache('config')) {
             return;
         }
-        $waitTime = $waitTime ?: (PHP_SAPI === 'cli' ? 60 : 3);
+        $waitTime = $waitTime ?: (getenv('MAGE_CONFIG_CACHE_LOCK_WAIT') ?: (PHP_SAPI === 'cli' ? 60 : 3));
```
fballiano, line 508: are we leaving 3 seconds as a default?

app/code/core/Mage/Core/Model/Config.php
midlan: Future 🙂

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: Open to changing, what's your proposal and justification?

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -502,17 +502,18 @@ public function getCache()
...
         $connection = Mage::getSingleton('core/resource')->getConnection('core_write');
         if (!$connection->fetchOne("SELECT GET_LOCK('core_config_cache_save_lock', ?)", [$waitTime])) {
             if ($ignoreFailure) {
                 return;
             } elseif (PHP_SAPI === 'cli') {
                 throw new Exception('Could not get lock on cache save operation.');
             } else {
+                Mage::log(sprintf('Failed to get cache save lock in %d seconds.', $waitTime), Zend_Log::NOTICE);
                 require_once Mage::getBaseDir() . DS . 'errors' . DS . '503.php';
```
colinmollenhour, line 517:
```suggestion
                require Mage::getBaseDir() . DS . 'errors' . DS . '503.php';
```

app/code/core/Mage/Core/Model/Config.php
colinmollenhour: After re-reading your previous comment:
> actually I think it's good to have a slightly bigger timeout because it forbids other requests (when timeout expires) from restarting the cache generation, that's what crashes servers, much more than requests piling up, at least in my experience

When the timeout occurs now, the timed out processes do not start saving the cache but rather they will exit with a 503 error. So, there is no risk of stampede but the timed out requests will obviously fail instead of potentially exhausting some other resource like PHP processes, RAM, web server connections, etc. In an ideal situation the systems are tuned so that the web server will handle these limits before they become a problem but a perfectly tuned server is probably pretty rare so I think it is safer to just consider 3 seconds to be a reasonable time to wait before failing. Adding the env var allows someone to adjust to their liking but I'm happy to concede that there is a better default value than 3 seconds if you make a good case. :)

colinmollenhour:
The changes in this PR are very minor since the bulk of the code was merged with #3533. It now just contains a README update, some syntax fixes and the commits and discussions addressing the concerns of @midlan and @fballiano.

app/code/core/Mage/Core/Model/Config.php
fballiano: Ideally 3 seconds is enough, but "real world poorly configured stores" I think it's borderline, I'd much rather prefer 10 seconds but actually it was already merged so...

fballiano (approved):

midlan (approved):

fballiano:
we need another review in order to merge

Sekiphp (approved):

Issue #3325, opened 2023-06-16
Deprecated notice when saving product attribute set


kyrena:
With OpenMage 20.x and PHP 8.1/8.2: `trim(): Passing null to parameter #1 ($string) of type string is deprecated`

![image](https://github.com/OpenMage/magento-lts/assets/78410399/a4c316fe-d907-4110-8dde-abdda211b6e8)

Without any changes (PHP 8.0), the query after the `foreach` is: `SELECT eav_attribute.* FROM eav_attribute WHERE (attribute_code='') AND (entity_type_id='4') AND (attribute_id!='1559')`

Possible solution?

```php
if (is_array($unique['field'])) {
    foreach ($unique['field'] as $field) {
        $select->where($field . '=?', trim((string)$data->getData($field)));
    }
} else {
    $select->where($unique['field'] . '=?', trim((string)$data->getData($unique['field'])));
}
```

It's weird.

fballiano:
solved by https://github.com/OpenMage/magento-lts/pull/3535

PR #3535, opened 2023-09-21
Fixed passing null for trim is deprecated in Mage_Core_Model_Resource_Db_Abstract

kyrena:
### Description

This fix `passing null to parameter #1 ($string) of type string is deprecated` for `trim` with PHP 8.1 / 8.2 / 8.3.
**Probably** it's not the best way, but it works with PHP 7.2 / 7.3 / 7.4 / 8.0 / 8.1 / 8.2 / 8.3.

To reproduce, go to any products attributes set, and save.

See also this [discussion comment](https://github.com/OpenMage/magento-lts/discussions/3025#discussioncomment-5195497) and #3325.

### Contribution checklist


app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
```diff
@@ -656,10 +657,10 @@ protected function _checkUnique(Mage_Core_Model_Abstract $object)
 
                 if (is_array($unique['field'])) {
                     foreach ($unique['field'] as $field) {
-                        $select->where($field . '=?', trim($data->getData($field)));
+                        $select->where($field . '=?', trim((string)$data->getData($field)));
```
kiatng, line 660:
```suggestion
                        if ($data->getData($field) !== null) {
                            $select->where($field . '=?', trim($data->getData($field)));
                        }
```

app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
```diff
@@ -656,10 +657,10 @@ protected function _checkUnique(Mage_Core_Model_Abstract $object)
...
                     }
                 } else {
-                    $select->where($unique['field'] . '=?', trim($data->getData($unique['field'])));
+                    $select->where($unique['field'] . '=?', trim((string)$data->getData($unique['field'])));
```
kiatng, line 663:
```suggestion
                } elseif ($data->getData($unique['field']) !== null) {
                    $select->where($unique['field'] . '=?', trim($data->getData($unique['field'])));
```

kiatng (commented):
Am I right to think that if it's null, there is no need to check for uniqueness?

[edit]
I checked, unique column allows multiple nulls, but not multiple empty strings.

app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
luigifab: It's probably better, I will try.

kyrena: No this does not work.

app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
kyrena: This is working when I save an attributes set, but are there side effects? I don't know.

app/code/core/Mage/Core/Model/Resource/Db/Abstract.php
kiatng: I  tested this by saving an attribute set with an existing name, it works just fine.

> Attribute set with the "Default" name already exists.

What error do you get?

kiatng (approved):
Tested.

fballiano (approved):

PR #3536, opened 2023-09-21
Updated composer.json to allow PHP 8.3

kyrena:
### Description

This PR allow to use PHP 8.3 with composer.
The list of changes is small: https://php.watch/versions/8.3

### Contribution checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     ],
     "type": "magento-source",
     "require": {
-        "php": ">=7.4 <8.3",
+        "php": ">=7.4 <8.4",
         "ext-ctype": "*",
         "ext-curl": "*",
         "ext-dom": "*",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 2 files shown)


elidrissidev:
Please run `composer update` after modifying `composer.json` to sync up `composer.lock` as well.

ADDISON74:
We have to check what challenges await us in PHP 8.3

https://php.watch/versions/8.3
https://www.zend.com/blog/php-8-3

This PR should also have a note in the README file that we do not guarantee that it works for production. In the case of version 8.2 we still find errors and warnings.

fballiano:
@ADDISON74 in the README we already have "PHP 7.4+ (PHP 8.0 is supported, PHP 8.1 supported but some warnings may be shown/logged, PHP 8.2 is usable but still being tested)" so I think it covers it.

we kinda have to allow this PR otherwise it's not even possible to test with php8.3, I don't think there's any other way

ADDISON74:
For those who are using **DDEV** for testing, PHP 8.3 is available too

https://ddev.readthedocs.io/en/latest/users/configuration/config/#php_version

ADDISON74 (approved):

elidrissidev (approved):

kyrena:
Yes sorry, I saw `composer update`, but I forgot.

PR #3537, opened 2023-09-21
Fixed empty exception when customer does not exist for products alerts

kyrena:
### Description

When a customer does not exist, products alerts will crash with an empty error message.
We encounter that when we after importing a database from production to dev/staging/local, without customers and orders, and testing products alerts.

### DIFF
--- a/app/code/core/Mage/ProductAlert/Model/Observer.php
+++ b/app/code/core/Mage/ProductAlert/Model/Observer.php
@@ -206,7 +206,7 @@ class Mage_ProductAlert_Model_Observer
                         if ($previousCustomer) {
                             $email->send();
                         }
-                        if (!$customer || !$customer->getId()) {
+                        if (!$customer->getId()) {
                             continue;
                         }
                         $previousCustomer = $customer;
--- a/app/code/core/Mage/ProductAlert/Model/Observer.php
+++ b/app/code/core/Mage/ProductAlert/Model/Observer.php
@@ -119,7 +119,7 @@ class Mage_ProductAlert_Model_Observer
                         if ($previousCustomer) {
                             $email->send();
                         }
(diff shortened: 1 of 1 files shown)


fballiano:
if the customer doesn't exist, wouldn't it be better to remove the alert?

kyrena:
I thort the same... But unsure.

fballiano:
cause otherwise the alert will continue to be processed forever :-\

app/code/core/Mage/ProductAlert/Model/Observer.php
```diff
@@ -119,7 +119,7 @@ protected function _processPrice(Mage_ProductAlert_Model_Email $email)
                         if ($previousCustomer) {
                             $email->send();
                         }
-                        if (!$customer) {
+                        if (!$customer || !$customer->getId()) {
```
kiatng, line 122: I think the original `if (!$customer) {` is always `false` because ` Mage::getModel('customer/customer')->load($alert->getCustomerId());` always return the customer model, even when the ID is invalid. So, this is a bug
```suggestion
                        if (!$customer->getId()) {
```

app/code/core/Mage/ProductAlert/Model/Observer.php
```diff
@@ -206,7 +206,7 @@ protected function _processStock(Mage_ProductAlert_Model_Email $email)
...
```
kiatng, line 209:
```suggestion
                        if (!$customer->getId()) {
```

kiatng:
There is a foreign key on `customer_id` on both the alert tables, but the check can be disabled during import.

fballiano (approved):

kiatng (approved):

PR #3539, opened 2023-09-21
Fixed passing null for explode is deprecated in Mage_Adminhtml_Block_System_Currency_Rate_Matrix

kyrena:
### Description

This fix `passing null to parameter #2 ($string) of type string is deprecated` for `explode` with PHP 8.1/8.2/8.3.

To got it, with a multi website instance, update the currency of one website, set for example _Chilean Peso / CLP_.
Then try to import currency rates, and if there is an import error, the deprecated notice is here.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/System/Currency/Rate/Matrix.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Currency/Rate/Matrix.php
@@ -71,7 +71,7 @@ class Mage_Adminhtml_Block_System_Currency_Rate_Matrix extends Mage_Adminhtml_Bl
 
         foreach ($array as $key => $rate) {
             foreach ($rate as $code => $value) {
-                $parts = explode('.', $value);
+                $parts = explode('.', (string)$value);
                 if (count($parts) === 2) {
                     $parts[1] = str_pad(rtrim($parts[1], 0), 4, '0', STR_PAD_RIGHT);
                     $array[$key][$code] = implode('.', $parts);


fballiano (approved):

ADDISON74 (approved):

sreichel:
Would it better to check for null, instead of type casting? (not only this PR)

ADDISON74:
> Would it better to check for null, instead of type casting? (not only this PR)

That's how we should proceed. I'm not a fan of type casting, especially when it's abused in too many places.

app/code/core/Mage/Adminhtml/Block/System/Currency/Rate/Matrix.php
```diff
@@ -71,7 +71,7 @@ protected function _prepareRates($array)
 
         foreach ($array as $key => $rate) {
             foreach ($rate as $code => $value) {
-                $parts = explode('.', $value);
+                $parts = explode('.', (string)$value);
```
kiatng, line 74: We expect `$value` to be a number, if it's `null`, set it to 0.
```suggestion
                $value ?? $value = 0;
                $parts = explode('.', $value);
```
The null coalescing operator above may not be a cup of tea for most, but I like it. See https://onlinephp.io/c/9a9a93. A conventional way:
```suggestion
                if ($value === null) {
                     $value = 0;
                }
                $parts = explode('.', $value);
```

fballiano: wouldn't it be better `$value ?? ''`? why exploding integer zero?

kiatng: $value is expected to be a number.

fballiano: explode expects a string https://www.php.net/explode

sreichel: `$value === null ? '' : explode(....`

kiatng (approved):
I think the original `$parts = explode('.', (string)$value);` is still the best as it'll work for all types. `$value` can be a number or null. For number, it can be int, float, number in string "1.23".

sreichel:
@kiatng its either a string o null? not?

kiatng:
> @kiatng its either a string o null? not?

```php
foreach ($rate as $code => $value) {
```

`$rate` can be from 
1. Read from the DB, `$value` is string
2. Line31: `$newRates = Mage::getSingleton('adminhtml/session')->getRates();`. If there is some custom code, `$rate` can be int or float.

Issue #3534, opened 2023-09-21
Attribute Mass Action JS Crash with Varien_Data_Form_Element_Label


Hanmac:
### Preconditions (*)

1. OpenMage 19.4.15

### Steps to reproduce (*)

1. have an EAV Attribute with `frontend_input` as `label`, which results in using `Varien_Data_Form_Element_Label` to render the field. Because i want the Attribute to be read only.
2. Now select multiple Products in the Grid and choose `Update Attributes` to get into the Attribute Mass Action.

### Expected result (*)

1. Working Attribute Mass Action.

### Actual result (*)

1. JS is crashing with because it tries to disable a HTML input with the attribute ID, but `Varien_Data_Form_Element_Label` didn't render the id around the text
```
Uncaught TypeError: $(...) is null
    disableFieldEditMode /js/mage/adminhtml/product.js:1006
    onCompleteDisableInited /js/mage/adminhtml/product.js:1025
```

### Possible solution
Either directly filter all attributes that uses `frontend_input` = `label`, Or have an event dispatched where you can filter the attribute collection yourself.

PR #3540, opened 2023-09-22
Excluded Varien_Data_Form_Element_Label from "edit attributes" mass action

Hanmac:
### Description (*)

- Label Attribute made with `Varien_Data_Form_Element_Label` doesn't have an HTML element for `disableFieldEditMode` or `enableFieldEditMode` to disable or enable
- first i excluded all attributes that are labels.
- i added a check in JS to prevent it from crashing even if the attributes are added to the Form.

1. Fixes OpenMage/magento-lts#3534 

1. have an EAV Attribute for `frontend_input` = `label`
2. go to the Product Grid and use the `Update Attributes` mass action
3. JS does crash, because `disableFieldEditMode` can't find an element with the given id.

 - [x] Add yourself to contributors list (i should already be added)

### DIFF
--- a/js/mage/adminhtml/product.js
+++ b/js/mage/adminhtml/product.js
@@ -1001,14 +1001,18 @@ function toogleFieldEditMode(toogleIdentifier, fieldContainer) {
 }
 
 function disableFieldEditMode(fieldContainer) {
-    $(fieldContainer).disabled = true;
+    if ($(fieldContainer)) {
+        $(fieldContainer).disabled = true;
+    }
     if ($(fieldContainer + '_hidden')) {
         $(fieldContainer + '_hidden').disabled = true;
     }
 }
 
 function enableFieldEditMode(fieldContainer) {
-    $(fieldContainer).disabled = false;
+    if ($(fieldContainer)) {
+        $(fieldContainer).disabled = false;
+    }
     if ($(fieldContainer + '_hidden')) {
         $(fieldContainer + '_hidden').disabled = false;
     }
--- a/app/code/core/Mage/Adminhtml/Helper/Catalog/Product/Edit/Action/Attribute.php
+++ b/app/code/core/Mage/Adminhtml/Helper/Catalog/Product/Edit/Action/Attribute.php
@@ -114,6 +114,7 @@ class Mage_Adminhtml_Helper_Catalog_Product_Edit_Action_Attribute extends Mage_C
                 ->getEntityType(Mage_Catalog_Model_Product::ENTITY)
                 ->getAttributeCollection()
                 ->addIsNotUniqueFilter()
+                ->addFieldToFilter('frontend_input', ['neq' => 'label'])
                 ->setInAllAttributeSetsFilter($this->getProductsSetIds());
 
             if ($this->_excludedAttributes) {


sreichel:
To make attributes read-only, you can lock them.

Please see https://github.com/sreichel/magento-StackExchange_LockAttributes

Hanmac:
> To make attributes read-only, you can lock them.
> 
> Please see https://github.com/sreichel/magento-StackExchange_LockAttributes

Your module doesn't work for me 
* I want my attributes blocked even for the Admin 
* Your lock Attributes stops them only in the Product Edit Tab, not in the Mass Action (which this issue/PR is about)

sreichel:
You can block it for admins too (via ACL).

I can add a config setting, to lock it for all. For mass-actions another observer could be added (?).

Hanmac:
> You can block it for admins too (via ACL).
Not if the Admin has All the Roles

> I can add a config setting, to lock it for all. For mass-actions another observer could be added (?).

Mass actions doesn't have an Event to Filter for the Attributes, they return the Collection, but only AFTER it has been loaded.
thats also what I have been complaining in the linked Issue

kiatng:
I cannot find `frontend_input` of type `label` . Is it added by 3rd-party with event `'adminhtml_product_attribute_types'`? Can someone shows me where it's added?

Hanmac:
> I cannot find `frontend_input` of type `label` . Is it added by 3rd-party with event `'adminhtml_product_attribute_types'`? Can someone shows me where it's added? image

The Attribute was added via sql update from one of our modules.
Label is a valid value for `Varien_Data_Form_Element_Label`

I didn't touch any other events

kiatng:
This PR is not easy to review:

1. The reviewers need a way to replicate the issue, but the input type is introduced in a custom code.
2. It's difficult to ascertain if the change is the best way to solve the issue.

I think the suggestion by @sreichel should be considered. An event-observer type of solution is better.

fballiano (approved):
I reproduced the error simply changing the frontend_input to label for an attribute.
Although I don't use label attributes it is still a valid variel_form_element and this PR fixes the problem so it is ok for me.

js/mage/adminhtml/product.js
```diff
@@ -1001,14 +1001,18 @@ function toogleFieldEditMode(toogleIdentifier, fieldContainer) {
 }
 
 function disableFieldEditMode(fieldContainer) {
-    $(fieldContainer).disabled = true;
+    if ($(fieldContainer)) {
```
fballiano, line 1004: from my tests, this check is not necessary, but it doesn't seem bad to have it anyway

sreichel:
From my POV it belongs to forms, not to product attributes. It seems like a "hack" to me.

However ...

sreichel:
@Hanmac im pretty sure it can be done via observer.

Please raise an issue and tell me ya needs.

- global disable (also for admins)
- prevent changing locked attributes in mass-actions
- ...

kiatng:
I can replicate the js error by changing an attribute's `frontend_input` to "label". However, I do not know to use this attribute in backend and frontend product page. If the `frontend_input` was "text" and its value is some sting, then on the backend product page I see this:
image

If there was no value, then I see this:
image

The attribute's value was rendered as "No" in frontend:
image

So, this PR does not make sense for me now. For the PR to be useful, is it possible to _properly_ add "label" as a valid   `frontend_input` type and explain how it is used?

kiatng:
Another possibility is to make use of the existing var `$this->_excludedAttributes` :

image

to dispatch an event. The var is hardcoded and not very useful currently. The event will finish the job.

Hanmac:
@kiatng There is also that does the job inside the block:
https://github.com/OpenMage/magento-lts/blob/5fad88396b4c7e58d359d08690425044834ec077/app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Action/Attribute/Tab/Attributes.php#L35-L38

js/mage/adminhtml/product.js
Hanmac: the problem is that `Varien_Data_Form_Element_Label` doesn't do a HTML Element with the Id of the attribute around the label text.

so then Prototype tries to access `$(fieldContainer)` it is null

kiatng:
@Hanmac You seem to know about the event that can be used to exclude the attributes of your choice. I tested the event and it works as expected, the attributes can be excluded without the js error. I do not understand why you still need to add `addFieldToFilter('frontend_input', ['neq' => 'label'])` in the core. Can you explain?

fballiano:
although it can be fixed via observer, isn't it anyway better to have it in the core?

Hanmac:
> @Hanmac You seem to know about the event that can be used to exclude the attributes of your choice. I tested the event and it works as expected, the attributes can be excluded without the js error. I do not understand why you still need to add `addFieldToFilter('frontend_input', ['neq' => 'label'])` in the core. Can you explain?
> 
> The script I used to exlude "label" attributes:
> 
> ```
>     // event 'adminhtml_catalog_product_form_prepare_excluded_field_list'
>     public function productMassActionExcludedFieldList($observer)
>     {
>         /** @var Mage_Adminhtml_Block_Catalog_Product_Edit_Action_Attribute_Tab_Attributes $block */
>         $block = $observer->getObject();
>         $excluded = $block->getFormExcludedFieldList();
>         $excluded = array_merge(
>             $excluded,
>             Mage::getResourceSingleton('catalog/attribute')->getAttributeCodesByFrontendType('label')
>         );
>         $block->setFormExcludedFieldList($excluded);
>     }
> ```

my problem was that the exludedFieldList event doesn't let me filter the attribute collection but more i need to find the attributes i want to exclude myself. (with additional queries/loading data)

in my code i had the attributes being hardcoded before, because i didn't know about `getAttributeCodesByFrontendType`

my boss said, Magento should handle that on its own (for example the JS error that still should be checked)

kiatng:
> although it can be fixed via observer, isn't it anyway better to have it in the core?

IMHO:
1. The "label" use case can only works by accessing the DB or by script to change the input type to "label" from other types, for new attributes, their values need to be set in the DB or by script. It's a job of a coder, which can write the observer. 
2. If the "label" use case is a common one, then we need to properly implement it, for example, "select" and "multiselect" types do not work when they are changed to "label". Because there is no UI for editing, if you need to change the attribute's value, you need to change it in the DB or by script.
3. Observer is the proper way to exclude attributes in mass action, especially there is an event designed for it.
4. If this PR is merged, there may be other cases that we need to `addFieldToFilter()`in the future. Perhaps we should consider removing the event since no one care to use it?

Hanmac:
> If this PR is merged, there may be other cases that we need to `addFieldToFilter()`in the future. Perhaps we should consider removing the event since no one care to use it?

my preference would have been if there was an event where i can have the collection to manually add my own filter.

the problem for me, while the getAttributes returns the collection (and i might overwrite the helper), the collection is already loaded because it has this part where it loads the attributes to do applyTo to them.
IMO wouldn't it be better if that could been handled via Query too?

kiatng:
@Hanmac OpenMage offers many ways to customize the code. I get that none of them is how you like it: direct manipulation of the collection without overwriting the class. I'm sorry but I cannot agree to this PR at this present state. Perhaps others in the community see it differently.

Flyingmana (approved):

fballiano:
@kiatng do you oppose to the merging of this one?

kiatng:
@fballiano Thanks for asking. My opinion is non-blocking to merge although I oppose it. 

This PR actually inspires discussion #3559, we need a way to document "new" uses of OpenMage. In this case, which is a hack of OpenMage. The doc would help other OpenMage users looking for non-editable attributes and how to do it.

PR #3542, opened 2023-09-22
Fixed PHPStan workflows not running on the correct PHP version

fballiano:
Without the `setup-php` step workflows are not running with the correct PHP version, this PR fixes it for PHPStan, I'll check more workflows later.

Also, this PR makes PHPStan run on php7.4, 8.2 and 8.3 for futureproofing.

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -12,9 +12,14 @@ jobs:
     strategy:
       matrix:
         operating-system: [ubuntu-latest]
-        php-versions: ['7.4', '8.2']
+        php-versions: ['7.4', '8.2', '8.3']
 
     steps:
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+    
       - uses: actions/checkout@v4
 
       - name: Check PHP version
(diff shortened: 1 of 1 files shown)


.github/workflows/phpstan.yml
```diff
@@ -8,9 +8,18 @@ on:
 jobs:
   phpstan:
     name: Analyze
-    runs-on: [ubuntu-latest]
+    runs-on: ${{ matrix.operating-system }}
+    strategy:
+      matrix:
+        operating-system: [ubuntu-latest]
+        php-versions: ['7.4', '8.2', '8.3']
```
ADDISON74, line 15: Did you rule out 8.0 and 8.1? Even if they are LTS for the PHP team, these versions are not for a few OS's that provide them. The versions benefit from periodic updates until the end of LTS.

fballiano: I thought:
- we have to test for the minimum supported version (7.4)
- I wanted to start testing for the upcoming PHP version (8.3)
- it seems right to test for the most recent supported version, which is also the most restrictive (8.2)
- not to overload github with too many workflow I didn't add more

kiatng (approved):

ADDISON74 (approved):

PR #3552, opened 2023-09-27
Removed weird comments

fballiano:
I found these weird comments in the source code:

`/**#@-*/`

I couldn't find where they come from or what's the meaning.

### DIFF
--- a/app/code/core/Mage/Admin/Model/User.php
+++ b/app/code/core/Mage/Admin/Model/User.php
@@ -70,7 +70,7 @@
  */
 class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
 {
-    /**#@+
+    /**
      * Configuration paths for email templates and identities
      */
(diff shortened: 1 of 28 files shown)


ADDISON74 (approved):

pquerner (approved):

elidrissidev:
Probably an ancient syntax for region folding, like for example you can do the following in VSCode to create a foldable code block with massactions as its title:
```php
//region massactions
/**
 * Current massactions
 * @var string
 */
public const ASSIGN_GROUP              = 'assign_group';
public const ATTRIBUTES                = 'attributes';
public const CANCEL_ORDER              = 'cancel_order';
public const CHANGE_MODE               = 'change_mode';
// more values...
//endregion
```

elidrissidev (approved):

kiatng:
Came from here https://github.com/OpenMage/magento-lts/commit/d0b97bacb84d061d98f7f765b9edf932d7ac6da8

PR #3290, opened 2023-06-01
Fixed customer not receiving order notification email when the email …

kiatng:
…is changed in customer module.

### Description (*)
See discussion #3289 

### Manual testing scenarios (*)

1. Change the email in frontend.
2. In backend, add a comment history in the order, tick checkbox _Notify Customer by Email_ and submit the comment
3. Without this PR, the email is sent to the old email address and not the email set in step 1.

### Contribution checklist (*)
 - [ ] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [ ] Add yourself to contributors list

PR #3560, opened 2023-09-29
Shell - change order email

m-overlund:
### Description (*)
OpenMage does not provide functionality to change email on an order - if for instance a customer has mistyped the email.

The script checks minimum length of email (5 chars) and order (9 chars) as default.

### Usage
`php shell/actions/change-email.php`


fballiano:
the shell scripts that could be implemented are many, I would much rather prefer to have a single utility (like magerun) instead of creating many scripts here :-)

ADDISON74:
I would make a rule of naming the files in the **/shell** directory. Regarding orders, start with order_, regarding customer with customer_, provided that there is at least one script for that category. At the moment we have only 3 files in /shell, abstract.php, indexer.php and log.php, which have been there for a long time and do not require change.

For this script I would set the name as **order_change-email.php**.

I leave it to everyone's discretion how a format would be the best choice, by using _ or - between words. I would follow the rule **category / action / target**. For example, if I need to change the email used in the order / change / email.

m-overlund:
i do agree on both points.
Would be nice if implemented functionality to n98-magerun.
However i would also like to see these basic functionalities in core OpenMage - but i dont know if the shell form is ideal for that.

ADDISON74:
The shell is the fastest way to implement simple actions. There are a lot of PHP script in StackExchange for different tasks. @sreichel knows better n98 and he can provide details if we can be implemented there such PRs. That would be great because the n98 is a very useful tool and I hope it will be updated for PHP 8.x soon.

If we are going to use the shell, then the idea shown above is similar to the way to remember the commands in n98. It only remains to set up how we name the files.

m-overlund:
agreed. 
But from a company point of view, changing a order email would normally be done by customer service - hence why i think shell script is not ideal as the only implementation.

ADDISON74:
In this case, an extension would be more suitable to allow a full editing of an order. There is a good offer from several developers. Until now I have never faced such a problem, but I have edited certain information directly in the database, correcting what cannot be done in the interface.

m-overlund:
agree - but shell like this is at least providing a way to do it, or have someone in our company do it.

sreichel:
I'd prefer to move it into another repo, like https://github.com/OpenMage/dev-translations/blob/main/src/shell/translations.php

And, i'd refactor the code, to "match" current shell scripts ...

@ADDISON74 magerun should be ready (had no time for final tests). Should be easy to add it not to n98-core, but into addon.

There is already something like this. https://github.com/kalenjordan/magerun-addons#assign-order-to-new-customer

Flyingmana:
We could create a new Repository in our Org for extending Magerun (via addon) with usefull tasks like this.

For the more UI centred approach there are a few different approaches, especially when thinking about posibilities to extend and scale it.
The easiest is probably writing a Report (or something similar) which can be exported as csv, and so Customer Service can open it and open the customers/Orders one by one to correct the data.
There are probably multiple more UX friendly and efficient approaches, but not sure how to design them to be easy extendable. Maybe the whole topic is a good thing for a separate module (which of course can also happen in the OpenMage Org)

sreichel:
Maybe put this on Awesome list ... https://github.com/kalenjordan/custom-reports @fab.

ADDISON74:
It's pretty much what I proposed, a repository for simple actions using CLI. In such of directory /actions, these files can be grouped in subdirectories such as customer, order and then their names will be simpler based on the format action-what_action(e.g. change-guest2customer.php).

ADDISON74:
Let's take a look to the other files in the shell directory.

1 - It must have a copyright block like the one bellow

```php
<?php
/**
 * OpenMage
 *
 * This source file is subject to the Open Software License (OSL 3.0)
 * that is bundled with this package in the file LICENSE.txt.
 * It is also available at https://opensource.org/license/osl-3-0-php
 *
 * @category   Mage
 * @package    Mage_Shell
 * @copyright  Copyright (c) 2022-2023 The OpenMage Contributors (https://www.openmage.org)
 * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
 */
```

2 - It should require_once 'abstract.php' for getting the right path related to Mage.php. I am not sure using public_html is correct for all cases.

3 - It should have the same format like log.php based on classes, having an usageHelp function., docblocks, ...

4 - You are using tabs instead of 2 blank spaces. It needs a better word processing.

kiatng:
This is remotely related to PR #3290, where it sends order transactional emails to address retrieve from `$order->getCurrentCustomerEmail()`. The email stored in the orders are for reference only. Wrong emails need to be corrected in the customer module.

m-overlund:
I have started trying to convert and adhere to the syntax of shell/log.php.
However Mage:: calls fails and running into Unable to start session error reports.

fballiano:
For the time being I'll close this PR, feel free to reopen it in case there's any update

PR #3563, opened 2023-09-30
Fixed missing layout parameter when choosing widget page group

ma4nn:
### Description (*)

This PR fixes the issue that the layout parameter is missing in some cases when choosing the page group on the widget instance page in backend.
As a result not all available block references will be shown in the select field.

This has already been tried to fix in https://github.com/OpenMage/magento-lts/pull/802, but the implementation does only work when choosing "Specified Page" in the page group select. For all other options like category or product pages (where no additional "Page" select is available) only the `default` layout handle will be used (instead of e.g. `catalog_product_view`) and therewith not all available blocks will be shown in "Block Reference".

1. In Backend go to _CMS > Widgets_ and create a new Widget instance
2. In the form section "Layout Updates" verify that all combinations of "Display On", "Page" and "Block Reference" produce the correct behaviour, i.e. the contents of the "Block Reference" select matches the chosen page
3. Verify step 2. especially for product pages vs. home page (as mentioned in https://github.com/OpenMage/magento-lts/pull/802)

### DIFF
--- a/app/design/adminhtml/default/default/template/widget/instance/edit/layout.phtml
+++ b/app/design/adminhtml/default/default/template/widget/instance/edit/layout.phtml
@@ -123,6 +123,7 @@ var pageGroupTemplate = '<div class="options-box page_group_container" id="page_
 '<div class="no-display ignore-validate pages group_container" id="pages_{{id}}">'+
     '<input type="hidden" class="container_name" name="__[container_name]" value="widget_instance[{{id}}][pages]" />'+
     '<input type="hidden" name="widget_instance[{{id}}][pages][page_id]" value="{{page_id}}" />'+
+    '<input type="hidden" class="layout_handle_pattern" name="widget_instance[{{id}}][all_pages][layout_handle]" value="default" />'+
     '<input type="hidden" class="for_all" name="widget_instance[{{id}}][pages][for]" value="all" />'+
     '<table cellspacing="0" class="option-header">'+
         '<col width="200" />'+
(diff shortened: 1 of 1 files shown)


fballiano:
@ma4nn I can't seem to replicate the issue or I didn't understand it correctly

Screenshot 2023-10-21 alle 11 55 26

it seems to me that the layout parameter is always passed

ma4nn:
@fballiano thanks for your feedback. As mentioned in the ticket description for "Display On = Specified Page" the `layout` parameter is correctly passed. But if you choose something else like "All Product Types" the `layout` parameter is empty and as a result the "Block Reference" select field is missing values like "Bottom Block Options Wrapper" which are only available for `catalog_product_view` layout handle.

fballiano:
select#layout_handle doesn't even exist in the DOM, interesting, I'm trying to think of why nobody noticed before but this PR seems to work fine.

fballiano:
before PR
Screenshot 2023-10-22 alle 16 13 56

after PR
Screenshot 2023-10-22 alle 16 13 22

I can't find any problem with this PR so I'm approving it

fballiano (approved):

ma4nn:
What about this mini-bugfix? Is any other reviewer able to verify and approve?

fballiano:
@ADDISON74 @kiatng @Flyingmana I think this one should be approved, I've tested it and it works

kiatng (approved):
Following the testing steps outlined: after creating a widget, and select it for editing, and set _Display On_ to _All Product Types_, the dropdown options for _Block References_ increases as shown by the screenshots captured by @fballiano:
image

ADDISON74 (approved):

PR #3564, opened 2023-10-01
ensure value is string to avoid type error in used library

Flyingmana:
### Description (*)
There was an error on PHP 8.2 about urlencode() expecting a string.
Should probably get fixed upstream in `\Zend_Locale_Data::getList` / `vendor/shardj/zf1-future/library/Zend/Locale/Data.php`

### Related Pull Requests
https://github.com/OpenMage/magento-lts/pull/3234 (resolves an error reported there)

have error reporting -1
cleared cache
login into admin area


Flyingmana:
although Iam not sure why we have this issue, as we depend on a version which should already have this fix https://github.com/Shardj/zf1-future/blob/release-1.23.5/library/Zend/Locale/Data.php#L336

fballiano:
I can't seem to reproduce the error, the version of the file that I have already has the cast to string:

Screenshot 2023-10-01 alle 21 37 24

Flyingmana:
yeah, not sure, maybe updating the source without updating the composer dependencies(thats what happened for me when I tested it first).

It was reported, so I think its at least worth to consider fixing it there, even as in theory it should not be possible to happen.

fballiano:
maybe it was using a OM version before the one when we strictly required zf1f version?

fballiano:
At the moment I'll close this, still can't reproduce, probably it's being fixed in a previous release ;-)

PR #3568, opened 2023-10-02
Fixed docblock onepage.phtml, cart.phtml, and success.phtml in checkout.

kiatng:
### Description (*)
Also removed deprecated getter in Mage_Checkout_Block_Onepage_Success.

### DIFF
--- a/app/code/core/Mage/Checkout/Block/Onepage/Success.php
+++ b/app/code/core/Mage/Checkout/Block/Onepage/Success.php
@@ -21,6 +21,9 @@
  *
  * @method $this setCanViewProfiles(bool $value)
  * @method $this setRecurringProfiles(Mage_Sales_Model_Recurring_Profile[] $value)
+ * @method string getOrderId()
+ * @method string getPrintUrl()
+ * @method string getViewOrderUrl()
  */
 class Mage_Checkout_Block_Onepage_Success extends Mage_Core_Block_Template
 {
@@ -28,17 +31,6 @@ class Mage_Checkout_Block_Onepage_Success extends Mage_Core_Block_Template
      * @deprecated after 1.4.0.1
      */
     private $_order;
-
-    /**
-     * Retrieve identifier of created order
-     *
-     * @return string
-     * @deprecated after 1.4.0.1
-     */
-    public function getOrderId()
-    {
-        return $this->_getData('order_id');
-    }
 
     /**
      * Check order print availability
@@ -49,28 +41,6 @@ class Mage_Checkout_Block_Onepage_Success extends Mage_Core_Block_Template
     public function canPrint()
     {
         return $this->_getData('can_view_order');
-    }
-
-    /**
-     * Get url for order detale print
-     *
-     * @return string
-     * @deprecated after 1.4.0.1
-     */
-    public function getPrintUrl()
-    {
-        return $this->_getData('print_url');
-    }
-
-    /**
-     * Get url for view order details
-     *
-     * @return string
-     * @deprecated after 1.4.0.1
-     */
-    public function getViewOrderUrl()
-    {
-        return $this->_getData('view_order_id');
     }
 
     /**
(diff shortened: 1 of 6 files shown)


fballiano (approved):

tmotyl (approved):

ADDISON74 (dismissed):

app/code/core/Mage/Checkout/Block/Onepage/Success.php
```diff
@@ -43,6 +45,17 @@ public function canPrint()
         return $this->_getData('can_view_order');
     }
 
+    /**
+     * Get url for view order details
+     *
+     * @return string
+     * @deprecated after 1.4.0.1
```
fballiano, line 52: if this is necessary probably we should un-deprecate it?

elidrissidev (dismissed):

app/code/core/Mage/Checkout/Block/Onepage/Success.php
kiatng: You are right!

fballiano (approved):

sreichel:
The methods are still used. Better keep them.

kiatng:
> The methods are still used. Better keep them.

Two methods were removed in this PR:

```php
    public function getOrderId()
    {
        return $this->_getData('order_id');
    }
    
    public function getPrintUrl()
    {
        return $this->_getData('print_url');
    }
```

As you can see, we use magic getter. I know you prefer not to use magic; either remove the methods or we need to remove the deprecated directive. I thought it's cleaner to remove the methods. Hope it's ok.

sreichel:
I don't see any benefit to change it to magic methods, that are slower.

kiatng:
Overall consistency can be a benefit.

sreichel:
Then it would be to replace magic ones?

sreichel:
@kiatng fyi ... a lot of added method annotations are only the for better IDE-support, not to make it common practice. 

In a perfect world they should be replaced by real methods, that are 4 times faster ... see my PR for getData().

kiatng:
Performance is an important consideration. Magic method is in the DNA, it predates the IDE annotations. It offers flexibility. It's our signature feature. It was a common practice, I would like to think that it is still. For this PR, I do think it's a right call to remove the 2 methods, to make it consistent with the code base.

On the other hand, we can introduce explicit setter and getter in the entire code base. If we are heading this direction, then this PR is wrong.  However, this is a larger discussion that needs to take place with a consensus decision. I'd be happy to revert this PR then.

app/code/core/Mage/Checkout/Block/Onepage/Success.php
```diff
@@ -21,6 +21,11 @@
  *
  * @method $this setCanViewProfiles(bool $value)
  * @method $this setRecurringProfiles(Mage_Sales_Model_Recurring_Profile[] $value)
+ * @method string getOrderId()
```
Flyingmana, line 24: this method was marked as deprecated before the change

app/code/core/Mage/Checkout/Block/Onepage/Success.php
```diff
@@ -21,6 +21,11 @@
...
+ * @method string getPrintUrl()
```
Flyingmana, line 25: this method was marked as deprecated before the change

app/code/core/Mage/Checkout/Block/Onepage/Success.php
```diff
@@ -51,22 +45,10 @@ public function canPrint()
...
-    public function getPrintUrl()
-    {
-        return $this->_getData('print_url');
-    }
-
     /**
      * Get url for view order details
      *
      * @return string
-     * @deprecated after 1.4.0.1
```
Flyingmana, line 69: what was the reason for making it not deprecated anymore?

PR #3577, opened 2023-10-04
Added label to system.xml

kyrena:
### Description

This PR add the label field to system.xml, this is an extract of #2739:

image

XML of the preview:

```xml
<head>
	<label>Crash test</label>
	<frontend_model>adminhtml/system_config_form_field_heading</frontend_model>
	<sort_order>1001</sort_order>
</head>
<label_1>
	<label>label</label>
	<value>Pouet</value>
	<frontend_type>label</frontend_type>
	<frontend_model>adminhtml/system_config_form_field_label</frontend_model>
	<sort_order>1002</sort_order>
</label_1>
<labelbold_2>
	<label>label bold</label>
	<value>Pouet Pouet</value>
	<bold>1</bold>
	<frontend_type>label</frontend_type>
	<frontend_model>adminhtml/system_config_form_field_label</frontend_model>
	<sort_order>1003</sort_order>
</labelbold_2>
```

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Label.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Label.php
@@ -29,11 +29,9 @@ class Mage_Adminhtml_Block_System_Config_Form_Field_Label extends Mage_Adminhtml
     protected function _getElementHtml(Varien_Data_Form_Element_Abstract $element)
     {
         $field = $element->getFieldConfig();
+        $element->setValue($field->value);
+        $element->setBold(!empty($field->bold));
 
-        $label = Mage::getModel('Varien_Data_Form_Element_Label');
-        $label->setValue($field->value);
-        $label->setBold(!empty($field->bold));
-
-        return $label->getElementHtml();
+        return $element->getElementHtml();
     }
 }
--- a/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Label.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Label.php
@@ -1,0 +1,33 @@
+<?php
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
+ * @license    http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Adminhtml system config label field renderer
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ */
+class Mage_Adminhtml_Block_System_Config_Form_Field_Label extends Mage_Adminhtml_Block_System_Config_Form_Field
+{
+    protected function _getElementHtml(Varien_Data_Form_Element_Abstract $element)
+    {
+        $field = $element->getFieldConfig();
+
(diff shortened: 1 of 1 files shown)


fballiano:
Screenshot 2023-10-04 alle 14 01 21

wouldn't it be better to use the frontend_type instead of the frontend_model?

fballiano:
also, I don't love the "store view" indicator for something that cannot change at a store view level, but I don't know if there's a way around that.

Screenshot 2023-10-04 alle 14 02 33

kyrena:
Oh yes, true, I can remove it.

kiatng:
How to use _Use Website_?
image

app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field/Label.php
```diff
@@ -0,0 +1,39 @@
...
+
+    protected function _getElementHtml(Varien_Data_Form_Element_Abstract $element)
+    {
+        $field = $element->getFieldConfig();
+
+        $label = Mage::getModel('Varien_Data_Form_Element_Label');
+        $label->setValue($field->value);
+        $label->setBold(!empty($field->bold));
+
+        return $label->getElementHtml();
```
kiatng, line 37: Simplify version:
```suggestion
    /**
     * @inheritdoc
     */
    public function render(Varien_Data_Form_Element_Abstract $element)
    {
        $element->unsScope()->unsCanUseWebsiteValue()->unsCanUseDefaultValue()->unsPath();
        $field = $element->getFieldConfig();
        $element->setValue($field->value);
        $element->setBold(!empty($field->bold));
        
        return parent::render($element);
```

luigifab: Yes, but it's not the way of `Mage_Adminhtml_Block_Customer_System_Config_Validatevat` and `Mage_Adminhtml_Block_System_Config_System_Storage_Media_Synchronize `, so it's probably not the good way to simplify?

But perhaps we can add phpdoc.

kiatng, line 37: Why is it necessary to get `$label = Mage::getModel('Varien_Data_Form_Element_Label');`?  Is it better to use the incoming `$element`?
```suggestion
        $element->setValue($field->value);
        $element->setBold(!empty($field->bold));

        return $element->getElementHtml();
```

elidrissidev:
> wouldn't it be better to use the frontend_type instead of the frontend_model?

`<frontend_type>` alone should be enough.

luigifab (approved):

fballiano (changes requested):
.

fballiano:
EDIT: apologize, my comment was wrong.

the current state of the PR is this:
Screenshot 2023-10-21 alle 15 13 14

but it still requires a frontend_model, which I don't really like

ma4nn:
Why not use the existing `Varien_Data_Form_Element_Label` or `Varien_Data_Form_Element_Note`?

kyrena:
Because we must pass config data from `$element->getFieldConfig()`:
```php
$label->setValue($field->value);
$label->setBold(!empty($field->bold));
```

ma4nn:
ok understood the idea now 👍  (the `Varien_Data_Form_Element_Label` can also handle the `value` but not the `bold` option)

But wouldn't it make more sense then to have a more flexible label with HTML values allowed in general (so the `bold` attribute is not necessary)? E.g. perhaps use
```xml
<frontend_type>note</frontend_type>
<frontend_model>adminhtml/system_config_form_field_label</frontend_model>
```
and then do a `$element->setText($field->value);`?

Though I agree with @fballiano that the naming is not really intuitive..

fballiano:
IMHO the current implementation is a bit too limited:
- I think it should support full HTML as @ma4nn pointed out, removing the need for ->setBold
- using fontend_type and not requiring a frontend_model would seem more intuitive to me

PR #3578, opened 2023-10-04
Exception is now printed on Mage_Catalog_ProductController::viewAction() in developer mode

kyrena:
### Description

When there is an error in `$viewHelper->prepareAndRender($productId, $this, $params);`, when developer mode is enabled, error is only logged, not displayed. This PR display the error to the user.

### Contribution checklist

### DIFF
--- a/app/code/core/Mage/Catalog/controllers/ProductController.php
+++ b/app/code/core/Mage/Catalog/controllers/ProductController.php
@@ -126,7 +126,7 @@ class Mage_Catalog_ProductController extends Mage_Core_Controller_Front_Action
                 } elseif (!$this->getResponse()->isRedirect()) {
                     $this->_forward('noRoute');
                 }
-            } else if (Mage::getIsDeveloperMode()) {
+            } elseif (Mage::getIsDeveloperMode()) {
                 Mage::printException($e);
             } else {
                 Mage::logException($e);
--- a/app/code/core/Mage/Catalog/controllers/ProductController.php
+++ b/app/code/core/Mage/Catalog/controllers/ProductController.php
@@ -126,6 +126,8 @@ class Mage_Catalog_ProductController extends Mage_Core_Controller_Front_Action
                 } elseif (!$this->getResponse()->isRedirect()) {
                     $this->_forward('noRoute');
                 }
+            } else if (Mage::getIsDeveloperMode()) {
+                Mage::printException($e);
             } else {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/controllers/ProductController.php
```diff
@@ -126,6 +126,8 @@ public function viewAction()
                 } elseif (!$this->getResponse()->isRedirect()) {
                     $this->_forward('noRoute');
                 }
+            } else if (Mage::getIsDeveloperMode()) {
```
fballiano, line 129:
```suggestion
            } elseif (Mage::getIsDeveloperMode()) {
```

app/code/core/Mage/Catalog/controllers/ProductController.php
```diff
@@ -126,6 +126,8 @@ public function viewAction()
                 } elseif (!$this->getResponse()->isRedirect()) {
                     $this->_forward('noRoute');
                 }
+            } elseif (Mage::getIsDeveloperMode()) {
+                Mage::printException($e);
```
elidrissidev, line 130: Missing redirection here

luigifab: Yes, `printException` stop the process, so it's useless to redirect.

kiatng:
Nice, but I think it should be the first condition after catch:

```php
        } catch (Exception $e) {
            if (Mage::getIsDeveloperMode()) {
                  Mage::printException($e);
            } elseif ($e->getCode() == $viewHelper->ERR_NO_PRODUCT_LOADED) {
```

What do you think?

luigifab:
I think it's not usefull, because for here there is no real exception: `throw new Mage_Core_Exception($this->__('Product is not loaded'), $this->ERR_NO_PRODUCT_LOADED);`.

elidrissidev (approved):

fballiano:
@kyrena how can I reproduce the error?

fballiano:
Since I didn't get any answer I started checkin the code, as @luigifab says it's really hard to trigger an exception there, if the product doesn't exists it goes to 404, but still, somebody could hook in some event and trigger an exception there. so, all and all, I think I agree with the change.

fballiano (approved):

kyrena:
@fballiano I don't remember how to reproduce the error...

PR #3574, opened 2023-10-03
Switched DOCTYPE to HTML5 for base theme

fballiano:
I thought it may be time to migrate away from XHTML 1.0 and switch the doctype for base templates to HTML5.

PR #3579, opened 2023-10-05
Switched DOCTYPE to HTML5 for adminhtml themes

fballiano:
In https://github.com/OpenMage/magento-lts/pull/3574 @m-overlund suggested to check also the adminhtml templates, this PR is for that reason.

### DIFF
--- a/app/design/adminhtml/default/default/template/empty.phtml
+++ b/app/design/adminhtml/default/default/template/empty.phtml
@@ -13,8 +13,8 @@
  * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
  */
 ?>
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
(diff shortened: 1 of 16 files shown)


fballiano:
@ADDISON74 I've added the 3 files but it's better to test the email templates now because that's the most delicate

ADDISON74 (dismissed):

ADDISON74:
I made this change a long time ago in the email templates. There could be problems with very, very old email clients.

ADDISON74:
To be sure, I will check today the modified email template with a very old email client. I guess that Outlook Express shipped with Windows XP could be the right choice.

fballiano:
😅 prehistoric email client 😂

ADDISON74:
Please separate that file from the email template in another PR because discussions are requested. The initial form was not suitable but it works. I have analyzed several email stand-alone clients and online messaging services and none of them use the original and modified version in the HTML part. The goal is to make sure that the message is not considered SPAM because of the chosen format.

fballiano:
@ADDISON74 i've removed it from this PR

ADDISON74 (approved):

elidrissidev (approved):

PR #3583, opened 2023-10-06
Removed text-shadow from disabled dropdown in Admin

elidrissidev:
### Description (*)

Minor UX improvement to disabled dropdowns in Admin that removes `text-shadow` for better readability.

| Before | After |
| --- | --- |
| Screen Shot 2023-10-06 at 10 58 11 AM | Screen Shot 2023-10-06 at 10 57 41 AM |

1. Go to any admin page that has a disabled dropdown, example: Configuration -> Payment Methods -> Cash On Delivery Payment.

### DIFF
--- a/skin/adminhtml/default/default/boxes.css
+++ b/skin/adminhtml/default/default/boxes.css
@@ -355,7 +355,7 @@ select.multiselect option       { padding:3px 4px; border-bottom:1px solid #ddd;
 .field-100 input.input-text { float:left; width:100% !important; border:0 !important; padding:0 !important; }
 @media screen and (-webkit-min-device-pixel-ratio:0) {
     select option:disabled,
-    select:disabled option { color:#c9c9c9!important;color:#cacaca!important; text-shadow:2px 2px 2px #000; }
+    select:disabled option { color:#cacaca!important; }
 }
 
 /* Form List */ /* Table for default form data */


ADDISON74:
Have you checked if it needs to be modified in any SCSS file?

OT: What a funny color value "#cacaca" chose the Magento team. This word in some languages means poop/shit.

elidrissidev:
> Have you checked if it needs to be modified in any SCSS file?

Yes, the OM theme had no override for those styles.

fballiano (approved):

ADDISON74 (approved):

ADDISON74:
Even I approved this PR, I noticed that it creates issues. Below are screenshots that prove it. Aheadworks Z-Blocks extension relies on the "select.mutiselect option" in boxes.css. 

**If "Use parent values" is checked, you can see that the text is no longer visible.**

screenshot_1

**If "Use parent values" is unchecked then it becomes visible**

screenshot_2

We should revert this PR or remove the color property, here is the effect below

screenshot_3

ADDISON74:
This is how it looks the multiselect revering this PR

screenshot_4

It is obvious we did not take into account that there is also the checked option which makes the background the same with the text.

ADDISON74:
or we can use a soft shadow

```css
text-shadow: 1px 1px 1px #000
```

Issue #3593, opened 2023-10-11
Dashboard Graph is skipping Months in 1y view


Hanmac:
### Preconditions (*)

1. Have OpenMage with Dashboard Graph enabled

### Steps to reproduce (*)

1. Open the Backend with Dashboard, and select 1y for period

### Expected result (*)

1. have a graph with all the months

### Actual result (*)

1. 
![testgraph](https://github.com/OpenMage/magento-lts/assets/225553/b71c5628-429a-4dd6-83af-ef5e97667922)

The Faulty Code piece:
```php
        while($dateStart->compare($dateEnd) < 0){
            switch ($this->getDataHelper()->getParam('period')) {
                case '24h':
                    $d = $dateStart->toString('yyyy-MM-dd HH:00');
                    $dateStart->addHour(1);
                    break;
                case '7d':
                case '1m':
                    $d = $dateStart->toString('yyyy-MM-dd');
                    $dateStart->addDay(1);
                    break;
                case '1y':
                case '2y':
                    $d = $dateStart->toString('yyyy-MM');
                    $dateStart->addMonth(1);
                    break;
            }
```

Results in this data from $dateStart (is there a logic behind this jumps for addMonth?)
```
01.01.2023 00:00:00
03.03.2023 00:00:00
01.04.2023 00:00:00
30.05.2023 00:00:00
01.06.2023 00:00:00
30.07.2023 00:00:00
01.08.2023 00:00:00
01.10.2023 00:00:00
```

and uses this for fields:
```
Array
(
    [0] => 2023-01
    [1] => 2023-03
    [2] => 2023-04
    [3] => 2023-05
    [4] => 2023-06
    [5] => 2023-07
    [6] => 2023-08
    [7] => 2023-10
)
```

this results in the graph missing a few months

Probably solution would be to use DateTime instead of Zend_Date ?

fballiano:
I've tested this on a customer's website and I can confirm:
`"chxl": "0:|01/2023|03/2023|04/2023|05/2023|06/2023|07/2023|08/2023|10/2023`

february and september are missing...

fballiano:
I've also done this:

```php
        echo $dateStart . "\n";
        $dateStart->addMonth(1);
        echo $dateStart . "\n";
        $dateStart->addMonth(1);
        echo $dateStart . "\n";
        $dateStart->addMonth(1);
        echo $dateStart . "\n";
        $dateStart->addMonth(1);
        echo $dateStart . "\n";
[...]

PR #3594, opened 2023-10-11
Fixed bug with backend dashboard about graph chart months calculations

fballiano:
Seems like Zend_Date has some bug calculating time differences when using timezones, for this reason the backend dashboard graph, when selecting `1Y` or `2Y` miscalculates the months to show and for this reason the graph is missing data.

More detailed explanation in bug report https://github.com/OpenMage/magento-lts/issues/3593

This PR rewrites the whole thing using `DateTime`, which seems more correct to me.

Fixes https://github.com/OpenMage/magento-lts/issues/3593


fballiano:
I preferred not to change the variables names, did you have the chance to test this PR?

elidrissidev:
@fballiano Can you try setting your timezone in configuration to Etc/GMT-12, cause for me it's still skipping some months.

download

fballiano:
@kiatng mmm interesting, since your TZ is `-12`, the first generated day becomes `2022-12-31 12:00:00` instead of `2023-01-01` so everything gets messed up.

I think that for `1Y` and `2Y` we shouldn't use the timezone at all, what do you think?

fballiano:
but I think timezone is needed for the hourly/24h reports, no?

empiricompany (approved):

boesbo (approved):

boesbo (approved):

Hanmac:
@fballiano are you going to revive this issue?

fballiano:
@Hanmac I left openmage as of today, I will maybe write a statement in the next days.

Hanmac:
@sreichel should we try to revive this MR?

PR #3596, opened 2023-10-13
Added before_cssjs and do not render the same block multiple times

luigifab:
### Description

I would like to add some JS code before `getCssJsHtml()` (for Sentry).
With this PR, it's now possible with `<reference name="before_cssjs">`.

But my block is rendered two times, first with `getChildHtml('before_cssjs')`, secondly with `getChildHtml()`.
With the change in `Abstract.php`, problem is now fixed.
Is it a good idea? I don't know.

### Contribution checklist


app/code/core/Mage/Core/Block/Abstract.php
```diff
@@ -1533,4 +1544,14 @@ protected function _isSecure()
     {
         return $this->_getApp()->getFrontController()->getRequest()->isSecure();
     }
+
+    /**
+     * Flag
+     *
```
fballiano, line 1550:
```suggestion
```

fballiano:
questions:
- what if you want a block to be rendered twice?
- wouldn't it work for you to add your block `before ='-'` in the head instead of also adding the `$this->getChildHtml('before_cssjs')`?

I guess the problem comes from the fact that the block is a "list" but adding the isRendered to the abstract class couldn't it be dangerous?

fballiano:
Since the unanswered question I'll close this at the moment. feel free to reopen in case there are news.

luigifab:
No it doesn't work with before, because the `getChildHtml` is too late.

Yes, this prevent multiple block rendering. But for me I didn't see any differences.

PR #3598, opened 2023-10-16
bugfix/LogVisitorSaveByRequest: added initServerData() to populate ServerData on saving visitor_log information

eneiasramos:
bugfix/LogVisitorSaveByRequest: added initServerData() to populate ServerData on saving visitor_log information

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Log/Model/Visitor.php
+++ b/app/code/core/Mage/Log/Model/Visitor.php
@@ -258,6 +258,7 @@ class Mage_Log_Model_Visitor extends Mage_Core_Model_Abstract
         }
 
         try {
+            $this->initServerData();
             $this->setLastVisitAt(Varien_Date::now());
             $this->save();
             $this->_session->setVisitorData($this->getData());


ADDISON74:
This is not a report based on the OM requested standard format. I would like you to provide information about the test environment and the steps to reproduce the error. Otherwise, I will close the report.

eneiasramos:
@ADDISON74 I'll provide you the information my dear :+1:

eneiasramos:
image

@ADDISON74 as you can see: only after the fix the URLs started to be recorded corretly in the database when changing the URL.

app/code/core/Mage/Log/Model/Visitor.php
```diff
@@ -258,6 +258,7 @@ public function saveByRequest($observer)
         }
 
         try {
+            $this->initServerData();
```
kiatng, line 261: The table `log_url_info` is updated here: https://github.com/OpenMage/magento-lts/blob/4b0be45ecfd569e3cc5716ec5ec2490c7559613c/app/code/core/Mage/Log/Model/Resource/Visitor.php#L69-L83

Instead of full init, perhaps we can selectively set the affected data:

```php
            $this->setHttpHost($this->_httpHelper->getHttpHost(true));
            $this->setRequestUri($this->_httpHelper->getRequestUri(true));
            $this->setHttpReferer($this->_httpHelper->getHttpReferer(true));
```

But I am not sure if this is the best, what do you think?

eneiasramos: @kiatng let me try to do this here my dear...

eneiasramos: @kiatng it worked! But, I think that duplicating the initServerData routines would not be an interesting alternative. That's why I kept the initServerData call inside the saveByRequest() method.

kiatng (approved):

eneiasramos:
@kiatng thank you my dear!

fballiano (approved):

eneiasramos:
@fballiano thank you my dear!

PR #3607, opened 2023-10-23
404 Responses Shouldn't Be Cached

Judx:
### Description (*)

404 Responses Shouldn't Be Cached: When a resource is not found we don't want the response to be cached. Caching a 404 response could lead to incorrect behavior for users who expect to see different content when they visit a page that previously returned a 404 error.

1. Call noroute -action
2. The HTTP Response should now include `Cache-Control` property with following values: `no-store, no-cache, must-revalidate, max-age=0`.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Controller/Rss/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Controller/Rss/Abstract.php
@@ -47,7 +47,6 @@ class Mage_Adminhtml_Controller_Rss_Abstract extends Mage_Adminhtml_Controller_A
         } else {
             $this->getResponse()->setHeader('HTTP/1.1', '404 Not Found');
             $this->getResponse()->setHeader('Status', '404 File not found');
-
             $this->_forward('noRoute');
             return false;
         }
--- a/app/code/core/Mage/Adminhtml/Controller/Rss/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Controller/Rss/Abstract.php
@@ -47,6 +47,7 @@ class Mage_Adminhtml_Controller_Rss_Abstract extends Mage_Adminhtml_Controller_A
         } else {
             $this->getResponse()->setHeader('HTTP/1.1', '404 Not Found');
             $this->getResponse()->setHeader('Status', '404 File not found');
+
             $this->_forward('noRoute');
             return false;
         }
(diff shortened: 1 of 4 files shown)


ADDISON74:
In my opinion, a correct observation. 

@colinmollenhour - what do you think about this PR?

colinmollenhour:
Given this is an admin controller, under what circumstances are responses being cached?

fballiano:
@colinmollenhour is always on point, I checked an admin 404 and it doesn't seem to be cached at a first glance:

Screenshot 2023-10-23 alle 17 55 32

app/code/core/Mage/Adminhtml/Controller/Rss/Abstract.php
```diff
@@ -47,6 +47,7 @@ protected function checkFeedEnable($code)
         } else {
             $this->getResponse()->setHeader('HTTP/1.1', '404 Not Found');
             $this->getResponse()->setHeader('Status', '404 File not found');
+
```
fballiano, line 50:
```suggestion
```

fballiano:
for CMS I already get "Cache-Control:no-store, no-cache, must-revalidate" so it this needed? but also, I understand the reason behind but attacking 404 pages couldn't be a vector for DOS attacks?

Judx:
> for CMS I already get "Cache-Control:no-store, no-cache, must-revalidate" so it this needed? but also, I understand the reason behind but attacking 404 pages couldn't be a vector for DOS attacks?

I believe these changes are needed. Indeed they dont necessary fix a bug but enforce the wanted.
I dont see how this change is related to denial of service -attack.

fballiano:
> I dont see how this change is related to denial of service -attack.

actually true, who wants to attack you won't respect cache policies anyway, sorry about that

Flyingmana:
One possible reason to not do this by default is reverse proxys like varnishe or cloudflare. They now need specialized configs to overwrite the caching for these routes.

It might be different when the content is user specific, but then it should not be a 404 but an unauthorized access, and the cache should be user specific, which also needs some special logic compared to our default behavior, so its maybe rather falling into the responsibility of the module providing this user based access limitation.

colinmollenhour:
When you are unauthenticated all admin routes return a 200 response which is the login page. When you are authenticated, why would you hit a 404 page other than mistakenly and why would any proxy cache that response even without the Cache-Control header? I'm not sure what I'm missing here but what was the motivation for opening this PR, it seems like it is fixing a problem that doesn't really exist.

joshua-bn:
Strong disagree here. Cache the 404. 

Also, there's `setHttpResponseCode()` that should be used. It is missing all over and incorrect responses are sent. This is a completely different issue. I think I brought it up before, but maybe only on my internal team discussions. I fixed it locally and see much better results.

Judx:
I close this PR because 1. There is no problem just enforcing something suggested and 2. The solution is disputed.

PR #3613, opened 2023-10-26
New feature: added `mage_run_installed_exception` event when uncatched exception is thrown

pquerner:
### Description (*)

This will add a new event called `mage_installed_exception`.
It will be thrown when OpenMage is installed and an uncatched exception is handled in the `Mage::run` method.

Using this new event it is possible to attach a exception facility, such as (Sentry)[https://sentry.io/].

### Related Pull Requests

None

None, it got converted to a discussion: https://github.com/OpenMage/magento-lts/discussions/3510

1. Throw exception anywhere thats unhandled by itself. For instance in an controller.
 ```
Mage::throwException('Dummy exception');
 ```
4. It will finally be catched in `Mage::run`
5. Listen to the new event, and do whatever you want with the exception. A use case has been discussed here: https://github.com/OpenMage/magento-lts/discussions/3510


app/Mage.php
```diff
@@ -751,6 +751,7 @@ public static function run($code = '', $type = 'store', $options = [])
             die();
         } catch (Exception $e) {
             if (self::isInstalled()) {
+                self::dispatchEvent('mage_installed_exception', ['exception' => $e]);
```
fballiano, line 754: @kiatng do we have anything against this? would a different name be better?

kiatng: image
If the first `if` block is removed, then there is already an event, which will be dispatched installed or not.

pquerner: I think the upper if (773 to 775) acts as a guard and is okay to stay.
Also because we immediately exit the program and not change "Location" to somewhere possibly outside of (current) context/control.

kiatng, line 754:
```suggestion
            try {
                self::dispatchEvent('mage_run_exception', ['exception' => $e]);
                if (!headers_sent() && !self::isInstalled()) {
```

You can do a test easily with `throw new Exception('test');` in any controller action. Not sure how to test installation but the check on `!isInstalled()` seems correct, without the not, it would redirect wrongly.

kiatng: I tested the following:

```php
           // remove the first if block lines 773-776
            try {
                self::dispatchEvent('mage_run_exception', ['exception' => $e]);
                if (!headers_sent() && !self::isInstalled()) {
               // ...
```

You can do a test easily with `throw new Exception('test');` in any controller action. Not sure how to test installation but the check on `!isInstalled()` seems correct, without the not, it would redirect wrongly.

fballiano:
@pquerner what do you think about this pr and its relation with https://github.com/OpenMage/magento-lts/pull/3957?

pquerner:
I think having 1 event as shown here is enough to even implement this new Error library or any library for that matter. I don't really see the need to have more than that event.

Also it makes sure that the exception caught can only be from cgi context, as the `::printException` can be called from different places, which ought to be handled either in the core or at the listener place (which I currently dont in the 1 module implementing)

fballiano:
not really because there's still the need to override the default error handler and that needs to be in the initEnvironment, so at least 2 events are needed.

I'll check if the event added in this pr could be used also in the context of Ignition and get back here.

@empiricompany

pquerner:
My sentry implementation uses `controller_action_predispatch` to change the default error handler. (cgi) and `default` for cli. 

The new event `core_app_init_environment_after` would make it more visible for cgi, I agree. But for cron this should not get executed if I read that code correctly.

fballiano:
we'll test the ignition module with a mix of your other PR and this one and see if it works, probably the mix of the two will be the best implementation ;-)

app/Mage.php
fballiano: I'm not sure about @kiatng, if it requires a modification of if it's ok as it is.

also, @pquerner do you think we could find a better name for the event? to me mage_installed_exception doesn't feel very clear, what about mage_run_exception_installed or mage_run_installed_exception, i'm not sure even about these ones...

pquerner: We have merged this in our core as event `mage_installed_run_exception` by now. I didnt touch this PR here and update yet again its name, because I sensed it would come up in a future discussion. ^^

Like I said, I like the guards in place as they currently are, but I can see why @kiatng wants it changed. I'd be fine with either.

fballiano: i'd kinda prefer mage_run_installed_exception cause keeps "mage_run" as "class_method" prefix, not that that's a standard.

in case of @kiatng's code, this new event wouldn't have any place to go right?

dunno... i guess it's a matter of tossing a coin?

pquerner: Would you expect a "printing" method to throw new events? To me thats a no. Although I could live with it. But its not really expected.

fballiano: mmm that's an interesting point

fballiano:
```suggestion
                self::dispatchEvent('mage_run_installed_exception', ['exception' => $e]);
```

docs/EVENTS.md
```diff
@@ -235,6 +235,7 @@
 | log_log_clean_after | 1.9.4.5 |
 | log_log_clean_before | 1.9.4.5 |
 | log_visitor_collection_load_before | 1.9.4.5 |
+| mage_installed_exception | TO_BE_DETERMINED |
```
fballiano, line 238:
```suggestion
| mage_run_installed_exception | 20.7.0 |
```

fballiano (approved):

fballiano:
since we kinda already had consensus, if nothing comes up I'll merge this tomorrow

PR #3615, opened 2023-10-27
New feature: added auto reply to contact form

kyrena:
### Description 

This PR allow admin to configure an email confirmation for the customer when a customer (registered or not) use the contact form.

Todo? remove changes from next branch (`Mage::throwException`, `_redirect`, ...) and add design to email template.

### Contribution checklist


ADDISON74:
This PR works as expected. 

**A few changes...**

1 - I would create a separate section called **Auto Reply**. It would look more elegant, in addition, there is an enable/disable option. See bellow

**BEFORE**

autoreply2


**AFTER**

autoreply

2 - One of the email templates must be identified and used for the auto reply, because messages sent to visitors have a default format (logo, ...). On this occasion, the contact template that the store receives should also be changed. It is ugly, a relic from the past. Of course, in other PR.

3 - The **subject** must be set as **Contact Form Confirmation**.

4 - The **message** in the body must be polite. E.g. "Thank you for contacting us. We will reply you shortly. Have a nice day."

fballiano:
@ADDISON74 I should have fixed all of your notes, from my tests the PR works as intended. I didn't work on point n.2 because I think it's a big topic for another PR (or more).

fballiano:
Since I changed a lot of stuff here, although I tested it, I won't put my green check on this, at the moment, and wait for other reviewers.

Flyingmana:
I think this needs a warning on the setting that an activated one can be used to spam email addresses.

maybe something for a future iteration(does not need to be part of this PR) a setting to limit it to logged in users, and making sure only the account email is used as sender

fballiano:
@Flyingmana very nice catch, I've added a warning message, which is absolutely necessary. At a later time we could enable ratelimiting for the contact us form.

app/locale/en_US/template/email/contact_form_auto_reply.html
```diff
@@ -0,0 +1,11 @@
...
+<!--@vars
+{"var data.name":"Sender Name",
+"var data.email":"Sender Email",
+"var data.telephone":"Sender Telephone",
+"var data.comment":"Comment"}
+@-->
+
+Thank you for contacting us, we will reply you shortly.
+<br>
+Have a nice day.
```
ADDISON74, line 11: I would evaluate if it is necessary to insert another paragraph for the situation when you receive the confirmation, without having used the contact form.

"In case you received this message without using the contact form on our website, please ignore it. Thank you for your understanding."

ADDISON74 (approved):

ADDISON74:
Please ignore my green stamp. I chose something else after making the comment in the code. I will check the new changes, at first glance they are fine.

@fballiano  - Did you make the change for point 1 as well?

fballiano:
> @fballiano - Did you make the change for point 1 as well?

yep

Screenshot 2024-02-16 alle 16 37 09

app/locale/en_US/template/email/contact_form_auto_reply.html
fballiano: @ADDISON74 I've also added this message to the template

ADDISON74:
The warning message does not resemble what already exists through the Backend. I will make you a proposal. 

The first word Warning should be in red.

fballiano:
I don't think it's possible to have html in that message (make it red)

ADDISON74:
"Auto Reply to Customer" is used incorrectly. When we set an auto reply, it is obvious that the message will be sent automatically to the person who contacts you. "Auto Reply" is enough in my opinion.

ADDISON74:
Html is possible, it needs to be wrapped into ```<![CDATA[]]>```

For example check this file `/app/code/core/Mage/Checkout/etc/system.xml`
https://github.com/OpenMage/magento-lts/pull/3836#issuecomment-1945787903

Adding "false" to docblock requires a check for not being false (e.g. "if ($param)" ... or something simiar)

fballiano:
> Html is possible, it needs to be wrapped into `<![CDATA[]]>`
> 
> For example check this file `/app/code/core/Mage/Checkout/etc/system.xml`

you live you learn :-D

fballiano:
> The warning message does not resemble what already exists through the Backend. I will make you a proposal.

I'll wait for this

app/code/core/Mage/Contacts/etc/system.xml
```diff
@@ -50,34 +50,62 @@
...
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
-                        </sender_email_identity>
+                        </email_template>
+                    </fields>
+                </email>
+                <auto_reply_to_customer translate="label">
+                    <label>Auto Reply to Customer</label>
+                    <sort_order>50</sort_order>
```
ADDISON74, line 83: "Auto Reply to Customer" is used incorrectly. When we set an auto reply, it is obvious that a message will be sent automatically to the person who contacts you. This person can be a customer, a potential customer or a simple visitor. 

"Auto Reply" is correct in my opinion. In addition, it is in the Contacts section and is self-explanatory

ADDISON74 (changes requested):

app/code/core/Mage/Contacts/etc/system.xml
```diff
@@ -50,34 +50,62 @@
...
+                    <show_in_default>1</show_in_default>
+                    <show_in_website>1</show_in_website>
+                    <show_in_store>1</show_in_store>
+                    <fields>
+                        <enabled translate="label comment">
+                            <label>Enable</label>
+                            <comment>Caution: auto reply can be abused to spam, ensure you have implemented necessary security measures (eg: enabling captcha) before proceeding.</comment>
```
ADDISON74, line 90: This is the comment I would use

```
<comment><![CDATA[<strong style="color:red">Warning!</strong> Enabling this feature may cause unwanted messages to be sent to people whose email address is used abusively. Please make sure that you have implemented security measures before activation (e.g. CAPTCHA, HoneySpam).]]></comment>
```

ADDISON74 (changes requested):

app/locale/en_US/template/email/contact_form_auto_reply.html
```diff
@@ -0,0 +1,13 @@
...
+"var data.email":"Sender Email",
+"var data.telephone":"Sender Telephone",
+"var data.comment":"Comment"}
+@-->
+
+Thank you for contacting us, we will reply you shortly.<br>
+Have a nice day.
+<br><br>
+In case you received this message without using the contact form on our website, please ignore it.<br>
+Thank you for your understanding.
```
ADDISON74, line 13: Thank you for your interest in the products sold by our company. This confirmation message is sent automatically as a result of your request made through the contact form. We will reply you shortly.

If someone else made the request on your behalf abusively, please ignore and delete our messages. Thank you for understanding.

ADDISON74 (changes requested):

app/code/core/Mage/Contacts/etc/system.xml
fballiano: done

app/locale/en_US/template/email/contact_form_auto_reply.html
fballiano: In my opinion this is way too much, at the moment I prefer the one that's already in the PR.

ADDISON74 (approved):

fballiano:
retested, works ok

PR #3619, opened 2023-10-29
Fixed deprecation warning if downloadable sample is a url

ma4nn:
### Description (*)
This PR fixes the error `Deprecated functionality: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated  in /app/htdocs/app/code/core/Mage/Downloadable/Helper/File.php on line 149`.
It occurs when a downloadable product is edited in backend that uses urls for the samples (as opposed to files).

1. Create a new downloadable product in backend
2. In tab "Downloadable Information" add a sample row with a URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -589,6 +589,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/leissbua"><img src="https://avatars.githubusercontent.com/u/68073221?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Michael Leiss</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://www.riseart.com/"><img src="https://avatars.githubusercontent.com/u/26821235?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Marcos Steverlynck</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/ahudock"><img src="https://avatars.githubusercontent.com/u/33500977?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Andy Hudock</b></sub></a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://www.vianetz.com/"><img src="https://avatars.githubusercontent.com/u/26252058?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Christoph Massmann</b></sub></a></td>
     </tr>
   </tbody>
 </table>
--- a/app/code/core/Mage/Downloadable/Helper/File.php
+++ b/app/code/core/Mage/Downloadable/Helper/File.php
@@ -124,11 +124,15 @@ class Mage_Downloadable_Helper_File extends Mage_Core_Helper_Abstract
      * Return full path to file
      *
      * @param string $path
-     * @param string $file
+     * @param string|null $file
      * @return string
      */
     public function getFilePath($path, $file)
     {
+        if ($file === null) {
+            return $path . DS;
+        }
+
         $file = $this->_prepareFileForPath($file);
 
         if (substr($file, 0, 1) == DS) {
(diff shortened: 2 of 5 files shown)


kiatng:
The method `getFilePath()` is called from many places, so I think it's better to fix at the source: https://github.com/OpenMage/magento-lts/blob/c7d6d554841a6fad29e99b25f2989b2704cb457c/app/code/core/Mage/Downloadable/Helper/File.php#L130-L130

```php
    public function getFilePath($path, $file)
    {
        if (!$file) {
            return $path . DS;
        }
// ...
```

ma4nn:
@kiatng thanks for your feedback. I didn't change that method intentionally for several reasons:

1. Lots of more tests would be required which makes that PR not that tiny any more (and in most of the other cases the parameter cannot be `null`)

1. I don't like those conditional arguments, imho it would be better to somehow later use Typehints and simply use `string $file` as parameter for that method (what kind of sense does a function `getFilePath()` make if no file is given?)

But I'm happy to change it if that's the way to go.

sreichel:
I think @kiatng is right. Before deprecation `null` was casted to empty string, ending up with `return $path . DS;`.

sreichel:
> 2\. I don't like those conditional arguments, imho it would be better to somehow later use Typehints and simply use `string $file` as parameter for that method

I agree in somehow later having typehints, but at this time we maybe have rector or phpstan who let you know about issues you mention.

In case that checks looks pretty safe. Even with typehints it avoids running str-functions on empty string.

app/code/core/Mage/Downloadable/Helper/File.php
```diff
@@ -124,11 +124,15 @@ protected function _moveFileFromTmp($baseTmpPath, $basePath, $file)
      * Return full path to file
      *
      * @param string $path
-     * @param string $file
+     * @param string|null $file
      * @return string
      */
     public function getFilePath($path, $file)
     {
+        if ($file === null) {
```
sreichel, line 132: `$file === null || $fille === ''`?

ma4nn: ok 🤷

README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-156-orange.svg" alt="All Contributors"></a>
+<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-157-orange.svg" alt="All Contributors"></a>
```
fballiano, line 3:
```suggestion
<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-156-orange.svg" alt="All Contributors"></a>
```
the format of this tag was changed

fballiano:
@ma4nn can you please check my last comment? I can't modify this PR because you didn't allow edits by admins. other than that I'd approve it

ma4nn:
@fballiano thanks I have rebased to `main` to resolve all conflicts.
I did activate modifications of the PR, but it seems like Github only allows this [for personal forks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). I will do future PRs from a personal fork then 👍

ADDISON74:
@ma4nn - Could you please check this PR https://github.com/sreichel/openmage-future/pull/8 which seems to be related to yours?

ma4nn:
@ADDISON74 I must admit the contribution process here in the OpenMage repo is really weired and not very encouraging 🙈  Why check another repo after I have updated this PR according to @fballiano? What do you mean with "check"? etc.

fballiano (approved):

ADDISON74:
@ma4nn - Thank you for your question and I assure you that you have no reason to worry. I personally appreciate any contribution. I noticed that your PR has a reference to another repository attached today. I looked over the other proposal, which is made by an important OpenMage contributor, who unfortunately no longer has the rights to post here and I can say that he has analyzed the issues discussed here. If it is a better PR then it should be in OpenMage, if the author agrees for his work to be brought here. If not, then I approve this PR after @kiatng expresses his opinion and we move on.

kiatng (approved):

PR #3634, opened 2023-11-08
Be more lenient in reading maintenance.ip addresses

loekvangool:
Currently, only commas are accepted to split IPs, and no spaces are allowed (only at the start/end of the file, not after the comma).

Event this input is currently invalid: `1.1.1.1, 1.1.1.2` because it will not trim the space.

This `preg_split()` allows splitting on spaces, commas, or new lines, such that these contents for `maintenance.ip` are valid:

```
1.1.1.1, 1.1.1.2  1.1.1.3
1.1.1.4


1.1.1.5,1.1.1.6
```

Result:
```php
array ( 0 => '1.1.1.1', 1 => '1.1.1.2', 2 => '1.1.1.3', 3 => '1.1.1.4', 4 => '1.1.1.5', 5 => '1.1.1.6', )
```


index.php
```diff
@@ -36,11 +36,12 @@
...
     $maintenanceBypass = false;
 
-    if (file_exists($maintenanceIpFile)) {
+    if (file_exists($maintenanceIpFile) && is_readable($maintenanceIpFile)) {
         /* if maintenanceFile and maintenanceIpFile are set use Mage to get remote IP (in order to respect remote_addr_headers xml config) */
         Mage::init($mageRunCode, $mageRunType);
         $currentIp = Mage::helper('core/http')->getRemoteAddr();
-        $allowedIps = explode(',', trim(file_get_contents($maintenanceIpFile)));
+        $allowedIps = preg_split('/[\ \n\,]+/', file_get_contents($maintenanceIpFile), 0, PREG_SPLIT_NO_EMPTY);
+        $allowedIps = array_map("trim", $allowedIps);
```
fballiano, line 44: do we need the trim since the split should remove double spaces?

sreichel, line 44:
```suggestion
	$allowedIps = preg_split('/[\ \n\t\,]+/', file_get_contents($maintenanceIpFile), 0, PREG_SPLIT_NO_EMPTY);
	foreach ($allowedIps as $index => $value) {
		if (!filter_var(trim($value), FILTER_VALIDATE_IP)) {
			unset($allowedIps[$index]);
		}
	}
```

Added tab-char.

loekvangool: Yup indeed `trim()` can be....🥁...trimmed.

loekvangool:
Made some changes. Didn't make much sense to `foreach...in_array()` it so turned it around a bit.

index.php
```diff
@@ -36,15 +36,16 @@
...
         Mage::init($mageRunCode, $mageRunType);
         $currentIp = Mage::helper('core/http')->getRemoteAddr();
-        $allowedIps = explode(',', trim(file_get_contents($maintenanceIpFile)));
-
-        if (in_array($currentIp, $allowedIps)) {
-            /* IP address matches, bypass maintenanceMode */
-            $maintenanceBypass = true;
+        $allowedIps = preg_split('/[\ \n\,]+/', file_get_contents($maintenanceIpFile), 0, PREG_SPLIT_NO_EMPTY);
+        foreach ($allowedIps as $allowedIp) {
+            if ($currentIp === $allowedIp && !filter_var($allowedIp, FILTER_VALIDATE_IP)) {
```
kiatng, line 45:
```suggestion
            if ($currentIp === $allowedIp && filter_var($allowedIp, FILTER_VALIDATE_IP)) {
```

However, I think the check for valid IP is unnecessary. If `$currentIp === $allowedIp` is true, it means `$allowedIp` must be a valid IP.  So I think the loop can be replaced with `in_array($currentIp, $allowedIps)`.

fballiano: I also think checking for valid IP is overkill, better to keep this code as small and fast as possible IMHO

loekvangool:
Agreed. Rewritten.

loekvangool:
@m-overlund anything to add for IPv6? I have no aspiration towards it tbh

m-overlund:
@loekvangool I'm not able to contribute sorry.
But the issue is, that with IPv6 you have privacy addresses, so, as i understand, you would have to be able to insert a subset of IPs like: `1234.1234.1234.0::/48`
Because your IP changes daily.

index.php
```diff
@@ -36,16 +36,12 @@
 if (file_exists($maintenanceFile)) {
     $maintenanceBypass = false;
 
-    if (file_exists($maintenanceIpFile)) {
-        /* if maintenanceFile and maintenanceIpFile are set use Mage to get remote IP (in order to respect remote_addr_headers xml config) */
+    if (file_exists($maintenanceIpFile) && is_readable($maintenanceIpFile)) {
```
kiatng, line 39: [PHP Doc](https://www.php.net/manual/en/function.is-readable.php):

> is_readable — Tells whether a file exists and is readable

```suggestion
    if (is_readable($maintenanceIpFile)) {
```

fballiano (approved):

sreichel:
Can we keep the comment?

fballiano:
the comment is still there but modified, how would you like that?

sreichel:
Either its too late, or too early :sunglasses: 

Missed that. Sry.

loekvangool:
> But the issue is, that with IPv6 you have privacy addresses, so, as i understand, you would have to be able to insert a subset of IPs like: `1234.1234.1234.0::/48`

Could be done in a follow up PR by introducing a lib like symfony ip utils.

kiatng (approved):

Issue #3625, opened 2023-11-01
A lot of exceptions after login in backend 


empiricompany:
### Reference PR:
The bug was introduced by this PR in loadRules(): an exception was added instead of this.
https://github.com/OpenMage/magento-lts/pull/2339/files#diff-1048571d4baf38d246708d5953af464948b5591c2c8d648036394d6101ba8022

### Possible solution
Rollback and remove `Mage::logException($e); ` in `app/code/core/Mage/Admin/Model/Resource/Acl.php::loadRules()`
I also recommend including the commented TODO code because it would be a feature to implement (the deletion of old resources),
that was this:
```
//$m = $e->getMessage();
//if ( eregi("^Resource '(.*)' not found", $m) ) {
    // Deleting non existent resource rule from rules table
    //$cond = $this->_write->quoteInto('resource_id = ?', $resource);
    //$this->_write->delete(Mage::getSingleton('core/resource')->getTableName('admin/rule'), $cond);
//} else {
    //TODO: We need to log such exceptions to somewhere like a system/errors.log
//}
```

in Mage_Admin_Model_Resource_Acl::loadRules() was added an exception 

### Preconditions (*)

1. Install OpenMage 20.2.0 - PHP 7.4 - MySQL 5.7
2. Install an extension that adds some resources to the admin_rule table (seems like the issue only occurs with deny permission resources).
![mageplaza_admin_rule](https://github.com/OpenMage/magento-lts/assets/5071467/75dcee40-2b85-4a4c-b268-985cfc773e3d)

3. Disable the previously installed extension from app/etc/modules/Vendor_Extension.xml.

### Steps to reproduce (*)

1. Login to the admin area.

### Expected result (*)

1. No exceptions should be logged for resources with deny permission.

### Actual result (*)

1. All resources with deny permission referring to the disabled extension have been logged with exceptions:

```
2023-11-01T08:44:04+00:00 ERR (3):
Zend_Acl_Exception: Resource 'admin/mageplaza_betterblog/new/action' not found in /var/www/html/htdocs/vendor/shardj/zf1-future/library/Zend/Acl.php:365
Stack trace:
#0 /var/www/html/htdocs/vendor/shardj/zf1-future/library/Zend/Acl.php(641): Zend_Acl->get()
#1 /var/www/html/htdocs/vendor/shardj/zf1-future/library/Zend/Acl.php(519): Zend_Acl->setRule()
[...]

PR #3642, opened 2023-11-10
Orphaned ACL resource exceptions are now logged only in developer mode

empiricompany:
### Description (*)
more details here:
https://github.com/OpenMage/magento-lts/issues/3625

Revert
https://github.com/OpenMage/magento-lts/pull/2339/files#diff-1048571d4baf38d246708d5953af464948b5591c2c8d648036394d6101ba8022

1. Fixes OpenMage/magento-lts#3625


sreichel:
This code is not complete.
Its not that easy to solve for disabled modules.

Better delete orphan records (and maybe add log otation, or patch it away?)

I see more pros then cons with logging.

empiricompany:
Sorry @sreichel this is not intended to resolve the orphan data but only to not log an exception every time i log in in the backend until we have a solution

Flyingmana:
I recommend to apply this change temporarily via the composer patcher if this causes you problems.
Iam against merging it, as logging this inconsistency is a good method the get aware of it.
And yes we should have a way to clean these up, maybe a script is ok?

The original code also had a (commented out) part in, which automatically cleans it up. Iam not a fan of this, but making it configurable would also be an option.

Would one of these options be acceptable?

fballiano:
I agree with @Flyingmana that logging is a good way to find and cleanup problems, and I agree with the sentiment of the original PR thus I'd leave it as it is, cleanup should be quite easy and there's https://github.com/OpenMage/magento-lts/pull/3647 on the way

empiricompany:
what if we wrap the exception in this if statement?

```
if (Mage::getIsDeveloperMode()) {
   Mage::logException($e);
}
```

Flyingmana:
> what if we wrap the exception in this if statement?
> 
> ```
> if (Mage::getIsDeveloperMode()) {
>    Mage::logException($e);
> }
> ```

acceptable for me

sreichel:
> what if we wrap the exception in this if statement?
> 
> ```
> if (Mage::getIsDeveloperMode()) {
>    Mage::logException($e);
> }
> ```

Perfect for me. It does not fill logs in production, but pops up in local env ... @Sdfendor

Sdfendor:
I agree, it's acceptable. 
The necessity for a cleanup/deletion mechanism for orphaned entries meanwhile stays the same.

m-overlund:
Perfect!

fballiano (approved):

m-overlund (approved):

ADDISON74 (approved):

Flyingmana:
can please someone provide a better title for the PR based on the new changes?

Flyingmana (approved):

empiricompany:
> can please someone provide a better title for the PR based on the new changes?

done

PR #3643, opened 2023-11-10
Get TinyMCE to 6.7.3  via composer :rocket: 

sreichel:
Update TinyMCE via composer/dependabot.

### Related Pull Requests

-  See OpenMage/magento-lts#3641

Note: composer plugin needs some rewrite ... `dev-master` for now.

---

From `composer audit`:

| Package           | tinymce/tinymce                                                                  |
| CVE               | CVE-2023-48219                                                                   |
| Title             | TinyMCE vulnerable to mutation Cross-site Scripting via special characters in unescaped text nodes    |
| URL               | https://github.com/advisories/GHSA-v626-r774-j7f8                                |
| Affected versions | >=6.0.0,<6.7.3|<5.10.9                                                           |
| Reported at       | 2023-11-15T18:32:34+00:00                                                        |


fballiano:
I was planning to work on something like this, the only thing I'd change is that I'd prefer the composer plugin to be owned by the openmage organization, if another enstablished standard one is not available.

sreichel:
Plugin can be moved to OM space for sure ... its just a fork. (tried all plugins, but non of them worked out of the box)

... but, it needs some rework before,

And needs write access ;)

sreichel:
You can fork it for backup?

PR #3644, opened 2023-11-11
Updated `.gitignore`

sreichel:
### Description (*)

- ignore `rector.php` config file for now
- ignore files prefixed with test
- `qodana.yaml` from phpstorm
- `.phpunit.result.cache`


.gitignore
```diff
@@ -80,3 +81,9 @@ phpstan*.neon
 # dev scripts loaded via composer
 /shell/update-copyright.php
 /shell/translations.php
+
+# test files
+**test*.php
```
kiatng, line 86: Sorry for my poor knowledge, but what does it match?

Flyingmana: if it only should match files prefixed with test, then I think there is a / missing and it should be `**/test*.php`

sreichel: @kiatng no default files, but in nearly every project hast some test files .;)

ADDISON74 (approved):

fballiano:
I understand where this comes from but I wouldn't add those extensions to the ignore because on my final projects I may want to have those files in the repo

sreichel:
I'd prefer "ignore all" first and only keep what belongs to core and have a customzed .gitignore in final project.

.gitignore
kiatng: It's true, I have them, just not name with the word "test'.

PR #3647, opened 2023-11-12
New feature: UI to list and remove orphaned ACL resources in backend.

kiatng:
### Description (*)
Orphaned ACL resources are left behind in the `admin_rule` table when modules are removed. This PR provides a UI to list these orphaned resources and delete them in backend > System > Permissions > Orphaned Resources.

### Related Pull Requests
See PR #2339 that log the orphaned resources.
See PR #3642 that revert #2339 to not cluttered exception.log.

1. Fixes OpenMage/magento-lts#3625

### Screenshots [edited]
At backend login:
image
Grid to list and delete:
image


sreichel:
Nice. Just did a visual review ... whats about disabled extensions?

(thats why ive nor completed commented code parts)

app/code/core/Mage/Admin/Model/Resource/Rules.php
```diff
@@ -80,4 +80,45 @@ public function saveRel(Mage_Admin_Model_Rules $rule)
...
+    }
+
+    /**
+     * Delete orphaned resources
+     *
+     * @param array $orphanedIds
+     * @return int
+     * @throws Mage_Core_Exception
+     */
+    public function deleteOrphanedResources(array $orphanedIds)
```
sreichel, line 103:
```suggestion
    public function deleteOrphanedResources(array $orphanedIds): int
```

app/code/core/Mage/Admin/Model/Resource/Rules.php
```diff
@@ -80,4 +80,45 @@ public function saveRel(Mage_Admin_Model_Rules $rule)
...
+    {
+        if (empty($orphanedIds)) {
```
sreichel, line 105:
```suggestion
        if ($orphanedIds === []) {
```

app/code/core/Mage/Admin/Model/Resource/Rules.php
```diff
@@ -80,4 +80,45 @@ public function saveRel(Mage_Admin_Model_Rules $rule)
...
+            return 0;
+        }
+
+        $resourceIds = Mage::getModel('admin/roles')->getResourcesList2D();
+        // Validate orphaned IDs are not in the list of valid resource IDs.
+        $validIds = array_intersect($orphanedIds, $resourceIds);
+        if (!empty($validIds)) {
```
sreichel, line 112:
```suggestion
        if ($validIds !== []) {
```

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
...
+        parent::__construct();
+        $this->setId('permissionsOrphanedResourceGrid');
+        $this->setDefaultSort('resource_id');
+        $this->setDefaultDir('asc');
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function _prepareCollection()
```
sreichel, line 34:
```suggestion
    protected function _prepareCollection(): static
```

Proposal ...

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
...
+        $resource->setResourceIdAsIdFieldName();
+
+        $this->setCollection($collection);
+        return parent::_prepareCollection();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function _prepareColumns()
```
sreichel, line 58:
```suggestion
    protected function _prepareColumns(): static
```

Proposal ...

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
...
+            'index' => 'resource_id'
+        ]);
+
+        return parent::_prepareColumns();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected function _prepareMassaction()
```
sreichel, line 71:
```suggestion
    protected function _prepareMassaction(): static
```

Proposal ...

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->getMassactionBlock()->addItem('delete', [
+            'label'    => Mage::helper('adminhtml')->__('Delete'),
+            'url'      => $this->getUrl('*/*/massDelete'),
+            'confirm'  => Mage::helper('adminhtml')->__('Are you sure you want to do this?')
+        ]);
+
+        return $this;
+    }
+
+    public function getRowUrl($row)
```
sreichel, line 85:
```suggestion
    public function getRowUrl($row): string
```

Proposal ...

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->_controller = 'permissions_orphanedResource';
+        $this->_headerText = Mage::helper('adminhtml')->__('Orphaned Resources');
+        parent::__construct();
+        $this->_removeButton('add');
+    }
+
+    /**
+     * @return string
+     */
+    protected function _toHtml()
```
sreichel, line 34:
```suggestion
    protected function _toHtml(): string
```

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource.php
```diff
@@ -0,0 +1,39 @@
+<?php
```
sreichel, line 1:
```suggestion
<?php

declare(strict_types=1)

```

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
+<?php
```
sreichel, line 1:
```suggestion
<?php

declare(strict_types=1)

```

app/code/core/Mage/Adminhtml/controllers/Permissions/OrphanedResourceController.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
sreichel, line 1:
```suggestion
<?php

declare(strict_types=1)

```

sreichel (commented):
Add strict types for new classes?

app/code/core/Mage/Admin/Model/Resource/Rules.php
kiatng: I like it.

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
kiatng: I have not seen this before, did a quick search, is this [PHP 8](https://php.watch/versions/8.0/static-return-type)?

sreichel: Shoul be valid for 7.4.

fballiano:
I think a n98-magerun action or a shell script would be enough because it's a one time maintainance thing, but good job anyway!

app/code/core/Mage/Admin/Model/Resource/Acl.php
```diff
@@ -131,9 +133,26 @@ public function loadRules(Mage_Admin_Model_Acl $acl, array $rulesArr)
                 } elseif ($rule['permission'] == 'deny') {
                     $acl->deny($role, $resource, $privileges, $assert);
                 }
+            } catch (Zend_Acl_Exception $e) {
+                if (
+                    !in_array($resource, $this->_orphanedResources)
+                    && strpos($e->getMessage(), "Resource '$resource' not found") !== false
+                ) {
```
fballiano, line 140:
```suggestion
                if (!in_array($resource, $this->_orphanedResources)
                    && strpos($e->getMessage(), "Resource '$resource' not found") !== false) {
```
phpcs

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2022-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 11:
```suggestion
 * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2022-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 11:
```suggestion
 * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```

luigifab:
@sreichel do you think that there are a lot of work with `find app/ shell/ lib/Mage/ lib/Magento/ lib/Varien/ -name "*.ph*" -type f -exec sed -i '1s/<?php/<?php declare(strict_types=1);/' {} ` ? (chatgpt powered)

💣

empiricompany:
Good job, but it would appear in all my clients' dashboards who are not developers and wouldn't know what to do. 
I agree with @fballiano that for these developer-related tasks, it is better to provide a utility in a shell script or n98-magerun.

**Question:** I haven't tested it yet, but what happens if I remove the permissions of a temporarily disabled extension? Will they be automatically recreated or will break the extension ?

Unfortunately, there is no system for uninstalling extensions. There are other things to consider, such as cleaning up the core_resource and core_config tables, and rolling back any database changes.
IMHO, I still believe it is better to remove the exception. Otherwise, every time I remove/disable an extension, the problem reoccurs.

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
kiatng: See [test](https://onlinephp.io/c/6d11a).. PHP7.4 throws
> Parse error: syntax error, unexpected 'static' (T_STATIC) in /home/user/scripts/code.php on line 3

app/code/core/Mage/Admin/Model/Resource/Acl.php
kiatng: How about  the doc on [PSR12](https://www.php-fig.org/psr/psr-12/):
image

Search for "multiple lines".

kiatng: [More info](https://en.wikipedia.org/wiki/PHP_Standard_Recommendation) on PSR.
> PSR-12	Extended Coding Style Guide: It extends, expands and replaces PSR-2, the coding style guide and requires adherence to PSR-1, the basic coding standard.[18]

kiatng: I'll stick with one-line, which will satisfies both PSR-2 and PSR-12.

kiatng:
> Add strict types for new classes?

I am not sure. Is it better to have an RFC on this? Or should we "sneak" this in? @Flyingmana what is your take?

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
sreichel: Mkay, `self` works for php74, but better leave it as it is.

sreichel:
We already have it in 

https://github.com/OpenMage/magento-lts/blob/16802cf42f4906eace4f710b7047e3126d124a83/app/code/core/Mage/Adminhtml/Helper/Config.php

May discuss it here #2748

> @sreichel do you think that there are a lot of work with

Yes, it is. Already work on it

kiatng:
> Nice. Just did a visual review ... whats about disabled extensions?

@sreichel So I performed 2 tests

Test 1. By disabling the modules `Nwdthemes_All` and `Nwdthemes_Revslider` in backend > System > Advanced. 
**Result 1:**
The notice `'The following resources are no longer available in the system: %s. You can delete them by <a href="%s">clicking here</a>.'` is **not rendered** at login to admin. There is **no record** in the grid.

Test 2. Removed the files in dir `app\etc\modules`: _Nwdthemes_All.xml_ and  _Nwdthemes_Revslider.xml_
**Result 2:** 
See the updated screenshots in description. The notice is rendered at login and the grid showed the orphaned resources.

> **Question:** I haven't tested it yet, but what happens if I remove the permissions of a temporarily disabled extension? Will they be automatically recreated or will break the extension ?

@empiricompany So I performed a test.

Test 3. Reinstate the files in dir `app\etc\modules`: _Nwdthemes_All.xml_ and  _Nwdthemes_Revslider.xml_. 
**Result 3:** 
Same as Result 1. No orphaned resources in grid. There is no orphaned resource to remove. This statement:

> Will they be automatically recreated or will break the extension ?

is not applicable.

kiatng:
@sreichel Thanks for your review, I incorporated all your suggestions except the return static type, which I think will break PHP7.4.

kiatng:
> I think a n98-magerun action or a shell script would be enough because it's a one time maintainance thing, but good job anyway!

@fballiano You are right, but I am not a fan of CLI. I am more comfortable with visual. This PR shows the orphaned resources, which before this, it's not that easy to know which ones are orphaned. I wrote this because I needed to update a couple of sites to v20.2 and encountered orphaned resources, e.g., poll was removed. 

> but it would appear in all my clients' dashboards who are not developers and wouldn't know what to do. I agree with @fballiano that for these developer-related tasks, it is better to provide a utility in a shell script or n98-magerun.

@empiricompany I was hoping you can test this PR and provide your feedback. Usually, a developer would remove a module. He may not know anything about orphaned resources. With this PR, he can check for it readily in the grid. The notice only appears once at login to admin. It may annoy other admin users but it will help the developer.

empiricompany:
@kiatng sorry but I didn't understand if from your tests, the removal of the ACL permissions corrupts the extension when it is re-enabled.

My personal opinion remains to move these utilities to n98-magerun, but if the majority agrees with this PR, we can find a compromise and display the messages only in developermode? as I mentioned in my PR: https://github.com/OpenMage/magento-lts/pull/3642

kiatng:
> @kiatng sorry but I didn't understand if from your tests, the removal of the ACL permissions corrupts the extension when it is re-enabled.

From my tests, the grid only shows orphaned resources which files are removed in dir `app\etc\modules`. It won't show resources disabled in backend. So, it **cannot** delete disabled resources since it is not shown. In case the code in this PR is called somewhere else to delete non-orphaned resources, it'll throw an exception `'The following resource(s) are not orphaned: %s'`.

If the modules files are removed in dir `app\etc\modules`, and the orphaned resources are removed, and then the modules are added back. In this case, it is just like installing a new module; its resources will be added back in.

kiatng:
>  we can find a compromise and display the messages only in developermode

My issue with this is that the notice is suppressed in production, which is not ideal. Alternative is to add a config flag.

empiricompany:
> If the modules files are removed in dir `app\etc\modules`, and the orphaned resources are removed, and then the modules are added back. In this case, it is just like installing a new module; its resources will be added back in. You'll need to set the ACL roles again for the module, as you would for a new install.

In this case, if the module is already installed because it is present in core_resource, are the ACL permissions automatically re-added? Or do we need to force the installation by removing the reference in core_resource and rolling back the database changes?
Sorry for the observation, but I wouldn't want to end up with "half-installed" extensions.

kiatng:
This PR is in productions. It works very well for me when updating OM to 20.3.

fballiano:
Screenshot 2024-01-31 alle 09 58 18

at a first glance it seems a bit unclear to me what this menu does, maybe "orphaned ACL resources"?

fballiano:
@ADDISON74 is "orphaned" the correct word here? cause i'd say something like "Orphan ACLs" but "orphaned" doesn't sound right to me.

ADDISON74:
That word has strange forms in certain phrases and can be either orphan or orphaned. For example, "orphan child" and "orphaned child" means the same thing, but according to Cambridge the orphaned version is more "comprehensive".

ADDISON74:
The link name in the menu is partially clear. ACL Resources / Orphaned ACL / Orphaned ACL Resources are a little bit better.

A short name must be found by referring to the other links in the menu which are shorter.

app/code/core/Mage/Adminhtml/etc/adminhtml.xml
```diff
@@ -145,6 +149,9 @@
                                     <blocks translate="title">
                                         <title>Blocks</title>
                                     </blocks>
+                                    <orphaned_resources translate="title">
+                                        <title>Orphaned Resources</title>
```
fballiano, line 153:
```suggestion
                                        <title>Orphaned ACLs</title>
```

what about this?

kiatng:
I actually chose the word _Resources_ from here: backend > System > Permissions > Roles > pick any role > Role Resources:
image

The resources listed are the ones orphaned and no longer listed here.

app/code/core/Mage/Adminhtml/etc/adminhtml.xml
kiatng: I changed it to "Orphaned Role Resources" to be consistent to existing label used in the core.

app/code/core/Mage/Adminhtml/controllers/Permissions/OrphanedResourceController.php
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
+ * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ */
+
+/**
+ * Class Mage_Adminhtml_Permissions_OrphanedResourceController
+ *
```
fballiano, line 20:
```suggestion
```

is this necessary?

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource.php
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 14:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Adminhtml/Block/Permissions/OrphanedResource/Grid.php
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Adminhtml
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 14:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

fballiano:
just some quick last minute formal fixes and I'll absolutely approve this

app/code/core/Mage/Adminhtml/controllers/Permissions/OrphanedResourceController.php
kiatng: Removed.

fballiano (approved):
I kinda prefer more verbose names expecially if they're more understandable :-)

empiricompany (approved):

fballiano:
another review and we're good to merge

empiricompany (approved):

tmewes (approved):

ADDISON74 (approved):
It works as described. Good job @kiatng.

PR #3648, opened 2023-11-12
Type casting and minor fixes in preparation for `declare strict_types=1`

luigifab:
### Description

This PR allow to add  `<?php declare(strict_types=1);`  in all files of OpenMage (@sreichel in #3647). This is only the beginning.

To add it in all files, run `find app/ shell/ lib/Mage/ lib/Magento/ lib/Varien/ -name "*.ph*" -type f -exec sed -i '1s/<?php/<?php declare(strict_types=1);/' {} +`

Apply this PR and enjoy backend.

To cancel changes: `find app/ shell/ lib/Mage/ lib/Magento/ lib/Varien/ -name "*.ph*" -type f -exec sed -i '1s/<?php declare(strict_types=1);/<?php/' {} +`

OpenMage 20.2.0, PHP 7.2 / 8.3.

### Contribution checklist


Flyingmana:
you need to be very careful with this, there are several situations where it changes the behavior of the code

luigifab:
Yes, I'm not sure to understand all possible impacts, but, for localhost it looks good!
And, with them, I found some critical fails of the life!

sreichel:
> there are several situations where it changes the behavior of the code

I did not see any changes in code behavoir during my tests. (not finished, but 80%) Either it works, or you see a fatal error. I think its not that hard to add strict types, its just **a lot of work** and manual testing. (unit tests would be nice here)

luigifab:
Looks good, perhaps extract enhancement for error message in another PR, and add `Mage::logException()`.
Else we can close it.

fballiano (dismissed):
I've read all the changes and, although some of them don't seem related to the PR to me (adding details on log exception) I think they could be merged, especially because this PR is not enforcing the strict_types, but it's only preparation work.

app/code/core/Mage/Core/Model/Layout.php
```diff
@@ -170,7 +170,7 @@ public function generateXml()
                             continue;
                         }
                         if (!isset($block->attributes()->ignore)) {
-                            $block->addAttribute('ignore', true);
+                            $block->addAttribute('ignore', 'true');
```
kiatng, line 173: Ref https://www.php.net/manual/en/language.types.declarations.php#:~:text=Strict%20typing%20¶,on%20a%20per-file%20basis:
> By default, PHP will coerce values of the wrong type into the expected scalar type declaration if possible. For example, a function that is given an int for a parameter that expects a string will get a variable of type string.

So `(string) true` is "1".
```suggestion
                            $block->addAttribute('ignore', '1');
```

app/code/core/Mage/Core/Model/Layout/Validator.php
```diff
@@ -143,6 +143,9 @@ public function isValid($value)
             try {
                 $value = new Varien_Simplexml_Element('<config>' . $value . '</config>');
             } catch (Exception $e) {
+                if (Mage::getIsDeveloperMode()) {
```
kiatng, line 146: This should be in a separate PR, lumping it here make it difficult to review because I'm not sure about the side-effects.

app/code/core/Mage/Customer/Model/Address/Config.php
```diff
@@ -99,9 +99,9 @@ public function getFormats()
             foreach ($this->getNode('formats')->children() as $typeCode => $typeConfig) {
                 $path = sprintf('%s%s', self::XML_PATH_ADDRESS_TEMPLATE, $typeCode);
                 $type = new Varien_Object();
-                $htmlEscape = strtolower($typeConfig->htmlEscape);
+                $htmlEscape = strtolower((string)$typeConfig->htmlEscape);
                 $htmlEscape = !($htmlEscape == 'false' || $htmlEscape == '0' || $htmlEscape == 'no'
-                    || !strlen($typeConfig->htmlEscape));
+                    || !strlen((string)$typeConfig->htmlEscape));
```
kiatng, line 104:
```suggestion
                    || !strlen($htmlEscape);
```

fballiano:
Agree with @kiatng, to speed things up I've removed from this PR the code that was not related to the strict_types thing.

app/code/core/Mage/Adminhtml/controllers/Cms/Wysiwyg/ImagesController.php
```diff
@@ -144,7 +144,7 @@ public function uploadAction()
             $targetPath = $this->getStorage()->getSession()->getCurrentPath();
             $result = $this->getStorage()->uploadFile($targetPath, $this->getRequest()->getParam('type'));
         } catch (Exception $e) {
-            $result = ['error' => $e->getMessage(), 'errorcode' => $e->getCode()];
+            $result = ['error' => true, 'message' => $e->getMessage()];
```
kiatng, line 147: This seems like a bug fix and should be in a separate PR.

app/code/core/Mage/Core/Model/Layout/Validator.php
```diff
@@ -233,9 +233,9 @@ public function validateTemplatePath(array $templatePaths)
         /** @var Varien_Simplexml_Element $path */
         foreach ($templatePaths as $path) {
             if ($path->hasChildren()) {
-                $path = stripcslashes(trim((string) $path->children(), '"'));
+                $path = stripcslashes(trim((string)$path->children(), '"'));
             }
-            if (strpos($path, '..' . DS) !== false) {
+            if (strpos((string)$path, '..' . DS) !== false) {
```
kiatng, line 238: Is this better?
```suggestion
            $path = $path->hasChildren()
                ? stripcslashes(trim((string)$path->children(), '"'))
                : (string)$path;
            if (strpos($path, '..' . DS) !== false) {
```
 For me it's easier to read, `$path` is typecast to string.

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
```diff
@@ -87,7 +87,7 @@ public function start($sessionName = null)
                 // no break
             default:
                 session_save_path($this->getSessionSavePath());
-                session_module_name($moduleName);
+                session_module_name((string)$moduleName);
```
kiatng, line 90: The param of `session_module_name()` can be type string or null. `$moduleName` type can be `Varien_Simplexml_Element|false`. I do not know how to test this. Can someone explain?

Also, `session_save_path($this->getSessionSavePath());` is same and should have the same treatment.

And these also:
line 80: `call_user_func($this->getSessionSavePath());`
line 81: `is_writable($this->getSessionSavePath())`

app/code/core/Mage/Customer/Model/Address/Config.php
```diff
@@ -101,7 +101,7 @@ public function getFormats()
                 $type = new Varien_Object();
                 $htmlEscape = strtolower((string)$typeConfig->htmlEscape);
                 $htmlEscape = !($htmlEscape == 'false' || $htmlEscape == '0' || $htmlEscape == 'no'
-                    || !strlen((string)$typeConfig->htmlEscape));
+                    || !strlen($htmlEscape);
```
ADDISON74, line 104: PHPStan reports an error. In line 104 there is a `)` missing before `;`

ADDISON74 (changes requested):

app/code/core/Mage/Customer/Model/Address/Config.php
fballiano:
```suggestion
                    || !strlen($htmlEscape));
```

app/code/core/Mage/Adminhtml/controllers/Cms/Wysiwyg/ImagesController.php
luigifab: Yes the original PR contains some minor bugfixes and addition for dev mode.

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
luigifab: Good point. I can't test anything at the moment.

app/code/core/Mage/Core/Model/Layout.php
luigifab: Why not.

app/code/core/Mage/Core/Model/Layout/Validator.php
fballiano: I'll commit @kiatng code but I'll convert the strpos to str_contains since we have the polyfills

app/code/core/Mage/Adminhtml/controllers/Cms/Wysiwyg/ImagesController.php
fballiano: I've removed this change to keep the PR clean

fballiano:
There's only one point open to fix for this PR.

app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
fballiano: @kiatng @luigifab I would remove the change to this file so that we can merge this PR, is it ok?

kiatng: Sure.

fballiano: done, i'll mark this conversation as resolved

fballiano (approved):

fballiano:
since all the comments and subsequent modification, I think this is mergeable

PR #3655, opened 2023-11-14
Fixed strip_tags(): Passing null to parameter #1 in Fulltext.php

kiatng:
### Description (*)
In the process of updating a couple of sites to PHP8.1. This was logged:

```
    [type] => 8192:E_DEPRECATED
    [message] => strip_tags(): Passing null to parameter #1 ($string) of type string is deprecated
    [file] => .../app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php
    [line] => 764
    [uri] => /index.php/admin/catalog_product/save/back/edit/tab/product_info_tabs_group_30/store/0/id/13641/key/debdde05d306ac8c1ac92215807e895d/
```

### DIFF
--- a/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php
+++ b/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php
@@ -761,7 +761,9 @@ class Mage_CatalogSearch_Model_Resource_Fulltext extends Mage_Core_Model_Resourc
             }
         }
 
-        $value = preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
+        if ($value !== null) {
+            $value = preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
+        }
 
         return $value;
     }
--- a/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php
+++ b/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php
@@ -761,9 +761,7 @@ class Mage_CatalogSearch_Model_Resource_Fulltext extends Mage_Core_Model_Resourc
             }
         }
 
-        if ($value !== null) {
-            $value = preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
-        }
+        $value = $value === null ? '' : preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
 
         return $value;
     }


Sdfendor (changes requested):
While this change solves the deprecation error it also changes the result of the code. 
See this following example: 
image

Currently in PHP 8.2 the deprecation is thrown but the built-in methods behave as before, so the result of `strip_tags(null)` is empty string, not null. Subsequently in cases where `$value` is null in line 764 it currently becomes empty string after this line is exexcuted. With your change `$value` stays null when returning in line 768. 

For BC sake and to avoid unforeseen consequences, I would therefore propose to set `$value` to empty string in the else branch of the condition: 
```php
if ($value !== null) {
	$value = preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
} else {
    $value = '';
}
```

kiatng:
I return `null` because that's what is being returned previously:
https://github.com/OpenMage/magento-lts/blob/bd00d61e833d6e9347d1a7001cb2573509424d88/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext.php#L732-L752

Sdfendor (approved):

m-overlund (approved):

elidrissidev (approved):

ADDISON74 (approved):

sreichel:
@kiatng i'd change it a bit to guarantee work with strict types turned on

https://onlinephp.io?s=PU_tCsIwDPy9Qt8hOKEbDnF__cAHUSmhxjnoZmlaQcR3t9ucEEhyyd0l-6O7OynyKxmLngoOvjVBh5cjPtTlTgoplk-0kRgOcOqjtRUEH6mCG1pOqU6x3lSganUZ128PT2juUMw8ZJjKEt5SZPnUJLmWNRtMvsU8P8J4Qd-Uv6UtKJVUsz_HeWq0J2fRULHIz7zKuY2L5A9quKztRgWnAzY865bDI9kTvb7Gzs1owj5SfAE%2C&v=8.2.12%2C8.1.25%2C8.0.30%2C7.4.33

simple: cast to string

```
	$value = preg_replace("#\s+#siu", ' ', trim(strip_tags((string)$value)));
```

or more strict:

```
	$value = is_scalar($value) ? (string)$value : '';
	$value = preg_replace("#\s+#siu", ' ', trim(strip_tags($value)));
```

sreichel:
I dont like casting to `(someting)` too, thats why i'd prefer the more strict one.

In case of scalar value casting is pretty safe.

sreichel:
> Taking a step back, is there a compelling reason to insert declare(strict_types=1) on existing classes?

- Without php does that type casting internally - and with a later version make it deprecated.
- Strict types ensure to pass correct type to internal php methods **and also validates `@method` annotations** !!!

  ```
  @method $this doSomething(<validates-type> $value)
  ```
I'm working on this, b/c it seems to be a large benefit for better code - even if we do not have typed parameters/returns.

PR #3656, opened 2023-11-14
Added `$product` param to `catalog_controller_product_init_before` event

kiatng:
### Description (*)
I am trying to change a product attribute based on some condition before rendering in frontend. I need to move the code a few lines below to add the product param:

image

It works well for me. Will it break something?


luigifab (approved):
I think that when `$productId` is undefined, the event is useless.
So I think it's not bad, but not tested.

sreichel:
Isnt there any other event? Eg `catalog_product_load_before`?

empiricompany:
this is a delicate matter

kiatng:
> Isnt there any other event? Eg `catalog_product_load_before`?

Yes, I considered it. In my case, I do not want to save the "dynamic" attributes, either in backend or frontend (I allow saving by frontend users). So, when product is loaded, I need to know where is it loaded and for what purpose. This event is clean and the purpose is frontend rendering.

fballiano:
I'm sure you thought about it but the catalog_controller_product_init doesn't work for you? it's just a few lines down

kiatng:
> I'm sure you thought about it but the catalog_controller_product_init doesn't work for you? it's just a few lines down

I need to do something about category `$categoryId = $params->getCategoryId();`. Event `catalog_controller_product_init` is too late.

fballiano (approved):

PR #3658, opened 2023-11-16
Updated TinyMCE to 6.8 via composer 🚀

sreichel:
Update TinyMCE via composer/dependabot.

(something went wrong in #3643, so new clean (?) PR)

### Related Pull Requests

-  See OpenMage/magento-lts#3641

---

From `composer audit`:

| Package           | tinymce/tinymce                                                                  |
| CVE               | CVE-2023-48219                                                                   |
| Title             | TinyMCE vulnerable to mutation Cross-site Scripting via special characters in unescaped text nodes    |
| URL               | https://github.com/advisories/GHSA-v626-r774-j7f8                                |
| Affected versions | >=6.0.0,<6.7.3|<5.10.9                                                           |
| Reported at       | 2023-11-15T18:32:34+00:00                                                        |


fballiano:
When I do a composer install, then I can't find the lib/tinymce folder:

Screenshot 2023-11-16 alle 09 57 50


am I doing something wrong?

fballiano:
tested and it works, notes:
- I still think that the new plugin should be under the openmage organization (but that can be done a a later stage)
- it seems to me that js/tinymce/langs folder is missing (it contains the translations for the editor)

sreichel:
`js/tinymce/langs` is not part of tinymce repo. Where does it comes from?

fballiano:
https://www.tiny.cloud/get-tiny/language-packages/

sreichel:
Added https://github.com/mklkj/tinymce-i18n that uses that :)

fballiano:
nice find

fballiano (approved):
tested, works for me and it is a good improvement

fballiano:
because of https://github.com/advisories/GHSA-v626-r774-j7f8 it would be good if somebody could review this PR, we'd need to release it

@kiatng @ADDISON74 @elidrissidev @Flyingmana @colinmollenhour

sreichel:
Only tested with git install!

If it does not work inside composer, i'v prepared a wrapper inside a magento-module. (or change plugin?)

Cons:

It needs to be updated hisself.

fballiano:
do you know for sure that it doesn't work with composer based projects? I can't find a way to test that case before it's merged

sreichel:
> do you know for sure that it doesn't work with composer based projects?

No, just an idea - i also could not test it.

May setup a test-scenario, but this would take me some days.

fballiano:
better to merge and test on dev-main, extremely easier

composer.json
```diff
@@ -26,13 +26,16 @@
...
         "pelago/emogrifier": "^7.0",
         "phpseclib/mcrypt_compat": "^2.0.3",
         "phpseclib/phpseclib": "^3.0.14",
         "shardj/zf1-future": "1.23.5",
+        "sreichel/composer-plugin-file-copy": "^1.0",
         "symfony/polyfill-php74": "^1.27",
         "symfony/polyfill-php80": "^1.27",
-        "symfony/polyfill-php81": "^1.27"
+        "symfony/polyfill-php81": "^1.27",
+        "tinymce/tinymce": "^6.7.3"
```
fballiano, line 38:
```suggestion
        "tinymce/tinymce": "^6.8"
```

sreichel:
@fballiano can you please update composer.lock?

(i have an complete different composer.json right now and cant update in time)

fballiano:
done!

fballiano (dismissed):

ADDISON74:
That vulnerability requires us to speed up how TinyMCE is updated. I haven't tested this PR yet, but I noticed that besides the composer files, the rest are just deleted files. Do the custom changes made by us in previous PR's still remain (those related to the menus, ...)?

fballiano:
@ADDISON74 yes

fballiano:
no need IMHO, it will be a normal release like all of the others, the vulnerability will be mentioned in the release notes ;-)

composer.json
```diff
@@ -26,13 +26,16 @@
         "colinmollenhour/magento-redis-session": "^3.2.0",
         "cweagans/composer-patches": "^1.7",
         "magento-hackathon/magento-composer-installer": "^3.1 || ^2.1 || ^4.0",
+        "mklkj/tinymce-i18n": "^23.11",
```
fballiano, line 29:
```suggestion
        "mklkj/tinymce-i18n": "^23",
```

fballiano (approved):

composer.json
sreichel: Why allow older versions? But bump it to 23.12?

sreichel:
> That vulnerability requires us to speed up how TinyMCE is updated.

Appove?

If composer install does not work, revert it.

ADDISON74 (approved):

fballiano:
tested on a composer based project (using "openmage/magento-lts": "dev-main") but it doesn't work, @sreichel maybe it's just the "copy" plugin that doesn't work? could you check it?

sreichel:
Mhh. Not 100% about this, I guess plugins work only on top-level composer. Maybe its easier to create a magento-module that updates tinymce via dependabot?

(or add install intructions ... ?)

fballiano:
also if I add the plugin in my main composer.json (the project's one) it doesn't work :-(

sreichel:
@fballiano will test it later. Until ... can you give it a last try with verbose `vvv` and see if the files maby copied to wrong location?

sreichel:
>we'd need to have the extra.file-copy in the top-level composer.json?

Seems so :(

Actually `extra.` config is read from root-composer only. No time to debug yet, but found this ...

https://github.com/wikimedia/composer-merge-plugin/blob/master/README.md#merge-extra ....

fballiano:
but it's another plugin, nah at this point it's kinda too many plugins IMHO.

sreichel:
Everything is better then manual file updates :(

Better to work on a `Mage_TineyMce` module?

sreichel:
Lets revert for now. :(

fballiano:
I think it should be implemented in the magento composer plugin but... ye...

sreichel:
If that plugin works, why not ... ? Same logic hat to be adapted to magento-composer-plugin.

fballiano:
because everybody would automatically (kinda) update to a new version of the magento-composer-plugin, but it's more difficult to make them install a new one (in their composer file)

ADDISON74:
I can't help but notice what a nice collaboration there was between contributors. It happened 9 months ago...

PR #3662, opened 2023-11-18
Rewrote `js/mage/translate.js` without prototypejs

fballiano:
needs extensive testing using language pack and checking that the frontend works as expected


pquerner:
What about ES5?

kiatng:
Yes!  #3645 has started.

luigifab:
With ES6 syntax, we will instant crash all old browsers that does not understand this syntax.
With 1% of the file in ES6, even if 99% of the file is in ES5, the whole file is crashed.

https://caniuse.com/es6

fballiano:
sure but es6 support is 98.31%... should we not use it because of opera mini and IE11? then let's move everything to the next branch

kiatng:
> sure but es6 support is 98.31%... should we not use it because of opera mini and IE11? then let's move everything to the next branch

`next` branch won't change anything on IE11. [Opera Mini](https://en.wikipedia.org/wiki/Opera_Mini) is only for android phones, which has a default chrome browser. So, it doesn't make sense to put this in `next`.

pquerner:
I think some other ticked mentioned transpiling.
In the core we can write in TypeScript and transpile it to ES5 or/and to ES6.

Or both and a feature flag decides whether or not to load the ES6 file.

I know that in our frontend no ES6 is allowed (sadly), since we have (still) lots of hits on older devices. And we don't want to break them (just yet?).

kiatng:
On @pquerner note, I vote for ES6 in `next` branch.

fballiano:
rebased on next

pquerner:
By all means, I don't want to keep OM "behind" with JavaScript usage. One can always just fork the release and manually patch things, or revert commits. Just wanted to give my 2 cts. :)

elidrissidev:
Pretty sure I saw ES6 syntax in a merged PR before, so it might be too late to discuss that 😅 I say just keep using ES6 it's almost 2024!

kiatng:
So, now what? Do we convert ES6 to ES5 in v20? Or do we carry on with ES6 in v20?

kiatng (dismissed):
Tested in backend browser console:
```
> Translator.translate('Please wait, loading...')
< 'Please wait, loading...'
> Translator.add('abc', 'xyz')
< undefined
> Translator.translate('abc')
< 'xyz'
> Translator.add({'foo': 'bar'})
< undefined
> Translator.translate('foo')
< 'bar'
```

js/mage/translate.js
```diff
@@ -8,32 +8,29 @@
  * @category    Mage
  * @package     js
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
- * @copyright   Copyright (c) 2022 The OpenMage Contributors (https://www.openmage.org)
+ * @copyright   Copyright (c) 2022-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 11:
```suggestion
 * @copyright   Copyright (c) 2022-2024 The OpenMage Contributors (https://www.openmage.org)
```

kiatng (approved):

empiricompany (approved):

ADDISON74 (approved):

PR #3663, opened 2023-11-19
Fixing warning about providing null instead of an expected string

alexh-swdev:
### Description (*)
Fixing  a null-vs-string deprecation:
2023-11-18T17:14:09+00:00 ERR (3): strlen(): Passing null to parameter #1 ($string) of type string is deprecated in /..../lib/Unserialize/Reader/Str.php on line 69

### DIFF
--- a/lib/Unserialize/Reader/Str.php
+++ b/lib/Unserialize/Reader/Str.php
@@ -66,7 +66,7 @@ class Unserialize_Reader_Str
         }
 
         if ($this->_status == self::READING_VALUE) {
-            $val = is_null($this->_value) ? "" : $this->value;
+            $val = is_null($this->_value) ? "" : $this->_value;
             if (strlen($val) < $this->_length) {
                 $this->_value .= $char;
                 return null;
--- a/lib/Unserialize/Reader/Str.php
+++ b/lib/Unserialize/Reader/Str.php
@@ -66,14 +66,15 @@ class Unserialize_Reader_Str
         }
 
         if ($this->_status == self::READING_VALUE) {
-            if (strlen($this->_value) < $this->_length) {
+            $val = is_null($this->_value) ? "" : $this->value;
+            if (strlen($val) < $this->_length) {
                 $this->_value .= $char;
                 return null;
             }
 
-            if (strlen($this->_value) == $this->_length) {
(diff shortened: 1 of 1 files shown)


lib/Unserialize/Reader/Str.php
```diff
@@ -66,14 +66,15 @@ public function read($char, $prevChar)
...
                 return null;
             }
 
-            if (strlen($this->_value) == $this->_length) {
+            if (strlen($val) == $this->_length) {
                 if ($char == Unserialize_Parser::SYMBOL_SEMICOLON && $prevChar == Unserialize_Parser::SYMBOL_QUOTE) {
-                    return (string)$this->_value;
+                    return (string)$val;
                 }
             }
```
kiatng, line 79: Note that the scope of `$_value` is at class level, you assign its value to a temporary var `$val` of local scope,  I do not know if it's necessary to have `$this->_value = $val` at line 77 before exiting.  Best to avoid the temp var all together: `if ($this->_value === null) $this->_value = '';` Or something like the following:
```php
            if ($this->_value === null) {
                $this->_value = $char;
                return null;
            }
            
            if (strlen($this->_value) < $this->_length) {
                $this->_value .= $char;
                return null;
            }

            if (strlen($this->_value) == $this->_length) {
                if ($char == Unserialize_Parser::SYMBOL_SEMICOLON && $prevChar == Unserialize_Parser::SYMBOL_QUOTE) {
                    return (string)$this->_value;
                }
            }
```
What do you think?

kiatng: Also, may be just declare like this `protected $_value = ''` is enough. Can you check?

alexh-swdev: I'll check, yes.
Exactly because it's a class field I used the local var to leave it alone. I was unsure if someone relied on the null-value. Probably overthinking.
I like your solution with assining the `$char` directly in case of null. I'll adjust it later.

kiatng (approved):

elidrissidev (approved):

fballiano:
@alexh-swdev @kiatng @elidrissidev this PR is causing https://github.com/OpenMage/magento-lts/discussions/3908

PR #3669, opened 2023-11-20
Rewrote `js/varien/accordion.js` without prototypejs

fballiano:
A quick test can be done going to the checkout, switching between the various steps of the checkout page uses this Accordion class.


midlan (approved):
not tested on my instance, just checked the code and it looks good to me

pquerner:
I wonder what happens when somebody overridden that prototyped function. Does it still work afterwards? Or does the new OpenMage version give a big hint in the release notes regarding these files/prototype removal?

fballiano:
there not much in the readme now, this PR is for the next branch.

kiatng (approved):

Issue #3672, opened 2023-11-22
`Undefined array key 0` warning at console Installation


akunzai:
### Preconditions (*)

1. OpenMage 20.2.0
2. PHP 8.2

See my [gist](https://gist.github.com/akunzai/8a14d7ee7b7a567f1d424f612477f08c)

### Steps to reproduce (*)

```sh
# start containers
docker compose up -d

# install OpenMage
bash ./install.sh
```

### Expected result (*)

Install success without errors

```sh
Checking database...
Installing OpenMage...
SUCCESS: ef6481244432fb89d109db9017743b1e
```

### Actual result (*)

```sh
Checking database...
Installing OpenMage...
ERROR: Error in file: "/var/www/html/app/code/core/Mage/Customer/sql/customer_setup/mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php" - Warning: Undefined array key 0  in /var/www/html/app/code/core/Mage/Core/Model/App.php on line 1006
```

Workaround: change `php.ini` to ignore warnings

```ini
error_reporting=E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED & ~E_WARNING
```

I wrote some simple code to reproduce exception at [App.php](https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Core/Model/App.php#L1006).

```php
<?php
function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
{
  $errno = $errno & error_reporting();
  if ($errno == 0) {
    return false;
  }
  throw new Exception($errstr);
}
set_error_handler('mageCoreErrorHandler');
error_reporting(E_ALL);
$website = 1;
$websites = [];
$id = false;
$code = null;
$websites[$code] = $website;
echo $websites[$code], "\n"; # [''] => 1
echo $websites[$id], "\n"; # [0] => Undefined array key 0
print_r($websites);
```

fballiano:
I've tested a clean installation without seeing your problem, can you provide a install.php script with all the parameters for a quick check?

sreichel:
I remember this error.

Can you test https://github.com/OpenMage/magento-lts/pull/3248 or install via n98-magerun?

akunzai:
@fballiano https://gist.github.com/akunzai/8a14d7ee7b7a567f1d424f612477f08c#file-install-php

sreichel:
@akunzai its nothing about your env ...

[...]

PR #3677, opened 2023-11-23
Fixed warning `Undefined array key 0` when installing OM via command line (dev mode on) (#3672)

akunzai:
### Description (*)

Handle posible false value in [App::getWebsite](https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Core/Model/App.php#L1006) from [Store::getWebsite](https://github.com/OpenMage/magento-lts/blob/1151717c034567f31722eb5d4666e0949977f80b/app/code/core/Mage/Core/Model/Store.php#L457) at [mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php](https://github.com/OpenMage/magento-lts/blob/1151717c034567f31722eb5d4666e0949977f80b/app/code/core/Mage/Customer/sql/customer_setup/mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php#L329)

1. Fixes OpenMage/magento-lts#3672

1. create docker env from this [gist](https://gist.github.com/akunzai/8a14d7ee7b7a567f1d424f612477f08c)
2. run the containers
3. change the source code in OpenMage container as this PR
4. run the `install.sh` from gist to install OpenMage


sreichel:
Not sure if its the best way, but i have no better idea right now. Needs some investigation.

fballiano (dismissed):
fine to me

sreichel:
false leads to another exception in behind.

I run into same and did not fix it that way. At least add a "todo" annotation.

Sekiphp:
This can provide unexpectable behavior of application. In ideal case method should throw Exception, that website not exists and fix places, where the error is provided.
`
I do not think, that if you call `Mage::app()->getWebsite(false)` you should get current website instance.

midlan:
Why not just fix the file `mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php`?

akunzai:
Thanks for everyone's advice. I made some changes to this PR, and I think that `null` should be enough to throw exceptions and clarify that the website does not exist!

midlan:
@akunzai I still believe we shouldn't modify one of the most commonly used method in code. Instead, let's focus on fixing the specific file where you discovered the issue.

app/code/core/Mage/Core/Model/Store.php
```diff
@@ -450,12 +450,12 @@ public function setWebsite(Mage_Core_Model_Website $website)
...
      * Retrieve store website
      *
-     * @return Mage_Core_Model_Website|false
+     * @return Mage_Core_Model_Website|null
      */
     public function getWebsite()
     {
         if (is_null($this->getWebsiteId())) {
-            return false;
+            return null;
```
midlan, line 458: Backcompatibility breaking change. This can cause huge problems inside openmage itself or many 3rd party modules. From my view, just fix the file `app/code/core/Mage/Customer/sql/customer_setup/mysql4-data-upgrade-1.4.0.0.7-1.4.0.0.8.php` alone

midlan (changes requested):

app/code/core/Mage/Core/Model/Store.php
akunzai: Ok, thanks for your advice.

app/code/core/Mage/Core/Model/Store.php
```diff
@@ -364,7 +364,7 @@ public function getConfig($path)
     public function initConfigCache()
     {
         /**
-         * Funtionality related with config separation
+         * Functionally related with config separation
```
fballiano, line 367:
```suggestion
         * Functionality related with config separation
```

I guess we'll never know the original idea of the author :-)

midlan (approved):
current solution is ok for me, approving

fballiano:
@akunzai silly question, did you test the current implementation of this PR?

akunzai:
@fballiano Yes, I tested the CLI installation via with current implementation.

sreichel:
It "works", but seems wrong as @Sekiphp pointed out.

akunzai:
Please explain your point, I didn't get it. or fire another PR to fix it?

fballiano:
@akunzai is the latest implementation of this PR tested?

@everybody Since this PR patches the install file, it seems to be ok to me and I'd merge it, anything against it?

sreichel:
> I do not think, that if you call `Mage::app()->getWebsite(false)` you should get current website instance.

This. But i had no better idea so far.

fballiano:
> > I do not think, that if you call `Mage::app()->getWebsite(false)` you should get current website instance.
> 
> This. But i had no better idea so far.

@sreichel in that case it could be fixed in a different place but could probably be another PR right?

sreichel:
I really dont know ... as said, i faced same issue and thought about to fix it that way, but it leads to another exception in background (needs to turn dev-mode on, or xedub on ... )

imho ... this need more investigation

sreichel:
I cant repoduce that from scratch ... 

If frequently run ddev install command (finally same as install.sh), but have no trouble. (beside to flush var/cache)

akunzai:
@fballiano 

The console installation success without any errors with the latest implementation of this PR.

CleanShot 2024-02-03 at 23 00 59@2x

sreichel:
@akunzai this error only occurs with dev-mode on.

You can set the environment variable to 0 in CLI too.

This "fix" seems wrong to me. (otherwise I'd have already made a pr)

akunzai:
@sreichel

1. In my [setup](https://gist.github.com/akunzai/8a14d7ee7b7a567f1d424f612477f08c), I have already turn on the dev-mode by set `MAGE_IS_DEVELOPER_MODE=1`, Are we using the same environment?
2. Can you provide the error details (Include the file and line number information)?

fballiano:
@akunzai I'm testing this PR now but I get this error:

`ERROR: Error in file: "openmage/app/code/core/Mage/Sales/sql/sales_setup/upgrade-1.6.0.3-1.6.0.4.php" - Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /Users/fab/Projects/openmage/lib/Varien/Db/Adapter/Pdo/Mysql.php on line 1783`

fballiano (approved):

fballiano:
I don't think it's that useful without fixing https://github.com/OpenMage/magento-lts/pull/3677#issuecomment-1928413416 (and probably more that would appear) but ok

PR #3686, opened 2023-12-03
New feature: ConfigurableSwatches now allows for auto-generation of the swatch image file based on color selection

empiricompany:
### Description (*)

I avoid installing third-party extensions to manage swatches that often need to be integrated into the template and cause issues and mess up the database.
But I also hate explaining to the client how to create and how to upload swatch images.
With this modification, it's possible to choose a color that will be used for dynamically creating the swatch.

The code only comes into play if a fallback image is not present in `media/wysiwyg/swatches/optionvalue.png`.
A resized image is created in cache directly to avoid conflicts with the original images in:
`media/catalog/swatches/STOREID/WidthxHeight/optionvalue.png`.
This way, it is perfectly compatible with the other modes that always take precedence.

demo
demo_picker
demo_delete

1. Configure color to be displayed as a swatch
2. Create a new configurable product with the `color` attribute
3. Create a new simple product with a color option that does not have a fallback image in `media/wysiwyg/swatches`

( If everything is correct, should display a broken image in product view and layered )

4. Navigate to Manage Attributes, edit the color attribute, and select a color for the option with the color picker

## Results:
A new swatch image is created on the fly directly in the cache `media/catalog/swatches/` foreach requested dimensions, both for the product view and layered navigation filters.

### Questions or comments

This feature is heavly inspired to magento 2 and can be extended to allow direct upload of swatch images in
`media/wysywyg/swatches`.
Use `<input type="color" />` for color picker that seems is widely supported in Chrome and Firefox https://caniuse.com/input-color

~~**It is not possible to set a default empty value for input type="color", so all colors are black by default, which is always better than a broken image (for now😉)**~~

I had to update the EAV module by adding a new table `eav_attribute_option_swatch` because it seem to me more consistent with the database schema. If you have any alternative solutions, please let me know.
I have tested also with the `Mage_ConfigurableSwatches` module disabled.

**Please comment only on the PR implementation; for anything else, it's best to open a separate discussion.**


fballiano:
great idea as usual, I'm not super familiar with the swatches, isn't it possible to generate a div with a css color background instead of having to create an image with a solid color? for sure you thought about it but I'm wondering, since it would avoid having files on the filesystem for a solid color.

empiricompany:
> great idea as usual, I'm not super familiar with the swatches, isn't it possible to generate a div with a css color background instead of having to create an image with a solid color? for sure you thought about it but I'm wondering, since it would avoid having files on the filesystem for a solid color.

Yes, it would be better to create the swatch using CSS or an SVG, but this solution was designed to be as backward compatible as possible with other themes that inherit the same templates and are not modified by this PR.

ADDISON74:
I flushed the cache before. I checkout this PR and I am getting the following error:

```php
Error in file: "/var/www/html/app/code/core/Mage/Eav/sql/eav_setup/upgrade-1.6.0.1-1.6.0.2.php" - SQLSTATE[HY000]: General error: 1005 Can't create table `db`.`eav_attribute_option_swatch` (errno: 121 "Duplicate key on write or update"), query was: CREATE TABLE `eav_attribute_option_swatch` (
  `value_id` int UNSIGNED NOT NULL auto_increment COMMENT 'Value Id' ,
  `option_id` int UNSIGNED NOT NULL default '0' COMMENT 'Option Id' ,
  `value` varchar(255) NULL default NULL COMMENT 'Value' ,
  `filename` varchar(255) NULL default NULL COMMENT 'Filename' ,
  PRIMARY KEY (`value_id`),
  UNIQUE `IDX_EAV_ATTRIBUTE_OPTION_VALUE_OPTION_ID` (`option_id`),
  CONSTRAINT `FK_EAV_ATTR_OPT_VAL_OPT_ID_EAV_ATTR_OPT_OPT_ID` FOREIGN KEY (`option_id`) REFERENCES `eav_attribute_option` (`option_id`) ON DELETE CASCADE ON UPDATE CASCADE
) COMMENT='Eav Attribute Option Swatch' ENGINE=INNODB charset=utf8 COLLATE=utf8_general_ci

#0 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(628): Mage::exception()
#1 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(424): Mage_Core_Model_Resource_Setup->_modifyResourceDb()
#2 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(308): Mage_Core_Model_Resource_Setup->_upgradeResourceDb()
#3 /var/www/html/app/code/core/Mage/Core/Model/Resource/Setup.php(223): Mage_Core_Model_Resource_Setup->applyUpdates()
#4 /var/www/html/app/code/core/Mage/Core/Model/App.php(441): Mage_Core_Model_Resource_Setup::applyAllUpdates()
#5 /var/www/html/app/code/core/Mage/Core/Model/App.php(347): Mage_Core_Model_App->_initModules()
#6 /var/www/html/app/Mage.php(740): Mage_Core_Model_App->run()
#7 /var/www/html/index.php(56): Mage::run()
#8 {main}
```

ADDISON74:
I ran the query below directly in the database and got the same error. I removed completely the CONSTRAIN line and I managed to create the table. It is not right, but there are some issues on that line.

```sql
CREATE TABLE `eav_attribute_option_swatch` (
  `value_id` int UNSIGNED NOT NULL auto_increment COMMENT 'Value Id' ,
  `option_id` int UNSIGNED NOT NULL default '0' COMMENT 'Option Id' ,
  `value` varchar(255) NULL default NULL COMMENT 'Value' ,
  `filename` varchar(255) NULL default NULL COMMENT 'Filename' ,
  PRIMARY KEY (`value_id`),
  UNIQUE `IDX_EAV_ATTRIBUTE_OPTION_VALUE_OPTION_ID` (`option_id`),
  CONSTRAINT `FK_EAV_ATTR_OPT_VAL_OPT_ID_EAV_ATTR_OPT_OPT_ID` FOREIGN KEY (`option_id`) REFERENCES `eav_attribute_option` (`option_id`) ON DELETE CASCADE ON UPDATE CASCADE
) COMMENT='Eav Attribute Option Swatch' ENGINE=INNODB charset=utf8 COLLATE=utf8_general_ci
```

ADDISON74:
At the first evaluation, you have to be careful with the initial values, because if there are already options with values, a black swatch is associated for all of them. In the Magento Sample Pack in size and color attributes, I am getting a black swatch for all options.

screen

Not all attributes should have swatches. Also, the options of an attribute do not have to get a value. There are situations when you may not want it to get a swatch, for example in size case, I don't need swatches.

What is the priority when I set an image on the product edit page to be a swatch. Based on this custom code, it takes the global one defined in attribute page or the one in the product image tab?

ADDISON74:
If I change the name of a color in the attribute options page (e.g. from Blue to Blue Print), in the /media/wysiwyg/swatches directory, there is not PNG file with the new name. The fallback should work but I didn't find any directory in /media/catalog/swatches.

empiricompany:
> I ran the query below directly in the database and got the same error. I removed completely the CONSTRAIN line and I managed to create the table. It is not right, but there are some issues on that line.
> 
> ```sql
> CREATE TABLE `eav_attribute_option_swatch` (
>   `value_id` int UNSIGNED NOT NULL auto_increment COMMENT 'Value Id' ,
>   `option_id` int UNSIGNED NOT NULL default '0' COMMENT 'Option Id' ,
>   `value` varchar(255) NULL default NULL COMMENT 'Value' ,
>   `filename` varchar(255) NULL default NULL COMMENT 'Filename' ,
>   PRIMARY KEY (`value_id`),
>   UNIQUE `IDX_EAV_ATTRIBUTE_OPTION_VALUE_OPTION_ID` (`option_id`),
>   CONSTRAINT `FK_EAV_ATTR_OPT_VAL_OPT_ID_EAV_ATTR_OPT_OPT_ID` FOREIGN KEY (`option_id`) REFERENCES `eav_attribute_option` (`option_id`) ON DELETE CASCADE ON UPDATE CASCADE
> ) COMMENT='Eav Attribute Option Swatch' ENGINE=INNODB charset=utf8 COLLATE=utf8_general_ci
> ```
> 
> This is the part in the PHP file which must be checked
> 
> ```
>     ->addForeignKey(
>         $installer->getFkName('eav/attribute_option_value', 'option_id', 'eav/attribute_option', 'option_id'),
>         'option_id',
>         $installer->getTable('eav/attribute_option'),
>         'option_id',
>         Varien_Db_Ddl_Table::ACTION_CASCADE,
>         Varien_Db_Ddl_Table::ACTION_CASCADE
>     )
> ```

I have retested the installation and I don't encounter any issues. Can you please check in the '`eav_attribute_option`' table if you have any duplicate option_id and if the schema is not corrupted?

empiricompany:
> If I change the name of a color in the attribute options page (e.g. from Blue to Blue Print), in the /media/wysiwyg/swatches directory, there is not PNG file with the new name. The fallback should work but I didn't find any directory in /media/catalog/swatches.

I have tested it, and it works correctly for me. The swatch with the new name (even with spaces) is dynamically created in the 'media\catalog\swatches' folder when you view the product page.

empiricompany:
> What is the priority when I set an image on the product edit page to be a swatch. Based on this custom code, it takes the global one defined in attribute page or the one in the product image tab?

The swatch created by the color picker is the last fallback, so all other methods take precedence.

empiricompany:
> At the first evaluation, you have to be careful with the initial values, because if there are already options with values, a black swatch is associated for all of them. In the Magento Sample Pack in size and color attributes, I am getting a black swatch for all options.
> 

Yes, the issue with black in the color picker is described in the description. We need to find a solution, including resetting the swatch to null.

For attributes that do not use swatches, simply not filling them out as we currently do, without loading any swatches in 'media/wysywyg/swatches', is sufficient.

In the description, you can find this and other issues to be resolved, such as the ability to directly upload an image in `media/wysywyg/swatches`. I hope someone is interested and can help me complete this PR because I have been short on time lately.

empiricompany:
> For attributes that do not use swatches, simply not filling them out as we currently do, without loading any swatches in 'media/wysywyg/swatches', is sufficient.

@ADDISON74 Sorry, I need to correct myself. By default, they are all saved black, so even saving the "size" swatch, for which we don't want any color. Therefore, this is a case to handle, and with a few lines of JavaScript, we can prevent it and save only the colors that are actually selected.

ADDISON74:
> I have retested the installation and I don't encounter any issues. Can you please check in the '`eav_attribute_option`' table if you have any duplicate option_id and if the schema is not corrupted?

For each individual PR I use an instance from scratch, OpenMage the latest updated version, based on the Magento Sample Pack. I don't make changes in the source code, I just take PRs and check them according to the description. This allows me to provide a clean feedback.

I have attached the **eav_attribute_option** table. I did not find duplicates on the **option_id** column. The scheme is fine. Still have the same error. 

I will use in tests OpenMage without the Magento Sample Pack.

empiricompany:
> > I have retested the installation and I don't encounter any issues. Can you please check in the '`eav_attribute_option`' table if you have any duplicate option_id and if the schema is not corrupted?
> 
> For each individual PR I use an instance from scratch, OpenMage the latest updated version, based on the Magento Sample Pack. I don't make changes in the source code, I just take PRs and check them according to the description. This allows me to provide a clean feedback.
> 
> I have attached the **eav_attribute_option** table. I did not find duplicates on the **option_id** column. The scheme is fine. Still have the same error.
> 
> I will use in tests OpenMage without the Magento Sample Pack.
> 
> [eav_attribute_option.csv](https://github.com/OpenMage/magento-lts/files/14236792/eav_attribute_option.csv)

I also use the same setup as you, so I don't understand why I didn't get an error. I must have fixed the error in the setup.

empiricompany:
> 1 - I would like to avoid adding swatches to all attributes of a certain type and to avoid the situation where all options receive the Black value. By default, each option should not have an associated value and leave it up to the user to decide. If it is not possible then it must be a white color with 100% transparency.

now if you not pick a color swatch is not saved, i've added also a class to stylish in future commit the color picker empty now is black by default by browser, but if you save only color picked get saved

fballiano:
This new feature is very nice so I'll test it out as soon as possible, I'm doing a lot of cleanup work to try to lower the numbers of PRs and process/merge old ones, but I didn't forget about this one.

app/code/core/Mage/Eav/Model/Resource/Entity/Attribute.php
```diff
@@ -288,6 +288,7 @@ protected function _saveOption(Mage_Core_Model_Abstract $object)
             $adapter            = $this->_getWriteAdapter();
             $optionTable        = $this->getTable('eav/attribute_option');
             $optionValueTable   = $this->getTable('eav/attribute_option_value');
+            $optionSwatchTable   = $this->getTable('eav/attribute_option_swatch');
```
fballiano, line 291:
```suggestion
            $optionSwatchTable  = $this->getTable('eav/attribute_option_swatch');
```

app/code/core/Mage/Eav/Model/Resource/Entity/Attribute/Option/Swatch.php
```diff
@@ -0,0 +1,28 @@
...
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Eav
+ * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
+ * @copyright  Copyright (c) 2019-2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 12:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Eav/sql/eav_setup/upgrade-1.6.0.1-1.6.0.2.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Eav
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 11:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

app/code/core/Mage/Eav/Model/Entity/Attribute/Option/Swatch.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * OpenMage
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available at https://opensource.org/license/osl-3-0-php
+ *
+ * @category   Mage
+ * @package    Mage_Eav
+ * @copyright  Copyright (c) 2023 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 11:
```suggestion
 * @copyright  Copyright (c) 2024 The OpenMage Contributors (https://www.openmage.org)
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -64,6 +64,9 @@
         <div class="hor-scroll">
             <table class="dynamic-grid" cellspacing="0"  cellpadding="0">
                 <tr id="attribute-options-table">
+                    <?php if ($this->isConfigurableSwatchesEnabled()): ?>
+                        <th><?php echo $this->__('Swatch'); ?></th>
```
fballiano, line 68:
```suggestion
                        <th><?php echo $this->__('Swatch') ?></th>
```

fballiano: I think we have to add the Swatch word to `app/locale/en_US`

fballiano:
> ... provided that it is strictly limited to the attribute that is used for the color swatch (e.g. color).

of course

> My opinion that a separate extension is preferable for such a change. 

since we already have swatches in the core I think this is a nice add for the core itself

fballiano:
> ... provided that it is strictly limited to the attribute that is used for the color swatch (e.g. color). 

I'm testing it now and I can't replicate what you say, only attributes marked as swatches have the color picker, so only these ones:

Screenshot 2024-02-21 alle 16 44 24

while all the others don't have the color picker:
Screenshot 2024-02-21 alle 16 45 12

ADDISON74:
I agree with you, only that attaching a "Color Picker" must be done strictly for the Color attribute. You cannot attach color picker for attributes like "Camera Megapixels", "Manufacturer" or "Shoe size".

This problem of logic and functionality is solved immediately provided that this PR offers the feature to upload an image (JPEG, GIF, PNG, WebP). In this case it is requested to analyze what happens if we deactivate a swatch attribute so that it is no longer a swatch, what orphan images remain in the directories, some data in the database.

fballiano:
> I agree with you, only that attaching a "Color Picker" must be done strictly for the Color attribute. You cannot attach color picker for attributes like "Camera Megapixels", "Manufacturer" or "Shoe size".

but it's not, it's only for the attributes marked as swatches.

ok maybe it would be better to have a second multiselect to select which attribute, among swatches, to have the color. but at the same time also no, because if it's a swatch it's a watch and will have a image/color anyway.

fballiano:
> This problem of logic and functionality is solved immediately provided that this PR offers the feature to upload an image (JPEG, GIF, PNG, WebP). In this case it is requested to analyze what happens if we deactivate a swatch attribute so that it is no longer a swatch, what orphan images remain in the directories, some data in the database.

That's not part of the PR but not even of the funcionality already in the core, I don't think it should be required for this specific PR to have file upload or the cleanup IMHO.

fballiano:
> What do you do when the selected attribute "Megapixel Camera" gets a color picker?

same as I posted at https://github.com/OpenMage/magento-lts/pull/3686#issuecomment-1957261316

Screenshot 2024-02-21 alle 17 34 39

no color picker becuase it's the attribute is not marked as a swatch, which seems the correct behavior

ADDISON74:
How many attributes can get a color picker? My opinion as only one, the `color` attribute.

If you select "Camera Megapixels" to be a swatch, because Magento allows you to use it in another way (Images section), that attribute will get the color picker, which is not logical.

fballiano:
I think it's kinda logical, if you want that attribute to be a swatch it means you'll either have a color or an image, but a swatch anyway.

as I said, maybe it could be better to have another multiselect for which swatch should have a color picker, but I also think it's ok this way, it doesn't get in the way and it could still be needed for some implementations.

app/code/core/Mage/ConfigurableSwatches/Helper/Productimg.php
```diff
@@ -263,6 +266,46 @@ public function getGlobalSwatchUrl(
         return Mage::getBaseUrl(Mage_Core_Model_Store::URL_TYPE_MEDIA) . $swatchImage;
     }
 
+    /**
+     * Create a swatch image for the given filename
+     *
+     * @param string $value
+     * @param int $width
+     * @param int $height
+     * @return string|null $destPath
```
fballiano, line 275: it seems to me that this method never returns null, am I missing something?

app/code/core/Mage/Eav/Model/Entity/Attribute/Option.php
```diff
@@ -34,4 +34,19 @@ public function _construct()
     {
         $this->_init('eav/entity_attribute_option');
     }
+
+    /**
+     * Retrieve swatch value
+     *
+     * @return Mage_Eav_Model_Entity_Attribute_Option_Swatch|bool
```
fballiano, line 41:
```suggestion
     * @return string|false
```

app/code/core/Mage/Eav/Model/Entity/Attribute/Option.php
```diff
@@ -34,4 +34,19 @@ public function _construct()
...
+     */
+    public function getSwatch()
```
fballiano, line 43:
```suggestion
    public function getSwatchValue()
```

not sure about this but since it returns the value... and in this case we can remove the `* Retrieve swatch value` comment

ADDISON74:
That color picker must be attached to only one attribute - Color. It is to be analyzed if it can be added to other attributes (e.g. Fabric Color, Shoe Color, ...). In this case instead of a drop-down for only one option we can have a multiselect. As you can see, it is mandatory to block the display of the color picker in the attribute edit page through an option in the section dedicated to swatches.

In addition, there are extensions that associate color or image swatches in dedicated Backend pages.

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +186,27 @@ var attributeOption = {
             this.updateItemsCountField();
         }
     },
+    swatch : function(event){
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
```
fballiano, line 190: I'd totally avoid using prototypejs

fballiano:
> That color picker must be attached to only one attribute - Color. It is to be analyzed if it can be added to other attributes (e.g. Fabric Color, Shoe Color, ...). In this case instead of a drop-down for only one option we can have a multiselect. As you can see, it is mandatory to block the display of the color picker in the attribute edit page through an option in the section dedicated to swatches.

I understand what you're saying but my opinion is that it's ok the way @empiricompany thought, all swatches can have images/color so I don't see it as a problem.

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +186,27 @@ var attributeOption = {
...
+                                                               // have table parent in safari
+        // Safari workaround
+        element.ancestors().each(function(parentItem){
```
fballiano, line 193: thi is also probably prototypejs

fballiano:
I tested the functionality backend/frontend and the precedence that has to be granted to files uploaded in media/wysiwyg/swatches and everything worked as expected :-)

maybe we can iron out some minor things but so far great job @empiricompany

ADDISON74:
> all swatches can have **images**/color so I don't see it as a problem.

The big problem is that this PR does not allow uploading images for swatches. This is where the discussion between us started.

Also, it adds the color Black to all the options of an attribute. Instead of querying a small table of 50 rows, we end up searching in a table with 10,000 rows. Offering a color picker for any swatch attribute is a logical error, but I will not insist on this topic.

fballiano:
> The big problem is that this PR does not allow uploading images for swatches. This is where the discussion between us started.

this was never present so it could be added in a later PR

> Also, it adds the color Black to all the options of an attribute. Instead of querying a small table of 50 rows, we end up searching in a table with 10,000 rows.

it doesn't get saved so that's incorrect:
Screenshot 2024-02-21 alle 18 17 08

> Offering a color picker for any swatch attribute is a logical error, but I will not insist on this topic.

we have different opinions here. the picker is shown only for attributes marked as swatches which I think can be considered ok

empiricompany:
> adds the color Black to all the options of an attribute

this is fixed in last commit, you see black in backend because is default of color picker by the browser but they are not applied to options. 

For me, the only thing missing is to stylize the color picker when empty, I've added a class for this

ADDISON74:
Let's analyze the following real situation, from a logical perspective.

1. In the **Backend** we select the product attributes that will be used as swatches.

2. Any selected attribute in the list will automatically get the Color Picker option. As a first observation, in the case of the Size attribute, it is obvious that we cannot use colors, it would be illogical if we did.

screenshot-swatch-1

What happens if I want the Color Picker only for the Color attribute? As the PR is created, it does not allow to do this and the Color Picker will be displayed in the editing page of all the selected attributes (e.g. Size attribute). This is not a logical behavior.

How do we solve this issue that creates confusion? Very simple, one more option is created in the Backend to control which attributes the Color Picker is displayed to. It would be good to correlate the selected swatch attributes with those that gets the Color Picker, for example, we enable the Color Picker for an attribute that is not set as a swatch.

screenshot-swatch-2

Based on this change we are able to control the Color Picker display. 

By default, no swatch attribute should get the Color Picker, giving the user the opportunity to choose one or more. At this stage, the issue is solved. There are attributes that will behave like swatches (Magento behavior) and some attributes that get the Color Picker.

ADDISON74:
One more observation. when I use the Color Picker, I would like the HEX option to be visible from the very beginning (if possible). It is easier to insert a HEX value (copy&paste), than entering RGB or HSL values in 3 boxes.

screenshot-swatch-3

empiricompany:
@ADDISON74 it seems that you have not understood the initial and simple purpose of this PR, I'm sorry but I have to say it, often your comments are an obstacle to the progress of this project, nothing personal but this attitude of wanting to emphasize own reasons without contradictory drives developers away. 

Here it seems that we have to do everything as you say and that's not always right. The problem is that it needs active participation like yours and @fballiano from more members

fballiano:
I'm sorry @ADDISON74 but I don't think the way this PR is implemented is wrong. Everybody is entitled to his/her own opinion but to me I think this implementation is ok, as I tried to explain in some different comments, and I don't think it's right for you to say that I'm wrong and that that's the only truth, you know what I'm saying?

empiricompany:
> The way this PR is implemented is wrong

You can express your opinions, but it's unfair that you want to impose them.
All the choices in this project cannot be decided by one person

ADDISON74:
I will rephrase my comments

1. You cannot enable the Color Picker for all swatch attributes.
2. This change may conflict with extensions already used

Faced with these opinions, I expect to be countered with logical arguments, not to be unfairly reproached for imposing changes, for creating an unproductive atmosphere in this group.

empiricompany:
> You cannot enable the Color Picker for all swatch attributes.

this is your opinion, it can be done and I have done it.
in my opinion the color picker is a small addition to what is already there and is minimally invasive. I don't want a new configuration to enable them on individual attributes, I just don't compile them.
it would be clearer if instead of black there was something like this
blank_swatch
and as I said a class is being added to do this (swatch-disabled), but I haven't had time yet


> This change may conflict with extensions already used

it is the price to pay for progress otherwise advanced swatch management will never be implemented and the platform will not evolve. my clients use different extensions and under no circumstances would it cause conflicts


However, I have already understood that here too I wasted my time unnecessarily. as long as we are only you, me and @fballiano

app/code/core/Mage/ConfigurableSwatches/Helper/Productimg.php
empiricompany: it should return null if it fails to create the swatch from the loaded image, so it creates it from the color picker

fballiano: If I'm not reading it wrong, it doesn't seem to do that check because it returns '' on line 284 and then I don't see any return statement until line 306 but $destPath seems to always be a string

ADDISON74:
>it can be done and I have done it.

To me this phrase sounds imposing. In fact, you want to enable Color Picker for **all** the attributes established as swatches, arguing that it is minimally invasive and a progress for OpenMage. Honestly speaking, to see the Color Picker on an attribute page like Size, where I don't want something like that, means that the programmer did his job only halfway.

However, you unfairly criticized me because I create an unproductive atmosphere in the community, for imposing my ideas, quite serious accusations. Please do not lose sight of the fact that for 4 years I have been analyzing and approving PRs.

It's a shame that you don't take into account the fact that I checked out this PR, that I spent time analyzing it. Teamwork means exchange of ideas, battle of arguments. If you consider that you have wasted your time, you disregard those who shared their opinions to improve the functionality. 

I conclude by saying that the arguments I asked you did not convince me.

empiricompany:
It's funny that you consider an imposition  when you type 

> You cannot enable the Color Picker for all swatch attributes.

to me seems, you always want to have the last word in discussions and attack anyone who contradicts you, I'd never be able to write / do something that suits you needs

ADDISON74:
Please dispel with the unfounded accusations. I didn't attack anyone as you do it now. You are rude and arrogant.

empiricompany:
just can't go on like this! @ADDISON74 You keep accusing me unjustly. 
Here, you are the arrogant, not me! Something must be done by admins about this harmful attitude!

ADDISON74:
You better limit yourself to the arguments that I politely asked you before, related to your PR. It's pathetic to ask the administrators for help after the accusations you brought against me. Please find them above.

Flyingmana:
@ADDISON74 please slow a bit down.

You did mention a lot of different details, and some of them are definitely no blockers and can get easily refined/improved in future iterations of the feature.

Lets simplify the questions a bit.

Is it possible to continue not using the new color picker?

Is a possibly no-optimal usability a blocker argument for anyone?

What happens if its active for a product/attribute but no color is choosen?

ADDISON74:
@Flyingmana - Better to let the author finish his work then you evaluate it. No matter how many explanations I provide, he will tell you that I determine the developers to stop collaborating, that I force everyone to do what I say. Serious but unproven accusations. Out of respect for the rest of you, I will no longer comment or review this gentleman's PRs.

app/code/core/Mage/ConfigurableSwatches/Helper/Productimg.php
empiricompany: you're right, I corrected it with false

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
empiricompany: I didn't remove prototypejs to avoid having mixed code, I think it can be the subject of a future PR

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -187,6 +228,15 @@ var attributeOption = {
                 Event.observe(buttons[i], 'click', this.remove.bind(this));
             }
         }
+    },
+    bindSwatchButtons : function(){
+        var buttons = $$('.swatch-option');
```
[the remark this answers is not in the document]
empiricompany: I didn't remove prototypejs to avoid having mixed code, I think it can be the subject of a future PR

app/code/core/Mage/Eav/Model/Entity/Attribute/Option.php
empiricompany: i've applied your suggestions but keep comment as "Retrieve swatch hex value"

empiricompany:
i've added an empty placeholder if no color is selected instead of default black
it's a bit hacky i don't know how is in mac and linux

306927438-db3f9dd3-f340-4508-93ca-d9df33d087ac

fballiano:
I see the new changes this way on macOS/chromium:

Screenshot 2024-02-22 alle 09 08 54

the placeholder are a bit bigger than the other colors, could it be a problem of my setup? What would you say if we made the "red line" (on the placeholder) light-gray instead? The red kinda makes me think that there's some error.

fballiano:
@Flyingmana 

> Is it possible to continue not using the new color picker?

sure, the current implementation (from my tests) is backward compatible, you can upload the swatches as you were doing for all these years and it will work. or you can use the picker to generate a swatch of that color. or mix and match the 2 things.

> What happens if its active for a product/attribute but no color is choosen?

(from my tests) nothing is saved on the new "_swatch" database table and the templates fall back to the previous implementation

fballiano:
@empiricompany I can't seem to find a way to "unset" a color from the color picker, and if I remove a record from the eav_attribute_option_swatch there's no way to remove the "resised" image from the media/catalog/swatches folder.

empiricompany:
> @empiricompany I can't seem to find a way to "unset" a color from the color picker, 

true, we need to add an x ​​to reset the color picker.
on click we can set input hidden to -1 and apply the "swatch-disabled" class to the input color in order to show placeholder.
then need to add a check when saving it and delete the row if input equal -1

> and if I remove a record from the eav_attribute_option_swatch there's no way to remove the "resised" image from the media/catalog/swatches folder.

maybe this is not necessary, when you change a color the swatches cache is emptied.
but if necessary before delete the record we have the filename to delete

fballiano:
> > and if I remove a record from the eav_attribute_option_swatch there's no way to remove the "resised" image from the media/catalog/swatches folder.
> 
> maybe this is not necessary, when you change a color the swatches cache is emptied. but if necessary before delete the record we have the filename to delete

that would be ok. I said so because if I remove the entry from the _swatch table, the frontend still seems to "see" the image from the media/catalog/swatches cache folder

fballiano:
> Unfortunately this depends on how specific browsers render the input color, there is no universal solution. in windows as shown on the screen it has the same dimensions. are you using safari on mac?

MacOs/Chromium

> it is usually represented with a red line, It is also represented like this in Magento 2 and other platforms.

are we saying M2 is a good place to get ideas from? heheheh
anyway, not a big deal obviously but I don't know, a red line crossing something doesn't seem to me a good way of telling somebody that something is not set, rather a forbidden or error (general discussion, not strictly related to this PR).

empiricompany:
> are we saying M2 is a good place to get ideas from? heheheh anyway, not a big deal obviously but I don't know, a red line crossing something doesn't seem to me a good way of telling somebody that something is not set, rather a forbidden or error (general discussion, not strictly related to this PR).

not to oppose me but it is a standard that we find in all programs for example on illustrator it is like this:
illustrator

fballiano:
@empiricompany Thanks, I'm didn't think about that :-)

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
fballiano: ok but each could be simply replaced by forEach right? which would probably be better anyway no?

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
fballiano: I think we shouldn't add prototypejs based new code, it would be less work/testing later when we'll finally convert everything to vanillajs.

anyway, I'll not oppose to this but also I can't collaborate on this PR because it's locked

empiricompany: what do you say we do this in a next PR? I don't have time now

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
empiricompany: sure i've unlocked editing, I am grateful if you contribute

fballiano:
> > > and if I remove a record from the eav_attribute_option_swatch there's no way to remove the "resized" image from the media/catalog/swatches folder.
> > 
> > 
> > maybe this is not necessary, when you change a color the swatches cache is emptied. but if necessary before delete the record we have the filename to delete
> 
> that would be ok. I said so because if I remove the entry from the _swatch table, the frontend still seems to "see" the image from the media/catalog/swatches cache folder

I'll only check this thing above and they it's fine to me.

fballiano:
moving to draft because there's the need to address https://github.com/OpenMage/magento-lts/pull/3686#issuecomment-1960172501

empiricompany:
@fballiano now you can delete colors
i have updated the screenshot demo in the description of PR

empiricompany:
to unify the color picker across all browsers what do you think about this? same height as input 26px
crossbrowser
ps. the empty color placeholder needs to be fixed

there is this strange rule:
html5-color-picker
I don't know why and where it is used

- - -

I also want to name the column something like "swatch fallback color" to be more precise

empiricompany:
@Flyingmana 

> Is it possible to continue not using the new color picker?

yes you can just not fill the color

> Is a possibly no-optimal usability a blocker argument for anyone?

the usability now seems optimal to me, it is now possible to delete a previously saved color
if by usability you mean for a non-color swatch attribute, like the size, simply don't fill the color
in the future extending this PR it will be possible to allow the uploading of an image as swatch

> What happens if its active for a product/attribute but no color is choosen?

nothing happens, the swatches continue to work following the fallback order. if no swatch exists, a broken image is shown as before

empiricompany:
@fballiano please share your screenshot i have stylish better the color picker to show the same in different OS

fballiano:
Sorry I'm kinda short on time today but here's my screenshot

Screenshot 2024-03-01 alle 12 51 10

empiricompany:
> Sorry I'm kinda short on time today but here's my screenshot

It seems acceptable to me

skin/adminhtml/default/openmage/form.css
```diff
@@ -73,246 +68,216 @@ button.save, button.ok_button, button.add, .form-button.ok_button {
...
+    border-radius: 2px; }
+  td:has(input.swatch-option) .swatch-option {
+    width: 26px !important;
+    height: 26px; }
+
+td:has(input.swatch-disabled) {
+  background-image: url('data:image/svg+xml,<svg id="e5490f2e-3f54-4acd-9496-57eed667c240" data-name="Livello 1" xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 28.41093 28.29249"><title>placeholder</title><rect x="1.20546" y="1.08375" width="26" height="26" fill="%23fff" stroke="%23ccc" stroke-miterlimit="10"/><line x1="0.70546" y1="27.58375" x2="27.70546" y2="0.70875" fill="none" stroke="%23ff003a" stroke-miterlimit="10" stroke-width="2"/></svg>');
+  background-repeat: no-repeat;
+  outline: 1px solid #c8c8c8;
+  border: 1px solid #fff;
```
fballiano, line 270: not a deal breaker but I kinda like it more without outline/border, is seems cleaner to me because it has many less lines, what do you think about that?

empiricompany: yes, it's a bit hacky I know but I always had to do it for emulate a color picker empty

without border / outline
whitout

original
original

fballiano: understood

app/code/core/Mage/Eav/Block/Adminhtml/Attribute/Edit/Options/Abstract.php
```diff
@@ -209,4 +211,15 @@ public function getAttributeObject()
     {
         return Mage::registry('entity_attribute');
     }
+
+    /**
+     * Check if configurable swatches module is enabled and attribute is swatch type
+     *
+     * @return boolean
```
fballiano, line 218:
```suggestion
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
             this.updateItemsCountField();
         }
     },
+    swatch : function(event){
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
```
fballiano, line 193:
```suggestion
        let element = event.target.closest('tr');
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
...
+            elementSwatchValue.value = event.target.value;
+            event.target.classList.remove('swatch-disabled');
+        }
+    },
+    swatchremove : function(event){
+        if (!confirm(event.target.getAttribute('data-msg-delete'))) {
+            return;
+        }
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
+                                                                // have table parent in safari
```
fballiano, line 218:
```suggestion
                                                                // have table parent in safari
```
this line seems misaligned?

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
...
+            elementSwatchValue.disabled = false;
+            elementSwatchValue.value = event.target.value;
+            event.target.classList.remove('swatch-disabled');
+        }
+    },
+    swatchremove : function(event){
+        if (!confirm(event.target.getAttribute('data-msg-delete'))) {
+            return;
+        }
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
```
fballiano, line 217:
```suggestion
        let element = event.target.closest('tr');
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -187,6 +256,24 @@ var attributeOption = {
                 Event.observe(buttons[i], 'click', this.remove.bind(this));
             }
         }
+    },
+    bindSwatchButtons : function(){
+        var buttons = $$('.swatch-option');
```
fballiano, line 261:
```suggestion
        let buttons = document.querySelectorAll('.swatch-option');
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -187,6 +256,24 @@ var attributeOption = {
...
+        for(var i=0;i<buttons.length;i++){
+            if(!$(buttons[i]).binded){
+                $(buttons[i]).binded = true;
+                Event.observe(buttons[i], 'change', this.swatch.bind(this));
+            }
+        }
+    },
+    bindSwatchRemoveButtons : function(){
+        var buttons = $$('.swatch-delete');
```
fballiano, line 270:
```suggestion
        let buttons = document.querySelectorAll('.swatch-delete');
```

app/locale/en_US/Mage_ConfigurableSwatches.csv
```diff
@@ -17,3 +17,5 @@
 "Swatch Dimensions in Product Listing","Swatch Dimensions in Product Listing"
 "Swatch Dimensions on Product Detail Page","Swatch Dimensions on Product Detail Page"
 "Width","Width"
+"Swatch","Swatch"
+"Are you sure to delete this color?","Are you sure to delete this color?"
```
fballiano, line 21:
```suggestion
"Are you sure to delete this fallback color?","Are you sure to delete this fallback color?"

```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
...
         }
     },
+    swatch : function(event){
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
+                                                               // have table parent in safari
+        // Safari workaround
+        element.ancestors().each(function(parentItem){
+           if (parentItem.hasClassName('option-row')) {
+               element = parentItem;
+               throw $break;
```
fballiano, line 199: question: is this right? couldn't find any $break in the javascript code

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
             this.updateItemsCountField();
         }
     },
+    swatch : function(event){
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
+                                                               // have table parent in safari
+        // Safari workaround
+        element.ancestors().each(function(parentItem){
```
fballiano, line 196:
```suggestion
        let parentItem = element;
        while ((parentItem = parentItem.parentNode) !== null) {
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -166,6 +189,52 @@ var attributeOption = {
...
+        }
+    },
+    swatchremove : function(event){
+        if (!confirm(event.target.getAttribute('data-msg-delete'))) {
+            return;
+        }
+        var element = $(Event.findElement(event, 'tr')); // !!! Button already
+                                                                // have table parent in safari
+        // Safari workaround
+        element.ancestors().each(function(parentItem){
```
fballiano, line 220:
```suggestion
        let parentItem = element;
        while ((parentItem = parentItem.parentNode) !== null) {
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -187,6 +256,24 @@ var attributeOption = {
...
             }
         }
+    },
+    bindSwatchButtons : function(){
+        var buttons = $$('.swatch-option');
+        for(var i=0;i<buttons.length;i++){
+            if(!$(buttons[i]).binded){
+                $(buttons[i]).binded = true;
+                Event.observe(buttons[i], 'change', this.swatch.bind(this));
+            }
```
fballiano, line 266:
```suggestion
            if (!buttons[i].dataset.binded) {
                buttons[i].dataset.binded = true;
                buttons[i].addEventListener('change', this.swatch.bind(this));
            }
```

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -187,6 +256,24 @@ var attributeOption = {
...
+        }
+    },
+    bindSwatchRemoveButtons : function(){
+        var buttons = $$('.swatch-delete');
+        for(var i=0;i<buttons.length;i++){
+            if(!$(buttons[i]).binded){
+                $(buttons[i]).binded = true;
+                Event.observe(buttons[i], 'click', this.swatchremove.bind(this));
+            }
```
fballiano, line 275:
```suggestion
            if (!buttons[i].binded) {
                buttons[i].binded = true;
                buttons[i].addEventListener('click', this.swatchremove.bind(this));
            }
```

skin/adminhtml/default/openmage/scss/form.scss
```diff
@@ -402,4 +402,45 @@ fieldset[disabled] button {
...
+  background-position: center center;
+  border-radius: 2px;
+  
+  .swatch-option {
+    opacity: 0;
+  }
+  .swatch-delete {
+    display: none;
+  }
 }
```
fballiano, line 446:
```suggestion
}

```

fballiano:
I tested the "remove fallback color" functionality and it works great

empiricompany:
@fballiano thanks a lot for all your suggestions! 
Regarding the deprecation of prototype code, can I safely merge it? Have you already tested it?

fballiano:
@empiricompany I will test my alternative code asap

fballiano:
- I rewrote the javascript part removing prototypejs
- removed safari hacks because as far as I could test, the hack is not needed anymore
- retested everything

imho this is ready to be merged

fballiano (approved):

fballiano:
@empiricompany can you give this one a test to confirm everything is working for you too?

empiricompany:
yep I've already tested!

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -64,6 +64,9 @@
         <div class="hor-scroll">
             <table class="dynamic-grid" cellspacing="0"  cellpadding="0">
                 <tr id="attribute-options-table">
+                    <?php if ($this->isConfigurableSwatchesEnabled()): ?>
+                        <th><?php echo $this->__('Swatch') ?></th>
```
Sekiphp, line 68: escape the string

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
```diff
@@ -75,7 +78,14 @@
                             <?php endif ?>
                         </th>
                     </tr>
-                    <tr class="no-display template" id="row-template">
+                    <tr class="no-display template" id="row-template">    
+                        <?php if ($this->isConfigurableSwatchesEnabled()): ?>
+                        <td style="position:relative">
+                            <input type="hidden" disabled class="swatch-value" name="option[swatch][{{id}}]" value="{{swatch}}" />
+                            <input type="color" class="input-color-html5 validate-hex-color-hash {{swatch_class}}" value="{{swatch}}" />    
+                            <span class="swatch-delete" data-msg-delete="<?php echo $this->__('Are you sure to delete this fallback color?'); ?>">X</span>
```
Sekiphp, line 86: escape

Sekiphp (changes requested):

app/code/core/Mage/ConfigurableSwatches/Helper/Productimg.php
```diff
@@ -263,6 +266,46 @@ public function getGlobalSwatchUrl(
...
+            self::SWATCH_CACHE_DIR,
+            Mage::app()->getStore()->getId(),
+            $width . 'x' . $height,
+            'media',
+            trim($filename, '/'),
+        ];
+        $destPath = implode('/', $destPathArr);
+        if (!is_dir(Mage::getBaseDir(Mage_Core_Model_Store::URL_TYPE_MEDIA) . DS . dirname($destPath))) {
+            $io = new Varien_Io_File();
+            $io->mkdir(Mage::getBaseDir(Mage_Core_Model_Store::URL_TYPE_MEDIA) . DS . dirname($destPath), 0777, true);
```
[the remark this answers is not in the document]
fballiano: there are many accorrences of 777 all over the codebase

Screenshot 2024-04-22 alle 10 29 56

actually all of the mkdir have a 777 so I think this topic should be part of another PR and @empiricompany is correct to use the default permission that's used everywhere else

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
fballiano: done

app/design/adminhtml/default/default/template/eav/attribute/options.phtml
fballiano: done

fballiano:
I have to retest this properly because it doesn't seem to save/delete the images from media/catalog/swatches anymore

fballiano:
I've retested everything and it works great, when modifying an attribute the cache folder gets deleted (correctly) and it gets recreated by the frontend

fballiano:
after thorough tests and long development, no BC is introduced, no errors seem to be logged because of this, we also addressed the last comments by @Sekiphp so it seems only right to merge this.

PR #3691, opened 2023-12-05
Allow an empty `content` field in a CMS Block and Page

loekvangool:
Currently, there needs to be some content in a CMS block. I propose that this restriction is removed. A HTML block can be a placeholder for future content.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Cms/Block/Edit/Form.php
+++ b/app/code/core/Mage/Adminhtml/Block/Cms/Block/Edit/Form.php
@@ -116,7 +116,7 @@ class Mage_Adminhtml_Block_Cms_Block_Edit_Form extends Mage_Adminhtml_Block_Widg
             'label'     => Mage::helper('cms')->__('Content'),
             'title'     => Mage::helper('cms')->__('Content'),
             'style'     => 'height:36em',
-            'required'  => true,
+            'required'  => false,
             'config'    => Mage::getSingleton('cms/wysiwyg_config')->getConfig()
         ]);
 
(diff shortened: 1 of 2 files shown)


luigifab (approved):
When using an empty static bloc in a cms page with:
`{{widget type="cms/widget_block" template="cms/widget/static_block/default.phtml" block_id="2"}}`
we have:
`... <div class="widget widget-static-block"></div> ...`
but ok.

loekvangool:
I would totally zap that empty block, which I actually do in my templates, but I'm worried about breaking some custom code if we just decide to not output anything.

sreichel:
Why not add a html comment?

loekvangool:
Add a HTML comment to workaround the rule that the block cannot be empty?

What does the rule add?

sreichel:
What's the benefit of empty content?

Should title be optional too? Do we need URL for pages?

Why to change this for all? You can rewrite that class to fit your custom workflow?

pquerner:
I vote to decline as well. I do not see any benefit.

loekvangool:
I'm emptying HTML blocks for promotions until the next promotion.

But whatever.

pquerner:
so create a block for it which changes the output if a promotion is currently available. I wouldnt touch the whole cms system for that and make a crucial field optional. for one, and only your usecase. many roads lead to rome.

loekvangool:
What exactly justifies the term "touch the whole CMS system"? It works just fine without any content.

Flyingmana:
The strength of OpenMage is the ability to do a lot without needing a developer.

Having fields required is usually a protective and usability focused measurement, when not filling something would lead to errors or extra effort (by repeating filling everything)

In a CMS block, which is very easy to edit again, and which has no visible potential for errors regarding the content field, the need to make or keep it required seems not to be given for me.

One could also work with non-space characters, or special html classes to imitate the empty content. But thats not sounding much like a best practice approach and prone to errors.

pquerner:
Whats next? Text fields only make sense for 500 characters, so the field should truncate.
Theres only so many things that should be changed.

Like I said, by building a observer or custom block or whatever else comes to mind, its better documented of _WHY_ this is wanted behaviour without removing the wheels from cars and saying, I could want to live in that car and not want to move, so it works fine without the wheels.

sreichel:
> I'm emptying HTML blocks for promotions until the next promotion.

Disable block until next promotion?

PR #3705, opened 2023-12-19
Revert "Updated TinyMCE to 6.8 via composer 🚀"

fballiano:
#3658 is a great idea but sadly we stumble upon some composer limitations that don't allow us to make it work (check the original PR comments to know about those) so we kinda have to revert it at the moment

Reverts OpenMage/magento-lts#3658


ADDISON74 (approved):

fballiano:
ok I'll take responsability for this and merge this also without the reviews since:
- the original author of #3658 agrees
- the main branch is in a non-working state
- we knew before merging #3658 that we needed to merge it in order to test it and since the test failed we've to revert and get the main branch to a working state

PR #3707, opened 2023-12-21
Updated TinyMCE to 6.8.2

fballiano:
it contains some security updates (as described in https://github.com/OpenMage/magento-lts/pull/3658) so we should merge this PR and release a new version of OM


ADDISON74 (approved):

empiricompany (approved):

kiatng (approved):

ADDISON74:
At this moment there two TinyMCE versions 6.8.4 and 7.2.1. We need an update for 6.8.4 in the first stage.

fballiano:
tinymce7 is GPL so we'll never be able to use it, I've already argued with them on twitter...

PR #3722, opened 2024-01-03
Allow setting admin user notification email receiver

Judx:
### Description (*)

This PR allows setting receiver email to "New Admin User Create Notification". Default is that email is sent to store general email address. In some cases it is necessary to allow changing the notification receiver.

### Related Pull Requests

Depends on and includes changes from https://github.com/OpenMage/magento-lts/pull/3721

No.

1. Pull changes.
2. Find and set email address in: System -> Admin -> Security -> Notification Email Receiver
3. Create new admin user and see that the notification email is received by the specified address.

### DIFF
--- a/app/code/core/Mage/Admin/Model/User.php
+++ b/app/code/core/Mage/Admin/Model/User.php
@@ -733,17 +733,21 @@ class Mage_Admin_Model_User extends Mage_Core_Model_Abstract
      * You can declare additional emails in Mage_Core general/additional_notification_emails/admin_user_create node.
      *
      * @param Mage_Admin_Model_User $user
+     * @param string|false $receiverEmail An optional receiver email address for the notification. Overrides default and additional emails.
      * @return $this
      */
-    public function sendAdminNotification($user)
+    public function sendAdminNotification($user, $receiverEmail = false)
     {
         // define general contact Name and Email
         $generalContactName = Mage::getStoreConfig('trans_email/ident_general/name');
         $generalContactEmail = Mage::getStoreConfig('trans_email/ident_general/email');
 
-        // collect general and additional emails
-        $emails = $this->getUserCreateAdditionalEmail();
-        $emails[] = $generalContactEmail;
+        // collect general and additional emails unless overridden
+        if ($receiverEmail && filter_var($receiverEmail, FILTER_VALIDATE_EMAIL)) {
+            $emails = [$receiverEmail];
+        } else {
+            $emails = array_merge($this->getUserCreateAdditionalEmail(), [$generalContactEmail]);
+        }
 
         /** @var Mage_Core_Model_Email_Template_Mailer $mailer */
         $mailer    = Mage::getModel('core/email_template_mailer');
--- a/app/code/core/Mage/Core/etc/system.xml
+++ b/app/code/core/Mage/Core/etc/system.xml
@@ -1258,7 +1258,7 @@
                         </extensions_compatibility_mode>
                         <create_admin_user_notification translate="label comment">
                             <label>New Admin User Create Notification</label>
-                            <comment>This setting enable notification when new admin user created.</comment>
+                            <comment>This setting enables notification when new admin user is created.</comment>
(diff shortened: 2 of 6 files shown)


app/code/core/Mage/Admin/Model/User.php
```diff
@@ -733,17 +733,21 @@ protected function _getDateNow($dayOnly = false)
...
     {
         // define general contact Name and Email
         $generalContactName = Mage::getStoreConfig('trans_email/ident_general/name');
         $generalContactEmail = Mage::getStoreConfig('trans_email/ident_general/email');
 
-        // collect general and additional emails
-        $emails = $this->getUserCreateAdditionalEmail();
-        $emails[] = $generalContactEmail;
+        // collect general and additional emails unless overridden
+        if ($receiverEmail && filter_var($receiverEmail, FILTER_VALIDATE_EMAIL)) {
```
pquerner, line 746: `Zend_Validate::is($receiverEmail, 'EmailAddress')` perhaps?

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1256,16 +1256,28 @@
                             <show_in_website>0</show_in_website>
                             <show_in_store>0</show_in_store>
                         </extensions_compatibility_mode>
-                        <crate_admin_user_notification translate="label comment">
+                        <create_admin_user_notification translate="label comment">
```
pquerner, line 1259: I would leave the typo as-is. Yes, you move the value to the new setting, but legacy code may still use the old xpath.

app/code/core/Mage/Core/etc/system.xml
```diff
@@ -1256,16 +1256,28 @@
...
                             <label>New Admin User Create Notification</label>
-                            <comment>This setting enable notification when new admin user created.</comment>
+                            <comment>This setting enables notification when new admin user is created.</comment>
```
pquerner, line 1261: This is ok to change.

app/code/core/Mage/Admin/Model/User.php
luigifab: Yes, because `FILTER_VALIDATE_EMAIL` is not used.

app/code/core/Mage/Core/etc/system.xml
luigifab: For legacy code, this can be used, if I remember correctly:
```xml
<crate_admin_user_notification deprecated="true">
    <config_path>admin/security/create_admin_user_notification</config_path>
</crate_admin_user_notification>
```

app/code/core/Mage/Core/etc/system.xml
ADDISON74: Before changing the comment, please check if there is a translation for that phrase. If exists you should change only the CSV file.

app/code/core/Mage/Core/etc/system.xml
pquerner: Where is that @luigifab ?

app/code/core/Mage/Core/etc/system.xml
fballiano: yep, needs to be changed in Mage_Core.csv instead of here, so that we don't break language packs

app/code/core/Mage/Core/etc/system.xml
fballiano, line 1259: In https://github.com/OpenMage/magento-lts/pull/3721 was decided not to create a breaking change to fix a typo, let's please address the comments made by other users if possible

fballiano (changes requested):

app/code/core/Mage/Core/etc/system.xml
luigifab: Here:

https://github.com/OpenMage/magento-lts/blob/6ea8cf0ae16ed17b7866443fa4a3ecc2292e498e/app/code/core/Mage/Paypal/etc/system.xml#L29-L40

Issue #3727, opened 2024-01-09
PHP 8.1 TypeError: imagecopymerge(): Argument #9 ($pct) must be of type int in GD when custom watermark image is applied


Tomasz-Silpion:
### Preconditions (*)

1. PHP 8.1 or newer.
2. Main branch or any recent version of OpenMage.

### Steps to reproduce (*)

1. Go to System -> Configuration -> Product Image Watermarks.
2. Add watermark images and set every opacity field to any value.

### Expected result (*)

Watermark is properly applied to the product image.

### Actual result (*)

TypeError: imagecopymerge(): Argument #9 ($pct) must be of type int, string given exception is thrown in lib/Varien/Image/Adapter/Gd2.php:504

kiatng:
I had the exact same thing [a week ago](https://makandracards.com/magento1/614548-uncaught-typeerror-imagecopymerge-argument-9-pct-must-be-of-type-int-string-given-in-lib-varien-image-adapter-gd2-php-570):

Uncaught TypeError: imagecopymerge(): Argument #9 ($pct) must be of type int, string given in .../lib/Varien/Image/Adapter/Gd2.php:570
Posted 7 days ago. Visible to the public. Linked content.
Uncaught TypeError: imagecopymerge(): Argument #9 ($pct) must be of type int, string given in /.../lib/Varien/Image/Adapter/Gd2.php:570
Stack trace:
#0 /.../lib/Varien/Image/Adapter/Gd2.php(570): imagecopymerge(Object(GdImage), Object(GdImage), 0, 0, 0, 0, 400, 400, '')
#1 /.../lib/Varien/Image.php(197): Varien_Image_Adapter_Gd2->watermark('/home/web/publi...', 0, 0, 30, false)
#2 /.../app/code/core/Mage/Catalog/Model/Product/Image.php(557): Varien_Image->watermark('/home/web/publi...')
#3 /.../app/code/core/Mage/Catalog/Helper/Image.php(362): Mage_Catalog_Model_Product_Image->setWatermark('default/images_...')
#4 /.../app/design/frontend/.../default/template/page/html/head.phtml(73): Mage_Catalog_Helper_Image->__toString()

Set the product watermark opacity to some integer value:

sreichel:
@all-contributors add @Tomasz-Silpion bug

PR #3728, opened 2024-01-09
Cast store config image opacity to int (#3727)

Tomasz-Silpion:
### Description (*)

This PR fixes PHP 8.1 exception `Argument #9 ($pct) must be of type int, string given in lib/Varien/Image/Adapter/Gd2.php:504` thrown when custom image opacity is set in admin configuration.

1. Fixes OpenMage/magento-lts#3727

1. Ensure that you are working with PHP 8.1 or newer.
2. Go to System -> Configuration -> Product Image Watermarks.
3. Add watermark images and set every opacity field to any value.
4. Check if PDP works properly images are properly rendered with or without the adjustment.

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Image.php
+++ b/app/code/core/Mage/Catalog/Helper/Image.php
@@ -143,7 +143,7 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
             Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_image")
         );
         $this->setWatermarkImageOpacity(
-            Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_imageOpacity")
+            (int) Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_imageOpacity")
         );
         $this->setWatermarkPosition(
             Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_position")
--- a/app/code/core/Mage/Catalog/Helper/Image.php
+++ b/app/code/core/Mage/Catalog/Helper/Image.php
@@ -143,7 +143,7 @@ class Mage_Catalog_Helper_Image extends Mage_Core_Helper_Abstract
             Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_image")
         );
         $this->setWatermarkImageOpacity(
-            (int) Mage::getStoreConfig("design/watermark/{$this->_getModel()->getDestinationSubdir()}_imageOpacity")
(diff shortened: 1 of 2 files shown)


kiatng:
Is it better to fix it at here https://github.com/OpenMage/magento-lts/blob/11eca9d518be7e8073556b64ba36cf2178b276fb/app/code/core/Mage/Catalog/Helper/Image.php#L490-L500

It may be reset by some custom code to `null`.

Tomasz-Silpion:
I agree where this approach actually needs adjustment of some other lines as well https://github.com/OpenMage/magento-lts/pull/3728/commits/c982305c705bc93ea0c3d5278bce9120edeb5e48

kiatng (approved):

sreichel:
Id prefer the original PR ...

... and may later change it to `public function setWatermarkImageOpacity(int $imageOpacity)`

app/code/core/Mage/Catalog/Helper/Image.php
```diff
@@ -488,12 +488,12 @@ protected function getWatermarkSize()
     /**
      * Set watermark image opacity
      *
-     * @param int $imageOpacity
+     * @param int|string $imageOpacity
```
fballiano, line 491: why adding string if we cast to int later?

Tomasz-Silpion: Because Mage::getStoreConfig would pass string there.

So either let's go back to original concept with cast before Mage::getStoreConfig (let's vote with ❤️) or new one with casting to int inside setWatermarkImageOpacity (let's vote with 🎉).

sreichel: There are several places where Mage::getStoreConfig is cast to int.

Why not add a helper method similiar to getStoreConfigFlag ... ?

```
public static function getStoreConfigAsInt(string $path, $store = null): int
{
    return (int) self::getStoreConfig($path, $store);
}
```

sreichel: @kiatng @fballiano pls share your opinion :)

sreichel: > why adding string if we cast to int later?

I think adding string would be correct, b/c string is valid too.

Without it code analyze may say "unnecessary casting too int" b/c from Docs only int is expected.

kiatng: Sorry for late reply, been very busy. `getStoreConfigAsInt()` and `getStoreConfigAsFloat()` is a good idea. I will support a PR for it. However, I have reservation on imposing strict type on existing functions such as `public function setWatermarkImageOpacity(int $imageOpacity)`, it'll break BC. 

I think adding string to `@param int|string $imageOpacity` is correct.

sreichel:
Think it can be closed.

fballiano:
should have been fixed by https://github.com/OpenMage/magento-lts/pull/3736

PR #3730, opened 2024-01-11
Fix autoincrement id fetching cache problem in import/export module

leissbua:
### Description (*)

Import/Export fetches an autoincrement id via getNextAutoincrement($tableName) by mysql SHOW TABLE STATUS.
That table information is cached in standard mySQL and should be cached. In first place I would say, that handling of the autoincrement register of a table should only be managed by mysql, but here we are, dealing with Magento.
The change is simply to set information_schema_stats_expiry to 0 before fetching the autoincrement. This will only be valid for the connection session and does not create a performance problem for the framework in other places, where caching is absolutely wanted. In general. import/export module is the only caller to that function, so to disable the information schema cache in general would not be a good idea.

Make sure information_schema_stats_expiry is set default 86400 seconds in mysql8. Import a configurable product in replace mode, having at least one child. Go to e.g. table catalog_product_super_attribute and check the autoincremented product_super_attribute_id.
Do nothing that could trigger a mysql cache refresh. Do the same import again, and you will end up having exactly the same product_super_attribute_id from the first import.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <p align="center">
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-156-orange.svg" alt="All Contributors"></a>
+<a href="#contributors-"><img src="https://img.shields.io/badge/all_contributors-157-orange.svg" alt="All Contributors"></a>
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/d/total.svg" alt="Total Downloads"></a>
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/license.svg" alt="License"></a>
@@ -589,7 +589,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://juhoholsa.com/"><img src="https://avatars.githubusercontent.com/u/15036353?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Juho Hölsä</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/seifer7"><img src="https://avatars.githubusercontent.com/u/13601073?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kane</b></sub></a></td>
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/Sdfendor"><img src="https://avatars.githubusercontent.com/u/2728018?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kevin Jakob</b></sub></a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/leissbua"><img src="https://avatars.githubusercontent.com/u/68073221?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>leissbua</b></sub></a></td>
     </tr>
   </tbody>
 </table>
--- a/README.md
+++ b/README.md
@@ -589,6 +589,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://juhoholsa.com/"><img src="https://avatars.githubusercontent.com/u/15036353?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Juho Hölsä</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/seifer7"><img src="https://avatars.githubusercontent.com/u/13601073?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kane</b></sub></a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/Sdfendor"><img src="https://avatars.githubusercontent.com/u/2728018?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kevin Jakob</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/leissbua"><img src="https://avatars.githubusercontent.com/u/68073221?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>leissbua</b></sub></a></td>
     </tr>
   </tbody>
--- a/app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
+++ b/app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
@@ -69,7 +69,7 @@ class Mage_ImportExport_Model_Resource_Helper_Mysql4 extends Mage_Core_Model_Res
     {
         if (!self::$instantInformationSchemaStatsExpiry) {
             // Set information_schema_stats_expiry to 0
-            $this->getReadAdapter()->query('SET information_schema_stats_expiry = 0;');
+            $this->_getReadAdapter()->query('SET information_schema_stats_expiry = 0;');
             self::$instantInformationSchemaStatsExpiry = true;
         }
     }
--- a/app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
+++ b/app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
@@ -54,8 +54,6 @@ class Mage_ImportExport_Model_Resource_Helper_Mysql4 extends Mage_Core_Model_Res
         if (empty($entityStatus['Auto_increment'])) {
             Mage::throwException(Mage::helper('importexport')->__('Cannot get autoincrement value'));
         }
-	
 	return $entityStatus['Auto_increment'];
     }
-
 }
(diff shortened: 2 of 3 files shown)


leissbua:
This is a serious bug, that gave us headaches when using import/export. The bug is real, and can easily be reproduced, by doing two small imports (single row) consecutively. Feedback would be appreciated.

fballiano:
@colinmollenhour what do you think about this?

colinmollenhour:
Lots of previous discussion here: https://github.com/OpenMage/magento-lts/pull/1721
Wow, I never saw any info that indicated that `SHOW TABLE STATUS` uses the information_schema stats cache as well so that is quite surprising but it is true, I verified it myself as well.

I think this PR should be accepted as well as #1721, but the SET statement should only be called once per session so a static variable should be added so it can be skipped on subsequent calls.

colinmollenhour:
As an aside, does anyone know of a situation where turning off the information schema stats cache has a real impact on performance? Maybe it should just be recommended to disable it globally as the side effects are not well-known or documented..

kiatng:
I am not sure, but `information_schema_stats_expiry` was introduced in [MySQL 8.0.3](https://dev.mysql.com/blog-archive/further-improvements-on-information_schema-in-mysql-8-0-3/):

image

Would this cause problem with MySQL 5.7?

kiatng:
For ref, alternate solution from https://github.com/mash2/cobby-magento/blob/master/src/app/code/community/Mash2/Cobby/Helper/Resource.php

```php
    /**
     * Find next value of autoincrement key for specified table.
     *
     * @param string $tableName
     * @throws Exception
     * @return string
     */
    public function getNextAutoincrement($tableName, $primaryKey)
    {
        $connection = Mage::getSingleton('core/resource')->getConnection('read');
        $select = $connection->select();
        $select->from($tableName, 'MAX('.$primaryKey.')');
        $result = (int)$connection->fetchOne($select) +1;
        return $result;
    }
```

The cobby module was original referenced in issue #1712, which it seems no longer use `showTableStatus()`.

leissbua:
@colinmollenhour I added a semaphore to only set schema stats expiry to 0 once. I am happy to receive some feeedback.

colinmollenhour:
> I am not sure, but `information_schema_stats_expiry` was introduced in [MySQL 8.0.3](https://dev.mysql.com/blog-archive/further-improvements-on-information_schema-in-mysql-8-0-3/)
> Would this cause problem with MySQL 5.7?

Yes, I think it would.. Great observation! Seems this should be wrapped in an `if` clause based on server version...

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -27,6 +27,13 @@ class Mage_ImportExport_Model_Resource_Helper_Mysql4 extends Mage_Core_Model_Res
     public const DB_MAX_PACKET_SIZE        = 1048576; // Maximal packet length by default in MySQL
     public const DB_MAX_PACKET_COEFFICIENT = 0.85; // The coefficient of useful data from maximum packet length
 
+    /**
+     * Semaphore to disable schema stats only once
+     *
+     * @var bool
+     */
+    private static $instantInformationSchemaStatsExpiry = false;
+
```
fballiano, line 36:
```suggestion

```
to fix phpcs

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,23 @@ public function getMaxDataSize()
...
         return $entityStatus['Auto_increment'];
     }
+
+    /**
+     * Set information_schema_stats_expiry to 0 if not already set.
+     */
+    public function setInformationSchemaStatsExpiry()
+    {
+        if (!self::$instantInformationSchemaStatsExpiry) {
+            // Set information_schema_stats_expiry to 0
```
fballiano, line 73:
```suggestion
```
this is probably not super necessary

fballiano:
> > I am not sure, but `information_schema_stats_expiry` was introduced in [MySQL 8.0.3](https://dev.mysql.com/blog-archive/further-improvements-on-information_schema-in-mysql-8-0-3/)
> > Would this cause problem with MySQL 5.7?
> 
> Yes, I think it would.. Great observation! Seems this should be wrapped in an `if` clause based on server version...

but also what about mariadb or other mysql derivatives?

sreichel:
> but also what about mariadb or other mysql derivatives?

https://dba.stackexchange.com/questions/323392/missing-information-schema-stats-expiry-in-mariadb-server-system-variables

colinmollenhour:
The simple solution would be to just wrap it in a try/catch and ignore any errors. 😁

leissbua:
I added the try catch as @colinmollenhour proposed and followed @fballiano hint and removed the unneeed comment

README.md
```diff
@@ -589,7 +589,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://juhoholsa.com/"><img src="https://avatars.githubusercontent.com/u/15036353?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Juho Hölsä</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/seifer7"><img src="https://avatars.githubusercontent.com/u/13601073?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kane</b></sub></a></td>
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/Sdfendor"><img src="https://avatars.githubusercontent.com/u/2728018?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kevin Jakob</b></sub></a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/leissbua"><img src="https://avatars.githubusercontent.com/u/68073221?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>leissbua</b></sub></a></td>
```
fballiano, line 592: I don't think removing Sdfendor was intended

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,26 @@ public function getMaxDataSize()
...
         if (empty($entityStatus['Auto_increment'])) {
             Mage::throwException(Mage::helper('importexport')->__('Cannot get autoincrement value'));
         }
         return $entityStatus['Auto_increment'];
     }
+
+    /**
+     * Set information_schema_stats_expiry to 0 if not already set.
+     */
+    public function setInformationSchemaStatsExpiry()
```
fballiano, line 74:
```suggestion
    public function setInformationSchemaStatsExpiry(): void
```

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,26 @@ public function getMaxDataSize()
     public function getNextAutoincrement($tableName)
     {
         $adapter = $this->_getReadAdapter();
+        try {
+            $this->setInformationSchemaStatsExpiry();    
+        } catch (Exception $e) {
+            //this will throw an exception for < mysql8
+        }
```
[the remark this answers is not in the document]
leissbua: Yes it does: SQL-Fehler [1193] [HY000]: Unknown system variable 'information_schema_stats_expiry'

leissbua:
I added all your suggestions into the fix, I would be interested in your opinions or other suggestions. From my point of view, this will solve a massive problem with the import/export module that damages configurable/simple relations in catalog_product_super_attribute table.

colinmollenhour (approved):
Looks good to me! Thanks!

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,26 @@ public function getMaxDataSize()
     public function getNextAutoincrement($tableName)
     {
         $adapter = $this->_getReadAdapter();
+        try {
+            $this->setInformationSchemaStatsExpiry();
+        } catch (Exception $e) {
+            //this will throw an exception for < mysql8
+        }
```
fballiano, line 63:
```suggestion
        $this->setInformationSchemaStatsExpiry();
```

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,26 @@ public function getMaxDataSize()
...
         return $entityStatus['Auto_increment'];
     }
+
+    /**
+     * Set information_schema_stats_expiry to 0 if not already set.
+     */
+    public function setInformationSchemaStatsExpiry(): void
+    {
+        if (!self::$instantInformationSchemaStatsExpiry) {
+            $this->_getReadAdapter()->query('SET information_schema_stats_expiry = 0;');
```
fballiano, line 77:
```suggestion
            try {
                $this->_getReadAdapter()->query('SET information_schema_stats_expiry = 0;');
            } catch (Exception $e) {}
```

I'd do it this way, so that the try/catch is not even called unless it's used

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
fballiano: @colinmollenhour exactly, I just sent the suggestion for the other part of the code ;-)

fballiano (dismissed):
fine by me

app/code/core/Mage/ImportExport/Model/Resource/Helper/Mysql4.php
```diff
@@ -49,11 +56,24 @@ public function getMaxDataSize()
...
+
+    /**
+     * Set information_schema_stats_expiry to 0 if not already set.
+     */
+    public function setInformationSchemaStatsExpiry(): void
+    {
+        if (!self::$instantInformationSchemaStatsExpiry) {
+            try {
+                $this->_getReadAdapter()->query('SET information_schema_stats_expiry = 0;');
+            } catch (Exception $e) {}
```
fballiano, line 75:
```suggestion
            } catch (Exception $e) {
            }
```

phpcs being picky

fballiano (approved):
fine by me

colinmollenhour (approved):

PR #3750, opened 2024-01-23
Fixed phpstan in Mage/Customer/controllers/AccountController.php

kiatng:
But I am not sure about this

```
			message: "#^Method Mage_Customer_AccountController\\:\\:preDispatch\\(\\) should return \\$this\\(Mage_Customer_AccountController\\) but empty return statement found\\.$#"
			count: 1
			path: app/code/core/Mage/Customer/controllers/AccountController.php
```

https://github.com/OpenMage/magento-lts/blob/55b14a38987e7988bbf3f3de95cb49faf2e5d28b/app/code/core/Mage/Customer/controllers/AccountController.php#L46-L60


app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -655,7 +655,7 @@ public function confirmAction()
      */
     public function confirmationAction()
     {
-        $customer = $this->_getModel('customer/customer');
```
fballiano, line 658: is the _getModel method still used/necessary at this point?

kiatng: It's still used. Is it better to replace them all with `Mage::getModel()` and marked `_getModel()` as deprecated?

fballiano: If the _getModel is still necessary, why changing it to Mage::getModel()? just curiousity

kiatng: PHPStan thinks `$customer` is `Mage_Core_Model_Abstract` and complains 

> message: "#^Cannot call method sendNewAccountEmail\\(\\) on Mage_Core_Model_Abstract\\|false\\.$#"

sreichel: `instanceof Mage_Customer_Model_Customer` check could solve it. (?)

kiatng: I didn't know that. But for [green coding](https://www.ibm.com/blog/green-coding/) sake,  why waste another CPU cycle.

sreichel: Mhh, or simply ...

```
/** @var Mage_Customer_Model_Customer $customer **/
$customer = $this->_getModel(...);
$collection = $customer->
```

kiatng: @sreichel I am lost, care to explain simply or make a code suggestion so that I can hopefully see the point?

sreichel: Sorry, my bad. Provide code suggestion later.

sreichel: @kiatng as was at the wrong place in code ...

I thought about "green coding"  ...

- adding instanceof-check is only one more function call and takes only a few bytes in memory usage (tested with xhprof)
- adding a new temporally variable may take more memory (not tested yet)
- adding inline DocBlock works, but you tell phpstan how to behave ... in my forked test-project i tried to avoid it)

Solution 1: add inline Docs

```
        /** @var Mage_Customer_Model_Customer $customer */ # added line
        $customer = $this->_getModel('customer/customer');
```

This works, but is just workaround b/c getModel(...) could return "false"  too. (it does not, but from static analyzis it could)

Solution 2: instanceof

```
        if (!$customer instanceof Mage_Customer_Model_Customer) {
            throw new Exception('Customer model must to extend Mage_Customer_Model_Customer');
        }
```

I'd prefer 2nd one.

luigifab:
I vote to mark `_getModel` methods as deprecated to use `getModel` directly.

app/code/core/Mage/Customer/controllers/AccountController.php
kiatng: In the controller's context, if `Mage::getModel('customer/customer')` return `false`, something is seriously off, an exception should be thrown. So I think the fix I have now is the simplest.

sreichel:
`_getModel()` offered the ability to add custom logic w/o overriding alls methods, where it was used.

Not sure if it is a good change,

kiatng:
In case `_getModel()` is used or overriden by 3rd-party code with custom logic, the only issue I can think of is that the IDE would complain about deprecated function, other than that, it shouldn't cause any problem. Or am I wrong?

sreichel:
@kiatng not sure if it is a real use-case, but you could rewrite `_getModel()` to something like ...

```
    public function _getModel($path, $arguments = [])
    {
        if ($path === 'customer/customer') {
            // change arguments or something else
        }
        return Mage::getModel($path, $arguments);
    }
```

... and it would work for every method. With this change you have to override all methods that used _getModel() befor.

kiatng:
@sreichel I am just trying to fix the complaints of my IDE, vscode + PHP Intelephense, complaining about undefined method `$customer->sendNewAccountEmail()`. And it so happens that it's PHPStan related. I do think that `_getModel()` is unnecessary.

sreichel:
@kiatng what wrong with adding `/** @var Mage_Customer_Model_Customer $customer **/` or even better check for instanceof?

Finally i agree with it, but its not what i prefer.

fballiano:
If we remove the _getModel method I think this should be considered breaking change because it would break overridden code, although it's something probably nobody did. So I'd just add `/** @var Mage_Customer_Model_Customer $customer **/` if that solved the problem.

kiatng:
> If we remove the _getModel method I think this should be considered breaking change because it would break overridden code, although it's something probably nobody did.

I agree `_getModel()` shouldn't be removed. At the same time, it's not a very useful function. So in this PR, it is deprecated:
image

fballiano (approved):

ADDISON74 (approved):

Issue #3751, opened 2024-01-23
Unable to save customer in admin console with DOB set to empty


massa-man:
### Preconditions (*)

1. OpenMage 19.5.2
2. PHP 7.4.33
3. MAGE_IS_DEVELOPER_MODE=true

### Steps to reproduce (*)

1. Open a customer record in Admin
2. Leave the date of birth field in the "Account information" section empty and click "save and continue edit"

### Expected result (*)

1. The contact information should be saved
2.

### Actual result (*)

1. The AJAX validation gets stuck, and the server responds to the ajax validation call with this:

```
<pre>Warning: A non-numeric value encountered  in /var/www/html/vendor/shardj/zf1-future/library/Zend/Locale/Math/PhpMath.php on line 97

#0 /var/www/html/vendor/shardj/zf1-future/library/Zend/Locale/Math/PhpMath.php(97): mageCoreErrorHandler(2, 'A non-numeric v...', '/var/www/html/v...', 97, Array)
#1 /var/www/html/vendor/shardj/zf1-future/library/Zend/Locale/Format.php(387): Zend_Locale_Math_PhpMath::Sub('0', '', 0)
#2 /var/www/html/vendor/shardj/zf1-future/library/Zend/Locale/Format.php(666): Zend_Locale_Format::toNumber('', Array)
#3 /var/www/html/vendor/shardj/zf1-future/library/Zend/Filter/NormalizedToLocalized.php(110): Zend_Locale_Format::toFloat('', Array)
#4 /var/www/html/lib/Varien/Data/Form/Filter/Date.php(72): Zend_Filter_NormalizedToLocalized->filter('')
#5 /var/www/html/app/code/core/Mage/Eav/Model/Attribute/Data/Abstract.php(191): Varien_Data_Form_Filter_Date->inputFilter('')
#6 /var/www/html/app/code/core/Mage/Eav/Model/Attribute/Data/Date.php(35): Mage_Eav_Model_Attribute_Data_Abstract->_applyInputFilter('')
#7 /var/www/html/app/code/core/Mage/Eav/Model/Form.php(355): Mage_Eav_Model_Attribute_Data_Date->extractValue(Object(Mage_Core_Controller_Request_Http))
#8 /var/www/html/app/code/core/Mage/Adminhtml/controllers/CustomerController.php(623): Mage_Eav_Model_Form->extractData(Object(Mage_Core_Controller_Request_Http), 'account')
[...]

PR #3752, opened 2024-01-23
Fixed Varien_Data_Form_Filter_Date::inputFilter() and Varien_Data_Form_Filter_Datetime::inputFilter() dealing with empty values

massa-man:
### Description (*)

Fixes issue with saving customer in admin with date of birth set to empty

1. Fixes OpenMage/magento-lts#3751

1. Open a customer record in Admin
2. Leave the date of birth field in the "Account information" section empty and click "save and continue edit"
3. The AJAX validation gets stuck, and the AJX server response shows a warning in php 7.4 and error in php 8.1


README.md
```diff
@@ -589,7 +589,7 @@ Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/d
     <tr>
       <td align="center" valign="top" width="14.28%"><a href="https://juhoholsa.com/"><img src="https://avatars.githubusercontent.com/u/15036353?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Juho Hölsä</b></sub></a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/seifer7"><img src="https://avatars.githubusercontent.com/u/13601073?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kane</b></sub></a></td>
-      <td align="center" valign="top" width="14.28%"><a href="https://github.com/Sdfendor"><img src="https://avatars.githubusercontent.com/u/2728018?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kevin Jakob</b></sub></a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://www.riseart.com/"><img src="https://avatars.githubusercontent.com/u/26821235?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Marcos Steverlynck</b></sub></a></td>
```
ADDISON74, line 592: I think it's a mistake to replace an existing contributor. You should have added yourself.

ADDISON74 (changes requested):

README.md
kiatng, line 592:
```suggestion
      <td align="center" valign="top" width="14.28%"><a href="https://github.com/Sdfendor"><img src="https://avatars.githubusercontent.com/u/2728018?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Kevin Jakob</b></sub></a></td>
      <td align="center" valign="top" width="14.28%"><a href="https://www.riseart.com/"><img src="https://avatars.githubusercontent.com/u/26821235?v=4" loading="lazy" width="100" alt=""/><br /><sub><b>Marcos Steverlynck</b></sub></a></td>
```

sreichel:
Please test this change/fix ... https://github.com/Shardj/zf1-future/pull/402

README.md
massa-man: Sorry, i did not realise I had done that. I run the collaborators JS script as instructed in the PR submission template and it seems it automatically replaced the last contributor instead of adding one at the bottom.

kiatng (dismissed):

ADDISON74:
@massa-man - Maybe you should consider the proposal made by @sreichel.

massa-man:
> @massa-man - Maybe you should consider the proposal made by @sreichel.

I have, see my comment in the Issue thread. While the update to `Zend_Locale_Math_PhpMath` would fix this issue, I still think we should merge this fix into the magento code, to avoid unnecessarily calling expensive code with an empty string. Zend_Locale is expensive, these simple lines of code will improve performance and avoid unnecessary CPU cycles:

```
        $filterInput = new Zend_Filter_LocalizedToNormalized([
            'date_format'   => $this->_dateFormat,
            'locale'        => $this->_locale
        ]);
        $filterInternal = new Zend_Filter_NormalizedToLocalized([
            'date_format'   => Varien_Date::DATE_INTERNAL_FORMAT,
            'locale'        => $this->_locale
        ]);

        $value = $filterInput->filter($value);
        $value = $filterInternal->filter($value);
```

lib/Varien/Data/Form/Filter/Datetime.php
```diff
@@ -51,6 +55,10 @@ public function inputFilter($value)
      */
     public function outputFilter($value)
     {
+        if (!$value) {
```
fballiano, line 58: are this checks ok? we don't use this kind of notation since a while, shouldn't it be if $value === null?

massa-man: Value here is actually an empty string since it comes from form submission, so if you want to tighten the code a bit, I am happy to do the following check instead:

`if ($value === null || $value === '') `

fballiano: I'm wondering cause `if !$value` would also be positive in case of $value = 0

massa-man: Yes, agree `if ($value === null || $value === '')` would be a stronger check. I'll update it.

pquerner: iirc that code is from the mage2 pr. Didnt check if thats still present.

sreichel: > I'm wondering cause `if !$value` would also be positive in case of $value = 0

Just from reading ... 0 needs no further validation/checks/... Looks okay for me.

fballiano (approved):

sreichel:
> to avoid unnecessarily calling expensive code

Yep. My suggestion was to fix it in time.

lib/Varien/Data/Form/Filter/Datetime.php
sreichel: Fully understand ... but 0 value would still cause unnecessary calls. At the end ... lets fix it.

fballiano: there are probably many more values that could be "managed" by us before going to the external library but at that point.. I think that, if there's a value, then it's the external library responsability.

sreichel: Yep. It will be fixed in zf1. If some calls can be avoided her ... why not? +1

fballiano: again, not to replicate the checks that are probably done by the external library. anyway.

sreichel (approved):

ADDISON74 (approved):

massa-man:
Thanks @fballiano for merging. Any chance we can cherry pick this commit into v19 branch? Is v19 branch still being maintained for patches?

I have also been testing v19 on PHP 8 and found a few issues (especially in the recurring profiles module) that I am happy to provide fixes for if we can push them into v19.

fballiano:
v19 is security patches only, we could backport a single one (this) but if we start re-backporting every bugfix we've like 200 to backport and it was voted to freeze v19 (still, security patches are ported anyway for another 2 years give or take)

fballiano:
but if you've installed v19 via composer you could try to add a composer patch of this PR: https://patch-diff.githubusercontent.com/raw/OpenMage/magento-lts/pull/3752.patch

PR #3758, opened 2024-01-25
Removed unused file js/mage/adminhtml/image.js

fballiano:
In the process of reviewing every javascript file, in order to rewrite everything without prototypejs, I found this image.js file that seems to be unused, I couldn't find any place where this file is linked and, removing it, I couldn't find any broken funcionality


kiatng (approved):
I cannot find any reference too. OM has no image resize and rotate in the backend.

luigifab (approved):

Issue #3760, opened 2024-01-26
Where is the code that dispatch the event "before_login_post_redirect"?


kiatng:
### Discussed in https://github.com/OpenMage/magento-lts/discussions/3745

<div type='discussions-op-text'>

<sup>Originally posted by **kiatng** January 22, 2024</sup>
Ref 
https://github.com/OpenMage/magento-lts/blob/e2e738a884f24e09e05e7ad098946ac4d7ff877d/app/code/core/Mage/Oauth/etc/config.xml#L63-L70

I cannot find where is it dispatched:
https://github.com/search?q=repo%3AOpenMage%2Fmagento-lts+before_login_post_redirect&type=code

~Any idea?~
</div>

I can verify that the event is required for OAuth customer (i.e., frontend access). Without it, after the customer login, the customer see the My Account page, which breaks the flow. The correct behavior is for the host to render the authorize page. Once the customer click the "Authorize" button, the host redirect to the consumer based on the callback URL.

Flyingmana:
my guess is it was either only in the enterprise edition, or exclusive to some modules implementing oauth.

fballiano:
I checked an M1.14 enterprise that I've here and there's nothing regarding that event :-(

sreichel:
Any code (or extension) to reproduce for lazy people like me`?

kiatng:
As requested: 

```php
    /**
     * REST API Client TEST
     * @link https://inchoo.net/magento/consuming-magento-rest-zend_oauth_consumer/
     */

    protected $_hostIndex = 0;
    protected $_hosts = [
        [
            'https://somedomain.com.my',
            'fa2d6a11353e5c0c58e179ad00070d64', // consumerKey
            '1c2dd1edd26f7f2413a19fe1449c1b21', // consumerSecret
        ],
        [
            'https://somedomain2.com.my',
            '2ec27fd255f0ef5e90ba47502c9c377a',
            'badb9ace83677378141860fa13a52b23',
        ]
    ];

    protected function _getOauthParams($withCallbackUrl = true): array
    {
        $host = $this->_hosts[$this->_hostIndex][0];
        $consumerKey = $this->_hosts[$this->_hostIndex][1];
        $consumerSecret = $this->_hosts[$this->_hostIndex][2];

        $params = [
            'siteUrl' => "$host/oauth",
            'requestTokenUrl' => "$host/oauth/initiate",
            'accessTokenUrl' => "$host/oauth/token",
[...]

PR #3761, opened 2024-01-26
Fixed bug on OAuth not redirecting to the authorize page after customer login

kiatng:
…er login.

### Description (*)
See issue #3760.

1. Fixes OpenMage/magento-lts#3760

1. Make a OAuth consumer test script such as [here](https://devdocs-openmage.org/guides/m1x/api/rest/introduction.html) or [here](https://inchoo.net/magento/consuming-magento-rest-zend_oauth_consumer/). [My test script](https://github.com/OpenMage/magento-lts/issues/3760#issuecomment-1912933414) was based on the latter.
2. Run customer (not the admin) script. You will be redirected to the host login page, after login, the page stays in host. With this PR, the authorize page is rendered, click the button "Authorize", the page will  be redirected to consumer.

Host login page:
image

After login, the host authorize page:
image

After authorization, it is back to the consumer to display a list of products.


fballiano:
I've used the REST api for a project in the past (Klaviyo and Zakeke use REST to talk to magento, but they don't do anything frontend-related) and I didn't have any problem setting those up also without this event.

Anyway I don't think adding this new event could create any problem so why not.

The naming is also weird, usually the "_before" is at the end of the event name, in that case it's the other way around.

fballiano (dismissed):

sreichel (approved):

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -184,6 +184,7 @@ public function loginPostAction()
      */
     protected function _loginPostRedirect()
     {
+        Mage::dispatchEvent('before_login_post_redirect', ['controller' => $this]);
```
fballiano, line 187: thinking about it, the name of this event is really unclear and since it wasn't working before... can't we just use a better name? should this also be added to the events list?

fballiano (approved):

ADDISON74 (approved):

PR #3762, opened 2024-01-26
Rewrote js/mage/adminhtml/variables.js without prototypejs

fballiano:
This can be tested going to "CMS -> static blocks", opening a block and they using the "insert variable" button, this js handles the opening of this overlay:

Screenshot 2024-01-26 alle 18 45 44

### DIFF
--- a/js/mage/adminhtml/variables.js
+++ b/js/mage/adminhtml/variables.js
@@ -21,7 +21,7 @@ var Variables = {
     overlayHideEffectOptions: null,
     insertFunction: 'Variables.insertVariable',
 
-    init: function (textareaElementId, insertFunction) {
+    init: function(textareaElementId, insertFunction) {
         if (document.getElementById(textareaElementId)) {
             this.textareaElementId = textareaElementId;
         }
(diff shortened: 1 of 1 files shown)


js/mage/adminhtml/variables.js
```diff
@@ -8,7 +8,7 @@
  * @category    Mage
  * @package     Mage_Adminhtml
  * @copyright   Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
- * @copyright   Copyright (c) 2022 The OpenMage Contributors (https://www.openmage.org)
+ * @copyright   Copyright (c) 2022-2024 The OpenMage Contributors (https://www.openmage.org)
```
pquerner, line 11: I still dont like that. A more common thing would be "see COPYRIGHT.md" file?! And only update once and not in 99999 files? :D 
</rant over>

fballiano: that can be discussed but it's outside the scope of this PR.

sreichel: @pquerner +1

At he one hand i'm okay with it, at the other it causes changes that can be avoided (for a cleaner history).

js/mage/adminhtml/variables.js
```diff
@@ -20,27 +20,28 @@ var Variables = {
     overlayShowEffectOptions: null,
     overlayHideEffectOptions: null,
     insertFunction: 'Variables.insertVariable',
-    init: function(textareaElementId, insertFunction) {
-        if ($(textareaElementId)) {
+
+    init: function (textareaElementId, insertFunction) {
```
kiatng, line 24: I am not used to the space after `function`. Is this a thing now?

fballiano: done ;-)

kiatng (approved):

empiricompany (approved):

ADDISON74 (approved):

PR #3767, opened 2024-01-28
Rewrote global search (in backend) without prototypejs

fballiano:
tested and working, both in openmage and legacy themes:

Screenshot 2024-01-28 alle 19 59 52


app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -31,21 +31,37 @@
             <input id="global_search" name="query" type="text" class="input-text" value="<?php if(!empty($query)): ?><?php echo $query ?><?php else: ?><?php echo Mage::helper('core')->quoteEscape($defSearch) ?><?php endif ?>" onfocus="if(this.value=='<?php echo Mage::helper('core')->quoteEscape($defSearch, true) ?>')this.value=''; " onblur="if(this.value=='')this.value='<?php echo $defSearch ?>';" />
             <div id="global_search_autocomplete" class="autocomplete"></div>
             <script type="text/javascript">
-                new Ajax.Autocompleter(
```
kiatng, line 34: The file https://github.com/OpenMage/magento-lts/blob/ecdf1e1325f30cc6f043447dac5f600990233beb/js/scriptaculous/controls.js#L43-L43 

is all about `autocompleter`. Does it make sense to rewrite `controls.js` with vanilla js?

fballiano: I would say no in the sense that I think that the goal was removing both prototype and scriptaculous and since control.js is part of scriptaculous I would drop it altogether (in a million year when this task is finished hehehe)

kiatng: How about creating a separate file: `js/openmage/ajaxcompleter.js`. It'll make it easier to remove prototype dependency in 3rd-party code. I use `Ajax.Completer` in my projects. If it's possible to retain the functionalities of `js/scriptaculous/controls.js` in `js/openmage/ajaxcompleter.js`, then it'll be very nice.

fballiano: that can be done (although it seems to be adding more work to a huge task) but this PR could still be valid

kiatng: It'll make it easier for this as well:  https://github.com/OpenMage/magento-lts/blob/ecdf1e1325f30cc6f043447dac5f600990233beb/js/varien/js.js#L307

fballiano: the thing is, if we rebuild prototype/scripta in another way, we'll kinda have the same library again 🧐 that's a philosophical question that puzzles me

ADDISON74: I can't help but note that these js libraries had their advantages. From a few lines of easy-to-understand code to a larger and more difficult-to-analyze code. But it is the painful way to get rid of unmaintained old things.

fballiano: they were super useful, they would still be (if they were maintained) and they would be 1/3 of the size probably

app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -31,21 +31,37 @@
...
-                    {
-                        paramName:"query",
-                        minChars:2,
-                        indicator:"global_search_indicator",
-                        updateElement:getSelectionId,
-                        evalJSON:'force'
+                document.getElementById('global_search').addEventListener('input', function() {
+                    let query = this.value;
+                    if (query.length >= 2) {
+                        document.getElementById('global_search_indicator').style.display = 'inline-block';
```
andrewcbi, line 37: should add this here:
			document.getElementById('global_search_autocomplete').style.display = 'block';

app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -31,21 +31,37 @@
...
+                        })
+                        .catch(error => {
+                            alert(error);
+                        });
                     }
-                );
-                function getSelectionId(li) {
-                    location.href = li.getAttribute('url');
-                }
+                });
```
andrewcbi, line 64: add this here (otherwise the search results don't go away):
					if (query.length < 2) {
						document.getElementById('global_search_autocomplete').style.display = 'none';
					}

app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -31,21 +31,37 @@
             <input id="global_search" name="query" type="text" class="input-text" value="<?php if(!empty($query)): ?><?php echo $query ?><?php else: ?><?php echo Mage::helper('core')->quoteEscape($defSearch) ?><?php endif ?>" onfocus="if(this.value=='<?php echo Mage::helper('core')->quoteEscape($defSearch, true) ?>')this.value=''; " onblur="if(this.value=='')this.value='<?php echo $defSearch ?>';" />
             <div id="global_search_autocomplete" class="autocomplete"></div>
```
andrewcbi, line 32: add display: none to this this div to start with the right display condition on load of the page

<div id="global_search_autocomplete" class="autocomplete" style="display: none"></div>

andrewcbi:
In file: /skin/adminhtml/default/default/boxes.css around line 628 should also add this to prevent the results jumping around as the indicator shows and hides:

`#global_search_indicator { position:absolute; margin-left:-20px; }`

with the adjustments above I have tested on the default magento theme and works well.

app/design/adminhtml/default/default/template/page/header.phtml
fballiano: done

app/design/adminhtml/default/default/template/page/header.phtml
fballiano: done

app/design/adminhtml/default/default/template/page/header.phtml
fballiano: done

fballiano:
> `#global_search_indicator { position:absolute; margin-left:-20px; }`

done this as well, thank you!

I think it's ready to be merged

app/design/adminhtml/default/default/template/page/header.phtml
```diff
@@ -31,21 +31,37 @@
...
+                            })
+                        })
+                        .then(response => {
+                            return response.text();
+                        })
+                        .then(htmlString => {
+                            document.getElementById('global_search_autocomplete').innerHTML = htmlString;
+                            Array.from(document.querySelectorAll('#global_search_autocomplete li')).forEach(li => {
+                                li.addEventListener('click', function() {
+                                    location.href = this.getAttribute('url');
```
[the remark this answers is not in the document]
fballiano: for sure, but this could be added at a later time if you want, PRs are always welcome

andrewcbi (approved):
have used in production for a bit over a month without issue, I agree that this is ok to merge

fballiano:
thank you!

PR #3773, opened 2024-01-30
Updated workflow

sreichel:
### Description (*)
- run Code-QL only when JS-Files changed
- run php syntax and sonar with all supported php versions (scheduled)
- run phpstan with all supported php versions
- moved config files
- deleted outdated changelog files


sreichel:
Dont know why it fails - the method exists.

sreichel:
Im suck atm, phpunit 9.6.16 is used. Everything looks okay ... any ideas?

fballiano:
I'm not sure about this, from a point of view it's nice to have the workflows run for every php version, on the other hand we're burning a lot of resources (and time) that are not super duper needed since some of the php versions generate have more or less the same behavior.

sreichel:
We changed that in the post to only run lowest/highest supported version b/c phpstan took ~15min. With bleedingEdge and current version it is down to 4 min.

However ... please lets find the phpunit issue first.

sreichel:
Missed that ... 

> PHPUnit 9.5.10 by Sebastian Bergmann and contributors.

Why? I dont see it ...

sreichel:
God damn it. Either grant me permissions or fix it on your own. Phpunit does not work right atm, and i can not abort test runs ....

sreichel:
Sonar runs with 9.5.28, phpunit with 9.5.10 ... but OM should use 9.6.16 ...

Flyingmana:
> Missed that ...
> 
> > PHPUnit 9.5.10 by Sebastian Bergmann and contributors.
> 
> Why? I dont see it ...

its using a different repository for the tests

> Run actions/checkout@v4
> Syncing repository: OpenMage/Testfield

The Testfield project was to have the Tests separated, so they can get reused by other forks and also compared with the original Magento Versions.

The install step mentions a lockfile

> Run composer install --prefer-dist --no-progress --ignore-platform-reqs
> Installing dependencies from lock file (including require-dev)
> Verifying lock file contents can be installed on current platform.

the lockfile of course is old and was not updated for quite a while, as nothing did break in the last 3 years 👀

sreichel:
Makes sense. Are you going to update it?

sreichel:
> its using a different repository for the tests

Why are the tests not in here?

... paid work?

Flyingmana:
> > its using a different repository for the tests
> 
> Why are the tests not in here?
> 
> ... paid work?

The Testfield project was to have the Tests separated, so they can get reused by other forks and (the results) also compared with the original Magento Versions.

PR #3774, opened 2024-01-31
Added whitelist to OAuth consumer callback URLs to allow custom URL scheme

kiatng:
…eme.

### Description (*)
I need to do something similar to this [stackoverflow](https://stackoverflow.com/questions/32067812/magento-oauth-authentication-can-not-handle-custom-url-scheme), I need to provide OAuth to a mobile app, in which the callback URL looks like this `myapp://call-back`. It doesn't work without the fix in the [stackoverflow](https://stackoverflow.com/questions/32067812/magento-oauth-authentication-can-not-handle-custom-url-scheme), or this PR.

This PR does 2 things:
1. Remove the URL validation on  _Callback URL_ and _Rejected Callback URL_ when saving OAuth Consumer in backend: 
image
2. Use the _Callback URL_ as a whitelist for custom scheme. This code ensures that the callback URL is either whitelisted or is a valid URL, so security is not compromised for custom URL schemes.

### DIFF
--- a/app/code/core/Mage/Oauth/Block/Adminhtml/Oauth/Consumer/Edit/Form.php
+++ b/app/code/core/Mage/Oauth/Block/Adminhtml/Oauth/Consumer/Edit/Form.php
@@ -92,7 +92,6 @@ class Mage_Oauth_Block_Adminhtml_Oauth_Consumer_Edit_Form extends Mage_Adminhtml
             'title'     => Mage::helper('oauth')->__('Callback URL'),
             'required'  => false,
             'value'     => $model->getCallbackUrl(),
-            'class'     => 'validate-url',
         ]);
 
         $fieldset->addField('rejected_callback_url', 'text', [
@@ -101,7 +100,6 @@ class Mage_Oauth_Block_Adminhtml_Oauth_Consumer_Edit_Form extends Mage_Adminhtml
             'title'     => Mage::helper('oauth')->__('Rejected Callback URL'),
             'required'  => false,
             'value'     => $model->getRejectedCallbackUrl(),
-            'class'     => 'validate-url',
         ]);
 
         $fieldset->addField(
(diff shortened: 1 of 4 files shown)


fballiano:
where is the whitelist stored? is it intended like a list of protocols/urls that bypass the validate?

kiatng:
> where is the whitelist stored? is it intended like a list of protocols/urls that bypass the validate?

In the column `callback_url` in the table `oauth_consumer`.

### Explanation

(I just learned this stuff, so this explanation is really for myself to digest the stuff.)

The callback URL is an optional param from the OAuth consumer (the client) to OAuth host. When it is not provided, OM as host will get it from the column `callback_url` in the table `oauth_consumer`. See https://github.com/OpenMage/magento-lts/blob/7a3e76522b5b519786f9594a16c1783a9de50e2c/app/code/core/Mage/Oauth/Model/Server.php#L402

Original code before this PR:
image

Lines 399 - 406: If the OAuth consumer doesn't provide the callback URL, it uses the  URL from table `oauth_consumer`; otherwise, it uses the the URL from provided from the OAuth Consumer. The URL is then saved in the table `oauth_token`. (Another possibility is `oob`, which is not affected by this PR. See [this](https://stackoverflow.com/questions/13522497/what-is-oob-in-oauth) and [this](https://developers.google.com/identity/protocols/oauth2/resources/oob-migration) for more info.)

That's the behavior. 

Now we need a way for mobile app with custom scheme to use OAuth. Currently, custom scheme such as `myapp://call-back` throws an exception, only schemes `http://` and `https://` are allowed. The suggested solution from this [stackoverflow](https://stackoverflow.com/questions/32067812/magento-oauth-authentication-can-not-handle-custom-url-scheme) is to add the custom scheme to `Zend_Uri`, which works but it won't fly. Another idea is to disable URL validation in OAuth, but it may compromise security. A colleague suggested a whitelist of the custom scheme. For whitelist to work, the callback URL is a required  param from the OAuth consumer, and the prefix must match what is stored in the column `callback_url` in the table `oauth_consumer`. Note that this value is set in the backend and required the _Current Admin Password_. So it is perfect to be used as whitelist. More importantly, using it as whitelist is BC, it doesn't affect URLs with standard schemes.

fballiano (approved):

ADDISON74 (approved):

PR #3778, opened 2024-02-02
Allow custom demo store notice, fixes #3777

sreichel:

1. Closes OpenMage/magento-lts#3777

### DIFF
--- a/app/code/core/Mage/Page/Helper/Data.php
+++ b/app/code/core/Mage/Page/Helper/Data.php
@@ -19,9 +19,9 @@
  */
 class Mage_Page_Helper_Data extends Mage_Core_Helper_Abstract
 {
-    public const XML_PATH_DEMO_NOTICE = 'design/head/demonotice';
+    public const XML_PATH_STORE_NOTICE_ENABLED = 'design/head/demonotice';
 
-    public const XML_PATH_DEMO_NOTICE_TEXT = 'design/head/demonotice_text';
+    public const XML_PATH_STORE_NOTICE_TEXT = 'design/head/store_notice_text';
 
     public const XML_PATH_BROWSER_CAPABILITIES_JAVASCRIPT = 'web/browser_capabilities/javascript';
 
--- a/app/code/core/Mage/Page/Helper/Data.php
+++ b/app/code/core/Mage/Page/Helper/Data.php
@@ -19,5 +19,11 @@
  */
 class Mage_Page_Helper_Data extends Mage_Core_Helper_Abstract
 {
+    public const XML_PATH_DEMO_NOTICE = 'design/head/demonotice';
+
+    public const XML_PATH_DEMO_NOTICE_TEXT = 'design/head/demonotice_text';
+
+    public const XML_PATH_BROWSER_CAPABILITIES_JAVASCRIPT = 'web/browser_capabilities/javascript';
+
     protected $_moduleName = 'Mage_Page';
 }
--- a/app/code/core/Mage/Page/etc/system.xml
+++ b/app/code/core/Mage/Page/etc/system.xml
@@ -119,6 +119,7 @@
                         <store_notice_text translate="label comment">
                             <label>Store Notice Text</label>
                             <sort_order>90</sort_order>
+                            <frontend_type>textarea</frontend_type>
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
--- a/app/code/core/Mage/Page/etc/system.xml
+++ b/app/code/core/Mage/Page/etc/system.xml
@@ -116,6 +116,14 @@
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
                         </demonotice>
+                        <demonotice_text translate="label comment">
+                            <label>Store Notice Text</label>
+                            <sort_order>90</sort_order>
+                            <show_in_default>1</show_in_default>
+                            <show_in_website>1</show_in_website>
+                            <show_in_store>1</show_in_store>
+                            <depends><demonotice>1</demonotice></depends>
+                        </demonotice_text>
                     </fields>
                 </head>
                 <header translate="label">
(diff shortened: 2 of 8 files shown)


sreichel:
Maybe better to add a model to trim value before saving config, but sounds like overhead?

app/design/adminhtml/default/default/template/page/notices.phtml
```diff
@@ -29,5 +29,9 @@
     </noscript>
 <?php endif ?>
 <?php if ($this->displayDemoNotice()): ?>
+    <?php if (strlen($this->getDemoNoticeText())): ?>
+    <p class="demo-notice"><?php echo $this->getDemoNoticeText() ?></p>
+    <?php else: ?>
     <p class="demo-notice"><?php echo $this->__('This is a demo store. Any orders placed through this store will not be honored or fulfilled.') ?></p>
```
kiatng, line 35: Why not add the default in `app/code/core/Mage/Page/etc/config.xml`?

```xml
    <default>
        <design>
            <head>
                <demonotice_text>This is a demo store. Any orders placed through this store will not be honored or fulfilled.</demonotice_text>
            </head>
        </design>
    </default>
```

sreichel: Nice idea, but strlen () is still required to not have an empty notice. (?)

sreichel: @kiatng this store default text to db. Do we want it?

kiatng: I suggest checking in the backend only, if the notice is empty; "Display Store Notice is enabled but the text is not set, please set it [here](link)."

app/code/core/Mage/Page/Helper/Data.php
```diff
@@ -19,5 +19,11 @@
  */
 class Mage_Page_Helper_Data extends Mage_Core_Helper_Abstract
 {
+    public const XML_PATH_DEMO_NOTICE = 'design/head/demonotice';
+
+    public const XML_PATH_DEMO_NOTICE_TEXT = 'design/head/demonotice_text';
```
kiatng, line 24: Drop the word "DEMO"?
```suggestion
    public const XML_PATH_STORE_NOTICE_ENABLED = 'design/head/demonotice';

    public const XML_PATH_STORE_NOTICE_TEXT = 'design/head/store_notice_text';
```

app/design/adminhtml/default/default/template/page/notices.phtml
sreichel: Thats why im asking .... w/o text set, just fallback to default message, Overhead?

What should happen?

- do not save empty text - forces the user to add something
- save empty text and inform the user

Thanks for input :)

app/code/core/Mage/Page/etc/system.xml
```diff
@@ -108,14 +108,22 @@
...
                             <frontend_type>select</frontend_type>
                             <source_model>adminhtml/system_config_source_yesno</source_model>
                             <sort_order>80</sort_order>
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
                         </demonotice>
+                        <demonotice_text translate="label comment">
+                            <label>Store Notice Text</label>
+                            <sort_order>90</sort_order>
```
kiatng, line 121:
```suggestion
                            <sort_order>90</sort_order>
                            <frontend_type>textarea</frontend_type>
```

app/design/adminhtml/default/default/template/page/notices.phtml
kiatng: May be also always include: "Display Store Notice is enabled, disable it here."

app/code/core/Mage/Page/etc/system.xml
sreichel: Whats the purpose of it?

app/design/adminhtml/default/default/template/page/notices.phtml
sreichel: It alll  started with a simple idea ...

... you want to make it perfect :)

Note to myself ... addd xml form validation ...

app/code/core/Mage/Page/etc/system.xml
kiatng: To expand the use cases and provide flexibility for HTML and fancy store notice with links etc. As an example use case:

```html
<span class="notification-content"><b>Notice to all Users:</b> <br> We will be performing a planned maintenance  from 09:00 PM to 09:30 PM on 23rd December 2024 GMT+8. During this time the store will not be accessible.
</span>
```

app/design/adminhtml/default/default/template/page/notices.phtml
fballiano: I'd also use the default value in XML

ADDISON74: Textarea is the right approach. The default phrase should be in the XML file. I will test it these days.

sreichel:
Change method name to `getStoreNoticeText()`?

ADDISON74:
I would add a comment bellow the textarea that HTML can also be used. This phrase must be in the translation files.

If I insert a static block {{ }} will it work?

sreichel:
Please suggest wording for comment.

Blocks will not work right now ... I'll add it later.

sreichel:
Question: should notice be the same for frontend and backend, or allow different ones?

kiatng:
> Question: should notice be the same for frontend and backend, or allow different ones?

Proposal: same for fronntend and backend, in backend, add hint and link to system configuration.

sreichel:
Addind that link makes it more complex.

With different notices for front/backend you can do this on your own.

sreichel:
> If its for a project of you only, then its ok.
>
> If its a plan in general for OpenMage, instead of renaming it, can you add a parallel one?
> It has a specific usecase and scope, and from our side we should not shift the scope, just because its similar to what we need.

"our side"?

You === our? Who are you speaking for?

ADDISON74:
This one was a good idea.

ADDISON74:
This one was a good idea. The editing of a hard coded paragraph in the Backend was a new feature. For months we have been sitting on the helmets, without bring anything new, a small visual improvement on the Frontend side.

What I don't understand is how certain people express an opinion that could be wrong, you close and delete the PR. Two people have contributed so far to this one and their time spent is wasted.

fballiano:
can't be reopened

PR #3782, opened 2024-02-02
Updated allcontributors badge to new format

fballiano:
As stated in [allcontributors faq](https://allcontributors.org/docs/en/bot/faq) there's a new format for the number-of-contributors badge, which is actually much better cause we don't have to update the number every time.

you can notice I've removed the "align center" because it was breaking the badge (for some weird github reason), the only way to fix it was to add an empty line but... these weird hack are forgotten in 2 seconds and then we'd break it at the next update to the readme.

result of this PR is:
Screenshot 2024-02-02 alle 22 32 03

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,8 @@
-<p align="center">
 [![All Contributors](https://img.shields.io/github/all-contributors/openmage/magento-lts?color=ee8449)](#contributors)
 <a href="https://packagist.org/packages/openmage/magento-lts"><img src="https://poser.pugx.org/openmage/magento-lts/d/total.svg" alt="Total Downloads"></a>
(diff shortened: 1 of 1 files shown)


ADDISON74 (approved):

sreichel:
Maybe you can test allcontributors bot at your repo?

fballiano:
will do but that doesn't have to be related or tied to this PR, which IMHO should be implemented anyway

sreichel (approved):

luigifab (approved):

luigifab (approved):

luigifab:
lol, new achievement unlocked: double approve!

PR #3784, opened 2024-02-03
Some Phpstan fixes

sreichel:
k/t

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Report/Grid/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Block/Report/Grid/Abstract.php
@@ -27,7 +27,7 @@ class Mage_Adminhtml_Block_Report_Grid_Abstract extends Mage_Adminhtml_Block_Wid
     /**
      * Column for grid to be grouped by
      *
-     * @var string
+     * @var string|null
      */
     protected $_columnGroupBy;
 
@@ -85,7 +85,7 @@ class Mage_Adminhtml_Block_Report_Grid_Abstract extends Mage_Adminhtml_Block_Wid
 
     /**
      * Add column to grid
-     * Overriden to add support for visibility_filter column option
+     * Override to add support for visibility_filter column option
      * It stands for conditional visibility of the column depending on filter field values
      * Value of visibility_filter supports (filter_field_name => filter_field_value) pairs
      *
@@ -122,7 +122,7 @@ class Mage_Adminhtml_Block_Report_Grid_Abstract extends Mage_Adminhtml_Block_Wid
     /**
      * Get allowed store ids array intersected with selected scope in store switcher
      *
-     * @return  array
+     * @return array
      */
     protected function _getStoreIds()
     {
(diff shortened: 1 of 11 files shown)


app/design/adminhtml/default/default/template/dashboard/graph.phtml
```diff
@@ -12,34 +12,17 @@
...
-            <img src="<?php echo Mage::getDesign()->getSkinUrl('images/dashboard-close.gif') ?>" onclick="toggleCal('<?php echo $this->getHtmlId() ?>')" />
-        </div>
-    </div>
-    <?php if($this->getCount()): ?>
-    <img src="<?php echo $this->getChartUrl() ?>" />
-    <?php else: ?>
-    <div align="center" style="width:587px;height:300px;"><?php echo $this->__('No Data Found') ?></div>
-    <?php endif ?>
-</div>
-<?php */ ?>
```
ADDISON74, line 37: Is it OK to remove this part of the code?

app/code/core/Mage/Adminhtml/Block/Report/Grid/Abstract.php
```diff
@@ -85,7 +85,7 @@ protected function _getAggregatedColumns()
 
     /**
      * Add column to grid
-     * Overriden to add support for visibility_filter column option
+     * Override to add support for visibility_filter column option
```
fballiano, line 88: I think overridden was correct in this comment

ADDISON74: It depends on how you relate to verb tenses. There is a present and a past action of the same verb.

fballiano: In this case the comment explains that the function was "overridden in order to add support for etc etc", if i'm reading the code correctly

app/design/adminhtml/default/default/template/dashboard/graph.phtml
sreichel: It is not used.

app/code/core/Mage/Adminhtml/Block/Report/Grid/Abstract.php
sreichel: @fballiano this was found with phpstorms spellcheck  ... lets keep it ond move on.

sreichel: @fballiano  either it describes what this metthod

- does (present) or
- did (past)

Im no native english speaker, but "Override" makes more sense for explain the purpose of it.

fballiano: that's why I took time to read the method and explain it here. and I commented because to me, after reading the code, it make more sense before. spell check doesn't have the context.
anyway.

app/design/adminhtml/default/default/template/dashboard/graph.phtml
sreichel: Please mark as resolved to move on.

app/code/core/Mage/Adminhtml/Block/Report/Grid/Abstract.php
sreichel: Me too. :yum: 

As said weird to think about it ... (independent from outcome)

... finally we saved two chars :P

(one for change, one for wrong spelling)

PR #3786, opened 2024-02-03
Removed LICENSE.html in favor of LICENSE.txt

sreichel:
### Description (*)

osl3

At my fork output is different ... only see content from that html file ...

Lets try and see, or revert it ... ?

### Related Pull Requests
1. Fixes OpenMage/magento-lts#1545

Maybe it should not be prefixed w/ license?


fballiano:
can't we just remove that file altogether?

sreichel:
It is used by web installer.

sreichel:
... the txt file is mentioned in every copyright block ...

fballiano:
I know but couldnt we use the txt for the web installer?

Il giorno sab 3 feb 2024 alle 17:48 Sven Reichel ***@***.***>
ha scritto:

sreichel:
Not tested, but I think it's for (html) formatting.

fballiano:
yep, but I think that something like an nl2br would kinda be enough for that use case ;-)

sreichel:
Tested. Looks okay. Only one special char in LICENSE.txt that i wont change,.

fballiano (approved):

kiatng (approved):

PR #3787, opened 2024-02-03
DDEV: added cronjob

sreichel:
### Description (*)

Adds ddev cronjob ...

### Fixed Issues (if relevant)

1. See OpenMage/magento-lts#3780

(have to run `ddev poweroff` or `ddev restart`)

@ADDISON74 pls test without `ddev get ddev/ddev-cron`! Files in place should work.

PR #3788, opened 2024-02-03
.gitignore Sample Data (used by DDEV install)

sreichel:
k/t

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,6 @@ phpstan*.neon
 # dev scripts loaded via composer
 /shell/update-copyright.php
 /shell/translations.php
+
+# sample data
+magento_sample_data_for_1.9.2.4.sql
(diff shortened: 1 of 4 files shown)


ADDISON74:
This PR also took over the modified files from another PR #3787, but we can approve them both. I will provide you my feedback.

.gitignore
```diff
@@ -80,3 +80,6 @@ phpstan*.neon
 # dev scripts loaded via composer
 /shell/update-copyright.php
 /shell/translations.php
+
+# sample data
+magento_sample_data_for_1.9.2.4.sql
```
fballiano, line 85: my personal opinion I don't think we should have every possible file in the gitignore

ADDISON74: When we use the ```ddev openmage-install``` command we get in the root directory that file and it stays there from now on. It should be ignored by git.

fballiano: then maybe it should be deleted afterwards

ADDISON74: It is debatable. That file is good for creating a new database quickly, without having to download the archive from the Internet.

fballiano: git is also very good at showing which files are there and shouldn't be, which allows for a better housekeeping of every project. there more we add to the gitignore the less effective housekeeping is, imho. that's why my opinion is that we shouldn't have every possible file for every possible use in the gitignore, which imho again is already cluttered and could be simplified

sreichel: As @ADDISON74 said, it is used by ddev install. To keep it is optional to repeat installing sample data without download it every time.

sreichel:
On hold ... install sample data can be improved.

PR #3793, opened 2024-02-04
DDEV: install housekeeping

sreichel:
### Description (*)

With this PR everything from sample data gets removed.

With `-k` (keep) flag only compressed `tgz` is kept now (to prevent downloading it again and again).

Before that PR sample data SQL-file was left in root directory.

### Related Pull Requests

1. See OpenMage/magento-lts#3788

1. `ddev openmage-install -q`
2. `ddev openmage-install -q -k`


ADDISON74 (approved):
It works as expected. Thank you.

We need one more green vote to merge this.

Flyingmana (approved):

.gitignore
```diff
@@ -57,6 +57,7 @@
 
 # ddev
 .ddev/config.yaml
+.ddev/.sampleData
```
fballiano, line 60: wouldn't it be possible to just have .ddev, removing the whole folder from git, is it needed for something else?

ADDISON74: I would not exclude this directory from git because it is a test environment, changes can be made in it and would be good to know about. Thanks to your comment I noticed that the DDEV configuration file is excluded and I don't think it's a good idea. I would ignore only the directory where the archive with the Magento Sample package was downloaded.

PR #3794, opened 2024-02-04
Removed openmage's dev-dependencies in PHPUnit workflow

Flyingmana:
in this context the dev dependencies are included in the Testfield repository, having them in both messes up phpunits class loading,


fballiano (approved):
I guess your explanation makes sense

ADDISON74 (approved):

PR #3798, opened 2024-02-05
Whitespace reformat of licence files

fballiano:
Just removed some empty lines and spaces (at the beginning of lines) to make it more similar to the original layout at https://opensource.org/license/osl-3-0-php/

also it looks better in the web install (because it doesn't have the empty first line):
Screenshot 2024-02-05 alle 15 21 48

and somehow it seems to have fixed the copyright symbol that was broken before:
Screenshot 2024-02-05 alle 15 21 53


ADDISON74:
That license information can't be scrolled? I don't see scroll to the right.

fballiano:
@ADDISON74 it can, the mac simply hides the scrollbars after a few secs:
Screenshot 2024-02-05 alle 17 18 21

ADDISON74 (approved):

kiatng (approved):

PR #3799, opened 2024-02-05
DDEV - Removing the sample archive from the root directory

ADDISON74:
### Important Notice
This is PR #3793 deleted by the author. For me this PR was important to install the Magento Sample Pack.

### Description (*)

The sample database was initially saved in the root directory. This PR deletes it or keeps it for a later install.

The **-k** flag is for keeping the downloaded archive to be used with new installations. The archive will be stored in the **.ddev/.sampleData** directory. This one must be ignored by git.

### Related Pull Requests

See OpenMage/magento-lts#3788

1. ```ddev openmage-install -q``` 
2. ```ddev openmage-install -q -k```

Thank you @sreichel for your contribution. I would have preferred to merge the PR created by you instead of wasting time to create it again.

### DIFF
--- a/.ddev/commands/web/openmage-install
+++ b/.ddev/commands/web/openmage-install
@@ -70,7 +70,7 @@ fi
 
 if [[ $INSTALL_SAMPLE_DATA =~ ^(yes|y) ]]; then
   SAMPLE_DATA_URL=https://github.com/Vinai/compressed-magento-sample-data/raw/master/compressed-magento-sample-data-1.9.2.4.tgz
-  SAMPLE_DATA_DIRECTORY="${ROOT}/.sampleData"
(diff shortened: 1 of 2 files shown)


fballiano:
If I'll be the one merging this one I'll make sure that @sreichel gets marked as coauthor in the commit

fballiano (dismissed):
tested, works correctly

fballiano (approved):

ADDISON74:
One more green stamp is required...

fballiano:
Since the original author was @sreichel I'll consider as if we already have 2 green (@ADDISON74's and mine) cause, since only a few people use ddev, it will stay stale here for no reason

Issue #3811, opened 2024-02-08
Fatal error:  Uncaught Error: Call to a member function addData() on null


ADDISON74:
OpenMage 20.3.0
Magento Sample Pack
PHP 8.3
MAGE_IS_DEVELOPER_MODE=1

**Steps to reproduce this issue**

1 - Go to the **Backend > System > Import / Export > Dataflow - Advanced Profiles**
2 - Click on the **[Add New Profile]** button
3 - Fill in "New Profile" and "Action XML" then click the [Save and Continue Edit] button

**You will get this error**

```php
Fatal error:  Uncaught Error: Call to a member function addData() on null in /var/www/html/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php:119
Stack trace:
#0 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(421): Mage_Adminhtml_System_Convert_ProfileController->editAction()
#1 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(255): Mage_Core_Controller_Varien_Action->dispatch()
#2 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Front.php(181): Mage_Core_Controller_Varien_Router_Standard->match()
#3 /var/www/html/app/code/core/Mage/Core/Model/App.php(358): Mage_Core_Controller_Varien_Front->dispatch()
#4 /var/www/html/app/Mage.php(760): Mage_Core_Model_App->run()
#5 /var/www/html/index.php(56): Mage::run()
#6 {main}
  thrown in /var/www/html/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php on line 119
```

This is the PHP file and the line:

/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php:119

kiatng:
I cannot replicate this error. 

Ref https://github.com/OpenMage/magento-lts/blob/c5b35aa9da03142873ddd22af7cca0ebcb077bd9/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php#L30-L53

which is called in https://github.com/OpenMage/magento-lts/blob/c5b35aa9da03142873ddd22af7cca0ebcb077bd9/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php#L108-L110

`$profile` cannot be null because of line 37 `$profile = Mage::getModel('dataflow/profile');`.

[...]

Issue #3810, opened 2024-02-08
Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated


ADDISON74:
OpenMage 20.3.0
Magento Sample Pack
PHP 8.3
MAGE_IS_DEVELOPER_MODE=1

**Steps to reproduce this issue**

1 - Go to the **Backend > System > Import / Export > Dataflow Profiles**
2 - **Click/Edit** on any of the listed profiles. In my case there are 6 profiles

**You will get this error**

```php
Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php on line 95

#0 /var/www/html/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php(95): mageCoreErrorHandler()
#1 /var/www/html/app/design/adminhtml/default/default/template/system/convert/profile/wizard.phtml(220): Mage_Adminhtml_Block_System_Convert_Gui_Edit_Tab_Wizard->getValue()
#2 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#3 /var/www/html/app/code/core/Mage/Core/Block/Template.php(310): Mage_Core_Block_Template->fetchView()
#4 /var/www/html/app/code/core/Mage/Core/Block/Template.php(323): Mage_Core_Block_Template->renderView()
#5 /var/www/html/app/code/core/Mage/Adminhtml/Block/Template.php(67): Mage_Core_Block_Template->_toHtml()
#6 /var/www/html/app/code/core/Mage/Adminhtml/Block/Widget/Container.php(293): Mage_Adminhtml_Block_Template->_toHtml()
#7 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(926): Mage_Adminhtml_Block_Widget_Container->_toHtml()
#8 /var/www/html/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tabs.php(47): Mage_Core_Block_Abstract->toHtml()
#9 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(925): Mage_Adminhtml_Block_System_Convert_Gui_Edit_Tabs->_beforeToHtml()
[...]

PR #3818, opened 2024-02-09
Fixed a couple of null parameter warnings in the ImportExport module

fballiano:
- Fixes https://github.com/OpenMage/magento-lts/issues/3810
- Fixes https://github.com/OpenMage/magento-lts/issues/3811

I created only one PR since both of the issues about about importexport

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
@@ -93,7 +93,7 @@ class Mage_Adminhtml_Block_System_Convert_Gui_Edit_Tab_Wizard extends Mage_Admin
 
         $value = $this->getData($key);
         if ($value === null) {
-            return '';
+            return $default;
         }
 
         return $this->escapeHtml(strlen($value) > 0 ? $value : $default);
--- a/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
@@ -92,6 +92,10 @@ class Mage_Adminhtml_Block_System_Convert_Gui_Edit_Tab_Wizard extends Mage_Admin
         }
 
         $value = $this->getData($key);
+        if ($value === null) {
+            return '';
+        }
+
         return $this->escapeHtml(strlen($value) > 0 ? $value : $default);
     }
 
--- a/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
+++ b/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
@@ -102,34 +102,25 @@ class Mage_Adminhtml_System_Convert_ProfileController extends Mage_Adminhtml_Con
         );
     }
 
-    /**
-     * Profile edit action
-     */
     public function editAction()
     {
         $this->_initProfile();
         $this->loadLayout();
 
+        // Set entered data if was error when we do save
         $profile = Mage::registry('current_convert_profile');
-
-        // set entered data if was error when we do save
         $data = Mage::getSingleton('adminhtml/session')->getConvertProfileData(true);
-
-        if (!empty($data)) {
+        if (!empty($data) && is_object($profile)) {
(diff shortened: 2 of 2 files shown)


app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
```diff
@@ -102,34 +102,25 @@ public function gridAction()
...
     public function editAction()
     {
         $this->_initProfile();
         $this->loadLayout();
 
+        // Set entered data if was error when we do save
         $profile = Mage::registry('current_convert_profile');
-
-        // set entered data if was error when we do save
         $data = Mage::getSingleton('adminhtml/session')->getConvertProfileData(true);
```
kiatng, line 112: The comment should be above `$data` as it is related to it:

```suggestion
        /** @var Mage_Dataflow_Model_Profile $profile */
        $profile = Mage::registry('current_convert_profile');

        // Repopulate user input values in the edit form in case of saving error 
        $data = Mage::getSingleton('adminhtml/session')->getConvertProfileData(true);
```

kiatng:
I think #3811 is unrelated to null deprecation and should be a separate PR should it be a bug. At this point, I am not sure it is.

app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
```diff
@@ -92,6 +92,10 @@ public function getValue($key, $default = '', $defaultNew = null)
         }
 
         $value = $this->getData($key);
+        if ($value === null) {
+            return '';
+        }
+
         return $this->escapeHtml(strlen($value) > 0 ? $value : $default);
```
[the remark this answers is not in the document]
fballiano: @kiatng true, I also did it in a previous PR but I didn't notice here :-D 

I did it in a different way (more similar to the other one, for readability), what do you think?

thanks!!

fballiano:
> I think #3811 is unrelated to null deprecation and should be a separate PR should it be a bug. At this point, I am not sure it is.

I just didn't want to create too many PRs in one go so I thought about merging at least the two that seemed a bit related to each other :-\

app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
fballiano: I reformatted it this way because it seems that comment belongs to this whole logical section:
Screenshot 2024-02-09 alle 09 56 10

I can revert it but those variables and that comment kinda belong together :-)

ADDISON74:
I just tested this PR.

**Case 1**
In the **Advanced Profiles** grid page, press the **[Add New Profile]** button. Then fill in "Profile Name" and "Actions XML" with some data. As soon as you press the **[Save]** button, you get the following error

```php
Fatal error:  Uncaught Error: Call to a member function getId() on null in /var/www/html/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php:117
Stack trace:
#0 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(421): Mage_Adminhtml_System_Convert_ProfileController->editAction()
#1 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(255): Mage_Core_Controller_Varien_Action->dispatch()
#2 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Front.php(181): Mage_Core_Controller_Varien_Router_Standard->match()
#3 /var/www/html/app/code/core/Mage/Core/Model/App.php(358): Mage_Core_Controller_Varien_Front->dispatch()
#4 /var/www/html/app/Mage.php(760): Mage_Core_Model_App->run()
#5 /var/www/html/index.php(56): Mage::run()
#6 {main}
  thrown in /var/www/html/app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php on line 117
```

**Case 2**
In my case there is already a profile in the grid and I will edit it. In the editing page, press any of the **[Save Profile]** or **[Save and Continue Edit]** buttons and the following message is displayed

screenshot

**There are two files affected if this PR is merged.**

ADDISON74:
This PR must be split into two because it does not solve any of the reported problems.

In the case of the **Profiles** grid. Press the **[Add New Profiles]** button. Fill in "Name" and press the **[Save and Continue Edit]** button. The following error appears at the top

screenshot

Issue #3814, opened 2024-02-08
Fatal error:  Uncaught Error: Call to a member function setOnclick() on false


ADDISON74:
OpenMage 20.3.0
Magento Sample Pack
PHP 8.3
MAGE_IS_DEVELOPER_MODE=1

**Steps to reproduce this issue**

1 - Go to the **Backend > Sales > Orders**
2 - Click on any row in the Order grid

**You will get this error**

```php
Fatal error:  Uncaught Error: Call to a member function setOnclick() on false in /var/www/html/app/code/core/Mage/Adminhtml/Block/Sales/Order/View/Giftmessage.php:91
Stack trace:
#0 /var/www/html/app/design/adminhtml/default/default/template/sales/order/view/giftmessage.phtml(40): Mage_Adminhtml_Block_Sales_Order_View_Giftmessage->getSaveButtonHtml()
#1 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#2 /var/www/html/app/code/core/Mage/Core/Block/Template.php(310): Mage_Core_Block_Template->fetchView()
#3 /var/www/html/app/code/core/Mage/Core/Block/Template.php(323): Mage_Core_Block_Template->renderView()
#4 /var/www/html/app/code/core/Mage/Adminhtml/Block/Template.php(67): Mage_Core_Block_Template->_toHtml()
#5 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(926): Mage_Adminhtml_Block_Template->_toHtml()
#6 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(643): Mage_Core_Block_Abstract->toHtml()
#7 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(583): Mage_Core_Block_Abstract->_getChildHtml()
#8 /var/www/html/app/design/adminhtml/default/default/template/sales/order/giftoptions.phtml(16): Mage_Core_Block_Abstract->getChildHtml()
#9 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#10 /var/www/html/app/code/core/Mage/Core/Block/Template.php(310): Mage_Core_Block_Template->fetchView()
#11 /var/www/html/app/code/core/Mage/Core/Block/Template.php(323): Mage_Core_Block_Template->renderView()
#12 /var/www/html/app/code/core/Mage/Adminhtml/Block/Template.php(67): Mage_Core_Block_Template->_toHtml()
[...]

PR #3821, opened 2024-02-09
Fixed call to a member function setOnclick() on false in Mage_Adminhtml_Block_Sales_Order_View_Giftmessage

fballiano:
Fixes https://github.com/OpenMage/magento-lts/issues/3814


kiatng:
Intuitively, if save button is missing, not able to edit and save the gift message, then isn't it correct to throw an error?

fballiano:
@kiatng I don't know how you do it but you got it right, and then debugging I found 2 instances of Mage_Gitftmessage instead of Mage_GiftMessage, which were there reason why this issue was happening.

kiatng (approved):
Good catch!

pquerner:
Magicento boys ;)

ADDISON74 (approved):

PR #3823, opened 2024-02-09
Varien_Directory_Collection::getLastDir(): "$this" can no longer be used in a plain function or method, since PHP 7.1

fballiano:
PHPCS is generating this error:

```
FILE: /Users/fab/Projects/openmage/lib/Varien/Directory/Collection.php
--------------------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
--------------------------------------------------------------------------------
 149 | ERROR | "$this" can no longer be used in a plain function or method
     |       | since PHP 7.1.
     |       | (PHPCompatibility.Variables.ForbiddenThisUseContexts.OutsideObjectContext)
--------------------------------------------------------------------------------
```

This method is static but has a $this in it we shouldn't be there.

Checking the way that method is used in the core:
Screenshot 2024-02-09 alle 09 33 13

I think we can safely remove that part of the method.


Flyingmana:
hmm, will check later, because I have a feeling this static method might have been used in a non-static way

fballiano:
Please everybody note that the EGC workflow fails on every PR with many errors:
https://github.com/OpenMage/magento-lts/actions/runs/7986772621/job/21807865811?pr=3823
and we never noticed

This PR is the start of the fix for all of those.

colinmollenhour (approved):

PR #3825, opened 2024-02-11
Fixed deprecated functionality for trim when creating DataFlow profile

ma4nn:
### Description (*)

This PR fixes the `trim()` deprecation warning when creating new DataFlow profiles with PHP 8.3.

1. Fixes OpenMage/magento-lts#3811

see https://github.com/OpenMage/magento-lts/issues/3811


ADDISON74:
I tested the PR and it solves the issue. I am waiting for opinions if it is the best approach.

fballiano:
now we get 2 PRs for the same issue..
https://github.com/OpenMage/magento-lts/pull/3818

ADDISON74:
> now we get 2 PRs for the same issue.. #3818

@kiatng suggested you to split it into two parts, since we had two separate reports.  If there are two approaches, then we will choose the best one.

@ma4nn - What is your opinion about the solution proposed in PR #3818?

ma4nn:
> now we get 2 PRs for the same issue..

@fballiano but your PR does not solve the issue imho..

@ADDISON74 does the solution in https://github.com/OpenMage/magento-lts/pull/3818 solve the issue for you? In my tests I still got the error. Because as far as I see it, the [change in the wizard tab](https://github.com/OpenMage/magento-lts/pull/3818/files#diff-3996e59de7b5e5887e6ca1169bb7830c8488adc47fcedfa2e65b5e1df86bba10) should only affect the creation of simple DataFlow profiles not advanced ones.

ADDISON74:
@ma4nn - I tested that PR and indeed the problem is not solved. Here are the details https://github.com/OpenMage/magento-lts/pull/3818#issuecomment-1937759303

ADDISON74 (approved):

app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
```diff
@@ -186,6 +186,7 @@ public function saveAction()
                     Mage::helper('adminhtml')->__('The profile has been saved.')
                 );
             } catch (Exception $e) {
+                Mage::logException($e);
```
fballiano, line 189: why add this?

app/code/core/Mage/Dataflow/Model/Profile.php
```diff
@@ -159,7 +159,7 @@ protected function _afterSave()
             ->setActionCode($this->getOrigData('profile_id') ? 'update' : 'create')
             ->save();
         $csvParser = new Varien_File_Csv();
-        $delimiter = trim($this->getData('gui_data/parse/delimiter'));
+        $delimiter = trim((string)$this->getData('gui_data/parse/delimiter'));
```
fballiano, line 162: if delimiter is null, then becomes '', shouldn't be a different default value? or is the right default generated later?

app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
ADDISON74: I asked myself the same question. If we solve the issue, there is no point in writing it in the exceptions file. On the other hand, if the issue is ever reproduced again, although unlikely and we are not in developer mode, it will be written there.

fballiano: I get that but the PR is to fix the trim warning so, like it's been told on other PRs, it should be mixed with other behaviors

ADDISON74: It is fine being removed.

ma4nn: You really want a separate PR for a simple exception logging😃? 
Feel free to modify the PR, but in my opinion the exception must be logged because thats what the exception log is for.

fballiano: Either we do that for all of for nobody. And since i got told to separate when there was a specific reason for merge, we apply it here then. And anyway i dont see why adding something that’s not related to the pr

ma4nn: Hm ok I don't know where you "have been told to", but when it is a rule it should be written somewhere so all contributors know it.

Secondly imho it **is** related to the PR because without it the error message about the `trim()` method would have never been visible.

app/code/core/Mage/Dataflow/Model/Profile.php
ma4nn: Sorry what is different about the logic than before? It's the same as PHP does internally (converting the `null` value to string)..

fballiano: I understand that but I was wondering if that could be fixed in a cleaner way (not cleaner than yours but cleaner than the original implementation)

kiatng:
I think that we should just focus on fixing null deprecation. It means that we avoid fixing any other types: int, bool, array, object, etc. So, the fix should be  checking for null: `$var ?? ''` or `$var === null` or similar. The reason is to preserve the original behavior. For example:

### Original Behavior
```php
$var = trim(['1' => 'foo']);
var_dump($var);
```

PHP8.2
> Fatal error: Uncaught TypeError: trim(): Argument #1 ($string) must be of type string, array given in ...

PHP7.4
> Warning: trim() expects parameter 1 to be string, array given in

### New Behavior 1
```php
$var = trim((string)['1' => 'foo']);
var_dump($var);
```

PHP8.2
> Warning: Array to string conversion in ...

PHP7.4
> Notice: Array to string conversion in ...

### New Behavior 2
```php
$var = trim(['1' => 'foo'] ?? '');
var_dump($var);
```

PHP8.2
> Fatal error: Uncaught TypeError: trim(): Argument #1 ($string) must be of type string, array given in ...

PHP7.4
> Warning: trim() expects parameter 1 to be string, array given in ...

### Conclusion
New Behavior 2 is the same as the Original Behavior. 

Ref https://onlinephp.io/c/72c58

app/code/core/Mage/Adminhtml/controllers/System/Convert/ProfileController.php
kiatng: In general, pull request should be small and focused on addressing one issue or feature at a time. It just makes reviewing easier. In this case, it's probably acceptable.

fballiano:
> I think that we should just focus on fixing null deprecation. It means that we avoid fixing any other types: int, bool, array, object, etc. So, the fix should be checking for null: `$var ?? ''` or `$var === null` or similar. The reason is to preserve the original behavior.

@kiatng in the case of this PR, the actual implementation is ok right?

kiatng:
The whole idea of moving to strict typing and the null deprecation is to avoid bugs, and to catch them early. So, if an array is pass to `trim()`, a fatal error should be thrown. If we do `trim((string)$value)`, depending on error config, there may not be any error. So, to be strict, we should only check for null:

```php
trim($value ?? ''); // other non-acceptable types will throw a fatal error, so that strict-typing makes sense.
```

app/code/core/Mage/Dataflow/Model/Profile.php
fballiano, line 162:
```suggestion
        $delimiter = trim($this->getData('gui_data/parse/delimiter') ?? '');
```

kiatng, line 162:
```suggestion
        $delimiter = trim($this->getData('gui_data/parse/delimiter') ?? '');
```

kiatng (approved):

Issue #3828, opened 2024-02-12
Deprecated functionality: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated


ADDISON74:
OpenMage 20.3.0
Magento Sample Pack
PHP 8.3
MAGE_IS_DEVELOPER_MODE=1

**Steps to reproduce this issue**

1 - Go to the **Backend > Customers**
2 - Click on any row in the grid
3 - Click on the **Wishlist** tab

**You will get this error**

```php
Deprecated functionality: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php on line 26

#0 [internal function]: mageCoreErrorHandler()
#1 /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php(26): htmlspecialchars()
#2 /var/www/html/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php(148): Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description->render()
#3 /var/www/html/app/design/adminhtml/default/default/template/widget/grid.phtml(153): Mage_Adminhtml_Block_Widget_Grid_Column->getRowField()
#4 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#5 /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php(240): Mage_Core_Block_Template->fetchView()
#6 /var/www/html/app/design/adminhtml/default/default/template/customer/tab/wishlist.phtml(16): Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist->getGridParentHtml()
#7 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#8 /var/www/html/app/code/core/Mage/Core/Block/Template.php(310): Mage_Core_Block_Template->fetchView()
#9 /var/www/html/app/code/core/Mage/Core/Block/Template.php(323): Mage_Core_Block_Template->renderView()
#10 /var/www/html/app/code/core/Mage/Adminhtml/Block/Template.php(67): Mage_Core_Block_Template->_toHtml()
[...]

PR #3829, opened 2024-02-12
Fixed null parameter warning in Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php

ADDISON74:
### Fix for issue #3828

OpenMage 20.3.0
Magento Sample Pack
PHP 8.3
MAGE_IS_DEVELOPER_MODE=1

**Steps to reproduce this issue**

1 - Go to the **Backend > Customers**
2 - Click on any row in the grid
3 - Click on the **Wishlist** tab

**You will get this error**

```php
Deprecated functionality: htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php on line 26

#0 [internal function]: mageCoreErrorHandler()
#1 /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php(26): htmlspecialchars()
#2 /var/www/html/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column.php(148): Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description->render()
#3 /var/www/html/app/design/adminhtml/default/default/template/widget/grid.phtml(153): Mage_Adminhtml_Block_Widget_Grid_Column->getRowField()
#4 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#5 /var/www/html/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist.php(240): Mage_Core_Block_Template->fetchView()
#6 /var/www/html/app/design/adminhtml/default/default/template/customer/tab/wishlist.phtml(16): Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist->getGridParentHtml()
#7 /var/www/html/app/code/core/Mage/Core/Block/Template.php(273): include('...')
#8 /var/www/html/app/code/core/Mage/Core/Block/Template.php(310): Mage_Core_Block_Template->fetchView()
#9 /var/www/html/app/code/core/Mage/Core/Block/Template.php(323): Mage_Core_Block_Template->renderView()
#10 /var/www/html/app/code/core/Mage/Adminhtml/Block/Template.php(67): Mage_Core_Block_Template->_toHtml()
#11 /var/www/html/app/code/core/Mage/Core/Block/Abstract.php(926): Mage_Adminhtml_Block_Template->_toHtml()
#12 /var/www/html/app/code/core/Mage/Core/Model/Layout.php(573): Mage_Core_Block_Abstract->toHtml()
#13 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(390): Mage_Core_Model_Layout->getOutput()
#14 /var/www/html/app/code/core/Mage/Adminhtml/controllers/CustomerController.php(498): Mage_Core_Controller_Varien_Action->renderLayout()
#15 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Action.php(421): Mage_Adminhtml_CustomerController->wishlistAction()
#16 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Router/Standard.php(255): Mage_Core_Controller_Varien_Action->dispatch()
#17 /var/www/html/app/code/core/Mage/Core/Controller/Varien/Front.php(181): Mage_Core_Controller_Varien_Router_Standard->match()
#18 /var/www/html/app/code/core/Mage/Core/Model/App.php(358): Mage_Core_Controller_Varien_Front->dispatch()
#19 /var/www/html/app/Mage.php(760): Mage_Core_Model_App->run()
#20 /var/www/html/index.php(56): Mage::run()
#21 {main}
```

screenshot

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
@@ -24,6 +24,6 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description 
     public function render(Varien_Object $row)
     {
         $value = $row->getData($this->getColumn()->getIndex());
-        return is_string($value) ? nl2br(htmlspecialchars($value)) : '';
+        return $value !== null ? nl2br(htmlspecialchars($value)) : '';
     }
 }
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
@@ -23,6 +23,6 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description 
 {
     public function render(Varien_Object $row)
     {
-        return nl2br(htmlspecialchars($row->getData($this->getColumn()->getIndex())));
+        return nl2br(htmlspecialchars((string)$row->getData($this->getColumn()->getIndex())));
     }
 }
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php
@@ -23,6 +23,7 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description 
 {
     public function render(Varien_Object $row)
     {
-        return nl2br(htmlspecialchars((string)$row->getData($this->getColumn()->getIndex())));
+        $value = $row->getData($this->getColumn()->getIndex());
+        return is_string($value) ? nl2br(htmlspecialchars($value)) : '';
     }
 }


ADDISON74:
There are two solutions, I chose the first one. Using a cast is easier to find and replace than a coalescing operator. 

**(string)$variabile** - previous behavior
**$variable ?? ''** - error suppression for TypeError on null only

```return nl2br(htmlspecialchars((string)$row->getData($this->getColumn()->getIndex())));``` 

or 

```return nl2br(htmlspecialchars(($row->getData($this->getColumn()->getIndex()) ?? ''));```

**Comment by kiatng** [comment]
I prefer your second solution although your first one works in this case.

**Comment by ADDISON74** [comment]
@kiatng - the same proposal as yours can be found here https://github.com/sreichel/openmage-future/pull/5/files. 

I made the changes, please do a test.

File: app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Wishlist/Grid/Renderer/Description.php [hunk 1]
```diff
@@ -23,6 +23,7 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Wishlist_Grid_Renderer_Description
 {
     public function render(Varien_Object $row)
     {
-        return nl2br(htmlspecialchars($row->getData($this->getColumn()->getIndex())));
+        $value = $row->getData($this->getColumn()->getIndex());
+        return is_string($value) ? nl2br(htmlspecialchars($value)) : '';
```
kiatng, line 27: This one modifies the original behavior. For example, if `$value` is an array, no error is thrown, making debug difficult. As our intention is to fix null deprecation in PHP8.1, we have to be careful if we are doing more than that. Is the side effect of doing more desirable? In this case, no errors are suppressed, is it what we want?

kiatng, line 27: >  It can't be an array without changing the related code.
 
It's possible for 3rd-party to change the related code. If that's the case, throwing a fatal error is the correct behavior:

```suggestion
        return $value !== null ? nl2br(htmlspecialchars($value)) : '';
```

PHP is moving towards strict typing (PHP9), which is essentially throwing fatal errors on wrong types; instead of implicitly type casting and not throw anything as in pre-PHP7. Throwing fatal errors in strict type is supposed to help debugging and catch the bugs earlier, and help us write dependable code. For null deprecation, we should focus on fixing the `null` part and leaves it to PHP to enforce the typing rule. If we start to do type checking, we would need to throw fatal errors (because that's what strict typing means.)

The problem we have with OpenMage is that `null` is the default value with magic getter. It didn't have any problem and was cool then, but now, sentiments have changed. The [opinion](https://www.techtarget.com/whatis/definition/strongly-typed) is that "weakly typed programming languages can produce erroneous or unpredictable results".

fballiano (approved):

kiatng (approved):

PR #3830, opened 2024-02-13
Updated the existing DDEV documentation

ADDISON74:
This PR aims to update the existing documentation in OpenMage about the DDEV test environment. Please note that this is just a small bag of elementary knowledge to be able to use it.

Those who haven't tried DDEV so far would do well to give it a chance. Since I discovered it a year and a half ago, I haven't tested OpenMage in Linux distributions. It runs quickly in Windows + WSL + Docker and it is so easy to install.

PR #3831, opened 2024-02-13
Added DDEV cronjob file

ADDISON74:
Without this change the OpenMage cronjob cannot be run in the DDEV test environment.

The original PR comes from here https://github.com/sreichel/openmage-future/pull/2 with the permission of its author.

In this PR https://github.com/OpenMage/magento-lts/pull/3830 I updated the DDEV documentation related to the cronjobs because it was no longer valid.

### DIFF
--- a/.ddev/web-build/openmage.cron
+++ b/.ddev/web-build/openmage.cron
@@ -1,0 +1,1 @@
+* * * * * /var/www/html/cron.sh


fballiano (approved):

ADDISON74:
@fballiano - merge this one.

fballiano:
I wouldn't like to have crons running on my dev environment but since you need the ddev-cron plugin in order to make it work and nobody else here seems to use ddev, I'll merge.

PR #3834, opened 2024-02-15
Updated cron shell script to call php binary with php.ini config argument

seifer7:
### Description (*)
Running PHP from CLI does not load the `php.ini` configuration file from the current directory [(confirmed here.)](https://www.php.net/manual/en/configuration.file.php)
This can result in unexpected behaviour.
Simply adding the `-c` parameter to the `php` binary call and specifying the Magento install directory will mean that the CLI php will more accurately reflect the environment that the HTTP server is using.

1. Apply the change to your `cron.sh` file
2. Allow your cron to run
3. Examine the results and ensure no new errors occurred

### DIFF
--- a/cron.sh
+++ b/cron.sh
@@ -33,10 +33,10 @@ INSTALLDIR=`echo $0 | sed 's/cron\.sh//g'`
 # prepend the installation path if not given an absolute path
 if [ "$INSTALLDIR" != "" -a "`expr index $CRONSCRIPT /`" != "1" ];then
     if ! ps auxwww | grep "$INSTALLDIR$CRONSCRIPT$MODE" | grep -v grep 1>/dev/null 2>/dev/null ; then
-    	$PHP_BIN $INSTALLDIR$CRONSCRIPT$MODE &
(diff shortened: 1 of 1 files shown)


pquerner:
This is BC and I think its out of scope to fix that. It should be documented somewhere, yes but not further.

ADDISON74:
I agree @pquerner. As long as there is no php.ini in the root directory of OpenMage, but only a php.ini.sample, I would not make this change in the code. I would mention it in the OpenMage documentation, in the section related to php.ini.sample. Practically, whoever wants to use the php.ini file must change that file manually. It is an omission for a particular situation, nothing more.

fballiano:
quick brainstorm: I think this PR would need to add a parameter to the script (eg: "cron.php -usephpini") which then checks for the existance of that file (or the file is directly passed as an argument) and then added to the cron.php. this way it wouldn't be a breaking change

seifer7:
> quick brainstorm: I think this PR would need to add a parameter to the script (eg: "cron.php -usephpini") which then checks for the existance of that file (or the file is directly passed as an argument) and then added to the cron.php. this way it wouldn't be a breaking change

I like this idea. I'll see what I can come up with and update this.

PR #3858, opened 2024-02-26
Removed Mage_Backup and Mage_PageCache from PHPStan configuration

fballiano:
These 2 modules were removed from the core so they could be removed from PHPStan's excludePaths configuration

### DIFF
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -17,10 +17,6 @@ parameters:
         - lib/Varien
         - shell
     excludePaths:
-        # removed and fixed modules (OM-2811, OM-2813)
-        - app/code/core/Mage/Backup/*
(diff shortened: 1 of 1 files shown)


ADDISON74:
Good observation and I agree with it. We should mention this change publicly, in case someone is using these modules in pre-production.

fballiano:
it's in the README since long time https://github.com/openmage/magento-lts/?tab=readme-ov-file#between-magento-1945-and-openmage-19x

ADDISON74 (approved):

kiatng (approved):

Sdfendor (approved):

S0FTWEX:
Is there any mistake? Because modules Mage_Backup and Mage_PageCache (and many others) are still present in asset  openmage-v20.10.0.zip.

PR #3862, opened 2024-03-01
Removed unused variables from Mage_Api_Model_Server_Handler_Abstract

fballiano:
Those two variables are unused since they're always re-set a few lines after the instanciation.


kiatng (approved):
[PHP doc](https://www.php.net/manual/en/language.variables.basics.php):
> It is not necessary to initialize variables in PHP however it is a very good practice. 

In this case, the variables are of mixed type, so agree to remove.

PR #3865, opened 2024-03-03
Fixed row click issue in product's upsell/crosssell/related grids

ADDISON74:
When you add/remove/set position in the **Related Products / Up-sells / Cross-sells** grids, you may face an important issue without realizing it. As the grid functionality was thought by the Magento team, when you click on a row, regardless of the position where the mouse pointer is, the row will be checked or unchecked. As soon as you press any of the [Save] buttons, products may appear or disappear in the grid by mistake. I faced this issue many times when I was editing the position and I didn't click exactly in the box but a little next to it.

related

This PR solves the issue and limits the row click strictly to the grid cell where the checkbox is (the first column). Unlike the Products grid, where the box must be checked strictly in the square, otherwise the product editing page is loaded, this modification is more permissive and the click can be in the entire cell where the box is located (td element).

### DIFF
--- a/app/code/core/Mage/Oauth/Model/Server.php
+++ b/app/code/core/Mage/Oauth/Model/Server.php
@@ -309,7 +309,7 @@ class Mage_Oauth_Model_Server
             if (self::REQUEST_TOKEN == $this->_requestType) {
                 $this->_validateVerifierParam();
 
-                if (!hash_equals($this->_token->getVerifier(), $this->_protocolParams['oauth_verifier'])) {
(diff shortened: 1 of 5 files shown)


ADDISON74:
@fballiano - I created this PR directly in Github. There are two commits coming from a previous PR. Please fix this issue for me. Thank you.

ADDISON74:
For this one 

/js/mage/adminhtml/grid.js

ADDISON74:
Many thanks. This one is ready for testing.

andrewcbi (approved):

andrewcbi:
Thank you for this one!

fballiano (approved):
tested, works. couldn't find any other grid affected by this change

PR #3866, opened 2024-03-03
Fixed row click issue in Category Products tab

ADDISON74:
When you use the **Category Products** tab in the **Categories** page, you may face an important issue without realizing it. As the grid functionality was thought by the Magento team, when you click on a row, regardless of the position where the mouse pointer is, the row will be checked or unchecked. As soon as you press the **[Save Category]** button, products may appear or disappear in the grid by mistake. I faced this issue many times when I was editing the position and I didn't click exactly in the box but a little next to it.

This PR solves the issue and limits the row click strictly to the grid cell where the checkbox is (the first column).

rowclick


andrewcbi (approved):

andrewcbi:
Thank you for this one too

fballiano (approved):
tested, works and doesn't affect any other functionality

kiatng (approved):

PR #3871, opened 2024-03-05
Removed Mage_Api2_Exception on HTTP_CREATE response in REST API

kiatng:
With this PR, it is now possible to response with  201 HTTP_CREATE on REST POST.

[edit] The following no longer applied for this PR.
### Description (*)
In REST POST, the response is always a redirect:
```json
{
  "headers": {
    "Server": "nginx",
    "Date": "Tue, 05 Mar 2024 10:14:11 GMT",
    "Content-type": "text/html; charset=UTF-8",
    "Transfer-encoding": "chunked",
    "Connection": "close",
    "X-powered-by": "PHP/8.1.27",
    "Set-cookie": "PHPSESSID=l71k09dfreagbgerrku0g9so59; expires=Thu, 07-Mar-2024 02:14:11 GMT; Max-Age=144000; path=/; domain=edumydevapi.scicom.com.my; secure; HttpOnly",
    "Expires": "Thu, 19 Nov 1981 08:52:00 GMT",
    "Cache-control": "no-store, no-cache, must-revalidate",
    "Pragma": "no-cache",
    "Location": ""
  },
  "status": 302,
  "message": "Moved Temporarily",
  "raw_body": "0\r\n\r\n"
}
```

In this PR, it redirects when `_create()` is truthy. If it's falsy:

```json
{
  "headers": {
    "Server": "nginx",
    "Date": "Tue, 05 Mar 2024 09:34:20 GMT",
    "Content-type": "text/html; charset=UTF-8",
    "Transfer-encoding": "chunked",
    "Connection": "close",
    "X-powered-by": "PHP/8.1.27",
    "Set-cookie": "PHPSESSID=516vegn87137l7bl22rv2n409o; expires=Thu, 07-Mar-2024 01:34:19 GMT; Max-Age=144000; path=/; domain=edumydevapi.scicom.com.my; secure; HttpOnly",
    "Expires": "Thu, 19 Nov 1981 08:52:00 GMT",
    "Cache-control": "no-store, no-cache, must-revalidate",
    "Pragma": "no-cache"
  },
  "status": 200,
  "message": "OK",
  "raw_body": "0\r\n\r\n"
}
```

1. ...
2. ...

### Questions or comments
An example of REST POST is create a new customer address:
https://github.com/OpenMage/magento-lts/blob/db3e41cbd31cce36d45a9b71b5d0c3b063897e74/app/code/core/Mage/Customer/Model/Api2/Customer/Address/Rest.php#L24-L62

Line 61: `return $this->_getLocation($address); ` returns a URL, which goes to lines 209-210:
image

But in my use case, I do not want to redirect after a POST. I want s 200 OK response to the client to indicate a successful operation.


fballiano:
is this changing a past behavior so should it be considered a breaking change?

kiatng:
Ref the [mdn doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location):
> The Location response header indicates the URL to redirect a page to.

```http
Location: <url>
```

What happen if url (`$newItemLocation`) is an empty string?  See issue  description:
image

The past behavior doesn't deal with empty location in header, it's always response with 302 redirect. However, different clients handle empty location differently.

Firefox renders "The page isn’t redirecting properly"
Opera & Chrome render a blank page
Mobile app: I'm not sure what happen

I'm not sure if that's desirable. 

The new behavior is to response with a 200 HTML status only if  `$newItemLocation` is falsy, else business as usual with 302 redirect. The question is: is this a breaking change? Normally, `$newItemLocation` is a non-empty string. Under what conditions will it be an empty string? I can only think of 2 possibilities:

1. A bug, a response of 200 won't make the case worse than the unknown client behavior
2. By intention, as in my use case, a 200 response to indicate to the client that the operation is successful

I would say that this PR is backward compatible: if  `$newItemLocation` is not an empty string, same behavior; if it's an empty string, new behavior (also feature) but it shouldn't disrupt existing code (if there is a bug) too badly.

pquerner:
I'd argue overwrite that class to have your wanted behaviour.

kiatng:
> I'd argue overwrite that class to have your wanted behaviour.

I would like to know your argument. Thanks for reviewing.

pquerner:
Other clients may parse the empty location string and now its not there. Or its stripped before sending out (nginx). Tons of different reasons not to touch it here. A class rewrite or configuration would be better suited imho.

fballiano:
I like [this explanation](https://github.com/OpenMage/magento-lts/pull/3871#issuecomment-1982207160) by @kiatng and I think it makes sense, but I've moved it to the `next` branch because it changes past behavior.

To me, it makes sense to have the new behavior, especially browsers render an error/emptypage when location is null

kiatng:
I'm OK with next branch.

fballiano (dismissed):

pquerner:
I still think this is a heavy change for no benefits.
302 empty Location could mean "redirect to self". Now we take it away with no benefit.

Why cant your action set a 200 and empty response?

kiatng:
@pquerner I don't know how to response with 200 on POST without redirection. If you would show me the way, I would be very grateful. Thanks for your review and opinion.

pquerner:
> @pquerner I don't know how to response with 200 on POST without redirection. If you would show me the way, I would be very grateful. Thanks for your review and opinion.

Override `\Mage_Api2_Model_Server::_dispatch`, or `\Mage_Api2_Model_Dispatcher::dispatch`, or `\Mage_Api2_Model_Response` (and implement `sendResponse`) or `echo ""; die();` in your action.

kiatng:
let me think and do some more tests.

kiatng:
I reverted the original changes on not redirect on empty location. I found away to returned [HTTP 201](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201):

> The HTTP 201 Created success status response code indicates that the request has succeeded and has led to the creation of a resource. The new resource, or a description and link to the new resource, is effectively created before the response is sent back and the newly created items are returned in the body of the message, located at either the URL of the request, or at the URL in the value of the [Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) header.

HTTP 201 shouldn't be an error but it's thrown as an error. So this PR fix this. As this is next branch and the fact that the const is not used in the core., it should be safe to change the behavior?

kiatng:
Is HTTP 201 the right way:
image
This is breaking change.

fballiano:
@kiatng does this PR have to be considered draft still?

fballiano:
@kiatng is this still considered a draft?

@pquerner what do you think about this last implementation?

kiatng:
I think this is good to go, especially for next branch. I just encountered 201 for a REST POST in a project when connecting to a host.

pquerner:
Agreed. 201 is better. Could you please remove all outdated info from the description? (I think the PR description should only mention now that for POST requests a HTTP 201 is returned?)
Or put it in a `<details>` block, like so:

<details>
  <summary>Outdated information</summary>
  Hey there
</details>

```
<details>
  <summary>Outdated information</summary>
  Hey there
</details>
```

fballiano (approved):

fballiano:
then I think we had enough discussion and it should be fine to merge :-) thank you!

pquerner:
Yea.. but if someone now takes a look at the PR its all unorganized and not clear (from the top comment) what it does. But whatever..

fballiano:
I'm sure Kiat will arrange it.

anyway there's already a "[edit] The following no longer applied for this PR." in the description, if we remove everything then the discussion in the PR itself won't make sense anymore no?

pquerner:
Thats why I also proposed the block `outdated information` ^^;

fballiano:
@pquerner I tried to apply it, is that ok?

kiatng:
For anyone who encountered the error “oauth_problem=nonce_used” in the REST client due to the 302 redirect, in your V1 file, return 201 like this:

```php
    protected function _create(array $data)
    {
        // ...
        $this->getResponse()->setHttpResponseCode(Mage_Api2_Model_Server::HTTP_CREATED);
    }
```

PR #3872, opened 2024-03-05
Removed deprecated and unused files from lib/Mage

fballiano:
**THIS PR TARGETS V21**

These files seem to be unused (unreferenced anywhere).
lib/Mage/Cache/Backend/Redis.php is also deprecated since many years (I've tested redis cache without this file and it works perfectly)


kiatng:
There may be local.xml with this setting:

```xml
<backend>Mage_Cache_Backend_Redis</backend>
```

That needs to change to this after this PR:
```xml
<backend>Cm_Cache_Backend_Redis</backend>
```

Ref https://github.com/colinmollenhour/Cm_Cache_Backend_Redis.

fballiano:
I'll make sure to write that in the release notes for v21

kiatng:
Found this: https://github.com/OpenMage/magento-lts/blob/304a23c4509c863b7f7784b535d294d93e78dd94/lib/Mage/DB/Mysqli.php#L77-L77

fballiano:
note that the PR targets `next`

kiatng (approved):

PR #3875, opened 2024-03-05
Rewrote getOpenMageVersion() to be faster

fballiano:
The `implode` shouldn't be necessary, so I made a try rewriting the method in order to make it easier and a tiny bit faster

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -176,25 +176,24 @@ final class Mage
      * Gets the current OpenMage version string
      * @link https://openmage.github.io/supported-versions.html
      * @link https://semver.org/
-     *
-     * @return string
      */
     public static function getOpenMageVersion(): string
     {
         $info = self::getOpenMageVersionInfo();
         $versionString = "{$info['major']}.{$info['minor']}.{$info['patch']}";
-        if ($info['stability'] || $info['number']) {
+        $stability = $info['stability'];
+        $number = $info['number'];
+
+        if ($stability || $number) {
             $versionString .= '-';
-            if ($info['stability'] && $info['number']) {
-                $versionString .= implode('.', [$info['stability'], $info['number']]);
+            if ($stability && $number) {
+                $versionString .= "{$stability}.{$number}";
             } else {
-                $versionString .= implode('', [$info['stability'], $info['number']]);
+                $versionString .= $stability . $number;
             }
         }
-        return trim(
-            $versionString,
-            '.-'
-        );
+
+        return trim($versionString, '.-');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


kiatng:
In my test, there is no significant diff. For **one thousand iterations**:

1. original: ~ 250 microsec
2. new: ~270 microsec; yes, it is worse

I suggest to close this PR.

app/Mage.php
```diff
@@ -176,25 +176,24 @@ public static function getVersionInfo()
...
-                $versionString .= implode('', [$info['stability'], $info['number']]);
+                $versionString .= $stability . $number;
             }
         }
-        return trim(
-            $versionString,
-            '.-'
-        );
+
+        return trim($versionString, '.-');
```
Sdfendor, line 196: We should be able to safely use `rtrim()` here, because we know the right side if anything needs potential trimming.

app/Mage.php
```diff
@@ -176,25 +176,24 @@ public static function getVersionInfo()
...
      * @link https://semver.org/
-     *
-     * @return string
      */
     public static function getOpenMageVersion(): string
     {
         $info = self::getOpenMageVersionInfo();
         $versionString = "{$info['major']}.{$info['minor']}.{$info['patch']}";
-        if ($info['stability'] || $info['number']) {
+        $stability = $info['stability'];
```
Sdfendor, line 184: We could use array destructuring here instead of two assignments 184-185. 
```php 
['stability' => $stability, 'number' => $number] = $info; 
```

app/Mage.php
fballiano, line 196:
```suggestion
        return rtrim($versionString, '.-');
```

@Sdfendor true!

fballiano:
this new version doesn't use implode, doesn't use trim, has less variable assignations and from my measurements it's more stable (in terms speed measurements, while when there are functions calling the measurements bounce a bit faster/slower).

Sdfendor (approved):
I like the cleaner structure and will approve. Just to mention it though, what do you think about a version with a switch instead: 
```php
public static function getOpenMageVersion(): string
{
    $info = self::getOpenMageVersionInfo();
    $versionString = "{$info['major']}.{$info['minor']}.{$info['patch']}";

    switch (true) {
        case ($info['stability'] && $info['number']):
            return "{$versionString}-{$info['stability']}.{$info['number']}";
        case ($info['stability']):
            return "{$versionString}-{$info['stability']}";
        case ($info['number']):
            return "{$versionString}-{$info['number']}";
        default:
            return $versionString;
    }
}
```
Does this improve readibility even further? 

An additional benefit for the future (when - hopefully soon - PHP 8 becomes the minimum version) is, that we then can use match(true) and transform the redundant returns into one return in front of the match statement.

fballiano:
mmmmm I kinda don't like the "switch true" :-\\\

PR #3877, opened 2024-03-06
Rewrote `directpost.js` without prototypejs

fballiano:


fballiano:
Directpost has been deprecated by authorize.net since year and it's not even possible to create an account for this service anymore, it should be removed from our codebase. Thus I'm closing this PR.

PR #3878, opened 2024-03-06
[HELP NEEDED] Added support for UPS Rest APIs

fballiano:
3rd of June 2024 UPS will shut down their SOAP APIs as stated on their website here:
https://developer.ups.com/oauth-developer-guide?loc=en_US

**All the websites using UPS will stop working on that date**

This PR is aims to port what M2 did to implement UPS Rest APIs with:
https://experienceleague.adobe.com/docs/commerce-knowledge-base/kb/troubleshooting/known-issues-patches-attached/ups-shipping-method-integration-migration-from-soap-to-restful-api.html

Most of the changes in M2 as in these 2 files:
- https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Ups/Model/Carrier.php (most complicated changes are here)
- https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Ups/Model/UpsAuth.php


fballiano:
I have something working

Screenshot 2024-03-10 alle 18 12 21

fballiano:
it kinda works now:

Screenshot 2024-03-12 alle 21 00 40

at least until creating an order, in the backend is shown correctly:
Screenshot 2024-03-12 alle 21 01 57

I don't know what other tests are needed after the order.

fballiano:
@Flyingmana I think we should add a notification in the the backend for all the users about this

Flyingmana:
we could do it over the notification repository, if it still works

iamniels:
UPS is sending out warnings about this to what it seems like all e-mail addresses in their address book. I have received 3 e-mails about this already.

app/code/core/Mage/Usa/Model/Shipping/Carrier/UpsAuth.php
```diff
@@ -0,0 +1,109 @@
...
+        $headers = [
+            "Content-Type: application/x-www-form-urlencoded",
+            "x-merchant-id: $clientId",
+            "Authorization: Basic " . base64_encode("$clientId:$clientSecret"),
+        ];
+        $authPayload = http_build_query([
+            'grant_type' => 'client_credentials',
+        ]);
+        try {
+            $ch = curl_init();
```
pquerner, line 50: Don't we wanna use the Varien way?

fballiano: tried it and wasn't working for some reason, also in some of this APIs there was already curl calls so I went that route and worked correctly

fballiano:
> UPS is sending out warnings about this to what it seems like all e-mail addresses in their address book. I have received 3 e-mails about this already.

I received a couple of them just yesterday, problem is many store manager are not the ones that will receive this emails, that's why it's important to put a notification in the backend

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +1697,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+    protected function setAPIAccessRequest()
+    {
+        $userId = $this->getConfigData('username');
+        $userIdPass = $this->getConfigData('password');
+        if ($this->getConfigFlag('mode_xml')) {
+            $authUrl = $this->_liveUrls['AuthUrl'];
+        } else {
+            $authUrl = $this->_defaultUrls['AuthUrl'];
+        }
+        return Mage::getModel('usa/shipping_carrier_upsAuth')->getAccessToken($userId, $userIdPass, $authUrl);
```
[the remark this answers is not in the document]
fballiano: I'd keep it as similar as possible to M2 because it will help porting future patches more easily

elidrissidev:
I'm unable to see the username or password fields in UPS config section, how did you fill them up when you were testing?

elidrissidev:
I'll try to work on this in the coming few days and see what I can do.

fballiano:
I did a minor update to fix a phpstan error and updated the task description for the things that still need to be done, the big part should already be done

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +1697,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+    private function processShippingRestRateForItem(
+        array $shipElement,
+        array $allowedMethods,
+        array $allowedCurrencies,
+        array &$costArr,
+        array &$priceArr,
+        bool $negotiatedActive
+    ): void {
+        $code = $shipElement['Service']['Code'] ?? '';
+        if (true or in_array($code, $allowedMethods)) { // TODO CHECK FAILS
```
fballiano, line 1950: this fails because $allowedMethods is populated wrong, and it's wrong because in ups.phtml the changeOriginShipment is not run for some reason, which changes the available methods depending on the origin

elidrissidev: Can you check again, I fixed a JS error that may have been causing this.

elidrissidev:
I pushed a fix for config fields visibility when switching between XML and REST types, basically the common fields should not exist in `onlyUpsXmlElements` nor `onlyUpsRestElements` since they will always get hidden in either type. I feel like for REST the fields for Username/Password should be replaced by Client ID/Secret as users may mistakenly assume they are their UPS account's username and password?

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: great! seems fixed!

fballiano:
great! thanks, I had it working on my environment without the changed to client_id/client_secret but this one is a better naming anyway.

I this it was user/password for the old XML apis but anyway it's not gonna be a problem in the future.

I think we could also remove the "gateway_url" but I'll check on that, anyway it's not a big trouble if it stays.

fballiano:
wow it's also green light from all our workflows

fballiano:
I've continued work on the shipping label part of the module, now what's remaining if the _sendShipmentAcceptRequestRest method (which I just copied from the XML part of the code)

ragnese:
Hi guys. We're users of OpenMage and I was looking to help finish the work for this PR. I've forked @fballiano's fork and was finishing the `_sendShipmentAcceptRequestRest` implementation when I came across a few questions/issues that I feel probably need to be addressed first.

First, with the shutdown of the UPS SOAP API coming in the next month-ish, it seems like there isn't much point in preserving the XML code at all anymore. I feel like we should just replace it with the REST code instead of bothering to let them live side-by-side. That would've probably been the right call when this work first started, but it probably doesn't make sense anymore since the XML code won't work by the time this ends up merged in a deployed.

Other issues I'll address by making comments in this PR.

For some of these suggestions/issues, do you think it would be better for us to lump a bunch of changes into a single PR, or should we suggest changes here and then only include the "ShipmentAccept" work in a separate PR against @fballiano's fork? (Or is there an altogether different approach we should be following?)

Cheers! I'll get to making comments now.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1584,13 +1455,403 @@ protected function _sendShipmentAcceptRequest(SimpleXMLElement $shipmentConfirmR
...
+
+        $shipParams['ShipmentRequest']['Shipment']['Service']['Code'] = $request->getShippingMethod();
+
+        $packagePart = [];
+        $customsTotal = 0;
+        $packagingTypes = [];
+        $deliveryConfirmationLevel = $this->_getDeliveryConfirmationLevel(
+            $request->getRecipientAddressCountryCode()
+        );
+        foreach ($packages as $packageId => $package) {
```
ragnese, line 1757: Is it correct to loop over the packages here?

The function `_doShipmentRequest` is only called by the abstract base class (`Mage_Usa_Model_Shipping_Carrier_Abstract`, line 317). When it calls `_doShipmentRequest`, it's inside a loop over `$request->getPackages()` already, so this implementation will loop over all packages quadratically. The original XML implementation (`_doShipmentRequestXML`=>`_formShipmentRequest` just looks for the "current" package on the $request object that is set by the base class, e.g., `$request->getPackageId()`, `$request->getPackageParams()`, etc.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +2054,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+        if ($this->getConfigFlag('include_taxes')) {
+            $rateParams['RateRequest']['Shipment']['TaxInformationIndicator'] = "Y";
+        }
+
+        if ($serviceCode !== null) {
+            $rateParams['RateRequest']['Shipment']['Service']['code'] = $serviceCode;
+            $rateParams['RateRequest']['Shipment']['Service']['Description'] = $serviceDescription;
+        }
+
+        foreach ($rowRequest->getPackages() as $package) {
```
ragnese, line 2150: Same concern as my comment in `_formShipmentRestRequest` (https://github.com/OpenMage/magento-lts/pull/3878/files#r1581854006). The base class seems to call the `_getQuotes` method for a single package, and the XML implementation of this also treats this as a single package request.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +2054,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+        $requestOption = $params['10_action'];
+        $headers = [
+            "Authorization: Bearer {$accessToken}",
+            "Content-Type: application/json"
+        ];
+        $debugData = [
+            'request' => $ratePayload
+        ];
+
+        try {
```
ragnese, line 2186: This is also something I'm trying to wrap my head around. Why are the `curl_*()` functions wrapped in a try-catch? These functions won't *ever* throw an Exception, AFAIK.

There are two cases where it might throw an `Error`, but it should never throw an `Exception`. If you pass an invalid url to `curl_init()`, then it *might* return `false` (though I've never seen it do that, even if you pass in total garbage); if you *then* pass that `false` into `curl_setopt` or `curl_exec`, then you might see a thrown `TypeError`. Or, if you try to call any of the `curl_*` functions when the php-curl extension isn't loaded, you'll also get a `FatalError` thrown.

But neither of those are subclasses of `Exception`, so the try-catches really don't make any sense to me. Am I missing something?

For error handling the cURL functions, we should be checking `curl_errno()` and `curl_error()`, or at least checking if `$responseData === false`.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +2054,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+            curl_setopt($ch, CURLOPT_URL, $url . $version . "/" . $requestOption);
+            curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+            curl_setopt($ch, CURLOPT_HEADER, 0);
+            curl_setopt($ch, CURLOPT_POST, 1);
+            curl_setopt($ch, CURLOPT_HTTPHEADER, $headers);
+            curl_setopt($ch, CURLOPT_POSTFIELDS, $ratePayload);
+            curl_setopt($ch, CURLOPT_TIMEOUT, 30);
+            curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, $this->getConfigFlag('verify_peer'));
+            $responseData = curl_exec($ch);
+            curl_close($ch);
```
ragnese, line 2197: If we were going to handle catching Exceptions/Errors, then `curl_close()` calls should be in a `finally {}` block, anyway.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1793,4 +2054,398 @@ protected function _getDeliveryConfirmationLevel($countyDest = null)
...
+            $debugData['result'] = ['error' => $e->getMessage(), 'code' => $e->getCode()];
+            Mage::logException($e);
+        }
+
+        if (!isset($responseData)) {
+            $responseData = '';
+        }
+
+        $this->_debug($debugData);
+        return $this->_parseRestResponse($responseData);
```
ragnese, line 2208: $responseData could be `false` here.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1584,13 +1455,403 @@ protected function _sendShipmentAcceptRequest(SimpleXMLElement $shipmentConfirmR
...
+        );
+
+        $result = new Varien_Object();
+        $this->_prepareShipmentRequest($request);
+        $rawJsonRequest = $this->_formShipmentRestRequest($request);
+        $accessToken = $this->setAPIAccessRequest();
+        $this->_debug(['request_quote' => $rawJsonRequest]);
+
+        if ($this->getConfigFlag('mode_xml')) {
+            $shipConfirmUrl = $this->_liveUrls['ShipRestConfirm'];
```
ragnese, line 1550: Why do some URLs use these hardcoded "default" vs "live" URLs, but some use the config.xml approach? Specifically, in for shipconfirm, the XML code checks `$this->getConfig('shipconfirm_xml_url')` and then falls back to `$this->_defaultUrls['ShipConfirm']`. Why doesn't this code add a config for 'shipconfirm_rest_url', and why doesn't the XML code ever use the `_liveUrls['ShipConfirm']` value?

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1584,13 +1455,403 @@ protected function _sendShipmentAcceptRequest(SimpleXMLElement $shipmentConfirmR
...
+        /** Rest API Payload */
+        $headers = [
+            "Authorization: Bearer {$accessToken}",
+            "Content-Type: application/json"
+        ];
+        $debugData = [
+            'request' => $rawJsonRequest
+        ];
+
+        try {
```
ragnese, line 1564: Same question about try-catching cURL stuff as above.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1584,13 +1455,403 @@ protected function _sendShipmentAcceptRequest(SimpleXMLElement $shipmentConfirmR
...
+            curl_close($ch);
+        } catch (Exception $e) {
+            $debugData['result'] = ['error' => $e->getMessage(), 'code' => $e->getCode()];
+            Mage::logException($e);
+        }
+
+        if (!isset($responseData)) {
+            $responseData = '';
+        }
+        $responseData = json_decode($responseData, JSON_THROW_ON_ERROR);
```
ragnese, line 1584: This call is incorrect. The flags are the fourth argument. This call should be `json_decode($responseData, false, 512, JSON_THROW_ON_ERROR);`.

It should also be wrapping in a try-catch if we don't want this method to be able to actually hard-fail.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: same thing, it's as similar as possible to the M2 version and probably there could be a situation where there are multiple packages

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: I've no idea why it was done this way but it's how the original module was done and how the M2 patch treats it

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: done

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: done, thanks!

fballiano:
> First, with the shutdown of the UPS SOAP API coming in the next month-ish, it seems like there isn't much point in preserving the XML code at all anymore. I feel like we should just replace it with the REST code instead of bothering to let them live side-by-side. That would've probably been the right call when this work first started, but it probably doesn't make sense anymore since the XML code won't work by the time this ends up merged in and deployed.

we have to preserve it until june, then I'll delete it for sure ;-)
but we also have to release this PR before june

> For some of these suggestions/issues, do you think it would be better for us to lump a bunch of changes into a single PR, or should we suggest changes here and then only include the "ShipmentAccept" work in a separate PR against @fballiano's fork? (Or is there an altogether different approach we should be following?)

as you wish, but let's keep the code as similar as possible to the M2 patch for future bugfix

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: I've removed the JSON_THROW_ON_ERROR at the moment, thanks!

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: I'll remove the try/catch but it was already used in the XML implementation of the module

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
ragnese: Fair enough. Upon looking at the Magento2 code, I see the similar approach.

However, I still think what's done here doesn't match M2. The M2 code has methods `getShipConfirmUrl()`, and `getShipAcceptUrl()`. `getShipConfirmUrl()` does seem to grab a config value for the url before falling back to the live vs default URLs:

```
    public function getShipConfirmUrl()
    {
        $url = $this->getConfigData('url');
        if (!$url) {
            if ($this->getConfigData('is_account_live')) {
                if ($this->getConfigData('type') == 'UPS_XML') {
                    $url = $this->_liveUrls['ShipConfirm'];
                } else {
                    $url = $this->_liveUrls['ShipRestConfirm'];
                }

                return $url;
            } else {
                if ($this->getConfigData('type') == 'UPS_XML') {
                    $url = $this->_defaultUrls['ShipConfirm'];
                } else {
                    $url = $this->_defaultUrls['ShipRestConfirm'];
                }

                return $url;
            }
        }

        return $url;
    }
```

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
ragnese: The M2 UPS Carrier class also overrides the parent class's `requestToShipment` and `returnOfShipment` methods. The for-loop is in those base methods, and the M2 code decided to override those so that the loop could be moved here.

We need to override those methods as well if we're going to loop here.

ragnese: M2 also updates the XML code to use multiple packages, which isn't being done here. So, if we do move to make this class more closely match M2's, then the XML code needs to be changed as well.

fballiano: I don't see how these 2 things are related, XML code will be dead in a few weeks

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
fballiano: I lost a bit of control here, but this patch is already working, you could make suggestions to the code or a PR to my repo if you want because this way it's harder for me to check.

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
ragnese: They're related because the behavior of the XML methods should match the behavior of the Rest methods. If one set of methods loops over the request packages, then they both should.

However, I'd argue that neither of them should.

For example, look at the `collectRates()` method. In OpenMage, `collectRates()` is called by ` Mage_Shipping_Model_Shipping::collectCarrierRates`, *inside a foreach* loop.

For the M2 code, the method `Magento\Shipping\Model\Shipping::collectCarrierRates` calls `collectRates()` on the Carrier, but *there is no foreach loop*.

So, for M2, they moved the package loop into the Carrier implementations. It doesn't make sense to loop over the packages in this class unless we're also going to change the other carriers and all of the code that uses carrier models.

ragnese:
Don't we also need to do `getTracking` for Rest?

fballiano:
About this PR, I'm waiting for some tests results by a customer, it seems to be working fine (for sure it's not perfect but it is what it is).

Since 3rd June is near the corner, in case I don't get any feedback from the community here, I'll still merge this PR and release 20.7.0 with a note about this particular feature.

And after june we'll remove the code about XML and re-release.

fballiano:
It's 7th may, regardless of the tests results I think it's time this PR gets merged and released to give merchants a few weeks to test it out and give feedback.

I'll merge and release 20.7.0 with a clear statement about this UPS module.

Let's hope for the best.

ragnese:
I know it's a little late now, but I just thought of something while I was working on bringing this change into our production system: Should we have used the "live" URLs (onlinetools.ups.com) for the defaults in config.xml, instead of the testing URLs (wwwcie.ups.com)?

Do you think it would be worth doing a quick patch for that?

fballiano:
@ragnese if you configure it in "live" mode it uses the live URL right? I think it does

fballiano:
I've received this:

Screenshot 2024-05-07 alle 18 00 08

ragnese:
First about the config params: The way we have it now, it'll use what's set in the config if it's there. Then it'll fallback to either the live or default URLs if the config value is falsey.

I'll look at the tracking error. It looks like the request is actually returning an error response and we're just not handling that correctly. So, we need to handle the bad response better, but we also need to figure out what the actual issue is with the request. I'm looking into it now.

fballiano:
what if we remove the configuration of the API URLs? I don't even know what it was done that way :-\

ragnese:
I agree that making them configurable was always a weird choice. But, that could turn into a breaking change for anyone who might be pointing to some custom proxy to escape a firewall or something... who knows what weird things people may be doing...

On the tracking error, it looks like you might have the error response string in that screenshot, but it's cut off (the "#1" line of the stack trace). Could you post what that response actually says? I bet it's that we don't have tracking permission on the auth. I noticed that the Admin panel config for the clientId and clientSecret list the permissions to acquire when setting up a UPS app, but it doesn't list "tracking" as one of the needed permissions, so that's probably it.

fballiano:
@ragnese the error is

Array ( [code] => TV0011 [message] => Missing transactionSrc )

ragnese:
Okay, that's an easy fix. We just need to add any kind of identifier that is "us". For other code I've used, we used our UPS merchantId, but since that's not something we ask for in the config, we could just put something like "OpenMage" and I think that'll be fine.

sebadiginet:
HI

There is one bug/problem - currency change is not working. Our calculation from UPS we get in PLN. We have same result in debug shipping_ups.log in PLN and EUR currency websites so its proper answer from API UPS. But checkout in website EUR currency we have not available any shipping method by UPS. Normally its working without problem with exchange made in magento for both previous version of UPS and for Fedex.

iamniels:
I ran into a bug, I got this error:

> Code: 110548
> Message: A shipment cannot have a KGS/IN or LBS/CM or OZS/CM as its unit of measurements

I checked the request and it contained:


```
"Package": [
                {
                    "PackagingType": {
                        "Code": "00",
                        "Description": "Packaging"
                    },
                    "Dimensions": {
                        "UnitOfMeasurement": {
                            "Code": "IN",
                            "Description": "Inches"
                        },
                        "Length": "5",
                        "Width": "5",
                        "Height": "5"
                    },
                    "PackageWeight": {
                        "UnitOfMeasurement": {
                            "Code": "KGS"
                        },
                        "Weight": "1.265"
                    }
                }
```
By removing the Dimensions the error has been resolved.

sebadiginet:
I got error today when try to configure firecheckout extension : 
looks that error with that line in Ups.php : $methods[$methodCode] = $methodData->getText();

Fatal error: Uncaught Error: Call to a member function getText() on string in /opt/lampp/htdocs/hpc/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php:1344 Stack trace: #0 /opt/lampp/htdocs/hpc/app/code/local/TM/FireCheckout/Model/System/Config/Source/Shipping/Allmethods.php(31): Mage_Usa_Model_Shipping_Carrier_Ups->getAllowedMethods() #1 /opt/lampp/htdocs/hpc/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php(463): TM_FireCheckout_Model_System_Config_Source_Shipping_Allmethods->toOptionArray(false) #2 /opt/lampp/htdocs/hpc/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php(229): Mage_Adminhtml_Block_System_Config_Form->initFields(Object(Varien_Data_Form_Element_Fieldset), Object(Mage_Core_Model_Config_Element), Object(Mage_Core_Model_Config_Element)) #3 /opt/lampp/htdocs/hpc/app/code/core/Mage/Adminhtml/Block/System/Config/Form.php(164): Mage_Adminhtml_Block_System_Config_Form->_initGroup(Object(Varien_Data_Form), Object(Mage_Core_Model_Config_Element), Object(Mage_Core_Model_Config_Element)) #4 /opt/lam in /opt/lampp/htdocs/hpc/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php on line 1344

fballiano:
@sebadiginet already solved here: https://github.com/OpenMage/magento-lts/pull/4013

please open new issues because it's much more difficult to track comments on old PRs

sebadiginet:
There is possibility to add estimated delivery time for new version of ups ? I have it ready working for fedex also with newer wsdl RateService_v31.wsdl which support new shipping methods like fedex regional - i can share with openmage to make upgrade for new version. Its described for example here https://magento.stackexchange.com/questions/147208/ups-api-estimated-delivery-date

PR #3917, opened 2024-03-29
RWD: better accessibility score removing double instance of the currency switcher from the DOM

fballiano:
In RWD we have 2 instances of the currency switcher in the DOM, one in header, one in footer:
Screenshot 2024-03-29 alle 19 26 36

This is not ideal because the DOM is big a google likes smaller DOMs, also it triggers an accessibility warning:
Screenshot 2024-03-29 alle 19 27 36

This is done because, on mobile, the currency switcher is shown in the footer instead of the header:
Screenshot 2024-03-29 alle 19 24 58

But in my opinion, a part from the DOM/accessibility problem, it's not very usable.

This PR removes the second currency switcher and handles the visibility of the one in the header for mobile view.

This is the header before this PR (with currency switcher in footer):
Screenshot 2024-03-29 alle 19 24 53

And this is after this PR:
Screenshot 2024-03-29 alle 19 24 35


fballiano:
I've just inverted the order of the 2 combobox because language-currency looks much better than currency-language:

Screenshot 2024-04-11 alle 10 17 47

Sekiphp (dismissed):
We are not using RWD in our websites, but this make sense to me :-)

fballiano:
I've just rebased it on `next` since technically could be considered a breaking change

PR #3920, opened 2024-04-01
PHP8.2 is now the minimum version for v21

fballiano:
v21 will have to be released sometimes (I hope very soon) but there's no point in releasing it in a few months supporting a PHP version that's already out of the active support.

Screenshot 2024-04-01 alle 07 39 28

Also, 8.1-8.2 don't have a lot of differences so, for people that need to upgrade to 8.1, it doesn't take a bigger effort to upgrade to 8.2.

Also, this PR allows for removing some of the polyfills we needed, making the final installation size smaller.

Side benefit, with this PR, PHPStan has to check 3800 files instead of 4300

### DIFF
--- a/.github/workflows/php-cs-fixer.yml
+++ b/.github/workflows/php-cs-fixer.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.1
+          php-version: 8.2
 
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.1
+          php-version: 8.2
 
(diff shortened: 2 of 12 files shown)


empiricompany (approved):

kiatng (dismissed):

tmewes (approved):

ADDISON74:
Even this PR benefits from approvals, the PHPStan error must be fixed before merging.

Related to this change, in production I use Debian 10, 11, 12 servers that have LTS's for PHP versions 7.3, 7.4, 8.2. For the first two I will benefit from support even after the period ends. Increasing the minimum requirement must be done with caution. I recently had to deal with a store where I started from PHP 7.0 perfectly functional. Moving it to 8.2 was crazy, so I left it on 7.4 for a while until I evaluate how long a complete migration will take.

fballiano:
@ADDISON74 phpstan errors comes from version 66 and will be fixed when https://github.com/OpenMage/magento-lts/pull/3880 gets merged, which by the way I should merge anyway.

phpstan 66 is in this PR because I had to run a full composer install since the requirements are changed (removed the polyfills)

fballiano:
I'll merge since it's approved and all checks are green

PR #3923, opened 2024-04-02
Add the column ID name to the <span> in absence of a link

loekvangool:
It's possible to determine the column name for admin grids, but not if they are not sortable. I propose to add this.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
@@ -115,7 +115,7 @@ abstract class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Abstract extends
     /**
      * @return string
      */
-    public function renderHeader(): string
+    public function renderHeader()
     {
         if ($this->getColumn()->getGrid()->getSortable() !== false && $this->getColumn()->getSortable() !== false) {
             $className = 'not-sort';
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
```diff
@@ -115,7 +115,7 @@ protected function _getInputValue(Varien_Object $row)
     /**
      * @return string
      */
-    public function renderHeader()
+    public function renderHeader(): string
```
fballiano, line 118: doesn't modifying the signature of public method break every module that overrides this method?

loekvangool: Indeed, I forgot to remove that yesterday

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
```diff
@@ -128,7 +128,7 @@ public function renderHeader()
                    . '" class="' . $className . '"><span class="sort-title">'
                    . $this->escapeHtml($this->getColumn()->getHeader()) . '</span></a>';
         } else {
-            $out = $this->escapeHtml($this->getColumn()->getHeader());
+            $out = '<span name="' . $this->getColumn()->getId() . '">'.$this->escapeHtml($this->getColumn()->getHeader()) . '</span>';
```
fballiano, line 131:
```suggestion
            $out = '<span name="' . $this->getColumn()->getId() . '">' . $this->escapeHtml($this->getColumn()->getHeader()) . '</span>';
```

fballiano:
Wouldn't it be possible to add the name to the `th` instead of adding a new DOM element?

Also, https://www.w3schools.com/tags/att_name.asp name isn't a valid attribute for `span` or for `a` so... why did they add it to the `a` element?

Issue #3925, opened 2024-04-05
PHP Error in frontend with improper request params


ioweb-gr:
### Preconditions (*)
1. Openmage 20.5.0

### Steps to reproduce (*)
1. Visit url https://demo.openmage.org/men/shirts.html?occasion=30,31 and check the php error

### Expected result (*)
1. The page opens and the filter is not applied but do not show error to the user. 

### Actual result (*)
1. PHP Error 

```
 Uncaught TypeError: strlen(): Argument #1 ($string) must be of type string, array given in /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Catalog/Model/Layer/Filter/Attribute.php:81\nStack trace:\n#0 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Catalog/Block/Layer/Filter/Abstract.php(81): Mage_Catalog_Model_Layer_Filter_Attribute->apply()\n#1 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Catalog/Block/Layer/Filter/Abstract.php(63): Mage_Catalog_Block_Layer_Filter_Abstract->_initFilter()\n#2 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Catalog/Block/Layer/View.php(125): Mage_Catalog_Block_Layer_Filter_Abstract->init()\n#3 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Core/Block/Abstract.php(299): Mage_Catalog_Block_Layer_View->_prepareLayout()\n#4 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Core/Model/Layout.php(474): Mage_Core_Block_Abstract->setLayout()\n#5 /var/www/vhosts/example.gr/httpdocs/app/code/core/Mage/Core/M.
```

fballiano:
I've created https://github.com/OpenMage/magento-lts/pull/3926 to fix this, please review the PR when you can

ioweb-gr:
Thanks I'll be able to test it tomorrow

PR #3926, opened 2024-04-05
Fixed null parameter warning in Mage_Catalog_Model_Layer_Filter_Attribute

fballiano:
Fixes https://github.com/OpenMage/magento-lts/issues/3925

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Layer/Filter/Attribute.php
+++ b/app/code/core/Mage/Catalog/Model/Layer/Filter/Attribute.php
@@ -77,7 +77,11 @@ class Mage_Catalog_Model_Layer_Filter_Attribute extends Mage_Catalog_Model_Layer
         if (is_array($filter)) {
             return $this;
         }
-        $text = $this->_getOptionText($filter) ?? '';
+        $text = $this->_getOptionText($filter);
+        if (!is_string($text)) {
+            return $this;
+        }
+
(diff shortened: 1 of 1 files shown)


ioweb-gr:
Nope it didn't work for me for a URL of the type 

/mobile-phones?core_count=41&gdiktia=168&nfcsosto=376&ram=50%2C54

or 
/mobile-phones?core_count=41&gdiktia=168&nfcsosto=376&ram=50,54

ioweb-gr:
That's because _getOptionText returns false

ioweb-gr:
How about this ? 

```php
        $text = $this->_getOptionText($filter);
        if ($filter && !empty($text)) {
            $this->_getResource()->applyFilterToCollection($this, $filter);
            $this->getLayer()->getState()->addFilter($this->_createItem($text, $filter));
            $this->_items = [];
        }
```

fballiano:
right, can you check this new version?

it seems to me that $text really needs to be a string for:
`$this->getLayer()->getState()->addFilter($this->_createItem($text, $filter));`
to work

ioweb-gr:
That also works

fballiano:
if you confirm me that it works I'll merge it right away and release a 20.6

ioweb-gr:
Yeah I already checked it, it didn't throw the error.

fballiano:
let's consider this as a review and merge it

PR #3927, opened 2024-04-06
Backend: removed invalid "name" element, removed redundant nobr spans, added data-column-id to grids

fballiano:
This PR rewrites https://github.com/OpenMage/magento-lts/pull/3923

In the process of reviewing https://github.com/OpenMage/magento-lts/pull/3923 I've seen that the `name` attribute is not valid for `span` or `a`, but the concept of the PR was interesting, so I did a clean up of the whole thing:
- removed invalid name attribute
- added data-column-id to the `th` instead
- removed the `nobr` span (there are way too many DOM nodes in OM, they need to be simplified) and added a `white-space:nowrap` to its container instead

Closes https://github.com/OpenMage/magento-lts/pull/3923

### DIFF
--- a/app/design/adminhtml/default/default/template/widget/grid.phtml
+++ b/app/design/adminhtml/default/default/template/widget/grid.phtml
@@ -114,7 +114,7 @@ $numColumns = count($this->getColumns());
                 <?php if ($this->getHeadersVisibility()): ?>
                     <tr class="headings">
                     <?php foreach ($this->getColumns() as $_column): ?>
-                        <th<?php echo $_column->getHeaderHtmlProperty() ?>><span class="nobr"><?php echo $_column->getHeaderHtml() ?></span></th>
+                        <th<?php echo $_column->getHeaderHtmlProperty() ?> data-column-id="<?php echo $_column->getId() ?>"><?php echo $_column->getHeaderHtml() ?></th>
                     <?php endforeach ?>
                     </tr>
                 <?php endif ?>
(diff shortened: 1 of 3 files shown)


fballiano:
@loekvangool please check this one instead

loekvangool:
I'm getting aroused! Great stuff

fballiano:
since I've tested this, I'll consider https://github.com/OpenMage/magento-lts/pull/3927#issuecomment-2041146575 as a review and merge it tomorrow if nothing against comes up

loekvangool:
Okay since you insist I accept my promotion to committer

app/design/adminhtml/default/default/template/widget/grid.phtml
```diff
@@ -114,7 +114,7 @@ $numColumns = count($this->getColumns());
                 <?php if ($this->getHeadersVisibility()): ?>
                     <tr class="headings">
                     <?php foreach ($this->getColumns() as $_column): ?>
-                        <th<?php echo $_column->getHeaderHtmlProperty() ?>><span class="nobr"><?php echo $_column->getHeaderHtml() ?></span></th>
```
kiatng, line 117: Not sure about removing the span element. It may break custom styling that select for it, for example:

```css
.grid tr.headings th span.nobr { ... }
```

fballiano: It is true (I thought about it as I'm very strict with other changes like modifying the signature of a public method) but as the same time the possibility of this happening and the minor side effect made me go for it.

fballiano: we could merge this in v21 but... it seems overkill for this.

fballiano: also, I've just removed the .grid tr.headings th span.nobr CSS line which was a hack for firefox 3... firefox 3...

kiatng (approved):

PR #3936, opened 2024-04-09
Converted all phtml files to short echo tag

fballiano:
This PR converts `<?php echo` to `<?=` (short echo tag) for all `phtml` files.  
It also removes the final `;` before the closing tag `?>`.

This PR:
- makes the overall size of the project smaller (removing 6 or 7 characters for every echo statement)
- makes the phtml files more readable (IMHO)

**Note**: short echo tag is not part of short tags (that could be disabled) and it's always available as stated here:
https://www.php.net/manual/en/language.basic-syntax.phptags.php


ADDISON74:
LGTM

ADDISON74:
I created a phtml template file that I loaded it in a CMS page. Here is the PHTML file content (/template/cms/test_page.phtml)

```php
<?php
$_modalTitle  = "Modal Title";
$_modalMessage = "Test";
?>

<?php if ($_modalMessage == "Test") $_modalBody = 'This is a test message' ?>

<!-- Button trigger modal -->
<button type="button" class="btn btn-primary" data-bs-toggle="modal" data-bs-target="#exampleModal">
  Launch demo modal
</button>

<!-- Modal -->
<div class="modal fade" id="exampleModal" tabindex="-1" aria-labelledby="exampleModalLabel" aria-hidden="true">
  <div class="modal-dialog">
    <div class="modal-content">
      <div class="modal-header">
        <h1 class="modal-title fs-5" id="exampleModalLabel"><?php echo $_modalTitle ?></h1>
        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
      </div>
      <div class="modal-body">
        <?php echo $_modalBody ?>
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-secondary" data-bs-dismiss="modal">Close</button>
      </div>
    </div>
  </div>
</div>
```

This is the the block loaded in the CMS page

```html
{{block type="core/template" template="cms/test_page.phtml"}}
```

If I load the page I see the button, then if I click it the modal window appears. 

Let's change the very first <?php tag. In the page content Modal Title text appears before the button which is not correct.

Let's go further and change the <?php tag in the if statement. The whole layout is not loaded. 

I would be cautious with this change because it has unexpected effects.

fballiano:
> Let's change the very first <?php tag to <?=. In the page content Modal Title text appears before the button which is not correct.

If I understood correctly what you mean that's not possible, `<?=` means `echo` so it can't be used when the php code requires logic or anything else than `echo`

ADDISON74:
I deleted by mistake the post, sorry for that, I tried to edit it.

If the change only concerns ```<?php echo to <?=``` it is not worth it, because we will have a mix with <?php tags where it cannot be replaced with <?=. For example in the case of a ```<?php if` ...``. My layout didn't load after making this change.

fballiano:
you can't use "<?= if" since it's not a valid php syntax

Flyingmana:
only for curiosity, is there a benefit in making it shorter?
Is it improving the parsing speed, or the opcache size?
Iam at least sure it does have no effect on the zipped size.

fballiano:
I tend to think that opcache is gonna be the same because it's already bytecode.

yes the file size difference is minimal but that's kinda the same reason why we removed mysql4 function and rewrite the headers of every file. I think :-)

mainly for me it's readability, less characters to read for your eyes and more characters you can fit on a single line.

also, when you see <?= you clearly understand that it's gonna be an echo and you don't really have to use to brain to discern if it's containing some logic or not, kinda reminds of single-responsibility-principle from afar

ADDISON74:
I don't know what to say, but I tend to disagree with this PR. I prefer using <?php as before because it is uniform and adopted for a long time. Tags are easy to find in a file. Adopting this change would mean having a mix of notations for the PHP code that creates confusion. I wouldn't want to create a special case for <?php echo mainly. In addition, it doesn't help anything. Deleting a few characters in hundreds of files does not bring any optimization benefit to the current processing power.

ADDISON74:
If there will be more people who want this implementation, then I would propose that ```<?php echo .... ?>``` being replaced by ```<? ...?>```. Just remove the = character and the effect is the same. Visually speaking it is an improvement, but I repeat, it is still beneficial for the programmer to find <?php echo in the file not <?= or <?.

Flyingmana:
> being replaced by <? ...?>. Just remove the = character and the effect is the same.

thats not true, this would be a very different construct which is not enabled on most instances and might even be deprecated at one point in php itself.

ADDISON74:
This PR is about a visual change. After 10 years of changing templates for me it does not bring any advantage. Variant <? ... ?> that I proposed is an alternative for the visual change, I tested and it is working. Anyway, I don't promote it and I agree that it could create issues. That's why my opinion is to leave things as they were for years. I prefer to see the echo function there instead of variable or string.

This PR will generate different opinions. I expect those having opinions for merging to make an effort to convince the others.

fballiano:
we understand you don't care for this, it's ok. as I said I think templates should be written as this PR shows, if nobody cares for this I'll close it.

> If there will be more people who want this implementation, then I would propose that `<?php echo .... ?>` being replaced by `<? ...?>`. Just remove the = character and the effect is the same. Visually speaking it is an improvement, but I repeat, it is still beneficial for the programmer to find <?php echo in the file not <?= or <?.

sorry but this sentence is completely wrong, that would be wrong and not working at all.

ADDISON74:
Please don't get me wrong, I appreciate any PR and the work everyone puts in for OpenMage, but I wanted to express my own opinion related to this one.

empiricompany:
IMHO I agree with all PRs that improve DX and in this case it would modernize the framework even if it has no benefit in terms of performance. 
I always hated the long `<?php echo` syntax but I continued to use it to comply with the standard

fballiano:
@empiricompany @joshua-bn if you like the PR please leave a review using this link https://github.com/OpenMage/magento-lts/pull/3936/files thank you!

empiricompany (approved):

empiricompany (commented):
> @empiricompany @joshua-bn if you like the PR please leave a review using this link https://github.com/OpenMage/magento-lts/pull/3936/files thank you! 

generally tested

ADDISON74:
I realized with sadness, it's true a bit late, that if I have different opinions, someone has to come and give a slap across my face. As long as merging is done with one approval, I don't think it matters what are my opinions from now on. This community project had some standards in which unfortunately I no longer find myself. Please accept my apologies for being honest. Before posting I used the unsubscribe button after 5 years of OpenMage. I wish you all success with this project.

fballiano:
Nobody merged yet. I told you I'm waiting for more feedback.
That doesn't mean if you don't like something then you should overrule other people's opinion.

fballiano:
Like nobody merged https://github.com/OpenMage/magento-lts/pull/3739 only because you were against it!

ADDISON74:
Discussions related to short echo tags date back more than 10 - 12 years if I'm not mistaken. I don't want to discuss this topic because it was discussed enough at that time, opinions being divided, but the majority was that they should be avoided. I mentioned earlier <? ... ?>, I was expecting someone to say that it was removed starting with PHP 8 based on a vote in 2019. The same can happen with <?= in the future. It is still there because it was supported by Rasmus Lerdorf. If it is a good or bad practice to use <?= in OpenMage template I'll let you find the answer yourself. Please note the "echo syntax" is used in all the themes out there and thousands of code examples someone may search for.

fballiano:
"short tags" have nothing to do with "short echo" tho. that's why "short echo" is always available also when "short tags" are disabled, as I pointed out in the description of the pr.

joshua-bn:
> @empiricompany @joshua-bn if you like the PR please leave a review using this link https://github.com/OpenMage/magento-lts/pull/3936/files thank you!

I've been following the discussion but haven't checked the PR.

PR #3945, opened 2024-04-19
Removed unused files from app/code/core/Mage/Adminhtml/Block/Catalog/Category/Tab

fballiano:
**This PR targets next**

These 2 files are never used, probably many years ago they were the general/design tabs of the category management page, but they're abandoned since I don't know how long.

I found this because I was debugging something and I couldn't wrap my head around the fact that my code wasn't running, wasted some time because of this leftover files.

They should be removed.


Sekiphp (approved):
I can confirm, that these 2 files are not used in category edit.

PR #3947, opened 2024-04-21
Rewrote `js/mage/adminhtml/accordion.js` without prototypejs

fballiano:
**This PR targets `next`**

To test this (I've tested and found no problem):
- open backend, catalog -> products, then open the tab "product alerts", the content of the page are 2 accordions
- open backend, customers -> manage customers, recent orders, shopping cart and wishlist are accordions (using ajax too)


andrewcbi (approved):
have tested and works as expected. thank you

PR #3948, opened 2024-04-21
Rewrote `js/mage/adminhtml/input-counter.js` without prototypejs

fballiano:
**This PR targets `next`**

This was very hard to test, I couldn't find anywhere where this file would originally be used (it's included in the html but it doesn't seem to run anywhere).

So
- I've created a product, with a "text" custom option, setting a maxlength for the custom option value
- I've created an order, trying to add the product from above, when the "configure" popup is shown then I'd copy past the inner part of the code from this PR (without the first line `window.addEventListener('DOMContentLoaded', function() {`)

The code is working as expected, this is the countdown when the input is still valid
Screenshot 2024-04-21 alle 17 07 21

and this is when it becomes invalid
Screenshot 2024-04-21 alle 17 07 26

### DIFF
--- a/js/mage/adminhtml/input-counter.js
+++ b/js/mage/adminhtml/input-counter.js
@@ -31,5 +31,7 @@ window.addEventListener('DOMContentLoaded', function() {
     document.querySelectorAll('.validate-length').forEach((elm) => {
         prepareForCountdown(elm);
         elm.addEventListener('keyup', countdown);
+        elm.addEventListener('paste', countdown);
+        elm.addEventListener('propertychange', countdown);
     });
 });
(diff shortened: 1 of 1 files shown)


js/mage/adminhtml/input-counter.js
```diff
@@ -1,40 +1,35 @@
...
+        }
+        return elm; // so you can chain
+    }
 
-    // deferred listener, only responds to keyups that issue from a matching element
-    document.on('keyup', '.validate-length', function(evt, elm) {
-        elm.countdown();
+    document.querySelectorAll('.validate-length').forEach((elm) => {
+        prepareForCountdown(elm);
+        elm.addEventListener('keyup', countdown);
```
andrewcbi, line 33: Would it be worth having this here too, I ended up needing these for a similar application to ensure paste operations ended up firing the counter also:
```
elm.addEventListener('paste', countdown);
elm.addEventListener('propertychange', countdown);
```

fballiano: done, thanks!

empiricompany (approved):

PR #3949, opened 2024-04-21
Removed unused file `js/varien/telephone.js`

fballiano:
**This PR targets `next`**

I couldn't find any reference to this file, so it could be safely removed (it's also heavily prototypejs based)


luigifab (approved):
Not sure the purpose of it, but for phone numbesr there is https://github.com/jackocnr/intl-tel-input

PR #3954, opened 2024-04-23
Added support for Laravel Ignition

fballiano:
Laravel Ignition is a pretty famous module for a better handling of error pages: https://flareapp.io/docs/ignition/introducing-ignition/overview

I thought it could be interesting for developers to have built-in, but absolutely optional (it's a pretty big package) in openmage.

**The logic is:**
- if there's a PHP error or exception
- and is in DeveloperMode
- and the spatie-ignition composer module is installed
- then the error is passed to ignition for the display

If DeveloperMode is disabled or the spatie-ignition package is not installed, OM behaves as usual.
It shouldn't introduce any performance degradation.

This is the error page for a PHP error (function doesn't exist):
Screenshot 2024-04-23 alle 13 25 13

This is the error page for a default PHP exception:
Screenshot 2024-04-23 alle 13 25 44

And this is a Mage::throwException:
Screenshot 2024-04-23 alle 13 25 58

**How to test**
1. `composer require "spatie/ignition"`
2. enable developer mode
3. modify one of your files in order to generate an error

PR #3957, opened 2024-04-24
Add new events for better integration with spatie-ignition

empiricompany:
### Description (*)

Add 3 new events around print exception and error handler to better integrate spatie-ignition via observer.
I have created an external module:
`composer require empiricompany/openmage_ignition`

### Related Pull Requests

https://github.com/OpenMage/magento-lts/discussions/3955
https://github.com/OpenMage/magento-lts/pull/3954

1. checkout the PR or patch `openmage/magento-lts`

```
        "patches": {
            "openmage/magento-lts": {
                "Add events for integration with spatie-ignition #3957": "https://github.com/OpenMage/magento-lts/pull/3957.patch"
            }
        }
```

2. install package:
`composer require empiricompany/openmage_ignition`

3. enable developer mode to print errors and exceptions


app/Mage.php
```diff
@@ -1002,6 +996,7 @@ public static function printException(Throwable $e, $extra = '')
 
             require_once(self::getBaseDir() . DS . 'errors' . DS . 'report.php');
         }
+        self::dispatchEvent('mage_print_exception_after', ['exception' => $e, 'extra' => $extra]);
```
fballiano, line 999: is this needed? since probably the mage_print_exception_before will end with a die...

and the normal error report.php also dies, there's probably no way anybody could reach the mage_print_exception_after

app/code/core/Mage/Core/Model/App.php
```diff
@@ -384,15 +384,10 @@ public function run($params)
...
-        if (Mage::getIsDeveloperMode() && class_exists('\Spatie\Ignition\Ignition')) {
-            \Spatie\Ignition\Ignition::make()
-                ->applicationPath(Mage::getBaseDir())
-                ->register();
-        } else {
-            $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
-        }
-
+        Mage::dispatchEvent('core_app_init_before', ['app' => $this]);
+        $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
```
fballiano, line 388: are you sure you don't want to invert these two? since you set another error handled in the module?

fballiano:
This has potential, I'd wait for some more development of the module itself and to define a couple of details in the code of this PR, then it may totally replace my implementation

app/code/core/Mage/Core/Model/App.php
empiricompany: good point
we can also add 2 events here _before and _after, then observe after event
and maybe better to name it more specific `core_app_initenv_*` cause we have more _init methods (_initCache(),  _initBaseConfig() etc..)
what do you think?

app/Mage.php
empiricompany: I was thinking about possible additional error tracking integrations after print / save reports
to me report.php call Error_Processor that render the template only after reach die() in printException(), correct?

app/code/core/Mage/Core/Model/App.php
fballiano: interesting topic, i'm not sure at the moment, from one side I wouldn't add many events (especially for a method that's called every time) from the other side they would be helpful in some cases, I'll think about it

app/Mage.php
fballiano: my bad, you're right, the die is AFTER this code so my point is invalid

app/code/core/Mage/Core/Model/App.php
```diff
@@ -384,15 +384,10 @@ public function run($params)
...
         date_default_timezone_set(Mage_Core_Model_Locale::DEFAULT_TIMEZONE);
+
```
fballiano, line 390:
```suggestion
Mage::dispatchEvent('core_app_init_environment_after', ['app' => $this]);

```

app/code/core/Mage/Core/Model/App.php
```diff
@@ -384,15 +384,10 @@ public function run($params)
...
     {
-        if (Mage::getIsDeveloperMode() && class_exists('\Spatie\Ignition\Ignition')) {
-            \Spatie\Ignition\Ignition::make()
-                ->applicationPath(Mage::getBaseDir())
-                ->register();
-        } else {
-            $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
-        }
-
+        Mage::dispatchEvent('core_app_init_before', ['app' => $this]);
```
fballiano, line 387:
```suggestion
```

app/code/core/Mage/Core/Model/App.php
fballiano: at the moment I'd only add this event to this class, it allows you to override what the method does and it should be enough at the moment

fballiano:
I'd also backport this to main

fballiano:
https://github.com/OpenMage/magento-lts/blob/main/docs/EVENTS.md also needs to be updated

app/code/core/Mage/Core/Model/App.php
```diff
@@ -384,15 +384,10 @@ public function run($params)
...
-            \Spatie\Ignition\Ignition::make()
-                ->applicationPath(Mage::getBaseDir())
-                ->register();
-        } else {
-            $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
-        }
-
+        $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
         date_default_timezone_set(Mage_Core_Model_Locale::DEFAULT_TIMEZONE);
+Mage::dispatchEvent('core_app_init_environment_after', ['app' => $this]);
```
fballiano, line 389:
```suggestion
        Mage::dispatchEvent('core_app_init_environment_after', ['app' => $this]);
```

sorry, I mistyped the spaces before

docs/EVENTS.md
```diff
@@ -236,6 +237,8 @@
 | log_log_clean_before | 1.9.4.5 |
 | log_visitor_collection_load_before | 1.9.4.5 |
 | mage_run_exception | 1.9.4.5 |
+| mage_print_exception_before | 21.0.0 |
+| mage_print_exception_after | 21.0.0 |
```
fballiano, line 241:
```suggestion
| mage_print_exception_before | 20.7.0 |
| mage_print_exception_after | 20.7.0 |
```

docs/EVENTS.md
```diff
@@ -193,6 +193,7 @@
 | controller_front_send_response_before | 1.9.4.5 |
 | controller_response_redirect | 1.9.4.5 |
 | core_app_run_after | 20.1.0 |
+| core_app_init_environment_after | 21.0.0 |
```
fballiano, line 196:
```suggestion
| core_app_init_environment_after | 20.7.0 |
```

app/Mage.php
```diff
@@ -964,14 +964,8 @@ public static function getIsDeveloperMode()
      */
     public static function printException(Throwable $e, $extra = '')
```
pquerner, line 965: This is also called in cron context.

fballiano: interesting, but it's up to the external module to handle that situation so it shouldn't be a problem here

pquerner: Should it? Yea maybe. Would be the first time tho. None other event to my knowledge just gets executed and the listener has to check if its from cli.

empiricompany: Good point. What do we have to do in this situation? Catch in ignition could be useful if we want to track in flare (not implemented yet) as well, but the output is not CLI-friendly.
I think it's a situation to handle in the external module.

fballiano: well in this case, since the error could be sent to Flare or another log collecting service... probably it's ok that the event triggers also in the cron environment

empiricompany: Regarding this PR, the event must be dispatched; then, the logic within the observer will must to handle it

fballiano (approved):
I think that also if we have a couple more events in the printexception it's no big deal, since it should never happen anyway.

the "init" one is necessary for sure in my opinion, much more clear than the controller_predispatch IMHO

so for me this is a yes

ps: i'll backport this to "main" too

fballiano:
@empiricompany now that mage_run_installed_exception is available, could you test your module using mage_run_installed_exception instead of mage_print_exception_before? 

for some reason your module is not working on my instance, I don't understand why but it's not calling the observers

empiricompany:
i've tested with events:
```xml
<mage_run_exception>
    <observers>
        <mm_ignition>
            <class>MM_Ignition_Model_Observer</class>
            <method>handleIgnitionException</method>
        </mm_ignition>
    </observers>
</mage_run_exception>
<mage_run_installed_exception>
    <observers>
        <mm_ignition>
            <class>MM_Ignition_Model_Observer</class>
            <method>handleIgnitionException</method>
        </mm_ignition>
    </observers>
</mage_run_installed_exception>
```
and work the same as with `mage_print_exception_before` for exception but...

I encountered another issue, I can't override the default PHP error handler by dispatching this event `core_app_init_environment_after` in Mage_Core_Model_App::_initEnvironment() because the module is not ready at this point.
so for example for a php error on undefined method like `Mage_Catalog_CategoryController::nofunc()` not works

instead with your hardcoded methods still works

```php
    /**
     * Initialize PHP environment
     *
     * @return $this
     */
    protected function _initEnvironment()
    {
        if (Mage::getIsDeveloperMode() && class_exists('\Spatie\Ignition\Ignition')) {
            \Spatie\Ignition\Ignition::make()
                ->applicationPath(Mage::getBaseDir())
                ->register();            
        } else {
            $this->setErrorHandler(self::DEFAULT_ERROR_HANDLER);
        }
```

This is a big problem that I currently don't know how to solve.

empiricompany:
These are the dispatched events:
```
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_before
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_after
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_before
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_after
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_before
2024-05-01T15:15:28+00:00 DEBUG (7): core_collection_abstract_load_after
2024-05-01T15:15:28+00:00 DEBUG (7): controller_front_init_before
2024-05-01T15:15:28+00:00 DEBUG (7): controller_front_init_routers
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): model_load_after
2024-05-01T15:15:28+00:00 DEBUG (7): core_abstract_load_after
2024-05-01T15:15:28+00:00 DEBUG (7): session_before_renew_cookie
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): core_locale_set_locale
2024-05-01T15:15:28+00:00 DEBUG (7): core_locale_set_locale
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): resource_get_tablename
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch
2024-05-01T15:15:28+00:00 DEBUG (7): customer_session_init
2024-05-01T15:15:28+00:00 DEBUG (7): customer_session_init
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch_catalog
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch_catalog
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch_catalog_category_view
2024-05-01T15:15:28+00:00 DEBUG (7): controller_action_predispatch_catalog_category_view
```

empiricompany:
I've done some tests, and it seems that we need an event at this point, immediately after initialization
init

fballiano:
I'd kinda lean towards using controller_front_init_before then

empiricompany:
I have changed the events used in my module:
https://github.com/empiricompany/openmage_ignition/blob/main/app/code/community/MM/Ignition/etc/config.xml

- controller_front_init_before
- mage_run_exception
- mage_run_installed_exception

So, there is no longer need for this PR.

fballiano:
does your module works 100% also my previous PR about Ignition stays in the core?

empiricompany:
> does your module works 100% also my previous PR about Ignition stays in the core?

From my test module works correcty (I invite you to test it), and since we now have an external module, there's no longer a need for the hard-coded loading of Ignition in the core.
So we can close this PR and revert this one: https://github.com/OpenMage/magento-lts/pull/3954.

PR #3958, opened 2024-04-25
Fixed: Fixer.io endpoint for currency updates

jankovandeventer:
I tested it, and it works, in our own project I used an overwrite to make sure we can keep collecting data.

### Description (*)
Recently there were some changes, this is the new endpoint if anyone is also using it 💸
The platform had a few changes over the years, but I think recently an endpoint was deprecated, resulting in no more updating our currency information.

The format is the same, just the url is changed and apiKey was renamed to access_key in the params.

### Related Pull Requests

I believe this solves: https://github.com/OpenMage/magento-lts/issues/3071

I believe this solves: https://github.com/OpenMage/magento-lts/issues/3071

1. set up a free account on fixer.io
2. add 2 or more currencies
3. go to System -> Manage Currencies -> Rates
4. Press import
5. Success

image

### Questions or comments
First PR for Open Mage, keep it up!

### DIFF
--- a/app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
+++ b/app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
@@ -36,7 +36,7 @@ class Mage_Directory_Model_Currency_Import_Fixerio extends Mage_Directory_Model_
      *
      * @var string
      */
-    protected $_url = 'https://api.apilayer.com/fixer/latest?apikey={{ACCESS_KEY}}&base={{CURRENCY_FROM}}&symbols={{CURRENCY_TO}}';
+    protected $_url = 'http://data.fixer.io/api/latest?access_key={{ACCESS_KEY}}&base={{CURRENCY_FROM}}&symbols={{CURRENCY_TO}}&format=1';
 
     /**
      * Information messages stack


app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
```diff
@@ -36,7 +36,7 @@ class Mage_Directory_Model_Currency_Import_Fixerio extends Mage_Directory_Model_
      *
      * @var string
      */
-    protected $_url = 'https://api.apilayer.com/fixer/latest?apikey={{ACCESS_KEY}}&base={{CURRENCY_FROM}}&symbols={{CURRENCY_TO}}';
+    protected $_url = 'http://data.fixer.io/api/latest?access_key={{ACCESS_KEY}}&base={{CURRENCY_FROM}}&symbols={{CURRENCY_TO}}&format=1';
```
pquerner, line 39: Why only http?

fballiano: https seems to work so it should be https

fballiano:
if I open https://api.apilayer.com/fixer/latest on the browser it seems to work, is it not working at all for you?

hirale:
https://apilayer.com/marketplace/fixer-api

image

pquerner:
Reads to me that apilayer is just a wrapper around the fixer api. So no change should be required.

https://apilayer.com/marketplace/fixer-api

fballiano:
@jankovandeventer why did you need it? did the flexer api not work for you?

hirale:
> @hirale what do you mean? that the currently used address is working fine?

the current one works fine.

app/code/core/Mage/Directory/Model/Currency/Import/Fixerio.php
jankovandeventer: HTTPS requires a paid account as far as I know, mostly legacy implementations just use 30 requests a month (one a day) and do not require a paid account, but we could make this configurable.

jankovandeventer:
Hi there, sorry for not responding sooner. When I use the credentials with the apilayer, I get an invalid authentication warning. 

image


image

But I realize my mistake, the fixer.io credentials do not work on the apilayer so I had to move the subscription over to their new platform and now it works.

image

Sorry to bother :-)

Issue #3952, opened 2024-04-22
In backend, add "click to copy text" feature


pauldpauld:
### Description (*)

I transfer new customer details into my accounting program, always using tedious "copy 'n paste", I have noticed that eBay and other websites have a "click to copy" button, or simply, click on a field which copied the text content.

![image](https://github.com/OpenMage/magento-lts/assets/1549372/9b2bda13-12df-47ac-af69-e3ac4e018cb2)

The above is my freight company page with a tracking number, clicking on the graphic copies the tracking number.

### Expected behavior (*)

Clicking on a customer name and/or address fields on the order page will copy that text to the clipboard.

### Benefits

It would make it much quicker and efficient for admin staff to copy many of the customer details (and or product details) if they don't already use a linked method.

### Additional information

Here is a sample from W3schools

```
` Step 1) Add HTML:
Example
The text field
<input type="text" value="Hello World" id="myInput">
<input type="text" value="Hello World" id="myInput">

The button used to copy the text
<button onclick="myFunction()">Copy text</button>
<button onclick="myFunction()">Copy text</button>

 <!-- The text field -->
<input type="text" value="Hello World" id="myInput">

<!-- The button used to copy the text -->
<button onclick="myFunction()">Copy text</button>
```
Step 2) Add JavaScript:
Example
function myFunction() {
  // Get the text field
  var copyText = document.getElementById("myInput");

  // Select the text field
  copyText.select();
  copyText.setSelectionRange(0, 99999); // For mobile devices

   // Copy the text inside the text field
  navigator.clipboard.writeText(copyText.value);

  // Alert the copied text
  alert("Copied the text: " + copyText.value);
} 
`

luigifab:
So good idea! See also #2774.
With Linux you can select any text to copy it :)

PR #3960, opened 2024-04-27
New feature: Added copyText function to admin order view page for copying order details

hirale:
Add copyText function to admin order view page for copying order details

### Description (*)

Include the implementation of the feature described in issue #3952, allowing users to easily copy order numbers, customer emails, and tracking numbers directly from the order page with a single click.

1. Fixes OpenMage/magento-lts#3952

1. fresh installation with sample data
2. add js function and modify templates
3. tested on chrome, works fine

image


pquerner:
I would like this to be in a seperate template, so it can get used more easily (without duplicating essentially) across the backend. What do you think?

Flyingmana:
can we do this over a special html attribute and css class combination, so the button is injected dynamically via js?

(just an idea, not a requirement from my side)

kiatng:
+1 on @Flyingmana 's idea. Instead of svg, may be use the [copy cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) on hover:
image

Flyingmana:
The Icon used is already good, what I see nowdays it already evolves to a standard recognizable copy icon.

hirale:
it's nice idea to add the buttons dynamically via js, but there are problems of styles for different elements.

fixed size 1em
image

without size
image

any ideas?

hirale:
To add a copy icon to an element, follow these steps:

1. Add the `copy-text` class to the target element. This class is used by JavaScript to identify the elements that need to have a copy icon appended.

2. Add the `data-text` attribute, which holds the text content to be copied.

3. Add the `data-copy-icon` attribute, which stores the styling information for the copy icon in JSON format, including height, width, and other style properties.

For example:

```html
<span class="copy-text" data-text="<?= $trackNumber ?>" data-copy-icon='{"style":"cursor:pointer;","height":"12px","width":"12px"}'><?= $trackNumber ?></span>
```

fballiano:
I really like the idea of this PR, what about having the SVG in a CSS and then just inject a div/span placeholder that gets styled by the css?

hirale:
I think it is more convenient to use the special attributes of html tag.

fballiano:
ok i mean, instead of injecting the whole svg you're injecting the placeholder that gets stylized by the css, the data-attribute stay the same

hirale:
@fballiano  do you mean this?

js/varien/js.js
``` javascript
/**
 * Adds copy icons to elements with the class 'copy-text'.
 */
function addCopyIcons() {
    const copyTexts = document.querySelectorAll('.copy-text');
    copyTexts.forEach(copyText => {
        const iconStyle = JSON.parse(copyText.getAttribute('data-copy-icon'));
        const svg = createSVGElement(iconStyle);
        copyText.parentNode.appendChild(svg);
    });
}


/**
 * Creates an SVG element with the specified iconStyle.
 *
 * @param {Object} iconStyles - An object containing the style properties for the SVG icon.
 * @param {string} [iconStyles.cursor='pointer'] - The cursor style for the SVG element.
 * @param {string} iconStyles.height - The height of the SVG element.
 * @param {string} iconStyles.width - The width of the SVG element.
 * @param {string} [iconStyles.margin='0'] - The margin of the SVG element.
 * @return {HTMLElement} The created SVG element.
 */
function createSVGElement(iconStyles) {
    const copyIcon = document.createElement('span');
    copyIcon.classList.add('icon-copy');
    copyIcon.setAttribute('onclick', 'copyText(event)');
    copyIcon.style.cursor = iconStyles.cursor || 'pointer';
    copyIcon.style.height = iconStyles.height;
    copyIcon.style.width = iconStyles.width;
        copyIcon.style.margin = iconStyles.margin || '0';

    return copyIcon;
}
/**
 * Copies the text from the data-text attribute of the clicked element to the clipboard.
 *
 * @param {Event} event - The event object triggered by the click event.
 * @return {Promise<void>} A promise that resolves when the text is successfully copied to the clipboard.
 */
function copyText(event) {
    let copyText = event.currentTarget.previousElementSibling.getAttribute('data-text');
    navigator.clipboard.writeText(copyText);
}
```
phtml example
``` phtml
<h4 class="icon-head head-account copy-text" data-text="<?= $_order->getRealOrderId() ?>" data-copy-icon='{"cursor":"pointer","margin":"8px 0 0 5px","height":"16px","width":"16px"}'><?php echo Mage::helper('sales')->__('Order # %s', $this->escapeHtml($_order->getRealOrderId())) ?> (<?php echo $_email ?>)</h4>
```

skin/adminhtml/default/openmage/scss/override.scss
``` css
.copy-icon {
  display: inline-block;
  background-image: url('images/copy-icon.svg');
  background-repeat: no-repeat;
  background-size: contain;
}
```

fballiano:
yep and the svg could also be inlined in the css (I kinda go for this route lately)

pquerner:
> yep and the svg could also be inlined in the css (I kinda go for this route lately)

So you love it when a base64 svg stares you in the face :D

fballiano:
svg sprites are hard(er than old-style sprites) and if we start having many svg that will be a problem, if the svg is inlined in the css that's more or less solved

js/varien/js.js
```diff
@@ -703,3 +703,47 @@ function buttonDisabler() {
...
+ * @return {HTMLElement} The created SVG element.
+ */
+function createSVGElement(iconStyles) {
+    const copyIcon = document.createElement('span');
+    copyIcon.classList.add('icon-copy');
+    copyIcon.setAttribute('onclick', 'copyText(event)');
+    copyIcon.style.cursor = iconStyles.cursor || 'pointer';
+    copyIcon.style.height = iconStyles.height;
+    copyIcon.style.width = iconStyles.width;
+        copyIcon.style.margin = iconStyles.margin || '0';
```
pquerner, line 736: Whitespace issue

js/varien/js.js
```diff
@@ -703,3 +703,47 @@ function buttonDisabler() {
...
+
+    return copyIcon;
+}
+/**
+ * Copies the text from the data-text attribute of the clicked element to the clipboard.
+ *
+ * @param {Event} event - The event object triggered by the click event.
+ * @return {Promise<void>} A promise that resolves when the text is successfully copied to the clipboard.
```
pquerner, line 744: Nothing is returned. (And also not needed I think)

js/varien/js.js
```diff
@@ -703,3 +703,47 @@ function buttonDisabler() {
...
+ */
+function copyText(event) {
+    let copyText = event.currentTarget.previousElementSibling.getAttribute('data-text');
```
pquerner, line 747: Can be `const`.

app/design/adminhtml/default/default/template/sales/order/view/tab/info.phtml
```diff
@@ -51,6 +51,13 @@
                 <?php  if ($_order->getTracksCollection()->count()) : ?>
                 <a href="#" id="linkId" onclick="popWin('<?php echo $this->helper('shipping')->getTrackingPopupUrlBySalesModel($_order) ?>','trackorder','width=800,height=600,resizable=yes,scrollbars=yes')" title="<?php echo $this->quoteEscape($this->__('Track Order')) ?>"><?php echo $this->__('Track Order') ?></a>
                 <br/>
+                    <?php foreach ($_order->getTracksCollection() as $_track): ?>
```
pquerner, line 54: I'd rather split this into its own PR, but I guess it shows a good use-case.
(Just wanted to note that)

fballiano:
when I click the copy icon on the order number I get:
Screenshot 2024-04-29 alle 22 46 47

hirale:
> when I click the copy icon on the order number I get: Screenshot 2024-04-29 alle 22 46 47

which browser?

hirale:
> chrome
maybe here is the reason. Should I add some checks before writeText function?
https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined
https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText
https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts

fballiano:
ahh maybe since in my local machine I don't have https, well I'd add a check as a first line in addCopyIcons

pquerner (approved):

app/design/adminhtml/default/default/template/sales/order/view/info.phtml
```diff
@@ -28,7 +28,7 @@ $orderStoreDate = $this->formatDate($_order->getCreatedAtStoreDate(), 'medium',
         endif; ?>
         <div class="entry-edit-head">
         <?php if ($this->getNoUseOrderLink()): ?>
-            <h4 class="icon-head head-account"><?php echo Mage::helper('sales')->__('Order # %s', $this->escapeHtml($_order->getRealOrderId())) ?> (<?php echo $_email ?>)</h4>
+            <h4 class="icon-head head-account copy-text" data-text="<?= $_order->getRealOrderId() ?>" data-copy-icon='{"cursor":"pointer","margin":"8px 0 0 5px","height":"16px","width":"16px"}'><?php echo Mage::helper('sales')->__('Order # %s', $this->escapeHtml($_order->getRealOrderId())) ?> (<?php echo $_email ?>)</h4>
```
fballiano, line 31: do we have to have all of this data in the data-copy-icon? can we have like a couple of classes or something easier?  if we have all the css in a json, at this point it's better to have the span already.

empiricompany: I don't understand the need to specify a CSS for each element in `data-copy-icon`

empiricompany: why not include all in css like this?
icon-copy

demo


keep the same icon size is better for me, but we can alway stylish in css like:

```
.entry-edit-head .icon-copy {
width:16px
height:16px;
}
```

hirale: My idea is to allow users to add the copy icon functionality to any element in the adminhtml without adding an element near the target element.

fballiano: I understand that but if you've to have all the css in a json data-attribute, then render the actual element via js... to me it's a no.

fballiano: minimum the width/height and the cursor should have defaults

hirale: ok, so create span via js or add placeholder in template? which is better?

fballiano: if the css in data-attribute was optional then I'd be ok with the current implementation. width/height to 16 by default (or 12, what makes more sense), and the cursor too. didn't check about the margins. then override the margins if necessary.

my opinion is partially positive because it's backend, if it was frontend this approach could cause content layout shift and should be avoided.

-----------------

actually
- if data-text was renamed to data-copy-text then would probably be no need for the class='copy-text' in order to identify the elements that need the copy icon next to it
- also instead of width/height in the data-copy-icon I'd just have a size, they have to be square, so
`<h4 class="icon-head head-account copy-text" data-text="<?= $_order->getRealOrderId() ?>" data-copy-icon='{"cursor":"pointer","margin":"8px 0 0 5px","height":"16px","width":"16px"}'>` could be `<h4 class="icon-head head-account" data-copy-text="<?= $_order->getRealOrderId() ?>" data-copy-icon='{"margin":"8px 0 0 5px","size":"16px"}'>` and only if the size if overriding the default size

empiricompany: i vote for remove all `data-copy-icon` and stylish it via css

fballiano: true, since you've to modify the phtml anyway... it not a super big benefit to have the data-attribute instead of just a <span

empiricompany:
I can't easly fork your master...
But i would propose to provide a visual feedback after copying. 
We can add a new class and then override the SVG icon like this

```css
class="icon-copy icon-copy-copied"
```
demo



```javascript
function copyText(event) {
    const copyText = event.currentTarget.previousElementSibling.getAttribute('data-text');
    navigator.clipboard.writeText(copyText);

    const copyTexts = document.querySelectorAll('.icon-copy');
    copyTexts.forEach(iconCopy => {
        iconCopy.classList.remove('icon-copy-copied');
    });
    
    event.currentTarget.classList.add('icon-copy-copied');
}
```

```css
.icon-copy-copied {
  background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" x="0px" y="0px" width="100" height="100" viewBox="0 0 30 30"><path d="M 26.980469 5.9902344 A 1.0001 1.0001 0 0 0 26.292969 6.2929688 L 11 21.585938 L 4.7070312 15.292969 A 1.0001 1.0001 0 1 0 3.2929688 16.707031 L 10.292969 23.707031 A 1.0001 1.0001 0 0 0 11.707031 23.707031 L 27.707031 7.7070312 A 1.0001 1.0001 0 0 0 26.980469 5.9902344 z"></path></svg>');}
```

fballiano:
nice idea! and then after a couple of seconds it could revert back to the original class

app/design/adminhtml/default/default/template/sales/order/view/info.phtml
```diff
@@ -114,7 +114,7 @@ $orderStoreDate = $this->formatDate($_order->getCreatedAtStoreDate(), 'medium',
                 </tr>
                 <tr>
                     <td class="label"><label><?php echo Mage::helper('sales')->__('Email') ?></label></td>
-                    <td class="value"><a href="mailto:<?php echo $this->escapeHtml($_order->getCustomerEmail()) ?>"><strong><?php echo $this->escapeHtml($_order->getCustomerEmail()) ?></strong></a></td>
+                    <td class="value"><a href="mailto:<?php echo $this->escapeHtml($_order->getCustomerEmail()) ?>" class="copy-text" data-text="<?= $_order->getCustomerEmail()?>" data-copy-icon='{"cursor":"pointer","margin":"2px 0 0 2px","height":"10px","width":"10px"}'><strong><?php echo $this->escapeHtml($_order->getCustomerEmail()) ?></strong></a></td>
```
empiricompany, line 117: all short syntax `<?=` must be `<?php echo `
There was a proposal, but it didn't succeed https://github.com/OpenMage/magento-lts/pull/3936

empiricompany:
> nice idea! and then after a couple of seconds it could revert back to the original class



```javascript
function copyText(event) {
    const copyText = event.currentTarget.previousElementSibling.getAttribute('data-text');
    navigator.clipboard.writeText(copyText);

    const copyIcon = event.currentTarget;
    copyIcon.classList.add('icon-copy-copied');
    setTimeout(() => {
        copyIcon.classList.remove('icon-copy-copied');
    }, 1000);
}
```

empiricompany:
> nice idea! and then after a couple of seconds it could revert back to the original class

i've update my code in the original comment to disapper after 1s

js/varien/js.js
```diff
@@ -703,3 +703,50 @@ function buttonDisabler() {
...
+ */
+function createSVGElement(iconStyles) {
+    const copyIcon = document.createElement('span');
+    copyIcon.classList.add('icon-copy');
+    copyIcon.setAttribute('onclick', 'copyText(event)');
+    copyIcon.style.cursor = iconStyles.cursor || 'pointer';
+    copyIcon.style.height = iconStyles.height;
+    copyIcon.style.width = iconStyles.width;
+    copyIcon.style.margin = iconStyles.margin || '0';
+
```
empiricompany, line 741: add some hints
```suggestion
    copyIcon.setAttribute('title', 'Copy text');
```

js/varien/js.js
```diff
@@ -703,3 +703,50 @@ function buttonDisabler() {
...
+ * Creates an SVG element with the specified iconStyle.
+ *
+ * @param {Object} iconStyles - An object containing the style properties for the SVG icon.
+ * @param {string} [iconStyles.cursor='pointer'] - The cursor style for the SVG element.
+ * @param {string} iconStyles.height - The height of the SVG element.
+ * @param {string} iconStyles.width - The width of the SVG element.
+ * @param {string} [iconStyles.margin='0'] - The margin of the SVG element.
+ * @return {HTMLElement} The created SVG element.
+ */
+function createSVGElement(iconStyles) {
```
empiricompany, line 733: better to rename this function to a more specific task name
```suggestion
function createIconCopyElement(iconStyles) {
```

fballiano:
In some resolutions it looks like this:

Screenshot 2024-04-30 alle 13 11 49

js/varien/js.js
fballiano: done, with a slightly different name :-)

js/varien/js.js
fballiano: done, using translator.translate

fballiano:
@empiricompany I've committed your code to switch the icon

app/design/adminhtml/default/default/template/sales/order/view/info.phtml
fballiano: fixed it

hirale: It's fine to use old echo format.

hirale:
I've added the `data-copy-text` attribute to the target element and styled it using a CSS class, as we discussed earlier.

fballiano (dismissed):
latest version is fine to me :-)

fballiano (approved):

fballiano:
@empiricompany I'd like to get your opinion on this last version :-)

empiricompany (approved):

skin/adminhtml/default/openmage/scss/override.scss
```diff
@@ -1160,3 +1160,22 @@ img[src$='grid-cal.gif'] {
...
 }
+
+.icon-copy {
+  display: inline-block;
+  background-image: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 111.07 122.88"><path d="M97.67 20.81h0l.01.02a13.38 13.38 0 0 1 9.46 3.93 13.32 13.32 0 0 1 3.9 9.42h.02v.02 75.28.01h-.02c-.01 3.68-1.51 7.03-3.93 9.46a13.32 13.32 0 0 1-9.42 3.9v.02h-.02-59.19-.01v-.02c-3.69-.01-7.04-1.5-9.46-3.93a13.37 13.37 0 0 1-3.91-9.42h0V34.2v-.01h.02c.01-3.69 1.52-7.04 3.94-9.46 2.41-2.4 5.73-3.9 9.42-3.91v-.02h.02l59.17.01h0zM.02 75.38L0 13.39v-.01h.02a13.44 13.44 0 0 1 3.93-9.46A13.37 13.37 0 0 1 13.37.01V0h.02 59.19c7.69 0 8.9 9.96.01 10.16H13.4h-.02v-.02c-.88 0-1.68.37-2.27.97-.59.58-.96 1.4-.96 2.27h.02v.01 3.17 58.81c0 8.26-10.15 8.72-10.15.01h0zm100.89 34.11V34.2v-.02h.02c0-.87-.37-1.68-.97-2.27-.59-.58-1.4-.96-2.28-.96v.02h-.01-59.19-.02v-.02c-.88 0-1.68.38-2.27.97-.59.58-.96 1.4-.96 2.27h.02v.01 75.28.02h-.02c0 .88.38 1.68.97 2.27s1.4.96 2.27.96v-.02h.01 59.19.02v.02c.87 0 1.68-.38 2.27-.97.59-.58.96-1.4.96-2.27h-.01 0 0 0z" fill-rule="evenodd"/></svg>');
+  background-repeat: no-repeat;
+  background-size: contain;
+  cursor: pointer;
+  width: 12px;
+  height: 12px;
```
empiricompany, line 1171: scss differ from compiled css, miss margins

skin/adminhtml/default/openmage/scss/override.scss
```diff
@@ -1160,3 +1160,22 @@ img[src$='grid-cal.gif'] {
...
+}
```
empiricompany, line 1172:
```suggestion
  margin: 2px 0 0 2px;
```

skin/adminhtml/default/openmage/scss/override.scss
```diff
@@ -1160,3 +1160,22 @@ img[src$='grid-cal.gif'] {
...
+
+.entry-edit-head .icon-copy {
+  width: 16px;
+  height: 16px;
+}
```
empiricompany, line 1177:
```suggestion
  margin: 8px 0 0 5px;
```

skin/adminhtml/default/openmage/scss/override.scss
fballiano: @empiricompany thanks so much! couldn't commit your suggestions directly cause they were missing the '}' but I'll make sure to have you as co-author in the final merge commit

PR #3755, opened 2024-01-25
Rewrote `js/mage/centinel.js` without prototypejs

fballiano:
- I've no idea how to test this file
- in the checkout page this file gets included and no errors are logged in the developer console

PR #3971, opened 2024-05-02
Removed obsolete module Mage_Centinel

fballiano:
**This PR targets `next`**

Mage_Centinel is a module that provided 3DS for payment methods in OpenMage. We're talking about 3DS 1.0, not 3DS 2.0 which is more or less mandatory for all credit card providers all over the world nowadays.

I've conducted quite a few tests and I'm pretty sure the module is not working anymore, when you enable it you will have this error message:
Screenshot 2024-05-02 alle 23 24 06

The XmlParser class is a reserved name in php8:
Screenshot 2024-05-02 alle 23 24 17

Also if you remove that class from the filesystem you will see:
Screenshot 2024-05-02 alle 23 24 32

So:
- 3DS 1.0 is obsolete and nobody should use it anymore, continuing to use the outdated Centinel module would expose merchants to potential security risks and compatibility issues with modern payment systems and regulations.
- the name "Centinel" doesn't really exist since 2011, when CardinalCommerce was acquired by Visa and the Centinel service became part of Visa's product offerings under the name "Visa Secure"
- the Centinel module is actually broken and doesn't technically work anymore

I think it's safe to remove this code from OpenMage v21.

Closes https://github.com/OpenMage/magento-lts/pull/3755


app/code/core/Mage/Payment/Model/Method/Cc.php
```diff
@@ -283,117 +278,4 @@ public function isAvailable($quote = null)
...
-            ->setOrderNumber($this->_getOrderId());
-        return $params;
-    }
-
-    /**
-     * Order increment ID getter (either real from order or a reserved from quote)
-     *
-     * @return string
-     */
-    private function _getOrderId()
```
kiatng, line 339: This and following functions don't seem to have anything to do with Centinel. Is it ok to delete?

kiatng:
Also need to cleanup here
image

app/code/core/Mage/Payment/Model/Method/Cc.php
fballiano: true, I didn't remove them at first but phpstan started complaining, you can see it here:
https://github.com/OpenMage/magento-lts/actions/runs/8931953986/job/24534948827

because those methods are private and not used so they should be un-usable by anybody

I think

fballiano:
@kiatng done! thanks for noticing!

luigifab:
Buy buy Centinel! Disabled for me.

empiricompany (approved):

kiatng (commented):
See https://docs.adyen.com/online-payments/3d-secure-for-regulation-compliance/

3-D Secure 1 and related technology are no longer supported by all major cards.

kiatng (approved):

PR #3976, opened 2024-05-07
Fixed UPS Rest API bugs

fballiano:
Let's use this PR to fix all the bugs we can find regarding the new UPS Rest API.

### DIFF
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -1205,7 +1205,7 @@ XMLAuth;
         if ($jsonResponse) {
             $responseData = json_decode($jsonResponse, true);
 
-            if (isset($responseData['trackResponse']['shipment'][0]['package'][0]['activity'])) {
+            if (isset($responseData['trackResponse']['shipment'])) {
                 $activityTags = $responseData['trackResponse']['shipment'][0]['package'][0]['activity'] ?? [];
                 $index = 1;
                 foreach ($activityTags as $activityTag) {
@@ -1257,9 +1257,7 @@ XMLAuth;
                     $index++;
                 }
                 $resultArr['progressdetail'] = $packageProgress;
-            } elseif (isset($responseData['trackResponse']['shipment'][0]['warnings'][0]['message'])) {
-                $errorTitle = $responseData['trackResponse']['shipment'][0]['warnings'][0]['message'];
-            } else {
+            } elseif (isset($responseData['response']['errors'][0]['message'])) {
                 $errorTitle = $responseData['response']['errors'][0]['message'];
             }
         }
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -754,9 +754,9 @@ XMLRequest;
 XMLRequest;
 
         $xmlRequest .= (
-        $params['49_residential'] === '01'
-            ? "<ResidentialAddressIndicator>{$params['49_residential']}</ResidentialAddressIndicator>"
-            : ''
+            $params['49_residential'] === '01'
+                ? "<ResidentialAddressIndicator>{$params['49_residential']}</ResidentialAddressIndicator>"
+                : ''
         );
 
         $xmlRequest .= <<< XMLRequest
@@ -2047,12 +2047,9 @@ XMLAuth;
         $countryShipper     = $params->getCountryShipper();
         $countryRecipient   = $params->getCountryRecipient();
 
-        if (($countryShipper == self::USA_COUNTRY_ID
-                && $countryRecipient == self::CANADA_COUNTRY_ID)
-            || ($countryShipper == self::CANADA_COUNTRY_ID
-                && $countryRecipient == self::USA_COUNTRY_ID)
-            || ($countryShipper == self::MEXICO_COUNTRY_ID
-                && $countryRecipient == self::USA_COUNTRY_ID)
+        if (($countryShipper == self::USA_COUNTRY_ID && $countryRecipient == self::CANADA_COUNTRY_ID)
+            || ($countryShipper == self::CANADA_COUNTRY_ID && $countryRecipient == self::USA_COUNTRY_ID)
+            || ($countryShipper == self::MEXICO_COUNTRY_ID && $countryRecipient == self::USA_COUNTRY_ID)
             && $method == '11' // UPS Standard
         ) {
             $containerTypes = [];
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -1222,34 +1222,28 @@ XMLAuth;
                         }
                         $dateArr = [];
                         $date = (string)$activityTag['date'];
-                        //YYYYMMDD
                         $dateArr[] = substr($date, 0, 4);
                         $dateArr[] = substr($date, 4, 2);
                         $dateArr[] = substr($date, -2, 2);
 
                         $timeArr = [];
                         $time = (string)$activityTag['time'];
-                        //HHMMSS
                         $timeArr[] = substr($time, 0, 2);
                         $timeArr[] = substr($time, 2, 2);
                         $timeArr[] = substr($time, -2, 2);
 
                         if ($index === 1) {
                             $resultArr['status'] = (string)$activityTag['status']['description'];
-                            $resultArr['deliverydate'] = implode('-', $dateArr);
-                            //YYYY-MM-DD
-                            $resultArr['deliverytime'] = implode(':', $timeArr);
-                            //HH:MM:SS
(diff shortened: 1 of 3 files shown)


app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1201,7 +1201,7 @@ protected function _parseRestTrackingResponse($trackingValue, $jsonResponse)
         if ($jsonResponse) {
             $responseData = json_decode($jsonResponse, true);
 
-            if ($responseData['trackResponse']['shipment']) {
+            if (isset($responseData['trackResponse']['shipment'])) {
```
fballiano, line 1204: This change solves https://github.com/OpenMage/magento-lts/pull/3878#issuecomment-2098924102 but the array structure still needs to be verified

ragnese: > but the array structure still needs to be verified

I think what you have here is enough. If this is a success (200) response from the API, the key(s) for response.trackResponse.shipment are, I think, always present. If they aren't there, that implies we got a non-happy response from UPS.

fballiano: for the error it needed to be $responseData['response'], for some reason, that's why I put this note ;-)

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1201,62 +1205,62 @@ protected function _parseRestTrackingResponse($trackingValue, $jsonResponse)
         if ($jsonResponse) {
             $responseData = json_decode($jsonResponse, true);
 
-            if ($responseData['trackResponse']['shipment']) {
+            if (isset($responseData['trackResponse']['shipment'][0]['package'][0]['activity'])) {
```
ragnese, line 1208: I'm not sure I agree with this change. If there isn't tracking info, that's not necessarily an error, right? If we give a tracking number that UPS doesn't have tracking for anymore (they only track for 120 days), do we want to display the error message: "For some reason we can\'t retrieve tracking info right now."? That's not quite right, IMO. 

I think the way you had it before was probably better. It's not wrong to return an empty array from this function if the request was successful, but didn't have any "activity".

UPS will return "{"trackResponse":{"shipment":[{"warnings":[{"code":"TW0001","message":"Tracking Information Not Found"}]}]}}" for such a case.

fballiano: exactly, that's why I did it, becuase in the elseif I'm catching the "warning", but let me check better

fballiano: i don't get it, if there's no activity the whole "if" won't do anything

ragnese: It's actually okay for the whole "if" to do nothing. If you follow the call chain, it eventually goes up to `Mage_Sales_Model_Order_Shipment_Track::getNumberDetail`, line 147, where it handles an empty tracking result just fine.

fballiano: yes but it will generated warning for non-existent  array keys I think

fballiano: since the whole if doesn't do anything if there's no activity, I think it's better to protect it with a full isset like now, I could remove the first elseif but... why? if there's a specific warning message we will return it

ragnese: > yes but it will generated warning for non-existent array keys I think

It won't. The `??` operator prevents PHP from complaining about it.

> if there's a specific warning message we will return it

The only reason I'd hesitate here is because I don't know every warning that UPS could send back. What if it sends something that is very "programmer-y" that only the site administrators should be seeing? End users won't want to see errors about missing request parameters or something. 

Maybe just logging warnings would make more sense?

fballiano: I see the point of weird messages from UPS but "tracking not found" I think it's pretty important.

Btw to me it's not finding any tracking :-\

ragnese: > I see the point of weird messages from UPS but "tracking not found" I think it's pretty important.

That one specifically *isn't* super important because if this function sets an empty trackingResult, then the calling classes handle it by displaying: `Mage::helper('sales')->__('No detail for number "%s"', $this->getNumber());`. So that case is handled "automatically" for us.


> Btw to me it's not finding any tracking :-\

Do you have a specific tracking number I can test? Is the order recent?

fballiano: I've reverted it, but I don't like it that much.

About tracking, sorry, it works, it's just weird that you can click "track order" also if there's no tracking number, but that's another something

ragnese:
Where is the code that displays the message for the adminhtml form item here?
Screenshot 2024-05-07 at 3 09 40 PM

fballiano:
it's in app/code/core/Mage/Usa/etc/system.xml

fballiano:
@ragnese what about we include this one https://github.com/magento/magento2/pull/38673/files?

ragnese:
> @ragnese what about we include this one https://github.com/magento/magento2/pull/38673/files?

Interesting. So the UPS API doesn't work if you mix metric and imperial units? I guess we probably should include that, then.

fballiano:
it doesn't work (I had the same problem since I selected "LBS" because the system was mixing KGS and INCHES and UPS didn't like that

ragnese:
> it doesn't work (I had the same problem since I selected "LBS" because the system was mixing KGS and INCHES and UPS didn't like that

Fair enough. Then we should definitely pull it in. (Though, it's pretty funny to think of a 5 cm x 5 cm x 5 cm package :p )

fballiano:
@ragnese done

ragnese:
Do you think this is worth pushing out as another point release as it is now? 

On the one hand, it would be nice to wait until we're pretty sure everything works well and only have one more release to fix the UPS REST issues. But, on the other hand, it doesn't seem like a good idea to let the tracking stuff stay broken longer than necessary.

Personally, I think the broken tracking is serious enough that this should be merged sooner rather than later and then just tackle future bugs/issues as they come.

fballiano:
I would wait a few more days, there are people testing this patch so there's no super hurry to do another release immediately, every communication has warnings about this. let's give it a few more days and if everything seems stable we'll go

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -1201,62 +1205,60 @@ protected function _parseRestTrackingResponse($trackingValue, $jsonResponse)
         if ($jsonResponse) {
             $responseData = json_decode($jsonResponse, true);
 
-            if ($responseData['trackResponse']['shipment']) {
+            if (isset($responseData['trackResponse']['shipment'])) {
                 $activityTags = $responseData['trackResponse']['shipment'][0]['package'][0]['activity'] ?? [];
-                if ($activityTags) {
```
ragnese, line 1206: Taking away this `if` changes the logic, and I'm not sure if it's correct.

If there is no activity, the old code will leave `$resultArr` as an empty array. This new version will set `$resultArr['progressdetail'] = $packageProgress`, no matter what. 

Then when `$resultArr` is passed to `setTrackingResultData`, the `if ($resultArr)` would evaluate to false in the old code, but true in the new code. This means that we will be appending a (empty) tracking-result-status to the tracking result in the new code, where the old code would have an empty tracking result.

So, I think this is incorrect and should be left as it was.

ragnese (changes requested):
We just need to address the comment about the "activity" if-statement.

ragnese (approved):

fballiano:
I'll now merge this since we had no ulterior feedback (my customer say this 20.7 + this PR work fine) and we'll have to release it publicly very soon.

fballiano:
Today I deployed this patch to production for a customer, tests:
- post deploy, without configuration change -> everything ok (still using UPS XML APIs)
- post deploy, after switching to Rest API -> shipping API works ok, tracking doesn't work, I get `[result] => {"response":{"errors":[{"code":"250002","message":"Invalid Authentication Information."}]}}`

fballiano:
my mistake, the production Rest API account wasn't setup correctly and didn't have "tracking" access enabled, enabling it fixed everything.

I think I can confirm everything works correctly.

I'll post an update in the next few days if there's something worth alerting you all.

mingjyou:
seems like free UPS shipping method do now work.

fballiano:
@mingjyou please provide more info so that we can test this

mingjyou:
Free Method: UPS Ground ( seems like do not allow to choose free method )

Minimum Order Amount for Free Shipping: 200

do not show free ups ground shipping when check out    On Wednesday, May 22, 2024 at 12:21:14 PM EDT, Fabrizio Balliano ***@***.***> wrote:  
 
 


@mingjyou please provide more info so that we can test this

PR #3982, opened 2024-05-08
RWD theme: Removed ancient normalize.css

fballiano:
**This PR targets `next`**

RWD includes an ancient version (2.0.1) of modernize.css, nowadays I've tried updating it to 8.0.1 (latest version, which is from 2018... so already very old) or removing it altogether and I don't see any difference.

I think it's safe to remove it and make the CSS a bit lighter.

Tested in chrome-firefox-safari.


pquerner (approved):

kiatng (approved):
Ref https://github.com/necolas/normalize.css

Issue #3441, opened 2023-08-11
Missing payment method CCsave crashes seeing older orders or invoices 


vincentRobot:
### Preconditions (*)
1. OpenMage 19.4.18 

### Steps to reproduce (*)

1. Try an view an older order or invoice that was paid by CCsave method from magento 1.9 it will crash the site

### Expected result (*)

1. we should be able to see the order instead the site crashes because that payment method CCsave is missing .

### Actual result (*)

1. site crashes :
![image](https://github.com/OpenMage/magento-lts/assets/53285000/2542bc16-713d-4520-aeb0-3ce370ce73bf)
![image](https://github.com/OpenMage/magento-lts/assets/53285000/a64d0448-6b2b-4fe6-8a10-2cbc6c4c655b)
exception log : 
[1481985719546.log](https://github.com/OpenMage/magento-lts/files/12322540/1481985719546.log)

It is possible to keep CCsave as a payment method (even if it's not PCI compliant) with OpenMage ? if not how I can 'mock a payment method" that will help me see older orders & invoices in the admin site ?

thanks

elidrissidev:
I don't know about PCI compliance, but the error you're facing is normal. As far as I see, the `Ccsave` method is still available in OpenMage, if you remove it while you have older orders using it, it's going to throw that error due to the payment method's class missing.

vincentRobot:
@elidrissidev : Thanks I agree the error looks normal under the circumstances. 
This is extremely weird since I didn't remove the payment method. OpenMage was installed via composer over an existing magento 1.9 site in order to keep my customizations. 
https://github.com/OpenMage/magento-lts#composer.
I was under the impression composer would overwrite the core files only but perhaps this created the problem ? 
Should I reinstall ? or perhaps you can tell me which files are responsible for the ccsave payment method and I will add them manually.

Thanks for your help

elidrissidev:
From my experience, even if there were issues during the process of copying files it shouldn't cause the files to get deleted since they existed since M1.

You should make sure the below file exists, as well as the blocks referenced by it like `payment/info_ccsave`:

https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Payment/Model/Method/Ccsave.php

elidrissidev:
Is your issue solved, @vincentRobot?

vincentRobot:
I think so @elidrissidev. 
I ended up getting help through the discord group.
[...]

PR #3984, opened 2024-05-08
Removed obsolete CCSave payment method

fballiano:
**This PR targets `next`**

CCSave is an old payment method that saved the credit card data directly in the database, this is not used (and nobody should use it) anymore since years.

I've checked my customers' projects and I didn't see any module extending these classes. Although it's possible that there's something out there, it should be part of some functionality that's not in use anymore.

**Note: this PR will break the possibility of viewing old orders payed with the ccsave method, as shown in https://github.com/OpenMage/magento-lts/issues/3441, my opinion is that having merchants still using ccsave in any way could be considered dangerous but probably they don't know (but I think they should and would prefer to know) so at least if the functionaly breaks on their website they will be able to understand that they were having a problematic situation, fix it and most importantly sanitize their database (you don't want to receive a security audit for them to discover you have credit cards stored in your database).**


fballiano:
PHPStan error will be solved by merging https://github.com/OpenMage/magento-lts/pull/3976

luigifab:
Perhaps it can be interesting to add a dummy payment method, and a SQL to clean and update db?

boesbo (approved):

boesbo (approved):

SuSiSonnenschein111:
Magento removed some lines to prevent this hack ... https://blog.sucuri.net/2019/07/magento-killer.html

Woulldn't it be easier to set

`protected $_canSaveCc   = false;`

in `Mage_Payment_Model_Method_Ccsave` to prevent that CC-info is stored (and re-add tha XML-code, to not break anything)?

This would leave some unused files, but should work like normal CC payment.

PR #3985, opened 2024-05-09
Fixed cannot sort by column in backend's grids

fballiano:
https://github.com/OpenMage/magento-lts/pull/3927 broke grid sorting functionality in the backend because the js was looking for the "name" attribute.

This PR rewrites the doSort methods fixing the bug

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Renderer/Abstract.php
@@ -124,7 +124,7 @@ abstract class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Abstract extends
             if ($this->getColumn()->getDir()) {
                 $className = 'sort-arrow-' . $dir;
             }
-            $out = '<a href="#" title="' . $nDir . '" class="' . $className . '"><span class="sort-title">'
+            $out = '<a href="#" name="' . $this->getColumn()->getId() . '" title="' . $nDir . '" class="' . $className . '"><span class="sort-title">'
                    . $this->escapeHtml($this->getColumn()->getHeader()) . '</span></a>';
         } else {
             $out = $this->escapeHtml($this->getColumn()->getHeader());
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -1232,20 +1232,16 @@ XMLAuth;
 
                         if ($index === 1) {
                             $resultArr['status'] = (string)$activityTag['status']['description'];
-                            $resultArr['deliverydate'] = implode('-', $dateArr);
-                            //YYYY-MM-DD
-                            $resultArr['deliverytime'] = implode(':', $timeArr);
-                            //HH:MM:SS
+                            $resultArr['deliverydate'] = implode('-', $dateArr); //YYYY-MM-DD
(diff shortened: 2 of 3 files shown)


empiricompany:
Sorry, but the patch solves the sorting issues in the default grid, 
but the module for advanced grids, widely used, [mage-enhanced-admin-grids](https://github.com/empiricompany/mage-enhanced-admin-grids) continues to not work.

The problem is here: https://github.com/OpenMage/magento-lts/commit/91b1944b2cad76349ce0e8899d6821eaaabd3ca7
where is being removed the attribute name

I have also try to change attribute name to data-name and update it here
https://github.com/empiricompany/mage-enhanced-admin-grids/blob/7aa9a82bc86d1ea87c07dd00db7251ff1ea96162/js/bl/customgrid/config.js#L226

but not resolve the issues, the sorting and filters are not working.

fballiano:
did you try element.parentNode.dataSet.columnId?

empiricompany:
> did you try element.parentNode.dataSet.columnId?

Tried now, but it doesn't work.

fballiano:
I've re-added the "name" attribute to the "a", which btw is also reported by phpstorm

Screenshot 2024-05-09 alle 11 35 42

empiricompany:
True, the tag ID should be used instead. 
For now, thank you for restoring "name". 

We should find a solution for these old, widely used extensions. For example, I was thinking of creating a dedicated clone repository of OpenMage and keep it updated then store all patches in the PRs then keep synchronized with the main one.
But i'm not sure i have so much time...

empiricompany (approved):

empiricompany:
Sorry, for it to work, we also need to revert " js/mage/adminhtml/grid.js" to version 20.6.0

https://raw.githubusercontent.com/OpenMage/magento-lts/v20.6.0/js/mage/adminhtml/grid.js

```diff
---
 js/mage/adminhtml/grid.js | 10 +++++-----
 1 file changed, 5 insertions(+), 5 deletions(-)

diff --git a/js/mage/adminhtml/grid.js b/js/mage/adminhtml/grid.js
index 55eb3b2ef05..9e2a168e5d4 100644
--- a/js/mage/adminhtml/grid.js
+++ b/js/mage/adminhtml/grid.js
@@ -141,15 +141,15 @@ varienGrid.prototype = {
 
     },
     doSort : function(event){
-        var element = Event.findElement(event, 'a');
+        event.preventDefault();
+        const element = event.target.closest('a');
+        const parentElement = element.parentNode;
 
-        if(element.name && element.title){
-            this.addVarToUrl(this.sortVar, element.name);
+        if (element && parentElement && parentElement.dataset.columnId && element.title) {
+            this.addVarToUrl(this.sortVar, parentElement.dataset.columnId);
             this.addVarToUrl(this.dirVar, element.title);
             this.reload(this.url);
         }
-        Event.stop(event);
-        return false;
     },
     loadByElement : function(element){
         if(element && element.name){
```

fballiano:
done

empiricompany:
For those who want to fix it immediately 20.0.7 without downgrade, you can apply the patch like this:

```json
{
    [...]
    "require": {
        "aydin-hassan/magento-core-composer-installer": "~2.1.0",
        "openmage/magento-lts": "^20.7.0",
    [...]
    "extra": {
        "enable-patching": true,
        [...]
        "patches": {
            "openmage/magento-lts": {
                "Fixed cannot sort by column in backend's grids #3985": "https://patch-diff.githubusercontent.com/raw/OpenMage/magento-lts/pull/3985.patch"
            }
        }
    }
```

fballiano:
I don't understand how the js thing can happen but whatever :-\

PR #3988, opened 2024-05-13
Updated PHPStan to 1.11.2

fballiano:


fballiano:
@Flyingmana @Sdfendor @kiatng could you take a look at this?

boesbo (approved):

fballiano:
I've retested this (mostly where the parent::construct() was removed, becuase the parent contructor is empty), just to be safe, can't see any problem.

I'll merge it.

F1Red5:
@fballiano we have a modified abstract class Mage_Catalog_Model_Product_Type_Abstract with some custom code in save() method. With the removal of parent::construct() it does not work anymore. Can you please revert it?

fballiano:
@F1Red5 could you override also the child class? I've removed it because phpstan complained that it's useless

Issue #4007, opened 2024-05-23
After upgrade to v20.7.0 I got some alert on Admin Login


starnetwork:
### Preconditions (*)

1. 20.7.0

### Steps to reproduce (*)

1. Login to Magento LTS Admin, click on Clicking here inside the alert box and getting Orphaned Role Resources empty page without any Roles.

### Expected result (*)

1. that magento admin will not show this alert.

### Actual result (*)

1. Attaching Screenshot of the error on admin login
2. Attaching screenshot of the error on "Clicking here"

![Screenshot 2024-05-23 132619](https://github.com/OpenMage/magento-lts/assets/626945/827134cf-a64b-43da-bf51-965c0323638b)
![Screenshot 2024-05-23 132647](https://github.com/OpenMage/magento-lts/assets/626945/f99e1632-3844-44e6-b05a-0ddca937490f)

fballiano:
you see the alert but then when you click there's the "no records"? really weird

@kiatng what do you think?

btw this is the original PR https://github.com/OpenMage/magento-lts/pull/3647

hirale:
confirmed.

fballiano:
@kiatng true, it happens only right after login, if you refresh the alert disappears

starnetwork:
@fballiano 
true, refresh and the alert disappears but it's back every time any admin do login again and again

fballiano:
I know, but it was important to tell that to the original author of the PR for debug

kiatng:
I did a test on the latest next branch:

adminhtml.xml, add a Test resource:
```xml
<config>
    <acl>
        <resources>
            <admin>
                <children>
                    <celera_crib translate="label" module="crib">
                        <title>Celera Modules</title>
                        <children>
                            <crib translate="title">
                                <title>Crib Services</title>
                                <sort_order>1000</sort_order>
                            </crib>
                            <test translate="title">
                                <title>Test</title>
                                <sort_order>1100</sort_order>
                            </test>
                        </children>
                    </celera_crib>
```

Then in backend > System > Roles > select a role and toggle the checkbox of the resource  > Save

![image](https://github.com/OpenMage/magento-lts/assets/1106470/be3c1002-8d00-464e-82d8-79ea3966d028)

Logout from backend.

Changed resource now to Test2 in adminhtml.xml 
```xml
                            <test2 translate="title">
                                <title>Test2</title>
                                <sort_order>1100</sort_order>
                            </test2>
```

[...]

PR #4022, opened 2024-05-31
Fixed issue #4007 bug on incorrect resources added to table `admin_ru…

kiatng:
…le`.

### Description (*)
This PR fixed the following:
1. Show the notice on  orphaned resources based on admin user ACL `system/acl/orphaned_resources`
2. Fixed bug on adding incorrect resources to table `admin_rule`

### Related Pull Requests
PR #3647 

1. Fixes OpenMage/magento-lts#4007

First, replicate the conditions in issue #4007 by adding `<action>some/action</action>` node in any `adminhtml.xml` file.  and refresh the cache.

1. Check the notice is not displayed when _Orphaned Role Resources_ is unchecked in admin > System > Permissions > Roles > click to edit a role
image
2. Logout and login as user with the role from step 1. No notice should be shown.
3. Logout and login as full admin user. Notice on orphaned resources is shown.
4. Click on the link and delete the orphaned resources.
5. Check that no invalid entries are added to table `admin_rule` by navigating to admin > System > Permissions > Roles > click to edit a role > Save (there is no need to edit the role). Before this PR, the `<action>` would be added. With this PR, they are not added.

### DIFF
--- a/app/code/core/Mage/Admin/Model/Resource/Acl.php
+++ b/app/code/core/Mage/Admin/Model/Resource/Acl.php
@@ -23,8 +23,6 @@ class Mage_Admin_Model_Resource_Acl extends Mage_Core_Model_Resource_Db_Abstract
 {
     public const ACL_ALL_RULES = 'all';
 
-    protected $_orphanedResources = [];
-
     /**
      * Initialize resource
      *
@@ -114,6 +112,7 @@ class Mage_Admin_Model_Resource_Acl extends Mage_Core_Model_Resource_Db_Abstract
      */
     public function loadRules(Mage_Admin_Model_Acl $acl, array $rulesArr)
     {
+        $orphanedResources = [];
         foreach ($rulesArr as $rule) {
             $role = $rule['role_type'] . $rule['role_id'];
             $resource = $rule['resource_id'];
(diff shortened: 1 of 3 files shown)


kiatng:
Here's the source of the bug:

image

This is a recursive function iterating the XML tree. On node `<action>` , `$resource->getName()` is string "action", `if (!in_array($resource->getName(), ['title', 'sort_order', 'children', 'disabled'])) {` is `true`, it is treated as a legitimate resource even though it is not.

Since all resources have children, as they need at least a `title`, so checking for non-empty children can validate real resources.

m-overlund:
The notification is still only shown once.
but now I'm able to see and delete the orphaned roles.

fballiano (approved):

fballiano:
it works for me, I applied the PR, logged in, I see the notice, click and the grid has values that I can remove. logout-login, no alert is shown anymore ;-)

fballiano:
@kiatng I've no way of testing if, merging this PR, the original situation would still work (I mean before the first round of removing orphaned ACLs) are you confident it still works in any case?

kiatng:
This PR is an improvement over the last one. But it's always better if more users can test it. There are quite a few people affected, should we wait for a couple more confirmations?

m-overlund (approved):

boesbo (approved):

fballiano:
I'll merge it as it is now, we've waited a bit, we've 1 green + 2 grays + the PR is by a maintainer so I think it's good to go

Issue #4023, opened 2024-05-31
Error when saving product with image without description


MateusM3:
### Preconditions (*)

1. OpenMage 20.8.0
2. PHP 8.2

### Steps to reproduce (*)

1. Open the Catalog -> Create Product -> Upload Images -> with out description (label) -> Save

### Expected result (*)

1. Saving the product with an empty label like previous versions, may be the version of PHP that is more careful regarding the array key not existing

### Actual result (*)

![error-img-default](https://github.com/OpenMage/magento-lts/assets/20692202/7a8f4323-2563-480b-9146-aedcc9e60fbd)

fballiano:
@MateusM3 can you test https://github.com/OpenMage/magento-lts/pull/4024?

MateusM3:
Thaks @fballiano ! But Its was partially corrected, still with the error on line 170, I corrected it as follows, taking your correction as an example
![image](https://github.com/OpenMage/magento-lts/assets/20692202/ed4a1727-730a-4dfd-ab7c-9c6ccb63104a)

Can you update your Pull request with this correction too?

kiatng:
I cannot reproduce the error on ver21 beta2.

fballiano:
@kiatng on v21 I see the same error, did you enable developer mode there too?

fballiano:
@MateusM3 how do you trigger the error you see on line 170? I can't reproduce it with my patch #4024

MateusM3:
@fballiano I tested it here and it worked, thank you!
Would it be possible to release a version 20.9.1 with this fix? I wanted to update here about having to overwrite the class

sreichel:
@all-contributors add @MateusM3 bug

PR #4024, opened 2024-05-31
Fixed undefined array index label_use_default in Mage_Catalog_Model_Product_Attribute_Backend_Media

fballiano:
Fixes https://github.com/OpenMage/magento-lts/issues/4023

There could be other ways to fix this problem but I would initialize those 2 array elements

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Attribute/Backend/Media.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Attribute/Backend/Media.php
@@ -242,11 +242,16 @@ class Mage_Catalog_Model_Product_Attribute_Backend_Media extends Mage_Eav_Model_
                 $image['value_id']      = $this->_getResource()->insertGallery($data);
             }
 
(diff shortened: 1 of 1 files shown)


boesbo (approved):

kiatng:
Why not use `empty()` to check:

```php
// from
$data['label']    = ($image['label'] === null || $image["label_use_default"]) ? null : $image['label'];
// to
$data['label']    = ($image['label'] === null || empty($image["label_use_default"])) ? null : $image['label'];
```

kiatng:
@fballiano You are right. I got confused with `if ($storeId === 0) $image["label_use_default"] = false;`. We want store 0 to have the default label, but in order to make sure of that, `$image["label_use_default"]` needs to be set to false, regardless of its prior value. It's counter intuitive. 

>  when the conditions are in || but not all positive

It's hard as hell for me too.

kiatng (approved):

PR #4019, opened 2024-05-28
Added stopPropagation to "copy text" buttons in backend

Caprico85:
### Description (*)
This PR adds `event.stopPropagation()` and `event.preventDefault()` to the new copyText() function.

The copy function is a very nice feature, which we already added to several other locations. Now our editors wanted to be able to copy values from backend grids, like this:

image

Adding the copy button can be done using a renderer. But the copy button is clicked, the event propagates up the DOM and triggers the row click, which navigates to the details page.

This is not what we wanted. I think, clicking the copy button should just do the copying and not do anything else. Therefore, calling `event.stopPropagation()` and `event.preventDefault()` to prevent all other actions should have no negative side effects.

### Manual testing scenarios (*)

1. A dirty way to add a copy button to a grid would be to edit Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Datetime::render() and add the line \
   `$data = '<span data-copy-text="' . $data . '">' . $data . '</span>';`
  right before the `return $data;`, like this \
   \
   image
   This adds a copy button to the date column in the orders grid
   image

2. Clicking the copy button copies the date. But it also navigates to the order details. With this PR applied, it only copies the date.

### Contribution checklist (*)
 - [X] Pull request has a meaningful description of its purpose
 - [X] All commits are accompanied by meaningful commit messages
 - [X] All automated tests passed successfully (all builds are green)
 - [X] Add yourself to contributors list

PR #4029, opened 2024-06-04
New Feature for developers: Copyable cells in grids

kiatng:
### Description (*)
Add `copyable` property to grid column. When the property is set to `true`, render the clickable copy icon in the cell:

image

### Related Pull Requests
See the main PR #3960 that renders the copy icon and the js that copies the target text. 
See the enhanced PR #4019 that copies text in the grid without navigating to the edit/view page.

On any grid class, set 'copyable'  => true,`; for example

```php
class Company_Module_Block_Adminhtml_Sales_Order_Grid extends Mage_Adminhtml_Block_Sales_Order_Grid
{
    protected function _prepareColumns()
    {
        // ...
        parent::_prepareColumns();
        $this->getColumn('real_order_id')->setCopyable(true); // Using setter.

        $this->addColumnAfter('ext_order_id', [
            'header'    => $this->__('STARS Order#'),
            'index'     => 'ext_order_id',
            'copyable'  => true, // Normal use.
        ], 'store_id');
       // ...
```

### Comments
In order for the legacy theme to work, I copy the related CSS from the openmage theme to the default/boxes.css.


skin/adminhtml/default/default/boxes.css
```diff
@@ -220,6 +220,21 @@ tr.dynamic-grid input.input-text { width:154px; }
 .available { color:#080; font-weight:bold; }
 .not-available { color:#800; }
 
+/* Grid - Copyable Cells */
+.icon-copy {
```
fballiano, line 224: can't we re-use the icon-copy class that's already in override.css?

kiatng: Not sure how to use the override.css in the openmage theme in the legacy.

fballiano: if it's not in the legacy theme maybe we should migrate it away from override.css, put it in the legacy so that it's available for both and we don't have to duplicate the classes

kiatng: You are right, we should.

kiatng:
Question: should the js be moved from `js\varien\js.js` to `js\mage\adminhtml\tools.js` since it's only available in backend.

fballiano:
@kiatng probably yes, I don't see that function be helpful on the frontend :-)

luigifab:
Beautiful!

kiatng:
Modern browsers require https to access the clipboard. For insecure http, you need to use Chrome. In the address bar, enter "chrome://flags/#Insecure-origins-treated-as-secure". Then input the URLs separated with a space in the box of _Insecure origins treated as secure_

fballiano:
tested, working perfectly

fballiano (approved):

fballiano:
since we didn't get any more feedback I'll merge

PR #4037, opened 2024-06-09
Preliminary PHP 8.4 support: Fixed implicitly nullable parameter deprecation warning

F1Red5:
See https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated


fballiano:
isn't modifying public method a breaking change?

Flyingmana:
I did not look through all of them yet, but if it only makes existing parameters nullable, which already use Null as default value, thats not likely to break anything.

fballiano:
but if they were overridden by some 3rd party code, the method signature changes and the user would get a warning/error no?

Flyingmana:
seems to be fine
https://3v4l.org/IKDae

fballiano:
ok I've also done some tests and the core of this PR is ok, but it has too much stuff, it adds rector and polifys that are not related to the PR itself, so I will remove those parts and only keep the "inplicitly nullable" part

PR #4039, opened 2024-06-10
Fixed `Mage_Uploader_Helper_File::getDataMaxSize()` when checking against different size units

F1Red5:
### Description (*)

Fixes `Mage_Uploader_Helper_File::getDataMaxSize()` when different units are used.

1. `min('1M', '1G')`returns `1G`

### DIFF
--- a/dev/tests/unit/Mage/Uploader/Helper/FileTest.php
+++ b/dev/tests/unit/Mage/Uploader/Helper/FileTest.php
@@ -113,10 +113,6 @@ class FileTest extends TestCase
                 1024,
                 '1024'
             ],
-            'byte' => [
-                1,
-                '1B'
-            ],
             'kilobyte' => [
                 1024,
                 '1K'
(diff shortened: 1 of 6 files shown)


dev/tests/unit/Mage/Uploader/Helper/FileTest.php
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    public function testGetDataMaxSize(): void
+    {
+        $mock = $this->getMockBuilder(Mage_Uploader_Helper_File::class)
+            ->setMethods(['getPostMaxSize', 'getUploadMaxSize'])
+            ->getMock();
+
+        $mock->expects($this->once())->method('getPostMaxSize')->willReturn('1G');
+        $mock->expects($this->once())->method('getUploadMaxSize')->willReturn('1M');
```
pquerner, line 86: What does this actually test here?

F1Red5: It is to simulate php.ini values for `post_max_size` and `upload_max_filesize` when post size is set to 1GB and upload size to 1MB.

pquerner: Thanks!

fballiano (approved):

fballiano:
when porting to `next` I'll not include the polyfills since they're not needed (`next` is running on 8.2 minimum)

PR #4040, opened 2024-06-15
Fixed cache issue in primary navigation block

empiricompany:
### Description (*)

On fresh block cache if i reload a PDP page the category of the product still marked as active in other pages if there is not a specific current category (homepage, contacts, cms page etc..)

https://github.com/OpenMage/magento-lts/issues/4041

1. Flush block cache
2. Load a PDP
3. Go to homepage

The category of first PDP loaded still flagged as active.
PS: The RWD theme does not have a CSS rule to highlight the active menu, so you must inspect the HTML.
active

This PR also adds css rule to highlight currente active category.

### DIFF
--- a/app/code/core/Mage/Catalog/Block/Navigation.php
+++ b/app/code/core/Mage/Catalog/Block/Navigation.php
@@ -74,6 +74,7 @@ class Mage_Catalog_Block_Navigation extends Mage_Core_Block_Template
             Mage::getDesign()->getPackageName(),
             Mage::getDesign()->getTheme('template'),
             Mage::getSingleton('customer/session')->getCustomerGroupId(),
+            Mage::getBlockSingleton('page/html_header')->getIsHomePage(),
             'template' => $this->getTemplate(),
             'name' => $this->getNameInLayout(),
             $this->getCurrenCategoryKey()
--- a/app/code/core/Mage/Catalog/Block/Navigation.php
+++ b/app/code/core/Mage/Catalog/Block/Navigation.php
@@ -74,7 +74,6 @@ class Mage_Catalog_Block_Navigation extends Mage_Core_Block_Template
             Mage::getDesign()->getPackageName(),
             Mage::getDesign()->getTheme('template'),
             Mage::getSingleton('customer/session')->getCustomerGroupId(),
-            Mage::getBlockSingleton('page/html_header')->getIsHomePage(),
             'template' => $this->getTemplate(),
             'name' => $this->getNameInLayout(),
             $this->getCurrenCategoryKey()
--- a/app/code/core/Mage/Catalog/Helper/Product.php
+++ b/app/code/core/Mage/Catalog/Helper/Product.php
@@ -348,6 +348,7 @@ class Mage_Catalog_Helper_Product extends Mage_Core_Helper_Url
             $category = Mage::getModel('catalog/category')->load($categoryId);
             $product->setCategory($category);
             Mage::register('current_category', $category);
+            Mage::register('current_entity_key', $category->getPath());
         }
 
         // Register current data and dispatch final events
--- a/skin/frontend/rwd/default/css/styles.css
+++ b/skin/frontend/rwd/default/css/styles.css
@@ -3554,7 +3554,8 @@ a.skip-link {
     border-bottom: none;
   }
   .nav-primary a:hover,
-  .nav-primary li:hover > a {
+  .nav-primary li:hover,
+  .nav-primary li.active > a {
     color: #0472ad;
   }
   .nav-primary .menu-active {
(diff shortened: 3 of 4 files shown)


fballiano:
I think you fixed a bug affecting 90% of magento1 (and 2) stores, where many disable the HTML_Block cache as a workaround.

will test asap!

F1Red5:
> a bug affecting 90% of magento1 (and 2) stores

... rly?

I thinks its (only) when people want different navigation for non-category CMS pages. Imho not a bug ...

Maybe (untested) if $`this->getCurrenCategoryKey()` is false/null better add full action path to cache info. (?)

`getIsHomePage()` is (as it says) for homepage only and a "half-way-fix".

Is this a "common" issue or something that could/should be fixed in an extension?

Imho.

empiricompany:
I thought I had solved it, but today I find a category selected again on the homepage, so it doesn't work correctly. 
I can't understand how the bug occurs.

@F1Red5 i thing the bug occur when root category id is 0

empiricompany:
Sorry for the confusion, I've finally figure out how this occur.
I've opened a separate issue and update the description on this PR.

empiricompany:
block Mage_Catalog_Block_Navigation is only used in catalog list

```xml
 <block type="core/text_list" name="top.menu" as="topMenu" translate="label">
    <label>Navigation Bar</label>
    <block type="page/html_topmenu" name="catalog.topnav" template="page/html/topmenu.phtml">
        <block type="page/html_topmenu_renderer" name="catalog.topnav.renderer" template="page/html/topmenu/renderer.phtml"/>
    </block>
</block>

```
i found Mage_Page_Block_Html_Topmenu read current_entity_key  from registry 

```php
/**
 * Retrieve current entity key
 *
 * @return int|string
 */
public function getCurrentEntityKey()
{
    if ($this->_currentEntityKey === null) {
        $this->_currentEntityKey = Mage::registry('current_entity_key')
            ? Mage::registry('current_entity_key') : Mage::app()->getStore()->getRootCategoryId();
    }
    return $this->_currentEntityKey;
}
```

so i've setted in PDP controller current_entity_key with the full category path like in category controller

I know it's intuitive to set current_entity_key with the product ID, but that would create many cache blocks for each product, so I set it with the current path. this way there is one cache foreach category path on PDP

empiricompany:
i can confirm it works, also on PDP the category of product it's selected now

empiricompany:
> @empiricompany can I test this problem without a full page cache plugin and with the RWD theme? cause I can't reproduce it on a vanilla OM installation (from a super quick test)

@fballiano Unfortunately, the latest Docker Desktop update completely destroyed my dev environment, but it's a problem I've always encountered on all installations without an FPC.

Are you sure you followed these steps?

1. Update the caches
2. Navigate to a product page
3. Go to the homepage

Issue: The category of the visited product should remain selected.

Additionally, I can say that I use the flat catalog and the config catalog/seo/product_use_categories = 1.

F1Red5:
@empiricompany never faced that issue and the category path should be in cache-info ...

```
    public function getCacheKeyInfo()
    {
        $shortCacheId = [
            'CATALOG_NAVIGATION',
            Mage::app()->getStore()->getId(),
            Mage::getDesign()->getPackageName(),
            Mage::getDesign()->getTheme('template'),
            Mage::getSingleton('customer/session')->getCustomerGroupId(),
            'template' => $this->getTemplate(),
            'name' => $this->getNameInLayout(),
            $this->getCurrenCategoryKey()
        ];
        $cacheId = $shortCacheId;

        $shortCacheId = array_values($shortCacheId);
        $shortCacheId = implode('|', $shortCacheId);
        $shortCacheId = md5($shortCacheId);

        $cacheId['category_path'] = $this->getCurrenCategoryKey();
        $cacheId['short_cache_id'] = $shortCacheId;

        return $cacheId;
    }
```

Maybe try https://github.com/AOEpeople/Aoe_TemplateHints for easier debugging.

F1Red5:
> Fix cache issue in navigation block

Mhh, can you share a screenshot with AoE-TH?

empiricompany:
i've checked it on latest main and i can confirm the issue.
But in rwd there is no css rule that highlight active menù item, so you must inspect html and you can view <li /> has class "active"

empiricompany:
In the last commit, I've highlighted the active category.

fballiano:
no no I knew there was no active category CSS, I was checking the source code for the "active" class, I'm tight with time lately, I'll try to recheck asap

fballiano:
It happens only when FLAT CATALOG is enabled (confirmed, also tested that your PR works and solves the problem).

I'm thinking, how is this happening? I mean, the fix doesn't seem to be strictly-related to the problem... anyway since it works I'm still approving, waiting a bit before merging to see if we get some other kind of feedback.

fballiano (approved):

empiricompany:
It's difficult to explain, but when you visit a PDP, a block cache of topmenu is stored without the specific tags of the category path. Then, when there is no category selected, like on the homepage, CMS, or contact pages, this cached block is retrieved without tags.

The PR also selects the category on the PDP according to the breadcrumbs.

Issue #4043, opened 2024-06-17
UPS REST _parseRestResponse doesn't return rates when only one service is available for a destination


iamniels:
After switching to UPS REST API, no rates were available for Cyprus. For Cyprus only one service is available on our UPS contract. In this case the response from UPS is:

```json
"RatedShipment": {
      "Service": {
        "Code": "65",
        "Description": ""
      },
      "RatedShipmentAlert": {
        "Code": "110971",
        "Description": "..."
      },
```
Instead of:
```json
"RatedShipment": [
      {
        "Service": {
          "Code": "07",
          "Description": ""
        },
```
So when there is only one service, an object is returned instead of an array. 

`UPS::_parseRestResponse` doesn't handle this case properly.

Can be fixed like this:
```php
    protected function _parseRestResponse(string $rateResponse)
    {
        $costArr = [];
        $priceArr = [];
        $errorTitle = '';
        if (strlen($rateResponse) > 0) {
            $rateResponseData = json_decode($rateResponse, true);
            if (@$rateResponseData['RateResponse']['Response']['ResponseStatus']['Description'] === 'Success') {
                $arr = $rateResponseData['RateResponse']['RatedShipment'] ?? [];
                if(isset($arr['Service'])){
                    $arr = [$arr];
                }
```

sreichel:
@all-contributors add @iamniels bug

PR #4044, opened 2024-06-17
UPS Rest API: handling cases where a single service is returned by UPS

fballiano:
Everything is explained in https://github.com/OpenMage/magento-lts/issues/4043

Fixes https://github.com/OpenMage/magento-lts/issues/4043

### DIFF
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -2341,6 +2341,10 @@ XMLAuth;
             $rateResponseData = json_decode($rateResponse, true);
             if (@$rateResponseData['RateResponse']['Response']['ResponseStatus']['Description'] === 'Success') {
                 $arr = $rateResponseData['RateResponse']['RatedShipment'] ?? [];
+                if (isset($arr['Service'])) {
+                    // Handling cases where a single service is returned by UPS
+                    $arr = [$arr];
+                }
                 $allowedMethods = explode(",", $this->getConfigData('allowed_methods') ?? '');
                 $allowedCurrencies = Mage::getModel('directory/currency')->getConfigAllowCurrencies();
                 foreach ($arr as $shipElement) {


app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
```diff
@@ -2341,6 +2341,10 @@ protected function _parseRestResponse(string $rateResponse)
             $rateResponseData = json_decode($rateResponse, true);
             if (@$rateResponseData['RateResponse']['Response']['ResponseStatus']['Description'] === 'Success') {
                 $arr = $rateResponseData['RateResponse']['RatedShipment'] ?? [];
+                if (isset($arr['Service'])) {
```
F1Red5, line 2344:
```suggestion
                if (array_key_first($arr)) {
```

F1Red5: Test: https://onlinephp.io?s=rZDRCoIwGIWvG-wdxghU6KZBXWTSRT1BXoqI6MQV6pgrCPHd2zTLYkYX7fKc8-_7z7_d8ZxDAMH8VFflEnnIavAxljT1c8YLWkq8QQ0EqHvYp-LKEjrWlLqvUi3h9QovRvKB1olgXLKq1C4erBaCtrXcJ5WYqcGQb_6EVdzwwe3YsRC6sF4hSmmifrP7KyyQFBfquH2GmDLklYHgGosovRTcVun4Fp3pLcqYqKXdIYKPZqGjp77OEPMMBCxDP0P0rWZmUzUKzE6oMO1bp6kfPA9NrNptegc%2C&v=8.0.30%2C7.4.33%2C8.3.0

fballiano: to me this is less readable

F1Red5: Mhhh, "If first array-key is a string ..."

Still prefer my suggestion, but another one ...

`if (!array_key_exists(0, $arr)) {`

F1Red5: Why do you mark it as resolved?

fballiano: because I prefer the code by @iamniels, easier to understand, also, 2 functions call instead of isset, there's no way it's better.

F1Red5: `Isset` goes wrong in this case:

```
$json3 = '{"RatedShipment": [
      {"Service": null}
]}';
```
Better? You prefer it. May let others think about it, too? Nope?

F1Red5: `isset` iterates over whole array, array_key_first does not. And is_string is not worth to mention.

F1Red5: In most case 'isset' can be replaced by more strict checks .... however ...

Flyingmana: https://stackoverflow.com/a/9522522

isset is O1, provide better proves next time

Flyingmana (approved):

empiricompany (approved):

app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
kiatng: A note from @F1Red5 : https://github.com/php/php-src/pull/3360

kiatng:
Ref issue #4043, the property `"RatedShipment"` is usually an array, but if there is only one item in it, then it is an object. 

`isset($arr['Service'])` looks for the key `"Service"` in `$arr` (JSON object) and return `true` if it's truthy.  If the key doesn't exist, it returns `false`. 

@F1Red5 proposal `is_string(array_key_first($arr))` returns `true` on the original JSON object, it doesn't check that the object has the property `"Service"`.

Based on this, I think testing `"RatedShipment"` for an object is may be better as it doesn't assume what keys `$arr` has. But may be it's necessary to check for the key "Service", so I don't know. I defer the decision to the PR author @fballiano.

fballiano:
the code object of this PR parses a maximum of 10 records, I would never trade readability and understandability of code for such a minor improvement in this context. if we were parsing millions then we should be having this talk, but having this talk now for this specific thing, in my opinion, doesn't make any sense. we should have this lever of participation with the other million things we should be doing instead of wasting energy for this.

I've personally tested all of the UPS implementation, @iamniels tested that this PR works for him (because he submitted it actually) so I really cannot understand all of this chat.

kiatng (approved):
@fballiano thanks for the clarification and your efforts!

PR #4047, opened 2024-06-19
Fixed ship-to address for UPS REST shipment API

ragnese:
### Description (*)
When a shipment is a "return", only the "ShipFrom" needs to change from the store owner to the customer. The "ShipTo" will already be the store owner and does not need to be changed.

### Related Pull Requests
N/A

N/A

Create a return shipment and try to get labels.

### Questions or comments
This is kind of funny, actually. If you look at the XML version of the code that creates the request object, you can see that the XML code does not REPLACE the "ShipFrom" and "ShipTo" when it's a return shipment--it actually ADDS another "ShipTo" node, which is ignored by UPS. Since UPS only looks at the first "ShipTo" node, the old XML code worked even though it was wrong. The REST code tried to mirror what the XML code does, but actually replaces the node instead of adding more.

### DIFF
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -1859,19 +1859,6 @@ XMLAuth;
             if ($request->getShipperAddressStateOrProvinceCode()) {
                 $address['StateProvinceCode'] = $request->getShipperAddressStateOrProvinceCode();
             }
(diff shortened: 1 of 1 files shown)


fballiano:
how do you create a return shipping label? I know only how to generate a normal shipping label, but I see know that the PDF is broken for me, it doesn't open.

fballiano:
@ragnese?

fballiano (approved):

PR #4059, opened 2024-07-01
Fixed undefined array key in Mage/Eav/Model/Config.php

kiatng:
### Description (*)
A ban person in OM community asked me to create this PR. I reviewed it and to me. this is a valid PR.

### Related Pull Requests
PR #4036

### DIFF
--- a/app/code/core/Mage/Eav/Model/Config.php
+++ b/app/code/core/Mage/Eav/Model/Config.php
@@ -571,7 +571,9 @@ class Mage_Eav_Model_Config
             }
             return $attributeCodes;
         } else {
-            return array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()]);
+            return isset($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()])
+                ? array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()])
+                : [];
         }
     }
 


app/code/core/Mage/Eav/Model/Config.php
```diff
@@ -571,7 +571,9 @@ public function getEntityAttributeCodes($entityType, $object = null)
             }
             return $attributeCodes;
         } else {
-            return array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()]);
+            return isset($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()])
```
pquerner, line 574: Wouldnt `return array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()] ?? [])` do the same?

davidhiendl: Changing array_keys to isset changes the return of the function  which is supposed to return attribute codes, not a boolean value.

pquerner: @davidhiendl It doesnt return boolean. I think you read that wrong. :)

davidhiendl: Yeah that code comment snippet function did cut of mid-change... fabulous.

fballiano: @pquerner I think `array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()] ?? [])` would still trigged the undefined array index warning

pquerner: @fballiano https://onlinephp.io/c/eebcc nope, unless I did something wrong.

Flyingmana:
@kiatng "@sreichel" got blocked permanently from this organization for toxic behavior, multiple violations of the CodeOfConduct and a clear statement he is not willing to apologize.

We can not accept any PRs which state to be created in his Name, so please remove the mention of him and avoid doing this again.

app/code/core/Mage/Eav/Model/Config.php
kiatng: `array_keys($this->_entityTypeAttributeIdByCode[$storeId][$entityType->getId()] ?? [])` works but not necessary better.

My break down:
1. The null coalescing operator (??) doc is [here](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op). `$result = $array['a'] ?? []` is the same as `$result = isset($array['a']) ? $array['a'] : []`
2. after the above, we have `array_kyes($result)`, an additional step if key 'a' is not set.

This `isset($array['a']) ? array_keys($array['a']) : []` works slightly better if key 'a' is not set. However, this statement `array_keys($array['a'] ?? [])` is shorter.

I vote for the lengthy one: it has clearer intent and works slightly better if the key is not set. Having said that, I am fine  either one.

Issue #4055, opened 2024-06-30
All attributes set to be used for search appear in layering navigation when there are no search results


ADDISON74:
Latest OM version cloned from the repository, DDEV + PHP 8.2 + MariaDB. 

Steps to reproduce this issue:

1. Set a few attributes with the option **[Used in Layered Navigation] as Filtrable (with results)**
2. Search in the Frontend for a word that doesn't return any result
3. Look at the layering navigation block. All your attributes appear with 0 results which is not a Magento correct behavior. If there are no results then no filterable attributes are displayed.

I investigated this issue in detail and I found that the changes made in this file

**/app/code/core/Mage/CatalogSearch/Model/Resource/Fulltext/Collection.php**

are causing the problem. 

**IMPORTANT NOTE:** I moved forward and I used the file version before the changes and the problem no longer occurs, but I am not able to search for words that returned results before. 

I compared the two versions and the changes are many so that I can get to the source of the problem.

fballiano:
thanks to `git bisect` I've identified that the problem is generated by commit fc97382ffcf9bb865346de078aafb7996564c7d1, PR https://github.com/OpenMage/magento-lts/pull/2993

@davidhiendl do you have any idea?

davidhiendl:
It's possible this problem is caused by newer PHP versions, will need some investigation. I cannot reproduce the problem on my 7.4/8.1 test env's.

davidhiendl:
Also cannot reproduce it on a clean 8.2 instance.

davidhiendl:
Just to be clear on the reproduction steps: 
- Add any option-type product attribute and assign it to an attribute-set
- assign a few products with options
- search for something that doesn't generate a result

ADDISON74:
It is not related to the PHP version. DDEV allows me to switch in a second any version from 7.4 to 8.3. If I am using a file version before your change in /app/core/local/... the layering navigation doesn't show the 0 values for all searchable attributes but I get another issue, the search feature is not functioning anymore even for words with results.

Steps for reproduction the issue are in the first post. It is not related to products with option.

ADDISON74:
It is not related to the PHP version. DDEV allows me to switch in a second any version from 7.4 to 8.3. If I am using a file version before your change in /app/core/local/... the layering navigation doesn't show the 0 values for all searchable attributes **but I get another issue, the search feature is not functioning anymore even for words with results**. Steps for reproduction the issue are in my initial post. It is not related to products with option.

[...]

PR #4063, opened 2024-07-01
Fixed regression introduced in #2993 where attributes are not correctly pre-filtered for the layered navigation

davidhiendl:
### Description (*)
Fix a regression introduced with the EAV overhaul where the filterable and filterable in search attribute options are not honored correctly.

Notes: For some reason the `Mage_CatalogSearch_Model_Layer::_prepareAttributeCollection` used to include a `is_visible` condition but not the parent for regular catalog layered nav. I've reproduced this behavior in the interest of having a usable fix quickly but I think it might cause the other issues found during the debugging of #4055

### Related Pull Requests
- introduced by https://github.com/OpenMage/magento-lts/pull/2993

- fixes https://github.com/OpenMage/magento-lts/issues/4055

See issue #4055

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Layer.php
+++ b/app/code/core/Mage/Catalog/Model/Layer.php
@@ -255,10 +255,8 @@ class Mage_Catalog_Model_Layer extends Varien_Object
     /**
      * Filter which attributes are included in getFilterableAttributes
      *
-     * @param Mage_Catalog_Model_Resource_Eav_Attribute $attribute
-     * @return  bool
      */
-    protected function _filterFilterableAttributes($attribute)
+    protected function _filterFilterableAttributes(Mage_Catalog_Model_Resource_Eav_Attribute $attribute): bool
     {
         return $attribute->getIsFilterable() > 0;
     }
--- a/app/code/core/Mage/Catalog/Model/Layer.php
+++ b/app/code/core/Mage/Catalog/Model/Layer.php
@@ -212,6 +212,9 @@ class Mage_Catalog_Model_Layer extends Varien_Object
             foreach ($setAttributeIds as $attributeId) {
                 if (!isset($attributes[$attributeId])) {
                     $attribute = $eavConfig->getAttribute(Mage_Catalog_Model_Product::ENTITY, $attributeId);
+                    if (!$this->_filterFilterableAttributes($attribute)) {
+                        continue;
+                    }
                     if ($attribute instanceof Mage_Catalog_Model_Resource_Eav_Attribute && $attribute->getIsFilterable()) {
                         $attributes[$attributeId] = $attribute;
                     }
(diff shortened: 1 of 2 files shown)


app/code/core/Mage/Catalog/Model/Layer.php
```diff
@@ -249,6 +252,17 @@ protected function _prepareAttributeCollection($collection)
         return $collection;
     }
 
+    /**
+     * Filter which attributes are included in getFilterableAttributes
+     *
+     * @param Mage_Catalog_Model_Resource_Eav_Attribute $attribute
+     * @return  bool
+     */
+    protected function _filterFilterableAttributes($attribute)
```
fballiano, line 261:
```suggestion
     */
    protected function _filterFilterableAttributes(Mage_Catalog_Model_Resource_Eav_Attribute $attribute): bool
```

app/code/core/Mage/CatalogSearch/Model/Layer.php
```diff
@@ -99,6 +99,17 @@ protected function _prepareAttributeCollection($collection)
         return $collection;
     }
 
+    /**
+     * Filter which attributes are included in getFilterableAttributes
+     *
+     * @param Mage_Catalog_Model_Resource_Eav_Attribute $attribute
+     * @return  bool
+     */
+    protected function _filterFilterableAttributes($attribute)
```
fballiano, line 108:
```suggestion
     */
    protected function _filterFilterableAttributes(Mage_Catalog_Model_Resource_Eav_Attribute  $attribute): bool
```

fballiano:
I can confirm this PR fixed the problem

app/code/core/Mage/CatalogSearch/Model/Layer.php
davidhiendl: Are we doing typed returns now, it would break 7.4 compatibility or has that been dropped?

fballiano: it would break 7.4 only for mixed types, which we can't use yet, but if it's a single type it should work fine:

https://onlinephp.io/c/f2834

ADDISON74:
I will test both PR that fixes the issues related to #2993. 

@davidhiendl - Thank you for your fast reaction.

ADDISON74:
@davidhiendl - Thank you for your fast reaction in solving this issue.

"Use in Layered Navigation" = USL
"Use in Search Results Layered Navigation" = USRLN

**Scenario 1**
ULN = Filterable (with results)
USRLN = No

For a word with results, all values ​​with zero will not be displayed for the attribute, which is correct, because since they are zero, they no longer need to be displayed.

**Scenario 2**
ULN = Filterable (with results)
USRLN = Yes

The same as before which is correct.

**Scenario 3**
ULN = Filterable (with results)
USRLN = No

For a word with **no results**, there are no attributes displayed in the layering navigation, which is correct.

**Scenario 4**
ULN = Filterable (with results)
USRLN = Yes

The same as before which is correct.

I'll let you decide if those changes proposed by @fbaliano are requested, Otherwise PR solves the problem reported by me.

fballiano:
since we have many instances of the same kind of typing in the main branch, I'll commit and merge those so that I can release 20.10 ;-)

PR #4068, opened 2024-07-02
Fixed null parameter warning in Mage_Wishlist_Block_Abstract

ADDISON74:
**Steps to reproduce this issue**

1. Log into a customer account
2. Add any product to Favorites
3. Once you are redirected to the Favorites page send the Favorites list to an email address

Take a look in this file **/var/log/system.log**. You will see this error, in my test with PHP version 8.3.8.

```Deprecated functionality: trim(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Wishlist/Block/Abstract.php on line 235```

### DIFF
--- a/app/code/core/Mage/Wishlist/Block/Abstract.php
+++ b/app/code/core/Mage/Wishlist/Block/Abstract.php
@@ -232,7 +232,7 @@ abstract class Mage_Wishlist_Block_Abstract extends Mage_Catalog_Block_Product_A
      */
     public function hasDescription($item)
     {
-        return ($item->getDescription() ?? '') != '';
+        return trim($item->getDescription() ?? '') != '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/code/core/Mage/Wishlist/Block/Abstract.php [hunk 1]
```diff
@@ -232,7 +232,7 @@ public function getEscapedDescription($item)
      */
     public function hasDescription($item)
     {
-        return trim($item->getDescription()) != '';
+        return ($item->getDescription() ?? '') != '';
```
fballiano, line 235:
```suggestion
        return trim($item->getDescription() ?? '') != '';
```

I think the trim is still needed because otherwise it changes the behaviour of the function ;-)

ADDISON74: I have nothing against. If it is needed then a way must be found to skip that error message.

fballiano: it is needed because otherwise a description with only spaces will be marked as "having a description", which is a different behavior from the current one.

but the `?? ''` is totally correct

PR #4073, opened 2024-07-03
Deprecated functionality: strlen() in the Tablerate.php file

ADDISON74:
In order to reproduce this issue use a table rate (uploading a CSV file in the Backend) in your test environment then follow the checkout steps to the end. 

Starting with PHP 8.1 to 8.3 you will get this error in the OM log file.

```
Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Shipping/Model/Resource/Carrier/Tablerate.php on line 131
```

My approach is using **strlen($variable ?? '')** but it could be **strlen((string) $variabile)** too. Please feel free to share your opinion for the best implementation.

### DIFF
--- a/app/code/core/Mage/Shipping/Model/Resource/Carrier/Tablerate.php
+++ b/app/code/core/Mage/Shipping/Model/Resource/Carrier/Tablerate.php
@@ -128,7 +128,7 @@ class Mage_Shipping_Model_Resource_Carrier_Tablerate extends Mage_Core_Model_Res
 
         $i = 0;
         $postcode = $request->getDestPostcode();
-        while (strlen($postcode) > 1) {
+        while (strlen($postcode ?? '') > 1) {
             $i++;
             $postcode = substr($postcode, 0, -1);
             $bind[':wildcard_postcode_' . $i] = "{$postcode}*";


app/code/core/Mage/Shipping/Model/Resource/Carrier/Tablerate.php
```diff
@@ -128,7 +128,7 @@ public function getRate(Mage_Shipping_Model_Rate_Request $request)
 
         $i = 0;
         $postcode = $request->getDestPostcode();
-        while (strlen($postcode) > 1) {
+        while (strlen($postcode ?? '') > 1) {
```
kiatng, line 131:
```suggestion
        $postcode = (string)$request->getDestPostcode();
        while (strlen($postcode) > 1) {
```
It's more efficient as it's outside the loop.

fballiano: good catch @ADDISON74, I like @kiatng version, it looks neater to me because it doesn't get executed inside the loop

fballiano (approved):

Issue #4076, opened 2024-07-03
Deprecated functionality: strlen(): Passing null to parameter #1 - Abstract.php


ADDISON74:
I am getting this error with PHP 8.3 in the system.log file:

```Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Wishlist/Controller/Abstract.php on line 88```

**/app/code/core/Mage/Wishlist/Controller/Abstract.php on line 88**

In order to reproduce this issue log into a customer account in the Frontend then add any product to the Wishlist. You will be redirected to the customer account. A part of the URL is **wishlist/index/index/wishlist_id/[PRODUCT_ID]/**. Press the [Add All to the Cart] button. In my case I am getting an error message because the product has custom options and I have to choose them. If you take a look into the /var/log directory you will see a nice system.log file, having 371 kB as size and inside the same line 1745 times.

**Comment by ADDISON74** [comment]
Maybe I am tired after a long busy day but this error is really funny. Looking into that file there is not strlen() function used.

**Comment by ADDISON74** [comment]
Commenting the line in the file the error doesn't appear in the system.log anumore.

```
 $qtysString = $this->getRequest()->getParam('qty');
        if (isset($qtysString)) {
            //$qtys = array_filter(json_decode($qtysString), '\strlen');
        }
```

**Comment by ADDISON74** [comment]
Commenting the bellow line in the file the error doesn't appear in the system.log anymore.

```php
 $qtysString = $this->getRequest()->getParam('qty');
        if (isset($qtysString)) {
            //$qtys = array_filter(json_decode($qtysString), '\strlen');
        }
```

This proves there is an issue in that line. The **array_filter** function filters the values of an array using a callback function. In this case the callback function is **strlen**. 

By searching the code I did not find an implementation of this function in the OpenMage, except the one in this file ```/app/code/core/Mage/Core/Helper/String.php```.

```php
/**
     * Retrieve string length using default charset
     *
     * @param string $string
     * @return int
     */
    public function strlen($string)
    {
        return is_null($string) ? 0 : iconv_strlen($string, self::ICONV_CHARSET);
    }
```

ADDISON74:
This is before Magento 1.8.1.0

```php
$qtys = $this->getRequest()->getParam('qty');
```

This is after Magento 1.8.1.0

```php
$qtysString = $this->getRequest()->getParam('qty');
[...]

PR #4083, opened 2024-07-04
Fixed null parameter warning in Mage_Wishlist_Controller_Abstract::allcartAction()

ADDISON74:
This PR is related to the issue #4076.

I am getting this error with PHP > 8.1 in the system.log file:

```Deprecated functionality: strlen(): Passing null to parameter #1 ($string) of type string is deprecated  in /var/www/html/app/code/core/Mage/Wishlist/Controller/Abstract.php on line 88```

**/app/code/core/Mage/Wishlist/Controller/Abstract.php on line 88**

In order to reproduce this issue log into a customer account in the Frontend then add any product to the Wishlist. You will be redirected to the customer account. A part of the URL is **wishlist/index/index/wishlist_id/[PRODUCT_ID]/**. Press the [Add All to the Cart] button. In my case I am getting an error message because the product has custom options and I have to choose them. If you take a look into the /var/log directory you will see a nice system.log file, having 374 kB as size and inside the same line 1745 times. Every line is for a product which is not in the wishlist. With a script someone could make the system.log file as big as he wants, an pretty fast.

**IMPORTANT NOTE**
This PR has 4 implementations inside (4 lines - 1 my choice, 3 commented), there may be others. Please share your opinions, at the end there will be only one line left. Once it is merged I will fix the same issue in the Import/Export module.

```php
array_filter($VARIABLE, fn ($tmpString) => strlen($tmpString ?? ''))
// $qtys = array_filter($VARIABLE, function ($tmpString) { return strlen($tmpString ?? ''); });
// $qtys = array_filter($VARIABLE, function ($tmpString) { return $tmpString !== NULL && strlen($tmpString); }); => From Drupal Community
// $qtys = @array_filter($VARIABLE, '\strlen');
```

### DIFF
--- a/app/code/core/Mage/Wishlist/Controller/Abstract.php
+++ b/app/code/core/Mage/Wishlist/Controller/Abstract.php
@@ -85,8 +85,9 @@ abstract class Mage_Wishlist_Controller_Abstract extends Mage_Core_Controller_Fr
 
         $qtysString = $this->getRequest()->getParam('qty');
         if (isset($qtysString)) {
-            $qtys = array_filter(json_decode($qtysString), fn($qtysString) => strlen($qtysString ?? ''));
-            // $qtys = array_filter(json_decode($qtysString), function ($qtysString) { return strlen($qtysString ?? ''); });
+            $qtys = array_filter(json_decode($qtysString), fn($tmpString) => strlen($tmpString ?? ''));
+            // $qtys = array_filter(json_decode($qtysString), function ($tmpString) { return strlen($tmpString ?? ''); });
+            // $qtys = array_filter(json_decode($qtysString), function ($tmpString) { return $tmpString !== NULL && strlen($tmpString); }); => From Drupal Community
             // $qtys = @array_filter(json_decode($qtysString), '\strlen');
         }
 
--- a/app/code/core/Mage/Wishlist/Controller/Abstract.php
+++ b/app/code/core/Mage/Wishlist/Controller/Abstract.php
@@ -85,7 +85,9 @@ abstract class Mage_Wishlist_Controller_Abstract extends Mage_Core_Controller_Fr
 
         $qtysString = $this->getRequest()->getParam('qty');
         if (isset($qtysString)) {
-            $qtys = array_filter(json_decode($qtysString), '\strlen');
+            $qtys = array_filter(json_decode($qtysString), fn($qtysString) => strlen($qtysString ?? ''));
+            // $qtys = array_filter(json_decode($qtysString), function ($qtysString) { return strlen($qtysString ?? ''); });
+            // $qtys = @array_filter(json_decode($qtysString), '\strlen');
         }
 
         /** @var Mage_Wishlist_Model_Item $item */
--- a/app/code/core/Mage/Wishlist/Controller/Abstract.php
+++ b/app/code/core/Mage/Wishlist/Controller/Abstract.php
@@ -85,10 +85,7 @@ abstract class Mage_Wishlist_Controller_Abstract extends Mage_Core_Controller_Fr
 
         $qtysString = $this->getRequest()->getParam('qty');
         if (isset($qtysString)) {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Wishlist/Controller/Abstract.php
```diff
@@ -85,7 +85,10 @@ public function allcartAction()
 
         $qtysString = $this->getRequest()->getParam('qty');
         if (isset($qtysString)) {
-            $qtys = array_filter(json_decode($qtysString), '\strlen');
+            $qtys = array_filter(json_decode($qtysString), fn($tmpString) => strlen($tmpString ?? ''));
```
kiatng, line 88: I vote line88 with arrow function, which is the same as line89.

fballiano: I also vote for 88 and removing the 3 lines that are commented out. I'd use a shorter variable name instead of $tmpString but ok who cares

app/code/core/Mage/Wishlist/Controller/Abstract.php
```diff
@@ -85,7 +85,10 @@ public function allcartAction()
...
+            // $qtys = array_filter(json_decode($qtysString), function ($tmpString) { return strlen($tmpString ?? ''); });
+            // $qtys = array_filter(json_decode($qtysString), function ($tmpString) { return $tmpString !== NULL && strlen($tmpString); }); => From Drupal Community
+            // $qtys = @array_filter(json_decode($qtysString), '\strlen');
```
fballiano, line 91:
```suggestion
            $qtys = array_filter(json_decode($qtysString), fn ($tmpString) => strlen($tmpString ?? ''));
```

this also fixed phpcs

ADDISON74: Great. Now all greens.

kiatng (approved):

app/code/core/Mage/Wishlist/Controller/Abstract.php
```diff
@@ -85,7 +85,7 @@ public function allcartAction()
 
         $qtysString = $this->getRequest()->getParam('qty');
         if (isset($qtysString)) {
-            $qtys = array_filter(json_decode($qtysString), '\strlen');
+            $qtys = array_filter(json_decode($qtysString), fn ($tmpString) => strlen($tmpString ?? ''));
```
kiatng, line 88: Remove space after `fn`:
```suggestion
            $qtys = array_filter(json_decode($qtysString), fn($tmpString) => strlen($tmpString ?? ''));
```

fballiano: @kiatng phpcs wants the space, in a previous commit there was no space and phpcs failed https://github.com/OpenMage/magento-lts/actions/runs/9800221620/job/27061795879

ADDISON74: In PR #4083 we had to use a space to avoid the issue related to PHPCS error.

fballiano (approved):
tested, works perfectly

app/code/core/Mage/Wishlist/Controller/Abstract.php
kiatng: Ref https://github.com/php-fig/fig-standards/blob/821ea5c5e1a30823f83e1bd56118861ded9ebbc6/accepted/PSR-2-coding-style-guide.md#6-closures:

> Control structure keywords MUST have one space after them; method and function calls MUST NOT.

Me think `fn($tmpString)` is a function call, but may be it is not.

sreichel:
On phpstan level 5 ...

>  ------ -------------------------------------------------------------------------------------------------------------------------- 
>    Line   Controller/Abstract.php                                                                                                   
>   ------ -------------------------------------------------------------------------------------------------------------------------- 
>    88     Parameter #2 $callback of function array_filter expects (callable(mixed): bool)|null, Closure(mixed): int<0, max> given.  
>   ------ --------------------------------------------------------------------------------------------------------------------------

PR #4089, opened 2024-07-07
Highlight popular countries in country select

ma4nn:
### Description (*)

This PR backports a feature from Magento 2 (in a nicer way with option groups) that allows to configure countries that will be shown on top of the country list ("popular countries").

Also the code duplication for `Mage_Checkout_Block_Onepage_Abstract::getCountryHtmlSelect()` and `Mage_Directory_Block_Data::getCountryHtmlSelect()` is removed.

In addition a new event `directory_get_top_countries` is added, that allows to customize the logic (e.g. to determine top countries automatically from quotes/orders).

Bildschirmfoto 2024-07-07 um 11 15 54

Repeat the following tests for 
- country dropdown in cart estimate shipping
- checkout billing
- checkout shipping
- customer account address edit
- customer account register

1. Without any configuration the country dropdown looks the same as before
2. Configure one top country in _System > Configuration > General > Countries Options > Top Countries_ and verify the country dropdown
3. Configure multiple top countries in _System > Configuration > General > Countries Options > Top Countries_ and verify the country dropdown

### DIFF
--- a/app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
+++ b/app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
@@ -18,6 +18,7 @@
  *
  * @category   Mage
  * @package    Mage_Checkout
+ * @phpstan-type Option array{label: string, value: non-falsy-string}
  *
  * @method \Mage_Sales_Model_Quote_Address getAddress()
  */
@@ -213,6 +214,7 @@ abstract class Mage_Checkout_Block_Onepage_Abstract extends Mage_Core_Block_Temp
 
         if ($options == false) {
             $options = $this->getCountryCollection()->toOptionArray();
+            $options = $this->sortCountryOptions($options);
             if ($useCache) {
                 Mage::app()->saveCache(serialize($options), $cacheId, $cacheTags);
             }
@@ -239,5 +241,29 @@ abstract class Mage_Checkout_Block_Onepage_Abstract extends Mage_Core_Block_Temp
     {
         return true;
     }
-    /* */
+
+    /**
+     * @template T of Option[]
+     * @param T $countryOptions
+     * @return array{0: array{label: string, value: Option[]}, 1: array{label: string, value: Option[]}}|T
+     */
+    private function sortCountryOptions(array $countryOptions): array
+    {
+        $topCountryCodes = $this->helper('directory')->getTopCountryCodes();
+        $headOptions = $tailOptions = [];
+
+        foreach ($countryOptions as $countryOption) {
+            if (in_array($countryOption['value'], $topCountryCodes)) {
+                $headOptions[] = $countryOption;
+            } else {
+                $tailOptions[] = $countryOption;
+            }
+        }
+
+        if (empty($headOptions)) {
+            return $countryOptions;
+        }
+
+        return [['label' => $this->__('Popular'), 'value' => $headOptions], ['label' => $this->__('Others'), 'value' => $tailOptions]];
+    }
 }
--- a/app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
+++ b/app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
@@ -18,7 +18,6 @@
  *
  * @category   Mage
  * @package    Mage_Checkout
- * @phpstan-type Option array{label: string, value: non-falsy-string}
  *
  * @method \Mage_Sales_Model_Quote_Address getAddress()
  */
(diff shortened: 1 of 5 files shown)


app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
```diff
@@ -195,31 +185,6 @@ public function getRegionHtmlSelect($type)
...
-
-        if ($options == false) {
-            $options = $this->getCountryCollection()->toOptionArray();
-            if ($useCache) {
-                Mage::app()->saveCache(serialize($options), $cacheId, $cacheTags);
-            }
-        }
-        return $options;
-    }
-
```
fballiano, line 222: If we remove a public method I think we have to move the PR to the `next` branch

ma4nn: Good point👍🏻 Didn't realize it was public because it seemed to be used only within that same class..
Alternative would be to leave it there (unused) and deprecated it?

kiatng: Vote to add `@deprecated` comment and suggest alternative method.

SuSiSonnenschein111:
Please re-check ... cache-key `DIRECTORY_COUNTRY_SELECT_STORE_` is used at several places ...

ma4nn:
> Please re-check ... cache-key `DIRECTORY_COUNTRY_SELECT_STORE_` is used at several places ... (not tested nothing, just an idea)

@SuSiSonnenschein111 thanks for your feedback (although it could have been a little bit more detailed).
By "several places" you mean `Mage_Catalog_Model_Product_Attribute_Source_Countryofmanufacture`? 

But I agree, the result before sorting should be cached, I've pushed the change and extended the testing scenarios.

Moreover probably `Mage_Catalog_Model_Product_Attribute_Source_Countryofmanufacture::getAllOptions()` and `Mage_Directory_Block_Data::getCountryHtmlSelect()` should also be refactored further to remove the code duplication..

app/code/core/Mage/Directory/Block/Data.php
```diff
@@ -169,4 +170,29 @@ public function getRegionsJs()
...
+     * @template T of Option[]
+     * @param T $countryOptions
+     * @return array{0: array{label: string, value: Option[]}, 1: array{label: string, value: Option[]}}|T
+     */
+    private function sortCountryOptions(array $countryOptions): array
+    {
+        $topCountryCodes = $this->helper('directory')->getTopCountryCodes();
+        $headOptions = $tailOptions = [];
+
+        foreach ($countryOptions as $countryOption) {
```
luigifab, line 184: it's "better" to not run the for each when `topCountryCodes` is empty

ma4nn: @luigifab better why? performance-wise this should be totally negligible..

app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
fballiano: ok for me too to deprecate it, maybe I'd do another PR on the next branch removing the method, if possible

app/code/core/Mage/Directory/Block/Data.php
fballiano: is it possible that $countryOptions could be null and generate a warning on the foreach?

ma4nn: > is it possible that $countryOptions could be null and generate a warning on the foreach?

no

app/code/core/Mage/Checkout/Block/Onepage/Abstract.php
ma4nn: alright, I re-added the public method 👍

fballiano (approved):
quickly tested on frontend and backend and works fine

app/code/core/Mage/Directory/etc/system.xml
```diff
@@ -279,6 +279,16 @@
                             <show_in_store>0</show_in_store>
                             <can_be_empty>1</can_be_empty>
                         </optional_zip_countries>
+                        <top_countries translate="label">
+                            <label>Top Countries</label>
```
fballiano, line 283: we've to add this to a language CSV

app/code/core/Mage/Directory/Block/Data.php
```diff
@@ -169,4 +170,29 @@ public function getRegionsJs()
...
+            } else {
+                $tailOptions[] = $countryOption;
+            }
+        }
+
+        if (empty($headOptions)) {
+            return $countryOptions;
+        }
+
+        return [['label' => $this->__('Popular'), 'value' => $headOptions], ['label' => $this->__('Others'), 'value' => $tailOptions]];
```
fballiano, line 196: I see that "Popular" is already present in some language file, but not the "directory" one

app/code/core/Mage/Directory/Block/Data.php
luigifab: yes, but this one + another one + another one......................

sreichel:
Looks good so far ..., but for some reason i have a wrong translation for country selction dropdown ...

country

It is not related to this PR ... i see the same on main-branch.

The only ccourence of "Allemagne" is in vendor/shardj/zf1-future/library/Zend/Locale/Data/fr.xml.

sreichel (dismissed):

sreichel (approved):

kiatng (approved):

sreichel:
@kiatng cs ... :P

sreichel (approved):

PR #4097, opened 2024-07-15
GitHub workflows: run PHPUnit locally, removed Sonar (too slow), removed unsupported phpunit reports

fballiano:
Out workflows are too slow, this also means we're using a lot of computational resources (meaning wasted CO2 and time).

This PR:
- run PHPUnit locally (also, not based on the openmage/testfield repository for a more standard implementation)
- removed unsupported phpunit reports
- ~~fixed publication of report to the PRs~~ removed publication of phpunit results (no other workflow we have does that, it's not working on PR unless we add some steps and it's not super beautiful anyway IMHO)
- removed Sonar (too slow, about 9 minutes each sun)


Flyingmana (dismissed):
I originally added sonarcube, but there was always a problem with properly syncing the file paths for the code coverage mapping.

fballiano:
from the EnricoMi/publish-unit-test-result-action clover and crap4j are not supported and they're slow to generate, but I've to check https://github.com/EnricoMi/publish-unit-test-result-action/blob/v2.17.0/README.md#support-fork-repositories-and-dependabot-branches before this PR fully works (it was working on my fork)

fballiano:
After reading https://github.com/EnricoMi/publish-unit-test-result-action/blob/v2.17.0/README.md#support-fork-repositories-and-dependabot-branches I've removed the publication of phpunit results (no other workflow we have does that, it's not working on PR unless we add some steps and it's not super beautiful anyway IMHO and it wasn't working also before). hope that's ok

PR #4105, opened 2024-07-20
Removed v19 related code from release workflow

sreichel:
There will be no "v19" release anymore, so we can remove that steps.

(or is it worth to find the issue?)


ADDISON74:
My opinion, it would be a good idea to forget about version 19, but the branch should be available for one more year. Although I stated before to patched it for the main reason to avoid issues for those who continue using it, recently I had to upgrade a big store to the latest version v20.0.10, due to a server upgrade. To my surprise everything went so smooth, and the extensions/modules required few changes to run PHP 8.3. I told the story here https://github.com/OpenMage/magento-lts/discussions/4082.

I encourage those who have not upgraded yet, to use a test environment such as DDEV and take the step to v20, instead of analyzing the migration to other platforms.

alexh-swdev:
But what about "your" promise `19.4.x will receive PATCH updates until April 4, 2025.` (See the main)?

(I am not using v19 anymore, nor have I promised to support it until next year ;) but...)

sreichel:
You are right, but I cannot answer it.

@Flyingmana @colinmollenhour

Flyingmana:
> You are right, but I cannot answer it.
> 
> @Flyingmana @colinmollenhour

Iam not a maintainer anymore, you have to ask my successor as soon as one was selected.

sreichel:
> Iam not a maintainer anymore, you have to ask my successor as soon as one was selected.

Okay. @colinmollenhour can you please port it to v19 and make a new release?

sreichel:
Maybe discuss to remove zip files at all.

PR #4111, opened 2024-07-22
Rector: CQ - UnusedForeachValueToArrayKeysRector

sreichel:
Refactor code step by step using rector ...

See https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#unusedforeachvaluetoarraykeysrector

Edit: there are some comments that state to dislike it - for personal prefs, for readability, for not having used it before ...

Imho ... it makes the code better. PHPMD will not complain about unused local variables. It makes it more readable, b/c i can see we deal only with the array keys. This sniff could be excluded, but isnt it better to make use of quality tools w/o ignore some rules?


ADDISON74 (approved):
I checked visually all the changes and they are fine.

app/code/core/Mage/Adminhtml/Block/Dashboard/Graph.php
```diff
@@ -215,7 +215,7 @@ public function getChartUrl($directUrl = true)
                     $dateStart->addMonth(1);
                     break;
             }
-            foreach ($this->getAllSeries() as $index => $serie) {
+            foreach (array_keys($this->getAllSeries()) as $index) {
```
empiricompany, line 218: Why change the code like this? It seems less readable and possibly less performant.

empiricompany: what about this?
https://stackoverflow.com/a/29556153 

For me, is a questionable choice

If it is among the rules of Rector, it is not obvious that we have to use it. It seemed more sensible to me to approve the short syntax for the echo <?=.

sreichel: I know that post at SE ... it looks like overhead, but "it's not always a bad thing if a code generates more opcode ".

I'd prefer it. No exclude rules for phpmd/rector ... Finally its about personal preferences.

Short echo syntax ... this would change so many places that are referenced in docs/posts (like at SE) for only to save some bytes for no benefit.

empiricompany: seems to me an example of "Premature optimization is the root of all evil."

short syntax is not for save bytes but readabily and esthetics

also another rule that adds time to the very long pipeline that I have to wait for every PR

ADDISON74: The discussion about the syntax is off-topic. It can be discussed elsewhere, it was proposed in a PR and I did not agree with it. In my opinion, the whole template  inherited from Magento is not readable at all, if I compare for example with Twig.

sreichel: Twig is also off-topic, but an interssting idea. :P

> seems to me an example of "Premature optimization is the root of all evil."

Well. I like clean code. Having variables that are not used looks not clean (to me).

empiricompany: This means that every time I have a foreach where I use only the keys, I have to use array_keys too.
This results in two internal loops of the same array.
What do others think about this? @colinmollenhour @kiatng @luigifab @drobinson @LeeSaferite @elidrissidev @tmotyl ?

sreichel: Performance isnt an issue. And how ofter does it happen with your own code?

empiricompany: I say if i have to commit new code after this PR i have to follow always this standard.
About performance it results into 2 loops of the same array multiplied by every instance where the code occurs

sreichel: Dont care about that two loops. Really.

You dont need to follow this standard. At the end there should be a working rector-config that changes it. (Working on it, but I wawnt to apply the rules step by step, because PRs would become to large)

sreichel: @empiricompany i am a fan of micro-optimization, but here I'd prefer "cleaner" code over it. 

An o/c rector rule can be disabled. Same for phpmd.

empiricompany: ok, but this for me is not cleaner it's a horror

elidrissidev: My 2 cents: I prefer just leaving the code alone unless there is noticeable performance or readability improvement in the new one, both of wish I don't see here.

ADDISON74: Using Rector is fine.

Now it depends on who writes the code, if he is a young programmer he will use the version proposed by Sven, which he is already used to. If he is an older programmer, it will be difficult for him to adapt to the new code. 

As for me, I think I have to adapt.

sreichel: Btw ... the the people behind rector are highly active. My last issue was fixed within 15 minutes and a new release including it was made today.

phpstan and rector etc.  are tools that are used, supported, sponsored by far far bigger projects ... why? It makes code better. Standardizes it.

We should use it too.

empiricompany: Soryy but I can't see this as a standard and I never seen it in other major framework

sreichel: Not a major one ....

https://github.com/search?q=repo%3Adrupal%2Fdrupal%20%22foreach%20(array_keys%22&type=code

empiricompany: Oh! I would have never expected that!

https://github.com/search?q=repo%3Alaravel%2Fframework+foreach+%28array_keys&type=code
https://github.com/search?q=repo%3Asymfony%2Fsymfony%20foreach%20(array_keys&type=code

Even though I still don't like it, and I believe they are also victims of these absurd rules, you have my approval.

sreichel: At the end of the day ... how often do you check that source-code?  Let it fix  and follow "common" rules. ;)

ADDISON74: Drupal is not a CMS for everyone, the cream of the programmers gathered here and I always found solutions from which to draw inspiration. Many, many years ago I had the choice to study between Joomla and Drupal and I stayed attached to Drupal, even though it's damn hard because they only use what is new and you can't really keep up with the developers.

sreichel: > Even though I still don't like it

;) 

I think all, ALL big projects use reactor/phpstan/phpcs to ensure some code quality.  Maybe it looks strange whem you have never used it, it makes sense to me. Readablity come form how often yo read it :P

sreichel: @ADDISON74 you cant compare Drupal or Typo3 with Joomla or Wordpress .... thats another level.

empiricompany (approved):

app/code/core/Mage/Adminhtml/Block/Dashboard/Graph.php
luigifab: I dislike it, I prefer `foreach ($things as $key = $nul) ...`

I'm also using Rector and PhpStorm ins pect code when I'm not traveling, with "groups enabled", but I disable some rules.

sreichel: I thought it would be a rule everybody agrees with ... :P

app/code/core/Mage/Adminhtml/Block/Page/Menu.php
```diff
@@ -164,7 +164,7 @@ protected function _buildMenuArray(Varien_Simplexml_Element $parent, $path = '',
 
         uasort($parentArr, [$this, '_sortMenu']);
 
-        foreach ($parentArr as $key => $value) {
+        foreach (array_keys($parentArr) as $key) {
             $last = $key;
         }
         if (isset($last)) {
```
kiatng, line 170:
```suggestion
        if (!empty($parentArr)) {
            $last = array_key_last($parentArr);
```

app/code/core/Mage/Adminhtml/Model/System/Config/Source/Admin/Page.php
```diff
@@ -109,7 +109,7 @@ protected function _buildMenuArray(?Varien_Simplexml_Element $parent = null, $pa
 
         uasort($parentArr, [$this, '_sortMenu']);
 
-        foreach ($parentArr as $key => $value) {
+        foreach (array_keys($parentArr) as $key) {
             $last = $key;
         }
         if (isset($last)) {
```
kiatng, line 115:
```suggestion
        if ($parentArr !== []) {
            $last = array_key_last($parentArr);
```

sreichel: Nice idea! Let me check.

sreichel: There is another rule for array_key_last that can be extended to cover that ...

https://github.com/rectorphp/rector/blob/main/rules/Php73/Rector/FuncCall/ArrayKeyFirstLastRector.php

sreichel: `array_key_last`  returns null for an empty array ... so it should be enought to check that `$last` is nut null.

kiatng (approved):

ADDISON74 (dismissed):

app/code/core/Mage/Payment/Block/Form/Cc.php
```diff
@@ -47,7 +47,7 @@ public function getCcAvailableTypes()
             $availableTypes = $method->getConfigData('cctypes');
             if ($availableTypes) {
                 $availableTypes = explode(',', $availableTypes);
-                foreach ($types as $code => $name) {
+                foreach (array_keys($types) as $code) {
                     if (!in_array($code, $availableTypes)) {
                         unset($types[$code]);
                     }
```
kiatng, line 53: Alternate:
```suggestion
                return array_diff_key($types, array_flip($availableTypes));
```

kiatng (approved):

ADDISON74 (approved):

kiatng (approved):

PR #4116, opened 2024-07-24
Phpstan fixes

sreichel:
### Description (*)

Mostly fixing docblock and adding a few inline docklocks and setting some default values.


app/code/core/Mage/ImportExport/Block/Adminhtml/Export/Filter.php
```diff
@@ -434,11 +434,11 @@ public function getFilterElementName($attributeCode)
...
      *
      * @param Mage_Catalog_Model_Resource_Eav_Attribute $row
-     * @return false
+     * @return string
      */
     public function getRowUrl($row)
     {
-        return false;
+        return '';
     }
```
empiricompany, line 442: Agreed for docblock notation, but this changes the behavior of the function, from bool to string

sreichel: Its to match parrent method.

Behavoir does not change. (Go to admin - system - import/export - export - select entity type product)

Should be fixed at some other places too.

sreichel:
```

message: "#^Return type \\(false\\) of method Mage_ImportExport_Block_Adminhtml_Export_Filter\\:\\:getRowUrl\\(\\) should be compatible with return type \\(string\\) of method Mage_Adminhtml_Block_Widget_Grid\\:\\:getRowUrl\\(\\)$#"
count: 1
path: app/code/core/Mage/ImportExport/Block/Adminhtml/Export/Filter.php
```

kiatng (approved):
LGTM

ADDISON74 (approved):

PR #4117, opened 2024-07-25
Remove `require_once` from Zend_ classes

sreichel:
### Description (*)

- see  https://github.com/OpenMage/magento-lts/discussions/3462#discussioncomment-9450261

Used that slightly modified script to create a patch using `ddev vendor-patches`.

### Related Pull Requests

- see OpenMage/magento-lts#3989

@elidrissidev maybe want to test it?

PR #4118, opened 2024-07-25
Replace ZF1Future with ZF1S

sreichel:
_... later ..._

ToDo:

- diff
- apply OM patches


ADDISON74:
What are the advantages changing ZF1- Future with ZF1S?

sreichel:
IMHO ... we dont have to add the complete Zend framework ... only the parts/packages we need.

And ... i'd like fix https://github.com/OpenMage/magento-lts/discussions/3462

https://github.com/OpenMage/magento-lts/pull/4117 did not solve it. As @elidrissidev and my tests confirmed ZF1S works for some reason. I'm to lazy to search for ... better spent a few hours he. I already did it with ZF1Future so it will not take that long again ...

sreichel:
Green for now ...

- diff for php8 changes
- apply OM patches
- add PRs to ZF1S (to replace that patches)

... done :)

elidrissidev:
There are some important changes that `zf1s` doesn't include by default such as the flag to prevent PDO from casting numeric types in PHP 8:

https://github.com/Shardj/zf1-future/blob/94d85abd45bdcb806b24a9adbe7c9eba65d54ba3/library/Zend/Db/Adapter/Pdo/Mysql.php#L113-L119

elidrissidev:
> (btw ... i'd like to have casting to int/not casting to string, .... drop php 7!!! :P)

I prefer that too, but there will be a lot of hard to find bugs, so I suggest leaving it as is for a separate PR in the future.

ADDISON74:
Don't fix something that works :)

I would not drop support for PHP 7.4, I am using servers running Debian 11 and it is entering the LTS program until June 2026 and which I will not upgrade any time soon.

sreichel:
> Don't fix something that works :)

Let us dream ... :P

O/c thats something that will not happen in near future.

empiricompany:
+1 to drop php 7 support
https://www.php.net/eol.php

sreichel:
On hold.

There are a lot of fixes in ZF1Future, but not in ZF1S.

I already started to prepare PRs, but this will take some time to make it good.

I like that mono-repo idea, but atm it seems to be a step backwards.

So we have to search for the not working autoloading from app/local|community|core ...

sreichel:
> Debian LTS: https://wiki.debian.org/LTS
Debian 11: https://www.debian.org/News/2021/20210814 => PHP 7.4 - End of August 2026

We have droped php7.3 support in 03/2023.

Ihho it would be great to could make use of fancy php8 stuff.

https://accesto.com/blog/php-performance-improvement-features/

sreichel:
Later ... maybe.

kiatng:
zf1s as I understand, split the ZF components into individual composer packages. The intention is to change the dead ZF1 gradually to something non-dead. So, for example, Zend PDF can be swapped with [FPDF](http://www.fpdf.org/). The problems:

1. Require big changes to the core
2. 3rd-party codes that rely on ZF will break

sreichel:
> 2\. 3rd-party codes that rely on ZF will break

Only if a not included Zend_* module is not included.

The biggest problem are the differencies between ZF1Future and ZF1S.

PR #4121, opened 2024-07-26
fix some 'undefined array key' warnings

alexh-swdev:
### Description (*)
Sometimes, the array key "value" is not available, showing warnings in the logfile. This PR adds some isset-checks to avoid them.

See for example
`Warning: Undefined array key "value"  in /var/www/XXX/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field.php on line 86`

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field.php
+++ b/app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field.php
@@ -83,7 +83,7 @@ class Mage_Adminhtml_Block_System_Config_Form_Field extends Mage_Adminhtml_Block
                 $defTextArr = [];
                 foreach ($options as $k => $v) {
                     if ($isMultiple) {
-                        if (is_array($v['value']) && in_array($k, $v['value'])) {
+                        if (isset($v['value']) && is_array($v['value']) && in_array($k, $v['value'])) {
                             $defTextArr[] = $v['label'];
                         }
                     } elseif (isset($v['value'])) {
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Select.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Select.php
@@ -54,7 +54,7 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Filter_Select extends Mage_Adminht
      */
     protected function _renderOption($option, $value)
     {
-        $selected = (($option['value'] == $value && (!is_null($value))) ? ' selected="selected"' : '');
+        $selected = (isset($option['value']) && ($option['value'] == $value && (!is_null($value))) ? ' selected="selected"' : '');
         return '<option value="' . $this->escapeHtml($option['value']) . '"' . $selected . '>' . $this->escapeHtml($option['label']) . '</option>';
     }
 
@@ -66,7 +66,7 @@ class Mage_Adminhtml_Block_Widget_Grid_Column_Filter_Select extends Mage_Adminht
         $html = '<select name="' . $this->_getHtmlName() . '" id="' . $this->_getHtmlId() . '" class="no-changes">';
         $value = $this->getValue();
         foreach ($this->_getOptions() as $option) {
-            if (is_array($option['value'])) {
+            if (isset($option['value']) is_array($option['value'])) {
                 $html .= '<optgroup label="' . $this->escapeHtml($option['label']) . '">';
                 foreach ($option['value'] as $subOption) {
                     $html .= $this->_renderOption($subOption, $value);


sreichel:
I'd prefer `array_key_exists()`

alexh-swdev:
In my personal older fix for Select, I indeed used `array_key_exists()` because I also prefer it. However, in Field.php there already is a `isset()` check for the array key so I had the impression OM does (or at least Magento did) prefer `isset` over `array_key_exists`.

I was curious about the actual difference so I looked up https://stackoverflow.com/a/3210982 . So the use of `isset` made sense and I used this.

But in the end, I don't care. If you insist I'll change it?

sreichel:
I try to avoid `isset()` whenever it is possible and use `is_null()` or `array_key_exists()` or similar. For me it makes it better readable.

ADDISON74 (dismissed):

sreichel (approved):

app/code/core/Mage/Adminhtml/Block/System/Config/Form/Field.php
```diff
@@ -83,10 +83,10 @@ public function render(Varien_Data_Form_Element_Abstract $element)
                 $defTextArr = [];
                 foreach ($options as $k => $v) {
                     if ($isMultiple) {
-                        if (is_array($v['value']) && in_array($k, $v['value'])) {
+                        if (array_key_exists("value", $v) && is_array($v['value']) && in_array($k, $v['value'])) {
```
S0FTWEX, line 86: The rest of code uses single quotes, I suggest to keep it up, so change it to `array_key_exists('value', $v)`.

alexh-swdev:
....hope you have write access :p

sreichel (approved):

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Select.php
```diff
@@ -66,7 +66,7 @@ public function getHtml()
         $html = '<select name="' . $this->_getHtmlName() . '" id="' . $this->_getHtmlId() . '" class="no-changes">';
         $value = $this->getValue();
         foreach ($this->_getOptions() as $option) {
-            if (is_array($option['value'])) {
+            if (array_key_exists('value', $option) && is_array($option['value'])) {
                 $html .= '<optgroup label="' . $this->escapeHtml($option['label']) . '">';
                 foreach ($option['value'] as $subOption) {
```
elidrissidev, line 71: If `$option['value']` is null then using in the foreach should still cause a warning, no?

sreichel (approved):

S0FTWEX (approved):

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Select.php
```diff
@@ -54,7 +54,16 @@ protected function _getOptions()
...
     protected function _renderOption($option, $value)
     {
-        $selected = (($option['value'] == $value && (!is_null($value))) ? ' selected="selected"' : '');
+        if (!array_key_exists('value', $option)) {
+            if (!array_key_exists('label', $option)) {
+                return '';
+            }
+
+            $lbl =  $this->escapeHtml($option['label']); // or just empty as well?
+            return '<option value="' . $lbl . '">' . $lbl . '</option>';
```
kiatng, line 63: What's the reason for setting label (instead of empty string) to option value?

alexh-swdev: I did it, so each option has a different / its own value, what the 'value' key in the map was supposed to provide.
If you think that's not required, we can leave the value attribute of the tag empty.

kiatng:
The description of this PR:

> Sometimes, the array key "value" is not available, showing warnings in the logfile

The warning happens because there are mistakes in supplying the options in custom grids and forms. When rendering the HTML select element, the rendering functions expect the var `$option` to have 2 keys: `value` and `label`. If these keys are absent, exception should be thrown. The warnings in the log file are very helpful and should not be silenced. Instead, the code that causes the error should be corrected. (We shouldn't accept bad code from third party.)

Therefore, I object to this PR.

alexh-swdev:
As I mentioned in https://github.com/OpenMage/magento-lts/pull/4125#issuecomment-2266627620 I think OM should handle this in  a proper way.
"Proper way" may be arguable:
* Showing  the option as good as possible (like this PR is done right now)
* Filtering it out and don't show the option at all (alternative, like written in the code comment)
* Throw an exception like you proposed.

Imho, the current way is neither fish nor fowl....

sreichel:
How about to log an InvalidConfigException?

sreichel:
How about to validate the config and log what's wrong?

kiatng:
> How about to validate the config and log what's wrong?

Nice idea. But if there is no developer, the log would just pile up. However `if ($isDevelopMode) $adminSession->addWarning($warn)`?

alexh-swdev:
Now, the log piles up, too, with the warnings ;) (ok, depending on cfg)

So I guess, this is not really solvable without BC, be it filter out wrong options and/or throwing exceptions, in 20.x, maybe it can be done in 21.x?

app/code/core/Mage/Adminhtml/Block/Widget/Grid/Column/Filter/Select.php
sreichel: I'd prefer an empty string.

Maybe you can share the  extension that causes the problem, to reproduce it?

alexh-swdev: Ok, I'll change it to empty string then.
It's an Amazon Pay add-on. I hope I can provide it in the next days

sreichel: Maybe its better to fix the extension itself. There some forks of it ...

https://github.com/kirchbergerknorr/magento1_creativestyle_amazon-payments
https://github.com/mothership-gmbh/magento1_creativestyle_amazon_payments


Or create a repo, or better update https://github.com/OpenMageModuleFostering/creativestyle_amazonpayments

Issue #3308, opened 2023-06-07
FedEx WSDLS will be retired on May 15, 2024, need to use RESTful API


xqiu:
### Description (*)
On FedEx https://www.fedex.com/en-us/developer/web-services/process.html page, it states: FedEx Web Services Tracking, Address Validation, and Validate Postal Codes WSDLS will be retired on May 15, 2024. The SOAP based FedEx Web Services is in development containment and has been replaced with FedEx RESTful APIs. To learn more and upgrade your integration from Web Services to FedEx APIs, please visit the [FedEx Developer Portal](https://developer.fedex.com/api/en-us/home.html).

### Expected behavior (*)
Open Mage USA module should create a FedEx RESTful shipping module

### Benefits
So that openMage can still be integrated with FedEx after May 15, 2024

### Additional information
N/A

fballiano:
@xqiu @ajbonner does M2 have the new API?

xqiu:
@fballiano Yes, see https://github.com/magento/magento2/issues/34642

fballiano:
well, it's now too late, we'll have to see what happens

fballiano:
Seems it's been moved to 21st Aug

<img width="1140" alt="Screenshot 2024-05-08 alle 10 25 26" src="https://github.com/OpenMage/magento-lts/assets/909743/57583748-145b-4254-b4ce-b41a3077df55">

vipin12117:
Use FedEx Rest API plugin auto address check with dimensional shipping with packaging rules

https://www.ecomplugins.com/magento2-fedex-shipping-carrier-with-rest-api

sreichel:
@vipin12117 deleted comment ... dont advertise M2 modules here.

xqiu:
I just used https://github.com/ShipStream/fedex-rest-php-sdk to implement a change to implement fedex ship rate collection and package tracking.  But the change requries 2 things that might not be OK to do pull request.  Please advice.

The commit: 
https://github.com/OpenMage/magento-lts/commit/c420e11e738cad4d2fa023a6e90b212f4ebfb84f

The problems:
1. uses shipstream/fedex-rest-sdk, but currently it's not updated with fedex tracking API schema, so need to use a patch branch for it to solve it simply.  https://github.com/ShipStream/fedex-rest-php-sdk/issues/3 

[...]

PR #4124, opened 2024-07-28
Make `php-8.1` minimum requirement

sreichel:
We removed 7.3 support  in 03/2003 ... now - plus one year later - its time to make the next step.

PHP8 offers a lot of good features, i'd like to use when writing extension but i am limited due to php7 support.

Worth a read ... https://accesto.com/blog/php-performance-improvement-features/

According to packagist.org stats only 20% still use php7.4 with latest releases.

https://packagist.org/packages/openmage/magento-lts/php-stats#20.10

---

Note 1: If you have modules/extensions that dont work with php8, drop me a note! This should be no reason to stay on an old php version!

Note 2: OpenMage will still proberly work with older versions too - as long we not introduce backwards-incomaptible changes.

Note3 : with https://github.com/rectorphp/rector you have a tool to refactor your code for php8.


composer.json
```diff
@@ -31,7 +32,6 @@
         "phpseclib/mcrypt_compat": "^2.0.3",
         "phpseclib/phpseclib": "^3.0.14",
         "shardj/zf1-future": "1.24.0",
-        "symfony/polyfill-php74": "^1.29",
         "symfony/polyfill-php80": "^1.29",
         "symfony/polyfill-php81": "^1.29",
         "symfony/polyfill-php82": "^1.29"
```
kiatng, line 37:
```suggestion
        "symfony/polyfill-php82": "^1.29",
        "symfony/polyfill-php83": "^1.29"
```

ADDISON74 (approved):

colinmollenhour:
Agreed with the changes, but I believe this should be in "next" so that it is in the next major version update.

> MAJOR version when you make incompatible API changes
> - ...
> - dependencies updated in major ways (e.g. dropping support of PHP 7.4 or MySQL 5.6)

> All new PRs are based on either "main" branch for MINOR and PATCH updates or "next" branch for MAJOR updates

Since v21 was not released yet this could still make it into v21.

sreichel:
https://github.com/OpenMage/magento-lts/discussions/4317

Closed for now. :(

sreichel:
ref https://github.com/OpenMage/magento-lts/issues/3308#issuecomment-2475316578

sreichel:
Bump.

[Poll](https://github.com/OpenMage/magento-lts/discussions/4317) 68% at the moment.

Current composer usage:

- https://packagist.org/packages/openmage/magento-lts/php-stats#20.11
- https://packagist.org/packages/openmage/magento-lts/php-stats#20.12

PHO stats

- https://stitcher.io/blog/php-version-stats-january-2025

sreichel:
> Since v21 was not released yet this could still make it into v21.

There will be no v21 - from the current "next" branch. Not?

I'd like to close "next" after backporting some changes: \*

- code: there should be much PR with breaking changes, that are really required
- admin theme: backport and ignore possible breaking changes. (Who ever used a custom admin theme?)
- rwd theme: i'd like to not throw awy that work made there. I'd move that PR to a new theme as started here: https://github.com/OpenMage/magento-lts/pull/4267

\* v21-betas should be removed from packagist. beta2 has wrong dependecies - and it is not used at all (see download stats)

kiatng:
I'm for PHP8+, but as pointed out by @colinmollenhour, the world seems not quite ready yet. I agree for v21, in which case, we can raise the bar to min PHP8.4. Something like 

- v20, PHP7.4, we have to decide when to freeze and when to stop support
- v21, PHP 8.4+, start when v20 is freeze? Avoid maintaining 2 versions concurrently
- v22, PHP 9+

Not sure if it's inline with https://github.com/OpenMage/rfcs/blob/main/accepted/0002-release-schedule.md

sreichel:
Moving to php8.1 is no BC-breaking change. You only cant update 20,13 (?).

v21/next contains breaking changes ... we cant use it as it is.

>  Avoid maintaining 2 versions concurrently

20, 21, 22? Why not continue with v20 to keep it simple?

kiatng:
> 20, 21, 22? Why not continue with v20 to keep it simple?

PHP8 has breaking changes and deprecations, so to me, it makes sense to have it in v21. It will reflect the major updates and ensure **clarity** for users and developers about the compatibility requirements.

addison74:
As a personal opinion, I would like to refer to OpenMage as one only version, with the possibility of releasing new versions periodicaly. I would use the project's front page or a pin to the discussions to prepare the community for major changes in advance, so that all those interested in upgrading can to read and take actions. For example, on date X the minimum PHP version will be 8.1, on date Y Prototype library will be removed from OpenMage version without Prototype. This v21/next would act more like a testing version until the new PRs are safe to be used in the main branch. All the changes that could generate BCs will be discussed in public then a deadline is set. Maybe there are people who want to sponsor some useful implementations and then those who implement them can be paid for their work.

I have been using PHP 8.3 and 8.4 for some time now, both in production and testing. Regarding this PR, I would add it first to the next branch (for testing), setting a date for a new version of OpenMage (for example July 1st 2025), get ready a warning effective from today and attach a discussion section for it. If there are no differences of opinion, on the set date PHP 8.1 becomes the minimum PHP for OpenMage.

sreichel:
> > 20, 21, 22? Why not continue with v20 to keep it simple?
> 
> PHP8 has breaking changes and deprecations, so to me, it makes sense to have it in v21. It will reflect the major updates and ensure **clarity** for users and developers about the compatibility requirements.

What do you think about that?

https://medium.com/@sampart/semantic-versioning-when-you-change-the-required-programming-language-version-16a3a3555c95

> **Conclusion**
> When you change the version of PHP supported by a project using Semantic Versioning, increment your _minor_ version number.

sreichel:
> there must be a v21 to remove PHP <8.1 since that would break ~42% of installations according to packagist.

Please check where the 42% came from. Its from users that run older OM-releases. Please see numbers for v20.11/20.12.

And ... a lot of current PRs are about fixing php8 errors/deprecations. Not much other fixes. If you need php74, you only/mostly miss some updates for php8.

kiatng:
Ref the medium article on PHP and release versioning. How about our own house rules:
image

sreichel:
> How about our own house rules:

In all that years i cant remember to have agreed to any of that RFCs! Imho .... thats something for larger projects.

sreichel:
Closed due to concerns.

colinmollenhour:
@sven, the Medium article you referenced makes a good case for it to be MINOR, one that I didn't think of before.. The fact that the composer package *can* prevent a user with the wrong PHP from upgrading is interesting - although in some cases I have used `--ignore-platform-reqs` so this would get me in trouble..

The purpose of the RFC is to make it clear to users what version specification it is safe for them to use given their updating practices.

Given that we have ceased to support non-composer installations (we no longer build a zip file), I could see taking this approach if everyone agrees to allow PHP requirements to change in a MINOR version so long as they are specified correctly in the composer.json file.

This would go against the Release Schedule guide, but if everyone wants it that way and there are no strong arguments against it, I would support this change to the release schedule document so it becomes official.

>  i cant remember to have agreed to any of that RFCs

There were 11 votes Yea and 0 Nea after months of discussion, so not sure what to say to that... https://github.com/OpenMage/rfcs/pull/10

Those are the rules currently, but it doesn't mean the rules can't change, just gotta make a proposal and get everyone onboard.

sreichel:
> > i cant remember to have agreed to any of that RFCs
> 
> There were 11 votes Yea and 0 Nea after months of discussion, so not sure what to say to that... [OpenMage/rfcs#10](https://github.com/OpenMage/rfcs/pull/10)

I only said that i not agreed/participated. For me it is too much for a small project like this.

colinmollenhour:
Looks like OpenMage is broken on PHP 8.0 so it seems already in practice the minimum version is 8.1... 🤣 

```
PHP Parse error:  syntax error, unexpected token ")" in .../vendor/symfony/css-selector/XPath/Extension/NodeExtension.php on line 63
```

Given we have a currently "broken" build and [the strong arguments](https://github.com/OpenMage/magento-lts/pull/4124#issuecomment-2664139672) for allowing PHP engine requirements to be a MINOR version change instead of MAJOR, I propose amending the Release Schedule and merging this PR. I'll follow up with a PR for the release schedule.

colinmollenhour (approved):
Approving since it is already not compatible with PHP 8.0 due to some previous update.

colinmollenhour:
Added PR to amend the release schedule document: https://github.com/OpenMage/rfcs/pull/12

sreichel:
> looks like OpenMage is broken on PHP 8.0 so it seems already in practice the minimum version is 8.1... 🤣

This should not happen, Can you confirm your `symfony/css-selector` is on v5.4.45?

kiatng:
I have one of the production servers on PHP7.4 running OM20.10.2. The composer lock entry:
```json
            "require": {
                "ext-dom": "*",
                "ext-libxml": "*",
                "php": "~7.3.0 || ~7.4.0 || ~8.0.0 || ~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0",
                "sabberworm/php-css-parser": "^8.7.0",
                "symfony/css-selector": "^4.4.23 || ^5.4.0 || ^6.0.0 || ^7.0.0"
            },
```
I have not encountered any issue.

sreichel:
composer.lock should have this

```json
            "name": "symfony/css-selector",
            "version": "v5.4.45",
            "source": {
                "type": "git",
                "url": "https://github.com/symfony/css-selector.git",
                "reference": "4f7f3c35fba88146b56d0025d20ace3f3901f097"
            },
```

colinmollenhour:
Ahh, maybe it's my mistake, then... I was trying to update the demo server and [ran composer update](https://github.com/OpenMage/OpenMage-demo-deployment/commit/88773ffac8fac0b3d95775a4a6d1d3a6b5310729), but maybe I ran it locally on a later PHP version than is used by the Nexcess server. Daniel updated the Nexcess config to use PHP 8.3 since then, so it's working now, but sounds like it would've been fine on PHP 8.0 had I run the composer update through Github actions. Oops...

addison74:
I would go a little further by setting the minimum PHP version to 8.2 in 3 -4 months from now. I have been using this version for a long time in a test environment and all the problems that arose I reported and they were resolved.

colinmollenhour:
Yes, at this point 8.1 is EOL in just 9 months.. although if we're only doing minor versions for PHP upgrades, then it won't be so hard to update again later.

colinmollenhour:
There are currently several errors preventing the checks from passing...

sreichel:
Yes, that why i set  i to draft. Will be fixed asap.

sreichel:
> I would go a little further by setting the minimum PHP version to 8.2 in 3 -4 months from now. I have been using this version for a long time in a test environment and all the problems that arose I reported and they were resolved.

Yep. I'd start with 8.1 here and in 2-3-4 months/releases bump version to 8.2.

addison74 (dismissed):

addison74 (approved):

colinmollenhour (approved):
Nice work @sreichel !

addison74 (approved):

PR #4147, opened 2024-08-08
PHPCS: ignore `mcrypt` related warnings

sreichel:
See https://github.com/OpenMage/magento-lts/actions/runs/10293142342/job/28488897656

Ignore all `mcrypt` related warnings b/c we use https://github.com/phpseclib/phpseclib.

Todo: (another PR)

Fix all other errors an do NOT continue on failure.


kiatng (approved):

ADDISON74 (approved):

mattdavenport:
Is there a case to be made here to keep the mcrypt code at all?

sreichel:
It may be used by old extensions?

mattdavenport:
No hard objections here - just wondering if we keep this code around forever or make a plan to remove at some point. Initially I had thought other projects were removing it completely, but upon second look they are keeping it around as well (e.g. [Joomla](https://issues.joomla.org/tracker/joomla-cms/38527)) for legacy support. LGTM.

mattdavenport (approved):

PR #4152, opened 2024-08-12
Rector: CQ - SimplifyIfReturnBoolRector

sreichel:
See:
- https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#simplifyifreturnboolrector

Rector; 
- added config file
- added to workflow


app/code/core/Mage/Adminhtml/Block/Notification/Toolbar.php
```diff
@@ -53,13 +53,8 @@ public function isShow()
...
         }
-        if ($this->getCriticalCount() == 0 && $this->getMajorCount() == 0 && $this->getMinorCount() == 0
-            && $this->getNoticeCount() == 0
-        ) {
-            return false;
-        }
-
-        return true;
+        return !($this->getCriticalCount() == 0 && $this->getMajorCount() == 0 && $this->getMinorCount() == 0
+            && $this->getNoticeCount() == 0);
```
kiatng, line 57: We can simplify with DeMorgan's Law `!(A && B) == (!A || !B)`:
```suggestion
        return $this->getCriticalCount() || $this->getMajorCount() || $this->getMinorCount() || $this->getNoticeCount();
```

sreichel: Step by step ... :)

https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#simplifydemorganbinaryrector

app/code/core/Mage/Adminhtml/Block/Sales/Items/Abstract.php
```diff
@@ -562,10 +562,7 @@ public function canShipPartially($order = null)
             $order = Mage::registry('current_shipment')->getOrder();
         }
         $value = $order->getCanShipPartially();
-        if (!is_null($value) && !$value) {
-            return false;
-        }
-        return true;
+        return !(!is_null($value) && !$value);
```
kiatng, line 565: DeMorgan's Law `!(A && B) == (!A || !B)`:

```suggestion
        return is_null($value) || $value;
```

kiatng, line 565: DeMorgan's Law `!(A && B) == (!A || !B)`:

```suggestion
        return is_null($value) || $value;
```

app/code/core/Mage/Adminhtml/Block/Sales/Items/Abstract.php
```diff
@@ -580,17 +577,11 @@ public function canShipPartiallyItem($order = null)
             $order = Mage::registry('current_shipment')->getOrder();
         }
         $value = $order->getCanShipPartiallyItem();
-        if (!is_null($value) && !$value) {
-            return false;
-        }
-        return true;
+        return !(!is_null($value) && !$value);
```
kiatng, line 580: DeMorgan's Law `!(A && B) == (!A || !B)`:

```suggestion
        return is_null($value) || $value;
```

app/code/core/Mage/Adminhtml/Block/Sales/Items/Abstract.php
```diff
@@ -580,17 +577,11 @@ public function canShipPartiallyItem($order = null)
...
     }
 
     public function isShipmentRegular()
     {
-        if (!$this->canShipPartiallyItem() || !$this->canShipPartially()) {
-            return false;
-        }
-        return true;
+        return !(!$this->canShipPartiallyItem() || !$this->canShipPartially());
```
kiatng, line 585: Apply DeMorgan's Law: `!(A || B) == (!A && !B)`
```suggestion
        return $this->canShipPartiallyItem()  && $this->canShipPartially();
```

app/code/core/Mage/Adminhtml/Block/Sales/Order/Invoice/Create/Form.php
```diff
@@ -102,10 +102,7 @@ public function hasInvoiceShipmentTypeMismatch()
     public function canShipPartiallyItem()
     {
         $value = $this->getOrder()->getCanShipPartiallyItem();
-        if (!is_null($value) && !$value) {
-            return false;
-        }
-        return true;
+        return !(!is_null($value) && !$value);
```
kiatng, line 105: DeMorgan's Law `!(A && B) == (!A || !B)`:

```suggestion
        return is_null($value) || $value;
```

app/code/core/Mage/Adminhtml/Controller/Action.php
```diff
@@ -395,13 +395,8 @@ protected function _validateSecretKey()
...
         }
-
-        if (!($secretKey = $this->getRequest()->getParam(Mage_Adminhtml_Model_Url::SECRET_KEY_PARAM_NAME, null))
-            || !hash_equals(Mage::getSingleton('adminhtml/url')->getSecretKey(), $secretKey)
-        ) {
-            return false;
-        }
-        return true;
+        return !(!($secretKey = $this->getRequest()->getParam(Mage_Adminhtml_Model_Url::SECRET_KEY_PARAM_NAME, null))
+            || !hash_equals(Mage::getSingleton('adminhtml/url')->getSecretKey(), $secretKey));
```
kiatng:
```suggestion
        return ($secretKey = $this->getRequest()->getParam(Mage_Adminhtml_Model_Url::SECRET_KEY_PARAM_NAME, null))
            && hash_equals(Mage::getSingleton('adminhtml/url')->getSecretKey(), $secretKey);
```

app/code/core/Mage/Adminhtml/Block/Notification/Toolbar.php
kiatng: Oh...so rector rules cannot be chained?

sreichel: With rulessets (chained rules) the PRs would grow to large. Mhhh.

app/code/core/Mage/Adminhtml/Controller/Action.php
kiatng: Continue on the conversation on chaining rector rules, I don't like the idea that it requires another PR to clean up the code above. This example from rector is really bad. It should be clean up before committing. Don't you think?

app/code/core/Mage/Adminhtml/Block/Notification/Toolbar.php
sreichel: Mhhh .... https://getrector.com/demo/ec03b365-db13-4e75-8ecd-d3773ae96870

kiatng: I created an issue: https://github.com/rectorphp/rector/issues/8781

app/code/core/Mage/Adminhtml/Controller/Action.php
sreichel: I'd prefer to revert last commit and add the DeMorgans rule separately.

Let's keep it simple. It only affects a few files.

Maybe add a new issue to not forget about it?

kiatng: I tried to apply the DeMorgan's rule to `!(!is_null($value) && !$value)`, not sure if I made the test correctly here: https://getrector.com/demo/6a07c84e-3a20-45c6-9513-d20329994eb8

Did I make a mistake in the test? It doesn't work, no code change.  If the above test is correct and if this PR is committed, there is no guarantee the convoluted code can be cleaned up by the rule later. So, now what?

kiatng: OK, so rector won't fix the convoluted bool expression now:
image

So, should we apply DeMorgan's manually?

sreichel: https://github.com/OpenMage/magento-lts/pull/4152/commits/392d0a88c2a8c68854106c79288e727fa2c1660c

No. I'd take it as it is and apply that rule later. I dont think its worth to change it manually, same pattern is already there.

sreichel:
Issues fixed in #4204

sreichel:
@kiatng can we continue w/o DeMorgan's Law?

kiatng:
I cannot give my approval because in my opinion, some changes are worse than present.

sreichel:
B/c of 6 out of +100 improvements?

Dont get me wrong, i agree with your suggestions, but i dont want to mix it up with automatic fixes.

How about to merge and verify manual changes again?

sreichel:
... at the other hand ... we can leave it open. With rector added as dependency we can skip it for now.

kiatng:
Or submit a PR to improve Rector (which I cannot do).

sreichel:
I'd really like to do, but i am not so familiar with rectors internal functionality.

I think i can write simple custom rules, but in this case ... there are a lot of possibly combinations of "&&", "||" that makes it difficult to make it safe for a general rule. I guess thats the reason why your issue has been closed.

fballiano:
this PR makes the code much harder to read, the benefit is only for the machine, not for the humans.

sreichel:
Disagree. Most of the changes look better to me.

- it safes a lot of lines
- how often do you look at core code?

sreichel:
It's a negated true/fails expression?

So what?

fballiano:
because I'm stupid I think that a negated OR is far less readable. but better make the code 1 line less but much harder to read so that only geniuses will take the time to read it.

again, the code is not for the machines, it's for the people, it should be understandable easily for the people.

only because rector has that rule it doesn't mean it should be applied.

sreichel:
I did not say you're stupid ... you're not.

I'd like to use rector to unify code.

For core, for extensions ...

kiatng:
I need to call out @sreichel comment:

> how often do you look at core code?

Even though it is a simple question, it is irrelevant. The question carries with it negative implications, which are not within standards of our [CoC](https://github.com/OpenMage/magento-lts?tab=coc-ov-file):

> ... we as contributors and maintainers pledge to making participation in our project and our community a harassment-free experience for everyone, regardless of ... level of experience, education, ...

and in particular the second standard:

> Being respectful of differing viewpoints and experiences

Note that our CoC was based on v1.4. The latest version is [2.1](https://www.contributor-covenant.org/version/2/1/code_of_conduct/). I encourage everyone here take the time to read it and think about it. The second standard in this version includes "opinions":

> Being respectful of differing opinions, viewpoints, and experiences

I urge all of us use language, English as well as PHP, that complies to the best standards.

sreichel:
There was no "negative implications"!

However ...

sreichel:
> I need to call out @sreichel comment:
> 
> > how often do you look at core code?

Maybe you got me wrong?

This PR would have added 7 negated return statements.

My question was how often you face these seven particular lines during development,

When you serach the codebase for `return !` you'll have +700 matches, so it is a very common pattern.  Decline to add 7 more b/c of "readablility" sound bit weired.

PR #4155, opened 2024-08-15
Fix strlen() null deprecation in AccountController.php

kiatng:
### Description (*)
```
    [type] => 8192:E_DEPRECATED
    [message] => strlen(): Passing null to parameter #1 ($string) of type string is deprecated
    [file] => .../app/code/core/Mage/Customer/controllers/AccountController.php
    [line] => 900
    [uri] => /customer/account/resetpassword/
```


app/code/core/Mage/Customer/controllers/AccountController.php
sreichel, line 1: Would it be better to use empty string instead of false?

With strict types passing bool to strlen results in a fatal error.

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -895,7 +895,7 @@ public function resetPasswordPostAction()
      */
     protected function getCustomerId()
     {
-        $customerId = $this->getRequest()->getQuery("id");
+        $customerId = $this->getRequest()->getQuery('id', false);
```
sreichel, line 898:
```suggestion
        $customerId = $this->getRequest()->getQuery('id', '');
```

Would it be better to use empty string instead of false?

With strict types passing bool to strlen results in a fatal error.

kiatng: Lines 904 - 906:
```php
            $customerId = count($customerCollection) === 1
                ? $customerCollection->getFirstItem()->getId()
                : false;
```
So, the default `false` is the expected behavior. If we modify this to empty string, is there a BC concern?

sreichel: Before you change default was `null`, now it 'string' ...

```
    /**
     * @return string|null
     * @throws Mage_Core_Exception
     */
    protected function getCustomerId()
    {
        $customerId = $this->getRequest()->getQuery('id');
        if (!is_null($customerId) && strlen($customerId) > 12) {
            $customerCollection = Mage::getModel('customer/customer')
                ->getCollection()
                ->addAttributeToSelect(['rp_customer_id'])
                ->addFieldToFilter('rp_customer_id', $customerId);
            $customerId = count($customerCollection) === 1
                ? $customerCollection->getFirstItem()->getId()
                : null;
        }

        return $customerId;
    }
```

kiatng: Now you removed `false` as a possible return value.

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -891,12 +891,12 @@ public function resetPasswordPostAction()
...
 
     /**
-     * @return string|false
+     * @return string|false|null
      */
     protected function getCustomerId()
     {
-        $customerId = $this->getRequest()->getQuery('id', false);
-        if (strlen($customerId) > 12) {
+        $customerId = $this->getRequest()->getQuery("id");
```
sreichel, line 898:
```suggestion
        $customerId = $this->getRequest()->getQuery('id');
```

sreichel (approved):

ADDISON74 (approved):

mattdavenport (dismissed):

app/code/core/Mage/Customer/controllers/AccountController.php
sreichel: I wanted to avoid antother return type ...  better use false as default and replace `is_null` check with `is_string`?

sreichel (approved):

ADDISON74 (approved):

PR #4159, opened 2024-08-21
Added rector as dev dev dependency

sreichel:
k/t


sreichel:
Wrong target ... closed.

PR #4160, opened 2024-08-21
Small DocBlock update

sreichel:
... detected with strict types on.


app/code/core/Mage/Adminhtml/Helper/Help/Mapping.php
```diff
@@ -231,7 +231,7 @@ protected function findInMapping($frontModule, $controllerName, $actionName)
     /**
      * Determine which version of docs should target onto
      *
-     * @return string
+     * @return false|Varien_Simplexml_Element
```
kiatng, line 234:
```suggestion
     * @return Mage_Core_Model_Config_Element|Varien_Simplexml_Element|false
```

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -524,7 +524,7 @@ public function getSessionHosts()
     /**
      * Retrieve session save method
      *
-     * @return string
+     * @return string|Varien_Simplexml_Element
```
kiatng, line 527:
```suggestion
     * @return Mage_Core_Model_Config_Element|Varien_Simplexml_Element|false|string
```

app/code/core/Mage/Core/Model/Session/Abstract.php
```diff
@@ -537,7 +537,7 @@ public function getSessionSaveMethod()
     /**
      * Get session save path
      *
-     * @return string
+     * @return string|Varien_Simplexml_Element
```
kiatng, line 540:
```suggestion
     * @return Mage_Core_Model_Config_Element|Varien_Simplexml_Element|false|string
```

app/code/core/Mage/Adminhtml/Helper/Help/Mapping.php
sreichel: Thought about it ....

I thinks with strict types on phpstan needs a class that implements `__toString()`.

`Mage_Core_Model_Config_Element` extends `Varien_Simplexml_Element` so i used this.

To narrow it down, how about `Mage_Core_Model_Config_Element|false`?

(but yes, add all types from parent method is also okay)

kiatng (approved):

ADDISON74 (approved):

PR #4138, opened 2024-08-02
PhpUnit: use database for tests, added tests

sreichel:
## Description

- [x] **install Openmage database for tests**

### GitHub workflow updates

- [x] merged phpunit and sonar, to not run tests twice
- [x] added unit-tests to php-cs-fixer/phpcs/phpstan checks

### UnitTests

- [x] added tests from [testfield](https://github.com/OpenMage/Testfield)
- [x] updated existing unit-tests
- [x] moved unit-tests to root directory
- [x] added composer scripts
- [x] **added ~150 tests**

---

### Tests

```shell
composer run phpunit:test
```

or with covarge:

```shell
composer run phpunit:coverage
```

and with local html covarge: (view build/coverage/index.html)

```shell
composer run phpunit:coverage-local
```

PR #4163, opened 2024-08-24
Add new method to get config value directly from DB bypassing cache.

kiatng:
### Description (*)
We can save a config value by
```php
Mage::getConfig()->saveConfig($path, $newValue);
```
However, if cache is enabled, after saving `$newValue`, the following will return the old value:
```php
Mage::getStoreConfig($path);
```

To get the new value, we need to refresh the cache with `Mage::getConfig()->reinit();`. 

In my use case, I needed to get and update config value in a cron running constantly.  It's not good idea to keep refreshing cache. (Also, I need to allow admin users to edit the value in the System Configuration. It turns out that values here are updated and not from the cache.)

With this PR, we can get the current value by
```php
Mage::getResourceSingleton('core/config')->getValue($path);
```

Over the years, others need similar feature, see [stackoverflow](https://stackoverflow.com/questions/14161149/magento-retrieving-uncached-system-configuration-values)

```php
$path = 'test/config';
Mage::getConfig()->saveConfig($path, 'foo');
echo Mage::getStoreConfig($path); // null
echo Mage::getResourceSingleton('core/config')->getValue($path); // foo
```


sreichel:
Can you please why why its not possible to create a PR at your repo?

- changed the method name (to follow naming pattern)
- added method in Mage_Core_Model_Config
- added UnitTest

See: https://github.com/sreichel/magento-lts/commit/136118a7fddbd9018c02437a49ce2c0703860f93.diff

sreichel:
Re-open please!

sreichel:
Re-open.

sreichel:
Re-open.

sreichel:
UnitTest requires a DB ... added in #4138

sreichel (approved):

colinmollenhour:
> In my use case, I needed to get and update config value in a cron running constantly.

It's up to you, but in my opinion, this is a pattern that should never be used since the config cache should only be updated by users or on deployment. I usually use core/flag instead, very convenient and fast and no need to touch cache. Just a thought.

sreichel:
Core config is not the right place for frequently changing values, but it "common" practice. Having a method to bypass cache is a +1.

I think most dont know about core/flag .... maybe worth a write-up?

kiatng:
I am aware of `core/flag` and use it. 

Some errors can only be captured with `error_get_last()`.
```php
// index.php, before Mage::run($mageRunCode, $mageRunType);
register_shutdown_function(function(){
    $err = error_get_last();
    if ($err && $err['type'] != E_WARNING) {
        $err['type'] = $err['type'] . ':' . array_search($err['type'], get_defined_constants(true)['Core']);
        $err['uri'] = $_SERVER['REQUEST_URI'] ?? $_SERVER['SCRIPT_NAME'];
        [$err['user'], $err['role']] = Mage::helper('base')->getSessionUser();
        Mage::getModel('core/flag', ['flag_code' => 'error_get_last'])
            ->loadSelf()
            ->setFlagData($err)
            ->save();
    }
});
```
And get the last error:
```php
$flag = Mage::getModel('core/flag', ['flag_code' => 'error_get_last'])->loadSelf();
```

But in my use case, the admin users need to view and edit the value. So I can't use `core/flag`.

sreichel:
> > In my use case, I needed to get and update config value in a cron running constantly.
> 
> It's up to you, but in my opinion, this is a pattern that should never be used since the config cache should only be updated by users or on deployment. I usually use core/flag instead, very convenient and fast and no need to touch cache. Just a thought.

I think there plenty of real use-cases for this ... e.g update/(re-)set a timestamp from backend, Set a value thats used in a cron-job ...

IMHO theres nothing wrong to provide this posibility.

ADDISON74 (approved):

PR #4164, opened 2024-08-24
PHPCS: fixed last PHPCompatibility check errors

sreichel:
- fixed errors (https://github.com/OpenMage/magento-lts/actions/runs/10537664925/job/29199341410)
- updated workflow to fail on error

### DIFF
--- a/.github/workflows/phpcs.yml
+++ b/.github/workflows/phpcs.yml
@@ -45,4 +45,4 @@ jobs:
 
       - name: PHPCodeSniffer
         run: php vendor/bin/phpcs -s --report=full --standard=${{ matrix.rules.path }}
-        continue-on-error: ${{ matrix.rules.path != '.phpcs.xml.dist' }}
+        continue-on-error: ${{ matrix.rules.path == '.phpcs.ecg.xml.dist' }}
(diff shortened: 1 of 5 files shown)


ADDISON74 (approved):

kiatng:
PHPCompatibility complaint is

> Since PHP 7.0, functions inspecting arguments, like func_get_args(), no longer report the original value as passed to a parameter, but will instead provide the current value. The parameter "$var" was used, and possibly changed (by reference), on line n.

But in the cases listed, it's all false, because the variables are not changed. So, isn't it a problem with PHPCompatibility and not the code?

Other fix is to add comment (see https://github.com/PHPCompatibility/PHPCompatibility/issues/1240)

```php
// phpcs:ignore PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.NeedsInspection -- Reason: value is unchanged.
```

sreichel:
> * Always use a call to any of the following functions - `func_get_arg()`, `func_get_args()` - at the top of a function before the rest of the logic.
>   This will make sure that the passed parameters have not been changed yet before they are captured and will prevent the error/warning altogether,
> * ...
> * If all else fails: PHPCS allows for the use of inline ignore annotations which when used correctly, can silence the error/warning from this sniff alone.

Followed this.

kiatng (approved):

ADDISON74 (approved):

PR #4165, opened 2024-08-24
Removed gitpod setup

sreichel:
- seems its not used at all (no questions, no issues ...)
- it was never updated and still uses unsupported php7.3

Get rid of it. Or are there any reasons to keep it?


kiatng (approved):

ADDISON74 (approved):

sreichel:
See https://github.com/OpenMage/magento-lts/pull/4175

PR #945, opened 2020-05-04
Get functional test suite working on GitHub Actions

colinmollenhour:
Goal: get the [Magento Test Framework](https://github.com/OpenMage/mtf) tests working via GitHub Actions so we actually have a robust functional test suite.

PR #4166, opened 2024-08-24
Removed docker setup

sreichel:
- there is DDEV that supports OpenMage
- there is https://github.com/OpenMage/docker-php .... merge it?

Dow we really need it? Isnt it better to simply use DDEV, that seems to be more flexible - and maintain the docker repo instead? Do we need it here in the main repo?


colinmollenhour:
I don't think it should be removed. Reasons:
- Not everyone uses ddev - it is the most robust option but I don't think it is ubiquitous enough to make it basically required.
- It provides a good blueprint for a production deployment, ddev doesn't do this at all
- If we are going to run things like GitPod, Codespaces, Devcontainer, Cypress, etc. we need a basic working container anyway, I don't think ddev is suitable for those
- It could be moved to a separate repo, but having a single command that launches a full environment in the repo itself is I think a valuable feature - for example if a person wants to make or try out a tiny patch without affecting their main environment and not set up ddev they can do it in seconds with this.

It's currently using images from https://github.com/colinmollenhour/docker-openmage/ but I would be fine consolidating these two somehow. When tinkering with FrankenPHP I learned of https://github.com/mlocati/docker-php-extension-installer which can greatly simplify maintaining custom PHP docker images so it really isn't that hard.

sreichel:
> * If we are going to run things like GitPod, Codespaces, Devcontainer, Cypress, etc. we need a basic working container anyway, I don't think ddev is suitable for those

You're right. For Gibtpod and some others we need the docker container, but for Cypress ... https://github.com/tyler36/ddev-cypress

colinmollenhour:
> You're right. For Gibtpod and some others we need the docker container, but for Cypress ... https://github.com/tyler36/ddev-cypress

Very nice. I was going to say if we use Cypress we will also want to run it in Github Actions which I think ddev is not well-suited.. but then I Googled and would have been proven wrong. :D https://github.com/ddev/github-action-setup-ddev

sreichel:
From google translate ... DDEV is a "Egg-laying jack of all trades" for development ;)

https://github.com/OpenMage/magento-lts/pull/945#issuecomment-2339510254 ? Let me know.

PR #4214, opened 2024-09-20
Fix wrong Preview link on CMS pages grid

Caprico85:
### Description (*)
If you have a CMS page where the identifier has 4 or more levels (e.g. `test/page/foo/bar`), the preview link on the CMS page grid will not work.

The preview link only shows the first 3 levels, where the URL key has 4 levels.

image

The preview link is generated in `Mage_Adminhtml_Block_Cms_Page_Grid_Renderer_Action`. This class uses `Mage::getModel('core/url')->getUrl(...)` to generate the link. `getUrl(...)` calls `Mage_Core_Model_Url::setReoutePath()`, which tries to parse the URL with the usual `module/controller/action` structure and cuts off the fourth level.

Using `getDirectUrl(...)` instead of `getUrl(...)` fixes this as `getDirectUrl(...)` takes the URL as is, not cutting off the fourth level.

1. Create a new CMS page. Set its identifier to `test/page/foo/bar` (or any other URL at least 4 levels deep).
2. Click the preview link on the CMS page grid.
3. You will get to the 404 page.
4. Check the URL of the link. It will be 'test/page/foo', not 'test/page/foo/bar' as specified in the identifier.

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Cms/Page/Grid/Renderer/Action.php
+++ b/app/code/core/Mage/Adminhtml/Block/Cms/Page/Grid/Renderer/Action.php
@@ -26,7 +26,7 @@ class Mage_Adminhtml_Block_Cms_Page_Grid_Renderer_Action extends Mage_Adminhtml_
             $href = $row->getPreviewUrl();
         } else {
             $urlModel = Mage::getModel('core/url')->setStore($row->getData('_first_store_id'));
-            $href = $urlModel->getUrl(
+            $href = $urlModel->getDirectUrl(
                 $row->getIdentifier(),
                 [
                     '_current' => false,


ADDISON74:
I confirm this bug with the latest OpenMage version + Magento Sample Data.

The **About Us** page has the URL key "about-magento-demo-store". Here are different scenarios increasing the path with one level

| URL Key  | Preview |
| about-magento-demo-store/1  | OK  |
| about-magento-demo-store/1/2  | OK  |
| about-magento-demo-store/1/2/3  | Wrong  |

For the last scenario, instead of being displayed in the status bar as "about-magento-demo-store/1/2/3" it is displayed as "about-magento-demo-store/1/2", 3 is missing. By increasing the levels number, the URL will remain the same limited to two levels.

ADDISON74 (approved):
Tested and it works as expected. Screenshot attached bellow for a 10-level URL.

screenshot3

sreichel (approved):
Did some tests too ...

PR #4229, opened 2024-09-26
Made swatches color-picker configurable

sreichel:
### Description (*)

... as suggested in related pull request.

- added new multi-select to config (see Catalog -> Configurable Swatches)
- added install script to not change anything for current installations

### Related Pull Requests

1. See OpenMage/magento-lts#3686


sreichel:
From my POV the current implementation is not complete. E.g. I dont need or want a color picker for "Size" attributes.  That makes no sense to me.

Further ... with an additional configuration we could add a file-upload for selected swatchable attributes.

This would be pretty cool for me.

This PR does not change the current behavoir or breaks anything. It only allows a more precise configuration.

app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -99,10 +100,26 @@ public function getSwatchAttributeIds()
...
+     * Get list of attributes that should use swatches
+     *
+     * @return array
+     */
+    public function getSwatchAttributeIdsUsesColorPicker()
+    {
+        if (is_null($this->_configAttributeIds)) {
+            $this->_configAttributeIds = [];
+            if (Mage::getStoreConfig(self::CONFIG_PATH_SWATCH_ATTRIBUTES_COLORPICKER)) {
+                $this->_configAttributeIds = explode(',', Mage::getStoreConfig(self::CONFIG_PATH_SWATCH_ATTRIBUTES));
```
sreichel, line 113:
```suggestion
                $this->_configAttributeIds = explode(',', Mage::getStoreConfig(self::CONFIG_PATH_SWATCH_ATTRIBUTES_COLORPICKER));
```

empiricompany:
For me it remains very cumbersome and not very intuitive but if we want it, must be explained well in the documentation that you need to configure the attribute as Swatch + configure it as a color picker to activate it.

+1 for implent image upload

sreichel:
> must be explained well in the documentation

To avoid this the install script was added ... by default every swatchable attribute will show the color-picker. So nothing changes. You have to manually de-select the attributes to not show the color-picker.

empiricompany:
this PR add a new configuration to enable the color picker for specific attribute not to disable it, so attributes selected as Swatch not show the color picker by default. correct?

sreichel:
No.

This install-script was added to enable the color-picker for all swatch-attributes. **You have manually to de-select them** to disable the color picker. (thats why i said i does not change anything for existing installations)

ADDISON74:
I have to read again the comments I had on #3686 to understand the change, then I'll give you feedback. empiricompany's initial proposal was not bad, but being a complex change in the code, there were no fresh forces to carry things through to the end. Anyway the effort made is appreciated.

sreichel:
For sample data config looks like this:

confsw

app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -99,10 +103,24 @@ public function getSwatchAttributeIds()
         return $this->_configAttributeIds;
     }
 
+    /**
+     * Get list of attributes that should use swatches
+     */
+    public function getSwatchAttributeIdsUsesColorPicker(): ?array
```
sreichel, line 109:
```suggestion
    public function getSwatchAttributeIdsUsesColorPicker(): array
```

app/code/core/Mage/ConfigurableSwatches/Model/System/Config/Source/Catalog/Product/Configattribute/Colorpicker.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * Attributes array
+     */
+    protected ?array $_attributes = null;
+
+    /**
+     * Retrieve attributes as array
+     */
+    public function toOptionArray(): ?array
```
sreichel, line 31:
```suggestion
    public function toOptionArray(): array
```

app/code/core/Mage/ConfigurableSwatches/Helper/Data.php
```diff
@@ -22,13 +22,17 @@ class Mage_ConfigurableSwatches_Helper_Data extends Mage_Core_Helper_Abstract
...
+    public const CONFIG_PATH_SWATCH_ATTRIBUTES_COLORPICKER = 'configswatches/general/swatch_attributes_colorpicker';
     public const CONFIG_PATH_LIST_SWATCH_ATTRIBUTE = 'configswatches/general/product_list_attribute';
 
     protected $_moduleName = 'Mage_ConfigurableSwatches';
 
     protected $_enabled = null;
+
     protected $_configAttributeIds = null;
 
+    protected ?array$_configAttributeIdsUsesColorPicker = null;
```
sreichel, line 34:
```suggestion
    protected ?array $_configAttributeIdsUsesColorPicker = null;
```

empiricompany:
please check this behavior: 
If there is no attribute selected as swatch, color picker option is empty.
I have to save the attribute as swatch then re-save for enable colorpicker

sreichel:
Thanks. I know save config twice is not ideally. I have an idea how to fix it, but I am not sure if it worth to spent time on it - as it probably not often changes.

Add JS ...

- to add/remove options in second selection when de/select a swatch attribute.

ADDISON74:
This PR works and thank you for reading my comments in the PR #3686.  There I tried to convince the participants without success.

Indeed, two simultaneous saves are required to set the configuration of an attribute.

However, I consider that the OpenMage implementation proposed in #3686 should have been inspired by Magento 2, but for that I will propose a separate discussion.

empiricompany:
in my opinion 2 separate saves is not intuitive, for this reason I remain of the idea that a new configuration is not needed but that simply, in the case of an attribute that does not use colors like size, you don't select any color and upload the individual images as default.

sreichel:
@empiricompany if we want to add a file-upload option, we have to select which attribute uses color-picker or upload.

I'd like to have it in ... maybe you have an idea for some javascript?

PR #4232, opened 2024-09-29
Admin config: added input validation & hide disabled options

sreichel:
Do not render options for disabled sections (e.g. disabled payment methods).

Added some input-validation to enforce number/digit imput.

### DIFF
--- a/app/code/core/Mage/Api/etc/system.xml
+++ b/app/code/core/Mage/Api/etc/system.xml
@@ -44,6 +44,7 @@
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
                             <show_in_store>1</show_in_store>
+                            <validate>required-entry validate-digits validate-zero-or-greater</validate>
                         </session_timeout>
                         <compliance_wsi translate="label">
                             <label>WS-I Compliance</label>
--- a/app/code/core/Mage/Authorizenet/etc/system.xml
+++ b/app/code/core/Mage/Authorizenet/etc/system.xml
@@ -42,6 +42,7 @@
                             <show_in_default>1</show_in_default>
                             <show_in_website>1</show_in_website>
                             <show_in_store>0</show_in_store>
+                            <depends><active>1</active></depends>
                         </payment_action>
                         <title translate="label">
                             <label>Title</label>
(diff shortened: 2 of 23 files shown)


empiricompany:
please provide a description including the reason of this PR

ADDISON74:
Nice work. This PR makes some changes in the **Backend > System > Configuration**. for example in **Google Sitemap > CMS Pages Options** section. If Frequency = "Do not show in sitemap" Priority and Show lastmod are displayed, although they should not be displayed.

**Before**

before

**After**

after

Similarly in the same section for **Generation Settings**, if it is not enabled, its 5 options are no longer displayed.

I will check all the sections so that nothing is omitted and then I will approve it.

ADDISON74:
The following sections need an evaluation:

**Catalog > Catalog > Product Alerts**
Do we need to display **Price Alert Email Template** and **Stock Alert Email Template**  dropdowns if both **Allow Alert When Product Comes Back to ...** are No?

**Sales > Shipping Methods > UPS**
If **Enabled for checkout** is No there are 7 inputs which are still visible. This is not happening for USPS, FedEx, DHL.

sreichel:
> Do we need to display **Price Alert Email Template** and **Stock Alert Email Template** dropdowns if both **Allow Alert When Product Comes Back to ...** are No?

I have hidden the template selector. 

> Sales > Shipping Methods > UPS
If Enabled for checkout is No there are 7 inputs which are still visible. This is not happening for USPS, FedEx, DHL.

The `<depends>` node is there. I dont know where its coming from.

ADDISON74:
The first issue is solved. The second one needs a solution, all those 7 inputs depends on the first one.

image
=

ADDISON74:
Just delete a node inside <ups>, for example the first one access_license_number, then reload the page. By setting "Enable for Checkout to No", all inputs disappear. There is something strange in the <ups> node.

sreichel:
I think there is some javascript that disables the input for the 7 fields.

I'd leave it as it is.

sreichel:
Missed your last comment ... I have an idea. Will test it later.

sreichel:
Can confirm deleting "access_license_number" works.

Lets fix it another time. Create issue for it.

ADDISON74 (approved):

kiatng (approved):

PR #4236, opened 2024-09-30
Added more PhpUnit tests

sreichel:
### Description (*)

- added some tests

### Comments

To exclude slow running tests;

```bash
phpunit --exclude Base,runInSeparateProcess
```


ADDISON74 (approved):

app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Websites.php
```diff
@@ -42,7 +42,7 @@ public function getProduct()
     /**
      * Get store ID of current product
      *
-     * @return int
+     * @return string
```
kiatng, line 45: https://github.com/OpenMage/magento-lts/blob/df37dba44808847ccbb90a4ec8a53cb6065e832c/app/code/core/Mage/Catalog/Model/Product.php#L215-L215
```suggestion
     * @return int|string
```

sreichel: I think i have xdebuged it, and it is always a string.
 
 The method annotation should be changed. (?)

kiatng: Consider 
```php
$product = Mage::getModel('catalog/product')
       ->setStoreId($storeId)
       ->load($id);
``` 
OpenMage is not type strict: `$storeId` and `$id` can be type string or int. However, the column `store_id` is type `smallint(5)` in the EAV tables, so `int` is more correct in docblock. Therefore, I suggest `int` first then `string`. We can leave the annotation as is.

sreichel: > However, the column store_id is type smallint(5) in the EAV tables, so int is more correct in docblock.

Thats true. Mhhh ... if we want to reflect the DB column type we should turn it into "int" - everywhere.

sreichel: Reverted some changes and cast to int (in a few places).

ADDISON74 (dismissed):

ADDISON74 (dismissed):

app/code/core/Mage/Eav/Model/Entity/Attribute/Backend/Abstract.php
```diff
@@ -222,7 +222,7 @@ public function getDefaultValue()
      *
      * @param Varien_Object $object
      * @throws Mage_Eav_Exception
-     * @return bool
+     * @return $this|bool
```
kiatng, line 225: I cannot see where it is returning `$this`. Can you point it out?

sreichel: It is to fix this:

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   app/code/core/Mage/Weee/Model/Attribute/Backend/Weee/Tax.php                                                                                                                             
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  41     Return type ($this(Mage_Weee_Model_Attribute_Backend_Weee_Tax)) of method Mage_Weee_Model_Attribute_Backend_Weee_Tax::validate() should be compatible with return type (bool) of method  
         Mage_Eav_Model_Entity_Attribute_Backend_Abstract::validate()                                                                                                                             
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 

```

kiatng: Nasty.

sreichel: Yep. I dont like it, but i dont have a better idea?

Docs only reflect what we have to deal with ...

app/code/core/Mage/Adminhtml/Block/Catalog/Product/Edit/Tab/Websites.php
sreichel: @kiatng any objections to cast store id to int?

kiatng: No objection to cast store ID to int.

app/code/core/Mage/Catalog/Model/Category.php
```diff
@@ -314,7 +314,7 @@ public function move($parentId, $afterCategoryId)
     /**
      * Retrieve default attribute set id
      *
-     * @return int
+     * @return string
```
kiatng, line 317: Attribute set ID is stored as int in the DB. Is it more correct to return as int? I'm aware that reading int from DB converts to string. It's problematic. But in the spirit of strict typing, it should be int. Your thought?

sreichel: Yes. Same as for "store_id". Cast to int?

kiatng (dismissed):

app/code/core/Mage/Catalog/Model/Product.php
```diff
@@ -217,7 +217,7 @@
  * @method $this setStoreIds(array $storeIds)
  * @method array getSwatchPrices()
  *
- * @method int getTaxClassId()
+ * @method string getTaxClassId()
```
kiatng, line 220: int?

ADDISON74 (approved):

kiatng (approved):

fballiano:
this PR should be about unit tests and it's mainly about phpstan, but also, did you notice it introduces 3-4 new errors in phpstan?

sreichel:
to write the tests, i have to view the files. when phpstan issues can be fixed at the same time ....

yes, i did, fixing one phpstan error can reveal others, but this can become endless story to fix all follow-up errors. Better +5 and -150 for now? They will be fixed later.

PR #2979, opened 2023-01-19
Prevent delete/disable CMS pages used as default pages

sreichel:
### Description (*)

This does NOT fixes the problem in #382 directly. Disabled pages are still listed in dropdown, but it prevents to disable/delete them.

### Fixed Issues (if relevant)

1. Closes OpenMage/magento-lts#382

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #4239, opened 2024-10-02
Prevent delete/disable CMS pages used in configuration

sreichel:
### Reproduce

Try to disable/delete configured `no-route` page.

### Related Pull Requests

1. See OpenMage/magento-lts#2979


empiricompany:
Isn't it better to put it in the controller? 

Resource should not do these checks

sreichel:
There are different checks in resource, e.g. `getIsUniquePageToStores`.

Having it in model allows to use ... `Mage::getModel('cms/page')->load(1)->getUsedInStoreConfigCollection()`

ADDISON74:
Tested an it does what it says. Do you need to add anything to it?

sreichel:
Maybe a unit test for the helper method, but this can be done in another PR.

ADDISON74:
I am testing this PR. For the 404 Not Found page (no-route). 

1. If I delete it the message is "Cannot delete page, it is used in "web/default/cms_no_route".

2. If I disable it I am getting two messages "Cannot disable page, it is used in "web/default/cms_no_route" and "The page has been saved." This can be cofusing.

Another observation changing the URL Key will have negative impact. For example, from no-route to no-route-2.

sreichel:
Having two messages is intended.

The page gets saved, but resets "is active" option to yes.

I have not tested changing url-key ... what happens? (I am not at home to test it)

ADDISON74:
Changing the **URL key** of that page is similar to actions such as delete or disable, because once the URL is changed that page will no longer be used and therefore it will be a problem. My opinion is that it should not be allowed to change the URL as long as it is used.

url_key

sreichel:
Good idea.

I think there are other scenarios to cover ...

- having multiple pages with same identifier for different store
- having different pages configured for different stores

sreichel:
Better leave it for another PR.

ADDISON74:
> Better leave it for another PR.

@sreichel - I agree. What else I would propose for this PR would be a different message, for example

**You cannot delete or disable the page nor change the URL Key as long as it is used as a Default Page in the configuration.**

One message covers all the situations and makes the things clear. Like a general warning what you cannot do.

sreichel:
> One message covers all the situations

This will not work. There is a difference between them.

When deleting an exception is thrown (red message) and the action is aborted.

When disabling the "save action" continues, all other changes are applied, but active state is set back and a warning (yellow message) is left.

kiatng (approved):
Tested

ADDISON74 (approved):

PR #4240, opened 2024-10-02
Fix placeholder image in WYSIWYG editor when using a different admin theme

Caprico85:
### Description (*)

A very long description for a very specific use case 😄.

If you insert an image into your content, but later you delete the image from the disk, the WYSIWYG editor shows a placeholder image:

image

If you have set a different admin theme, a broken image will be shown instead:

image

Or worse, if no width and height is set in the generated  tag, nothing will be shown at all:

image

I would expect the placeholder image to always be shown, no matter which admin theme is used

---

This is caused as follows:

The WYSIWYG editor generates the HTML code like this:

```html
<p><img src="{{media url="wysiwyg/test.jpg"}}" alt=""></p>
```

To resolve the {{media}} tag, the editor sends a request to the server:

image

The request will be handled in Mage_Adminhtml_Cms_WysiwygController::directiveAction(). There, the placeholder image will be used if the original image cannot be found. The path to the placeholder image is generated like this:

```php
public function getSkinImagePlaceholderPath()
{
    return Mage::getModel('core/design_package')->getSkinBaseDir(['_area' => 'adminhtml']) . DS .
        self::WYSIWYG_SKIN_IMAGE_PLACEHOLDER_FILE;
}
```

This is the problem. It takes the skin directory **of the current theme** (`skin/adminhtml/default/mytheme/`) and concatenates the path to the placeholder image. This works fine for the `default` admin theme. But to correctly support other admin themes, is has to take fallbacks to the default theme into account. Using `getFilename()`, as in this PR, fixes that.

1. On a CMS page, product or anywhere else where you can use the WYSIWYG editor, add an image to the content and save.

image

2. Delete the newly uploaded image from the folder media/wysiwyg/.

3. Refresh the admin page. You will see the placeholder image as expected (as we have not yet set an admin theme)

image

4. Set a different admin theme by setting this in your app/etc/local.xml:

```xml
<config>
    <stores>
        <admin>
            <design>
                <theme>
                    <default>mytheme</default>
                </theme>
            </design>
        </admin>
    </stores>
</config>
```

You don't need to create any directories. All file requests will transparently fall back to the default theme (well, almost all).

5. Clear the config cache.

6. Reload the admin page where you added the image. You will see a broken image.

image

7. Apply my PR. The placeholder image is back.
 
(for extra fun, subscribe to exception mails in errors/local.xml and get email bombed whenever somebody edits a page with broken images)

### DIFF
--- a/app/code/core/Mage/Cms/Model/Wysiwyg/Config.php
+++ b/app/code/core/Mage/Cms/Model/Wysiwyg/Config.php
@@ -127,8 +127,7 @@ class Mage_Cms_Model_Wysiwyg_Config extends Varien_Object
      */
     public function getSkinImagePlaceholderPath()
     {
-        return Mage::getModel('core/design_package')->getSkinBaseDir(['_area' => 'adminhtml']) . DS .
-            self::WYSIWYG_SKIN_IMAGE_PLACEHOLDER_FILE;
+        return Mage::getDesign()->getFilename(self::WYSIWYG_SKIN_IMAGE_PLACEHOLDER_FILE, ['_type' => 'skin']);
     }
 
     /**


sreichel:
@Hanmac @theroch please re-check admin-template fall-back PR (we talked about). May it fixes this too.

empiricompany (approved):

ADDISON74 (approved):

PR #4247, opened 2024-10-04
Replaced `=== null` with `is_null()`

sreichel:
### Description (*)

Thats an old idea to unify this, but thanks to @tmewes (https://github.com/MahoCommerce/maho/pull/35) who brought it up.

Instead of changing to `=== null` i used `is_null()` instead.

Reason, confirmed in tests ...

- https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/4015#issuecomment-568508507

@tmewes there is no rector-rule, but a (risky) one for [PHP-CS-Fixer](https://cs.symfony.com/doc/rules/language_construct/is_null.html). _jtlyk_

Note: there are a few loose comparisons (`==` /`!=`) that i did not touch)

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -673,7 +673,7 @@ final class Mage
      */
     public static function app($code = '', $type = 'store', $options = [])
     {
-        if (self::$_app === null) {
+        if (is_null(self::$_app)) {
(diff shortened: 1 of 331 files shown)


sreichel:
Rector?

PR #4268, opened 2024-10-11
Added polyfill-8.3/8.4 in Composer

ADDISON74:
I have added polyfill-8.3 in the composer file.


sreichel:
We don't use these methods in core, but why not.

Please use composer require command and add updated composer.lock too.

sreichel:
@colinmollenhour tech question. How it is possible to create new branches here? (https://github.com/OpenMage/magento-lts/branches/all)

I thopught we disabled it times ago?

sreichel (approved):

kiatng (dismissed):

sreichel (approved):

kiatng (approved):

PR #4276, opened 2024-10-14
Some phpstan fixes

sreichel:
k/t

### DIFF
--- a/app/code/core/Mage/Index/Model/Event.php
+++ b/app/code/core/Mage/Index/Model/Event.php
@@ -83,7 +83,7 @@ class Mage_Index_Model_Event extends Mage_Core_Model_Abstract
     /**
      * Get related process object
      *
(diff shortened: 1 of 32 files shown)


app/code/core/Mage/Index/Model/Event.php
```diff
@@ -71,7 +71,7 @@ protected function _construct()
     /**
      * Specify process object
      *
-     * @param Mage_Index_Model_Process $process
+     * @param Mage_Index_Model_Process|null $process
```
kiatng, line 74: How about line 86? `* @return Mage_Index_Model_Process` to `* @return Mage_Index_Model_Process|null`?

sreichel: Thanks.

ADDISON74 (approved):

kiatng (approved):

Issue #4286, opened 2024-10-17
[PHP 8.4] Constant E_STRICT is deprecated


ADDISON74:
**[Test Environment]**
- Windows 11 + WSL
- Docker 4.34.4
- DDEV (Apache 2.4.62, PHP 8.4.ORC1, MariaDB 10.11.9) with MAGE_IS_DEVELOPER_MODE=1
- Magento Sample Data

I just upgraded DDEV to 1.2.5. Accessing the Frontend I am getting this error in the browser window

```php
Deprecated:  Constant E_STRICT is deprecated in /var/www/html/app/code/core/Mage/Core/functions.php on line 168

```

For this error more information here https://php.watch/versions/8.4/E_STRICT-deprecated#:~:text=The%20E_STRICT%20constant%20is%20deprecated,be%20removed%20in%20PHP%209.0.

PR #4289, opened 2024-10-18
PHP84: ignore deprecated E_STRICT

sreichel:
### Description (*)

https://php.watch/versions/8.4/E_STRICT-deprecated#:~:text=The%20E_STRICT%20constant%20is%20deprecated,be%20removed%20in%20PHP%209.0

### Related Pull Requests

1. See OpenMage/magento-lts#4286


ADDISON74 (approved):

app/code/core/Mage/Core/functions.php
```diff
@@ -165,7 +165,7 @@ function mageCoreErrorHandler($errno, $errstr, $errfile, $errline)
         case E_USER_NOTICE:
             $errorMessage .= 'User Notice';
             break;
-        case E_STRICT:
+        case PHP_VERSION_ID < 80400 && E_STRICT:
```
kiatng, line 168: The `case` is based on `switch($errno)`, which is not a bool comparison.

sreichel: I dont get it. We switch to operation based on $errno, with an additional condition. I dont have a better idea right now.

ADDISON74: https://github.com/sebastianbergmann/phpunit/issues/5956

kiatng, line 168:
```suggestion
        case 2048: // E_STRICT prior to PHP8.4
```

ADDISON74: It is much better using the error value in this case.

ADDISON74 (dismissed):

ADDISON74 (approved):

kiatng (approved):

PR #4293, opened 2024-10-18
Update Mage_SalesRule_Model_Quote_Discount

Hanmac:
### Description (*)

* DiscountDescription should only be string or null (not array)
* strlen needs a type check for string before
* sales_quote_address_discount_item needs to be done after calculator->process

1. ...
2. ...


hirale:
About DiscountDescrption, check this file
 https://github.com/OpenMage/magento-lts/blob/main/app%2Fcode%2Fcore%2FMage%2FSalesRule%2FModel%2FValidator.php#L1039

sreichel:
Wouldnt it be better to dispatch an `_after` event?

Hanmac:
> Wouldnt it be better to dispatch an `_after` event?

The problem right now is that the existing event is either called before or after the calculator process

If you want a new event, you probably want one before process

kiatng:
I agree with the change.

sreichel (approved):

kiatng (approved):

PR #4297, opened 2024-10-19
[PHP 8.4] fgetcsv() - The $escape parameter must be provided as its default value will change

ADDISON74:
Starting with PHP version 8.4 the fifth parameter of the fgetcsv() function must be specified, otherwise the following message will be displayed

```php
Deprecated functionality: fgetcsv(): the $escape parameter must be provided as its default value will change  in /var/www/html/lib/Varien/File/Csv.php on line 80
```

The PHP default value for the escape parameter is ```'\\'``` and I declared it at the beginning in some files.

Since we support the PHP version 7.4, I did not use ```escape:``` in front of the variable like others who provided similar solutions.

### DIFF
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -209,7 +209,8 @@ class Varien_Io_File extends Varien_Io_Abstract
         if (!$this->_streamHandler) {
             return false;
         }
-        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure);
+        $escape = '\\';
+        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, $escape);
     }
 
     /**
@@ -241,8 +242,9 @@ class Varien_Io_File extends Varien_Io_Abstract
         if (!$this->_streamHandler) {
             return false;
         }
-
-        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure);
+        
+        $escape = '\\';
+        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, $escape);
     }
 
     /**
--- a/lib/Varien/Io/File.php
+++ b/lib/Varien/Io/File.php
@@ -210,7 +210,7 @@ class Varien_Io_File extends Varien_Io_Abstract
             return false;
         }
         $escape = '\\';
-        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, $escape);
+        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure, $escape);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Varien/Io/File.php
```diff
@@ -209,7 +209,8 @@ public function streamReadCsv($delimiter = ',', $enclosure = '"')
         if (!$this->_streamHandler) {
             return false;
         }
-        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure);
+        $escape = '\\';
+        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, $escape);
```
ADDISON74, line 213: PHPStan is right, this was an existing issue before this PR. 

The function ```streamReadCsv($delimiter = ',', $enclosure = '"')``` gets only two parameters comparing with ```streamWriteCsv(array $row, $delimiter = ',', $enclosure = '"')``` which gets $row as parameter.

sreichel: $row should be changed back to 0.

lib/Varien/Io/File.php
```diff
@@ -209,7 +209,8 @@ public function streamReadCsv($delimiter = ',', $enclosure = '"')
         if (!$this->_streamHandler) {
             return false;
         }
-        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure);
+        $escape = '\\';
+        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure, $escape);
```
sreichel, line 213:
```suggestion
        return @fgetcsv($this->_streamHandler, 0, $delimiter, $enclosure, '\\');
```

sreichel (dismissed):
lgtm

ADDISON74:
@sreichel - All green now.

sreichel (approved):

kiatng (approved):

S0FTWEX:
After a bit of investigation I found some other occurrences in the code that were overlooked. Is it possible to add them as well so that the OpenMage code is compatible with PHP 8.4?
Here are the mentioned files and line numbers in the code (OpenMage v20.15.0):
```
./app/code/core/Mage/Paypal/Model/Report/Settlement.php : 249
./app/code/core/Mage/Dataflow/Model/Convert/Iterator/Http.php : 36
./lib/Varien/Io/File.php : 201
./vendor/shardj/zf1-future/library/Zend/Auth/Adapter/Http/Resolver/File.php : 160
```

kiatng:
> After a bit of investigation I found some other occurrences in the code that were overlooked. Is it possible to add them as well so that the OpenMage code is compatible with PHP 8.4? 

@S0FTWEX Would you be able to create a PR? It's the only way to get it into OpenMage codebase.

addison74:
For the first 3 files, PR must be created in our repository. For the rest, it must be in the ZF1-Future repository.

PR #4298, opened 2024-10-19
[PHP 8.4] fputcsv() - The $escape parameter must be provided as its default value will change

ADDISON74:
Starting with PHP version 8.4 the fifth parameter of the fputcsv() function must be specified, otherwise the following message will be displayed

```php
Deprecated functionality: fputcsv(): the $escape parameter must be provided as its default value will change  in /var/www/html/app/code/core/Mage/ImportExport/Model/Export/Adapter/Csv.php on line 108 in /var/www/html/app/code/core/Mage/Core/functions.php:197
```

The PHP default value for the escape parameter is ```'\\'``` and I declared it at the beginning in some files.

Since we support the PHP version 7.4, I did not use ```escape:``` in front of the variable like others who provided similar solutions.

### DIFF
--- a/lib/Varien/File/Csv.php
+++ b/lib/Varien/File/Csv.php
@@ -22,6 +22,7 @@ class Varien_File_Csv
     protected $_lineLength = 0;
     protected $_delimiter = ',';
     protected $_enclosure = '"';
+    protected $_escape = '\\';
 
     public function __construct()
     {
@@ -115,7 +116,7 @@ class Varien_File_Csv
     {
         $fh = fopen($file, 'w');
         foreach ($data as $dataRow) {
-            $this->fputcsv($fh, $dataRow, $this->_delimiter, $this->_enclosure);
+            $this->fputcsv($fh, $dataRow, $this->_delimiter, $this->_enclosure, $this->_escape);
         }
         fclose($fh);
         return $this;
--- a/lib/Varien/File/Csv.php
+++ b/lib/Varien/File/Csv.php
@@ -22,7 +22,6 @@ class Varien_File_Csv
     protected $_lineLength = 0;
     protected $_delimiter = ',';
     protected $_enclosure = '"';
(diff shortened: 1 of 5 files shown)


lib/Varien/File/Csv.php
```diff
@@ -115,7 +116,7 @@ public function saveData($file, $data)
     {
         $fh = fopen($file, 'w');
         foreach ($data as $dataRow) {
-            $this->fputcsv($fh, $dataRow, $this->_delimiter, $this->_enclosure);
+            $this->fputcsv($fh, $dataRow, $this->_delimiter, $this->_enclosure, $this->_escape);
```
ADDISON74, line 119: This seems to be a PHPStan error in my opinion. Starting with PHP 8.4 all 5 parameters are required.

sreichel: Changes in this file can be reverted ... it does not not use php fputcsv(), but Varien_File_Csv::fputcsv()

lib/Varien/Io/File.php
```diff
@@ -241,8 +241,9 @@ public function streamWriteCsv(array $row, $delimiter = ',', $enclosure = '"')
         if (!$this->_streamHandler) {
             return false;
         }
-
-        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure);
+        
+        $escape = '\\';
+        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, $escape);
```
sreichel, line 246:
```suggestion
        return @fputcsv($this->_streamHandler, $row, $delimiter, $enclosure, '\\');
```

sreichel:
Pls run `ddev composer run php-cs-fixer:fix`

ADDISON74:
@sreichel - All green now.

sreichel (approved):

kiatng (approved):

PR #3482, opened 2023-08-31
Add PHPMD to workflow

sreichel:
### Description (*)

Added PHPMD to worklow.

### Related Pull Requests

1. See OpenMage/magento-lts#2771

### Manual testing scenarios (*)

1. `ddev phpmd <path>`

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #2771, opened 2022-12-02
PHPMD: added basic config

sreichel:
### Description (*)

Last tool to add ... [PHPMD](https://phpmd.org/)

- [x] added composer (dev) dependency (already there)
- [x] added default config file
- [x] suppress error message for `@` operator

Todo (new PRs)
- [ ] add inline-comments for common errors
  - [x] ErrorControlOperator (this PR)
  - [ ] UnusedFormalParameter (#2786 )
- [ ] add baseline file
- [ ] add to github workflow

#### Test

`vendor/bin/phpmd app/code/core/Mage/ text .phpmd.dist.xml`

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)
 - [x] Add yourself to contributors list

PR #4303, opened 2024-10-23
Code quality: added PHPMD, updated workflow

sreichel:
### PHPMD

Added to workflow. The config is there since years, but it never was used b/c of to much "common" errors.

- renamed a lot of variables that does not **fit camel-case**.
  **Note:** Underscore-prefix is still allowed but may be removed (?).
- suppressed some errors we cant change in DOCs (usage of Superglobals, silence operator).
- excluded some sniff to re-check later
- baselined some error that could be reviewed in time (empty catch blocks, unused private methods e.g.)

### PHPCS: ECG standard

This check was useles. It had only onyl rule (`Ecg.Performance.Loop.ModelLSD`) enabled and every files was excluded.

Added rules. Some of them could/should be removed/fixed seperatly.

### phpstorm

In phpstorms phpcs-config you can only select one config file for code-sniffs, but we have three (default, phpcompa, ecg). Added `.phpcs.xml` to allow syntax highlighting for all sniffs. Just add this file as custom ruleset.

**Note:** for custom rules use local `phpcs.xml`

### Minor changes

- dot-prefixed all dev-files (phpunit, rector, ... for cleaner basedir)
- added some script commands to composer.json
- removed a lot of `phpcs:ignore comments` (not all)

### Related Pull Requests

- see OpenMage/magento-lts#2400
- see OpenMage/magento-lts#2771
- see OpenMage/magento-lts#3482


kiatng (approved):

addison74 (approved):

app/code/core/Mage/Core/functions.php
```diff
@@ -294,7 +295,7 @@ function mageParseCsv($string, $delimiter = ',', $enclosure = '"', $escape = '\\
  *
  * @SuppressWarnings(PHPMD.ErrorControlOperator)
  */
-function is_dir_writeable($dir)
+function isDirWriteable($dir)
```
fballiano, line 298: isn't this a breaking change?

PR #4305, opened 2024-10-24
Dont show disabled modules in "Disable Modules Output" config

sreichel:
### Description (*)

Don't show disabled modules in admin - system - config - advanced - advanced

1. disable a module
2. flush cache
3. go to admin - system - config - advanced - advanced

Disabled module is still there.


ADDISON74 (approved):
Tested and it works as expected.

kiatng:
Instead of not showing, is it better to inform the users that the module is inactivated in the global config? An example:
image

sreichel:
Thought about it, but it makes no sense.

- it is only displayed if it is still there in filesystem
- it is disabled. Its confusing.

I put it on draft b/c ...

- pass the modules array to observer (not only array keys)
- add observer
- hide disabled modules there

sreichel:
> Instead of not showing, is it better to inform the users that the module is inactivated in the global config? An example: image

I tried, but it would require some more changes.

app/code/core/Mage/Adminhtml/Block/System/Config/Form/Fieldset/Modules/DisableOutput.php
```diff
@@ -49,6 +39,22 @@ public function render(Varien_Data_Form_Element_Abstract $element)
         return $html;
     }
 
+    public function getModules(): array
+    {
+        $modules = (array)Mage::getConfig()->getNode('modules')->children();
```
kiatng, line 44: Won't this break BC? Before this PR:
```php
$modules = array_keys((array)Mage::getConfig()->getNode('modules')->children());
```
It is a simple array with int keys. 

After this PR, the keys are string, 'Mage_Core' etc. It may break 3rd-party observer.

sreichel: Yes, it would be. [But] I think the observer was useles before. Will change it.

kiatng:
Alternative:

```php
   // File app/code/core/Mage/Adminhtml/Block/System/Config/Form/Fieldset/Modules/DisableOutput.php
    protected function _getFieldHtml($fieldset, $moduleName)
    {
        $isInactive = (string) Mage::getConfig()->getNode('modules')->$moduleName->active !== 'true';
        // ...
        $field = $fieldset->addField(
            $moduleName,
            'select',
            [
                'name'          => 'groups[modules_disable_output][fields][' . $moduleName . '][value]',
                // ...
                'disabled'      => $isInactive,
                'title'         => $isInactive ? 'Inactivated in /etc/modules' : '',
        //...
```

sreichel:
Wouldnt this not bypass the observers?

kiatng:
> Wouldnt this not bypass the observers?

Yes, observer is not required. And if you want to not show the modules: `if ($isInactive) return ''`. The is active check is microseconds. It's a simpler implementation, no additional loop on modules.

sreichel:
> The is active check is microseconds.

I've tested it. Running `Mage::getConfig()->getNode('modules')` 60 times is far slower.

kiatng:
```
$isInactive = (string) Mage::getConfig()->getNode('modules')->$moduleName->active !== 'true';
```

My tests:

1. with all cache disabled, 5 microseconds
2. with all cache enabled, 8 microseconds

Really weird. Is it just me or for you too?

justinbeaty:
> ```
> $moduleName = 'Mage_Core';
> $isInactive = (string) Mage::getConfig()->getNode('modules')->$moduleName->active !== 'true';
> ```
> 
> My tests:
> 
> 1. with all cache disabled, 5 microseconds
> 2. with all cache enabled, 8 microseconds, yes it's worst!
> 
> Really weird. Is it just me or for you too?

I actually don’t think the modules xml section is cached at all, probably so you can disable a broken module. But I’m not at my desk to confirm this.

kiatng:
Another test:

```php
        $modules = array_keys(Mage::getConfig()->getNode('modules')->asArray());
        foreach ($modules as $moduleName) {
            $result[$moduleName] = (string) Mage::getConfig()->getNode('modules')->$moduleName->active !== 'true';
        }
```
1. Config cache disabled, 650 microseconds
2. Config cache enabled, 930 microseconds, it's better to disable cache?

sreichel:
My tests:

```php
require_once('app/Mage.php');
umask(0);
Mage::app()->loadArea('adminhtml'); #to call the observer

function test4305_Observer() {
    $test = new Mage_Adminhtml_Block_System_Config_Form_Fieldset_Modules_DisableOutput();
    $test->getModules();
}

function test4305_Config() {
    for ($i = 0; $i <= 60; $i++) {
        $isInactive = (string)Mage::getConfig()->getNode('modules')->Mage_Core->active !== 'true';
    }
}
```
---

> * Config cache disabled, 650 microseconds
> 
> * Config cache enabled, 930 microseconds, it's better to disable cache?

No.

Config cache enabled (last 5) vs config cache disabled.

xh_cache1

I'd recommand to test https://github.com/OpenMage/magento-lts/pull/4325.

PR #4306, opened 2024-10-24
Fix deprecated message for strlen first value null

leissbua:

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/ImportExport/Model/Import/Entity/Product/Type/Abstract.php
+++ b/app/code/core/Mage/ImportExport/Model/Import/Entity/Product/Type/Abstract.php
@@ -171,7 +171,7 @@ abstract class Mage_ImportExport_Model_Import_Entity_Product_Type_Abstract
                             'is_static'        => $attribute->isStatic(),
                             'apply_to'         => $attribute->getApplyTo(),
                             'type'             => Mage_ImportExport_Model_Import::getAttributeType($attribute),
-                            'default_value'    => strlen($attribute->getDefaultValue())
+                            'default_value'    => strlen($attribute->getDefaultValue() ?? '')
                                                   ? $attribute->getDefaultValue() : null,
                             'options'          => $this->_entityModel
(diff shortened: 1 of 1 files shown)


ADDISON74:
@leissbua - How did you reproduce the error that led to the proposed change?

app/code/core/Mage/ImportExport/Model/Import/Entity/Product/Type/Abstract.php
```diff
@@ -171,7 +171,7 @@ protected function _initAttributes()
                             'is_static'        => $attribute->isStatic(),
                             'apply_to'         => $attribute->getApplyTo(),
                             'type'             => Mage_ImportExport_Model_Import::getAttributeType($attribute),
-                            'default_value'    => strlen($attribute->getDefaultValue())
+                            'default_value'    => strlen($attribute->getDefaultValue() ?? '')
                                                   ? $attribute->getDefaultValue() : null,
```
kiatng, line 175:
```suggestion
                            'default_value'    => $attribute->getDefaultValue(),
```

sreichel: @kiatng if `$attribute->getDefaultValue()` is an empty string it had returned `null`, now it would be empty string.

leissbua:
@ADDISON74 we experienced deprecation warnings, that first parameter passed to can not be null. From my pov as null can no longer be passed this has to be fixed in general for all strpos strlen and so on.

sreichel:
> this has to be fixed in general for all strpos strlen and so on

I think we should not blindly add echecks everywher.e. Better check for the cause. imho. Maybe check for null before or add getter()-method.

sreichel (approved):
Steps to reproduce would be nice.

null seems to be a possible return type ...

```php
    public function getDefaultValue()
    {
        return $this->_getData('default_value');
    }
```

kiatng (approved):

Issue #4312, opened 2024-10-25
[PHP 8.4] Warning: foreach() argument must be of type array|object, string given


addison74:
**This is the error message** 

Warning: foreach() argument must be of type array|object, string given  in /var/www/html/app/design/adminhtml/default/default/template/widget/grid.phtml on line 163

```php
            <?php if ($_multipleRows = $this->getMultipleRows($_item)):?>
                <?php foreach ($_multipleRows as $_i):?> <= LINE 163
                <tr>
                    <?php $i=0;foreach ($this->getMultipleRowColumns($_i) as $_column): ?>
                        <td class="<?php echo $_column->getCssProperty() ?> <?php echo ++$i==$numColumns-1?'last':'' ?>">
                            <?php echo (($_html = $_column->getRowField($_i)) != '' ? $_html : '&nbsp;') ?>
                        </td>
                    <?php endforeach ?>
                </tr>
                <?php endforeach ?>
            <?php endif ?>
```

I am getting this error when running extension MageWorx SEO Suite Ultimate. Just accesss **Backend > Catalog > SEO Breadcrumbs > Manage Breadcrumbs Priority for Category**. Maybe other extensions could produce the same issue.

This must be solved using **is_iterable()** function

```php
            <?php if ($_multipleRows = $this->getMultipleRows($_item)):?>
                <?php if (is_iterable($_multipleRows)): ?>
                    <?php foreach ($_multipleRows as $_i):?>
                <tr>
                    <?php $i=0;foreach ($this->getMultipleRowColumns($_i) as $_column): ?>
                        <td class="<?php echo $_column->getCssProperty() ?> <?php echo ++$i==$numColumns-1?'last':'' ?>">
                            <?php echo (($_html = $_column->getRowField($_i)) != '' ? $_html : '&nbsp;') ?>
                        </td>
                    <?php endforeach ?>
                </tr>
                <?php endforeach ?>
                <?php endif ?>
```

TangLiang:
`Class extends Mage_Adminhtml_Block_Widget_Grid
{
    public function getMultipleRows($item)
    {
        return false;
    }`
find the file which extends Mage_Adminhtml_Block_Widget_Grid and just rewrite this function getMultipleRows, i face the same issue when i add category collection as grid with ajax tab for custom module,as $category has this function getChildren() with returns string if it has children ids

PR #4318, opened 2024-10-27
[PHP 8.4] Warning: foreach() argument must be of type array|object, string given

addison74:
This PR fixes issue #4312.


sreichel:
Can u please add a debug breakpoint here ...

```

```

addison74:
I just debugged the code. 

**$_multipleRows** which stores the function return is a string with this value "10,11,12,13".

sreichel:
Can you pla add stack trace?

sreichel:
> I just debugged the code.
> 
> **$_multipleRows** which stores the function return is a string with this value "10,11,12,13".

Cant cast to array there?

Closed. Module should be fixed.

PR #4320, opened 2024-10-29
Fix `Mage_Adminhtml_Block_Template::isOutputEnabled()` for invalid module

sreichel:
### Description (*)

`Mage_Adminhtml_Block_Template::isOutputEnabled()` returns false now for non-existing module.

```php
$test = new Mage_Adminhtml_Block_Template();
var_dump($test->isOutputEnabled('Mage_Catalog'));
var_dump($test->isOutputEnabled('XXX')); # returns true befor PR
```

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Template.php
+++ b/app/code/core/Mage/Adminhtml/Block/Template.php
@@ -40,7 +40,7 @@ class Mage_Adminhtml_Block_Template extends Mage_Core_Block_Template
     }
 
     /**
-     * Check whether or not the module output is enabled
+     * Check whether the module output is enabled
      *
      * Because many module blocks belong to Adminhtml module,
      * the feature "Disable module output" doesn't cover Admin area
@@ -50,10 +50,7 @@ class Mage_Adminhtml_Block_Template extends Mage_Core_Block_Template
      */
     public function isOutputEnabled($moduleName = null)
     {
-        if ($moduleName === null) {
-            $moduleName = $this->getModuleName();
-        }
-        return !Mage::getStoreConfigFlag('advanced/modules_disable_output/' . $moduleName);
+        return Mage::helper('core')->isModuleOutputEnabled($moduleName);
     }
 
     /**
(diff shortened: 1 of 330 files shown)


ADDISON74 (dismissed):

app/code/core/Mage/Adminhtml/Block/Template.php
```diff
@@ -50,10 +50,7 @@ public function getFormKey()
      */
     public function isOutputEnabled($moduleName = null)
     {
-        if ($moduleName === null) {
-            $moduleName = $this->getModuleName();
-        }
```
justinbeaty, line 55: Removal of these lines technically changes the behavior. If `$moduleName = null`, before we would check `Mage_Adminhtml` and now we would check `Mage_Core`.

However, I only see two calls to this method, and in both cases `$this->isOutputEnabled('Mage_AdminNotification')`

tests/unit/Mage/Adminhtml/Block/TemplateTest.php
```diff
@@ -0,0 +1,80 @@
...
+    public function provideIsOutputEnabled(): Generator
+    {
+        yield 'Mage_Core' => [
+            true,
+            'Mage_Core',
+        ];
+        yield 'Not_Exits' => [
+            false,
+            'Not_Exits',
+        ];
```
justinbeaty, line 68:
```suggestion
        yield 'Not_Exist' => [
            false,
            'Not_Exist',
        ];
```

tests/unit/Mage/Adminhtml/Block/TemplateTest.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+        $this->assertSame($expectedResult, $this->subject->isOutputEnabled($moduleName));
+    }
+
+    public function provideIsOutputEnabled(): Generator
+    {
+        yield 'Mage_Core' => [
+            true,
+            'Mage_Core',
+        ];
```
justinbeaty, line 64:
```suggestion
        yield 'Mage_Adminhtml' => [
            true,
            null,
        ];
        yield 'Mage_Core' => [
            true,
            'Mage_Core',
        ];
```

Not tested but I think this would work to make sure null param works

sreichel: works, just renamed and added test. thanks.

justinbeaty (approved):

ADDISON74 (approved):

kiatng (approved):

PR #4323, opened 2024-10-30
Cache `isModuleOutputEnabled` or `isModuleEnabled` status

sreichel:
### Description (*)

If `isModuleOutputEnabled` or `isModuleEnabled` is called multiple times for the same Module, get config once.

### Related Pull Requests

1. see OpenMage/magento-lts#4320

1. add xdebug breapoints to `isModuleOutputEnabled()`
2. ...

```php
    var_dump($test->isModuleOutputEnabled('Mage_Catalog'));
    var_dump($test->isModuleOutputEnabled('Mage_Catalog'));
    var_dump($test->isModuleOutputEnabled('Mage_Catalog'));
    var_dump($test->isModuleOutputEnabled('a'));
    var_dump($test->isModuleOutputEnabled('a'));
    var_dump($test->isModuleOutputEnabled('b'));
```


app/code/core/Mage/Core/Helper/Abstract.php
```diff
@@ -42,6 +42,10 @@ abstract class Mage_Core_Helper_Abstract
      */
     protected $_layout;
 
+    protected array $modulesDisabled = [];
```
kiatng, line 45: Traditionally, protected var has prefix '_'. Do we want to break tradition?

sreichel: i'd not stick with old traditions.

i think we had PRs that added protected properties w/o underscore.

justinbeaty:
XML / DB config is already cached, how much time is this actually saving?

sreichel:
xh1

For the example above,

sreichel:
> even though the "Disable Modules Output" can be configured at the store scope

I did not see that and did not test it. Thanks.

Config is alerady cached, so it brings no benfits. Check if module is enabled still save some calls. (seems even faster ... numbers from ddev + xhprof + xhgui)

justinbeaty (approved):

sreichel:
https://github.com/OpenMage/magento-lts/pull/4320#discussion_r1821212246
> However, I only see two calls to this method

Updated:

- moved methods to parent abstract class ... so its also available for frontend block classes too
- added trait to provide same funtionality to models, resource,-models, ... (instead of adding helper methods multiple times)

kiatng (dismissed):

kiatng (approved):

addison74 (approved):

Issue #1284, opened 2020-10-26
Cannot add to wishlist from tag page - Customer account


ADDISON74:
**(OpenMage v20.0.4 + RWD theme) on (Debian 9 + Virtualmin).**

1. In backend activate module output for Mage_Tag (System -> Configuration -> Advanced),

2. Go to frontend and visit any product you like. Add tags for it. In my case it is Khaki Bowery Chino Pant and it has two tags chino and pants. 

3. Go to customer account. From the left side click on My Tags. On this page you will see your added tags (url: /tag/customer).

![mytags](https://user-images.githubusercontent.com/8360474/97148962-a6fa5180-1774-11eb-90c3-c27626f44516.jpg)

4. Click on a tag. On this page you will see the associated products with your tag (url: /tag/customer/view/tagId/21).

![wishlist_1](https://user-images.githubusercontent.com/8360474/97149017-c09b9900-1774-11eb-8f0a-33d36a117838.jpg)

5. Click on [Add to Wishlist] button (url: /wishlist/index/add/product/456/). 

6. My Wishlist page is loaded (url: /wishlist/index/) but the product was not added to list.

![wishlist_1](https://user-images.githubusercontent.com/8360474/97149160-00628080-1775-11eb-830f-da305970414e.jpg)

My personal opinion is that Magento Team did not added JS script for this action. Initially I thought it was related to a missing formkey but I found it in template.

Thank you.

PR #4329, opened 2024-11-03
Fix: cannot add to wishlist from tag page - Customer account

sreichel:
### Description (*)

Fixes add to wishlist URL.

1. Fixes OpenMage/magento-lts#1284


ADDISON74 (approved):
Tested and it solves the reported issue.

kiatng (approved):

PR #4346, opened 2024-11-12
PHPStan: some fixes

sreichel:
k/t


app/code/core/Mage/Paypal/Model/Payflowlink.php
```diff
@@ -365,7 +365,6 @@ protected function _getOrderFromResponse()
     /**
      * Build request for getting token
      *
-     * @param Mage_Sales_Model_Order_Payment|Mage_Payment_Model_Info $payment
      * @return Varien_Object
```
kiatng, line 368:
```suggestion
     * @return Mage_Paypal_Model_Payflow_Request
```

fballiano:
what does "k/t" mean?

addison74 (approved):

app/code/core/Mage/SalesRule/Model/Quote/Discount.php
```diff
@@ -134,7 +134,7 @@ protected function _aggregateItemDiscount($item)
     /**
      * Add discount total information to address
      *
-     * @return $this
+     * @return $this|array
```
kiatng, line 137: I cannot see how it can return an array.

sreichel: To match `parrent::fetch(()`, that returns an array.

kiatng (approved):

PR #4349, opened 2024-11-12
DOCs: update

sreichel:
@colinmollenhour i think we can point docs to https://OpenMage.github.io/magento-lts/.

(instead of forwarding to openmage.org)


colinmollenhour:
I tried everything I could find to get rid of the redirect... Not sure why I couldn't, but I was able to get the site published to https://docs.openmage.org/
Note that I had to switch the publish mode to Classic Pages to get it to work and Github created the commit https://github.com/OpenMage/magento-lts/commit/7806f0e529da17a85df6f82c09420116a1a60f26 for me automatically. Do you see any issue with this file being here?

image

README.md
```diff
@@ -14,367 +14,11 @@ level of backwards compatibility to the official releases.
...
-  - [New Events](#new-events)
-  - [Changes to SOAP/WSDL](#changes-to-soapwsdl)
-- [Development Environment with ddev](#development-environment-with-ddev)
-- [PhpStorm Factory Helper](#phpstorm-factory-helper)
-- [PhpStorm File-Watcher for SCSS files](#phpstorm-file-watcher-for-scss-files)
-- [Public Communication](#public-communication)
-- [Maintainers](#maintainers)
-- [License](#license)
-- [Contributors](#contributors-)
+See: https://OpenMage.github.io/magento-lts/
```
colinmollenhour, line 19:
```suggestion
See: https://docs.openmage.org/
```

sreichel: My settings ...

ghp

As long as automatic deployment works docs.openmage.org is even better. Thanks

sreichel:
Added some content from https://github.com/OpenMage/documentation ... (just to have it here).

@colinmollenhour i think the RTD page should be closed (possibly forwarded) to not have outdated content out there. Archive repo?

colinmollenhour:
Yes, I agree that site should be folded into the new docs site. I think the docs being in the same repo will be a big improvement because the issues and PRs will be all in one place. For example, I wasn't even aware of https://github.com/OpenMage/documentation/pull/12.. Also when a BC change is made we can require the docs to be updated in the same PR.

colinmollenhour:
I updated the README and index page and archived the old docs repo.

sreichel:
Thanks.

Having docs here should be more contributor-friendly.

Ping @m-michalis Want to update https://github.com/OpenMage/documentation/pull/12?

m-michalis:
> Ping @m-michalis Want to update [OpenMage/documentation#12](https://github.com/OpenMage/documentation/pull/12)?

In the end, I never did implement K8s cluster with OpenMage. I _played_ with it but that's it. If you see any value over my PR, you are free to do as you please.

sreichel:
@colinmollenhour 

Something is wrong ... check ...

- https://www.openmage.org/magento-lts/install.html
- https://www.openmage.org/magento-lts/about.html

colinmollenhour:
@sreichel I'm not sure we can fix that, I think it is a redirect problem because the path "/magento-lts/" matches the repo name and GitHub pages is treating it as a request to the magento-lts repo site rather than the OpenMage/OpenMage.github.io repo.. I think the best solution is to relocate these pages to a different subdirectory.

sreichel:
Thats (nearly fixed).

## Update

- fixed syntax in API section
- add blog with all releases and a few guides (1)
- minor fixes/improvements

Note 1:

About having blog posts here ...

- CONs
  - <100kb "more"
  - no multilang support atm, mkdocs can, but is it worth?

- PROs
  - no other repo, easier to update
  - context is searchhable, current blog is not

colinmollenhour:
I think my preference is to keep blog posts in the main site so that writing and publishing them can follow a different process than code and documentation changes. Docs make sense, but blog posts could be random thoughts, announcements, etc.. I don't think we want that in our code repo so it's more about separation of concerns for me.

sreichel:
> I think my preference is to keep blog posts in the main site

Removed

- https://www.openmage.org/2024/10/03/payment-method-support-for-brazil.html
- https://www.openmage.org/2023/01/03/do-you-really-need-more.html
- https://www.openmage.org/2019/01/20/after-m1-eol.html

... but kept under guides ...

- https://www.openmage.org/2023/05/01/customize-your-openmage.html

sreichel:
Updated modules section

Source: https://github.com/fballiano/awesome-openmage

Todo

- check https://github.com/aleron75/mageres/blob/master/README-M1.md
- add you extensoins ...

kiatng (approved):

addison74 (approved):

sreichel:
@colinmollenhour have you changed something? Guess not ...

https://docs.openmage.org does not work, but this does ...

https://www.openmage.org/magento-lts/

colinmollenhour:
Strange.. I just set the custom domain again so let's see if that works..

sreichel:
Thanks. Seems to work now.

PR #4356, opened 2024-11-15
Cleanup:  removed broken `Mage_Core_Controller_Front_Router`

sreichel:
### Description (*)

> Fatal error:  Uncaught Exception: Notice: Array to string conversion  in /var/www/html/app/code/core/Mage/Core/Model/Store.php on line 580

```php
$config = new Mage_Core_Model_Config_Element('<?xml version="1.0" encoding="utf-8" ?><test-url />');
$router = new Mage_Core_Controller_Front_Router($config);
var_dump($router->getUrl());
```


app/Mage.php
```diff
@@ -444,7 +444,7 @@ public static function getStoreConfigFlag($path, $store = null)
     /**
      * Get base URL path by type
      *
-     * @param string $type
+     * @param Mage_Core_Model_Store::URL_TYPE_* $type
```
kiatng, line 447: I have not seen this before, not sure if that's a valid doc block. Alternative:
```suggestion
     * @param string $type One of Mage_Core_Model_Store::URL_TYPE_* constants
```

kiatng:
Any idea where is this class `Mage_Core_Controller_Front_Router` reference in the core? I didn't find any, should it be marked `@deprecated`?

app/Mage.php
justinbeaty: It seems that it is supported https://phpstan.org/writing-php-code/phpdoc-types#literals-and-constants

kiatng: Thanks, learn something new.

kiatng (dismissed):

sreichel:
> should it be marked `@deprecated`?

Better to remove it? It was a fatal error in any case ... not usable at all.

kiatng:
Yes, it's not functional, make sense to remove.

kiatng:
I was thinking removing the entire file.

kiatng (approved):

addison74 (approved):

kiatng (approved):

PR #4357, opened 2024-11-15
Fixed null deprecation in `Mage_Catalog_Model_Product_Option_Type_Text::validateUserValue()`

kiatng:
```
    [type] => 8192:E_DEPRECATED
    [message] => trim(): Passing null to parameter #1 ($string) of type string is deprecated
    [file] => .../app/code/core/Mage/Catalog/Model/Product/Option/Type/Text.php
    [line] => 36
```

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Option/Type/Default.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Option/Type/Default.php
@@ -54,9 +54,12 @@ class Mage_Catalog_Model_Product_Option_Type_Default extends Varien_Object
      */
     protected $_formattedOptionValue = null;
 
-    public function getUserValue(): string
+    /**
+     * @return string|array
+     */
+    public function getUserValue()
     {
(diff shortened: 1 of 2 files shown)


sreichel:
Its an issue in core or 3rd-party?

kiatng:
It happened in custom code to create an order programmatically, I agree that null check for values can be placed here. But I feel the core should be more tolerant because it works before.  It's a matter of opinion, I have no issue if others feel otherwise.

sreichel:
Mhh, instead of adding null-checks everywhere (not only in this PR), why not to override the magic method calls?

```php
    public function getUserValue(): string
    {
        return (string) $this->getDataByKey('user_value');
    }
```

sreichel:
you made the change in the wrong file. check https://github.com/kiatng/magento-lts/pull/2/files

sreichel (approved):

addison74 (approved):

PR #4362, opened 2024-11-17
Added methods to access different sessions

sreichel:
### Description (*)

...w/o have to type `Mage::getSingleton('some/session')` ...


fballiano:
wouldn't it be be easier to just have Mage::getAdminSession() Mage::getCoreSession(), it seem it would have less impact

justinbeaty:
While I like @fballiano's idea more, I'm not sure I really see the need for this change at all? There are tons of other models called with `Mage::getSingleton()`, why make special method for these and not others? What if I make a custom session model, then there's an inconsistency between how I get session models.

kiatng:
With copilot, and especially cursor AI, more and more typing are done by them, hence less and lees need for this PR.

sreichel:
> wouldn't it be be easier to just have Mage::getAdminSession() Mage::getCoreSession(), it seem it would have less impact

What do you mean?

Add helper methods for "core" and "admin" sessions only (and leave everything else at it is)?

> There are tons of other models called with Mage::getSingleton(), why make special method for these and not others?

Absolutly aggree! E.g. there is `Mage::getSingleton('eav/config')` w/ ~ 100 occurences ...

Yes. This can also be "fixed/improved" by adding `getEavConfig()` or something like that.

> With copilot, and especially cursor AI, more and more typing are done by them, hence less and lees need for this PR.

I dont use any of that. Now ... in phpstorm i can use `[$this->]` + `g` + `S` ... for autocomplete `get[some]Session()`

sreichel:
Yes, Its only a wrapper. (that was the idea of it :smile:)

> Why then not also special methods for getting regular models too?
>
> $ git grep "Mage::getModel('catalog/product')" | wc -l
> 147

+1 for  `getProductModel()` or `getCatalogProductModel()`. In another PR?

justinbeaty:
Personally I am just not a fan... For one, it's adding references to different modules in Mage_Core. Also, where does it end? Which models are important enough and which ones aren't?

sreichel:
> For one, it's adding references to different modules in Mage_Core.

True. But traits are extendable. We can still split it into modules. (?)

> Also, where does it end?
> Which models are important enough and which ones aren't?

I think i have covered ALL session types .

PR #4368, opened 2024-11-18
Fix validateHttpVersion to be RFC 1945 and RFC 7230 compliant

leissbua:
### Description (*)

`Mage_HTTP_Client_Curl` is expecting the HTTP Response Header to have 3 elements.

1. HTTP-Version 
2. Status-Code
3. Reason-Phrase

However, the Reason-Phrase is optional, e.g. Apple Pay does not return OK Reason-Phrase on the current `validationURL()`endpoint which is HTTP 1.1. I fixed the code to allow 2 or 3 elements as it does for HTTP 2.0 in the current version


sreichel (commented):
.

leissbua:
What does "." mean @sreichel

sreichel:
@leissbua was a comment on your first commit. The change you have reverted,

kiatng (approved):
Ref https://www.rfc-editor.org/rfc/rfc7230#section-3.1.2:

> The reason-phrase element exists for the sole purpose of providing a textual description associated with the numeric status code, mostly out of deference to earlier Internet application protocols that were more frequently used with interactive text clients. A client SHOULD ignore the reason-phrase content.

sreichel:
Thanks for that info.

sreichel (approved):

sreichel:
For HTTP 1.0

https://datatracker.ietf.org/doc/html/rfc1945#section-6.1.1

PR #4377, opened 2024-11-23
Added `JsonSerializable` interface to `Varien_Object`

sreichel:
### Description (*)

Thanks to @justinbeaty.

Only rewrote tests to check before/after.

### Related Pull Requests
- see https://github.com/MahoCommerce/maho/pull/55

### Questions or comments

Not only adding **a useful feature**, it also reveals in unittest, that `Varien_Object::getData()` is not fully working with nested data - returning `null`.

Btw ... i follow your PRs. It would be nice to have them here to. :smiley:


justinbeaty:
> Btw ... i follow your PRs. It would be nice to have them here to.

Maybe in time... but also feel free to backport them too. Btw, I think `lib/Varien/Object.php` should have the Maho copyright header. 

```php
 * @copyright  Copyright (c) 2024 Maho (https://mahocommerce.com)
```

It wasn't in https://github.com/MahoCommerce/maho/pull/55 because that file had already been updated by Fabrizio.

sreichel:
@justinbeaty with all respect - no. Should we start to add every contributor who changed a few lines to copyright?

IMHO ... I think porting 5 lines requires nochange in copyright header.

fballiano:
In my opinion copyright could be even so much simpler

`@copyright openmage.org, magento.com`
(in this order because as far as i could see it should be reverse order, so first mentioned is the last added)

sparing a lot of bytes (in disk space and bandwith) that don't really mean anything to anybody, "magento, inc." doesn't exist anymore since a million years, "the openmage contributors" I think it's loo long (the article especially to me is sore)

justinbeaty:
@sreichel a thought experiment:

Say MageOne opensourced their code, and has included "Copyright 2024 Paddox GmbH" in their changed files. If we port changes from that codebase, we would have to include their copyright header, right? It should not be any different if changes are ported from Maho.

However, I would probably concede this specific PR might be trivial enough and not even be copyrightable. But this discussion applies to more than just this PR.

@fballiano 
> in this order because as far as i could see it should be reverse order, so first mentioned is the last added

I have always heard that new copyright notes go below. Also I *really* wouldn't remove the Magento Inc notice. I'm pretty sure the sale from Magento, Inc to eBay was a stock sale, but I'm not sure about from eBay to Adobe. In any case, I would find it likely Adobe might still have Magento Inc active in some way or form.

But also, I like the separate lines still. Git compresses files right?

fballiano:
still think it's a lot of characters that don't mean anything to anybody, but legal is legal sadly and we will have to die with those lines

```
* @copyright   Magento, Inc. (https://magento.com)
 * @copyright   OpenMage Contributors (https://openmage.org)
```

I would at least do something like this, remove the year and the "copyright (c)" and the "www."

Personally don't like the website within parenthesis either

sreichel:
> However, I would probably concede this specific PR might be trivial enough and not even be copyrightable. But this discussion applies to more than just this PR.

Its only about this specific PR!

Lets assume i would adapt most parts of your autoloader-change ... no need to discuss about adding that.

However ...  can add that later.

Btw ... should we update Magentos copyright info, when we port back changes? (getData(), getDataByKey(), ...)?

fballiano:
> @justinbeaty with all respect - no. Should we start to add every contributor who changed a few lines to copyright?
> 
> IMHO ... I think porting 5 lines requires nochange in copyright header.

many times you update the copyright when fixing one sigle typo.

anyway I'm no longer a contributor, it's a different project. but also, openmage is mentioned everywhere on maho's website while it seems that on this repository maho's name is the secret nobody has to know about. so it's a matter of good will.

sreichel:
> @copyright   Magento, Inc. (https://magento.com)

I'd not touch Magentos CR.

Removing www ... yes.

colinmollenhour:
I would refrain from changing the Copyright that is not ours unless and until we backport changes from the original holder. If the entity is "dead", it just shows that the original Copyright holder is now "dead", but that doesn't mean the notice should be removed.

Interestingly, M2 source code still says `* Copyright © Magento, Inc. All rights reserved.` so I'm not so sure it is dead?

> Btw ... should we update Magentos copyright info, when we port back changes? (getData(), getDataByKey(), ...)?

I assume you mean from M2? That's tough because unless it is clearly a copy, it is likely more of a discovery of a possibility for improvement, and the final changes to OM will liklely be materially different than the initial change for M2. But if it "looks and smells" like a copy, it would probably be safest to add/update the copyright info in the same PR as well.

fballiano:
> Interestingly, M2 source code still says `* Copyright © Magento, Inc. All rights reserved.` so I'm not so sure it is dead?

right, I thought it would say adobe, I'm sure I've seen some adobe headers somewhere.
I thought linking to magento.com would kinda have the same effect

> I assume you mean from M2? That's tough because unless it is clearly a copy, it is likely more of a discovery of a possibility for improvement, and the final changes to OM will liklely be materially different than the initial change for M2. But if it "looks and smells" like a copy, it would probably be safest to add/update the copyright info in the same PR as well.

if something is ported from any project than license of that project applies, in case of OSL3 there's at least this point:

> 6) Attribution Rights. You must retain, in the Source Code of any Derivative Works that You create, all copyright, patent, or trademark notices from the Source Code of the Original Work, as well as any notices of licensing and any descriptive text identified therein as an “Attribution Notice.” You must cause the Source Code for any Derivative Works that You create to carry a prominent Attribution Notice reasonably calculated to inform recipients that You have modified the Original Work.

sreichel:
> if something is ported from any project than license of that project applies, in case of OSL3 there's at least this point:

This may be true when using a library, bot not for 5 line that is trivial code.

> to inform recipients that You have modified the Original Work.

Shouldnt Maho report changes back? Tthe "original work" is OpenMage.

colinmollenhour:
> Shouldnt Maho report changes back? Tthe "original work" is OpenMage.

It says "inform recipients", not "inform original authors", so for example, neither OpenMage nor Maho have any obligation to inform Magento, Inc. (assuming it exists) of the derivative works we are creating, only the recipients of the derivative works (people who choose to download them). We do this by way of updating the source code headers to provide such notice, so this obligation is met if we copy the Copyright lines.

colinmollenhour:
I don't think there is any "minimum" amount of change that makes a copyright/license no longer apply, unfortunately. So I think even if a single line is sourced from anyone not directly contributing to OpenMage (e.g. Maho, Adobe, MageOne, etc.), we should observe any additional licenses or copyrights and make sure the proper attribution has been made accordingly. For this PR, I believe that means adding the copyright line if the changes are being backported from Maho code. If someone donates code to us via a PR, then there is nothing to add in terms of licensing and it is covered under the "OpenMage community".

sreichel:
> I don't think there is any "minimum" amount of change

Lawyer speak. "reasonably calculated".  :+1: 

I'd never ever thought to discuss about copyright change for adding 5 lines (from a forked project that takes ervy PR.)

colinmollenhour:
I don't think this needs to be closed; it seems the attribution discussion is worth having... not trying to prolong it.

I read "reasonably calculated" as pertaining to the prominence of the attribution notice, not the degree of change that constitutes a derivative work.

sreichel:
@colinmollenhour its not worth for this change.

PR #4382, opened 2024-11-25
Inserted sample nginx config for API in api_nginx-frontend.conf

kiatng:
### Description (*)
This PR provides a sample config for API in api_nginx-frontend.conf, which took me a while to figure out.

### Related Pull Requests
PR #1209 
PR #1210


colinmollenhour:
Perhaps the REST API should be default exposed? It was mainly the XMLRPC API that I figured didn't need to be exposed by default.

kiatng:
> Perhaps the REST API should be default exposed? It was mainly the XMLRPC API that I figured didn't need to be exposed by default.

I'm not sure. My guess is that SOAP is the common one. REST is tied to OAuth, limited use case. I don't think many use the new JSON-RPC. We do not have to expose a default API. `\api.php` takes care of any API.

colinmollenhour (approved):

addison74 (approved):

PR #4399, opened 2024-11-30
Fix: InvalidParamTypeProvidedNULL: fixed login() method

eneiasramos:
### Description (*)

I got an error using API when applying the last commits to my master branch.

Apparently any API call is generating the error: Invalid param type provided ("null")

Applying the fix the error disappears.

image

image

1. ...
2. ...


kiatng:
I have this commit 41ea3c68a39ff694fdd07d4c53ee5b6a8ffa6f13 in production, which uses JSON-RPC with basic auth with   requests daily, so far nothing bad is reported.

eneiasramos:
@kiatng

Another consideration to do is that apiKey parameter can be null ( apiKey = null ).
However the invalid_request_param fault is triggered in this case.

```
if (empty($username) || empty($apiKey)) {
            $this->_fault('invalid_request_param');
            return;
        }
```

So it doesn't make sense that apiKey could be null at this point :)

sreichel:
> Another consideration to do is that apiKey parameter can be null ( apiKey = null ).
However the invalid_request_param fault is triggered in this case.

We should not change methods parameters. Default `null` does not do anything. Removing it should not change you problem.

eneiasramos:
@sreichel 

Actually the change solved the problem :)

https://github.com/gamuzatech/tolucastore-openmage/commit/afec611a08cbeee8b0fb0dc8cdd1ef4c56f85e25

image

sreichel:
@eneiasramos  i dont see how doc type should have impact on it ...  please check.

sreichel:
https://github.com/OpenMage/magento-lts/pull/4399

app/code/core/Mage/Api/Model/Server/Handler/Abstract.php
```diff
@@ -204,10 +204,10 @@ protected function _prepareResourceModelName($resource)
      * Login user and Retrieve session id
      *
      * @param string $username
-     * @param string|null $apiKey
+     * @param string $apiKey
      * @return stdClass|string|void
      */
-    public function login($username, $apiKey = null)
+    public function login($username, $apiKey)
```
sreichel, line 210:
```suggestion
    public function login($username, $apiKey = null)
```

sreichel (approved):

addison74 (approved):

eneiasramos:
@sreichel 

Sorry for the delay in responding.

The adjustment worked as expected.

Thank you very much for your support :)

sreichel:
Thanks for reporting & PR :)

PR #4113, opened 2024-07-24
Install TinyMCE 7.5 via composer

sreichel:
### Description (*)

I've updated the copy-plugin to work if you install OpenMage as composer dependency.

### Related Pull Requests
- see OpenMage/magento-lts#3658
- see OpenMage/magento-lts#3705
- closes OpenMage/magento-lts#4403 

### Update

- moved block classes from Adminhtml/Cms to Cms/Adminhtml (and deprecated old ones)

### Todo

- for another PR ... add another config option to always use editor
- for another PR ... disable editor if TinyMce is not installed

PR #4403, opened 2024-12-07
Tinymce 69

sreichel:
### Description (*)

As long as TinyMCE 7 is not tested (https://github.com/OpenMage/magento-lts/pull/4113), lets install latest v6.8 release.


addison74 (approved):

addison74:
However, periodic updates should be made to v6. In the past it was done by Fabrizio Balliano.

sreichel:
We updated to v7 ... ?!?

addison74:
You're right. Version 6 has been removed from the distributions and is now installed with Composer. That's fine.

Issue #4401, opened 2024-12-05
Coupon not valid when cart promo rule end date is today.


fabianaromagnoli:
### Preconditions

1. OpenMage version 20.6.0
2. PHP 8.0

### Steps to reproduce
1. Setup a cart promo rule with a specific coupon. Set both start date and end date with current date.
2. Go to fronted, add a product to cart and go to the cart page
3. Try to apply the coupon

### Expected result
The coupon is valid and discount is applied

### Actual result

The coupon is not valid and discount is not applied

In [this piece of code](https://github.com/OpenMage/magento-lts/blob/cfa9fbbe6781a9073a9e1862382d89acfd97256e/app/code/core/Mage/SalesRule/Model/Validator.php#L182) the coupon expiration date is compared to the current datetime, but coupon expiration date always has 00:00:00 time, so that if the expiration day is today, that check cannot pass. Dates should be compared without considering time, like it's done [in the Collection](https://github.com/OpenMage/magento-lts/blob/cfa9fbbe6781a9073a9e1862382d89acfd97256e/app/code/core/Mage/SalesRule/Model/Resource/Rule/Collection.php#L153) where the promo validity dates are compared to $now, and $now is a date without time.

fabianaromagnoli:
This patch seems to work fine:

```
--- a/app/code/core/Mage/SalesRule/Model/Validator.php
+++ b/app/code/core/Mage/SalesRule/Model/Validator.php
@@ -179,7 +179,9 @@ protected function _canProcessRule($rule, $address)
                         return false;
                     }
                     // check coupon expiration
-                    if ($coupon->hasExpirationDate() && ($coupon->getExpirationDate() < Mage::getModel('core/date')->date())) {
+                    $couponExpirationDay = Mage::getModel('core/date')->date(Varien_Date::DATE_PHP_FORMAT, $coupon->getExpirationDate());
+                    $today = Mage::getModel('core/date')->date(Varien_Date::DATE_PHP_FORMAT);
+                    if ($coupon->hasExpirationDate() && ($couponExpirationDay < $today)) {
                         $rule->setIsValidForAddress($address, false);
                         return false;
                     }
```

addison74:
Please create a PR to help us checking this issue.

PR #4409, opened 2024-12-09
The coupon is not valid and discount is not applied

fabianaromagnoli:
It solves the issue reported here #4401.

**Preconditions**
OpenMage version 20.6.0
PHP 8.0

**Steps to reproduce**
Setup a cart promo rule with a specific coupon. Set both start date and end date with current date.
Go to fronted, add a product to cart and go to the cart page
Try to apply the coupon

**Expected result**
The coupon is valid and discount is applied

**Actual result**
The coupon is not valid and discount is not applied

In [this piece of code](https://github.com/OpenMage/magento-lts/blob/cfa9fbbe6781a9073a9e1862382d89acfd97256e/app/code/core/Mage/SalesRule/Model/Validator.php#L182) the coupon expiration date is compared to the current datetime, but coupon expiration date always has 00:00:00 time, so that if the expiration day is today, that check cannot pass. Dates should be compared without considering time, like it's done [in the Collection](https://github.com/OpenMage/magento-lts/blob/cfa9fbbe6781a9073a9e1862382d89acfd97256e/app/code/core/Mage/SalesRule/Model/Resource/Rule/Collection.php#L153) where the promo validity dates are compared to $now, and $now is a date without time.


sreichel:
LGTM, but i'd like to suggest to move the chkeck into its own method ...

```php
$couponExpirationDay = Mage::getModel('core/date')->date(Varien_Date::DATE_PHP_FORMAT, $coupon->getExpirationDate());
$today = Mage::getModel('core/date')->date(Varien_Date::DATE_PHP_FORMAT);
if ($coupon->hasExpirationDate() && ($couponExpirationDay < $today)) {
```

to 

```php
if ($this->checkCouponExpirationDay($coupon)) {
```

```php
public function checkCouponExpirationDay(NS $coupon)
{
    #...
)
```

sreichel:
Id change the code to not call `getExpirationDate` before `hasExpirationDate`.

fabianaromagnoli:
@sreichel Thanks for pointing out! What do you think about the refactoring in commit 3cff24f48ab0f4e59d5ea82361fabe746368a2bf? The very bad thing before this commit was that `$couponExpirationDay` had a not-null value even when the coupon had no expiration date. Now it is explicitly `null` if coupon has no expiration date.

sreichel:
Mhh, i think that the original code works as expected.

To make a coupon valid only for this this set "to date" to the next day. (thats 00:00:00) - but that unintential.

I have tested your code, but it does enot work for me. I have a diff of on day in comparison. Wait for PR ...

mattdavenport:
What about ensure the `to_date` is always at the end of the day (e.g. `23:59:59`). Maybe in `Mage_SalesRule_Model_Rule::_afterSave()`? We could expand the functionality of price rules to include time as well. I could see some use cases there "Sale ends at 6PM Friday, use coupon code X". Likely a separate discussion though.

sreichel:
> What about ensure the `to_date` is always at the end of the day (e.g. `23:59:59`).

Thats what idid in https://github.com/OpenMage/magento-lts/pull/4456, when from-date and to-date is the same day.

mattdavenport:
Ah I missed that one. Yeah that implementation LGTM. I propose closing this in favor of #4456.

> Note: we cannot set 23:59:59 for every end-date w/o breaking behavoir of existing rules (?)

Agreed. I think only in this case (same day to/from) this is valid.

sreichel:
@fabianaromagnoli can you please test the linked PR?

sreichel:
@all-contributors add @fabianaromagnoli bug

PR #4411, opened 2024-12-09
New PHPStan Plugin Test

justinbeaty:
### Description (*)
This is test to use a new PHPStan plugin with OM instead of radically changing the macopedia one. Presumably you'd fork the module into a new namespace.

I also cleaned up a lot of the `Mage::getModel()` etc methods in `app/Mage.php`, `Mage_Core_Model_Config`, `Mage_Core_Model_Layout`, etc classes. I did change the ordering of some of those methods, so please use split diff view.

Some notable changes:

`Mage::helper('foo')` now returns `false` instead of throwing an error. This is technically a breaking change, but in reality it's okay because you'd just get an error on the next line of code where you try to call some method on that helper. This is in line with all of the other methods like `Mage::getModel()` which can return false.

I removed `Mage_Core_Model_Config::getModuleSetup()` method. It was from a really old Magento version and not applicable anymore.

I removed some empty string default params, so `Mage::getModel($modelClass = '', $arguments = [])` is now `Mage::getModel($modelClass, $arguments = [])`. Because calling getModel with no value just returns false.

I added some `@throws` annotations.

### Related Pull Requests
https://github.com/macopedia/phpstan-magento1/pull/6

1. Checkout the branch and composer install. Then run phpstan.

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1165,31 +1165,6 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     }
 
     /**
-     * Get module setup class instance.
-     *
-     * Defaults to Mage_Core_Setup
-     *
-     * @param string|Mage_Core_Model_Config_Element $module
-     * @return object
-     */
-    public function getModuleSetup($module = '')
-    {
-        $className = 'Mage_Core_Setup';
-        if ($module !== '') {
-            if (is_string($module)) {
-                $module = $this->getModuleConfig($module);
-            }
-            if (isset($module->setup)) {
-                $moduleClassName = $module->setup->getClassName();
-                if (!empty($moduleClassName)) {
-                    $className = $moduleClassName;
-                }
-            }
-        }
-        return new $className($module);
-    }
-
-    /**
      * Get base filesystem directory. depends on $type
      *
      * If $moduleName is specified retrieves specific value for the module.
@@ -1332,20 +1307,48 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     }
 
     /**
+     * Retrieve class name from config.xml node
+     */
+    public function getNodeClassName(string $path): string
+    {
+        $config = Mage::getConfig()->getNode($path);
+        if (!$config) {
+            return false;
+        }
+        return $config->getClassName();
+    }
+
+    /**
+     * Retrieve class instance from config.xml node
+     *
+     * @param string $path
+     * @return Mage_Core_Model_Abstract|false
+     */
+    public function getNodeClassInstance($path)
+    {
+        $className = $this->getNodeClassName($path);
+        if ($className === false || !class_exists($className)) {
+            return false;
+        }
+        // phpcs:ignore Ecg.Classes.ObjectInstantiation.DirectInstantiation
+        return new $className();
+    }
+
+    /**
      * Retrieve class name by class group
      *
      * @param   string $groupType currently supported model, block, helper
-     * @param   string $classId slash separated class identifier, ex. group/class
+     * @param   string $classAlias slash separated class identifier, ex. group/class
      * @param   string $groupRootNode optional config path for group config
      * @return  string
      */
-    public function getGroupedClassName($groupType, $classId, $groupRootNode = null)
+    public function getGroupedClassName($groupType, $classAlias, $groupRootNode = null)
     {
         if (empty($groupRootNode)) {
             $groupRootNode = 'global/' . $groupType . 's';
         }
 
-        $classArr = explode('/', trim($classId));
+        $classArr = explode('/', trim($classAlias));
         $group = $classArr[0];
         $class = !empty($classArr[1]) ? $classArr[1] : null;
 
@@ -1397,8 +1400,8 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
     /**
      * Retrieve block class name
      *
-     * @param   string $blockType
-     * @return  string
+     * @param string $blockType
+     * @return string
      */
     public function getBlockClassName($blockType)
     {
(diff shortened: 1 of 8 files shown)


justinbeaty:
> > * I removed some empty string default params, so `Mage::getModel($modelClass = '', $arguments = [])` is now `Mage::getModel($modelClass, $arguments = [])`. Because calling getModel with no value just returns false.
> 
> Please not on plublic methods.
> 
> https://onlinephp.io/c/d4544

You have it backwards. It's okay to remove the empty string default value on the base class:

https://onlinephp.io/c/c0165

sreichel:
> * I removed `Mage_Core_Model_Config::getModuleSetup()` method. It was from a really old Magento version and not applicable anymore.

Own PR? Should be merged fast.

sreichel:
> You have it backwards. It's okay to remove the empty string default value on the base class:
> 
> https://onlinephp.io/c/c0165

"Backwards"? A is Mage, B my module, that will not work with that change.

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -1411,109 +1416,145 @@ public function getBlockClassName($blockType)
...
+        // phpcs:ignore Ecg.Classes.ObjectInstantiation.DirectInstantiation
+        $obj = new $className($constructArguments);
+        Varien_Profiler::stop('CORE::create_object_of::' . $className);
+        return $obj;
+    }
+
+    /**
+     * Retrieve resource helper class name
+     */
+    public function getResourceHelperClassName(string $moduleAlias): string|false
```
sreichel, line 1521: "string|false" union type is php8? Possible to change return type for new methods to null?

justinbeaty: I thought I caught all of those. Put in `@return` annotation. Best to leave it as `string|false` to match the other `getXxxClassName` methods.

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -1331,21 +1306,51 @@ public function getPathVars($args = null)
...
+        return $config->getClassName();
+    }
+
+    /**
+     * Retrieve class instance from config.xml node
+     *
+     * @param string $path
+     * @return Mage_Core_Model_Abstract|false
+     */
+    public function getNodeClassInstance($path)
```
sreichel, line 1329:
```suggestion
    /**
     * Retrieve class instance from config.xml node
     *
     * @return Mage_Core_Model_Abstract|false
     */
    public function getNodeClassInstance(string $path)
```

justinbeaty: This is actually not a new method. I moved it up a bit in the class definition since it wasn't relevant to the other code surrounding it.

sreichel: Mhh, should not do reviews online. Thanks

sreichel:
Mhh, no checks on draft PR?

app/code/core/Mage/Core/Model/Config.php
sreichel: HAHA. i did not knew it :) Why isnt this default?

> I debated not moving the methods at all to make review easier, but the random ordering put in by Mage devs just made me mad..

Yep. There are rules for phpcs-fixer that re-order methods .... coming next year ;)

sreichel:
> This is test to use a new PHPStan plugin with OM **instead of radically changing the macopedia one**. Presumably you'd fork the module into a new namespace.

@tmotyl s plugin work well for years. Id prefer to follow yor PR there and update it - instead of replacing.

justinbeaty:
> > This is test to use a new PHPStan plugin with OM **instead of radically changing the macopedia one**. Presumably you'd fork the module into a new namespace.
> 
> @tmotyl s plugin work well for years. Id prefer to follow yor PR there and update it - instead of replacing.

I can update the PR there with the latest updates. But I didn’t know how @tmotyl feels about such a big change to the codebase.

The reason I have changes in this PR to Mage, Mage_Core_Model_Config, etc is to enable more features in the PHPStan plugin. So the new plugin won’t work with non-OM codebases. Will this be a problem? If the plugin gets released as a new major version then it’s okay, but going forward?

sreichel:
> to enable more features in the PHPStan plugin. So the new plugin won’t work with non-OM codebases. Will this be a problem?

I have (roughly) testesd your PR at macppedia plugin and i tooked good.

The changes made here are to "enable more features for the plugin" or the "plugin won’t work with non-OM codebases"?

sreichel:
> If the plugin gets released as a new major version then it’s okay, but going forward?

Going forward? dont know. ists on @tmotyl. Maybe moving it to OM namespace?

sreichel:
@justinbeaty we still need support for php7.4 :(

justinbeaty:
Hm, I found an error. Go to Admin Dashboard > Promotions > Catalog Price Rules and  edit any rule. You get error: `Invalid block type: rule/conditions`.

This is because of an `instanceof Mage_Core_Block_Abstract` check but we the block doesn't extend it:
```php
class Mage_Rule_Block_Conditions implements Varien_Data_Form_Element_Renderer_Interface
{
```

We can revert the behavior with the following patch, but the phpdoc is wrong because it doesn't always return `Mage_Core_Block_Abstract`.


```diff
diff --git a/app/code/core/Mage/Core/Model/Layout.php b/app/code/core/Mage/Core/Model/Layout.php
index fe27d08466..1acfa2ea6e 100644
--- a/app/code/core/Mage/Core/Model/Layout.php
+++ b/app/code/core/Mage/Core/Model/Layout.php
@@ -590,18 +590,25 @@ class Mage_Core_Model_Layout extends Varien_Simplexml_Config
     /**
      * @param string $type
      * @return Mage_Core_Block_Abstract
      * @throws Mage_Core_Exception
      */
     public function getBlockSingleton($type)
     {
         if (!isset($this->_helpers[$type])) {
-            $helper = $this->_getBlockInstance($type);
-            $helper->setLayout($this);
+            $className = Mage::getConfig()->getBlockClassName($type);
+            if ($className === false || !class_exists($className)) {
+                Mage::throwException(Mage::helper('core')->__('Invalid block type: %s', $type));
+            }
+            // phpcs:ignore Ecg.Classes.ObjectInstantiation.DirectInstantiation
+            $helper = new $className();
+            if ($helper instanceof Mage_Core_Block_Abstract) {
+                $helper->setLayout($this);
+            }
             $this->_helpers[$type] = $helper;
         }
         return $this->_helpers[$type];
     }
 
     /**
      * Retrieve helper object
      *
```

PR #4415, opened 2024-12-13
Mini DOCBlock update

sreichel:


sreichel:
Fixes some issues found during work on magerun v3 release.

https://github.com/netz98/n98-magerun/pull/1484

sreichel:
Ignore CS, fixed in https://github.com/OpenMage/magento-lts/pull/4425

app/code/core/Mage/Catalog/Model/Category.php
```diff
@@ -45,10 +45,10 @@
...
  * @method bool getInitialSetupFlag()
  * @method $this setInitialSetupFlag(bool $value)
- * @method bool getIsActive()
- * @method $this setIsActive(bool $value)
- * @method bool getIsAnchor()
- * @method $this setIsAnchor(bool $value)
+ * @method int getIsActive()
+ * @method $this setIsActive(int $value)
+ * @method int getIsAnchor()
+ * @method $this setIsAnchor(int $value)
```
colinmollenhour, line 51: Are there going to be negative consequences of putting the return type we wish we had instead of the one we actually have? The return type is actually likely to be string "0|1" because of the horrible decision of someone long ago to make PDO return strings for all data types. But it could also be bool or int because of inconsistencies in the use of the setter. That is:

```php
$category = Mage::getModel('catalog/category')->load(1);
var_dump($category->getIsActive()); // string "0|1"
$category->setIsActive(true);
var_dump($category->getIsActive()); // bool true
```

sreichel: In the past, when i added method annotions i took the values i have found. E.g. bool, but in PDO its either string or int.  bool is wrong for DB values .

magerun uses (int) constants for active/inactive, thats looks  more correct to me. 

Id like to continue https://github.com/OpenMage/magento-lts/pull/4376 and add getter/setter methods to reflect the correct type ... later.

sreichel:
> because of the horrible decision of someone long ago to make PDO return strings for all data types

... until mysql8.

Thats one reason for https://github.com/OpenMage/magento-lts/pull/4376

app/code/core/Mage/Catalog/Model/Category.php
colinmollenhour: Sure, just something to be thinking about. :)

colinmollenhour (approved):

addison74 (approved):

PR #4458, opened 2025-01-06
Cleanup: removed js/jscolor

sreichel:
### Description (*)

jscolor.js is not used.


addison74:
In the past, someone else suggested deleting this file and I pointed out that it is used in extensions like MageWorx Advanced Product Options.

sreichel:
Should we really keep files that are not used?

- you can still grab that files from a previous release
- if requested often (i doubt) we can add a repo (please not)

With next composer-plugin [update ](https://github.com/OpenMage/composer-plugin/commit/67ea3dc84c8599f3e1ada15d8a4ac45c1356704d#diff-9a87c9809a5da5928c343193bc82c48f7b037f58c2851cc07912ded82cb549edR94)it it possible to add custom fiiles from NPM/unpkg.com.

Cou can add this to composer extra ...

```
        "openmage-unpkg-packages": {
            "@eastdesire/jscolor": {
                "version": "2.5.2",
                "source": "",
                "target": "js/jscolor",
                "files": [
                    "jscolor.js"
                ]
            }
        }
```

### Note:

This is just an example. Current version of jscolor is 1.4.5. That's so old and not even avalaible at https://unpkg.com/ or https://www.npmjs.com.

Lets remove it. We can still help to update that extension.

addison74 (approved):

kiatng (approved):

PR #4467, opened 2025-01-09
Shorter copyright header

sreichel:
### Description (*)

No need to mention license again.

- closes https://github.com/OpenMage/magento-lts/discussions/3137


Axepih:
Hi,

What do you think of what piwigo has done?
https://github.com/Piwigo/Piwigo/commit/fc193f7474afe133a7692b479d8c6ce6717930c1

The copyright is available in a single file COPYING.txt 
https://github.com/Piwigo/Piwigo/blob/master/COPYING.txt

sreichel:
I think we should keep at in files for now.

Put it into a single file should be discussed in detail.

api.php
```diff
@@ -1,16 +1,10 @@
...
- *
- * This source file is subject to the Open Software License (OSL 3.0)
- * that is bundled with this package in the file LICENSE.txt.
- * It is also available at https://opensource.org/license/osl-3-0-php
- *
  * @category   Mage
  * @package    Mage
  * @copyright  Copyright (c) 2006-2020 Magento, Inc. (https://www.magento.com)
- * @copyright  Copyright (c) 2021-2024 The OpenMage Contributors (https://www.openmage.org)
+ * @copyright  Copyright (c) 2021-2025 The OpenMage Contributors (https://www.openmage.org)
```
fballiano, line 7: these lines (bumping the year) should be avoided in the context of this PR

colinmollenhour:
I say remove the date entirely as the git repo provides ample evidence of when a file was created or last modified if anyone ever needed any proof. In the US, Copyright expiration is 70 years, so I just don't see this being an issue. The code is open-source, so I'm not even sure what purpose it serves us adding a Copyright so keep it as simple as possible.
I also like the Piwigo approach suggested by @Axepih.

api.php
fballiano: The comments I make are for the good of this project. Also, you blocked me and unblocked me so, if you don't want my comments why did you unblock me.

sreichel:
@colinmollenhour we can follow @Axepih suggestion, but please do it in a another PR. 

Please close or merge.

colinmollenhour:
I'm thinking about it this way: If you wanted to use part of a 100-year old book because the 70 year copyright expired, but the author and the author's family just kept refreshing the copyright date every year claiming that it was as fresh as the day it was written, would this be at all defensible in court? I don't think so, but updating individual files that aren't receiving changes seems similar to this analogy in my opinion.

Here is a proposed `COPYING.txt` file:

```
COPYRIGHT

    OpenMage - an eCommerce platform

    Copyright © 2006-2020 Magento, Inc. (https://magento.com)
    Copyright © 2017-2025 The OpenMage Contributors (https://openmage.org)
    Copyright © 2024-2025 Maho (https://mahocommerce.com)

SOFTWARE LICENSE

    The free software license of OpenMage themes (located in app/design/) and skins (located in skin/) is the
    Academic Free License (AFL 3.0) that is bundled with this package in the file LICENSE_AFL.txt.
    It is also available at https://opensource.org/license/afl-3-0-php

    The free software license of TinyMCE that may be bundled and used exclusively with this project is a proprietary license
    that is bundled with this package in the file LICENSE_TINYMCE.txt.
    It is also available at https://github.com/OpenMage/magento-lts/blob/main/LICENSE_TINYMCE.txt

    The free software license of all other OpenMage code included with this package unless otherwise noted in the source file
    or source file directory is The Open Software License 3.0 (OSL 3.0) that is bundled with this package in the file LICENSE.txt.
    It is also available at https://opensource.org/license/osl-3.0.php    
```

And an example header block:
```
/*
This file is part of OpenMage.
For copyright and license information, please view the COPYING.txt file that was distributed with this source code.
*/
```

I don't know what the benefit of using the `@license` annotation is, but it could be adapted to that if needed.

sreichel:
Updated as suggested.

addison74:
I would suggest
```
/*
This code protected. For more information related to the copyright and 
license information please read the COPYING.txt file
*/
```

I asked ChatGPT to provide a short text and here is the first result

```
The copyright and license information has been moved to a separate file named COPYING.txt. Please refer to this file for the full details.
```

sreichel:
@addison74 can you please create a PR to my branch to change wording?

addison74:
At this point I am not firmly convinced what the most appropriate wording would be. It is an important change and requires more time for analysis. I ask others to give their opinion too.

colinmollenhour:
> At this point I am not firmly convinced what the most appropriate wording would be. It is an important change and requires more time for analysis. I ask others to give their opinion too.

To avoid getting stuck, we will need more specific concerns or alternatives. You offered alternative wording for the sentences, but they seem basically the same to me (aside from the grammatical error in the first one 😉)? I'm not emotionally attached to the text I proposed; as long as it gets the job done, it's fine by me. However, Sven already made the changes and AFAIK there are no concerns that have not been addressed.

I added a comment to the original discussion so anyone watching can have a say. However, to me it seems this has been discussed quite a bit in different places over the years, so let's get er done!.

fballiano:
Hi @colinmollenhour 

>     Copyright © 2006-2020 Magento, Inc. (https://magento.com)
>     Copyright © 2017-2025 The OpenMage Contributors (https://openmage.org)
>     Copyright © 2024-2025 Maho (https://mahocommerce.com)

thanks for thinking about us, since our project has many unique parts that can't be labeled as "copyright by openmage", this version, at a first glance, can't probably work for us, but that will be something we'll have to tackle later internally, you'll rightfully decide what you feel it's best for your project.

justinbeaty:
I very much support that `COPYING.txt` file. 👍 

For PHPDoc I'll share some things I found when playing with the official phpdoc package to build the documentation.

The first docblock in a file is the file level docblock. This isn't used correctly in our codebases, I'm referring to these lines:

https://github.com/OpenMage/magento-lts/blob/7e5a1414b77f4f1cf4e0b61f91dd476952020073/app/code/core/Mage/Core/Model/App.php#L4-L9

When you build that into the static documentation website, there's an index page you can navigate to that lists every PHP file in alphabetical order. If it's not bad enough that we have dozens of `Abstract.php`, `Data.php`, `Grid.php`, and other files with the same name, according to phpdoc they all contain "OpenMage". So you can't know if you're clicking the right `IndexController.php` or whatever other file you're looking for quickly.

Instead of "OpenMage", the title of the file level docblock should be a short description of what's in that file. The long description (two new lines after title) can optionally contain more information.

All phpdoc tags present in the file level docblock apply to everything else defined in that file. So tags like `@category`, `@package`, and `@copyright` defined on the file docblock are copied to all class and other docblocks in that file. Thus, the duplicated `@category` and `@package` tags in many of our files are unnecessary, you only need duplicates if you want to override the value from the file docblock somewhere else.

In my opinion, file level docblocks are useful for only these files in our repo: phtml files, install scripts, `app/code/core/Mage/Core/functions.php`, `app/bootstrap.php`, `app/Mage.php`, and maybe a few others. That's because those files don't contain a singular PHP class definition, instead a file level docblock is needed to cover every variable, function, or piece of top level code defined in it.

Most of our PHP files do contain just a single class definition, and it's sort of confusing having two top level docblocks. Nobody could explain it which is why I installed the official package. I am away from my computer at the moment but I'm fairly certain dropping the file level docblock and keeping only the class definition docblock didn't cause any problems. It just meant the file had no file level meta information, but that's not any worse than having the file's description be "OpenMage". I can't imagine anyone wanting to write both a file description and a class description, or how they would differ, so I don't see why we need both.

So whatever the first docblock is, it would contain the `@copyright` tag pointing to `COPYING.txt`. My rationale for removing it from the file level docblock and adding it to the class docblock is since there's only one thing in the file, the PHP class, having the `@copyright` on the class docblock covers the code under copyright. *There is no other code in that file to be copyrighted.*

A PHP file can contain multiple classes each with a different `@copyright` tag, perhaps because the code was pulled from a project with a compatible license. But these are all edge cases, and having the file docblock makes sense in a scenario like this so you can explain what the heck you're doing inlining a 3rd party library into your class that extends `Mage_Adminhtml_Block_` , or whatever else you're doing...

Of course we can use a script to convert everything. However some Magento class files have really bad phpdoc titles and/or descriptions. Most are basically just repeating the class name. Rewriting those isn't necessary, but it's good to keep new ones descriptive and over time fix the old.

A Few Extra Thoughts:
1. I would keep the `@license` tag in each file, even if it just the license name without the full URL. I'm imagining a scenario where a file from `app/code/Mage/*` or `lib/*` is distributed without it being clear to the recipient where in the repository the file came from, and thus how `COPYING.txt` applies to it.
2. Is 2006-2020 the current lowest/highest year for copyright Magento, Inc.? Were changes from M2 backported at a later date? Any files that have or should have a copyright Adobe?

colinmollenhour:
Ahh, I thought or assumed there had already been some degree of code sharing in both directions and adding it to COPYING.txt would simplify things rather than (or in addition to) adding a `@copyright` line to files that received patches from Maho.

I wasn't making an assumption that Maho would adopt the same approach for file headers. Ultimately, we're using the same license which is FOSS, so I don't think it makes any difference to the end user whether it's in individual files or the main file, and it would mostly be a matter of attribution. Personally, I'd rather have a higher level of acknowledgment of sharing between the projects and a more liberal amount of sharing as a result.

To address @justinbeaty's comment that just arrived, thanks for detailing how PHPDoc handles the doc blocks. It seems to me that these days with smart IDEs that there is nearly zero value to generating a giant website or document detailing all of the classes and methods and these doc blocks are mostly useful for navigating within the IDE. Therefore, it is best to keep the "copying" info out of the main docblocks so that the coder isn't having to read them all the time. For example, in the highlighted section I would not want to see "This file is a part of OpenMage..." so I think keeping separate blocks is best regardless of being "correct" with respect to PHPDoc.
image

> I would keep the @license tag in each file...
I sort-of agree, but if we don't care about generating a website from the doc blocks and only care about the IDE, from a legal perspective, it doesn't make a difference. I.e. "the file didn't have a @license annotation so I figured it was public domain" is not a valid defense when there is a line right there that says "For copyright and license information, please..." which also should make it obvious that it's not ok to copy the code without the mentioned COPYING.txt file. That said, I'm not opposed to adding it back, is that hard to do @sreichel? I'm not sure what method you're using to make all of these changes. 

> Is 2006-2020 the current lowest/highest year for copyright Magento, Inc.?
To my knowledge, there was never a M1 release with any other copyright headers, so I don't see that we have any obligation to add this for them retrospectively. Can you even download original M1 code from an official source anymore?

justinbeaty:
I would agree IDEs are more important than a static documentation website, but I do like both. :smiley:

I'm back at my computer now, so to be clear here's an example, but I don't have a current PHPStorm license to check how it shows.

```php
<?php
/**
 * Base html block
 *
 * Example usage, etc...
 * 
 * @category   Mage
 * @package    Mage_Core
 * @copyright  See COPYING.txt
 * @license    https://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
 *
 * @method $this setContentHeading(string $value)
 * @method $this setDestElementId(string $value)
 * @method $this setFormAction(string $value)
 * @method $this setIdSuffix(string $value)
 * @method $this setProduct(Mage_Catalog_Model_Product $value)
 * @method $this setDisplayMinimalPrice(bool $value)
 */
class Mage_Core_Block_Template extends Mage_Core_Block_Abstract
{
```

colinmollenhour:
Thanks for the example, that looks great to me, a nice improvement. If you agree, do you mind applying that method to class files @sreichel ? No changes needed for all other files (phtml, css, etc).

justinbeaty:
I think there is room for improvement on the phtml files too, for example:

https://github.com/OpenMage/magento-lts/blob/e41e8f2ea42a3376c7264f0c46c57be87092d88e/app/design/adminhtml/default/default/template/eav/attribute/options.phtml#L1-L23

It could be something like:
```php
<?php
/**
 * Attribute manage options form
 *
 * This template is responsible for rendering the form at Catalog > Attributes > Manage Attributes > Manage Labels / Options tab.
 * You can customize it by... (etc)
 * 
 * @category    design
 * @package     default_default
 * @copyright   This file is part of OpenMage[...]
 * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
 */

/** @var Mage_Eav_Block_Adminhtml_Attribute_Edit_Options_Abstract $this */
?>
<div>
<ul class="messages">
    <li class="notice-msg">
<!-- ... -->
```

I'm not tied to any format specific format or wording either. Whatever is decided we should test on various IDEs. It would be nice to make this the final docblock update.

justinbeaty:
And my memory could be hazy, so I'll confirm my findings before @sreichel were to change anything in this PR. I would like to hear other people's opinions too. 😃

sreichel:
> I'm not opposed to adding it back, is that hard to do @sreichel? I'm not sure what method you're using to make all of these changes.

Really often I use "my time".

The PR was about to remove some "duplicate" content.

Not touching the years. Not touching the copyrights ...

justinbeaty:
I had a chance to double check things with the official phpdoc package. Regardless of the outcome of this PR, here are some more notes:

1\. The `@category` tag is deprecated:
> This was necessary since the [@package](https://docs.phpdoc.org/guide/references/phpdoc/tags/package.html) tag, as defined in the original Standard, did not allow for more than one hierarchy level. Since this has changed this tag SHOULD NOT be used.
> Source: https://docs.phpdoc.org/guide/references/phpdoc/tags/category.html

2\. Just the `@package` tag is needed, and the values we already have for those tags are fine. They *could* be more specific, i.e. `Mage_Catalog_Model` instead of just `Mage_Catalog`, but it feels unnecessary for now.

3\. Using the `@package (base|rwd|default)_default` tag in phtml files seems pointless to me. Template files don't define any classes, functions, or global constants, so there's nothing to associate into the namespace. Instead the documentation site shows three namespaces (`base_default`, `rwd_default`, and `default_default`) but they all just show an empty table of contents. If anything, I believe the `@package` value should refer to the module that the template file belongs to, so `@package Mage_Page` for the file `template/page/2columns-left.phtml`. 
<details><summary>Expand screenshot</summary>
<p>

</p>
</details> 

---

## Updated docblock template proposal

Not a final docblock template by any means, just updating the previous post to remove `@category` and from all docblocks, and additionally remove `@package` from phtml files.

```php
# app/code/core/Mage/Core/Block/Template.php
<?php 

/**
 * Base html block
 *
 * Example usage, etc. Can use markdown:
 * ```php
 * $filename = Mage::getDesign()->getTemplateFilename($this->getTemplate(), $params);
 * ```
 *
 * @package    Mage_Core
 * @copyright  This file is part of OpenMage. For copyright and license information, please view the COPYING.txt file that was distributed with this source code.
 * @license    Open Software License (OSL 3.0)
 *
 * @method $this setContentHeading(string $value)
 * @method $this setDestElementId(string $value)
 * @method $this setFormAction(string $value)
 * @method $this setIdSuffix(string $value)
 * @method $this setProduct(Mage_Catalog_Model_Product $value)
 * @method $this setDisplayMinimalPrice(bool $value)
 */
class Mage_Core_Block_Template extends Mage_Core_Block_Abstract
```

```php
# app/design/adminhtml/default/default/template/eav/attribute/options.phtml
<?php

/**
 * Attribute manage options form
 *
 * This template is responsible for rendering the form at Catalog > Attributes > Manage Attributes > Manage Labels / Options tab.
 * You can customize it by... (etc)
 *
 * @copyright   This file is part of OpenMage. For copyright and license information, please view the COPYING.txt file that was distributed with this source code.
 * @license     Academic Free License (AFL 3.0)
 */

/** @var Mage_Eav_Block_Adminhtml_Attribute_Edit_Options_Abstract $this */
?>
<div>
<ul class="messages">
```

---


## Screenshots

Here's some screenshots of how each option (current main branch, this PR, and my proposed docblock) render.

1\. Class: `Mage_Core_Block_Template`

This is the page that shows the *Class Definition* of `Mage_Core_Block_Template`. Both the current and this PR render the same, so I only show two screenshots.

The current screenshot shows a useless "category: Mage" tag when we already know we're in the Mage package. There is also no copyright or license information. Unrelated, but I added some markdown to that file in the proposed version just to show what's possible.

<details><summary>Expand screenshots</summary>
<p>


</p>
</details> 

2\. File: <code>app/code/core/Mage/Core/Block/Template.php</code>
This is the page that shows the *File Level* metadata in the file where `Mage_Core_Block_Template` is defined. Personally I think the proposed screenshot looks the best, and IMO there's no need to show copyright info here, because the `@copyright` is attached to the class.

<details><summary>Expand screenshots</summary>
<p>



</p>
</details> 

3\. File: <code>app/design/adminhtml/default/default/template/eav/attribute/options.phtml</code>
This is just a template file, but both current and this PR screenshots lack any information about the file, even though there is this docblock, because that second docblock doesn't attach to anything, except potentially a variable defined after it. 
```php
<?php
/**
 * OpenMage
 *
 * This source file is subject to the Academic Free License (AFL 3.0)
 * that is bundled with this package in the file LICENSE_AFL.txt.
 * It is also available at https://opensource.org/license/afl-3-0-php
 *
 * @category    design
 * @package     default_default
 * @copyright   Copyright (c) 2021-2024 The OpenMage Contributors (https://www.openmage.org)
 * @license     https://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
 */
?>
<?php
/**
 * Attribute options control
 *
 * @see Mage_Eav_Block_Adminhtml_Attribute_Edit_Options_Abstract
 * @var Mage_Eav_Block_Adminhtml_Attribute_Edit_Options_Abstract $this
 */
?>
```
<details><summary>Expand screenshots</summary>
<p>



</p>
</details>

addison74:
I support moving most of the copyright into a root file. Just a small comment related to this line 

"@copyright  This file is part of OpenMage. For copyright and license information, please view the COPYING.txt file that was distributed with this source code."

If it is required make it smaller, for example

"@copyright  For copyright and license information please read the COPYING.txt file."

colinmollenhour:
I like all the proposed changes by @justinbeaty and @addison74.

Sorry if I hijacked the original PR purpose, @sreichel, but I think it was only after you added the workflow to update the copyright years so I didn't think that was outside of the scope. I was just asking about what method you were using in case you had a smarter method than mine so I could help. I'd probably just use a the obvious "Find and Replace" personally, but I'm sure there are other more clever ways.. I'll make the changes if you prefer me to do it, just let me know.

justinbeaty:
@colinmollenhour have you tried those updated docblocks in your IDE? Presumably you only see the class docblocks and not the file level docblocks anywhere, so just want to make sure those look good on your end.

I should also check if putting the `@var ... $this`  and any `@see` tags in the file level docblock works. It is a bit cleaner, but I don't know what others think about it. For example:
```php
<?php

/**
 * Adminhtml manage attribute options form block
 *
 * Long description...
 *
 * @var         Mage_Eav_Block_Adminhtml_Attribute_Edit_Options_Abstract $this  <-- This line
 * @see         Some value...
 * @copyright   For copyright and license information please read the COPYING.txt file
 * @license     Academic Free License (AFL 3.0)
 */
?>
```

colinmollenhour:
I think for `@var` it should be in its own comment for best IDE compatibility. Are there other IDEs besides PhpStorm and VS Code (and plugins/variants) that are relevant these days?

sreichel:
@justinbeaty @colinmollenhour @addison74 thanks for suggestions.

To reproduce output i've added phpdocumentor.

To not run it run for whole code you can add this to root.

Takes 10s to have docs at `http://localhost:63342/magento-lts/build/phpdoc-test/`

<details><summary>phpdoc.xml</summary>

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<phpdocumentor
        configVersion="3"
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xmlns="https://www.phpdoc.org"
        xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/phpDocumentor/phpDocumentor/master/data/xsd/phpdoc.xsd"
>
    <setting name="template.color" value="blue"/>

    <paths>
        <output>build/phpdoc-test/api</output>
        <cache>build/phpdoc-test/cache</cache>
    </paths>

    <version number="3.0.0">
        <api>
            <source dsn=".">
                <path>app/code/core/Mage/Admin</path>
                <path>COPYING.txt</path>
            </source>
            <extensions>
                <extension>php</extension>
                <extension>txt</extension>
            </extensions>
            <visibility>public</visibility>
            <visibility>protected</visibility>
            <visibility>private</visibility>
        </api>
    </version>
</phpdocumentor>

```
</details> 

Run

```bash
vendor/bin/phpdoc run
```

---

From my tests ...

- phpDocumentor does not render class-blocks correctly when file-block exists. So it makes sense to combine it, like https://github.com/OpenMage/magento-lts/pull/4467#issuecomment-2641508674.
- category-tag can be removed

```

For php-files we have this atm? Added link ...

/**
 * Class description
 *
 * @copyright   For copyright and license information read COPYING.txt file
 * @copyright   ... additional info ....
 * @link        /COPYING.txt
 * @package     Mage_Xml
 * 
 * ... methods, etc ...
 */
class Some_Class
```

(note: slash is requird to linkt to copying.txt)

@colinmollenhour i am happy for every improvent, but its bit wiered ... there are other PRs that deserve more attention. 

I have used simple "find&replace", later with regex pattern. Now i think about a script to automate all suggestion ... lol. "Problem" is, it takes time to process 5k files (again and again). Espaccially running on old hardware (core-i3, 16gb :cry:)

> I'll make the changes if you prefer me to do it, just let me know.

As for every PR, pleae fork and submit a PR to improve.

PS: can you please prepare a subdomain to publish phpdocumentor?

colinmollenhour:
Yeah, as often happens, a small suggestion becomes a bigger one... But I think we arrived at a good solution.

What subdomain do you want? phpdoc.openmage.org?

sreichel:
I looks good, but how to do it all files?

How to move class descripion on top, put methods below copright?

You are fine with regex, any ideas for complex replace pattern?

phpdoc.openmage.org is good :+1:

sreichel:
> thanks for thinking about us, since our project has many unique parts that can't be labeled as "copyright by openmage"

Whats sooo unique?

- you did a f*** great job with replacing prototype.js
- justin made some good changes to autoloader
- ...
- ... but majoity of your commits comes from here

Why not to put that energy into a "RWD v2" or "Maho"-Theme? Why add CLI commands instead of supporting n98?

sreichel:
> thanks for thinking about us, since our project has many unique parts that can't be labeled as "copyright by openmage"

I'd like to add the Maho-theme as addition to RWD., so we need to place copyrights in phtmls too.

colinmollenhour:
> Whats sooo unique?

@sreichel This is why there are not enough maintainers. 😞

sreichel:
See https://github.com/OpenMage/magento-lts/pull/4624

PR #4478, opened 2025-01-13
PHPUnit: unify to static methods

sreichel:
### Description (*)

Phpunit allows both - dynamic and static method calls. Nothing wrong, but when using "phpstan/phpstan-strict-rules" <s>or "phpstan/phpstan-phpunit"</s> it causes a lot of noise.

Added php-cs-fixer-rule to make phpunits assertions static.


sreichel:
Its only with phpstan-strict-rules.

PR #4337, opened 2024-11-07
Fix duplicate block rendering with getSortedChildren()

justinbeaty:
### Description (*)

If you try to add a block to a `core/text_list`, you'll get unexpected behavior if another block has been defined with the same name or alias. I consider this a bug because it's not that both blocks are rendered, but the second block is rendered twice.

For example, this renders "ÄÄ":
```xml
<block type="core/text_list" name="test">
    <block type="core/text" name="test_a">
        <action method="setText"><text>A</text></action>
    </block>
    <block type="core/text" name="test_a">
        <action method="setText"><text>Ä</text></action>
    </block>
</block>
```

While the first block will be overwritten by `$this->_children[$alias] = $block;`, a duplicate array element will be added to `$this->_sortedChildren`. 

Then when we loop through the sorted children in `Mage_Core_Block_Text_List`, we will encounter the block name `test_a` twice, call `getBlock('test_a')` twice, and run `toHtml()` twice.

https://github.com/OpenMage/magento-lts/blob/ca4f3217b7a0cb8d0b5b00df7c519922f55d99f3/app/code/core/Mage/Core/Block/Text/List.php#L27-L38

You could just not add duplicate block names, but it's actually useful if you want to override a block in a more specific handle without having to add `<action method="unsetChild">`.

I also replaced `unset()` with `array_splice()` in the `unsetChild` method. This seems like a better solution than the solution merged from PR #1108 which I reverted.

### Related Pull Requests
#1108

### Manual testing scenarios (*)
Add the following into your local.xml:

```xml
<?xml version="1.0"?>
<layout version="0.1.0">
    <default>
        <reference name="content">

            <block type="core/text" name="test_1_title">
                <action method="setText"><text><![CDATA[<h5>Test 1</h5>]]></text></action>
            </block>
            <block type="core/text_list" name="test_1">
                <block type="core/text" name="test_1_a">
                    <action method="setText"><text>A</text></action>
                </block>
                <block type="core/text" name="test_1_a">
                    <action method="setText"><text>Ä</text></action>
                </block>
            </block>

            <block type="core/text" name="test_2_title">
                <action method="setText"><text><![CDATA[<h5>Test 2</h5>]]></text></action>
            </block>
            <block type="core/text_list" name="test_2">
                <block type="core/text" name="test_2_a">
                    <action method="setText"><text>A</text></action>
                </block>
                <block type="core/text" name="test_2_b">
                    <action method="setText"><text>B</text></action>
                </block>
[...]

PR #4480, opened 2025-01-14
Fix regression bug for duplicate block rendering with getSortedChildren()

bucha:
### Description (*)

This is a fix for regression bug I noticed while updating one of our projects from OpenMage 20.10.2 to 20.11.0 which I was able to trace back to #4337

Within this project we've basically implemented a holepunch-system in which we run the following code to replace a given block by another block:

```php
$layout->unsetBlock($targetBlockName);
$layout->setBlock($targetBlockName, $replacementBlock);
$parent->setChild($targetBlockAlias, $replacementBlock);
```

Running OpenMage 20.10.2 this approach worked perfectly.
On OpenMage 20.11.0 however we experienced a problem:
One of the blocks we replace using this method is inside a list-block. A list-block will first sort its children and then render them in order to append their html output to its own output. We noticed that the output of the replacement-block was not in the list-block's output any longer. When looking at this phenomenon in an xdebug session I noticed that the list-block contained the replacement-block in its `$_children` but `$_sortedChildren` was empty.

It turns out that the recently introduced call to `$this->unsetChild($alias);` in `Mage_Core_Block_Abstract` was responsible for this behavior:

https://github.com/OpenMage/magento-lts/blame/v20.11.0/app/code/core/Mage/Core/Block/Abstract.php#L468

I have therefor removed the call to `$this->unsetChild($alias);` and instead introduced a check in `insert()` whether a given block is already known in `$_sortedChildren` and if so it will be removed from `$_sortedChildren` (before then being appended again).

This should be in line with @justinbeaty 's original intentions which I also checked against the unit-tests of #4337.

### Related Pull Requests
- #4337
- #1108

I've added another test within the unit tests:

```php
	public function testSortInstructionsAfterReplaceChild()
	{
		$layout = Mage::getModel('core/layout');

		$parentBlock = $layout->createBlock('core/text_list', 'parent');

		$childBlockA = $layout->createBlock('core/text', 'target_block')->setText('A');
		$parentBlock->insert($childBlockA, '', false, 'child');

		$childBlockB = $layout->createBlock('core/text', 'target_block')->setText('B');

		// Replacing the block but keeping its order within the parent
		$layout->unsetBlock('target_block');
		$layout->setBlock('target_block', $childBlockB);
		$parentBlock->setChild('child', $childBlockB);

		$this->assertSame('B', $parentBlock->toHtml());
	}
```


bucha:
Yes it's my first PR

sreichel:
Thank you.

Please run `composer run php-cs-fixer:fix`

justinbeaty:
@bucha Thank you, looking at the code it looks fine since your new check is right before the other sorted children logic. And if the unit tests are passing then it seems all okay (thanks also for fixing that one test.)

sreichel (approved):

addison74 (approved):

sreichel:
@all-contributors add bucha code

PR #4483, opened 2025-01-15
PHP-CS-Fixer: ordered class elements

sreichel:
### Description (*)

I'd relly like to make progress on https://github.com/OpenMage/magento-lts/pull/4411, but @justinbeaty made a change that makes reviews bit complicated.

Methods got re-ordered by its names .

I really like that!

When adding methods/properties i often ask myself where to put it ... 

Easy solution ... automate it.

This PR adds https://cs.symfony.com/doc/rules/class_notation/ordered_class_elements.html.

### Note

It moves underscore-prefixed methods like "_construct" (single underscore) to the bottom ... deal with it :)


justinbeaty:
Even if I was the inspiration for this PR, I'm not sure how I feel about it. Often times method are grouped together by their role, and this would make it harder to read the code just so we can have another rule to be applied.

Also to be clear, the other PR I made I didn't reorder the methods to be in alphabetical order, but instead regrouped them so they logically made sense.

PR #4485, opened 2025-01-15
Drop ZIP-archive support

sreichel:
### Description (*)

Since v20.11.0 there is no zip-archive (including vendor) created. I dont know why - and i dont want to.

composer should be available on the cheapest hostsing you need to run OM. Zip only causes issued ... please drop it.


colinmollenhour (approved):
To my knowledge, nobody has complained that the archive quit working, so it seems there is no reason to keep it.

kiatng (approved):

addison74 (approved):

toejam34:
> To my knowledge, nobody has complained that the archive quit working, so it seems there is no reason to keep it.

I did ! 

Its a lot easier for us Wintel engineers to upload a zip and unzip!

PR #4489, opened 2025-01-15
Updated SonarQube action

sreichel:
> This action is deprecated and will be removed in a future release. Please use the sonarqube-scan-action action instead. The sonarqube-scan-action is a drop-in replacement for this action.

@colinmollenhour there is another waring ...

> Warning: Running this GitHub Action without SONAR_TOKEN is not recommended

Can you please add one? Maybe it resolves missing coverage reports.


colinmollenhour:
There was already a SONAR_TOKEN created 3 years ago, but I just updated it with a token created with my account that has the OM org.

sreichel:
Tested. Its still empty. :question: 

```yml
SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
```

fballiano:
secrets are not visible in all types of workflows: https://github.com/pypa/gh-action-pypi-publish/discussions/49#discussioncomment-162653

sreichel:
Please ignore that token problem for now. 

The sonar-action needs to be updated.

PR #4491, opened 2025-01-16
TinyMCE: potential fix for code scanning alert: Inefficient regular expression

sreichel:
Potential fix for [https://github.com/OpenMage/magento-lts/security/code-scanning/102](https://github.com/OpenMage/magento-lts/security/code-scanning/102)

To fix the problem, we need to modify the regular expression to remove the ambiguity that causes exponential backtracking. We can achieve this by ensuring that the sub-expression inside the repetition does not have multiple ways to match the same string. Specifically, we can replace `(?:\\.|[^'])` with a more precise pattern that avoids ambiguity.

In this case, we can use a negative lookahead to ensure that we match any character except a single quote or a backslash followed by any character. This approach eliminates the ambiguity and prevents exponential backtracking.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._


colinmollenhour:
Perhaps a few negative/positive tests using regex101.com or similar to prove the patch does not break any common cases would be in order and help reviewers not have to do the heavy lifting of figuring out if an AI solution is any good or not.

sreichel:
This affectsto patterns in TinyMce config ,,,,

```
    encodeWidgets: function (content) {
        return content.gsub(/\{\{widget(.*?)\}\}/i, function (match) {
            var attributes = this.parseAttributesString(match[1]);
```

and

```
    decodeWidgets: function (content) {
        return content.gsub(/<img([^>]+id=\"[^>]+)>/i, function (match) {
            var attributes = this.parseAttributesString(match[1]);
```

Either images/widgets still work - or not.

colinmollenhour (approved):
I assume it is still working, so that is great, but is there a possible security implication if it is able to match patterns that it didn't match before? Looks innocuous, I'm just not that familiar with the TinyMCE integration to know what strings are the input and where the output is used. Otherwise looks like a good fix.

addison74 (approved):

PR #4496, opened 2025-01-19
Changed `getOpenMageVersion()`

sreichel:
Tests work.

We dont need to upgrade the verion for every release anymore, but 3rd-party cant get the version info anymore.

In a composer-based procet you can detect the installed version easily. Do we need to have hardcoded version number?


sreichel:
Tests work.

We dont need to upgrade the verion for every release anymore, but 3rd-party cant get the version info anymore.

In a composer-based procet you can detect the installed version easily. Do we need to have hardcoded version number?

app/Mage.php
```diff
@@ -178,64 +180,65 @@ public static function getVersionInfo()
...
-        $versionString = "{$info['major']}.{$info['minor']}.{$info['patch']}";
-
-        if ($info['stability'] && $info['number']) {
-            return "{$versionString}-{$info['stability']}.{$info['number']}";
-        }
-        if ($info['stability']) {
-            return "{$versionString}-{$info['stability']}";
+        $versionfromGit = self::getVersionFromGit($withHas);
+        if ($versionfromGit) {
+            return $versionfromGit;
```
Hanmac, line 187: Wouldn't that collide with other repos that uses openMage?

if my repo has tags, then it would return my Repo tags instead of the OpenMage tags

sreichel: Missed to push last commits. Check for composer version has to come first. 

Will update PR tomorrow.

sreichel: We have to ways to install ...

- git clone ... gibt branch + hash is shown
- composer ... branch is shown
- having a fork with own tags ... your tag will be shown (you know what you have forked)

PR #4500, opened 2025-01-21
Fixed null deprecation in Mage_Eav_Model_Attribute_Data_Text

kiatng:
### Description (*)
```
    [type] => 8192:E_DEPRECATED
    [message] => trim(): Passing null to parameter #1 ($string) of type string is deprecated
    [file] => .../app/code/core/Mage/Eav/Model/Attribute/Data/Text.php
    [line] => 62
    [uri] => /checkout/onepage/saveBilling/
```

### DIFF
--- a/app/code/core/Mage/Eav/Model/Attribute/Data/Text.php
+++ b/app/code/core/Mage/Eav/Model/Attribute/Data/Text.php
@@ -37,7 +37,7 @@ class Mage_Eav_Model_Attribute_Data_Text extends Mage_Eav_Model_Attribute_Data_A
      * Validate data
      * Return true or array of errors
      *
-     * @param array|string $value
+     * @param string|bool|null $value
      * @return bool|array
      */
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Eav/Model/Attribute/Data/Text.php
```diff
@@ -51,7 +51,7 @@ public function validateValue($value)
             $value = $this->getEntity()->getDataUsingMethod($attribute->getAttributeCode());
         }
 
-        if ($attribute->getIsRequired() && strlen($value) == 0) {
+        if ($attribute->getIsRequired() && strlen((string) $value) === 0) {
```
sreichel, line 54:
```suggestion
        if ($attribute->getIsRequired() && (string) $value === '') {
```

sreichel (approved):

addison74 (approved):

Issue #4510, opened 2025-01-27
[Api2] [Catalog] Product URL returned by REST Api is wrong store


Hanmac:
### Preconditions (*)

1. 19.4.15 to 20.12+
2. Mage_Api2 enabled

### Steps to reproduce (*)

1. Enable Guest in REST API2 and the Attribute that they can receive Product URL
2. 

### Expected result (*)

1. [Screenshots, logs or description]
2. Correct URL

### Actual result (*)

1. [Screenshots, logs or description]
2. Wrong Store

My Test URL
`https://myserver/api/rest/products/sku?store=5`

The JSON output:
```json
{
  "entity_id": 1343,
  "type_id": "simple",
  "url": "https:\/\/myserver\/index.php\/admin\/catalog\/product\/view\/id\/1343\/"
}
```

see that it is returning `admin` because it ignores the store parameter or `getDefaultStoreView`
(it even ignores the `adminhtml` router in `local.xml`, but that is a different thing)

The problem is this piece:

https://github.com/OpenMage/magento-lts/blob/fc9a8ca634323663f6f585727f28ac9ee9611f72/app/code/core/Mage/Catalog/Model/Api2/Product/Rest.php#L121

a few lines above, the WebsiteId is set from the store I used as parameter: 
https://github.com/OpenMage/magento-lts/blob/fc9a8ca634323663f6f585727f28ac9ee9611f72/app/code/core/Mage/Catalog/Model/Api2/Product/Rest.php#L106

changing it to this solves the issue:
```php
$productData['url'] = $product->getProductUrl();
```

@sreichel if you want, I can make an MR soon.

PR #4511, opened 2025-01-27
Api2: Fixes getProductUrl

Hanmac:
### Description (*)

This fixes REST API2 Product `url` to use the `store` parameter.

### Related Pull Requests

- fixes OpenMage/magento-lts#4510

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Api2/Product/Rest.php
+++ b/app/code/core/Mage/Catalog/Model/Api2/Product/Rest.php
@@ -100,8 +100,6 @@ abstract class Mage_Catalog_Model_Api2_Product_Rest extends Mage_Catalog_Model_A
      */
     protected function _prepareProductForResponse(Mage_Catalog_Model_Product $product)
     {
-        /** @var Mage_Catalog_Helper_Product $productHelper */
-        $productHelper = Mage::helper('catalog/product');
         $productData = $product->getData();
         $product->setWebsiteId($this->_getStore()->getWebsiteId());
         // customer group is required in product for correct prices calculation
@@ -118,7 +116,7 @@ abstract class Mage_Catalog_Model_Api2_Product_Rest extends Mage_Catalog_Model_A
 
         if ($this->getActionType() == self::ACTION_TYPE_ENTITY) {
             // define URLs
-            $productData['url'] = $productHelper->getProductUrl($product->getId());
+            $productData['url'] = $product->getProductUrl();
             /** @var Mage_Checkout_Helper_Cart $cartHelper */
             $cartHelper = Mage::helper('checkout/cart');
             $productData['buy_now_url'] = $cartHelper->getAddUrl($product);


sreichel:
Mhhh, cant reproduce atm ... my json-response has the correct frontend-url.

Hanmac:
Use multiple stores with `web/url/use_store` enabled
Then the Guest Api doesn't use the Frontend Route
So it can't decide what the current store should be

then this line:
https://github.com/OpenMage/magento-lts/blob/488c02150da4249d7ce1555020d025240ae71949/app/code/core/Mage/Catalog/Helper/Product.php#L62
does load the product without store id set

Hanmac:
Use multiple stores with `web/url/use_store` enabled
Then the Guest Api doesn't use the Frontend Route
So it can't decide what the current store should be

then this line:
https://github.com/OpenMage/magento-lts/blob/488c02150da4249d7ce1555020d025240ae71949/app/code/core/Mage/Catalog/Helper/Product.php#L62
does load the product without store id set

Hanmac:
Use multiple stores with `web/url/use_store` enabled
Then the Guest Api doesn't use the Frontend Route
So it can't decide what the current store should be

then this line:
https://github.com/OpenMage/magento-lts/blob/488c02150da4249d7ce1555020d025240ae71949/app/code/core/Mage/Catalog/Helper/Product.php#L62
does load the product without store id set

Hanmac:
Use multiple stores with `web/url/use_store` enabled
Then the Guest Api doesn't use the Frontend Route
So it can't decide what the current store should be

then this line:
https://github.com/OpenMage/magento-lts/blob/488c02150da4249d7ce1555020d025240ae71949/app/code/core/Mage/Catalog/Helper/Product.php#L62
does load the product without store id set

sreichel:
The fix is valid, but i still have no admin-url in response - its aleays a frontend-url.

- Use multiple stores with web/url/use_store enabled (thats default in sample data)
- enabled "add store code to url"

-> return default store url (not admin)

After this fix i get the correct frontend-url.

sreichel:
https://magento-lts.ddev.site/api/rest/products/ace000?store=0

> https://magento-lts.ddev.site/default/aviator-sunglasses.html

https://magento-lts.ddev.site/api/rest/products/ace000?store=1

> https://magento-lts.ddev.site/default/aviator-sunglasses.html

https://magento-lts.ddev.site/api/rest/products/ace000?store=2

> https://magento-lts.ddev.site/french/aviator-sunglasses.html

https://magento-lts.ddev.site/api/rest/products/ace000?store=3

> https://magento-lts.ddev.site/german/aviator-sunglasses.html

Hanmac:
@sreichel you can try to debug it too, but for me, when getUrlModel is called, the StoreId of the Product is still null,
and then app says the current store is admin

https://github.com/OpenMage/magento-lts/blob/488c02150da4249d7ce1555020d025240ae71949/app/code/core/Mage/Catalog/Model/Product.php#L368-L374

sreichel:
Checked and everything is fine.

Please debug `Mage_Api2_Model_Resource::_getStore()`.

Hanmac:
> Checked and everything is fine.
> 
> Please debug `Mage_Api2_Model_Resource::_getStore()`.

like is said before, it doesn't matter. While the `_getSore` method is called and loads the store for the given parameter, it doesn't alter the default store.

So when this line later happens, it will create a new Product object with the store parameter set to null, and then it will fall back to the current store (which is admin for the api)
https://github.com/OpenMage/magento-lts/blob/fc9a8ca634323663f6f585727f28ac9ee9611f72/app/code/core/Mage/Catalog/Model/Api2/Product/Rest.php#L121

sreichel:
> The fix is valid, but i still have no admin-url in response - its always a frontend-url.

You got me wrong. Your fix works - but for me the behavoir was different before the fix. I had no admin url,

sreichel (approved):

Hanmac:
@sreichel makes me wonder what does this `$productHelper->getProductUrl` does different for you than for me, that you get a different store.

Also for later, we might need to think about deprecating the getProductUrl from the Product Helper because it only causes problems in multi store if using product_id?

The only other place that uses Mage_Catalog_Helper_Product::getProductUrl is `Mage_Catalog_Block_Seo_Sitemap_Product`,
and the Sitemap uses the Product Item and not the Id

kiatng (approved):
Tested with https://github.com/kiatng/openmage-shooter

PR #4601, opened 2025-02-05
Fix error in column renderer when value is empty and not null

aamant:
### Description (*)

fix class Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Number when value is empty


sreichel:
Thanks. Whats the error message?

Can you provide steps to reproduce?

aamant:
This error occurs when the expected value is empty but not null

Capture d’écran 2025-02-05 à 10 05 15

sreichel:
> This error occurs

Please add error message and stacktrace.

aamant:
Capture d’écran 2025-02-05 à 12 11 36

aamant:
Capture d’écran 2025-02-05 à 12 11 36

colinmollenhour (changes requested):
The value of `empty('0')` is `true`, so this would change the behavior such that '0' renders the default value rather than '0'.

aamant:
I've updated the test

kiatng (approved):
My test: https://onlinephp.io/c/38a62

colinmollenhour (approved):

PR #4612, opened 2025-02-07
Replace Zend_Validation

sreichel:
### Description (*)

WIP (90%), but ready to review.

This attempts to replace another Zend_ library.

I did not choose laminas (as earlier suggested in #3199), nor symfony. No framework, straight-forward validation.

See https://github.com/Respect/Validation

### Related Pull Requests

- see OpenMage/magento-lts#3199


app/code/core/Mage/Checkout/Model/Type/Onepage.php
```diff
@@ -506,7 +508,7 @@ protected function _processValidateCustomer(Mage_Sales_Model_Quote_Address $addr
             }
         } elseif (self::METHOD_GUEST == $this->getQuote()->getCheckoutMethod()) {
             $email = $address->getData('email');
-            if (!Zend_Validate::is($email, 'EmailAddress')) {
+            if (!v::email()->validate($this->getEmail())) {
```
sreichel, line 511:
```suggestion
            if (!v::email()->validate($email)) {
```

Hanmac:
sad `no symfony` noises xD


Also, you might need to look at this:
```
@deprecated Calling `validate()` directly from rules is deprecated. Please use {@see \Respect\Validation\Validator::isValid()} instead.
```

sreichel:
> sad `no symfony` noises xD

I tried, but writing rules is more intuitive with that ... imho.

> Also, you might need to look at this:
> 
> ```
> @deprecated Calling `validate()` directly from rules is deprecated. Please use {@see \Respect\Validation\Validator::isValid()} instead.
> ```

I am aware of it, but latest version requies php 8.1.

Hanmac:
for the current changes in this MR, Symfony Validator might be easy too.

the bigger use would be that we could validate entire objects like Customer Address at once

Also the Intl/Translation support is nice, so we could have the errors already translated by the System

mattdavenport:
I would be in favor of switching this to Symfony validation. Looking at this from a bigger picture, it would be best if we picked a library of components (e.g. Symfony/Laminas/etc.) and stuck to those throughout the project when replacing outdated libraries. IMO not doing this will increase the maintenance burden long term with differing packages/conventions to keep track of.

tests/unit/Mage/Api/Model/UserTest.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * @dataProvider provideValidateApiUserData
+     * @param array|true $expectedResult
+     * @group Mage_Api
+     * @group Mage_Api_Model
+     */
+    public function testValidate($expectedResult, array $methods): void
+    {
+        $mock = $this->getMockBuilder(Subject::class)
+            ->setMethods([
```
Hanmac, line 45: If the `$methods` array is controlled by the dataProvider,
why not loop the data there?

`->setMethods(array_keys($methods))`
 and
```php
foreach ($methods as $key => $value) {
  $mock->method($key)->willReturn($value);
}
```

with that, you might be able to reduce the amount of duplicated lines

sreichel:
> I would be in favor of switching this to Symfony validation

Did it. Looks good, but found some changes ... wait for #4758.

app/code/core/Mage/Admin/Model/User.php
```diff
@@ -615,18 +632,26 @@ public function validate()
 
         if (isset($password)) {
             $minAdminPasswordLength = $this->getMinAdminPasswordLength();
-            if (Mage::helper('core/string')->strlen($password) < $minAdminPasswordLength) {
-                $errors->append(Mage::helper('adminhtml')
-                    ->__('Password must be at least of %d characters.', $minAdminPasswordLength));
+            $violations = new ArrayObject();
```
Hanmac, line 633: i would like it if the part with the validatePassword would be put into an extra Function, easier to rewrite
(like adding extra validation rules via Modules)

app/code/core/Mage/Customer/Model/Customer.php
```diff
@@ -1074,101 +1074,119 @@ public function setStore(Mage_Core_Model_Store $store)
...
-        if (!$this->getId() && !Zend_Validate::is($password, 'NotEmpty')) {
-            $errors[] = Mage::helper('customer')->__('The password cannot be empty.');
-        }
+        $email = $this->getEmail();
+        $violations->append($validator->validateEmail(
+            value: $email,
+            message: Mage::helper('customer')->__('Invalid email address "%s".', $email),
+        ));
 
+        $password = $this->getPassword();
```
Hanmac: Same there, I would like it to have the Password Validation in an extra Function

Maybe to not duplicate code with `validateResetPassword`?

sreichel: Very nice catch!

`validatePassword` was not finished yet to make it re-usable. 

Updated `validatePassword` and added extra methods as suggested.

app/code/core/Mage/Admin/Model/User.php
sreichel: Updated.

sreichel:
Working on some test, but thats hopefully ready for review.

sreichel:
@Hanmac  are the chanfes okay for you? Any further suggestions?

Think only file uploads have to be tested manually. Everything else should be covered by unit tests.

Hanmac (dismissed):

sreichel:
Merged with one green.

Issue #4501, opened 2025-01-21
Exception noise from OAuth and REST (API2 )


kiatng:
I have a project with OAuth REST which is interfacing with a mobile app. The exception.log is filling up with millions of lines due to unnecessary logs:

Mage_Api2_Exception: oauth_problem=token_rejected in .../app/code/core/Mage/Api2/Model/Auth/Adapter/Oauth.php:49
Mage_Api2_Exception: Missing required params in .../app/code/core/Mage/Api2/Model/Resource.php:629
Mage_Oauth_Exception in .../app/Mage.php:648, .../app/code/core/Mage/Oauth/Model/Server.php(421)

Each exception has 11 lines or more.

### Preconditions (*)
Any version.

### Steps to reproduce (*)
May be by unit tests?

### Expected result (*)

1. Do not `Mage::logException($e)` for `Mage_Api2_Exception` and `Mage_Oauth_Exception`? 
2. Only log one line per exception?

### Actual result (*)
Noise in `exception.log` making it hard to detect the signal, problems that require attention.

sreichel:
> Mage_Api2_Exception: Missing required params in .../app/code/core/Mage/Api2/Model/Resource.php:629

Please post that lines. Looks like modified code. (?)

kiatng:
> Please post that lines. Looks like modified code. (?)

The project has lots of custom API2. When the client requests certain info but fail some validations, an exception is thrown, an example of bad request:
```php
$this->_critical('Missing nationality', Mage_Api2_Model_Server::HTTP_BAD_REQUEST);
```

https://github.com/OpenMage/magento-lts/blob/00de7b5f4db329d16bff33ca497f288ebe369cd2/app/code/core/Mage/Api2/Model/Resource.php#L613-L626

sreichel:
> Line 629 is `throw new Mage_Api2_Exception($message, $code);` in my project.

In core thats line 625 ...

Isnt it correct to log "$this->_critical('Missing nationality'"?

kiatng:
Sure, it is OK to log, but it's unnecessary to include the trace in the log.

kiatng:
The exceptions are log here: https://github.com/OpenMage/magento-lts/blob/cf16ac85e2f0767f4a544c280410ac1ac3034576/app/code/core/Mage/Api2/Model/Server.php#L104-L107

What's the best way to handle it?

PR #4642, opened 2025-02-22
Fix issue #4501 Exception noise from OAuth and REST (API2 )

kiatng:
### Description (*)
See issue #4501. This PR added `$shouldLog` as an optional parameter in `Mage_APi2_Exception`, any code that extends it and overrides the constructor with the old signature would break. 

### Related Pull Requests

- fixes OpenMage/magento-lts#4501


app/code/core/Mage/Api2/Model/Resource.php
```diff
@@ -613,21 +613,25 @@ protected function _render($data)
...
-    protected function _critical($message, $code = null)
+    protected function _critical($message, $code = null, $shouldLog = true)
     {
         if ($code === null) {
             $errors = $this->_getCriticalErrors();
             if (!isset($errors[$message])) {
                 throw new Exception(
                     sprintf('Invalid error "%s" or error code missed.', $message),
-                    Mage_Api2_Model_Server::HTTP_INTERNAL_ERROR,
+                    Mage_Api2_Model_Server::HTTP_INTERNAL_ERROR
```
kiatng:
```suggestion
                    Mage_Api2_Model_Server::HTTP_INTERNAL_ERROR,
```

sreichel:
> any code that extends it and overrides the constructor with the old signature would break.

https://github.com/OpenMage/magento-lts/issues/4501#issuecomment-2606224769

How about to add `_criticalNotLoggable()` method and a new exception?

kiatng:
> > any code that extends it and overrides the constructor with the old signature would break.
> 
> [#4501 (comment)](https://github.com/OpenMage/magento-lts/issues/4501#issuecomment-2606224769)
> 
> How about to add `_criticalNotLoggable()` method and a new exception?

I have thought about new exception class. I am hesitant because I do not think there is any custom code that would extend it, if there is, it's easy to fix. This avoid adding more code.

Hanmac (dismissed):
I think this is okay for now?

Another approach would be to check the log-level (like show them on debug/info?)

app/code/core/Mage/Api2/Exception.php
```diff
@@ -14,18 +14,36 @@
  */
 class Mage_Api2_Exception extends Exception
 {
+    /**
+     * Log the exception in the log file?
+     * @var bool
+     */
+    protected $_shouldLog = true;
```
sreichel:
```suggestion
    protected bool $shouldLog = true;
```

app/code/core/Mage/Api2/Exception.php
```diff
@@ -14,18 +14,36 @@
...
+     * @param bool $shouldLog
      */
-    public function __construct($message, $code)
+    public function __construct($message, $code, $shouldLog = true)
     {
         if ($code <= 100 || $code >= 599) {
             throw new Exception(sprintf('Invalid Exception code "%d"', $code));
         }
 
+        $this->_shouldLog = $shouldLog;
```
sreichel:
```suggestion
        $this->shouldLog = $shouldLog;
```

app/code/core/Mage/Api2/Exception.php
```diff
@@ -14,18 +14,36 @@
...
+
+    /**
+     * Check if exception should be logged
+     *
+     * @return bool
+     */
+    public function shouldLog()
+    {
+        return $this->_shouldLog;
+    }
```
sreichel:
```suggestion
    /**
     * Check if exception should be logged
     */
    public function shouldLog(): bool
    {
        return $this->shouldLog;
    }
```

app/code/core/Mage/Api2/Exception.php
```diff
@@ -14,18 +14,34 @@
  */
 class Mage_Api2_Exception extends Exception
 {
+    /**
+     * Log the exception in the log file?
+     * @var bool
+     */
```
sreichel:
```suggestion
    /**
     * Log the exception in the log file?
     */
```

sreichel (approved):

sreichel:
Already reviewed by @Hanmac .

addison74:
This one has a PHPstan error. Did you check it?

sreichel:
All lights green here?

Hanmac:
@addison74 there was an outdated PHPstan error, that got fixed with a later commit

PR #4668, opened 2025-03-07
FEAT: disable advanced search on storeview scope

sreichel:
This is a port of one of my extensions. It has only 265 downloads for M1, but 30k for Magento2. Worth to add 2 observers?

For me it was usefull as many stores dont have well configured attributes that gave advanced search a bad UX. It was better to disable it.

Note: i followed M2s SRP (single-responsibility principle) for observers and created a new class with only one `excetute`-method. The interface for that will come in a seperate PR.

### DIFF
--- a/app/code/core/Mage/Catalog/Helper/Search.php
+++ b/app/code/core/Mage/Catalog/Helper/Search.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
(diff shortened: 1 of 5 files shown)


addison74 (approved):
LGTM

kiatng (approved):

PR #4693, opened 2025-03-13
rector: `PowToExpRector`

sreichel:
See https://getrector.com/rule-detail/pow-to-exp-rector ... php56

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,9 +1,7 @@
 <?php
-/*
-* This document has been generated with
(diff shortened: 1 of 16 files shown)


addison74 (approved):

Hanmac (approved):

Hanmac:
```php
$symbol = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
$exp = floor(log($bytes) / log(1024));
```

Technically, these are the wrong symbols, (KB=1000, KiB=1024) but that might be for another time

PR #4696, opened 2025-03-14
Update `release-drafter.yml`

sreichel:
Last one did not work ... next try.

If it will not work there is `expclude-labels` option as last chance.


addison74 (approved):

kiatng (approved):

sreichel:
Meh ... did not work too. Final PR coming ...

PR #4706, opened 2025-03-15
rector: applied all rules up to `php-74`

sreichel:
... skipped some that did not work properly.

### DIFF
--- a/.rector.php
+++ b/.rector.php
@@ -16,7 +16,7 @@ use Rector\TypeDeclaration\Rector as TypeDeclaration;
 try {
     return RectorConfig::configure()
         ->withPhpSets(
-            php71: true,
+            php73: true,
         )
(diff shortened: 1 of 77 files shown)


Hanmac:
`} catch (Mage_Core_Exception|Exception $e) {`

Doesn't Mage_Core_Exception extend from Exception?
I'm bit surprised it didn't got simplified out

sreichel:
You cant simplify as long its one of them.

> As of PHP 7.1.0, a [catch](https://www.php.net/manual/en/language.exceptions.php#language.exceptions.catch) block may specify multiple exceptions using the pipe (|) character. This is useful for when different exceptions **from different class hierarchies** are handled the same.

Hanmac (approved):

kiatng (approved):

PR #4697, opened 2025-03-14
[FEAT] Redirect single search result to product page

sreichel:
### Description (*)

When product search has only one result, redirect to product page.

- new config options added (enabled by default)

### Manual testing scenarios (*)

1. use sample data
2. search for "wonderland"
3. should redirect to a books product page

PR #4712, opened 2025-03-16
[FEAT] Added `Mage_Core_Observer_Interface` with `execute()` method

sreichel:
Extracted that from #4697.


pquerner:
Does that mean the XML structure also changes? (When you "register" a observer)

sreichel:
> Does that mean the XML structure also changes? (When you "register" a observer)

There is no change in XML structure.

If the observer implements the interface, you dont need to set a method in XML, because it uses `execute` .

PR #4720, opened 2025-03-19
Fix MySQL connection for DDEV setup

stasadev:
### Description (*)

`ddev openmage-admin` doesn't import sample data because DDEV don't use `MYSQL_PWD` anymore:

- https://github.com/ddev/ddev/pull/6887

```
$ ddev openmage-install -q
...
Importing Sample Data into the database...
ERROR 1045 (28000): Access denied for user 'db'@'172.19.0.2' (using password: YES)
...
```

Use DDEV quickstart for OpenMage from:

- https://github.com/ddev/ddev/pull/7091

Sample data should be successfully imported.

### DIFF
--- a/.ddev/commands/web/openmage-admin
+++ b/.ddev/commands/web/openmage-admin
@@ -29,14 +29,14 @@ fi
 RANDOM_STRING=$({ tr -dc A-Za-z0-9 </dev/urandom | head -c 32 ; })
 
 if [[ "${ACTION}" =~ ^(update|u) ]]; then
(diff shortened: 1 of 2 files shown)


colinmollenhour (approved):

stasadev:
The only affected DDEV versions are v1.24.2 and v1.24.3, so if you are using an earlier version, it should work as expected.

addison74 (approved):

kiatng:
@all-contributors please add stasadev for code

Issue #4732, opened 2025-03-26
[BUG]  Taxvat, Dob, and Gender Attributes Ignore WEBSITE Scope and Only Work on GLOBAL Scope


empiricompany:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The Taxvat attribute (along with other account-level attributes such as dob and gender) only works at the GLOBAL scope and does not respect custom values set at the WEBSITE scope.

For example, if Taxvat is disabled GLOBALLY but required for a specific WEBSITE, it does not appear on the Account Edit form.

![config_website](https://github.com/user-attachments/assets/7266b627-d0e6-44ff-9599-268b57d0446b)

The issue comes from `app\code\core\Mage\Customer\Block\Widget\Taxvat.php `

```php
class Mage_Customer_Block_Widget_Taxvat extends Mage_Customer_Block_Widget_Abstract
{
    public function _construct()
    {
        parent::_construct();
        $this->setTemplate('customer/widget/taxvat.phtml');
    }

    /**
     * @return bool
     */
    public function isEnabled()
    {
        return (bool)Mage::getStoreConfig('customer/address/taxvat_show'); <==== possible fix read from config
        return (bool)$this->_getAttribute('taxvat')->getIsVisible();
    }

    /**
     * @return bool
     */
    public function isRequired()
    {
        return (bool)Mage::getStoreConfig('customer/address/taxvat_show') == 'req'; <==== possible fix read from config
        return (bool)$this->_getAttribute('taxvat')->getIsRequired();
    }

    /**
     * @return Mage_Customer_Model_Customer
     */
    public function getCustomer()
    {
        return Mage::getSingleton('customer/session')->getCustomer();
    }
}
```

Called from various templates such as `app\design\frontend\rwd\default\template\customer\form\edit.phtml`

```php
<?php $_dob = $this->getLayout()->createBlock('customer/widget_dob') ?>
<?php if ($_dob->isEnabled()): ?>
    <li><?php echo $_dob->setDate($this->getCustomer()->getDob())->toHtml() ?></li>
<?php endif ?>
<?php $_taxvat = $this->getLayout()->createBlock('customer/widget_taxvat') ?>
<?php if ($_taxvat->isEnabled()): ?>
    <li><?php echo $_taxvat->setTaxvat($this->getCustomer()->getTaxvat())->toHtml() ?></li>
<?php endif ?>
[...]

PR #4745, opened 2025-04-09
Fix website-specific attribute values not loading correctly

empiricompany:
## Description

This PR fixes an issue where website-specific attribute values (stored in tables like `customer_eav_attribute_website`) are not loaded correctly for EAV attributes. The problem affects attributes like `taxvat` that should respect website-specific configurations.

## Problem

The issue occurs in the EAV attribute loading process. When loading entity attributes in `Mage_Eav_Model_Config::_loadEntityAttributes()`, the code calls:

```php
$entityAttributes = $entityType->newAttributeCollection()
    ->addStoreLabel($storeId)
    ->getData();
```

The `newAttributeCollection()` method in `Mage_Eav_Model_Entity_Type` calls `_getAttributeCollection()`, which creates a collection using:

```php
$collection = Mage::getModel('eav/entity_attribute')->getCollection();
```

This always returns an instance of `Mage_Eav_Model_Resource_Entity_Attribute_Collection`, which does not have the `setWebsite()` method needed to load website-specific attribute values.

For entity types like 'customer' that have website-specific attributes, this means the website-specific values are never loaded, even though they exist in the database.

## Proposed Solution

The solution has two parts:

### 1. Fix the attribute collection creation

Modify the `_getAttributeCollection()` method in `Mage_Eav_Model_Entity_Type` to use the entity's configured attribute collection class instead of always using the base collection:

```php
protected function _getAttributeCollection()
{
    $collectionClass = $this->getEntityAttributeCollection();
    $collection = Mage::getResourceModel($collectionClass);
    
    $objectsModel = $this->getAttributeModel();
    if ($objectsModel) {
        $collection->setModel($objectsModel);
    }

    return $collection;
}
```

This leverages the existing `getEntityAttributeCollection()` method, which returns the collection class specified in the entity type's XML configuration. By default, it returns 'eav/entity_attribute_collection', but modules can override this for specific entity types.

### 2. Fix the attribute loading process with cache

Modify the `_loadEntityAttributes()` method in `Mage_Eav_Model_Config` to check for website-specific collections and set the website ID:

```php
protected function _loadEntityAttributes($entityType, $storeId)
{
    // preload attributes in array form to avoid instantiating
    // models for every attribute even if it is never accessed
    $collection = $entityType->newAttributeCollection()
        ->addStoreLabel($storeId);

    // if collection supports per-website values, set website id
    if ($collection instanceof Mage_Eav_Model_Resource_Attribute_Collection) {
        $websiteId = Mage::app()->getStore($storeId)->getWebsiteId();
        $collection->setWebsite($websiteId);
    }

    $entityAttributes = $collection->getData();
    
    // ... rest of the method ...
}
```

Using `instanceof` instead of `method_exists()` is more reliable and efficient because:

1. **Type safety**: It ensures we're only calling `setWebsite()` on collections that are designed to handle website-specific values
2. **Performance**: Type checking is faster than method existence checking
3. **Clarity**: It explicitly shows that only collections extending `Mage_Eav_Model_Resource_Attribute_Collection` support website-specific values
4. **Consistency**: It follows object-oriented principles by relying on the type hierarchy rather than duck typing

Currently, only two collections in the core code implement the abstract class with `setWebsite()`:
- `Mage_Customer_Model_Resource_Attribute_Collection`
- `Mage_Customer_Model_Entity_Attribute_Collection` (which extends the first one)

This is consistent with the database schema, as only the Customer module has a `customer_eav_attribute_website` table to store website-specific attribute values. Other entity types like Catalog do not have equivalent tables and their collections do not extend the abstract class with `setWebsite()`.

For example, the Customer module can specify:
```xml
<entity_attribute_collection>customer/resource_attribute_collection</entity_attribute_collection>
```

With these changes, entity types that have website-specific attributes can provide their own attribute collection class that implements the `setWebsite()` method, and the attribute loading process will correctly use this method when caching attributes.

## Diagram

```mermaid
flowchart TD
    A[Mage_Eav_Model_Config::_loadEntityAttributes] --> B[entityType->newAttributeCollection]
    B --> C[Mage_Eav_Model_Entity_Type::_getAttributeCollection]
    
    subgraph "Before Fix"
        C -->|Always creates| D["Mage.getModel(eav/entity_attribute).getCollection"]
        D -->|Returns| E[Mage_Eav_Model_Resource_Entity_Attribute_Collection]
        E -->|No setWebsite method| F[Website-specific values not loaded]
    end
    
    subgraph "After Fix"
        C2[Mage_Eav_Model_Entity_Type::_getAttributeCollection] -->|Uses configuration| G["this.getEntityAttributeCollection"]
        G -->|Returns collection class| H["Mage.getResourceModel(collectionClass)"]
        H -->|For customer entity| I[Mage_Customer_Model_Resource_Attribute_Collection]
        I -->|Has setWebsite method| J[Website-specific values loaded correctly]
    end
```

## Benefits

1. **Fixes website-specific attribute values**: Attributes like `taxvat` will now correctly respect website-specific configurations.
2. **Generic solution**: Works for all entity types, not just customer.
3. **Configurable**: Each module can specify its own attribute collection class.
4. **Respects Magento architecture**: Uses the existing XML configuration system.
5. **Minimal change**: Only modifies two methods in two classes.
6. **Type-safe implementation**: Uses proper OOP principles with instanceof checks.
7. **Cache-aware**: Ensures website-specific values are respected even when attributes are cached.

## Testing

1. Create a multi-website setup
2. Configure different settings for the `taxvat` attribute on different websites
3. Verify that the attribute respects the website-specific configuration
4. Test with cache enabled to ensure website-specific values are still respected

**Note**: This solution requires thorough testing across different environments and configurations as it affects a core part of the EAV system. Comprehensive regression testing is strongly recommended to ensure no unintended side effects occur in other parts of the system that rely on attribute collections.

## Backward Compatibility

This change is backward compatible as it:
1. Does not change any public APIs
2. Uses existing configuration mechanisms
3. Falls back to the default behavior if no custom collection class is specified
4. Only applies website-specific behavior to collections that explicitly support it

## Related Issues

This fixes the following issues:
- [Issue #4732](https://github.com/OpenMage/magento-lts/issues/4732): Website-specific attribute values not loading correctly


empiricompany:
Now I’ve encountered a problem when the cache is enabled.

empiricompany:
Now it works correctly even with EAV cache enabled.

kiatng (approved):
Tested.

Issue #4749, opened 2025-04-12
[BUG] Products with same position value appear duplicated across pagination in category view


gdemarsico:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The same product appears on multiple pages when browsing through the pagination, specifically when sorting by position and multiple products have the same position value.

### Expected Behavior

Each product should appear exactly once in the paginated listing.

## Steps to Reproduce
1. Have multiple products in a category with the same position value (typically 0)
2. Apply a filter to the category 
3. Navigate between pages 1 and 2
4. Notice that some products (in my case, product ID 5034) appear on multiple pages

### Environment

```markdown
- OpenMage: 20.12.3
- php:8.3
```

## Proposed Solution
Modify the class Mage_Catalog_Model_Resource_Product_Collection change `addAttributeToSort` method to include entity_id as a tiebreaker criterion:

```php
if ($attribute == 'position') {
    if (isset($this->_productLimitationFilters['category_id'])) {
        $this->getSelect()->order(['cat_index.position ' . $dir, 'e.entity_id ' . $dir]);
    }
    return $this;
}
```

Hanmac:
Closed because of merged MR

PR #4750, opened 2025-04-12
Fix products duplicate across pagination when they have the same posi…

gdemarsico:
### Issue #4749 
When filtering products in a category with pagination enabled, some products can appear duplicated across different pages if they have the same value in the "position" field.

## Steps to Reproduce
1. Have multiple products in a category with the same position value (typically 0)
2. Apply a filter to the category 
3. Navigate between pages 1 and 2
4. Notice that some products (in my case, product ID 5034) appear on multiple pages

## Expected Behavior
Each product should appear exactly once in the paginated listing.

## Actual Behavior
The same product appears on multiple pages when browsing through the pagination, specifically when sorting by position and multiple products have the same position value.

## Proposed Solution
Modify the `addAttributeToSort` method to include entity_id as a tiebreaker criterion:

```php
if ($attribute == 'position') {
    if (isset($this->_productLimitationFilters['category_id'])) {
        $this->getSelect()->order(['cat_index.position ' . $dir, 'e.entity_id ASC']);
    }
    return $this;
}
```

Instead of only ordering by position. This would ensure a deterministic order even when multiple products have the same position value.

## Environment
- OpenMage version: [your version]
- PHP version: [your version]
- MySQL version: [your version]

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
@@ -1520,7 +1520,7 @@ class Mage_Catalog_Model_Resource_Product_Collection extends Mage_Catalog_Model_
             // optimize if using cat index
             $filters = $this->_productLimitationFilters;
             if (isset($filters['category_id']) || isset($filters['visibility'])) {
-                $this->getSelect()->order('cat_index.position ' . $dir);
+                $this->getSelect()->order(['cat_index.position ' . $dir, 'e.entity_id ' . $dir]);
             } else {
                 $this->getSelect()->order('e.entity_id ' . $dir);
             }


kiatng:
I tried testing in sample category, where the position for all products is set to 1. I set 4 products to show in grid view. I cannot replicate the issue.

gdemarsico:
> I tried testing in sample category, where the position for all products is set to 1. I set 2 products to show in grid view. I cannot replicate the issue. I tested with Accessories > Shoes and filter for price. I tested turned on and off flat and category.
> 
> Ref https://magento.stackexchange.com/questions/4829/product-list-sometimes-show-duplicate-products.

Hi, I import the products with magmi, and category position is no seted, so all products have 0 as default value. I this scenario i have found the bug

kiatng (approved):
If all the position values are 0, the `ORDER BY` clause does not provide a deterministic order. This means the database is free to return rows in any order for rows with the same position value. By adding `e.entity_id` as a secondary sorting parameter, the order becomes consistent and ensures uniqueness. This eliminates duplicates across different pages.

Hanmac (approved):

PR #4768, opened 2025-04-21
Fix: removed not working code

sreichel:
Not working b/c incompatibility to interface.


kiatng:
How about
https://github.com/OpenMage/magento-lts/blob/0262423ae568cca01bc98e9b728fd91ad23d6f90/app/code/core/Mage/Api/etc/api.xml#L123-L126

PR #4753, opened 2025-04-16
Add Mage_Csp Module

Hanmac:
### Description (*)

Big Inspired by #4721, but extended with:
- Seperate base config for frontend and backend in the Magento settings.
- Also allowing Layout Updates to add more directives.
- (Should I also add Events for this?)

### Related Pull Requests

- see OpenMage/magento-lts#4721

### Fixed Issues (if relevant)

- fixes OpenMage/magento-lts#<issue_number>

### Manual testing scenarios (*)

1. ...
2. ...

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [x] All commits are accompanied by meaningful commit messages
 - [x] All automated tests passed successfully (all builds are green)

PR #4776, opened 2025-04-24
Content Security Policy (CSP) Implementation a new approach

empiricompany:
# Content Security Policy (CSP) Implementation in OpenMage

This PR adds comprehensive support for Content Security Policy (CSP) headers in OpenMage, significantly improving application security against XSS attacks and other client-side security vulnerabilities. The module was developed based on the PR https://github.com/OpenMage/magento-lts/pull/4753 by @Hanmac, **to whom all credit should be given**. I decided to create a completely separate PR because there were many changes and the approach is very different. I don't want to impose my solution but rather propose a more robust implementation following the OpenMage standards that guarantees the same functionality, such as the ability to specify directives per module and the separation of frontend and backend directives.

## Features

- **Modular configuration**: Each module can specify its own CSP directives in the default section areas:
  - `global/csp/<directive-name>`: Applied to both frontend and backend
  - `frontend/csp/<directive-name>`: Applied only to the frontend
  - `adminhtml/csp/<directive-name>`: Applied only to the backend (administration panel)

- **Administration interface**: Administrators can:
  - Enable/disable CSP separately for frontend and admin areas
  - Configure "Report Only" mode for each area
  - Configure "Report URI" for each area
  - Split headers for each directive to avoid potential web server/browser limitations
  
  split_headers
  **IMPORTANT:** Read this https://content-security-policy.com/examples/multiple-csp-headers/

  - Add custom directives
  - View directives added by modules (shown as readonly and non-modifiable) and scope (`<global />,` `<frontend />`, `<admihtml />`)
  - All CSP configuration for **Frontend** can be custimezed per store view
  - Ability to convert <meta> directives into HTTP headers

  config
  config_frontend

- **Pre-configured security**: The module includes safe directives in the global section (like 'self' for most directives) providing a secure baseline configuration

- **Support for all standard CSP directives**:
  - default-src
  - script-src
  - style-src
  - img-src
  - connect-src
  - font-src
  - frame-src
  - object-src
  - media-src
  - form-action

## Advantages of This Approach

The key technical advantages of this implementation include:

- **Efficient configuration caching**: Directives from config.xml are properly cached, minimizing performance impact
- **Strategic event integration with presentation layer separation**: Using `http_response_send_before` ensures headers are applied at the optimal moment in the request lifecycle while preventing themes from injecting potentially malicious directives
- **Developer-friendly format**: The configuration structure aligns with OpenMage conventions, making adoption intuitive
- **Safe defaults**: The module is disabled by default with "Report Only" mode enabled to prevent breaking functionality

## Configuration Example in a Module

Modules can specify CSP directives in their `config.xml` file in three different areas. This configuration approach follows the same pattern used for `global>ignore_user_agents` in OpenMage:

```xml
<config>
    <!-- Global directives applied to both frontend and backend -->
    <global>
        <csp>
            <script-src>
                <my-service>example.com</my-service>
                <my-cdn>cdn.example.com</my-cdn>
            </script-src>
            <img-src>
                <my-images>images.example.com</my-images>
            </img-src>
        </csp>
    </global>
    
    <!-- Frontend-specific directives -->
    <frontend>
        <csp>
            <connect-src>
                <my-api>api.example.com</my-api>
            </connect-src>
            <frame-src>
                <my-widget>widgets.example.com</my-widget>
            </frame-src>
        </csp>
    </frontend>
    
    <!-- Admin-specific directives -->
    <adminhtml>
        <csp>
            <script-src>
                <admin-tool>admin-tools.example.com</admin-tool>
            </script-src>
            <style-src>
                <admin-styles>admin-styles.example.com</admin-styles>
            </style-src>
        </csp>
    </adminhtml>
</config>
```

Each directive can have multiple entries, and they will be properly merged with directives from other modules and user configurations.
You can see for each directive in which area it was added:
config_scope

## Resources
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy
- https://csp-evaluator.withgoogle.com/

### Ability to add directives from views
The ability to add directives from the presentation layer was deliberately left out. It could be easily implemented, but IMHO views should not have this kind of "power". In cases where themes include external resources, directives can be easily added by checking the browser console.

**= UPDATE =**
I implemented support for adding directives via `<meta>`, but it’s strongly discouraged according to the documentation.

For this reason, all  `<meta>` directives added via layout update are merged into the HTTP header by default. If you want to enable directive insertion through  `<meta>`, you must explicitly it in the configuration.

merge_meta

It’s possible to add them through a layout update like this:

```xml
<layout version="0.1.0">
    <checkout_onepage_index>
        <reference name="csp_meta">
            <action method="addDirective"><directive>default-src</directive><value>*.default-from-layout-update.com</value></action>
            <action method="addDirective"><directive>script-src</directive><value>*.script-from-layout-update.com</value></action>
            <action method="addDirective"><directive>style-src</directive><value>*.style-from-layout-update.com'</value></action>
        </reference>
    </checkout_onepage_index>
</layout>
```
This will print in the head (in report-only or enforce mode according to the configuration):
```html
<meta http-equiv="Content-Security-Policy-Report-Only" content="default-src *.default-from-layout-update.com; script-src *.script-from-layout-update.com; style-src *.style-from-layout-update.com'; report-uri https://xxx.report-uri.com/r/d/csp/reportOnly" />
```

These directives will be difficult to debug because they won’t appear in the configuration like those declared by the modules.
**IMPORTANT:** Read this https://content-security-policy.com/examples/meta/

### API Responses and Event Handling

CSP headers are currently added to all responses, including API responses, by observing the `http_response_send_before` event. This shouldn't create problems since CSP directives are only interpreted by browsers and add minimal overhead to API responses. 
However, it raises the question:
Is it better to hook into an event that only affects frontend/HTML responses, or is it preferable to have CSP headers consistently applied to all responses regardless of type?
Netalico solution use "controller_action_predispatch"
https://github.com/netalico/magento-1-csp/blob/master/app/code/local/Netalico/Csp/etc/config.xml

### Report URI Implementation

The current implementation include configuration for a `report-to` directive declared in `Reporting-Endpoints` header, which could be valuable for collecting violation reports to external service like https://report-uri.com.

Future enhancements could include:
- Implementing an internal reporting system that collects violation reports and notifies administrators
- Creating an interface that allows administrators to easily whitelist reported domains after verification

### Enforce only specific frontend area
I'm thinking about implementing the possibility to enable CSP enforcement only in specific areas, for example on the checkout_onepage_index controller — although I'm not sure I'll go through with it, since I believe the entire site should include security headers.

### Default directives
Does it make sense to include default directives for services like Google Analytics, PayPal, Stripe, etc.?
These can change over time and there will never be a complete list — it will always be specific to each use case.
It feels like the same discussion we had about `ignore_user_agents`.

### Nonce helper
A nonce helper could also be implemented to use in the views for scripts and automatically add it to the directives.
Resource: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP#nonces


Hanmac:
I like the different Input fields, but I would have tried it with less classes. (I might try to get it working in my MR)

One of the reasons why I wanted the layout update approach was so I could enable different urls depending on the page. Like if Captcha is enabled and it uses external source, I could enable it only for the ones with captcha.
(Maybe even with Events)

empiricompany:
Hey @Hanmac, thanks for your work and inspiration!

> I like the different Input fields, but I would have tried it with less classes. (I might try to get it working in my MR)
> 
> One of the reasons why I wanted the layout update approach was so I could enable different urls depending on the page. Like if Captcha is enabled and it uses external source, I could enable it only for the ones with captcha. (Maybe even with Events)

In my view, CSPs should be set at the controller level, where the response and its headers are handled.
To me, it doesn’t make sense to enable a host only for certain pages—if it’s safe, it can be whitelisted for the entire site.
Also, this way I can't get a clear backend-side view of which hosts are active, who is whitelisting them and where.
I prefer this separation of responsibilities.

app/code/core/Mage/Csp/Helper/Data.php
```diff
@@ -0,0 +1,145 @@
...
+
+        return $policy;
+    }
+
+    /**
+     * Get global policy for a specific directive
+     * global/csp/<directiveName>
+     * @param string $directiveName
+     */
+    public function getGlobalPolicy($directiveName = 'default-src'): array
```
sreichel, line 102:
```suggestion
     * @param value-of<self::CSP_DIRECTIVES> $directiveName
     */
    public function getGlobalPolicy(string $directiveName = 'default-src'): array
```

app/code/core/Mage/Csp/Helper/Data.php
```diff
@@ -0,0 +1,145 @@
...
+        return [];
+    }
+
+    /**
+     * Get area policy for a specific directive
+     * (adminhtml|frontend)/csp/<directiveName>
+     * @param string $area
+     * @param string $directiveName
+     */
+    public function getAreaPolicy($area = Mage_Core_Model_App_Area::AREA_FRONTEND, $directiveName = 'default-src'): array
```
sreichel, line 117:
```suggestion
     * @param Mage_Core_Model_App_Area::AREA_FRONTEND|Mage_Core_Model_App_Area::AREA_ADMINHTML $area
     * @param value-of<self::CSP_DIRECTIVES> $directiveName
     */
    public function getAreaPolicy(string $area = Mage_Core_Model_App_Area::AREA_FRONTEND, string $directiveName = 'default-src'): array
```

app/code/core/Mage/Csp/Helper/Data.php
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+
+    /**
+     * Get system policy for a specific directive
+     * csp/(adminhtml|frontend)/<directiveName>
+     * @param string $area
+     * @param string $directiveName
+     */
+    public function getStoreConfigPolicy($area = Mage_Core_Model_App_Area::AREA_FRONTEND, $directiveName = 'default-src'): array
```
sreichel, line 133:
```suggestion
     * @param Mage_Core_Model_App_Area::AREA_FRONTEND|Mage_Core_Model_App_Area::AREA_ADMINHTML $area
     * @param value-of<self::CSP_DIRECTIVES> $directiveName
     */
    public function getStoreConfigPolicy(string $area = Mage_Core_Model_App_Area::AREA_FRONTEND, string $directiveName = 'default-src'): array
```

colinmollenhour:
Awesome, work guys! CSP is the only real way to stop MageCart style attacks so this is major.

I think allowing the rules to be added via layouts and events is useful so that extensions can inject their CSP policies without overriding controllers or requiring manual update to the config or global config. For example, you may want an obnoxious add-on the main site, but exclude it from the checkout or account pages, or it may only apply to product pages (e.g. a reviews plugin) so adding it everywhere just bloats your header sizes. Keep in mind that some third-party add-ons don't just require a single rule, some of the worst ones may require a bunch of rules (Facebook is one of the worst if I recall correctly).

Hanmac:
> I think allowing the rules to be added via layouts and events is useful so that extensions can inject their CSP policies without overriding controllers or requiring manual update to the config or global config. **For example, you may want an obnoxious add-on the main site, but exclude it from the checkout or account pages, or it may only apply to product pages (e.g. a reviews plugin) so adding it everywhere just bloats your header sizes.** Keep in mind that some third-party add-ons don't just require a single rule, some of the worst ones may require a bunch of rules (Facebook is one of the worst if I recall correctly).

@colinmollenhour This is what my #4753 can do

imo it would have been better if they did suggestions in the comments instead of making a new MR

app/code/core/Mage/Csp/Block/Meta.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        if (!in_array($directive, Mage_Csp_Helper_Data::CSP_DIRECTIVES)) {
+            return $this;
+        }
+
+        if (!isset($this->_directives[$directive])) {
+            $this->_directives[$directive] = [];
+        }
+
+        $this->_directives[$directive][] = $value;
```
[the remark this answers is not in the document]
empiricompany: why not with underscore like other properties like _data in page/html_head?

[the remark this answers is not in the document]
sreichel: We talked about in https://github.com/OpenMage/magento-lts/discussions/2748

It seems arround 50/50 for consistency vs code quality. I'd prefer 2nd and go for best practice.

phpmd complains about it (supressed atm), also Sonar does.

sreichel:
> imo it would have been better if **they** did suggestions

Erm, they? I have made same suggestions to your PR. 

It should not be that hard to merge everthing together to have possibility of layout updates and everthing else. (?)

Hanmac:
> > imo it would have been better if **they** did suggestions
> 
> Erm, they? I have made same suggestions to your PR.
> 
> It should not be that hard to merge everthing together to have possibility of layout updates and everthing else. (?)

I wasn't talking about you, don't get offended

colinmollenhour:
> This is what my #4753 can do

I know, I was agreeing that this is a valuable functionality. I don't see why the two approaches couldn't be combined so that users and dev could both add rules as they saw fit. Personally, I'd rather have the CSP rules managed in a repo than in a database, but I know lots of people prefer to do things through the admin UI and the UI fields can always be hidden with a system.xml override so both methods could be supported simultaneously.

empiricompany:
@Hanmac @colinmollenhour 
How were you thinking of adding these features?
- Via a layout XML update, like it's currently done for <meta> (we could remove the meta addition and collect them all in the header instead)
- Via an event that hooks in before all CSP policies and directives are collected (in the event, it would be possible to check the current route and apply conditions)
- Any other ideas?

Hanmac:
@empiricompany i wanted to test your `Mage_Adminhtml_Block_System_Config_Form_Field_Csp_Hosts` in my PR with a bit changes, but had some problems with that.

For this PR, i wanted to check if its possible to not have different classes for this, but have the Type be provided somehow else
like have it guess the directive name or something, but i need to debug that

sreichel:
Maybe an interesting idea https://github.com/basecom/magento2-csp-split-header

empiricompany:
> Maybe an interesting idea https://github.com/basecom/magento2-csp-split-header

Ok, we can split the headers for each directive, but this module does not allow setting different headers based on the route. 
It seems to me that even in Magento 2 this issue does not arise.

app/code/core/Mage/Csp/Block/Meta.php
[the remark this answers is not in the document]
colinmollenhour: @sreichel I propose we let it go for this PR. Looks like you already allowed it for phpmd. Hopefully Sonar will evolve to allow it as well, or looks like it is possible to write a plugin to allow it.

colinmollenhour:
> How were you thinking of adding these features?

My bad, looks like with the `csp_meta` block you can add updates through the layout or could use an event observer at any time to add directives, so that completely covers what I had in mind. 👍 

A doc page explaining the ways it can be used would be fantastic. :)

app/code/core/Mage/Csp/Block/Meta.php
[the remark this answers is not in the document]
sreichel: It was only a suggestion. When you or Tony prefers underscores ... lets go. :)

app/code/core/Mage/Csp/etc/config.xml
```diff
@@ -0,0 +1,173 @@
...
+                    <maxcdn-font>maxcdn.bootstrapcdn.com</maxcdn-font>
+                </font-src>
+                <frame-src>
+                    <paypal-frame>www.paypal.com</paypal-frame>
+                    <amazon>payments.amazon.com</amazon>
+                </frame-src>
+                <form-action>
+                    <paypal-form>www.paypal.com</paypal-form>
+                    <paypal-secure-form>securepayments.paypal.com</paypal-secure-form>
+                </form-action>
```
colinmollenhour, line 165: These defaults are the only thing I'm not sure about. For merchants that don't use these it is a lot of wasted XML that bloats the cache that is loaded and unserialized on every request. Also, once users click "save", we can no longer control these defaults in the future. Some ideas to address it would be:
- Move these to a documentation page under "Common Directives" and let the user add them manually.
- Add a form field for "presets" that the user can choose, and choosing the preset loads the directives from code rather than XML. That way the XML is not bloated and the code can be updated in future versions. Also, the community or other extensions could add their own presets to the list if it was implemented in such a way as to allow it (e.g. an XML node of class names to get directives from).

That said, it's not a blocker. Having CSP is super important.

colinmollenhour (dismissed):

app/code/core/Mage/Csp/etc/config.xml
sreichel: > * Move these to a documentation page under "Common Directives" and let the user add them manually.

:+1:

empiricompany:
> > How were you thinking of adding these features?
> 
> My bad, looks like with the `csp_meta` block you can add updates through the layout or could use an event observer at any time to add directives, so that completely covers what I had in mind. 👍 

Currently the directives specified via the Layout Update helper are added as <meta>. My proposal was to collect and add them at the end to the request header. 

We haven't yet implemented an observer that dispatches when adding headers.

app/code/core/Mage/Csp/etc/config.xml
empiricompany: Also in my opinion they should be included in the documentation and linked in the configuration.

Hanmac:
> I like the different Input fields, but I would have tried it with less classes. (I might try to get it working in my MR)

I finally found a way I have been looking for:

`dom_import_simplexml($element->getData('field_config'))->getNodePath()`

The Element is set in `Mage_Adminhtml_Block_System_Config_Form_Field_Array_Abstract::_getElementHtml` before `_toHtml` is called

With that you get the full config path like this one:
`/config/sections/design/groups/package/fields/ua_regexp`

then you can parse that path back again into `[design, package, ua_regexp]` 

So you only need one singular Class instead of multiple ones.

-----

Or a different approach would be like `Mage_Adminhtml_Block_System_Config_Form_Fieldset_Modules_DisableOutput`,
where the renderer is set on the Fieldset and the Entire list of Options like `script-src` and `script-src` are Generated

empiricompany:
> > I like the different Input fields, but I would have tried it with less classes. (I might try to get it working in my MR)
> 
> I finally found a way I have been looking for:
> 
> `dom_import_simplexml($element->getData('field_config'))->getNodePath()`
> 
> The Element is set in `Mage_Adminhtml_Block_System_Config_Form_Field_Array_Abstract::_getElementHtml` before `_toHtml` is called
> 
> With that you get the full config path like this one: `/config/sections/design/groups/package/fields/ua_regexp`
> 
> then you can parse that path back again into `[design, package, ua_regexp]`
> 
> So you only need one singular Class instead of multiple ones.
> 
> Or a different approach would be like `Mage_Adminhtml_Block_System_Config_Form_Fieldset_Modules_DisableOutput`, where the renderer is set on the Fieldset and the Entire list of Options like `script-src` and `script-src` are Generated

If you open a PR on my [Mage_Csp branch](https://github.com/empiricompany/openmage/tree/Mage_Csp), I can test and review it.

empiricompany:
I added the ability to split headers for each CSP directive.

Hanmac:
I need to experiment a bit with that, because my field_config idea doesn't really work with the bigger Fieldset one
but maybe one step at a time

(it's kinda funky that the backend model and the source model gets to know the path, but the element and renderer does not)

Hanmac:
@sreichel maybe in a smaller MR,
it would be better to make `Mage_Adminhtml_Block_System_Config_Form`
give the field_path/config_path similar to
`'field_config'          => $element,` ?

Like it does with Backend and Source Model?

empiricompany:
sorry @Hanmac you can create a fork of this PR please?

empiricompany:
@Hanmac It makes sense for me to review your changes before merging them into my PR.

I just checked out your changes and everything broke.

`Fatal error: Uncaught Error: Cannot instantiate abstract class Mage_Adminhtml_Block_System_Config_Form_Field_Csp_Hosts`

empiricompany:
The issue is that the `Mage_Adminhtml_Block_System_Config_Form_Field_Csp_Hosts` class remained abstract in your changes.

However, I prefer my more explicit and clearer solution. For me, this code segment isn't straightforward to understand:

```php
/** @var Varien_Data_Form_Element_Abstract $element */
$element = $this->getElement();
// NodePath: /config/sections/csp/groups/frontend/fields/default-src
$nodePath = dom_import_simplexml($element->getData('field_config'))->getNodePath();
[,,,,,$area,,$directiveName] = explode('/', $nodePath);

**Comment by empiricompany** [comment]
Another valid reason is that this approach would allow us to discuss and improve your solution, since for example any directive declared in `config.xml` could potentially be injected into `getArrayRows()`.

**Comment by sreichel** [comment]
"Thanks"

**Comment by empiricompany** [comment]
> Then we could have discuss why you didn't do that

@Hanmac It's a good idea but this is my pull request - let's avoid arrogance, please. I've modified your code to only allow specific directives and areas.

**Comment by Hanmac** [comment]
> @Hanmac It's a good idea but this is my pull request

Then why didn't do make an PR on my one too?

**Comment by sreichel** [comment]
You work on the same idea. 

Stick your heads together. Damn.

**Comment by Hanmac** [comment]
> Another valid reason is that this approach would allow us to discuss and improve your solution, since for example any directive declared in `config.xml` could potentially be injected into `getArrayRows()`.

I would have done it the other way around with the Fieldset idea
doing using CSP_DIRECTIVES, then the different directive wouldn't be needed anymore

**Comment by empiricompany** [comment]
> You work on the same idea.

This isn't true - it's a completely different approach. Introducing new bugs by committing directly to my PR forces me to fix them and imposes your idea over mine. That's not correct. This isn't the FOSS spirit.

**Comment by sreichel** [comment]
> Let's stay calm @sreichel

Of course, but i dont understand your dicussion.  Sloooooow down a little bit? (both)

**Comment by sreichel** [comment]
btw ... i did not check the commits, but it should not be pushed directly to another repo, but .... didnt you made the same mistake some days ogo? not?

**Comment by empiricompany** [comment]
> btw ... i did not check the commits, but it should not be pushed directly to another repo, but .... didnt you made the same mistake some days ogo? not?

I don't understand you and you're creating too much controversy. I'm working on my repository like any other PR.

Constructive suggestions for this PR are welcome. If it's good as-is, we can merge it.
For everything else, I don't want further discussion.

**Comment by sreichel** [comment]
> I don't understand you

Same here. I agreed with you .... 

... and i dont understand the discussion between you and Hans. Same idea/goal with different a approach ... 

(you should work together ... thats all i wanted to say)

**Comment by colinmollenhour** [comment]
These are both big PRs so it is hard to review and compare without a deep dive.. @Hanmac you are very clearly credited in the description of the PR, does it matter whose PR is finally merged? If you truly think this is a worse PR than yours then just state the reasons so everyone can understand and help decide. But, if you think it is generally an improvement *or at least can live with the changes* and recognize that to some users it may be an improvement, then just get onboard and help iron out the issues on this PR and close yours - everyone will still recognize the hard work you put into laying the foundation and getting it to completion.

**Comment by sreichel** [comment]
PhpStran Level 10 ... https://github.com/OpenMage/magento-lts/commit/8b501d87d611f42376acd10c68c0318194046ca5

Please check commented lines.

File: app/code/core/Mage/Csp/Block/Meta.php (hunk 4 above)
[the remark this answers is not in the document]
Comment (empiricompany) [reply app/code/core/Mage/Csp/Block/Meta.php]: In the end, we'll do it your way.

File: app/code/core/Mage/Csp/Model/Observer/Abstract.php [hunk 6]
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Common method to add CSP headers for a specific area
+     *
+     * @param Mage_Core_Model_App_Area::AREA_FRONTEND|Mage_Core_Model_App_Area::AREA_ADMINHTML $area
+     * @throws Zend_Controller_Response_Exception
+     */
+    protected function addCspHeaders(Varien_Event_Observer $observer, string $area): void
+    {
+        /** @var Mage_Core_Controller_Response_Http $response */
+        $response = $observer->getEvent()->getDataByKey('response');
```
empiricompany, line 23: @sreichel Why in your patch do you want to use getDataByKey('response') instead of getResponse()?

empiricompany:
I've added a default setting to convert and merge all `<meta>` directives into **HTTP headers**. The description is located at the point "*Ability to add directives from views*".

This allows adding:
```xml
<checkout_onepage_index>
    <reference name="csp_meta">
        <action method="addDirective"><directive>default-src</directive><value>*.default-from-layout-update.com</value></action>
        <action method="addDirective"><directive>script-src</directive><value>*.script-from-layout-update.com</value></action>
        <action method="addDirective"><directive>style-src</directive><value>*.style-from-layout-update.com</value></action>
    </reference>
</checkout_onepage_index>

**Comment by Hanmac** [comment]
@empiricompany the only tricky one is this Route: `adminhtml_index_login`
Because it doesn't really have a Head-Block you can access. (and can't really add meta tags)

But you could add it to `form.additional.info` if shouldMergeMeta is enabled?

**Comment by Hanmac** [comment]
> @Hanmac For the `adminhtml_index_login` and `adminhtml_index_forgotpassword` handles, if I add it to `form.additional.info` when in `<meta>` mode, the meta tag output is incorrectly added to the `<body>`.

Yeah, that's what i mean.

But if i would enable, `shouldMergeMeta` then the Update should be moved to the Header

File: app/code/core/Mage/Csp/Model/Observer/Abstract.php (hunk 6 above)
Comment (sreichel) [reply app/code/core/Mage/Csp/Model/Observer/Abstract.php:23]: `getResponse()` is a magic method that finally calls Varien_Object::getData() with some unnecessary checks. getDataByKey() is avoids some overhead.

**Comment by sreichel** [comment]
Please grant permission to create PRs. I'd like to add some tests.

Two small bugs in admin config:

- the is no form validation, so its possible to save empty input
- `self`  or `data:`  appears twice.  (global and ...)

**Comment by sreichel** [comment]
Some code could by shorter, but .... https://youtrack.jetbrains.com/issue/WI-81337/Argument-type-does-not-match-the-declared-when-using-class-alias

File: app/code/core/Mage/Csp/Model/Observer/Abstract.php [hunk 7]
```diff
@@ -0,0 +1,97 @@
...
+
+        $directives = $helper->getPolicies($area);
+        if (empty($directives)) {
+            return;
+        }
+
+        // Merge meta directives if needed
+        if ($helper->shouldMergeMeta($area)) {
+            $blockCspMeta = Mage::app()->getLayout()->getBlock('csp_meta');
+            if ($blockCspMeta && $blockCspMeta instanceof Mage_Csp_Block_Meta) {
```
sreichel, line 42:
```suggestion
            if ($blockCspMeta instanceof Mage_Csp_Block_Meta) {
```

empiricompany:
> Please grant permission to create PRs. I'd like to add some tests.
> 
> Two small bugs in admin config:
> 
> - there is no form validation, so its possible to save empty input
> - `self`  or `data:`  appears twice.  (global and ...)

sure, you can create PR on my branch

sreichel:
- [ ] revert visibility change from private to protected, add new private property?
- [ ] avoid new entries in phpstan baseline
- [ | add form validation, avoid empty/duplicated entries 
- [ ] see comment from @colinmollenhour, do we need all that default value in xml
- [ ] chore: update labeler config
- [ ] tests: add some tests (unit/e2e)

My checklist for next days :)

sreichel:
> sure, you can create PR on my branch

Mhh, i cant select your repo.

om-fork-push

sreichel:
See latest commits https://github.com/sreichel/magento-lts/tree/Mage_Csp

empiricompany:
strange, if you look there is also a PR of @Hanmac that I have just merged

Hanmac:
> > sure, you can create PR on my branch
> 
> Mhh, i cant select your repo.


I somehow had the same problem, but i don't remember how i fixed it.

I think I changed it to myRepo:Mage_Csp <= empiricompany:Mage_Csp,
and then told github to reverse the comparison 

@sreichel you can try to click on this Link: https://github.com/empiricompany/openmage/compare/Mage_Csp...sreichel:magento-lts:Mage_Csp

empiricompany:
@sreichel simply fork my branch https://github.com/empiricompany/openmage/tree/Mage_Csp and create a pull request
This branch is not behind the upstream OpenMage/magento-lts:main

app/code/core/Mage/Csp/Model/Observer/Abstract.php
empiricompany: It's an unnecessary over-optimization. I prefer to keep clearer and more fluent code. I haven't found other code like this. In observers, like in `Mage_Wishlist_Model_Observer::processAddToCart()`, `getResponse()` is always used. Maybe in the future, a magic method `getResponse()` could return the correct type as well.

app/code/core/Mage/Csp/etc/config.xml
empiricompany: I am unable to implement a "presets" feature; it can be added in the future. I propose creating detailed documentation for the new CSP module and providing examples for most players.

sreichel:
@empiricompany thats what I did. I'll test it again.

app/code/core/Mage/Csp/Model/Observer/Abstract.php
sreichel: getDataByKey() comes from M2 and was added 8 month ago, so it's only used in a few places atm. But I've added it to avoid that overhead. It's not an over-microoptimization, it's clearly faster.

Hanmac (approved):

sreichel (approved):

sreichel:
@empiricompany @Hanmac can i merge?

I tried to update Sonar setting, but its not possible for free plan. We have to bypass this check.

Hanmac:
> @empiricompany @Hanmac can i merge?
> 
> I tried to update Sonar setting, but its not possible for free plan. We have to bypass this check.

@sreichel i would say "go"

Sonar is such a crybaby,
and the only problem is one duplicated file.

Issue #4778, opened 2025-04-25
[BUG] Incorrect product option price calculation in new orders in the admin console


massa-man:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When creating a new order in magento admin console for a product that includes custom options, the total product price is incorrectly calculated using the option price for the default store instead of the option price of the new order store.

The product total in the products list is correctly calculated using the correct option price for the order store, but the order totals (the subtotal specifically) is calculated using the product option price for the admin (default) store

### Expected Behavior

I would expect the subtotal of the order to match the correct amount calculated using the order store for the option prices

### Steps To Reproduce

1. In a multi-store magento installation 
2. Create a product that has custom options with different FIXED prices in different websites/stores
3. In the admin console, create a new order in a specific store
4. Add the product created to the order and configure the product to add the option(s) with specific prices
5. The product list section shows the correct total for the product and option in the order store
6. However, the order subtotal in the totals section is calculated with the product price in the order store but the option price in the admin store

<img width="1266" alt="Image" src="https://github.com/user-attachments/assets/69321ace-561e-4f6d-9976-98ac9bae715b" />

<img width="1213" alt="Image" src="https://github.com/user-attachments/assets/8ee9f835-000d-4aac-9066-aae1a30a6f4c" />

<img width="605" alt="Image" src="https://github.com/user-attachments/assets/7e951e6c-7e1c-4dbc-81be-c55ccb1c7c7a" />

### Environment

```markdown
- OpenMage: 20.12.2
- php: 8.2
```

massa-man:
I have done some investigation on this issue and the problem is the following:

When the quote items are loaded as a collection, it assigns the products to the quote items, but for some reason magento is using the application store when loading the products into the quote items collection, instead of the quote store, loading the incorrect Product option prices and product option value prices (for multi select options)

This is the trace:

1. When working with the quote items, we use `Mage_Sales_Model_Quote::getItemsCollection()`
2. When we iterate the item collection, magento will load the items in the collection and will call `Mage_Sales_Model_Resource_Quote_Item_Collection::_afterLoad()` as with any other collection object
3. This method calls `$this->_assignProducts()` to load and assign the product to each quote item
4. This method loads a collection of products using the product IDs from the quote items and loads as well the product options for each product, specifically, in line 169, it prepares the product collection and calls `Mage_Catalog_Model_Resource_Product_Collection::addOptionsToResult()`
[...]

PR #4779, opened 2025-04-25
Fix issue with incorrect product option pricing in admin console

massa-man:
Fixes BUG 4778

### Description (*)
Passes the correct store ID to the product options loaded in the product collection, to solve issue https://github.com/OpenMage/magento-lts/issues/4778

- fixes OpenMage/magento-lts#4778

1. In a multi-store magento installation
2. Create a product that has custom options with different FIXED prices in different websites/stores
3. In the admin console, create a new order in a specific store
4. Add the product created to the order and configure the product to add the option(s) with specific prices
5. The product list section shows the correct total for the product and option in the order store
6. However, the order subtotal in the totals section is calculated with the product price in the order store AND the option price in the ADMIN store

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
@@ -1463,7 +1463,7 @@ class Mage_Catalog_Model_Resource_Product_Collection extends Mage_Catalog_Model_
             $productIds[] = $product->getId();
         }
         if (!empty($productIds)) {
-             $storeId = $this->getStoreId();
+            $storeId = $this->getStoreId();
             $options = Mage::getModel('catalog/product_option')
                 ->getCollection()
                 ->addTitleToResult($storeId)
--- a/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
+++ b/app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
@@ -1458,16 +1458,12 @@ class Mage_Catalog_Model_Resource_Product_Collection extends Mage_Catalog_Model_
      */
     public function addOptionsToResult()
     {
-        $storeId = $this->getStoreId();
-        if ($storeId === null) {
-            $storeId = Mage::app()->getStore()->getId();
-        }
-
         $productIds = [];
         foreach ($this as $product) {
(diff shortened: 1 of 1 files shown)


app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
```diff
@@ -1458,17 +1458,22 @@ protected function _addTaxPercents()
...
+        $storeId = $this->getStoreId();
+        if ($storeId === null) {
+            $storeId = Mage::app()->getStore()->getId();
+        }
+
         $productIds = [];
         foreach ($this as $product) {
             $productIds[] = $product->getId();
         }
         if (!empty($productIds)) {
```
kiatng, line 1470:
```suggestion
        $productIds = [];
        foreach ($this as $product) {
            $productIds[] = $product->getId();
        }
        if (!empty($productIds)) {
             $storeId = $this->getStoreId();
```

1. There is no need to check if `$storeId` is `null`, see https://github.com/OpenMage/magento-lts/blob/7fa24378f5601b513092480ff76044377287f065/app/code/core/Mage/Catalog/Model/Resource/Collection/Abstract.php#L51-L71
2. Declare var close to where it is needed, within the `if` condition. It is not needed outside of the condition.

massa-man:
Thanks for the suggested edits @kiatng , I have accepted them into the PR

app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
```diff
@@ -1463,12 +1463,13 @@ public function addOptionsToResult()
...
+             $storeId = $this->getStoreId();
```
kiatng, line 1466:
```suggestion
            $storeId = $this->getStoreId();
```

kiatng: Extra space. Note that if you allow PE editing, it will help to merge your PR.

kiatng: Also note that PHP VS Fixer picked up on this resulted in failed run.

kiatng (approved):

empiricompany (approved):

PR #4799, opened 2025-05-05
Added `empiricompany/openmage_ignition` as dependency

sreichel:
I am not a fan of adding more and more features, but this should be in by default.

https://github.com/empiricompany/openmage_ignition/

(and ... we have added an event to make that work. use it :))


empiricompany:
@sreichel I don't understand this and other PRs opened and closed immediately after.

sreichel:
Something went wrong, see commit history. There will be a new PR.

sreichel:
> @sreichel I don't understand this and other PRs opened and closed immediately after. 

Erm, when you agree with that PR, please add it on your own 😛

Why should I add your code? 👍

PR #4811, opened 2025-05-14
workflow: update check-files.yml

sreichel:
... fixed typo.

Spellcheck should work now for #4810


sreichel:
Check for phtml seems also wrong ...

PR #4828, opened 2025-06-09
Allow empty qty-item field when creating a shipment

loekvangool:
While creating a shipment, one has to put a `0` into every order line that is out of scope for this shipment, empty values are blocked by JS, while the PHP code can take empty values without issue.

This PR allows simply clearing the `qty-item` field.

1. Try and see if there are any logs, errors, or functional bugs


sreichel (approved):
Works for me.

addison74 (approved):

PR #4834, opened 2025-06-20
rector: `RemoveUnusedVariableInCatchRector`

sreichel:
See https://getrector.com/rule-detail/remove-unused-variable-in-catch-rector

### DIFF
--- a/app/Mage.php
+++ b/app/Mage.php
@@ -703,7 +703,7 @@ final class Mage
             } else {
                 self::$_app->init($code, $type, $options);
             }
-        } catch (Mage_Core_Model_Session_Exception $e) {
+        } catch (Mage_Core_Model_Session_Exception) {
             header('Location: ' . self::getBaseUrl());
             die;
         } catch (Mage_Core_Model_Store_Exception $e) {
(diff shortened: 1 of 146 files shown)


sreichel:
Pls ignore Sonar.

Hanmac:
i'm more curious why the Exceptions aren't used, not even logged with `Mage::logException`?
And in most of the cases where i would throw a different exception, i would use the old exception as previous one.

Also since most of Mage core stuff does throw Throwable instead of Exception now,
wouldn't it be better to catch Throwable instead, or is that too dangerous?

sreichel:
I think there is some space for improvements. Want to oen a new issue or discussion for that?

app/Mage.php
```diff
@@ -703,7 +703,7 @@ public static function init($code = '', $type = 'store', $options = [], $modules
             } else {
                 self::$_app->init($code, $type, $options);
             }
-        } catch (Mage_Core_Model_Session_Exception $e) {
+        } catch (Mage_Core_Model_Session_Exception) {
             header('Location: ' . self::getBaseUrl());
             die;
         } catch (Mage_Core_Model_Store_Exception $e) {
```
kiatng, line 709: I do not understand the rationale of this PR. Why `$e` isn't remove here?

sreichel: I don't know. Think I have to report it to rector.

sreichel: Maybe it is passed to next catch-statement. I dont know. Ignore it for now?

sreichel:
Pls ignnore sonar.

(in free version it is not possible to adjust that limit)

sreichel:
Merged with one green.

PR #4835, opened 2025-06-20
rector: `ChangeSwitchToMatchRector`

sreichel:
See https://getrector.com/rule-detail/change-switch-to-match-rector

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1240,26 +1240,14 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
         $codePool = (string) $this->getModuleConfig($moduleName)->codePool;
         $dir = $this->getOptions()->getCodeDir() . DS . $codePool . DS . uc_words($moduleName, DS);
 
-        switch ($type) {
-            case 'etc':
-                $dir .= DS . 'etc';
-                break;
-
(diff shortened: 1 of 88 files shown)


addison74 (dismissed):

app/code/core/Mage/Core/Model/Config.php
```diff
@@ -1240,26 +1240,14 @@ public function getModuleDir($type, $moduleName)
...
-                $dir .= DS . 'locale';
-                break;
-        }
+        match ($type) {
+            'etc' => $dir .= DS . 'etc',
+            'controllers' => $dir .= DS . 'controllers',
+            'sql' => $dir .= DS . 'sql',
+            'data' => $dir .= DS . 'data',
+            'locale' => $dir .= DS . 'locale',
+            default => str_replace('/', DS, $dir),
```
Hanmac: For default, isn't str_replace called twice?

kiatng:
`match()` was introduced in PHP 8.0.0.  Are we dropping support for PHP7.4?

>There is currently no polyfill package that can provide the match() expression from PHP 8.0 for earlier PHP versions.

sreichel:
> Are we dropping support for PHP7.4?

We dropped 7.4 support with 20,14.0 release. (https://github.com/OpenMage/magento-lts/pull/4124) ;)

Hanmac (approved):

PR #4868, opened 2025-07-06
rector: `InlineArrayReturnAssignRector`

sreichel:
- see https://getrector.com/rule-detail/inline-array-return-assign-rector


sreichel:
Safe to merge.

PR #4875, opened 2025-07-06
rector: `SimplifyEmptyArrayCheckRector`

sreichel:
- see https://getrector.com/rule-detail/simplify-empty-array-check-rector


sreichel:
Merging only the pretty clear PR.

PR #4876, opened 2025-07-07
rector: `EmptyOnNullableObjectToInstanceOfRector`

sreichel:
- see https://getrector.com/rule-detail/empty-on-nullable-object-to-instance-of-rector

### DIFF
--- a/app/code/core/Mage/Admin/Model/Redirectpolicy.php
+++ b/app/code/core/Mage/Admin/Model/Redirectpolicy.php
@@ -39,7 +39,7 @@ class Mage_Admin_Model_Redirectpolicy
         ?Zend_Controller_Request_Http $request = null,
         $alternativeUrl = null
     ) {
-        if (empty($request)) {
+        if (!$request instanceof Zend_Controller_Request_Http) {
             return null;
         }
         $countRequiredParams = ($this->_urlModel->useSecretKey()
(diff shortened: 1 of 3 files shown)


app/code/core/Mage/Admin/Model/Redirectpolicy.php
```diff
@@ -39,7 +39,7 @@ public function getRedirectUrl(
         ?Zend_Controller_Request_Http $request = null,
         $alternativeUrl = null
     ) {
-        if (empty($request)) {
+        if (!$request instanceof Zend_Controller_Request_Http) {
```
kiatng, line 42: Is this better:
```suggestion
        if (is_null(request)) {
```

given line 39:
```
        ?Zend_Controller_Request_Http $request = null,
```

app/code/core/Mage/Shipping/Model/Carrier/Abstract.php
```diff
@@ -176,7 +176,7 @@ protected function _getAllowedContainers(?Varien_Object $params = null)
         if (empty($containersAll)) {
             return [];
         }
-        if (empty($params)) {
+        if (!$params instanceof Varien_Object) {
```
kiatng, line 179: Similarly 
```suggestion
        if (is_null($params)) {
```

app/code/core/Mage/Admin/Model/Redirectpolicy.php
sreichel: I'd prefer instanceof  checks if possible,

app/code/core/Mage/Shipping/Model/Carrier/Abstract.php
sreichel: Guess it would be replaced by https://github.com/OpenMage/magento-lts/pull/4869 rule.

PR #4891, opened 2025-07-21
Allow PHP 8.5

sreichel:
### Questions or comments

phpstan errors are fixed in https://github.com/OpenMage/magento-lts/pull/4889


addison74 (dismissed):

addison74 (dismissed):

.vendor-patches/MAG-1.1.1.patch
```diff
@@ -0,0 +1,11 @@
+--- /dev/null
++++ ../library/Zend/Mime.php
+@@ -38,7 +38,7 @@
+     public const ENCODING_BASE64          = 'base64';
+     public const DISPOSITION_ATTACHMENT   = 'attachment';
+     public const DISPOSITION_INLINE       = 'inline';
+-    public const LINELENGTH               = 72;
++    public const LINELENGTH               = 200;
```
kiatng, line 8: For most use cases in 2025, increasing the line length to 200 is not a concern. But if absolute maximum compatibility is required, especially with legacy systems, you should be aware of the historical 76-character recommendation.

76 characters per line is the best practice for MIME-encoded data to ensure universal interoperability. While modern systems can often accept longer lines, 76 remains the safest and most widely compatible standard.

sreichel: @kiatng i've only moved the patches from remote repo. What is your suggestion here?

kiatng: Sorry, I didn't know these files are auto-generated.

sreichel:
We already had two green flags ... merge.

Issue #4886, opened 2025-07-18
[BUG] Notification Security causes false positive on Proxy with Connection established


Hanmac:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

`curl -I https://domain.tld/app/etc/local.xml`

```
HTTP/1.1 200 Connection established

HTTP/1.1 403 Forbidden
date: Thu, 17 Jul 2025 16:00:15 GMT
strict-transport-security: max-age=16070400; includeSubDomains
content-type: text/html; charset=iso-8859-1
```

This function returns true:
https://github.com/OpenMage/magento-lts/blob/c76a1919e2c3132b50173d755367f2ba32fccdc3/app/code/core/Mage/Adminhtml/Block/Notification/Security.php#L53-L65

### Expected Behavior

`_isFileAccessible` to return false

### Steps To Reproduce

Needs a Proxy that sends `HTTP/1.1 200 Connection established` first

### Environment

```markdown
- OpenMage:20.3.0+
- php: 8.1+
```

### Anything else?

The Core Problem might be `Zend_Http_Response::extractCode`

sreichel:
Its also possible to fix it in `Varien_Http_Adapter_Curl::read()` (?)

Hanmac:
> Its also possible to fix it in `Varien_Http_Adapter_Curl::read()` (?)

I haven't had time yet to debug what exactly is going on
So I don't know if Adapter Curl would have both bodies
Or if it only fails because extractCode returns the first one 

Need to try next week

sreichel:
> Or if it only fails because extractCode returns the first one

Tested. It only returns the first one..

PR #4892, opened 2025-07-21
Remove proxy response code from response body

sreichel:

- fixes OpenMage/magento-lts#4886

### Questions or comments

@Hanmac can please test (i am nut sure about the linebreaks), or close if you have an better idea.

### DIFF
--- a/lib/Varien/Http/Adapter/Curl.php
+++ b/lib/Varien/Http/Adapter/Curl.php
@@ -192,8 +192,14 @@ class Varien_Http_Adapter_Curl implements Zend_Http_Client_Adapter_Interface
             $response = trim($response[1]);
         }
 
-        if (stripos($response, "Transfer-Encoding: chunked\r\n")) {
-            $response = str_ireplace("Transfer-Encoding: chunked\r\n", '', $response);
+        $responsePart = "HTTP/1.1 200 Connection established\r\n";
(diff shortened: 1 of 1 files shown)


lib/Varien/Http/Adapter/Curl.php
```diff
@@ -192,8 +192,14 @@ public function read()
             $response = trim($response[1]);
         }
 
-        if (stripos($response, "Transfer-Encoding: chunked\r\n")) {
-            $response = str_ireplace("Transfer-Encoding: chunked\r\n", '', $response);
+        $responsePart = "HTTP/1.1 200 Connection established\r\n";
+        if (stripos($response, $responsePart)) {
```
Hanmac: Because the response starts with "HTTP/1.1 200 Connection established\r\n",
`stripos` returns 0.

You need to test against `!== false`

sreichel: Or `=== 0`?

Hanmac: It is very likely that it only appears at first position, so === 0 should be fine

we probably don't need to case-insensitive, so `str_starts_with` could be fine too

sreichel: Thought about it, but stripos is also used for ""Transfer-Encoding: chunked"

Issue #4933, opened 2025-09-17
Validate email address for forgottenPassword


pquerner:
When an email is checked if the password reset is allowed or not, the database is asked first if the email address is allowed or not, and only afterwards its checked if the given email address (via POST) is a valid email address.

In my opinion it must be the other way around. And thats what we've seen in some reports:

```
a:5:{i:0;s:338:"SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (utf8_general_ci,IMPLICIT) and (utf8mb4_general_ci,COERCIBLE) for operation '=', query was: SELECT COUNT(*) FROM `customer_flowpassword` AS `main_table` WHERE (`email` = 'can I ask you a question please''����%2527%2522\\\'\\\"') AND (`requested_date` > '2025-09-16 01:01:43')";i:1;s:1821:"#0 /var/www/magento/lib/Varien/Db/Statement/Pdo/Mysql.php(103): Zend_Db_Statement_Pdo->_execute(Array) 
#1 /var/www/magento/app/code/core/Zend/Db/Statement.php(290): Varien_Db_Statement_Pdo_Mysql->_execute(Array) 
#2 /var/www/magento/lib/Zend/Db/Adapter/Abstract.php(480): Zend_Db_Statement->execute(Array) 
#3 /var/www/magento/lib/Zend/Db/Adapter/Pdo/Abstract.php(238): Zend_Db_Adapter_Abstract->query('SELECT COUNT(*)...', Array) 
#4 /var/www/magento/lib/Varien/Db/Adapter/Pdo/Mysql.php(501): Zend_Db_Adapter_Pdo_Abstract->query('SELECT COUNT(*)...', Array) 
#5 /var/www/magento/lib/Zend/Db/Adapter/Abstract.php(828): Varien_Db_Adapter_Pdo_Mysql->query('SELECT COUNT(*)...', Array) 
#6 /var/www/magento/lib/Varien/Data/Collection/Db.php(218): Zend_Db_Adapter_Abstract->fetchOne(Object(Varien_Db_Select), Array) 
#7 /var/www/magento/app/code/core/Mage/Customer/Model/Flowpassword.php(84): Varien_Data_Collection_Db->getSize() 
#8 /var/www/magento/app/code/core/Mage/Customer/controllers/AccountController.php(742): Mage_Customer_Model_Flowpassword->checkCustomerForgotPasswordFlowEmail('can I ask you a...') 
[...]

PR #4934, opened 2025-09-17
#4933 fix: check email validity before attempting to save email or check if password request is allowed

pquerner:
### Description (*)

This pull request moved the email address validity check to the top, which means that invalid email addresses are no longer saved nor checked in the database, before making sure it is even a valid email address.

- fixes OpenMage/magento-lts#4933

1. Go to password reset page in frontend
2. Enter invalid email address (disable frontend javascript validation by removing the CSS classes)
3. Backend should disallow invalid email addresses now

### DIFF
--- a/app/code/core/Mage/Customer/controllers/AccountController.php
+++ b/app/code/core/Mage/Customer/controllers/AccountController.php
@@ -715,6 +715,13 @@ class Mage_Customer_AccountController extends Mage_Core_Controller_Front_Action
     {
         $email = (string) $this->getRequest()->getPost('email');
         if ($email) {
+            if (!Zend_Validate::is($email, 'EmailAddress')) {
+                $this->_getSession()->setForgottenEmail($email);
+                $this->_getSession()->addError($this->__('Invalid email address.'));
+                $this->_redirect('*/*/forgotpassword');
+                return;
+            }
+            
             $flowPassword = Mage::getModel('customer/flowpassword');
             $flowPassword->setEmail($email)->save();
 
@@ -727,13 +734,6 @@ class Mage_Customer_AccountController extends Mage_Core_Controller_Front_Action
 
             if (!$flowPassword->checkCustomerForgotPasswordFlowIp()) {
                 $this->_getSession()->addError($this->__('You have exceeded requests to times per hour from 1 IP.'));
-                $this->_redirect('*/*/forgotpassword');
-                return;
-            }
-
-            if (!Zend_Validate::is($email, 'EmailAddress')) {
-                $this->_getSession()->setForgottenEmail($email);
-                $this->_getSession()->addError($this->__('Invalid email address.'));
                 $this->_redirect('*/*/forgotpassword');
                 return;
             }


app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -715,6 +715,13 @@ public function forgotPasswordPostAction()
     {
         $email = (string) $this->getRequest()->getPost('email');
         if ($email) {
+            if (!Zend_Validate::is($email, 'EmailAddress')) {
+                $this->_getSession()->setForgottenEmail($email);
```
pquerner, line 719: One could argue that we also set the forgotten email back to empty string `""`, to make sure not to possibly break anything on the frontend. But I'll leave it as is for now.

Internally we've set it to empty string, just to make extra sure we dont pass in invalid data back to the frontend - even if its sesssion based and was user input.

kiatng: If I typed an invalid email, I would like to know what was submitted. That's just personal preference. But this change makes sense to me.

kiatng (approved):

kiatng:
I tried to save an invalid email but can't get through the frontend js validation. Can you provide some invalid email addresses for testing?

kiatng:
I tested with a bogus email with valid format; the system saves on every submission beyond the allowed times. Shouldn't we avoid saving beyond the allowed times for this type of emails as well, besides invalid formatted emails?

app/code/core/Mage/Customer/controllers/AccountController.php
Hanmac: An empty string wouldn't even pass the `if ($email) {` check

So I think your current code should be fine
but can you try to fix the PHP-CS-Fixer part?

It wants you to remove the extra spaces before the $flowpassword line

pquerner: It doesnt for the next request, thats true.
Which is why the user must enter the email address again. If you pass in `setForgottenEmail($email);` the email address field on the frontend will be pre-filled with that value. This may or may not be wanted - for invalid email addresses. And because I didnt want to change this, I left this comment up for discussion (or at the very least as information)

pquerner:
> I tested with a bogus email with valid format; the system saves on every submission beyond the allowed times. Shouldn't we avoid saving beyond the allowed times for this type of emails as well, besides invalid formatted emails?

It is true that inserts are done first and then checked afterwards. But I didn't want to touch that logic part.

pquerner:
PHPStan fails, but not because of my changes. No idea what to do in that case. :D Anyone?

Hanmac:
> PHPStan fails, but not because of my changes. No idea what to do in that case. :D Anyone?

I added an extra MR #4937

kiatng:
If the intent of this PR is to lower write amplification: fewer DB writes under spam/abuse, then we need to consider that correctly formatted emails would still get written to the DB until it exceeded 24 hours or per IP limits.

pquerner:
Yes, but please lets do 1 issue 1 pull request. I didnt want to get this all too bloated.

kiatng:
I tasked the AI for a solution, it came up with this:

```php
    public function forgotPasswordPostAction()
    {
        $email = (string) $this->getRequest()->getPost('email');
        if ($email) {
            $flowPassword = Mage::getModel('customer/flowpassword');
            $flowPassword->setEmail($email); // kiat: no save

            if (!$flowPassword->checkCustomerForgotPasswordFlowEmail($email)) {
                // ...
            }

            if (!$flowPassword->checkCustomerForgotPasswordFlowIp()) {
                // ...
            }

            if (!Zend_Validate::is($email, 'EmailAddress')) {
                // ...
            }

            // ...

            try {
                $flowPassword->save(); // kiat: insert save after try before if
                if ($customerId) {
                    /** @var Helper $helper */ // kiat: should remove
                    // ...
                }
            } catch (Exception $exception) {
                // ...
            }
//....
```
I tested it and it works! You can still move the email validation up if desired. The log only saved up to the limit set in the system config.

pquerner:
As I said, this wasnt the intend of this pull request and I feel like it should be done in its own PR.
Of course if you or others feel like it absolutely must be included here, then please go ahead and add it.

addison74:
PHPStan errors.

kiatng (approved):

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -731,33 +737,29 @@ public function forgotPasswordPostAction()
...
                 ->setWebsiteId(Mage::app()->getStore()->getWebsiteId())
                 ->loadByEmail($email);
 
             $customerId = $customer->getId();
-            if ($customerId) {
-                try {
+
+            try {
+                $flowPassword->setEmail($email)->save();
+                if ($customerId) {
```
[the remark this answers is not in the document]
kiatng: This comment is incorrect because the email should be saved regardless if the customer exists in order for the rate limit checks to work.

app/code/core/Mage/Customer/controllers/AccountController.php
```diff
@@ -715,8 +715,14 @@ public function forgotPasswordPostAction()
...
+                $this->_getSession()->setForgottenEmail($email);
+                $this->_getSession()->addError($this->__('Invalid email address.'));
+                $this->_redirect('*/*/forgotpassword');
+                return;
+            }
+
             $flowPassword = Mage::getModel('customer/flowpassword');
-            $flowPassword->setEmail($email)->save();
 
             if (!$flowPassword->checkCustomerForgotPasswordFlowEmail($email)) {
```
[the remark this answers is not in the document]
kiatng: This comment is incorrect because the `setEmail()` has no bearing on the `checkCustomerForgotPasswordFlowEmail()` and therefore can be safely removed.

app/code/core/Mage/Customer/controllers/AccountController.php
sreichel: @pquerner i'll mark it as "resolved" for green flags ....

kiatng: Agree not to set `setForgottenEmail('');`. However, I cannot resolve this comment in order to merge this PR. If you agree, please resolve.

PR #4943, opened 2025-09-27
Handling of disabled CMS pages in System Configuration's Default Pages 

addison74:
This PR addresses [issue #382](https://github.com/OpenMage/magento-lts/issues/382), where disabled CMS pages were listed as selectable options for the "Default Page" under **System > Configuration > Web > Default Pages**.

### **Fix details:**

- The source model for the CMS page dropdown (Mage_Adminhtml_Model_System_Config_Source_Cms_Page) has been updated.
- The dropdown now only lists CMS pages that are enabled (is_active = 1) and are assigned to the currently selected store view.
- This prevents disabled pages from being selected as the home page and ensures correct filtering by store/storeview scope.

### **How the fix works:**

- Filters the page collection using addFieldToFilter('is_active', 1) to only include enabled pages.
- Uses addStoreFilter($storeId) to ensure only pages available for the current store are shown.
- The store ID is detected from the current admin configuration scope.

### **Impact:**

- Admins will only see enabled CMS pages relevant to the selected store when configuring the home page.
- This prevents accidental selection of disabled or mis-scoped CMS pages.

**Credits to Copilot.**


addison74:
It seems that this PR only solves only the situation when ```Store View = All Store View```. If ```Status = Enable``` the page name appears in the list, and if it is ```Disable``` then it does not appear anymore.

It doesn't fix the scenario when ```Store View = {English, French, German}``` and ```Status = Enable```. By changing ```Current Configuration Scope = {English, French, German}```, the page name no longer appears in the list.

One_Enable

Two_Enable

addison74:
It should work correctly now. I have sorted the list ascending. I checked all scenarios for a CMS page and then the scope in the configuration. Please note this dropdown list appears on both the CMS Home Page and the CMS No Cookies Page. As an observation Copilot was unable to provide a viable solution. ChatGPT based on my description solved it immediately. It is surprising, because Copilot uses ChatGPT too.

kiatng:
This solves half the problem. The other half is that when a page is disabled, it needs to check if it is one of the default pages. If it is, throws a notice. Alternatively, on the page, it needs to inform the user the page is used as default and disable the button that disabled the page.
image

addison74:
It seems the PR's we merged today fixed the Reactor issues.

This PR was converted to draft because I have to make more test based on new scenarios. I consider it would be best to display all available pages in those dropdown lists.  Pages that are disabled cannot be selected. This avoids confusion for those who are not experts in OpenMage Backend. I would also insert a warning under the dropdowns informing that certain pages in the list may be disabled and to activate them, the page status must be edited.

The issue reported by @kiatng is implemented and functioning and I will open a PR soon. It is much better there are two PR's for two different issues/discussions.

sreichel:
I am sure I had a PR ready that prevents to disable a CMS page, that is used as default page. Have to search for it.

addison74:
Don't worry it is already done with a lot of features :)

sreichel:
There is only one file changed. Went someting wrong?

addison74:
I am testing this PR again then implement a better version covering all scenarios.

addison74:
It needs more tests and completely other approach. Please don't spend time on it, I will assign both changes to myself.

kiatng:
I apologized for reporting a feature that was already implemented, PR #4239. Anyway, I still think this PR is needed.

addison74:
Please don't worry about. I will take care off this PR. I close it because it needs more features and testing, but fortunately I covered all.

sreichel:
>  it needs more features and testing

This i a great PR to play around with cypress, to test different scenarios.

Give it a try ;)

addison74:
Even it is closed let's share some opinions for the final PR.

## Problem
There are three dropdown lists in the ```Web > Default Pages``` section for **CMS Home Page**, **CMS No Route Page**, and **CMS No Cookies Page**. The current behavior is that any page in the list can be set, although it can be disabled. This creates an error when these pages are used, for example, you won't get the Home Page, but a message that you have not set the page in the Frontend (until the PR #4969 is merged).

The implementation I would like to do takes into account several situations that I have analyzed:

## Hidden or Displayed and Disabled in the list?
- If the page is hidden, then a comment should be visible under the dropdown, which should inform the reason why an existing page would not appear there. This avoid the confusion for some admins.
- If it is displayed it should be disabled, then another phrase in the comment should inform that the page must be edited and the option enabled.

##  What happens if there are a lot of pages even tons?
I have encountered situations with hundreds of pages appearing in the list and scrolling is time consuming (blog with a tree).
- It is important that all the three lists must be sorted ascending.
- Not all existing pages must be used, some could be articles, information pages. In this case an option must be added when editing a CMS page to allow not to be displayed in the list. The default behavior and backward compatibility is to display the page. Basically under ``Status`` [Enabled/Disabled] to add ``Default Page``` [Yes/No], thus the list will be reduced strictly to the pages that have this clear purpose.

## In conclusion
1. If a page is not a Default Page it will not appear in the list. the number of Default pages displayed is limited.
2. If a page is a Default Page and its status is Disabled, then it appears in the list but cannot be selected.
3. The comment message bellow each dropdown will clarify both situation, why not displaying some pages because they are not default, as well as why the reason some are disabled.

I appreciate any feedback. If what I have mention before is enough, I will move on to implementing the change.

kiatng:
@addison74 That is a very detail analysis. Perhaps we should move the configuration of the default pages from the system configuration to the CMS pages, so that everything is in one place, which should simplify things. Perhaps on `!Mage::app()->isSingleStoreMode()`, add a left tab on the edit page to set the default page for store and website. Or add an additional column in the grid to show if a page is used as default page.

PR #4945, opened 2025-09-27
Update rector configuration to skip specific files

addison74:
This pull request makes targeted adjustments to the Rector configuration in `.rector.php` to address processing issues and compatibility with PHP 8.0 features. The main changes involve updating the list of Rector rules being applied and skipping specific files that cause issues during processing.

Configuration updates for Rector rules:

* Removed the `Php80\Class_\AnnotationToAttributeRector` rule from the list of applied Rector rules, likely due to pending compatibility or stability issues with PHP 8.0.

File processing exclusions:

* Added `lib/Varien/Db/Select.php`, `lib/Varien/File/Uploader.php`, and `lib/Varien/Io/File.php` to the list of files to be skipped during Rector processing, addressing specific processing issues with these files.Removed AnnotationToAttributeRector from the list and added Varien files to skip processing.

### DIFF
--- a/.rector.php
+++ b/.rector.php
@@ -121,7 +121,6 @@ try {
             # skip: causes issues
             Php74\Assign\NullCoalescingOperatorRector::class,
             Php80\Catch_\RemoveUnusedVariableInCatchRector::class, # todo: TMP
-            Php80\Class_\AnnotationToAttributeRector::class, # todo: wait for php80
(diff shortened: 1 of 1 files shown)


sreichel:
I think this has to be reported to rector.

addison74:
I tried to update the Rector configuration thinking that only those 3 files were the problem, but with this update even more problems appear. It is obvious that there must be a problem in the configuration and it is not solved by skipping the files. So I will close the PR.

PR #4959, opened 2025-10-01
Fix missing TinyMCE license files + tests

sreichel:
(it will fail for v8)

edit: composer-plugin added license-files for TinyMCE v7.x, but not for v8.x (or later). Fixed in v3.1.


kiatng (approved):

sreichel:
composer-plugin was updated in #4976. 

PR only contain unittest ... merge.

Issue #4908, opened 2025-08-13
[BUG] no login possible post-Update 20.13.0 -> 20.14.0 (and still 20.15.0)


alexh-swdev:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Since the last update from 20.13.0 to 20.14.0, admin login is not working anymore. I get a "Invalid Form Key". It does not matter whether I am using the legacy admin theme or the OM admin theme. The issue is still persisting after 20.14.0 -> 20.15.0 upgrade.

<img width="699" height="306" alt="Image" src="https://github.com/user-attachments/assets/2b299567-ae99-4784-8607-d33cdc75b96b" />

### Expected Behavior

Login is working.

### Steps To Reproduce

1. upgrade
2. go to the backend url
3. try to login

### Environment

```markdown
- OpenMage:20.14.0 / 20.25.0
- php: 8.4
```

alexh-swdev:
Turn's out, 'samesite' is not working as (maybe) expected ....

```
Index: src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php b/src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php
--- a/src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php	(revision 9dc957c7baf1d7ff03e889ed6a7f0a5a07b79413)
+++ b/src/app/code/core/Mage/Core/Model/Session/Abstract/Varien.php	(date 1755097194401)
@@ -106,7 +106,7 @@
             'domain'   => $cookie->getConfigDomain(),
             'secure'   => $cookie->isSecure(),
             'httponly' => $cookie->getHttponly(),
-            'samesite' => $cookie->getSameSite(),
+//            'samesite' => $cookie->getSameSite(),
         ];
 
         if (!$cookieParams['httponly']) {

```

sreichel:
https://github.com/OpenMage/magento-lts/issues/4855

alexh-swdev:
ok...in dev (as on my PC) and local (as "in my home network") staging I don't have https....

sreichel:
Temp solution use DDEV, but o/c it should be fixed.

massa-man:
[...]

Issue #4855, opened 2025-07-02
Install magento-lts-20.14.0 with ip address issue


TangLiang:
From previous closed issue
#4844
I have test many time with different ENV(php 8.4),I install with empty database and clear magento-lts-20.14.0

If install with ip address the frontend and backend cookies will not be set,but with domain will be ok.
I have debug and find where effect and cause this issue.

Compare with magento-lts-20.13.0 as magento-lts-20.13.0 install with ip there is no problem

This is the magento-lts-20.13.0 file

```````````````````````
Mage_Core_Model_Session_Abstract_Varien
`````````````````````````````````
````````````````````````````````
// session cookie params
        $cookieParams = [
            'lifetime' => (int) $cookie->getLifetime(),
            'path'     => $cookie->getPath(),
            'domain'   => $cookie->getConfigDomain(),
            'secure'   => $cookie->isSecure(),
            'httponly' => $cookie->getHttponly(),
        ];

        if (!$cookieParams['httponly']) {
            unset($cookieParams['httponly']);
            if (!$cookieParams['secure']) {
                unset($cookieParams['secure']);
                if (!$cookieParams['domain']) {
                    unset($cookieParams['domain']);
                }
            }
        }

        if (isset($cookieParams['domain'])) {
            $cookieParams['domain'] = $cookie->getDomain();
        }

        call_user_func_array('session_set_cookie_params', array_values($cookieParams));
```````````````````````````````````````````
This is the magento-lts-20.14.0 file
`````````````````````````````````````
$cookieParams = [
            'lifetime' => (int) $cookie->getLifetime(),
            'path'     => $cookie->getPath(),
            'domain'   => $cookie->getConfigDomain(),
            'secure'   => $cookie->isSecure(),
            'httponly' => $cookie->getHttponly(),
            'samesite' => $cookie->getSameSite(),
        ];

        if (!$cookieParams['httponly']) {
            unset($cookieParams['httponly']);
            if (!$cookieParams['secure']) {
                unset($cookieParams['secure']);
                if (!$cookieParams['domain']) {
                    unset($cookieParams['domain']);
                }
            }
        }

        if (isset($cookieParams['domain'])) {
            $cookieParams['domain'] = $cookie->getDomain();
        }

        session_set_cookie_params($cookieParams);
`````````````````````````````````````

[...]

PR #4827, opened 2025-06-06
Use sameSite setting in the session cookie

gschaden:
Session cookies should contain the samesite setting from the configuration

### Manual testing scenarios (*)
1. Open frontend page and check the cookies of the server response

### SonarQube Note
SonarQube may flag the use of insecure cookie settings in this PR. However, these are configuration options—not hardcoded values. Actual security enforcement depends on how the store is configured. These warnings are considered false positives in this context.

### Contribution checklist (*)
 - [x] Pull request has a meaningful description of its purpose
 - [ ] All commits are accompanied by meaningful commit messages
 - [ ] All automated tests passed successfully (all builds are green)

PR #4960, opened 2025-10-02
Fix admin login for non-secure cookies

sreichel:
### Description (*)
Setting cookie SameSite to "None" breaks admin login for unsecured pages.

- fixes OpenMage/magento-lts#4855
- fixes OpenMage/magento-lts#4908

### Related PR
- see OpenMage/magento-lts#4827


sreichel:
Only one notable change: delete param and fallback to browser default.

```php
        if (!$cookie->isSecure() && $cookieParams['samesite'] === CookieSamesite::NONE) {
            // PHP doesn't allow to set SameSite=None for non-secure cookies
            unset($cookieParams['samesite']);
        }
```

Merged.

PR #4963, opened 2025-10-02
Fix and improve the cms page warning message when disabling

addison74:
## Problem

When disabling a CMS page, OpenMage must prevent disabling if the page is used as a default page (Home, No Route, No Cookies) for any store view or website.  
The warning message should clearly indicate for which scopes (Default Config, Website, Store View) the page is in use.  
The grouping and ordering of scopes in this warning message must be consistent and not creating confusion for admins.  

## Implementation

- Refactored the scope grouping logic in `Mage_Cms_Model_Page::_beforeSave()` to consistently use `"Default Config"` as the global scope label, and to properly order Website and Store View labels.
- Ensured that all arrays used for collecting scope usage are associative arrays with string keys, not numeric lists.
- Added a PHPDoc annotation at the `$usage` initialization point to guide PHPStan in type deduction.
- Removed unnecessary type casting and redundant PHPDoc annotations in the loop, keeping the code clean.
- Updated comments to clarify the ordering and grouping logic for future maintenance.

## Solutions

- The warning message now displays the scope usage in the following order:  
  **Default Config**, **Website Name**, **Website Name > Store View Name**

  Example:  
  ```
  This page is used as Home Page in (Default Config, Main Website, Main Website > English, Main Website > French)
  ```
- The delimiters can be easy changed in the source code.
- PHPStan errors related to array key access and type mismatch have been resolved by:
  - Ensuring population of `$usage` only uses string keys.
  - Providing a precise PHPDoc type on `$usage` initialization.
  - Avoiding the use of numeric indexes anywhere in the grouping logic.
- The code and comments are now clearer for future contributors and for static analysis tools.

## Additional Notes

- No changes were made to the underlying logic for preventing disable; only the message formatting and internal grouping were improved.
- The solution is compatible with all Magento LTS versions and does not affect backward compatibility.

## Image for doing the changes 
default_pages

## Image for the message with one change CMS Home Page
disable_message_1

## Image for the message with two changes CMS Home Page
disable_message_2

## Image for the message with three changes CMS Home Page, CMS No Route Page and CMS No Cookies Page
disable_message_3


kiatng:
I tested without this PR:
image

PR #4239 
https://github.com/OpenMage/magento-lts/blob/24cd60d87a3a454b4d901eb9d4acc81fe738320e/app/code/core/Mage/Cms/Model/Resource/Page.php#L76-L87

So, is this PR unnecessary? I apologize for not testing this earlier.

addison74:
@kiatng - you are right, but the default behavior is only valid for that combination, if you make changes  in the Current Configuration Scope > Default Pages you will see that it no longer works. Just change in Default Pages > Madison Island > English the CMS Home Page to Company then disable the Company page and see what is happening.

I have addressed all the scenarios in this PR, including yours.

I will analyze an improvement to this PR https://github.com/OpenMage/magento-lts/pull/4239, which does not solve the problem I reported.

sreichel:
@addison74 the issue has two sides ...

- disable/delete a page that is use in config .... covered with my PR?
- change config to a disabled page.... not covered?

This is a really good PR to write cypress tests, to reproduce different steps.

(edit: my warning could be improved, to name website/store :+1: )

addison74:
@sreichel, @kiatng: Over time it was a mistake to inflate PRs with too many changes. Very few were finalized. That's why I chose to create PRs for each issue separately, to make it easier to review.

This PR is strictly dedicated to disabling a CMS page action. It checks if that page is used in the configuration and if so, then provides where to make the changes. 

I have other PRs in my list:
- the CMS Home Page, CMS No Route Page and CMS Cookies Page lists in the config. I closed a PR that implemented because there are more scenarios to fix.  I am working on that.
- deleting a CMS page that is used in config, I only covered the saving action.
- improving the message in your PR #4329, to have the same format message when trying to change the URL. it should provide where are the places.

In conclusion, I won't add anything because all the other must be resolved in other PRs. I prefer strict PRs, but easily reviewable by anyone.

addison74:
## Before
https://we.tl/t-aft86m68fr

## After
https://we.tl/t-aR5fLHIBJi

@sreichel, @kiatng - please watch these video captures that prove the initial issue, and how this PR solves it.

## As observations:
1. the initial implementation in OpenMage does indeed detect changes, but only for the "Default Config" configuration. the rest are no longer detected and the page cand be saved even it is disabled.

2. notice the messages displayed after running the _beforeSave() method. there are two and they create confusion. one shows that the save was successful, the other that the change cannot be made.

Both observations are resolved in this PR.

Pay attention the links are valid for 3 days from today. There are 500 MB in total, I don't have time to edit to present an inline small version here.

Looking forward for comments related to this PR.

addison74:
I closed it in favor of #4969.

addison74:
@sreichel - I'll let you decide if a cleanup is mandatory for https://github.com/OpenMage/magento-lts/pull/4239, considering that we have 2 open PRs that unify the behavior for the disable, delete and change URL key actions.

kiatng:
I feel that the error message in this PR is a lot more useful for the users. We should incorporate it into #4239.

PR #4967, opened 2025-10-03
Fixed ability to disable or delete CMS pages

kiatng:
### Description (*)
@addison74 discovers a bug in PR #4239
>  if you make changes in the Current Configuration Scope > Default Pages you will see that it no longer works. Just change in Default Pages > Madison Island > English the CMS Home Page to Company then disable the Company page and see what is happening.

### Related Pull Requests
- see OpenMage/magento-lts#4239
- see OpenMage/magento-lts#4963

image

### Questions or comments
To do unit tests?


addison74:
@kiatng - with all due apologies, you misunderstood my reply. I didn't find anything wrong with PR #4239, I just pointed out that it needs to be improved. It works as described and needs a detailed warning message like in #4963.

That quote is the reply to you regarding your report that there would be the implementation from PR #4963 and I replied that I agree with your comment that it provides an warning, but it only covers one situation, they are actually much more. I recommended to check another CMS page like Company in the sample pack and play with its settings in the configuration scope. You will see that you can disable it without any warning.

You need to check the scenarios strictly in the Backend and you will convince yourself that OpenMage before that PR allows saving disabled pages, with one exception, which is the one you received that warning message.

I am ready to review this PR. Please give me more details what it fixes.

kiatng:
Duplicate with PR #4969

PR #4969, opened 2025-10-03
Fix and improve the cms page warning message when disabling or changing the URL key

addison74:
## Problem

When disabling a CMS page or changing its URL Key, OpenMage must prevent the action if the page is used as a default page (Home, No Route, No Cookies) for any store view or website.  
The warning message should clearly indicate for which scopes (Default Config, Website, Store View) the page is in use.  
The grouping and ordering of scopes in this warning message must be consistent and not creating confusion for admins.  

## Implementation

- Refactored the scope grouping logic in `Mage_Cms_Model_Page::_beforeSave()` to consistently use `"Default Config"` as the global scope label, and to properly order Website and Store View labels.
- Ensured that all arrays used for collecting scope usage are associative arrays with string keys, not numeric lists.
- Added a PHPDoc annotation at the `$usage` initialization point to guide PHPStan in type deduction.
- Removed unnecessary type casting and redundant PHPDoc annotations in the loop, keeping the code clean.
- Updated comments to clarify the ordering and grouping logic for future maintenance.

## Solutions

- The warning message now displays the scope usage in the following order:  
  **Default Config**, **Website Name**, **Website Name > Store View Name**

  Example:  
  ```
  This page is used as Home Page in (Default Config, Main Website, Main Website > English, Main Website > French)
  ```
- The delimiters can be easy changed in the source code.
- PHPStan errors related to array key access and type mismatch have been resolved by:
  - Ensuring population of `$usage` only uses string keys.
  - Providing a precise PHPDoc type on `$usage` initialization.
  - Avoiding the use of numeric indexes anywhere in the grouping logic.
- The code and comments are now clearer for future contributors and for static analysis tools.

## Additional Notes

- No changes were made to the underlying logic for preventing disable; only the message formatting and internal grouping were improved.
- The solution is compatible with all OpenMage/Magento LTS versions and does not affect backward compatibility.
Please this PR closes #4963 which disables only. Actually it joins both action Disabling and Changing the URL Key.

## Image for doing the changes 
default_pages

## Image for the message with one change CMS Home Page
disable_message_1

## Image for the message with two changes CMS Home Page
disable_message_2

## Image for the message with three changes CMS Home Page, CMS No Route Page and CMS No Cookies Page
disable_message_3


sreichel:
@copilot should have added type hints and retirn types to new methods.

app/code/core/Mage/Cms/Helper/Data.php
```diff
@@ -79,4 +79,152 @@ public function canUseCanonicalTag($store = null): bool
...
     }
+
+    /**
+     * Returns an array of usages for a CMS page identifier in critical config (Home, No Route, No Cookies)
+     * Each entry describes where the page is used (Default Config, Website, Store View), ordered like in admin config.
+     *
+     * @param string $identifier
+     * @return array
+     */
+    public function getUsageScopes($identifier)
```
sreichel:
```suggestion
    /**
     * Returns an array of usages for a CMS page identifier in critical config (Home, No Route, No Cookies)
     * Each entry describes where the page is used (Default Config, Website, Store View), ordered like in admin config.
     */
    public function getUsageScopes(string $identifier): array
```

app/code/core/Mage/Cms/Helper/Data.php
```diff
@@ -79,4 +79,152 @@ public function canUseCanonicalTag($store = null): bool
...
+     *
+     * Examples:
+     * - ['A']         => 'A'
+     * - ['A', 'B']    => 'A and B'
+     * - ['A', 'B', 'C'] => 'A, B, and C'
+     *
+     * @param array $items Array of strings to join
+     * @return string
+     */
+    public function joinWithCommaAnd(array $items)
```
sreichel:
```suggestion
     * @param array $items Array of strings to join
     */
    public function joinWithCommaAnd(array $items): string
```

sreichel: Could be moved to Mage_Core_Helper_Array

app/code/core/Mage/Cms/Model/Page.php
```diff
@@ -172,4 +172,63 @@ public function isUsedInStoreConfig(?array $paths = []): bool
...
+     * - For URL Key change, blocks if the old identifier is referenced in config and shows detailed warning.
+     *
+     * The list of usages is formatted using Mage_Cms_Helper_Data::getUsageScopes for proper English grammar.
+     *
+     * Throws a Mage_Core_Exception if the page is in use as a default page and disabling or changing URL key is attempted.
+     *
+     * @return Mage_Cms_Model_Page
+     * @throws Mage_Core_Exception
+     */
+    protected function _beforeSave()
```
sreichel, line 190: Could use `sprintf`` for building strings.

app/code/core/Mage/Cms/Helper/Data.php
sreichel: I have to check the code i've used .... It should be similiar. Maybe we can extend it.

app/code/core/Mage/Cms/Helper/Data.php
addison74: I agree that moving the joinWithCommaAnd() method to Mage_Core_Helper_Array is the right approach from an architectural perspective. This function is generic and not specific to the CMS module, so centralizing it in the core array helper makes it more reusable and easier to maintain across the project. I will proceed with this change in a separate commit.

app/code/core/Mage/Cms/Model/Page.php
addison74: Regarding the use of sprintf for building strings in the _beforeSave() method I agree that using this function leads to cleaner and more maintainable code, and it also improves localization support when combined with translation helpers. I will plan to refactor the relevant code to use sprintf for string formatting in a future commit.

app/code/core/Mage/Cms/Helper/Data.php
sreichel: ... or ermove it. Do we really need extra code to spilt by ", and"? Comma-separated would be okay too.

app/code/core/Mage/Cms/Helper/Data.php
sreichel: Code from app/code/core/Mage/Cms/Model/Resource/Page.php could be re-used. (?)

addison74: I have checked it and I can say that it is not implemented similarly. An extent is not necessary, since the solution works. The approval of this PR will make it mandatory to evaluate the remaining methods that are no longer needed, nor the local phrases.

Adding type hints and return types for sure improves code clarity, helps prevent type-related bugs, and aligns with modern PHP standards, making the code easier to maintain and analyze. However, in platforms like OpenMage/Magento 1, introducing strict typing may cause compatibility issues with older extensions or overrides that expect looser type requirements, so this change needs to be carefully considered and tested. Personally, I would prefer not to implement it for the moment.

sreichel: > However, in platforms like OpenMage/Magento 1, introducing strict typing may cause compatibility issues with older extensions or overrides that expect looser type requirements, so this change needs to be carefully considered and tested. Personally, I would prefer not to implement it for the moment.

Thanks for AI answer :P

Adding type hints and return types to new methods cant break anything. We did it for all PRs since months.

sreichel: https://github.com/OpenMage/magento-lts/pull/4923#issuecomment-3339077275

app/code/core/Mage/Cms/Helper/Data.php
addison74: Regarding moving the method joinWithCommaAnd() to Mage_Core_Helper_Array, I would prefer not to do so at this stage, as the function is currently specific to these message constructions. If we find a need for it elsewhere in the future, we can consider relocating it for better reuse.

As for the suggestion to remove the construction using "and" and just use comma separation: while this would simplify the code, I have intentionally chosen to include "and" for grammatical correctness and improved readability, especially when listing two or three items with lots of scopes. Using the conjunction provides clarity and makes the messages more natural for users.

Suggested implementation
Scope 1, Scope 2, Scope 3

Current implementation (using 'and')
Scope 1 and Scope 2                 // for two items
Scope 1, Scope 2, and Scope 3 // for three items

Actually, using only commas is acceptable for short lists, but it does not provide the same clarity for longer ones. Additionally, the delimiter is easily configurable, so anyone who prefers a different implementation can adjust it as needed.

sreichel:
#### disable a page

This breaks current behavior. Other changes made e.g. to content should be saved. (PR for [cypress test](https://github.com/OpenMage/magento-lts/blob/72f50baea639c591e7be8a6629a4178a74ebce95/cypress/e2e/openmage/backend/cms/page.cy.js))

#### change url key

With this PR, how to change the url key for a page, that is used in config?

addison74:
> #### disable a page
> This breaks current behavior. Other changes made e.g. to content should be saved. (PR for [cypress test](https://github.com/OpenMage/magento-lts/blob/72f50baea639c591e7be8a6629a4178a74ebce95/cypress/e2e/openmage/backend/cms/page.cy.js))
> 
> #### change url key
> With this PR, how to change the url key for a page, that is used in config?

Regarding changes to Status and URL key, you are absolutely correct. In the current implementation, any other permitted modifications (such as content updates) will not be saved if the page is used in the ```Web > Default Pages```, which is indeed a regression from previous behavior. I will make the changes to ensure only blocked fields (Status and URL key) are restricted, while all other allowed updates are properly saved.

As for changing the URL key when the page is set as a Default Page in configuration, the solution is already mentioned in the warning message displayed to the user. You simply have to click on the "Default Pages" link provided, and, based on the complete information in the warning, about where the page is used, update your selection in the dropdown list within the configuration. Once the page is no longer set as a Default Page, you will be able to change its URL key.

PR #4970, opened 2025-10-03
Improve CMS Page URL Key comment with implications for reference updates

addison74:
## What does this PR improve?

This PR enhances the comment shown for the "URL Key" field in the CMS Page admin form. The new note informs administrators that changing the URL Key may require updating all references to the page in CMS blocks, widgets, layouts, templates, external/internal links, and any custom extensions or modules.

## Why was this change made?

Previously, the only help text was "Relative to Website Base URL," which is technically accurate but does not warn about breaking references elsewhere in the site or custom code. Modifying the URL Key for a CMS Page can cause broken links or missing content if the identifier is used in:
- CMS blocks (e.g. `{{page url="identifier"}}`)
- Widgets
- Layout XML or templates
- External/internal links (emails, SEO, documentation)
- Custom extensions or modules

Adding this note helps administrators understand the potential impact and avoid unintended issues after changing the URL Key.

## Before

before

## After

after

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Cms/Page/Edit/Tab/Main.php
+++ b/app/code/core/Mage/Adminhtml/Block/Cms/Page/Edit/Tab/Main.php
@@ -57,7 +57,7 @@ class Mage_Adminhtml_Block_Cms_Page_Edit_Tab_Main extends Mage_Adminhtml_Block_W
             'title'     => Mage::helper('cms')->__('URL Key'),
             'required'  => true,
             'class'     => 'validate-identifier',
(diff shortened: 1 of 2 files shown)


sreichel:
This will break comments for all shops that use locale packs.

addison74:
> This will break comments for all shops that use locale packs.

Over time, I have not been a supporter of making changes only in locale files. There are situations where it becomes time consuming to find a message because one is defined in a PHP file and another in a locale CSV file, requiring extra checks to understand what the actual correspondence is between the displayed message and its translation. However, I have been mostly alone in advocating for implementing changes directly in PHP files. Therefore, I will adjust the PR as requested.

PR #4972, opened 2025-10-04
DDEV: allow optional sample data

sreichel:
Allows to install alternative sample data.

Sample Data will be updated frequently to cover more scenarios in tests.


sreichel:
DDEV option only ... merged.

PR #4991, opened 2025-10-13
Fixes for Phpstan 2.1.31 and Rector 2.2.3

sreichel:
Related ...

- closes OpenMage/magento-lts/pull/4984
- closes OpenMage/magento-lts/pull/4985

### DIFF
--- a/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
+++ b/app/code/core/Mage/Usa/Model/Shipping/Carrier/Ups.php
@@ -848,7 +848,6 @@ XMLRequest;
         $costArr = [];
         $priceArr = [];
         if (trim($xmlResponse) !== '') {
-        if (strlen(trim($xmlResponse)) > 0) {
             $xmlResponse = str_replace('<?xmlversion="1.0"?>', '<?xml version="1.0"?>', $xmlResponse);
             $xml = new Varien_Simplexml_Config();
             $xml->loadString($xmlResponse);


app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -338,8 +338,8 @@ public function validateHash($password, $hash)
     public function getVersionHash(Mage_Core_Model_Encryption $encryptionModel)
```
Hanmac, line 338: If the parameter isn't used anymore, maybe deprecate it?

app/code/core/Mage/Adminhtml/Block/System/Convert/Gui/Edit/Tab/Wizard.php
```diff
@@ -92,7 +92,7 @@ public function getValue($key, $default = '', $defaultNew = null)
         }
 
         $value = $this->getData($key);
-        return $this->escapeHtml($value !== null && strlen($value) > 0 ? $value : $default);
+        return $this->escapeHtml($value !== null && (string) $value !== '' ? $value : $default);
```
Hanmac, line 95: What of them rector/phpstan or pslam is against `!empty($value)`?

sreichel: Guess it includes multiple checks that makes it less readable.

app/code/core/Mage/Core/Helper/Data.php
sreichel: Its not an optional parameter.

PR #4997, opened 2025-10-13
rector: `NewlineBeforeNewAssignSetRector`

sreichel:
- see https://getrector.com/rule-detail/newline-before-new-assign-set-rector


sreichel:
Only empty lines. Merged.

PR #4998, opened 2025-10-13
rector: `CountArrayToEmptyArrayComparisonRector`

sreichel:
- see https://getrector.com/rule-detail/count-array-to-empty-array-comparison-rector


sreichel:
Bypass. Merged.

PR #4999, opened 2025-10-13
rector: `RemoveNonExistingVarAnnotationRector`

sreichel:
- see https://getrector.com/rule-detail/remove-non-existing-var-annotation-rector


sreichel:
Docblocks only. Merged.

PR #5010, opened 2025-10-15
Unified creating of common blocks

sreichel:
This prevents calling methods on non existing blocks.

- added some helper methods for common block (rout, head, breadcrumbs)
- deprecated `get Block()` in favor of `getBlockByName()` that returns `null` or existing block, allowing to use PHP´8 null safe operator


Hanmac:
I dislike that the new functions still returns false on missing block.
(that functions can return false instead of null is something I dislike about old magento core)
If they returned null instead, then you could use  `?->` the Null-Safe Operator

```php
    $this->getLayout()->getBlockAdminhtmlHead()?->setCanLoadTinyMce(true);
```

Hanmac:
I'm with copilot there: `Breaking the method chain reduces readability`

sreichel:
Okay, change it.

app/code/core/Mage/Checkout/controllers/Multishipping/AddressController.php
```diff
@@ -142,9 +139,8 @@ public function editAddressAction()
                 ->setErrorUrl(Mage::getUrl('*/*/*', ['id' => $this->getRequest()->getParam('id')]))
                 ->setBackUrl(Mage::getUrl('*/*/selectBilling'));
 
-            if ($headBlock = $this->getLayout()->getBlock('head')) {
-                $headBlock->setTitle($addressForm->getTitle() . ' - ' . $headBlock->getDefaultTitle());
-            }
+            $head = $this->getLayout()->getBlockHead();
+            $head?->setTitle($addressForm->getTitle() . ' - ' . $head->getDefaultTitle());
```
sreichel, line 143:
```suggestion
            $head?->setTitle($addressForm->getTitle() . ' - ' . $head?->getDefaultTitle());
```

Hanmac:
i don't know why `getDefaultTitle` is needed when both `title_prefix` and `title_suffix` exist
🤷‍♂️ 
but that might be a different issue

sreichel:
You are right. "Reference in new issue" and move on?

kiatng (dismissed):

sreichel:
@kiatng set it on draft again.

There are much more blocks that coud be loaded that way, but this would make Mage_Core_Model_Layout bigger and bigger.

How about adding traits for that?

PR #5021, opened 2025-10-22
Remove CSS/JS merge functionality

### LINKED ISSUE #4740: [FEAT] Warning in the backend if css/js files are missing
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Is it possible to show a warning, if expected asset files like js or css are missing when refreshing the caches in the backend?

Background:
Yesterday I realized that since the last update, the admin -> system -> configuration section was unusable. (This was due to a merge error on my side, not from OM core) All other (or at least the daily used) areas in the backend and fronted where still working. The problem was: only there, wrong js urls have been generated for all js-files (<full path on filesystem> instead of (<domain>/<relaive path from doc root>) so they could not get loaded.

I then made a quick&dirty fix to handle this and today I realized that I accidentally removed a required js file. This broke js merging and seemed to create the unusable paths. An error message would have helped a lot :)

### Expected Behavior

If during cache refresh in the backed a css or js file is missing, a warning pops up.

### Benefits

_No response_

### Anything else?

_No response_

sreichel:
- closes OpenMage/magento-lts#4740


Hanmac (dismissed):

app/code/core/Mage/Core/Helper/Data.php
```diff
@@ -1015,7 +909,7 @@ public function getEscapedCSVData(array $data)
                 $value = (string) $value;
 
                 $firstLetter = substr($value, 0, 1);
-                if ($firstLetter && in_array($firstLetter, ['=', '+', '-'])) {
+                if (in_array($firstLetter, ['=', '+', '-'])) {
```
sreichel:
```suggestion
                if ($firstLetter && in_array($firstLetter, ['=', '+', '-'])) {
```

Hanmac (approved):

PR #5024, opened 2025-10-22
Cache attribute values for catalog rules

sreichel:
### Description (*)

@rvelhote thanks for your comment here https://github.com/OpenMage/magento-lts/issues/3127#issuecomment-3419977010

- fixes OpenMage/magento-lts#3127

Bildschirmfoto vom 2025-10-22 20-05-42

Bildschirmfoto vom 2025-10-22 19-53-47

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -4,6 +4,11 @@ categories:
   - title: '🚨 Security'
     labels:
       - 'security'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'fix'
+      - 'bugfix'
+      - 'bug'
   - title: '🚀 Features'
     labels:
       - 'new feature'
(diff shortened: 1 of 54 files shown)


app/code/core/Mage/Catalog/Model/Resource/Product/Collection.php
```diff
@@ -866,7 +873,9 @@ public function getAllAttributeValues($attribute)
             $res[$row['entity_id']][$row['store_id']] = $row['value'];
         }
 
-        return $res;
+        $this->attributesValueCache[$attributeId] = $res;
+
+        return $this->attributesValueCache[$attributeId];
```
Hanmac: I don't know if the checker would like it, but what about:

```php
return $this->attributesValueCache[$attributeId] = $res;
```

sreichel:
```suggestion
        return $this->attributesValueCache[$attributeId] = $res;
```

sreichel:
Merged with one green.

PR #5026, opened 2025-10-24
Fix media gallery actions on duplicate and product deletion procedures

m-michalis:
### Description (*)
Use cases:

> A product with 3 images is being duplicated 10 times. This will result in extra 10 products that all of them use the original product's images, even though the images are being copied. So there are going to be 30 residual images that are not used anywhere.

> When a product gets deleted, it's image files are not being deleted. (probably because of the above in fear)

This commit does:

- [fix] media gallery to properly propagate store view data
- [fix] use new (copied) images on product duplication
- [feature] on product deletion the images will be deleted. Since the product duplication flow uses new images this is now safe. For compatibility, images that are used by other products since this change, will not be affected.

For [fix] parts do the same with/without this commit:
1. Spin up ddev environment with sample data.
2. Duplicate a product.
3. Check the new product's image,thumb,small attributes + whole gallery.
4. Check `<root>/media/catalog/product/<path_to_image(s)>`

For [feature] part (before this commit):
1. Spin up ddev environment with sample data.
2. Duplicate product. (new product will use same images)
3. Apply this commit.
4. Duplicate product. (new product will use new copied images)
5. Delete the duplicated products.
6. Check old product's images that are still there and that the 2nd duplicated product's images are removed.

### Questions or comments
Feedback I guess?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,10 @@
 /app/etc/modules/Cm_RedisSession.xml
 /lib/Credis
 
+# MM Ignition
+/app/code/community/MM
+/app/etc/modules/MM_Ignition.xml
+
 # ChartJs library
 /js/lib/chartjs
 
--- a/app/code/community/MM/Ignition
+++ b/app/code/community/MM/Ignition
@@ -1,1 +1,0 @@
-../../../../vendor/empiricompany/openmage_ignition/app/code/community/MM/Ignition
--- a/app/etc/modules/MM_Ignition.xml
+++ b/app/etc/modules/MM_Ignition.xml
@@ -1,1 +1,0 @@
-../../../vendor/empiricompany/openmage_ignition/app/etc/modules/MM_Ignition.xml


addison74:
I appreciate for your PR. I tried in the past to initiate a discussion in Magento Forum because product duplication + image handling creates real issues: when a product with images is duplicated (I prefer the word Clone for our days than Duplicate), the new product rarely reuses all images and they get removed or replaced, which leaves orphan image files. 

A few simple requests / expectations before a deep code review:

Duplication: when duplicating a product that has images, ask the admin whether to copy the images or not. Simpler: add a configuration option “Include images when duplicating products (Yes/No)” and/or a small prompt on Duplicate when the source product has images. This should be configurable so stores can choose the default behavior.

Removing images from a product: should the physical file be deleted immediately? My preference is to keep files (don’t delete immediately) so an orphan-cleaner utility  (Fballiano's Orphan Image Cleaner) can decide when to remove them. Please make this behavior configurable; if you implement immediate deletion then ensure you only delete when no other DB references exist.

Deleting a product: same as point 2 — prefer to keep files by default (or move them to a trash folder and let a cron/cleaner remove them after N days). If you choose to delete immediately, verify there are zero references before removing the file.

Please split behavior changes into separate PRs if this PR touches both duplication logic and deletion/cleanup logic — they should be reviewed independently. If you want to go further, add tests for these cases (duplicate with/without images, deleting image/product when image is shared) and ensure you check all tables that may reference the media before any physical delete.

m-michalis:
> ...
>
> A few simple requests / expectations before a deep code review:
> 
> Duplication ...
> Removing images from a product ...
> Deleting a product ...
> 
> Please split behavior changes into separate PRs if this PR touches both duplication logic and deletion/cleanup logic — they should be reviewed independently.
> ...

@addison74 Thanks. Will do.

app/code/core/Mage/Catalog/Model/Product/Attribute/Backend/Media.php
```diff
@@ -689,24 +697,28 @@ protected function _copyImage($file)
 
     /**
      * @param Mage_Catalog_Model_Product $object
+     * @param $newId
      * @return $this
      */
-    public function duplicate($object)
+    public function duplicate($object, $newId)
```
sreichel, line 703: Please avoid adding new params.

app/code/core/Mage/Catalog/Model/Resource/Product/Attribute/Backend/Media.php
```diff
@@ -159,47 +159,84 @@ public function deleteGalleryValueInStore($valueId, $storeId)
...
-     * @param array $newFiles
-     * @param int $originalProductId
+     * @param array[] $newImagesMap
+     * @param int[] $attributeIds
+     * @param int $oldProductId
      * @param int $newProductId
      * @return $this
      */
-    public function duplicate($object, $newFiles, $originalProductId, $newProductId)
+    public function duplicate($object, $newImagesMap, $attributeIds, $oldProductId, $newProductId)
```
sreichel, line 168: Same here.

PR #5038, opened 2025-10-27
rector: `FirstClassCallableRector`

sreichel:
- https://getrector.com/rule-detail/first-class-callable-rector


Hanmac:
When touching the callable, maybe lessen this restriction for `filter_condition_callback`?

https://github.com/OpenMage/magento-lts/blob/8e22f8ba87f25576780b485e263a84c62867a757/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php#L572-L574

It stops me to add a column with `filter_condition_callback` via layout updates.

sreichel:
@Hanmac do you mean `instanceof self` check?

Hanmac:
> @Hanmac do you mean `instanceof self` check?

Yeah, that extra `instanceof self` check prevents the callback from being defined in a Helper or extra Model

sreichel:
It was part of a secuity fix and requested to change back in past. Please not.

Hanmac:
> > @Hanmac do you mean `instanceof self` check?
> 
> Yeah, that extra `instanceof self` check prevents the callback from being defined in a Helper or extra Model

It isn't really a security feature

I just want to add a Column via Layout Update and this Helper:
https://github.com/OpenMage/magento-lts/blob/2220264c13150a311bbdba92d3411548d591c26f/app/code/core/Mage/Core/Model/Layout.php#L315

But I can't, because inside the helper, I can't get the right callback because it wants an array like this `[$this, method]`

sreichel:
- see https://github.com/OpenMage/magento-lts/pull/5434

PR #4405, opened 2024-12-07
PHPStan v2

sreichel:
### Description (*)

- updated to phpstan v2
- config cleanup
- removed classes w/ missing interface methods (fatal error)
- some minor fixes

PR #5042, opened 2025-10-27
PhpStan: removed classes with incompatible interface

sreichel:
- see https://github.com/OpenMage/magento-lts/pull/4405


Hanmac:
Probably forgot to remove this lines too?
https://github.com/OpenMage/magento-lts/blob/21f04ef69be4fd17b56e1dd43ccbda7e1a799d06/app/code/core/Mage/Api/etc/api.xml#L116-L119

sreichel:
Not sure if this PR was complete.

Thanks @Hanmac 

Will be back in October ✌️

PR #5048, opened 2025-11-01
PHP8: Allow null hash in validateHashByVersion method

sreichel:
> Exception: Deprecated functionality: password_verify(): Passing null to parameter #2 ($hash) of type string is deprecated  in /var/www/html/app/code/core/Mage/Core/Model/Encryption.php on line 147

Found in unit tests.


sreichel:
Thanks @kiatng.

PR #5065, opened 2025-11-02
phpcs-fixer: `ordered_types`

sreichel:
- see https://cs.symfony.com/doc/rules/class_notation/ordered_types.html


sreichel:
Merged w/o review. Tests only and splitted catch blocks.

PR #5069, opened 2025-11-04
Phpunit: `Mage_Wishlist`

sreichel:
Added some tests.


sreichel:
Merged w/o review. Tests only. (And made one method fluent)

PR #5070, opened 2025-11-06
PhpStan: fix wrong params for `strlen`

sreichel:
... added some Docs, renamed some short vars.


sreichel:
Merge w/o review. Should be safe.

PR #5077, opened 2025-11-07
PhpStan: docblock fixes

sreichel:


sreichel:
Merged as it is 99% docblock updates.

PR #5102, opened 2025-11-15
Replace `Zend_Measure` with `php-units-of-measure`

sreichel:
- see https://github.com/PhpUnitsOfMeasure/php-units-of-measure
- all methods covererd witg unit tests


Hanmac:
Why can't phpstan find: `PhpUnitsOfMeasure\PhysicalQuantity\Length`? 🤔

sreichel:
What do you mean? One commit was not complete, but that should be fixed.

Hanmac (dismissed):

sreichel:
Merged with one green.

If using other lenths/weigths, add them to helper classes and rector.

PR #5103, opened 2025-11-16
Remove call to `widgets.magentocommerce.com`

sreichel:
Dont load icons for admin notification from external url.


sreichel:
Nicer icons, better css ...

... PR welcome.

PR #5117, opened 2025-11-27
Fix wrong tier price displayed when some tier prices are higher than special price

empiricompany:
### Description (*)

This PR fixes a bug where tier prices display incorrect values when a product has a special price that is lower than some tier prices.

**The Problem:**

When a product has tier prices configured and a special price is set, two different filtering mechanisms are applied:

1. **HTML rendering** (`Mage_Catalog_Block_Product_Abstract::getTierPrices()`): Only tier prices lower than the final price are rendered in HTML. Each rendered tier gets a sequential CSS class (`.tier-0`, `.tier-1`, etc.)

2. **JavaScript config** (`Mage_Catalog_Helper_Product_Type_Composite::prepareJsonProductConfig()`): The `tierPrices` array is built from ALL tier prices (via `$product->getTierPrice()`), regardless of whether they are greater than the special price.

This causes an **index mismatch** between the HTML elements and the JavaScript array.

**Example:**
- Product with special price: €12
- Tier 0: €10 for 5 units (shown as `.tier-0` in HTML)
- Tier 1: €15 for 10 units (NOT shown - greater than special price)
- Tier 2: €8 for 20 units (shown as `.tier-1` in HTML)

The JavaScript `tierPrices` array contains 3 elements [diff0, diff1, diff2], but HTML only has 2 elements with classes `.tier-0` and `.tier-1`.

When the JS tries to update `.tier-1` using `tierPrices[1]`, it applies the **wrong tier price difference** (from the hidden tier instead of the third tier).

**The Fix:**

Added filtering in `prepareJsonProductConfig()` to skip tier prices that are greater than or equal to the final price, maintaining consistency with the HTML filtering logic.

### Related Pull Requests
- N/A

- N/A (discovered during development)

1. Create a simple product with base price €20
2. Add tier prices: €15 for qty 5, €18 for qty 10, €10 for qty 20
3. Set special price €16
4. Go to product page in frontend
5. **Before fix:** The tier price for qty 20 shows incorrect price (uses the difference from the hidden €18 tier)
6. **After fix:** The tier price for qty 20 shows correct price €10

### Questions or comments

The same filtering logic from `Mage_Catalog_Block_Product_Abstract::getTierPrices()` (line 416) is now applied in the helper to ensure the JS array indices match the rendered HTML elements.

### DIFF
--- a/tests/unit/Mage/Catalog/Model/ProductTest.php
+++ b/tests/unit/Mage/Catalog/Model/ProductTest.php
@@ -25,6 +25,7 @@ use OpenMage\Tests\Unit\OpenMageTest;
 final class ProductTest extends OpenMageTest
 {
     use BoolTrait;
+
     use CatalogTrait;
 
     private static Subject $subject;
--- a/tests/unit/Mage/Catalog/Model/UrlTest.php
+++ b/tests/unit/Mage/Catalog/Model/UrlTest.php
@@ -23,7 +23,9 @@ use Varien_Object;
 final class UrlTest extends OpenMageTest
 {
     use CatalogTrait;
+
     use IntOrNullTrait;
+
     use UrlTrait;
 
     private static Subject $subject;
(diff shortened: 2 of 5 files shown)


sreichel:
Thanks.

sreichel:
Tested. Merged.

Thanks @empiricompany.

PR #5126, opened 2025-12-03
Replace `Zend_Log` with `monolog`

sreichel:
### Description (*)

Replaces Zend_Log with monolog (https://github.com/Seldaek/monolog).

Added rector rules, so you can rector on your custom code, to replace Zend_Log constants.


sreichel:
Merged after tests.

Hanmac:
I was thinking if we should add `psr/log` too, but that doesn't support filenames.

But I could add a small helper/model class that does connect PsrLog to MageLog.
(using null filename and let it fall back to default)

Hanmac:
I also think there might be a small BC if someone still uses RFC log levels like `LOG_DEBUG`

while Monolog below can detect that and change it to their loglevel, we need to do it ourselves for the `$levelValue > $maxLogLevel` comparison
(sadly, their constant `RFC_5424_LEVELS` isn't public)

I will make a small MR for this

Hanmac:
Another thing is, that we might add some Events for the Monolog\Logger creation.
Or have the Channel-Name (`OpenMage`) be filename specific.

Or another thing, allow special Paths like `php://stdout` and `php://stderr`
(the file-extension check currently blocks this)

sreichel:
> Or another thing, allow special Paths like `php://stdout` and `php://stderr`
> (the file-extension check currently blocks this)

https://github.com/OpenMage/magento-lts/pull/5148/commits/1536dda34a7805f82e000b329cadb4bfbb49803d

Can you please test?

sreichel:
> Another thing is, that we might add some Events for the Monolog\Logger creation.
> Or have the Channel-Name (`OpenMage`) be filename specific.

Add to new issue (feature request) to not forget about it?

PR #5144, opened 2025-12-08
[Core] add PsrLogger

Hanmac:
### Description (*)

* Also adds `Mage_Core_Helper_PsrLogger` as way to implement `LoggerInterface` using Mage Log function

### Related Pull Requests

- see https://github.com/OpenMage/magento-lts/pull/5126#issuecomment-3626104132

1. use `LOG_DEBUG` on `Mage::log`
2. ...


app/Mage.php
```diff
@@ -906,8 +906,24 @@ public static function log($message, $level = null, $file = '', $forceLog = fals
             $levelValue = $level->value;
         } elseif (is_null($level)) {
             $levelValue = Level::Debug->value;
+        } elseif (is_string($level) && !is_numeric($level)) {
+            // PSR 3 Log level
+            $levelValue = Level::fromName($level)->value;
         } else {
             $levelValue = (int) $level;
+            // change RFC_5424 Log Level into Monolog.
```
sreichel: I think thats not needed.

I remember to have tested it and monolog turns PFC5424 levels into psr3 interally.

Hanmac: Yes, later in the Monolog function

But not before comparing `$levelValue > $maxLogLevel`
Where levelValue might be <= 7 (Zend_Log value) and maxLogLevel >= 100 (Monolog value)

sreichel: See it ,...

sreichel (dismissed):

Hanmac:
Not an "accessibility bypass" like SonarQube thinks

sreichel:
Note: there is a PSR-Handler

- not listed here: https://github.com/Seldaek/monolog/blob/main/doc/02-handlers-formatters-processors.md
- see: https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/PsrHandler.php

There a serveral ideas to support mor handlers (and possible replace firegento/logger), Any help is welcome. :)

Hanmac:
> * not listed here: https://github.com/Seldaek/monolog/blob/main/doc/02-handlers-formatters-processors.md
> 
> * see: https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/PsrHandler.php

it is listed there, at the bottom of the handler in the "Wrappers / Special Handlers" section:
> Can be used to forward log records to an existing PSR-3 logger


But this PR is a different direction:
* the PsrHandler from Monolog is for "Monolog -> Psr"
* this PR is for "Psr -> Magento -> Monolog"

Hanmac:
> (and possible replace firegento/logger)

That's what got me thinking first, i added firegento/logger to another project recently.
And after looking into how it works, my first thought was that it wasn't working with new Monolog changes anymore.

sreichel:
Only from reviewing the code ... i think changing log level to monolog values coud cause problems.

Updated PR to use RFC5424 levels.

sreichel:
@Hanmac is it okay to wait for 20.18 that may be released in 1-2 month (maybe earlier)

Hanmac:
> @Hanmac is it okay to wait for 20.18 that may be released in 1-2 month (maybe earlier)

I'm not in a hurry for this

sreichel:
Okay. Want to release 20.17 asap.

sreichel:
Possible to add a test? ( mostly copy&paste from model-test?)

Hanmac:
@sreichel while adding testcases (and trying so mock stuff to the logger),

i noticed this section i copied from the main code of Mage:

```php
$logger = self::getModel('core/logger');
if ($logger instanceof Mage_Core_Model_Logger) {
    $logger->log($message, $level, $file, $forceLog, $context);
}
```

My thought, shouldn't this be a singleton model?
Because right now, each time Mage::log is called, it creates a new Logger Model, with new Monolog Handler created inside.

PR #5145, opened 2025-12-08
Fix docblocks to show correct Mage_Sales_Model_Order_Invoice return / var type

ajbonner:
### Description (*)
In Mage_Checkout_OnepageController the return docblock for _initInvoice() shows it returning the Mage_Sales_Model_Service_Order when it's returning $invoice which is the return value of Mage_Sales_Model_Service_Order::prepareInvoice(array $qtys = []): Mage_Sales_Model_Order_Invoice.

So this PR updates the return type docblock and the @var annotation to reflect the correct type. No executable code changes other than docs.

1. Open app/code/core/Mage/Checkout/controllers/OnepageController.php
2. Navigate to \Mage_Checkout_OnepageController::_initInvoice (line 551)
3. Trace line 559 to the prepareInvoice() invocation and compare types returned by that method vs the @var annotation on line 558. $invoice is then returned which is an Mage_Sales_Model_Order_Invoice not a Mage_Sales_Model_Service_Order

 - [*] Pull request has a meaningful description of its purpose
 - [* ] All commits are accompanied by meaningful commit messages
 - [*] All automated tests passed successfully (all builds are green)

### DIFF
--- a/.phpstan.dist.baseline.neon
+++ b/.phpstan.dist.baseline.neon
@@ -1363,12 +1363,6 @@ parameters:
 			path: app/code/core/Mage/Checkout/controllers/CartController.php
 
 		-
-			rawMessage: 'Call to an undefined method Mage_Sales_Model_Service_Order::register().'
-			identifier: method.notFound
-			count: 1
-			path: app/code/core/Mage/Checkout/controllers/OnepageController.php
-
-		-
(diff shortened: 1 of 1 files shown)


ajbonner:
Looking at the PHPStan output, this seems to fix one of the baseline errors as well, so the baseline would need to be updated to remove the fixed failing check.

```Error: Ignored error pattern "Call to an undefined method Mage_Sales_Model_Service_Order::register()." (method.notFound) in path /home/runner/work/magento-lts/magento-lts/app/code/core/Mage/Checkout/controllers/OnepageController.php was not matched in reported errors.```

register() is not  found because PHPStan was looking at Mage_Sales_Model_Service_Order rather than  Mage_Sales_Model_Order_Invoice

sreichel:
Thanks.

PR #5168, opened 2025-12-18
[Core] adds ClockInterface

Hanmac:
### Description (*)

This MR makes `Mage_Core_Helper_Data` implement `Psr\Clock\ClockInterface` to be used for other libraries.

### Related Pull Requests

- see OpenMage/magento-lts#5131

1. ...
2. ...


sreichel:
Time for one (or two) line description?

sreichel:
@Hanmac rdy?

Hanmac:
Hm, I might refactor this MR later to use a new `Mage_Core_Helper_Clock`,
that would make it much cleaner.

Hanmac:
I think I should update the other places where `Carbon::now()->` is used where it gets either directly formatted or formatted into a string

sreichel:
Nice :)

sreichel (dismissed):

sreichel:
`Mage::helper('core/clock')->now()` is used in a lot of models (and controllers) ....

- maybe add a helper method (like `$model->getValidationHelper()` ?
- <s>imho `core/clock` could be changed to `core/time` (???)</s>
- add `getTimestamp()` method (maybe more?)


```
Mage::helper('core/clock')->now()->getTimestamp()
```

```
Mage::helper('core/clock')->getTimestamp()
```

???

Hanmac:
I'm not sure yet in renaming it to `core/time` ... I think it might cause confusion with the `core/date` model

sreichel:
Okay. Not every idea is a good idea. :)

sreichel:
Great :)

> Binary operation "+" between string and int<0, 10> results in an error.

Think there are some operations where `Carbon::now/parse()->addSeconds()` (or similiar can be used ... had not finished that ... unrelated to this error)

Hanmac:
> Great :)
> 
> > Binary operation "+" between string and int<0, 10> results in an error.
> 
> Think there are some operations where `Carbon::now/parse()->addSeconds()` (or similiar can be used ... had not finished that ... unrelated to this error)

These two places work directly with years, for stuff like "2000-2026", probably not something to use a full Carbon/Datetime Object for

sreichel:
Note to myself ... add rector rule to replace Carbon::now ....

Hanmac:
> Note to myself ... add rector rule to replace Carbon::now ....

It depends, I was specific looking for `now()->getTimestamp` and `now()->format`, these don't try to alter the object.

Because `Carbon::now` returns a mutable Carbon `DateTime`,
while the new Helper implementing the ClockInterface returns a `DateTimeImmutable`, which as the name suggests, is immutable. (and without the Carbon helpers)

That's why i didn't touch the other places with `Carbon::now`, for now.

Like this:
https://github.com/OpenMage/magento-lts/blob/18055a0536404e439546bd8f7ba239b84406544e/app/code/core/Mage/Oauth/Model/Resource/Nonce.php#L41

sreichel:
> It depends, I was specific looking for `now()->getTimestamp` and `now()->format`, these don't try to alter the object.

Rector should only replace these two.

sreichel (approved):

sreichel:
Rdy to merge?

addison74 (approved):

PR #5198, opened 2025-12-31
fix patches for cweagans/composer-patches v2

M-arcus:
patch header fixed for PR https://github.com/OpenMage/magento-lts/pull/5121


sreichel:
@all-contributors add M-arcus code

PR #5222, opened 2026-01-09
rector: `RemoveUnusedVariableAssignRector`

sreichel:
- see https://getrector.com/rule-detail/change-if-else-value-assign-to-early-return-rector


.phpstan.dist.baseline.neon
```diff
@@ -1164,6 +1152,12 @@ parameters:
 			count: 4
 			path: app/code/core/Mage/Adminhtml/controllers/Sales/OrderController.php
 
+		-
+			rawMessage: Variable $order might not be defined.
+			identifier: variable.undefined
+			count: 1
+			path: app/code/core/Mage/Adminhtml/controllers/Sales/OrderController.php
```
Hanmac, line 1159: Probably should fix this? otherwise, it could crash?

PR #5254, opened 2026-01-27
fix(security): Remove phantom checkbox behavior in ACL role editor

mattdavenport:
When new ACL resources are added to the system, roles without explicit rules for those resources would show them as "checked" in the UI based on parent permission inheritance. However, the actual runtime permission check uses different logic (falls back to null/root), so users may not actually have access.

This creates a security risk: when an admin edits a role and saves it without unchecking these phantom checkboxes, the permissions get explicitly granted - potentially giving users access to features they were never intended to have.

The fix removes the fallback inheritance logic in the UI. New resources will now appear unchecked by default, requiring explicit selection by the admin. This prevents accidental permission grants.


sreichel:
@mattdavenport nice. :+1:

Issue #5252, opened 2026-01-27
[FEAT] @return $this if able


Hanmac:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Functions like this, that returns `$this` or `patent::` like this:

https://github.com/OpenMage/magento-lts/blob/892e482c895b638c3379ddbed4dd8ace6fd6a823/app/code/core/Mage/Uploader/Block/Single.php#L18-L26

https://github.com/OpenMage/magento-lts/blob/892e482c895b638c3379ddbed4dd8ace6fd6a823/app/code/core/Mage/Uploader/Block/Abstract.php#L117-L155

should return `@return $this` or `: self` if able

### Expected Behavior

Returns `$this`

### Benefits

My coworker found phpstan warnings, but i currently can't find where openmage would ignore them

### Anything else?

I don't know if there is a Rector for this to check

PR #5266, opened 2026-02-11
Return $this if able

Hanmac:
### Description (*)

i don't know if i got all of them, but i changed them to `@return $this` or `@inheritDoc` if able

### Related Pull Requests

- fixes OpenMage/magento-lts#5252

1. ...
2. ...

### DIFF
--- a/app/code/core/Mage/Page/Block/Template/Links/Block.php
+++ b/app/code/core/Mage/Page/Block/Template/Links/Block.php
@@ -136,9 +136,8 @@ class Mage_Page_Block_Template_Links_Block extends Mage_Core_Block_Template
     /**
      * Set last list flag
      *
-     * @param  bool                                 $value
-     *                                                     return Mage_Page_Block_Template_Links_Block
-     * @return Mage_Page_Block_Template_Links_Block
+     * @param  bool  $value
+     * @return $this
      */
     public function setIsLast($value)
     {


sreichel:
Love it.

app/code/core/Mage/Sales/Block/Adminhtml/Customer/Edit/Tab/Agreement.php
Hanmac, line 109: i'm not 100% sure about this change

my PHPStrom was complaining that `Mage_Adminhtml_Block_Widget_Grid` is not `$this`
even all methods `return $this`

maybe need to revert this one🤔

sreichel: imho ... i have added a lot of docblocks and not all of them are 100% correct.

if it helps to impove fluent interface id keep. if it wrong, we still can fix it later.

Hanmac:
I also noticed that a clean `: $this` return type is still not really in PHP: https://wiki.php.net/rfc/this_return_type

and a small example like this causes errors on psalm:

```php
<?php declare(strict_types = 1);

class A {
    /**
     * @return $this
     */
	public function func(): static
	{
		return $this;
	}
}
class B extends A {
	#[Override]
	public function func(): static
	{
		return parent::func();
	}
}
class C extends B {
	#[Override]
    /**
     * @return $this
     */
	public function func(): static
	{
		return A::func();
	}
}
```


https://psalm.dev/r/940e0d520f

kiatng (approved):

PR #5272, opened 2026-02-17
chore: composer update

sreichel:
... and some fixes for rector/phpstan

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Newsletter/Grid/Filter/Status.php
+++ b/app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Newsletter/Grid/Filter/Status.php
@@ -19,7 +19,7 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Newsletter_Grid_Filter_Status exten
     public function __construct()
     {
         self::$_statuses = [
-            null                                        => null,
+            ''                                          => null,
             Mage_Newsletter_Model_Queue::STATUS_SENT    => Mage::helper('customer')->__('Sent'),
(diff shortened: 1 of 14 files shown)


app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Newsletter/Grid/Filter/Status.php
```diff
@@ -19,7 +19,7 @@ class Mage_Adminhtml_Block_Customer_Edit_Tab_Newsletter_Grid_Filter_Status exten
     public function __construct()
     {
         self::$_statuses = [
-            null                                        => null,
+            ''                                          => null,
```
Hanmac, line 22: Does this change work with the Condition function below?
It would check against empty string instead of null

sreichel: https://www.php.net/manual/en/language.types.array.php

> [Null](https://www.php.net/manual/en/language.types.null.php) will be cast to the empty string, i.e. the key null will actually be stored under ""

app/code/core/Mage/Adminhtml/Block/Customer/Edit/Tab/Newsletter/Grid/Filter/Status.php
Hanmac, line 55: This function is equal to `Mage_Adminhtml_Block_Widget_Grid_Column_Filter_Select::getCondition` and could be removed?

sreichel: Same for `Mage_Adminhtml_Block_System_Email_Template_Grid_Filter_Type`.

Next PR.

PR #5285, opened 2026-03-01
Fix: save newsletter template

sreichel:
### Description (*)

Cant save existing newsletter template b/c "template_type" has to be an integer value, string given.

- added getter/setter
- added save success message
- added cypress test

### Related Pull Requests

- see OpenMage/magento-lts#4612

1. go to admnin - newsletter - template
2. save an existing template


app/code/core/Mage/Newsletter/Model/Template.php
```diff
@@ -426,4 +443,18 @@ public function getTemplateText()
...
+    /**
+     * @param  Mage_Core_Model_Template::TYPE_* $type
+     * @return $this
+     */
+    public function setTemplateType(int $type)
+    {
+        return $this->setData('template_type', $type);
+    }
+
+    public function getTemplateType(): int
```
B3Hana:
```suggestion
    public function getTemplateType(): ?int
```
This functon is called in `app/design/adminhtml/base/default/template/newsletter/template/edit.phtml` while the model is new.

app/code/core/Mage/Adminhtml/controllers/Newsletter/TemplateController.php
```diff
@@ -164,6 +164,7 @@ public function saveAction()
             }
 
             $template->save();
+            $this->_getSession()->addSuccess(Mage::helper('newsletter')->__('The template has been saved.'));
             $this->_redirect('*/*');
```
B3Hana, line 168:
```suggestion
            $this->_redirect('*/*');
            return;
```
otherwise the message isn't shown, because of the forward.

The forward could also be replaced by a redirect to edit, i guess?
`$this->_redirect('*/*/edit', ['id' => $this->getRequest()->getParam('id')]);`

app/code/core/Mage/Newsletter/Model/Template.php
```diff
@@ -77,31 +75,50 @@ public function validate()
...
             value: $templateType,
-            message: $message,
+            choices: [Mage_Core_Model_Template::TYPE_TEXT, Mage_Core_Model_Template::TYPE_HTML],
+            message: Mage::helper('adminhtml')->__(
+                'The value %1$s you selected for "%3$s" is not a valid choices %2$s.',
+                '{{ value }}',
+                '{{ choices }}',
+                'template_type',
+            ),
         ));
```
B3Hana, line 95: Whats the point of this check? isn't the type hardcoded, changing it uses a flag with a hardcoded const?

B3Hana (changes requested):

app/code/core/Mage/Adminhtml/controllers/Newsletter/TemplateController.php
sreichel: Thanks for review.

I had return in initial commit https://github.com/OpenMage/magento-lts/pull/5285/changes/706d190bfc0ed0b23cde571027f0212f8287abd7

Forwarding ....

I want to unify grids later .... some have no "save & continue" button

save -> index page
s&c -> edit page

app/code/core/Mage/Newsletter/Model/Template.php
sreichel: `validate()` is called from `_beforeSave()`

`$model->setTemplateType(666)->save()` will fail now.

cypress/e2e/openmage/backend/newsletter/template.cy.js
sreichel:
```suggestion
        validation.hasErrorMessage('You must give a non-empty value for field \'template_code\'.', { match: 'have.text' });
```

app/code/core/Mage/Adminhtml/controllers/Newsletter/TemplateController.php
```diff
@@ -164,7 +164,10 @@ public function saveAction()
...
+            return;
+            return;
```
sreichel:
```suggestion
            return;
```

app/code/core/Mage/Adminhtml/controllers/Newsletter/TemplateController.php
sreichel: imho save should redirect to index and save and continue to edit page.

some edit pages have no save and continue button yet

sreichel:
@all-contributors add @B3Hana code

app/code/core/Mage/Newsletter/Model/Template.php
B3Hana: so a general safeguard, makes sense. thanks

B3Hana (approved):

app/code/core/Mage/Newsletter/Model/Template.php
sreichel: Thanks for review!

Hanmac:
`validateType` might not be needed, because `validateChoice` already restricts it to `[1, 2]`

But better safe than sorry?

Hanmac (approved):

sreichel:
> `validateType` might not be needed, because `validateChoice` already restricts it to `[1, 2]`
> 
> But better safe than sorry?

Good one ... not sure/untested if validateChoice is type-safe. (1 vs "1").

sreichel:
Want to add a new test case?

Hanmac:
> > `validateType` might not be needed, because `validateChoice` already restricts it to `[1, 2]`
> > But better safe than sorry?
> 
> Good one ... not sure/untested if validateChoice is type-safe. (1 vs "1").

https://github.com/symfony/symfony/blob/5bcd98749de899a7de2e0052e9234ca028cde867/src/Symfony/Component/Validator/Constraints/ChoiceValidator.php#L100
checks:
```php
} elseif ($constraint->match xor \in_array($value, $choices, true)) {
```

this says that strict makes it type-safe:
https://www.php.net/manual/en/function.in-array.php
> If the third parameter strict is set to true then the in_array() function will also check the types of the needle in the haystack.

so no extra test case needed

sreichel:
PR to remove additional check?

PR #5300, opened 2026-03-09
Fix toggle between region address and non region address on paypal side

leissbua:
### Description (*)

Fix stale `region` / `region_id` values in quote addresses when returning from PayPal Express Checkout.

When PayPal Express returns the shipping or billing address, Magento imports the exported address data using `_setExportedAddressData()`. However, this method only overwrites fields that are present in the exported PayPal address.

If the PayPal address **does not contain a region/state value**, the existing `region` and `region_id` fields on the already loaded quote address remain unchanged. Since quote addresses may already exist in the database, this can result in **stale region values being kept from a previous address**.

Example scenario:

1. Customer starts checkout with address
   `Cologne, NRW, DE`
2. Customer changes the address in PayPal to
   `Berlin, DE` (no region returned by PayPal)
3. Magento imports the address but keeps the existing region
4. Resulting address becomes
   `Berlin, NRW, DE` (incorrect)

This PR fixes the issue by **explicitly resetting `region` and `region_id` on the quote address before importing the PayPal address** when the exported PayPal address does not contain a region.

The reset is applied in `Mage_Paypal_Model_Express_Checkout::returnFromPaypal()` for both shipping and billing addresses.

This ensures that stale region data stored in the quote or database cannot persist when PayPal provides an address without region/state information.

---

### Related Pull Requests

* none

---

* fixes OpenMage/magento-lts#<issue_number>

---

1. Configure PayPal Express Checkout.
2. Start checkout with an address containing a region/state (e.g. `NRW`).
3. Click **PayPal Express Checkout**.
4. In PayPal, change the shipping address to one **without a region/state**.
5. Return to Magento.
6. Verify that the quote shipping and billing addresses **no longer contain the previous region/state**.

Before this fix:

```
Berlin, NRW, DE
```

After this fix:

```
Berlin, DE
```

Also verify that:

* addresses **with a region** are imported correctly
* shipping rates are recalculated normally
* billing address import behaves the same way

---

### Questions or comments

This change only affects the PayPal Express return flow and ensures correct state handling for already-loaded quote addresses. It does not alter PayPal API handling or address import logic outside the Express Checkout return process.

---

* [x] Pull request has a meaningful description of its purpose
* [ ] All automated tests passed successfully (all builds are green, SonarCloud checks are not required to merge)


sreichel (approved):
Untested.

lgtm

sreichel:
@leissbua thanks.

sreichel:
@leissbua Thank you. :heart:

leissbua:
@sreichel Thanks and keep up the good work.

Issue #4979, opened 2025-10-09
[BUG] Mage_SalesRule_Model_Observer->addProductAttributes produces a slow query


EliasKotlyar:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Hello

We are experiencing a performance Problem with the following Query:
Query:
```sql
SELECT DISTINCT ea.attribute_code
FROM `salesrule_product_attribute` AS `a`
 INNER JOIN `eav_attribute` AS `ea` ON ea.attribute_id = a.attribute_id
```
Time: 0.33641290664673 sec

We traced it down to following code:
```php
    public function getActiveAttributes($websiteId, $customerGroupId)
    {
        $read = $this->_getReadAdapter();
        $select = $read->select()
            ->from(
                ['a' => $this->getTable('salesrule/product_attribute')],
                new Zend_Db_Expr('DISTINCT ea.attribute_code'),
            )
            ->joinInner(['ea' => $this->getTable('eav/attribute')], 'ea.attribute_id = a.attribute_id', []);
        return $read->fetchAll($select);
    }
```

It would be great if we could improve performance here. Also we can remove unused parameters. See the following MR

### Expected Behavior

Query would run faster

### Steps To Reproduce

Run any test using sales_flat_rules

### Environment

```markdown
- OpenMage:
- php:
```

### Anything else?

Query has been optimized to:
```sql
SELECT ea.attribute_code
FROM eav_attribute AS ea
INNER JOIN (
  SELECT DISTINCT attribute_id FROM salesrule_product_attribute
) AS a ON ea.attribute_id = a.attribute_id;
```

Which is a slightly faster version.

PR #5315, opened 2026-03-11
Issue 4979

EliasKotlyar:
### Description (*)

Optimizes the SQL query in `Mage_SalesRule_Model_Resource_Rule::getActiveAttributes()` to improve performance.

The original query used `SELECT DISTINCT ea.attribute_code` with a direct `INNER JOIN` between `salesrule_product_attribute` and `eav_attribute`, which resulted in slow execution (~0.34 sec) due to the `DISTINCT` operating on the full joined result set.

The optimized query uses a subselect to first get `DISTINCT attribute_id` from `salesrule_product_attribute`, and then joins this smaller, deduplicated result set against `eav_attribute`. This is faster because the deduplication happens on a single integer column before the join, rather than after.

Additionally:
- Removed unused `$websiteId` and `$customerGroupId` parameters from `getActiveAttributes()`, as they were never used in the query body.
- Changed from `fetchAll()` to `fetchCol()` since only a single column (`attribute_code`) is needed, and updated the caller in `Mage_SalesRule_Model_Observer::addProductAttributes()` accordingly.

**Before (slow):**
```sql
SELECT DISTINCT ea.attribute_code
FROM salesrule_product_attribute AS a
INNER JOIN eav_attribute AS ea ON ea.attribute_id = a.attribute_id
```

**After (optimized):**
```sql
SELECT ea.attribute_code
FROM eav_attribute AS ea
INNER JOIN (
  SELECT DISTINCT attribute_id FROM salesrule_product_attribute
) AS a ON ea.attribute_id = a.attribute_id
```

### Related Pull Requests

- see OpenMage/magento-lts#4979

- fixes OpenMage/magento-lts#4979

1. Ensure you have sales price rules with product attribute conditions configured in the admin panel.
2. Navigate to the storefront and add products to the cart so that sales rules are evaluated.
3. Verify that the correct sales rules still apply and discounts are calculated as expected.
4. Enable MySQL query logging or use a profiler to confirm that the query against `salesrule_product_attribute` / `eav_attribute` now uses the subselect pattern and executes faster than before.
5. Verify that the `catalog_product_collection_load_before` event still correctly adds the required product attributes to the collection select.

### Questions or comments

The `$websiteId` and `$customerGroupId` parameters were removed from `getActiveAttributes()` because they were never actually used in the SQL query — the method always returned all attribute codes regardless of website or customer group.

### DIFF
--- a/app/code/core/Mage/SalesRule/Model/Observer.php
+++ b/app/code/core/Mage/SalesRule/Model/Observer.php
@@ -229,8 +229,8 @@ class Mage_SalesRule_Model_Observer
     /**
      * Remove catalog attribute condition by attribute code from rule conditions
      *
-     * @param Mage_Rule_Model_Condition_Combine|Mage_Rule_Model_Action_Collection $combine
-     * @param string $attributeCode
+     * @param Mage_Rule_Model_Action_Collection|Mage_Rule_Model_Condition_Combine $combine
+     * @param string                                                              $attributeCode
      */
     protected function _removeAttributeFromConditions($combine, $attributeCode)
     {


sreichel:
Hey, can you please update.

(and give a better title)

PR #5327, opened 2026-03-18
Cleanup: removed mysql4 classes

sreichel:
### Description (*)

- removed classes
- removed deprecatedNodes from XML
- removed deprecatedNode support
- removed calls to non-existing models

### DIFF
--- a/.phpstan.dist.baseline.php
+++ b/.phpstan.dist.baseline.php
@@ -2,12 +2,6 @@
 
 $ignoreErrors = [];
 $ignoreErrors[] = [
-	'rawMessage' => 'Call to an undefined method Mage_Core_Model_Abstract::addError().',
-	'identifier' => 'method.notFound',
-	'count' => 1,
-	'path' => __DIR__ . '/app/Mage.php',
-];
-$ignoreErrors[] = [
 	'rawMessage' => 'Construct empty() is not allowed. Use more strict comparison.',
 	'identifier' => 'empty.notAllowed',
 	'count' => 10,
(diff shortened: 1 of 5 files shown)


Hanmac:
i added extra Issue on the MM Repo:
https://github.com/empiricompany/openmage_ignition/issues/10

sreichel:
@Hanmac  openmage_ignition is installed by default. I think its okay to have it in phpstorm metafiles. (dont know why i have removed it in prev PR)

Can we proceed?

Hanmac:
> @Hanmac openmage_ignition is installed by default. I think its okay to have it in phpstorm metafiles. (dont know why i have removed it in prev PR)
> 
> Can we proceed?

For the ignition thing, i made an Issue in their repository.
So that they can fix it at the source, and we can remove them later again.

sreichel:
> Might remove the "_mysql4_" upgrade scripts in later MR?

Script?

Hanmac:
> > Might remove the "_mysql4_" upgrade scripts in later MR?
> 
> Script?

hm it seems they got removed with another MR before? I forgot to check first

but there is still some missing like this one:
https://github.com/OpenMage/magento-lts/blob/main/app/code/core/Mage/Core/sql/core_setup/mysql4-data-upgrade-1.6.0.2-1.6.0.3.php

sreichel:
most of the mysql4-scripts (everything lower 1.6?) have been removed. not sure about this one, but i kept it b/c its used. (?)

PR #5328, opened 2026-03-18
phpstan: update  `Mage::getModel()` docs

sreichel:
### Description (*)

`Mage::getModel()` will return an object, even its no child of Mage_Core_Model_Abstract.

Updated docs.

### DIFF
--- a/app/code/core/Mage/Core/Model/Config.php
+++ b/app/code/core/Mage/Core/Model/Config.php
@@ -1488,9 +1488,9 @@ class Mage_Core_Model_Config extends Mage_Core_Model_Config_Base
      *
      * Will instantiate Mage_Catalog_Model_Resource_Product
      *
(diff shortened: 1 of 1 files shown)


Hanmac:
I personally think its bad code-style if a model doesn't extend from `Mage_Core_Model_Abstract`

sreichel:
@Hanmac same for helpers ...

sreichel:
> I personally think its bad code-style if a model doesn't extend from `Mage_Core_Model_Abstract`

It tried ... extending MCMA works for 99% models, but leads to some incompatible method declarations for some classes.

Lets update helpers/models in another PR.  This one is obnly about to correct the docblocks.

sreichel:
> I personally think its bad code-style if a model doesn't extend from `Mage_Core_Model_Abstract`

Lets fix it later.

PR #5329, opened 2026-03-19
Visibility: remove deprecated functions

Hanmac:
### Related Pull Requests

1. ...
2. ...

### Questions or comments

@sreichel should the functions in `Mage::getSingleton('catalog/product_visibility')` like `getVisibleInCatalogIds()` be changed to static? if they return fixed array anyway?

### DIFF
--- a/app/code/core/Mage/Adminhtml/Model/Sales/Order/Random.php
+++ b/app/code/core/Mage/Adminhtml/Model/Sales/Order/Random.php
@@ -79,7 +79,7 @@ class Mage_Adminhtml_Model_Sales_Order_Random
                 ->addAttributeToFilter('status', [
                     'in' => Mage::getSingleton('catalog/product_status')->getVisibleStatusIds(),
                 ])
-                ->setVisibility(Mage::getSingleton('catalog/product_visibility')->getVisibleInSearchIds())
+                ->setVisibility(Mage::getSingleton('catalog/product_visibility')::getVisibleInSearchIds())
                 ->load();
         }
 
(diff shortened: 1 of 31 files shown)


sreichel:
Removing deprecated code is a very good idea. Imho we should keep methods that forward to their replacements.

We should be start removing code that is depecrated since 1.4/1.5 and can be considered as dead code. The visibility filters are used in older extensions for sure. It was often suggested on [stackoverflow](https://magento.stackexchange.com/a/211067/46249).

Hanmac:
@sreichel see https://github.com/MahoCommerce/maho/pull/699 where the functions got made static
(and called via `Mage_Catalog_Model_Product_Visibility::` instead)

sreichel:
@Hanmac change to static call is fine. My comment was only about not deleting the methods.

sreichel:
> @sreichel see [MahoCommerce/maho#699](https://github.com/MahoCommerce/maho/pull/699) where the functions got made static (and called via `Mage_Catalog_Model_Product_Visibility::` instead)

... not sure about replacing singleton pattern ... as you said here https://github.com/MahoCommerce/maho/pull/699#issuecomment-4090516386

I disagree with Fab ... he is right, changing/rewrite these methods could require more changes, but it keeps OM "rewriteable".

https://github.com/OpenMage/magento-lts/pull/5319/changes ... what bout this? Change these too?

Hanmac:
> We should be start removing code that is depecrated since 1.4/1.5 and can be considered as dead code. The visibility filters are used in older extensions for sure. It was often suggested on [stackoverflow](https://magento.stackexchange.com/a/211067/46249).

See this anwser:

> KeyurShah I use [github.com/OpenMage/magento-lts](https://github.com/OpenMage/magento-lts) (1.9.3.7) and there we possibly remove deprecated methods in the future. – [sv3n](https://magento.stackexchange.com/users/46249/sv3n) Commented Jan 26, 2018 at 9:19

So next time! These functions should be removed

sreichel:
> So next time! These functions should be removed

In general ... its time to discuss the strict BC promise. 

We always avoided to change method signatures (making static here), but we are limiting ourselfs. I'd like to see more typehinting, strict types ...

PR #3640, opened 2023-11-10
Improve performance for grids with permission checks

sreichel:
### Description (*)

Some grids (eg orders) check whether user is allowed to click/view ... but its valididated for each row.

With this PR its only checked once.

PR #5332, opened 2026-03-20
Improve performance for grids with permission checks

sreichel:
### Related Pull Requests

- see OpenMage/magento-lts#3640


sreichel:
@kiatng https://github.com/OpenMage/magento-lts/pull/3640#issuecomment-1806737471 ... later :)

PR #5345, opened 2026-03-23
Fix ignored phpstan errors

Hanmac:
### Related Pull Requests

1. ...
2. ...

### Questions or comments

> Comparison operation ">" between *NEVER* and 1 results in an error.

This happens because the options are filtered to `string[]` first.
Meaning that `is_array($option)` never is true.
And PHPStan causes a stupid error https://github.com/phpstan/phpstan/issues/14355

@sreichel does this mean that this part is currently broken because it can't be true?

### DIFF
--- a/app/code/core/Mage/Bundle/Model/Product/Type.php
+++ b/app/code/core/Mage/Bundle/Model/Product/Type.php
@@ -1008,7 +1008,7 @@ class Mage_Bundle_Model_Product_Type extends Mage_Catalog_Model_Product_Type_Abs
         $optionQty  = $buyRequest->getBundleOptionQty();
 
         $option     = (is_array($option)) ? array_filter($option, fn(mixed $o) => (int) $o !== 0) : [];
-        $optionQty  = (is_array($optionQty)) ? array_filter($optionQty, fn(mixed $o) => (float)$o !== 0.0) : [];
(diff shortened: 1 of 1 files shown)


Hanmac:
> Deprecated: ctype_digit(): Argument of type array will be interpreted as string in the future

Ugh i got noticed that ctype_digit also doesn't like possible array type
might need to find a better way to check


the Code previously was using `intval` which works interestingly with arrays:
https://3v4l.org/59Gnt#v8330

but phpstan didn't like that:
https://phpstan.org/r/34ca0d36-d6c0-4506-aff9-06078849ba62
that's why i tried to change it

sreichel:
Tried to change it month ago ... imho there is nothing wrong with `\inval(...)` or `intval` (as string param). Should be fixed at phpstan side.

Hanmac:
> Tried to change it month ago ... imho there is nothing wrong with `\inval(...)` or `intval` (as string param). Should be fixed at phpstan side.

[intval](https://www.php.net/manual/en/function.intval.php) Page:
> Empty arrays return 0, non-empty arrays return 1. 

imo this could be problematic


another way would be to use `fn($o) => \intval($o) !== 0` instead

Hanmac:
@sreichel this should be ready now

Hanmac:
`fn($o) => (bool)\intval($o)` could have worked too, but `!== 0` is more explicit

PR #5366, opened 2026-03-27
chore: deprecated replaced Zend classes

sreichel:
Added stub files for phpstan.

### DIFF
--- a/dev/phpstan/stubs/Zend/Log.stub
+++ b/dev/phpstan/stubs/Zend/Log.stub
@@ -1,6 +1,6 @@
 <?php
 
 /**
- * @deprecated Use Mage_Core_Model_Logger instead; it uses Monolog as the underlying implementation.
+ * @deprecated use Mage_Core_Model_Logger instead; it uses Monolog as the underlying implementation
  */
 class Zend_Log {}


dev/phpstan/stubs/Zend/Measure.stub
```diff
@@ -0,0 +1,6 @@
+<?php
+
+/**
+ * @deprecated Use Mage_Core_Helper_Measure_Length and Mage_Core_Helper_Measure_Weight instead.
```
sreichel:
```suggestion
 * @deprecated use Mage_Core_Helper_Measure_Length and Mage_Core_Helper_Measure_Weight instead
```

dev/phpstan/stubs/Zend/Log.stub
```diff
@@ -1,6 +1,6 @@
 <?php
 
 /**
- * @deprecated use Monolog instead
+ * @deprecated Use Mage_Core_Model_Logger instead; it uses Monolog as the underlying implementation.
```
sreichel:
```suggestion
 * @deprecated use Mage_Core_Model_Logger instead; it uses Monolog as the underlying implementation
```

PR #5368, opened 2026-03-28
rector: use constants for store config paths

sreichel:
### Description (*)

This is something i wanted do for a very long time ... replace store config paths string with its XML path constant.

Its simple search and replace, but its annoying. Added rectore rule for replacements.

On hold till next rector release.

- see https://github.com/rectorphp/rector/issues/9713

### Related Pull Requests

- see OpenMage/magento-lts#5333


sreichel:
Q: move constants from helper to model?

sreichel:
See https://github.com/OpenMage/magento-lts/pull/5444

PR #5381, opened 2026-03-30
chore: Mini docblock update

sreichel:
Mini docblock update


app/design/frontend/rwd/default/template/wishlist/item/list.phtml
```diff
@@ -16,8 +16,7 @@
     <thead>
         <tr>
             <?php
-                /* @var $column Mage_Wishlist_Block_Customer_Wishlist_Item_Column */
-                foreach ($columns as $column):
+            foreach ($columns as $column):
```
sreichel:
```suggestion
                foreach ($columns as $column):
```

PR #5403, opened 2026-04-02
fix: invalid type for `getStore()`

sreichel:
### Description (*)

Revert. Wrong and not working.

Thanks @Hanmac 

### Related Pull Requests

- see OpenMage/magento-lts#5090


Hanmac:
The problem is `protocolDirective`, it should not have called `getSafeStore` in the first place

https://github.com/OpenMage/magento-lts/blob/cc9de00eb06a7768680db8f1a120bcdd8ad42934/app/code/core/Mage/Core/Model/Email/Template/Filter.php#L441-L446

```php
$isSecure = Mage::app()->getStore($params['store'] ?? null)->isCurrentlySecure();
```

anything else crashes anyway in an Exception sooner or later in the code

PR #5406, opened 2026-04-02
phpstan: use `@phpstan-type` for config store-id param

sreichel:
... to not repeat ...

@Hanmac only an idea ... same could be used for escape-methods ...


sreichel:
Im pretty sure there are more files that use configs store-id ... this was 1st search & replace, more later (?)

PR #5413, opened 2026-04-03
rector: `CompleteMissingIfElseBracketRector`

sreichel:
- see https://getrector.com/rule-detail/complete-missing-if-else-bracket-rector

wait for ...

- see https://github.com/rectorphp/rector/issues/9725
- see https://github.com/rectorphp/rector/issues/9726


sreichel:
No changes with 2.4.1

PR #5415, opened 2026-04-04
rector: `RemoveUnusedVariableAssignRector` (no automatic changes)

sreichel:
- see https://getrector.com/rule-detail/remove-unused-variable-assign-rector

Rector only removes method var, not the method call ... (thanks)

### Related Pull Requests

- see OpenMage/magento-lts#5392

### DIFF
--- a/app/code/core/Mage/Catalog/Model/Product/Type.php
+++ b/app/code/core/Mage/Catalog/Model/Product/Type.php
@@ -224,7 +224,7 @@ class Mage_Catalog_Model_Product_Type
 
             $types = self::getTypes();
             foreach ($types as $typeId => $typeInfo) {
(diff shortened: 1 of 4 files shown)


app/code/core/Mage/Core/Model/Resource/Translate/String.php
```diff
@@ -207,7 +207,8 @@ public function saveTranslate($string, $translate, $locale = null, $storeId = nu
         if ($row = $write->fetchRow($select, $bind)) {
             $original = $string;
             if (str_contains($original, '::')) {
-                [$scope, $original] = explode('::', $original);
+                // ignorerd scope
```
sreichel:
```suggestion
                // ignored scope
```

PR #5416, opened 2026-04-04
rector: `TernaryFalseExpressionToIfRector`

sreichel:
- see https://getrector.com/rule-detail/ternary-false-expression-to-if-rector

wait for https://github.com/rectorphp/rector/issues/9728


sreichel:
No changes with 2.4.1

PR #5434, opened 2026-04-08
rector: `ArrayToFirstClassCallableRector`

sreichel:
- see https://getrector.com/rule-detail/array-to-first-class-callable-rector


sreichel:
Wait for https://github.com/phpstan/phpstan-src/pull/5474

Hanmac:
Good thing that the new `validateColumnFilterCallback` isn't final, if I really want to add some callbacks that are not part of the Grid

Hanmac (dismissed):

Hanmac:
@sreichel try `callable-array` ?
But that seems to have a problem itself anyway:
https://github.com/phpstan/phpstan/issues/3842
https://github.com/phpstan/phpstan-src/pull/5548

PR #5457, opened 2026-04-10
chore: ref #5456

sreichel:


app/code/core/Mage/Usa/Model/Shipping/Carrier/Usps.php
```diff
@@ -152,16 +152,16 @@ public function setRequest(Mage_Shipping_Model_Rate_Request $request)
             $rawRequest->setService('ALL');
         }
 
-        $container = $request->getUspsContainer() ? $request->getUspsContainer() : $this->getConfigData('container');
-
+        $requestContainer = $request->getUspsContainer();
+        $container = $requestContainer ? $requestContainer : $this->getConfigData('container');
```
Hanmac, line 156: Our rector doesn't like 
`$request->getUspsContainer() ?? $this->getConfigData('container')`, right?
or was that only for `?:`?

sreichel:
`??` is fine, only `?:` should be avoided

app/code/core/Mage/Usa/Model/Shipping/Carrier/Usps.php
Hanmac: @sreichel would you be okay with writing it like this?

sreichel: > Disallow short ternary operator (?:) - implies weak comparison, it's recommended to use null coalesce operator (??) or ternary operator with strict condition.

https://github.com/phpstan/phpstan-strict-rules

:)

sreichel: > @sreichel would you be okay with writing it like this?

would it be okay to fix 55 baselined errors too? :sunglasses:

Hanmac: I would prefer `??` if able instead of writing such parameters twice

sreichel: Id like it ... using var is my bad habbits :)

sreichel:
@Hanmac new PR?

PR #5472, opened 2026-04-14
Mage_Widget: fix placeholder images not respecting possible SkinDir

### LINKED ISSUE #5359: [Bug] Widget Placeholder got broken by Admin Design change
See this part:
https://github.com/OpenMage/magento-lts/blob/c6f4014a9d1f7f893952c549c9a9b7860f74ff4c/app/code/core/Mage/Widget/Model/Widget/Config.php#L48-L73

and this part:
https://github.com/OpenMage/magento-lts/blob/c6f4014a9d1f7f893952c549c9a9b7860f74ff4c/app/code/core/Mage/Widget/Model/Widget.php#L242-L246

the images are now under `skin/adminhtml/base/default/images/widget/cms__widget_block.gif`
but the SkinBaseDir for openmage design is `skin/adminhtml/openmage/default/` and it doesn't fall back to parent design

It should not use getBaseDir and getBaseUrl, but uses the real `getSkinUrl` with fallback.

I just don't know yet what would be the best way to check if file exists and if not, use `default.gif`
(because some checks are done in JS)

Hanmac:
### Description (*)

This fixes the problem that `images/widget` URLs and Filenames were ignoring other Skin Layout Updates.

First i use glob to get available images in `*/*/images/widget/*.gif`, take their basenames and filter for unique entries.
Then I use `getSkinUrl` to get the correct URL for JS. This causes it to fallback to the base/default Skin.

### Related Pull Requests

- fixes OpenMage/magento-lts#5359

1. ...
2. ...


Hanmac:
@sreichel what do you think of this fix?

sreichel:
Not sure ..

<s>Do we need a fallback here? Would it be easier to add a placeholder image instead?</s>

How about https://github.com/symfony/finder?

Too much overhead?

Hanmac:
Hm, another way would be to go through the theme folders:
like A → B → C and get the files there.

But i thought one big glob might be faster.

(there is a con right now, that if in a totally unrelated theme D,
there is another widget, it would try to find it, but then would fail on the Skin URL)

It was using `Varien_Data_Collection_Filesystem`, but i didn't think it works with glob paths.
(`Mage_Cms_Model_Wysiwyg_Images_Storage_Collection` is the only other place that uses this Collection)

I don't know if `symfony/finder` would be so much of a difference? Might be too much if I only want to check some file names?

Hanmac:
@sreichel is this ready, or should I try to address the corner case with unrelated image in template D?

sreichel:
lgtm

sreichel (approved):

PR #5476, opened 2026-04-16
[PHP83] rector: added `Deprecated` php attribute

sreichel:
- added Deprecated Attribute (https://php.watch/versions/8.4/Deprecated)

### Note:

With php84+ and dev-mode on using a deprecated method/class will throw an exception.


sreichel:
Wait for php83 and phpunit12.

PR #5477, opened 2026-04-17
Downgraded composer/composer

sreichel:
### Description (*)

- see https://github.com/advisories/GHSA-wg36-wvj6-r67p
- see https://github.com/Cotya/magento-composer-installer/issues/212

### Questions or comments

Remove magento-composer-installer?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
     },
     "require-dev": {
         "ext-xmlreader": "*",
-        "composer/composer": "^2.8",
+        "composer/composer": ">= 2.2.27 <2.3 || >= 2.9.6",
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0.0",
         "friendsofphp/php-cs-fixer": "^3.6",
         "macopedia/phpstan-magento1": "^1.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "416833ac523a8f6849aea118fa0c1c2b",
+    "content-hash": "d1874cd06be8c1e9f4ef1237444ca19b",
     "packages": [
         {
             "name": "carbonphp/carbon-doctrine-types",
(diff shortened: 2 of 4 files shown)


Hanmac:
@sreichel is the failed test something we can fix there? or for another MR?

sreichel:
Another PR. Had no time to check why it fails ....

Hanmac:
Older composer version https://packagist.org/packages/composer/composer#2.2.27
depends on [react/promise](https://packagist.org/packages/react/promise): ^1.2 || ^2.7

this fix https://github.com/reactphp/promise/commit/501b4aa15121cd015d9f6f548cbda4c27bc16ced
was merged into 2.x, but there isn't a release with that fix yet:
https://github.com/reactphp/promise/issues/274

sreichel:
> Remove magento-composer-installer?

...

Hanmac:
> Remove magento-composer-installer?

This just moves the problem to the next repos
while Magento itself doesn't need `magento-composer-installer`,
what happens if you use other Magento dependencies via composer?
I assume they fail without it, right?


Edit:
i found this:
https://github.com/composer/installers

but that's only for:

* magento-library
* magento-skin
* magento-theme

sreichel:
Support for themes, libs, etc can be added to https://github.com/OpenMage/composer-plugin

Also ... instead of copy/symlink files to app\code we can load them from vendor. *thinking loud*

sreichel:
> Older composer version https://packagist.org/packages/composer/composer#2.2.27 depends on [react/promise](https://packagist.org/packages/react/promise): ^1.2 || ^2.7
> 
> this fix [reactphp/promise@501b4aa](https://github.com/reactphp/promise/commit/501b4aa15121cd015d9f6f548cbda4c27bc16ced) was merged into 2.x, but there isn't a release with that fix yet: [reactphp/promise#274](https://github.com/reactphp/promise/issues/274)

We can temp fix it via composer patch ....

Hanmac:
> Also ... instead of copy/symlink files to app\code we can load them from vendor. _thinking loud_

Sounds interesting, I wonder how complicated it gets for non-php files like skin or locale 🤔

ajbonner:
I'd love to see the back of composer installer... but I think it's been so embedded in extension distribution there would need to be some robust scaffolding around its removal. 

One point to raise when thinking about 'vendoring' extensions - code pool overrides. A lot of extensions take advantage of code pool shadowing, so replacing wholesale a app/code/core/Mage_xxx or lib/Varien class (we do that one a lot!!) path with app/code/{community,local} would lose that behaviour and autoloading would need to take that into account.

That's doable though.

sreichel:
@ajbonner its planned, but not for today or tomorrow. :)  Lets brainstorm a later time.

PR #5484, opened 2026-04-18
fix: invalid compare `strlen() < 0`

sreichel:
Or more strict ... 

```php
        if (!is_string($actionsXML) || $actionsXML === ''
            || @simplexml_load_string('<data>' . $actionsXML . '</data>', null, LIBXML_NOERROR) === false
        ) {
```

### DIFF
--- a/app/code/core/Mage/Dataflow/Model/Profile.php
+++ b/app/code/core/Mage/Dataflow/Model/Profile.php
@@ -87,8 +87,7 @@ class Mage_Dataflow_Model_Profile extends Mage_Core_Model_Abstract
     {
         parent::_beforeSave();
         $actionsXML = $this->getDataByKey('actions_xml');
-        // @phpstan-ignore-next-line because of https://github.com/phpstan/phpstan/issues/10570
-        if ($actionsXML !== null && strlen($actionsXML) < 0
+        if (is_string($actionsXML) && $actionsXML !== ''
             && @simplexml_load_string('<data>' . $actionsXML . '</data>', null, LIBXML_NOERROR) === false
         ) {
             Mage::throwException(Mage::helper('dataflow')->__('Actions XML is not valid.'));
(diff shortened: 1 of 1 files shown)


Hanmac:
I haven't checked the code yet, can actions_xml be null?

And if yes is that allowed?

sreichel:
> I haven't checked the code yet, can actions_xml be null?

Can be anything?

`$this->setData('actions_xml', mixed)`

Hanmac:
I think we should apply the copilot suggestion and checking for non-null non-string too

sreichel:
> for non-null non-string

non-null? So throw exception on null, or not?

This was my question ...

```php
if (!is_string($actionsXML) || $actionsXML === '' ||
```

... requires a non-empty, valid xml string. Or do i miss something?

Hanmac:
I meant more in a way that allows null values

right now these are allowed too

app/code/core/Mage/Dataflow/Model/Profile.php
```diff
@@ -87,8 +87,7 @@ protected function _beforeSave()
     {
         parent::_beforeSave();
         $actionsXML = $this->getDataByKey('actions_xml');
-        // @phpstan-ignore-next-line because of https://github.com/phpstan/phpstan/issues/10570
-        if ($actionsXML !== null && strlen($actionsXML) < 0
+        if (is_string($actionsXML) && $actionsXML !== ''
```
sreichel:
```suggestion
        if ($actionsXML !== null && is_string($actionsXML) && $actionsXML !== ''
```

@Hanmac

Hanmac: 👍

sreichel:
Rdy.

Hanmac (approved):

PR #5534, opened 2026-04-24
cypress: add `data-test` selectors to admin widget buttons

sreichel:
### Related Pull Requests

- see OpenMage/magento-lts#5308
- see OpenMage/magento-lts#5515

Use ...

```php
$this->getLayout()->createBlock('adminhtml/widget_button')
    ->setData([
        # ...
        'test_id'        => 'add', #new
    ]),
```


app/code/core/Mage/Adminhtml/Block/Widget/Button.php
```diff
@@ -38,6 +38,15 @@ public function getOnClick()
         return $this->getDataByKey('on_click');
     }
 
+    public function getTestId(): ? string
```
sreichel:
```suggestion
    public function getTestId(): ?string
```

PR #5551, opened 2026-04-29
phpstan: variable.implicitArray

sreichel:
### Description (*)

Minor phpstan fixes.


app/code/core/Mage/Catalog/Block/Layer/State.php
```diff
@@ -53,10 +53,12 @@ public function getClearUrl()
             $filterState[$filter->getRequestVar()] = $filter->getCleanValue();
         }
 
+        $params                 = [];
         $params['_current']     = true;
         $params['_use_rewrite'] = true;
         $params['_query']       = $filterState;
         $params['_escape']      = true;
+
```
Hanmac, line 61: Wouldn't it be better to have `$params = [ ... ]`?

with `[ ... ]` being a Multiline Array?

Or better `return Mage::getUrl('*/*/*', [ ... ]);`

sreichel: I think this will be to much work.

Lets keep it simple.

Hanmac: @sreichel i meant to write the params array like this:

```php
$params = [
    '_current'     => true,
    '_use_rewrite' => true,
    '_query'       => $filterState,
    '_escape'      => true,
];

```

or like this:
```php
return Mage::getUrl('*/*/*', [
    '_current'     => true,
    '_use_rewrite' => true,
    '_query'       => $filterState,
    '_escape'      => true,
]);
```

i found a rector for similar stuff, but not quite the right thing:
https://getrector.com/rule-detail/inline-array-return-assign-rector

Edit: i asked if there is an extra rule: https://github.com/rectorphp/rector/discussions/9747
there isn't one right now

sreichel: Yeah, to automate it, but atm i dont wont to spent much time for writing rules.

(there should be some project/OM-related rules first e.g. change to `getDataById()`)

btw ... we also have CodeSniffer/Beatifier, cs-fixer with additional ECS rules, but no one covers this

PR #5578, opened 2026-05-05
Fix wrong DocBlock and getter in adminhtml grid classes

B3Hana:
### Description (*)
The base getRowUrl function tried to access a parent method, which doesn't exist, resulting in a getData magic call.
But because the $row (Model) is used as an index, it would result in an error in Varien_Object if you set the `row_url` as data in the grid.

The setRenderer function of the column should expect a renderer, not a string, as the getRenderer call is chained with function calls of the `Mage_Adminhtml_Block_Widget_Grid_Column_Renderer_Abstract `. It is also the return type of the getter.

### Related Pull Requests
-

-

Don't override the getRowUrl in a grid and set `$this->setData('row_url', 'url');`, it will result in an illegal offset error.

### Questions or comments
-

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Grid.php
@@ -2038,7 +2038,7 @@ class Mage_Adminhtml_Block_Widget_Grid extends Mage_Adminhtml_Block_Widget
      */
     public function getRowUrl($row)
     {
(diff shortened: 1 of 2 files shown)


Hanmac (approved):

sreichel (approved):

sreichel:
Thx for fixing parrent call ... skip that dozen times.

PR #5598, opened 2026-05-21
dependency: symfony/mime v6.4.40

Hanmac:
Because Dependabot didn't want to do that, so I did myself

This fixes the "PHP Security Check / check" that is blocking the other MRs


Hanmac:
@sreichel merging this early fixes all the other MRs

PR #5601, opened 2026-05-22
phpstan: checks around getConfig + asArray

Hanmac:
### Description (*)

Some fixes regarding:
* `getConfig()` which might return null when called as static,
* `getNode()` which might return false,
* `asArray()` which might return string and not array. (says PHPStan)

### Related Pull Requests

- see https://github.com/OpenMage/magento-lts/pull/5590#issuecomment-4515216357

1. ...
2. ...


app/code/core/Mage/Payment/Model/Config.php
```diff
@@ -98,9 +98,19 @@ protected function _getMethod($code, $config, $store = null)
...
+        $_types = Mage::getConfig()->getNode('global/payment/cc/types');
 
-        uasort($_types, ['Mage_Payment_Model_Config', 'compareCcTypes']);
+        if ($_types === false) {
+            return [];
+        }
+
+        $_types = $_types->asArray();
+
+        if (is_string($_types)) {
```
sreichel:
```suggestion
        if (!is_array($_types)) {
```

app/code/core/Mage/Api2/Helper/Data.php
```diff
@@ -104,7 +110,12 @@ public function getUserTypes()
      */
     public function getRequestInterpreterAdapters()
     {
-        return (array) Mage::app()->getConfig()->getNode(self::XML_PATH_API2_REQUEST_INTERPRETERS);
+        $node = Mage::app()->getConfig()->getNode(self::XML_PATH_API2_REQUEST_INTERPRETERS);
+        if (!$node instanceof Varien_Simplexml_Element) {
+            return [];
+        }
+
+        return (array) $node;
```
sreichel: Whats the  idea?

Hanmac: In case getNode doesn't return the Varien_Simplexml_Element, but string instead (if it has no children)
probably empty string `''`.

`(array) $node` would have turned that into `['']` instead of empty array as expected.

Edit: ups, that was the problem with `asArray`, not with `getNode`

app/code/core/Mage/Payment/Model/Config.php
```diff
@@ -98,9 +98,19 @@ protected function _getMethod($code, $config, $store = null)
      */
     public function getCcTypes()
     {
-        $_types = Mage::getConfig()->getNode('global/payment/cc/types')->asArray();
+        $_types = Mage::getConfig()->getNode('global/payment/cc/types');
 
-        uasort($_types, ['Mage_Payment_Model_Config', 'compareCcTypes']);
+        if ($_types === false) {
```
sreichel: instanceof?

lib/Varien/Data/Collection/Db.php
```diff
@@ -899,7 +899,7 @@ public function addFilterToMap($filter, $alias, $group = 'fields')
     {
         if (is_null($this->_map)) {
             $this->_map = [$group => []];
-        } elseif (is_null($this->_map[$group])) {
+        } elseif (!isset($this->_map[$group])) {
```
sreichel: array_key_exists (+ is_null)?

lib/Varien/Data/Collection/Db.php
```diff
@@ -66,7 +66,7 @@ class Varien_Data_Collection_Db extends Varien_Data_Collection
     /**
      * Fields map for correlation names & real selected fields
      *
-     * @var null|array{fields: array<string, string>}
+     * @var null|array<string, array<string, string|Zend_Db_Expr>>
```
sreichel, line 69: Can you try to keep the named field?

(not sure why atm, but i did it for some reason?)

lib/Varien/Data/Collection/Db.php
Hanmac: At this place, checking for `array_key_exists` alone is enough, the function wants to set `$group` if it doesn't exist.

Checking for null should not be needed?

lib/Varien/Data/Collection/Db.php
Hanmac: > (not sure why atm, but i did it for some reason?)

You got baited by `addFilterToMap` having `$group = 'fields'` as default parameter.

Hanmac:
Stupid PHPStrom, lied to me that in `&& (bool) (int)`, the `(bool)` was unnecessary -,-

sreichel:
`Pls dont add too much. Better smaller PRs.

Hanmac:
@sreichel yeah, that should be enough for now in Mage_Payment_Helper

sreichel:
imho ... its already to big ... w/ various changes.

DocBlock fixex are worth a small PR ...

PR #5675, opened 2026-07-08
Phpunit10

### LINKED ISSUE #5674: PHPUnit10 setMethods
Also, PHPUnit >= 10 has the problem with `setMethods`
https://github.com/OpenMage/magento-lts/blob/997ce9dd540934b6a855e0cd98de5921d77a8224/tests/unit/OpenMageTest.php#L36-L38
we need to split it between `onlyMethods` and `addMethods`.
But `addMethods` is deprecated too:
https://github.com/sebastianbergmann/phpunit/issues/5320

Example:
https://github.com/OpenMage/magento-lts/blob/997ce9dd540934b6a855e0cd98de5921d77a8224/tests/unit/Mage/Admin/Model/BlockTest.php#L43-L46

These functions can't be mocked with `onlyMethods`, because they don't exist:
```php
[
                'getBlockName' => 'test/block',
                'getIsAllowed' => '1',
]
```
So we either need to add them as Interfaces, #5104
Or we need to mock `__call` instead, which is ugly in itself

example with `__call` mock:
```php
$mock = $this->createPartialMock(Subject::class, ['__call']);
$mock->method('__call')->willReturnCallback(fn(string $method) => $methods[$method]);
```

Do we really need to use MockObject there?
Wouldn't it be better to use real Object and use `setData` to set the values?
Like this, with updated `provideValidateAdminBlockData`
```php
self::$subject->setData($data);
```

_Originally posted by @Hanmac in https://github.com/OpenMage/magento-lts/issues/5650#issuecomment-4913604865_

Hanmac:
### Description (*)

Update to PHPUnit 10 for later MRs.

### Related Pull Requests

- fixes OpenMage/magento-lts#5674

1. ...
2. ...

### DIFF
--- a/tests/unit/Mage/Oauth/Model/ConsumerTest.php
+++ b/tests/unit/Mage/Oauth/Model/ConsumerTest.php
@@ -20,7 +20,6 @@ use OpenMage\Tests\Unit\OpenMageTest;
 
 final class ConsumerTest extends OpenMageTest
 {
-    // @phpstan-ignore property.onlyWritten
     private static Subject $subject;
 
     #[Override]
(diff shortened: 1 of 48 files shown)


Hanmac:
@sreichel for some reason, when i test `EnvironmentConfigLoaderTest` local on my machine, i don't get any errors?

tried everything, i can't reproduce it

Hanmac:
@sreichel looks like the only thing breaking are the empty testcases

because:
https://github.com/sebastianbergmann/phpunit/commit/7b7b5dd8d2a381a8bf3ad35cf4ce919316221f3e

Hanmac:
I finally found the option to disable the warnings and make it pass 😂

sreichel (dismissed):

Hanmac:
Reset CatalogRule/SalesRule changes, that looks like a bigger bug

tests/unit/Mage/Oauth/Model/ConsumerTest.php
```diff
@@ -33,26 +32,24 @@ public static function setUpBeforeClass(): void
...
     }
 
-    public function validateDataProvider(): Generator
+    public static function validateDataProvider(): Generator
     {
         $validData = [
-            'setKey'    => str_repeat('x', 32),
-            'setSecret' => str_repeat('x', 32),
+            'key'    => str_repeat('x', 32),
+            'secret' => str_repeat('x', 32),
```
Hanmac, line 54: 1) i messed up the data below, it should be the new array keys
2) the test code isn't correctly looking for an exception as phpunit wants them. Need to rewrite them.
3) because of the ugliness, it says it is duplicated via TokenTest. But that's just Sonar

Hanmac:
@sreichel @addison74 I finally got the Sonar Code Scan clean

addison74 (approved):

sreichel (approved):

sreichel:
❤️

PR #5688, opened 2026-07-15
Prevent override of existing buttons in getButtonsHtml

B3Hana:
### Description (*)
`Mage_Adminhtml_Block_Widget_Container` lets you add buttons with a custom `sort_order`.
This means it is possible for multiple buttons to get added with the same `sort_order` by different modules.

The `getButtonsHtml` function doesn't take this case into account and overrides a previously added button with another button, if it has the same `sort_order`.

### Related Pull Requests
-

-

Add 2 buttons with the same custom `sort_order`, the later one will replace the first.

### Questions or comments
If the increment of a button exceeds an already defined button, it will be placed after it, instead of before.
example: 
`button_a` with sort_order `19`
`button_b` with sort_order `20`
`button_c` with sort_order `19` will be `21`, after `button_b`

### DIFF
--- a/app/code/core/Mage/Adminhtml/Block/Widget/Container.php
+++ b/app/code/core/Mage/Adminhtml/Block/Widget/Container.php
@@ -232,8 +232,14 @@ class Mage_Adminhtml_Block_Widget_Container extends Mage_Adminhtml_Block_Templat
         foreach ($this->_buttons as $cachedButtons) {
             $buttons = [];
             foreach ($cachedButtons as $buttonId => $data) {
-                $buttons[$data['sort_order']]['buttonId'] = $buttonId;
-                $buttons[$data['sort_order']]['data'] = $data;
+                $sortOrder = $data['sort_order'];
+
+                while (isset($buttons[$sortOrder])) {
+                    $sortOrder++;
+                }
+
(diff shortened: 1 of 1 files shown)


addison74:
LGTM. The change fixes an issue where buttons with identical ```sort_order``` values overwrite each other because the array is keyed by sort_order. Incrementing the key until an unused slot is found preserves all buttons while keeping the existing rendering logic intact.

The only optional improvement would be to avoid missing sort_order values, for example:
```php
$sortOrder = (int)($data['sort_order'] ?? 0);
```
if there is any chance the key is not always present.

Hanmac:
@addison74 sort_order is always defined there:

https://github.com/OpenMage/magento-lts/blob/965dbd2ffd4e34526aba006085e8918da273b1e7/app/code/core/Mage/Adminhtml/Block/Widget/Container.php#L82

addison74:
In this case, I see no problem in not approving this PR.

addison74 (approved):

Hanmac (approved):

Hanmac:
A different way would be to use `uasort` to sort `$cachedButtons`, like:
```php
uasort($cachedButtons, fn($a, $b) => $a['sort_order'] <=> $b['sort_order']);
```

PR #5709, opened 2026-07-27
Fix MariaDB 12.3 compatibility in cart

Caprico85:
### Description (*)
MariaDB 12.3 adds `TO_DATE()` as a function. See [here](https://mariadb.com/docs/release-notes/community-server/12.3/mariadb-12.3-changes-and-improvements#compatibility-features), first line in "Compatibility Features"..

This collides with the unquoted column name `to_date` in app/code/core/Mage/SalesRule/Model/Resource/Rule/Collection.php.

### Related Pull Requests
none

none

Add a product to cart. On MariaDB 12.3, opening the cart will result in:
> SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'is null or to_date >= '2026-07-27') AND (`is_active` = '1') AND (`main_table`...' at line 2, query was: SELECT `main_table`.* FROM `vn_salesrule` AS `main_table` INNER JOIN `vn_salesrule_customer_group` AS `customer_group_ids` ON main_table.rule_id = customer_group_ids.rule_id AND customer_group_ids.customer_group_id = 1 WHERE (EXISTS (SELECT 1 FROM `vn_salesrule_website` AS `website` WHERE (website.website_id IN (1)) AND (main_table.rule_id = website.rule_id))) AND (from_date is null or from_date <= '2026-07-27') AND (to_date is null or to_date >= '2026-07-27') AND (`is_active` = '1') AND (`main_table`.`coupon_type` = '1') ORDER BY sort_order ASC

### Questions or comments
I didn't check everything for MariaDB 12.3 compatibility. So far, I only noticed this issue with 12.3.

 - [X] All automated tests passed successfully (all builds are green, SonarCloud checks are not required to merge)

### DIFF
--- a/app/code/core/Mage/SalesRule/Model/Resource/Rule/Collection.php
+++ b/app/code/core/Mage/SalesRule/Model/Resource/Rule/Collection.php
@@ -146,8 +146,8 @@ class Mage_SalesRule_Model_Resource_Rule_Collection extends Mage_Rule_Model_Reso
                     ),
                     [],
                 )
-                ->where('from_date is null or from_date <= ?', $now)
-                ->where('to_date is null or to_date >= ?', $now);
+                ->where('`from_date` is null or `from_date` <= ?', $now)
+                ->where('`to_date` is null or `to_date` >= ?', $now);
 
             $this->addIsActiveFilter();
 


Hanmac:
Maybe use `prepareSqlCondition` ? Or would that be overkill?
With `quoteIdentifier` it would look like this:

```php
->where($connection->prepareSqlCondition($connection->quoteIdentifier('from_date'), [
    ['null' => true], ['to' => $now],
]))
->where($connection->prepareSqlCondition($connection->quoteIdentifier('to_date'), [
    ['null' => true], ['from' => $now],
]));
```

addison74 (approved):

Hanmac:
@addison74 your opinion?
prepareSqlCondition would be too much, right?

addison74:
I think ```prepareSqlCondition()``` is a bit too heavy for this one. Since the break is specifically **to_date** becoming a function name in MariaDB 12.3 I’d keep it simple and just quote the date columns in this query. No issues found from Copilot review. That should fix the issue with minimal risk. Let's see what others think.

Caprico85:
I think it's a matter of preference. `prepareSqlCondition` is the "cleaner", more framework-oriented solution. It's affected by any future framework changes, e.g, if MySQL/MariaDB changes its query syntax (if that will ever happen). But it takes longer to understand the logic, especially the `OR` condition. That's why I used the string with backticks as a minimal change. I can change it to `prepareSqlCondition`, that's not a problem.

Hanmac (approved):
I think good enough for now?

PR #5757, opened 2026-09-01
fix(security): deny ACL resources that have no explicit rule for a role

mattdavenport:
## Problem

`Mage_Admin_Model_Resource_Rules::saveRel()` writes one explicit `allow`/`deny` row per ACL
resource when a role is saved. A resource added to `adminhtml.xml` after that save has no row.
`Mage_Admin_Model_Config::loadAclResources()` registers every resource with its parent, and
`Zend_Acl::isAllowed()` walks the resource parents, so the new resource inherits the nearest
ancestor's rule. A role with `allow` on `admin/system` silently gains every resource a later
module adds under `admin/system`.

Reproduction (no DB needed):

```php
$acl = new Mage_Admin_Model_Acl();
$acl->add(new Mage_Admin_Model_Acl_Resource('admin'));
$acl->add(new Mage_Admin_Model_Acl_Resource('admin/system'), 'admin');
$acl->add(new Mage_Admin_Model_Acl_Resource('admin/system/newthing'), 'admin/system');
$acl->addRole(new Mage_Admin_Model_Acl_Role_Group('G1'));
$acl->allow('G1', 'admin/system');
var_dump($acl->isAllowed('G1', 'admin/system/newthing')); // bool(true)
```

## Relation to #5254

#5254 (mine) removed the role editor's fallback that displayed inherited resources as checked.
Its description claimed the runtime "falls back to null/root, so users may not actually have
access". That claim was wrong: the runtime does inherit, and the fallback display was correct.
After #5254 the editor shows a new resource as unchecked while the role can use it.

This PR fixes the runtime instead of the display. With unlisted resources denied at load time,
the editor's unchecked box and the runtime agree, so #5254 does not need to be reverted.

## Change

`Mage_Admin_Model_Resource_Acl::loadAcl()` now calls `denyUnlistedResources()` after the
rules load. For every group role without an `all` **allow** row (`saveRel()` writes an `all` deny
row for every other role), every registered resource that has no explicit rule gets a `deny`. User roles inherit from their group through Zend_Acl role
inheritance, so only group roles need rules.

Cost: one `deny()` per (group role, unlisted resource) pair. Roles saved through the editor
have rows for almost every resource, so the count is small. The ACL is built at login and on
`reload_acl_flag`, not per request.

## Behavior change

A role that gained access to a resource only through inheritance loses it. That is the
intended effect; the role editor never showed or saved that access. Merchants who want it
open the role and tick the resource.

## Tests

`tests/unit/Mage/Admin/Model/Resource/AclTest.php` builds the ACL from in-memory role and
rule rows: unlisted child of an allowed parent is denied; explicit rows are unchanged; the
`all` allow row keeps everything; an `all` deny row does not; a user role follows its group's deny.

### DIFF
--- a/.phpstan.dist.baselines/argument.type.php
+++ b/.phpstan.dist.baselines/argument.type.php
@@ -47,16 +47,6 @@ $ignoreErrors[] = [
     'path' => __DIR__ . '/../app/code/core/Mage/Admin/Model/Resource/Acl.php',
 ];
 $ignoreErrors[] = [
-    'rawMessage' => 'Parameter #2 $rolesArr of method Mage_Admin_Model_Resource_Acl::loadRoles() expects array, array|null given.',
-    'count' => 1,
-    'path' => __DIR__ . '/../app/code/core/Mage/Admin/Model/Resource/Acl.php',
-];
-$ignoreErrors[] = [
-    'rawMessage' => 'Parameter #2 $rulesArr of method Mage_Admin_Model_Resource_Acl::loadRules() expects array, array|null given.',
-    'count' => 1,
-    'path' => __DIR__ . '/../app/code/core/Mage/Admin/Model/Resource/Acl.php',
-];
-$ignoreErrors[] = [
     'rawMessage' => 'Parameter #4 $assert of method Zend_Acl::allow() expects Zend_Acl_Assert_Interface|null, object|null given.',
     'count' => 2,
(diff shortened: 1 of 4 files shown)


Hanmac (approved):
The Code-Duplication is a false positive

Hanmac:
@mattdavenport now your MR is green (and I fixed some other PHPStan stuff)

addison74 (approved):

PR #5774, opened 2026-09-14
fix(security): use CSPRNG for uniqHash and Newsletter randomSequence

philipsorensen:
### Description

`Mage_Core_Helper_Data::uniqHash()` generated its value from
`md5(uniqid(microtime() . mt_rand(), true))`, and
`Mage_Newsletter_Model_Subscriber::randomSequence()` selected characters with
`mt_rand()`. Neither `uniqid()` nor `mt_rand()` is a cryptographically secure
random source, yet both feed security-sensitive tokens:

- `uniqHash()` produces the **admin and customer password-reset link tokens**
  (`Mage_Admin_Helper_Data::generateResetPasswordLinkToken()` and
  `Mage_Customer_Helper_Data::generateResetPasswordLinkToken()`), the **wishlist
  sharing code**, and the **newsletter subscriber confirmation key**
  (`Mage_Newsletter_Model_Resource_Subscriber`).
- `randomSequence()` produces the newsletter subscriber confirmation key on the
  model path.

This brings these generators in line with the CSPRNG the project already adopted
for `getRandomString()`, which uses `random_int()`.

Changes:

- `uniqHash()` -> `$prefix . bin2hex(random_bytes(16))`. The output shape is
  unchanged (the prefix followed by 32 hexadecimal characters), so callers that
  rely on the length/charset — including DOM id generation — are unaffected.
- `randomSequence()` -> `random_int()` for character selection.
- Extended the `uniqHash()` unit test to assert the 32-hex output shape and that
  successive calls do not collide.

This is defense-in-depth hardening: it removes predictable RNG from token
generation. There is no change to token format, storage, or verification.

### Related Pull Requests
- n/a

- n/a

1. Trigger the customer "Forgot Your Password?" flow; confirm the reset link
   still works end to end and `rp_token` is 32 hexadecimal characters.
2. Share a wishlist; confirm the sharing link still resolves to the shared list.
3. Subscribe to the newsletter; confirm the confirmation-email link still works.
4. Run `composer phpunit:test`; `Mage_Core_Helper_DataTest::testUniqHash` passes.

 - [x] All automated tests passed successfully (all builds are green, SonarCloud checks are not required to merge)


philipsorensen:
Addressed the Copilot review: un-skipped `SubscriberTest` and added a runnable `testRandomSequence()` that asserts the `[a-z0-9]` alphabet, default (32) and requested lengths, and non-collision across successive calls. `random_int()` needs no DB, so the model instantiates cleanly in the unit bootstrap. Passing locally (4 assertions).

addison74 (approved):

